PR #14, opened 2023-06-01
Move libtelio build jobs to GitHub actions

tomaszklak:
### Description

This adds build jobs for all supported architecture/os combinations excluding windows (which is done in parallel). 

The change to the old style toolchain file is due to the fact that github `actions-rs/toolchain` action only supports old style.

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated


dfetti:
+0.5

rust-toolchain.toml
```diff
@@ -1,10 +0,0 @@
-[toolchain]
```
packgron, line 1: Why was this removed ?

packgron: I would prefer to sill keep rust-toolchain.toml, targets and profile can be left out

packgron:
+1.5 nitpick about rust-toolchain file

rust-toolchain.toml
tomaszklak: @packgron I mention that in the description - basically the semi official actions in `action-rs` don't support new style of toolchain file and only works with the old style. I would prefer to limit the number of places where we have the version hardcoded and this way seemed best.

Alternatively I could have both toolchain file at the same time or (even more duplication) have only toml version and explicitly encoded the version across many invocations of actions in the build.yml

packgron: Oh I see, than let's keep this

PR #16, opened 2023-06-04
Move windows build job to github actions

jjanowsk:
### Description
Move windows build job to github actions
Additionally, in second commit I've added FFI tests to this workflow. In gitlab we used to run this test on release linux build in a `test` job which was fetching artifacts from `linux-build` job. In github it is possible but not trivial to fetch this artifact to the Tests workflow and I think it's not worth the hustle but I leave it for you to decide while I'm on vacation 🍹 One note about the second linker search path in `ffi/tests/CMakeLists.txt`. The old one can be removed once we remove this test from gitlab pipeline


ci/build_libtelio.py
```diff
@@ -33,38 +36,52 @@
...
-        wintun_path,
-        rutils.get_distribution_path(
-            config.target_os, config.arch, "wintun.dll", config.debug
-        ),
-    )
-    shutil.copyfile(
-        sqlite_path,
-        rutils.get_distribution_path(
-            config.target_os, config.arch, "sqlite3.dll", config.debug
-        ),
```
mathiaspeters, line 50: Is sqlite no longer necessary?

Jauler:
+1.5

mathiaspeters:
+1.0

PR #18, opened 2023-07-20
LLT 3986, 3987, 3988 Improvements after security review

jjanowsk:
- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
             runner: windows-2022
     runs-on: ${{ matrix.runner }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
           path: libtelio
       - name: Install packages
@@ -75,7 +75,7 @@ jobs:
         run: ./libtelio/ci/setup_ndk.sh
         if: ${{ matrix.target_os == 'android' }}
       - name: Setup MSVC
-        uses: egor-tensin/vs-shell@v2
+        uses: egor-tensin/vs-shell@9a932a62d05192eae18ca370155cf877eecc2202 # v2.1
         if: ${{ matrix.target_os == 'windows' }}
       - name: Build libtelio
         run: python3 libtelio/ci/build_libtelio.py build ${{ matrix.target_os }} ${{ matrix.arch }}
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -65,7 +65,7 @@ jobs:
             runner: windows-2022
     runs-on: ${{ matrix.runner }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
```
tomaszklak, line 68: Can you add a comment with the version that the hash points to. Here it would be something like `# v3`. This would also applies to all other places where you have replaced the hash.

CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @NordSecurity/libtelio-team
+.github/workflows @Jauler @jjanowsk @packgron @tomaszklak
```
tomaszklak, line 2: Why not give everyone from `libtelio-team` same access rights to whole repo?

jjanowsk: Because this one directory is important from security perspective since changes to workflow files can compromise security of the repo. But it's up to @packgron and @Jauler  to decide if we want this line or not

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,39 @@
+name: Scorecard security
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: "43 5 * * 0"
```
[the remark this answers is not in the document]
jjanowsk: Not really. This scorecard is more of a repository related security evaluation then code evaluation. So it only make sense to run on `main`. And this `on:` field works as `or`. So this will be run on every push to `main` and on every change to branch protection settings and additionally every Sunday.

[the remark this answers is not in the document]
tomaszklak: If I make a change that lowers our score shouldn't I be able to know that during pull request review? Why postpone the check after the code is already on main?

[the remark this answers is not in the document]
jjanowsk: I think it would be cumbersome to manually review results of this tool for every PR. It contain some checks that are really repository dependent, e.g. it checks history of the branch to calculate the score, so you not always have direct influence on the score. So because of this I don't think it makes sense to run it no branches. Also this seems to be recommended way to run it

CODEOWNERS
packgron: As this is indeed a critical place, for now let's keep it with a list of people who know implications. We can always expand moving forward.

packgron (approved):
+1.0

tomaszklak:
+1 :1st_place_medal:

tomaszklak (approved):
+1 :1st_place_medal:

lcruz99:
+1

Jauler (approved):
+1.0

PR #24, opened 2023-07-24
CI improvements

jjanowsk:
### Description
Few improvements to the CI:
1. Gitlab pipeline is run automatically for PRs where feature branch is residing inside NordSecurity repository. This way there is no need to add `run tests` label in order to trigger gitlab pipelines for internal PRs.
2. build, linters, tests and fuzzing were put into one `CI` workflow in order to use `check-all-green` and `skip-duplicate-actions` actions more easily
3. `skip-duplicate-actions` is used to skip actions if multiple of them are running for the same sha. We have this situation because we run our actions for both `push` and `pull_request` events. Alternatively we could also not run them for push (leave only push to master) but the advantage of keeping both of them is that you can push your feature branch and check the actions without needing to open a PR. I was sometimes using it on Gitlab and I think we should leave this behavior and only think about dropping it if it turns out that this `skip-duplicate-actions` does not work correctly.
4. `check-all-green` is used to collect status checks from multiple jobs in order to easily add it into the settings under: `Require status checks to pass before merging`. Without it we would need to specify ~30 of them there and modify every time we change our CI.
5. Top level permission added to `dod.yml`

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+name: CI
+on: [push, pull_request]
+permissions: {}
+
+jobs:
+  skipper:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: read
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@12aca0a884f6137d619d6a8a09fcc3406ced5281 # v5.3.0
+        with:
+          concurrent_skipping: 'same_content_newer'
+
+  build:
+    needs: skipper
+    if: needs.skipper.outputs.should_skip != 'true'
+    uses: ./.github/workflows/build.yml
+  linters:
+    needs: skipper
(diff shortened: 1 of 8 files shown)


gytsto:
`check-all-green` doesn't seem to check `Gitlab` job status or am I missing something?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+  skipper:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: read
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@v5
```
lukjok, line 15: Change version to the commit SHA

gytsto (approved):
+1

.github/workflows/ci.yml
jjanowsk: Done

tomaszklak (approved):
+1, lgtm

PR #27, opened 2023-07-25
Llt 3626 firewall speedup

tomaszklak:
### Description
**Already reviewed**
### Description
This MR improves the performance of telio-firewall as measured by the internal benchmarks. Improvements are separated into distinct commits for easier reviewing:
- Replace default hashing algorithm of standard hash map with a rustc_hash. The default one in known to be quite slow but gives some benefits (eg. protection against [Hash flooding](https://en.wikipedia.org/wiki/Collision_attack#Hash_flooding)). The new one is used in the rustc compiler and is much faster and works well for relatively small keys.
- The common path for the outbound tcp packets has been cleaned up to avoid any unnecessary work.
- LRU cache has been reimplemented. The one we used previously was using BTreeMap which is in general slower than HashMap and [it also performed many operations in O(n)](https://github.com/maidsafe/lru_time_cache/issues/143). Our new implementation is based on the [LinkedHashMap](https://docs.rs/hashlink/latest/hashlink/linked_hash_map/struct.LinkedHashMap.html) (hash map with custom order for iteration). To ensure correctness property based tests verify that the new implementation behaves the same way as the old one.
- In some places peer whitelist was searched twice for the same packet which is now no longer the case.
- Instead of getting the same key/value pair twice we now keep one reference where applicable.
- Since lru cache is in the utils crate all methods had been marked as always inline

Results of comparing performance in benchmarks of this MR (and IPv6) with the main commit f183bdf4f4150a5fd6c50943ca90e2614a4805bf (without IPv6) are below. The only benchmarks that are not better are first two which check how much time we spend in processing inbound tcp packets when there are 0 peers - which is a artificial case.

### Conformity
- [ ] [Changelog entry]

Closes LLT-3626

[firewal.benchmark.results.txt](https://github.com/NordSecurity/libtelio/files/12158504/firewal.benchmark.results.txt)

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,6 +28,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


lcruz99:
+2

lcruz99 (approved):
+2.0

gytsto:
changelog was not updated, gitlab pipelines was not finished :smile:

tomaszklak:
> changelog was not updated
Changelog already includes the information that telio-firewall supports ipv6 and this kind of speedup that brings it back to the old performance is not something that is I think interesting enough for the target audience, to merit a separate entry.

PR #33, opened 2023-07-26
Next CI improvements

jjanowsk:
1. Move clippy-mac from gitlab pipeline and unify arguments between linux, win and mac jobs
2. Remove skip-duplicate-actions and replace it with our own check. The idea is as follows: for all internal PRs `push` trigger is enough. It basically is equivalent to how pipelines work on gitlab. `pull request` trigger is only needed because pipelines would not otherwise be triggered for PRs from fork because obviously `push` is not triggered for them. So lets just run our pipelines only for `push` and  lets use  `pull request` trigger only for pull requests from forks.
3. Improve gitlab triggering: Again, similarly to point 2 run them only on 'push' trigger for our PRs and  'pull request target' in case of PRs from forks, but run them for forks only once when 'run tests' label is added. 

One more note:
If you look at the workflows run for this PR you will notice that gitlab is triggered twice. It's correct behavior. It is because gitlab triggering work in target branch context (because `pull_request_target` trigger is used). That means that the code from main is checked out to run this workflow, not the code from this branch. Once this is merged and becomes a main PRs opened to the main will use workflow with updates from this branch. I've tested that it works correctly by opening dummy PR with this branch set as a target branch. And it works :)

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,38 +3,32 @@ on: [push, pull_request]
 permissions: {}
 
 jobs:
-  skipper:
-    runs-on: ubuntu-latest
-    permissions:
-      actions: write
-      contents: read
-    outputs:
-      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+  trigger:
+    if: |
+      github.event_name == 'push' ||
+      (
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+      )
+    runs-on: ubuntu-22.04
     steps:
-      - id: skip_check
-        uses: fkirc/skip-duplicate-actions@12aca0a884f6137d619d6a8a09fcc3406ced5281 # v5.3.0
-        with:
-          concurrent_skipping: 'same_content_newer'
+      - name: CI Trigger
+        run: echo "Triggering CI"
 
   build:
-    needs: skipper
-    if: needs.skipper.outputs.should_skip != 'true'
+    needs: trigger
     uses: ./.github/workflows/build.yml
   linters:
-    needs: skipper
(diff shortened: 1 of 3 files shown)


gytsto (dismissed):

.github/workflows/ci.yml
```diff
@@ -3,38 +3,20 @@ on: [push, pull_request]
...
-    steps:
-      - id: skip_check
-        uses: fkirc/skip-duplicate-actions@12aca0a884f6137d619d6a8a09fcc3406ced5281 # v5.3.0
-        with:
-          concurrent_skipping: 'same_content_newer'
-
   build:
-    needs: skipper
-    if: needs.skipper.outputs.should_skip != 'true'
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
```
tomaszklak, line 7: Isn't this a bigger set than what is described in the PR? I would expect to have a condition like (event_name = push) || (different full names).

tomaszklak: I think the third option is the best/safest. I wouldn't count on github to never add new event type.

jjanowsk: It's not about github adding new type. It's about us having defined only those two at the top of the workflow file like this:
```
name: CI
on: [push, pull_request]
```

jjanowsk: But I'll update it so the intent is clear

jjanowsk: Fixed, I've also added trigger job back, so this huge `if` does not have to be repeated 5 times. Jobs that need other jobs are also cancelled if the ones they need are canceled.

tomaszklak (approved):
+1.5

PR #34, opened 2023-07-26
[LLT-3626] Add firewall benchmarks for IPv6

tomaszklak:
Packets sent/received in benchmarks are abstracted to allow for uniform handling of both IPv4 and IPv6 packets (in the same way as we did in the firewall tests). This doubles number of benchmarks.

At the same time different number of packets to send/receive is removed and only the bigger number of packets is used. This halves the number of benchmarks which in the end results in keeping the same number of benchmarks overall.

### Description
*--write your description here--*

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-firewall/benches/firewall_bench.rs
+++ b/crates/telio-firewall/benches/firewall_bench.rs
@@ -1,40 +1,97 @@
-use std::fmt::Display;
+use std::fmt::{Debug, Display, Formatter};
+use std::net::SocketAddr;
+use std::ops::Deref;
+use std::rc::Rc;
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use pnet_packet::tcp::TcpFlags;
 use telio_crypto::SecretKey;
-use telio_firewall::firewall::tests::{make_tcp, make_udp};
+use telio_firewall::firewall::tests::{make_tcp, make_tcp6, make_udp, make_udp6};
 use telio_firewall::firewall::{Firewall, StatefullFirewall};
 
 const PEER_COUNTS: [usize; 8] = [0, 1, 2, 4, 8, 16, 32, 64]; // max 60 peers: https://meshnet.nordvpn.com/getting-started/meshnet-explained#meshnet-scalability
-const PACKET_COUNTS: [u64; 2] = [1_000, 100_000];
+const PACKET_COUNT: u64 = 100_000;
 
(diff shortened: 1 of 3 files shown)


crates/telio-firewall/benches/firewall_bench.rs
```diff
@@ -1,40 +1,97 @@
...
-    packets: u64,
+    packet: Packet,
 }
 
 impl Display for Parameter {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{self:?}")
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let ip = if self.packet.src.is_ipv4() { 4 } else { 6 };
+        write!(f, "peers: {}, ip: v{ip}", self.peers)
```
jjanowsk, line 67: I assume it cannot look like: `"peers: {self.peers}, ip: v{ip}"` because of some syntax limitaion?

jjanowsk (approved):
lgtm +1

crates/telio-firewall/benches/firewall_bench.rs
tomaszklak: yes, only identifier are allowed: https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html#future-possibilities

crates/telio-firewall/src/firewall.rs
```diff
@@ -844,8 +844,10 @@ pub mod tests {
 
         let mut ip = MutableIpv4Packet::new(&mut raw).expect("TCP: Bad IP buffer");
         set_ipv4(&mut ip, IpNextHeaderProtocols::Tcp, IPV4_HEADER_MIN, ip_len);
-        let source: SocketAddrV4 = src.parse().expect("UDPv4: Bad src address");
-        let destination: SocketAddrV4 = dst.parse().expect("UDPv4: Bad dst address");
+        let source: SocketAddrV4 = src
+            .parse()
+            .expect(&format!("TCPv4: Bad src address: {src}"));
```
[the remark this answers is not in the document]
tomaszklak: Good point, I've added it here beacouse I needed it but forgot to update other places. Thanks!

jjanowsk (approved):

lcruz99:
+1

lcruz99 (approved):
+1

PR #36, opened 2023-07-27
[LLT-4026] Integration test for AAAA records

dfetti:
### Description
Added an integration test for AAAA records and made some improvements suggested by Tomasz

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-dns/src/zone.rs
+++ b/crates/telio-dns/src/zone.rs
@@ -1,7 +1,7 @@
 use async_trait::async_trait;
 use std::{
     collections::{HashMap, HashSet},
-    net::{IpAddr, Ipv4Addr, Ipv6Addr},
+    net::IpAddr,
     str::FromStr,
 };
 use trust_dns_client::rr::{rdata::SOA, DNSClass, LowerName, Name, RData, Record, RecordType};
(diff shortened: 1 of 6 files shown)


changelog.md
```diff
@@ -4,6 +4,7 @@
 * LLT-4002: Review log-levels of spammy log messages
 * LLT-3990: Add initial heartbeat interval to feature config
 * LLT-4008: Add support for AAAA records in telio-dns
+* LLT-4026: Add integration test for AAAA records
```
tomaszklak, line 7: I'm not sure this kind of internal change should be part of the changelog. I think that target audience for changelog is not interested in changes not observable for `libtelio` caller/user.

crates/telio-dns/src/zone.rs
```diff
@@ -21,11 +21,22 @@ use crate::forward::ForwardAuthority;
...
+    /// Both types of IP
+    Both(Ipv4Addr, Ipv6Addr),
+}
+
 /// Records (aka zone files) are instructions that live in authoritative
 /// DNS servers and provide information about a domain including what IP
 /// address is associated with that domain and how to handle requests
 /// for that domain.
-pub type Records = HashMap<String, (Option<Ipv4Addr>, Option<Ipv6Addr>)>;
+pub type Records = HashMap<String, TelioRecord>;
```
tomaszklak, line 39: Sorry for not spotting this in the previous review but is there a reason why we artificially restrict this type to have max two IPs for given domain? Normally you can get multiple IPv4 and IPv6 records so maybe instead of the `TelioRecord` this should be just `pub type Records = HashMap<String, Vec<std::net::IpAddr>>`? (Sadly there is no standard/common type that represents non empty vector).
This is just me thinking out loud, fell free to completely ignore if you think current approach is better.

tomaszklak (dismissed):
+2, some small comments but nothing blocking.

tomaszklak (approved):
+2, lgtm! :rocket:

PR #40, opened 2023-07-28
Resolve LLT-4012: fix firewall, so tcp won't be stuck in last_ack state

gytsto:
#### Root cause:
when analyzing the following sequence from perspective of Node1:
image
The rx_alive as well as tx_alive will both be set to false whenever the FIN&ACK packet is sent out, consequently, the tcp_cache entry will be removed here: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-firewall/src/firewall.rs#L454C1-L454C44. Thus upon receiving the ACK no tcp_cache entry will exist and the packet will be dropped.

#### Reproduction steps  
1. Using boringtun start meshnet
2. Add one peer to meshnet which is not allowed to do the incoming connections
3. Create TCP connection using nc between those nodes
4. Shutdown the connection from the peer (added in step 2) side
5. Observe connection left over in LAST_ACK stage

#### Solution
After FIN is being sent/recieved, shutdown tx / rx and then wait for last ACK before removing `tcp_cache` entry

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -573,6 +573,7 @@ impl StatefullFirewall {
                 }
                 return false;
             }
+            // restarts cache entry timeout
             tcp_cache.get(&key);
             telio_log_trace!("Accepting TCP packet {:?} {:?}", ip, peer);
             return true;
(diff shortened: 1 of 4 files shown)


crates/telio-firewall/src/firewall.rs
```diff
@@ -468,7 +468,6 @@ impl StatefullFirewall {
                 *tx_alive = false;
                 if !*rx_alive {
                     telio_log_trace!("Removing TCP conntrack entry {:?}", e.key());
-                    e.remove();
```
baltrus1, line 471: Correct me if I'm wrong, but If the connection is not removed, and the connection is marked as not remote_initiated(meaning our side created the connection), the remote will send new SYN and new tcp connection will be made, but our firewall will think we created that connection. This could maybe be fixed by checking rx_alive to filter received packets, but not updating cache timeout, so the cache entry will just expire eventually.

nat-lab/tests/test_mesh_firewall.py
```diff
@@ -525,3 +525,283 @@ async def on_stdout(stdout: str) -> None:
...
+
+        output_notifier.notify_output(
+            f"connect to [{CLIENT_ALPHA_IP}] from (UNKNOWN) [{CLIENT_BETA_IP}]",
+            connected_event,
+        )
+
+        output_notifier.notify_output("LAST_ACK", last_ack_event)
+        output_notifier.notify_output("TIME_WAIT", time_wait_event)
+
+        # registering on_stdout callback on both streams, cuz most of the stdout goes to stderr somehow
```
tomaszklak, line 627: I would move it as close as possible to the line that the comment refers to, to avoid confusion. In this case I think this comment should be before line 642.

crates/telio-firewall/src/firewall.rs
tomaszklak: Maybe when `!tx_alive && !rx_alive && !conn_remote_initiated` we should not update ttl and remove the entry when we receive `ack`?

tomaszklak (commented):
Please clean up the commit history so that what was changed and why. Also the PR description nicely explains the problem but is missing any explanation of the chosen solution.

nat-lab/tests/test_mesh_firewall.py
gytsto: fixed

crates/telio-firewall/src/firewall.rs
```diff
@@ -567,7 +561,13 @@ impl StatefullFirewall {
                 telio_log_trace!("Connection {:?} closing", key);
                 if let Some(connection) = tcp_cache.get_mut(&key) {
                     connection.rx_alive = false;
-                    if !connection.tx_alive {
+                }
+            } else if flags & TcpFlags::ACK == TcpFlags::ACK {
+                if let Some(connection) = tcp_cache.get_mut(&key) {
```
tomaszklak, line 566: we already have a `connection_info` so there should be no need for another `get`.

crates/telio-firewall/src/firewall.rs
```diff
@@ -1260,6 +1260,11 @@ pub mod tests {
             assert_eq!(fw.process_outbound_packet(&make_peer(), &make_tcp(us, them, TcpFlags::FIN)), true);
             assert_eq!(fw.tcp.lock().unwrap().len(), 1);
             assert_eq!(fw.process_inbound_packet(&make_peer(), &make_tcp(them, us, TcpFlags::FIN)), true);
+            assert_eq!(fw.tcp.lock().unwrap().len(), 1);
```
tomaszklak, line 1263: The other end can never send a `ACK` but some other packets and I think that this will prevent us from ever expiring the cache entry (since any accepted packet right now bumps the timeout by ttl).

I'm also to be honest not sure what to do with such packets, should be change current bevaiour and drop them?

If we want to handle that, this might require us to switch from `get` to `peek` and bump the timeout only if we are not waiting for closing `ACK`.

nat-lab/tests/test_mesh_firewall.py
```diff
@@ -525,3 +525,282 @@ async def on_stdout(stdout: str) -> None:
...
+                generate_connection_tracker_config(
+                    ConnectionTag.DOCKER_CONE_CLIENT_2,
+                    derp_1_limits=ConnectionLimits(1, 1),
+                ),
+            )
+        )
+
+        client_alpha = await exit_stack.enter_async_context(
+            telio.Client(
+                connection_alpha, alpha, telio.AdapterType.BoringTun
```
tomaszklak, line 563: I didn't realize this last time but I think we should expect to see same behaviour on both boring tun and linux native adapters. If that is the case can you parametrize both tests?

packgron: Do we have proper firewall config in nat-lab tests? Because firewall hooks only properly exist for boringtun.

crates/telio-firewall/src/firewall.rs
gytsto: changed

crates/telio-firewall/src/firewall.rs
gytsto: >>> I'm also to be honest not sure what to do with such packets, should be change current bevaiour and drop them?

idk actually, redid so it drops them now :shrug:

crates/telio-firewall/src/firewall.rs
```diff
@@ -567,13 +561,19 @@ impl StatefullFirewall {
...
+            {
+                if flags & TcpFlags::ACK == TcpFlags::ACK {
+                    telio_log_trace!("Removing TCP conntrack entry {:?}", key);
+                    tcp_cache.remove(&key);
+                    return true;
+                }
+                return false;
             }
-
+            tcp_cache.get(&key);
```
tomaszklak, line 576: Please add a comment explaining that the `get` is called here only to restart the cache entry timeout.

More importantly I think that this timeout update should only happen when `!(!connection_info.tx_alive && !connection_info.rx_alive && !connection_info.conn_remote_initiated)`

crates/telio-firewall/src/firewall.rs
```diff
@@ -546,7 +540,7 @@ impl StatefullFirewall {
         let flags = tcp_packet.get_flags();
         let mut tcp_cache = unwrap_lock_or_return!(self.tcp.lock(), false);
 
-        if let Some(connection_info) = tcp_cache.get(&key) {
+        if let Some(connection_info) = tcp_cache.peek(&key) {
```
tomaszklak, line 543: Please add a test that would fail before this change. Maybe something like sending non `ACK` packets when `!connection_info.tx_alive
                && !connection_info.rx_alive
                && !connection_info.conn_remote_initiated` and verify that such packets don't extend the entry lifetime.

crates/telio-firewall/src/firewall.rs
gytsto: >>> More importantly I think that this timeout update should only happen when !(!connection_info.tx_alive && !connection_info.rx_alive && !connection_info.conn_remote_initiated)

I think it already does that :smile: , but yeah, comment is missing, will add it

tomaszklak: You are right... I thought for some reason we are outside of the outer `if let Some...` :facepalm:

gytsto: updated

crates/telio-firewall/src/firewall.rs
gytsto: added test `firewall_tcp_connection_timeout_is_not_updated_after_fin`

tomaszklak (approved):
+1, lgtm!

PR #41, opened 2023-07-31
nat-lab: Refactor event handling

lcruz99:
### Description
Some nat-lab tests were not handling events correctly, causing flakyness. After initial investigation on test_direct_connection_endpoint_gone() flakyness it was found that most calls to wait_for_event functions were checking historic events where they should instead be waiting for new events/states (node & derp servers). Also it was using wait_for_any_derp_state which looks for a state in any derp, where they should be checking on every derp (eg: blocking every derp servers connection).
In this PR, the checks were splitted into two distint category of functions, `wait_for_state_<derp/peer>` and `wait_for_event_<derp/peer>`. `wait_for_state_<derp/peer>` functions wait for a state report that matches with the given arguments, and `wait_for_event_<derp/peer>` functions wait for a new event report on stdout that matches the given states list. The difference is that the state functions checks (can also wait) for that state, while the event functions exclusively wait for new events from the instant they are called ~~OR they can take a asyncio.Event() as argument which is later set() in a way that the caller is able to launch the function before executing a specific action, ensuring that events are caught and thus avoiding data races~~ (Removed).

Note1: critical changes were introduced in telio.py and the rest are adjustments.
Note2: to avoid bigger MR, I decided to tackle [LLT-3880](https://nordsec.atlassian.net/browse/LLT-3880)
and [LLT-3846](https://nordsec.atlassian.net/browse/LLT-3846) on a separated MR.
Note3: probably a lock could solve the need for wait_for_event functions, it would synchronize quick event changes (Disconnected -> Connecting -> Connected). 

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -166,9 +166,7 @@ class AdapterType(Enum):
 
 class Runtime:
     _output_notifier: OutputNotifier
-    _peer_info: Dict[str, PeerInfo]
     _peer_state_events: List[PeerInfo]
-    _derp_state: Dict[str, DerpServer]
     _derp_state_events: List[DerpServer]
     _started_tasks: List[str]
     _stopped_tasks: List[str]
@@ -176,9 +174,7 @@ class Runtime:
 
     def __init__(self) -> None:
         self._output_notifier = OutputNotifier()
-        self._peer_info = {}
         self._peer_state_events = []
-        self._derp_state = {}
         self._derp_state_events = []
         self._started_tasks = []
         self._stopped_tasks = []
@@ -213,66 +209,88 @@ class Runtime:
         public_key: str,
         states: List[State],
         paths: List[PathType],
-        new_event_explicitly: bool = False,
     ) -> None:
-        def _get_events() -> Set[PeerInfo]:
-            return {
-                event
-                for event in self._peer_state_events
-                if event.public_key == public_key
-                and event.state in states
-                and event.path in paths
-            }
-
-        old_events = _get_events()
         while True:
             peer = self.get_peer_info(public_key)
-            events = _get_events()
-            if (
-                not new_event_explicitly
-                and (
-                    (peer and peer.path in paths and peer.state in states)
-                    or any(events)
-                )
-            ) or (new_event_explicitly and set(events) - set(old_events)):
-                self._peer_state_events = list(set(self._peer_state_events) - events)
+            if peer and peer.path in paths and peer.state in states:
+                return
+            await asyncio.sleep(0.01)
+
+    async def notify_peer_event(
+        self,
+        public_key: str,
+        states: List[State],
+        paths: List[PathType],
+    ) -> None:
+        def _get_events() -> List[PeerInfo]:
+            return [
+                peer
+                for peer in self._peer_state_events
+                if peer
+                and peer.public_key == public_key
+                and peer.path in paths
+                and peer.state in states
+            ]
+
+        old_events = _get_events()
+
+        while True:
+            new_events = _get_events()[len(old_events) :]
+            if new_events:
                 return
             await asyncio.sleep(0.01)
 
     def get_peer_info(self, public_key: str) -> Optional[PeerInfo]:
-        return self._peer_info.get(public_key)
+        events = [
+            peer_event
+            for peer_event in self._peer_state_events
+            if peer_event.public_key == public_key
+        ]
+        if events:
+            return events[-1]
+        else:
+            return None
 
     async def notify_derp_state(
-        self, server_ip: str, states: List[State], new_event_explicitly: bool = False
+        self,
+        server_ip: str,
+        states: List[State],
     ) -> None:
-        def _get_events() -> Set[DerpServer]:
-            return {
-                event
-                for event in self._derp_state_events
-                if event.ipv4 == server_ip and (event.conn_state in states)
-            }
-
-        old_events = _get_events()
         while True:
             derp = self.get_derp_info(server_ip)
-            events = _get_events()
-            if (
-                not new_event_explicitly
-                and (
-                    (derp and (derp.ipv4 == server_ip and derp.conn_state in states))
-                    or any(events)
-                )
-            ) or (new_event_explicitly and set(events) - set(old_events)):
-                self._derp_state_events = list(set(self._derp_state_events) - events)
+            if derp and derp.ipv4 == server_ip and derp.conn_state in states:
+                return
+            await asyncio.sleep(0.01)
+
+    async def notify_derp_event(
+        self,
+        server_ip: str,
+        states: List[State],
+    ) -> None:
+        def _get_events() -> List[DerpServer]:
+            return [
+                event
+                for event in self._derp_state_events
+                if event.ipv4 == server_ip and event.conn_state in states
+            ]
+
+        old_events = _get_events()
+
+        while True:
+            new_events = _get_events()[len(old_events) :]
+            if new_events:
                 return
             await asyncio.sleep(0.01)
 
     def get_derp_info(self, server_ip: str) -> Optional[DerpServer]:
-        return self._derp_state.get(server_ip)
+        events = [event for event in self._derp_state_events if event.ipv4 == server_ip]
+        if events:
+            return events[-1]
+        else:
+            return None
 
     def _set_peer(self, peer: PeerInfo) -> None:
         assert peer.public_key in self.allowed_pub_keys
-        self._peer_info[peer.public_key] = peer
         self._peer_state_events.append(peer)
 
     def _handle_node_event(self, line) -> bool:
@@ -291,7 +309,6 @@ class Runtime:
         return False
 
     def _set_derp(self, derp: DerpServer) -> None:
-        self._derp_state[derp.ipv4] = derp
         self._derp_state_events.append(derp)
 
     def _handle_derp_event(self, line) -> bool:
@@ -332,29 +349,35 @@ class Events:
         )
         await event.wait()
 
-    async def wait_for_peer_state(
+    async def wait_for_state_peer(
         self,
         public_key: str,
         state: List[State],
         path: List[PathType],
     ) -> None:
-        await self._runtime.notify_peer_state(public_key, state, path, False)
+        await self._runtime.notify_peer_state(public_key, state, path)
 
-    async def wait_for_new_peer_state(
+    async def wait_for_event_peer(
         self,
         public_key: str,
-        state: List[State],
-        path: List[PathType],
+        states: List[State],
+        paths: List[PathType],
     ) -> None:
-        await self._runtime.notify_peer_state(public_key, state, path, True)
+        await self._runtime.notify_peer_event(public_key, states, paths)
 
-    async def wait_for_derp_state(self, server_ip: str, states: List[State]) -> None:
-        await self._runtime.notify_derp_state(server_ip, states, False)
-
-    async def wait_for_new_derp_state(
-        self, server_ip: str, states: List[State]
+    async def wait_for_state_derp(
+        self,
+        server_ip: str,
+        states: List[State],
     ) -> None:
-        await self._runtime.notify_derp_state(server_ip, states, True)
+        await self._runtime.notify_derp_state(server_ip, states)
+
+    async def wait_for_event_derp(
+        self,
+        server_ip: str,
+        states: List[State],
+    ) -> None:
+        await self._runtime.notify_derp_event(server_ip, states)
 
 
 class Client:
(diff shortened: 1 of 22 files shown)


nat-lab/tests/test_network_switch.py
```diff
@@ -332,33 +328,33 @@ async def test_mesh_network_switch_direct(
             await testing.wait_long(ping.wait_for_next_ping())
 
         await network_switcher.switch_to_secondary_network()
-
         await alpha_client.notify_network_change()
 
         await testing.wait_lengthy(
-            alpha_client.wait_for_any_derp_state(
```
gytsto, line 339: I see this doesn't work? waiting for all connection states I mean...

What if you were to write something like this:
```
async with run_async_contexts(
    [
        alpha_client.wait_for_event_on_any_derp([telio.State.Disconnected]),
        alpha_client.wait_for_event_on_any_derp([telio.State.Connecting]),
        alpha_client.wait_for_event_on_any_derp([telio.State.Connected]),
    ]
) as futures:
    # wait for futures to schedule
    await asyncio.sleep(0)
    await alpha_client.notify_network_change()
    for fut in futures:
        await fut
```

nat-lab/tests/test_node_state_flickering.py
```diff
@@ -84,34 +84,28 @@ async def test_node_state_flickering(
...
         )
 
         with pytest.raises(asyncio.TimeoutError):
             await testing.wait_defined(
                 asyncio.gather(
-                    client_alpha.wait_for_any_new_node_event(beta.public_key),
-                    client_alpha.wait_for_any_new_derp_state(),
-                    client_beta.wait_for_any_new_node_event(alpha.public_key),
-                    client_beta.wait_for_any_new_derp_state(),
+                    client_alpha.wait_for_any_state_peer(beta.public_key),
```
gytsto, line 103: events here, not states

gytsto (changes requested):
General issues:
1. Split derp / peer / conn _tracker events in separte gathers for better indication (also, `asyncio.gather` doesn't ensure you, that coroutines will be called in order you provided)
2. It feels like we have too much `wait_for_derp_(state/event)` functions, kinda getting confusing
3. providing `asyncio.Event()` in those wait functions not needed.
4. also need to thoroughly check tests if logic was not changed accidentally

nat-lab/tests/telio.py
```diff
@@ -213,57 +213,76 @@ async def notify_peer_state(
...
-                )
-            ) or (new_event_explicitly and set(events) - set(old_events)):
-                self._peer_state_events = list(set(self._peer_state_events) - events)
+            new_events = _get_events()[len(old_events) :]
+            if new_events:
                 return
             await asyncio.sleep(0.01)
 
     def get_peer_info(self, public_key: str) -> Optional[PeerInfo]:
         return self._peer_info.get(public_key)
```
gytsto, line 252: I think this can be changed to, therefore we wouldn't need `self._peer_info`
```
def get_peer_info(self, public_key: str) -> Optional[PeerInfo]:
    return [
        peer_event
        for peer_event in self._peer_state_events
        if peer_event.public_key == public_key
    ][-1]
```

nat-lab/tests/telio.py
```diff
@@ -213,57 +213,76 @@ async def notify_peer_state(
...
+                if peer
+                and peer.public_key == public_key
+                and peer.path == path
+                and peer.state in states
+            ]
 
         old_events = _get_events()
+
+        if event:
+            event.set()
```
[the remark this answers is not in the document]
lcruz99: I think you missed the point. Event is there to synchronize tasks, after we initialize old_events the event is set() to inform the main thread that the event waiting task is ready (we got the historic events, no dataraces will happen), giving green light to execute the test actions. Therefore, moving it to While loop is trivial and adds unnecessary cpu iterations. 

You can do `future = run_async(notifiy_peer_event(...))`,  however how do you ensure that notifiy_peer_event() have started and old_events been fetched before starting test actions? You could await on that future but that would block. So I came up with Event() because it truly launchs the task asynchrounisly and confirms we are ready to move on.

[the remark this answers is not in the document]
lcruz99: Just to add that, we do not want to wait for the function return, but for the function signal that the old events have been fetched, allowing us to start the test actions without eventual dataraces.

nat-lab/tests/telio.py
```diff
@@ -213,57 +213,76 @@ async def notify_peer_state(
...
                 for event in self._derp_state_events
-                if event.ipv4 == server_ip and (event.conn_state in states)
-            }
+                if event.ipv4 == server_ip and event.conn_state in states
+            ]
 
         old_events = _get_events()
+
+        if event:
+            event.set()
```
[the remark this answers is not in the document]
lcruz99: Replied above

nat-lab/tests/test_node_state_flickering.py
lcruz99: good catch

nat-lab/tests/telio.py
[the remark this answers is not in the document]
gytsto: Can you show some use cases of it, cuz I cannot wrap my head around, why is it actually needed and why it is not used anywhere. :confused:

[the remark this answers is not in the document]
lcruz99: Oops, it seems I did some last minute changes that I didn't remember, sorry about that, you are right.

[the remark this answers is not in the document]
lcruz99: So currently we are not using them, I'll have them removed. 

The thought behind it was to avoid dataraces. For instance, if we have an event that we would like to catch but it was logged so quickly after the test action that we couldn't call wait_for_event() in time, missing it.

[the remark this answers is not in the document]
lcruz99: It technically still can happen, see: https://github.com/NordSecurity/libtelio/pull/41/files#diff-778fbedd2d660ca2c2322076ce67ff83072b808e239308da0bbfc3f130add155R331

nat-lab/tests/telio.py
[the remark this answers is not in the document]
lcruz99: will remove

nat-lab/tests/telio.py
```diff
@@ -332,29 +351,56 @@ async def message_done(self, message_idx: int) -> None:
...
+        states: List[State],
+        paths: PathType,
+        event: Optional[asyncio.Event] = None,
     ) -> None:
-        await self._runtime.notify_peer_state(public_key, state, path, True)
+        await self._runtime.notify_peer_event(public_key, states, paths, event)
 
-    async def wait_for_derp_state(self, server_ip: str, states: List[State]) -> None:
-        await self._runtime.notify_derp_state(server_ip, states, False)
+    async def wait_for_any_state_peer(
```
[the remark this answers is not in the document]
lcruz99: removed

nat-lab/tests/test_direct_connection.py
```diff
@@ -80,13 +80,19 @@
 
 
 @pytest.mark.asyncio
-@pytest.mark.timeout(150)
+@pytest.mark.timeout(120)
 @pytest.mark.parametrize(
-    "endpoint_providers, client1_type, client2_type, reflexive_ip",
-    UHP_conn_client_types,
+    "endpoint_providers, client1_type, client2_type",
```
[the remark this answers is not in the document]
lcruz99: test leftovers, reverted

nat-lab/tests/test_network_switch.py
lcruz99: that would work I think.. is it worth to check those intermediate events when we only care about [telio.State.Connected]? 


I have the same question for other tests: if we want to wait for a node event with Direct path is it worth to check [telio.State.Connected] Derp event and [telio.State.Connected] node event? IMO we should have those intermediate steps on a just one test case to test specifically that, on all other tests it looks redundant.

nat-lab/tests/telio.py
```diff
@@ -447,67 +493,120 @@ async def simple_start(self):
...
+        states: List[State],
+        path=PathType.Relay,
+    ) -> None:
+        await self.get_events().wait_for_state_peer(public_key, states, [path])
 
-    async def disconnect(self, public_key, path=PathType.Relay) -> None:
-        await self.get_events().wait_for_peer_state(
-            public_key, [State.Disconnected], [path]
-        )
+    async def wait_for_any_state_peer(
```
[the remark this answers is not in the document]
lcruz99: removed

nat-lab/tests/telio.py
```diff
@@ -447,67 +493,120 @@ async def simple_start(self):
...
+                    states,
+                )
+                for derp in DERP_SERVERS
+            ]
+        ) as futures:
+            while not any([fut.done() for fut in futures]):
+                await asyncio.sleep(0.01)
 
-    async def wait_for_new_derp_state(
+    async def wait_for_any_state_on_any_derp(
```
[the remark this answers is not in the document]
lcruz99: removed

nat-lab/tests/telio.py
lcruz99: Changed

lcruz99:
> General issues:
> 
>     1. Split derp / peer / conn _tracker events in separte gathers for better indication (also, `asyncio.gather` doesn't ensure you, that coroutines will be called in order you provided)
> 
>     2. It feels like we have too much `wait_for_derp_(state/event)` functions, kinda getting confusing
> 
>     3. providing `asyncio.Event()` in those wait functions not needed.
> 
>     4. also need to thoroughly check tests if logic was not changed accidentally

1. Splitted some coroutings to 2 futures (asyncio.gather)  for better readability. You're right however coroutines inside gather() were not order dependent
2. Removed specific cases of `wait_for_derp_(state/event)` functions
3. Removed `asyncio.Event()` `wait_for_derp_(state/event)` functions
4. Hope everything's fine :)

nat-lab/tests/test_network_switch.py
gytsto: I'd say it is important to track all the events, even if they are not really relevant to test. This way we are making sure, that correct flow is executed. But maybe let's leave this for the future.

And about direct path. It is important to check for derp first, since w/o derp, there cannot be any direct connection.

nat-lab/tests/telio.py
```diff
@@ -166,7 +166,6 @@ class AdapterType(Enum):
 
 class Runtime:
     _output_notifier: OutputNotifier
-    _peer_info: Dict[str, PeerInfo]
```
gytsto, line 169: simillar logic should apply to `_derp_state` , should be removed and used last event from `_derp_state_events`

nat-lab/tests/telio_test.py
```diff
@@ -67,247 +65,98 @@ async def test_set_derp_state(self) -> None:
...
     @pytest.mark.asyncio
     async def test_handle_node_event(self) -> None:
         runtime = Runtime()
         runtime.allowed_pub_keys = set(["AAA"])
+
         assert runtime.handle_output_line(
             'event node: "{"identifier":"tcli","public_key":"AAA","state":"connected","is_exit":true,"is_vpn":true,"ip_addresses":[],"allowed_ips":[],"endpoint":null,"hostname":null,"allow_incoming_connections":false,"allow_peer_send_files":false,"path":"relay"}"'
         )
-        peer_info = runtime.get_peer_info("AAA")
-        assert peer_info is not None and peer_info.state == State.Connected
```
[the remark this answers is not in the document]
gytsto: nvm actually

nat-lab/tests/telio_test.py
```diff
@@ -328,43 +177,14 @@ async def test_explicit_new_derp_state(self) -> None:
...
         )
 
-        # This should timeout even though derp connection state is connected, since no new event is generated
-        with pytest.raises(asyncio.TimeoutError):
-            await testing.wait_short(
-                events.wait_for_new_derp_state("1.1.1.1", [State.Connected])
-            )
-
-        # Start waiting for new event before it is being generated
-        async with run_async_context(
```
gytsto, line 346: why remove this?

gytsto:
- I see you missed alot of comments :smile: 
- `telio_test.py` is missing tests for `notify_peer_event` and `notify_derp_event`
- also commit history should be cleaned up

I hope I didn't forget anything else :laughing: but let's wait for you to battle current comments first

nat-lab/tests/test_network_switch.py
lcruz99: > And about direct path. It is important to check for derp first, since w/o derp, there cannot be any direct connection.
That's the point,

lcruz99: My though is to simplify tests as much as possible, to have transparent tests so that we easily understand them: adding common setup/destroy methods, cleaning test redundancy (example above), limit tests to one behaviour per test case, etc.

nat-lab/tests/telio.py
lcruz99: missed that, fixed!

lcruz99: however, shouldn't we make `_derp_state_events/_peer_state_events` atomic? At the moment we have different threads writing/reading them without synchronization

nat-lab/tests/test_network_switch.py
gytsto: I think we will do this in the future, with nat-lab framework improvements. For now this PR is not about this I guess

lcruz99: correct 😄

nat-lab/tests/telio_test.py
lcruz99: Removed this test initially because it didn't make sense for states, nevertheless added new tests

lcruz99:
> * I see you missed alot of comments 😄

still learning Github, it seems the comments I missed only appeared in 'Files changed tab', not that intuitive..

>     * `telio_test.py` is missing tests for `notify_peer_event` and `notify_derp_event`

Added tests

>     * also commit history should be cleaned up

Will be cleaned before merging 😄

nat-lab/tests/test_derp_connect.py
```diff
@@ -241,9 +273,15 @@ async def test_derp_reconnect_3clients() -> None:
         #    |         |           |
         # [ALPHA]    [BETA]     [GAMMA]
 
-        await testing.wait_lengthy(check_derp_connection(alpha_client, DERP1_IP, True))
-        await testing.wait_lengthy(check_derp_connection(beta_client, DERP2_IP, True))
-        await testing.wait_lengthy(check_derp_connection(gamma_client, DERP3_IP, True))
+        await testing.wait_lengthy(
```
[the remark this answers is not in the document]
lcruz99: Added

nat-lab/tests/test_derp_connect.py
```diff
@@ -209,9 +230,15 @@ async def test_derp_reconnect_3clients() -> None:
         #
         # (* - connection to the escaped client, but DERP does not know):
 
-        await testing.wait_lengthy(check_derp_connection(alpha_client, DERP1_IP, True))
-        await testing.wait_lengthy(check_derp_connection(beta_client, DERP2_IP, True))
-        await testing.wait_lengthy(check_derp_connection(gamma_client, DERP2_IP, True))
+        await testing.wait_lengthy(
```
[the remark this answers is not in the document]
lcruz99: Added

nat-lab/tests/test_derp_connect.py
```diff
@@ -45,41 +47,49 @@ async def test_derp_reconnect_2clients() -> None:
...
         #  [GW1]     [GW2]
         #   /           \
         # [ALPHA]     [BETA]
 
         # an iptables rule is placed in order to reject connections and
         # send a TCP reset to the client BETA
         await exit_stack.enter_async_context(
-            beta_client.get_router().break_tcp_conn_to_host(DERP1_IP)
+            beta_client.get_router().break_tcp_conn_to_host(
+                str(DERP_SERVERS[0]["ipv4"])
```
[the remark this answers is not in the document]
lcruz99: dunno.. changed to `DERP1_IP`

nat-lab/tests/test_derp_connect.py
```diff
@@ -45,41 +47,49 @@ async def test_derp_reconnect_2clients() -> None:
 
         # As the wireguard protocol routing scheme is based on the public key
         # a handshake is needed to let both clients aware of the key of each other
-        await testing.wait_long(alpha_client.handshake(beta.public_key))
-        await testing.wait_long(beta_client.handshake(alpha.public_key))
+        await testing.wait_lengthy(
+            asyncio.gather(
+                alpha_client.wait_for_state_derp(DERP1_IP, [State.Connected]),
+                beta_client.wait_for_state_derp(DERP1_IP, [State.Connected]),
+                alpha_client.wait_for_state_peer(beta.public_key, [State.Connected]),
```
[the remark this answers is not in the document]
lcruz99: done

nat-lab/tests/test_connection_states.py
```diff
@@ -85,24 +85,18 @@ async def test_connected_state_after_routing(
...
 
-        await testing.wait_long(
-            asyncio.gather(
-                client_alpha.wait_for_any_derp_state([telio.State.Connected]),
-                client_beta.wait_for_any_derp_state([telio.State.Connected]),
-            )
-        )
-
-        await testing.wait_long(
+        await testing.wait_lengthy(
```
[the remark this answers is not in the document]
lcruz99: you're right, splitted

nat-lab/tests/test_connection_states.py
```diff
@@ -17,32 +17,32 @@
...
-            ConnectionTag.WINDOWS_VM,
-            telio.AdapterType.WireguardGo,
-            marks=pytest.mark.windows,
-        ),
-        pytest.param(
-            ConnectionTag.MAC_VM,
-            telio.AdapterType.Default,
-            marks=pytest.mark.mac,
-        ),
+        # pytest.param(
```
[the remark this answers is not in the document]
lcruz99: yess, uncommented

nat-lab/tests/telio.py
```diff
@@ -213,66 +209,88 @@ async def notify_peer_state(
...
-            events = _get_events()
-            if (
-                not new_event_explicitly
-                and (
-                    (peer and peer.path in paths and peer.state in states)
-                    or any(events)
-                )
-            ) or (new_event_explicitly and set(events) - set(old_events)):
-                self._peer_state_events = list(set(self._peer_state_events) - events)
+            new_events = _get_events()[len(old_events) :]
```
arvydasv, line 238: What does the square brackets do?

lcruz99: They assign to new_events a slice of the list returned by _get_events(). The slice is from len(old_events) index to the end of the list, which translates to every event except the old ones ones (on old_events).

arvydasv:
+1

nat-lab/tests/telio.py
gytsto: Idk if we need it for now, but you could create Jira issue and add relation to LLT-4054

gytsto (approved):
:+1: don't forget to update commit messages

nat-lab/tests/telio.py
lcruz99: After some thinking, I've concluded that's not worth to make them atomic. We have only one thread writing them and even if race conditions happen it's not critical as the readers (wait_for_events) will just read them again with no side effects (notify state/event functions)

PR #42, opened 2023-07-31
Resolve LLT-4051: Reenable flaky tests with XFAIL pytest

gytsto:
### Problem
whenever flaky test appeared, we disabled it by skipping it. Sometimes losing some important information.
### Solution
mark them as `possible to fail` using XFAIL. XFAIL mark allows to indicate that tests might fail (being falky). Test will run, but no traceback will be reported when it fails. And also, end result (exit code) will not be affected

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated


lcruz99:
+2

lcruz99 (approved):
+2

PR #45, opened 2023-07-31
Remove nagger job

tomaszklak:
This job is right now just a dummy. It builds dylint but runs zero checks with it. Since it didn't use specific version of dylint, it started to fail when the dylint executable started to require rust version newer than what we use. Instead of fixing it now, let's remove it now. It can be re-added once we have our lints open sourced and ready to be used here.

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated


dfetti:
+2

jjanowsk (approved):
+2

PR #30, opened 2023-07-25
Add telio version compatability test

arvydasv:
Add a test for telio compatibility with it older version. 

This required changing the natlab docker image.

### :ballot_box_with_check: Definition of Done checklist
- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

PR #46, opened 2023-08-01
Add git commit message requirements

packgron:
### Description

Adding document that indictates how project's commit messages should look like.

- [x] Commit history is clean
- [x] Commit message adhere to [requirements](docs/git_commit_messages_requirements.md)
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/docs/git_commit_messages_requirements.md
+++ b/docs/git_commit_messages_requirements.md
@@ -1,0 +1,115 @@
+# Git Commit Message Requirements
+
+## 1. Use the body to explain why this commit exists and on a high level what was done
+
+Good
+```
+Fix buffer overflow
+
+The buffer overflow was caused by off-by-one error when copying strings.
+The source string buffer includes the null terminator, while the
+destination string buffer does not.
+```
+
+Bad
+```
+Fix buffer overflow
+
+Add -1 to the length of bytes being copied from the source string buffer
+to the destination string buffer.
+```
+
+## 2. When there is a body, its separated from subject by a blank line
+
+Note: subject line is mandatory, body is optional
+
+Good
+```
+Fix buffer overflow
+
+The buffer overflow was caused by off-by-one error when copying strings.
+The source string buffer includes the null terminator, while the
+destination string buffer does not.
+```
+
+Bad
+```
+Fix buffer overflow
+The buffer overflow was caused by off-by-one error when copying strings.
+The source string buffer includes the null terminator, while the
+destination string buffer does not.
+```
+
+## 3. Limit the subject line to 50 characters
+
+Good
+```
+Fix buffer overflow
+```
+
+Bad
+```
+Fix buffer overflow that was caused by off-by-one error when copying strings
+```
+
+## 4. Capitalize the subject line
+
+Good
+```
+Fix buffer overflow
+```
+
+Bad
+```
+fix buffer overflow
+```
+
+## 5. Do not end the subject line with a period
+
+Good
+```
+Fix buffer overflow
+```
+
+Bad
+```
+Fix buffer overflow.
+```
+
+## 6. Use the imperative mood in the subject line
+
+Good
+```
+Fix buffer overflow
+```
+
+Bad
+```
+Fixed buffer overflow
+
+Fixing buffer overflow
+
+Buffer overflow fixed
+```
+
+## 7. Wrap the body at 72 characters
+
+Good
+```
+Fix buffer overflow
+
+The buffer overflow was caused by off-by-one error when copying strings.
+The source string buffer includes the null terminator, while the
+destination string buffer does not.
+```
+
+Bad
+```
+Fix buffer overflow
+
+The buffer overflow was caused by off-by-one error when copying strings. The source string buffer includes the null terminator, while the destination string buffer does not.
+```
+
+## Sources
+
+- [How to write git commit messages](https://cbea.ms/git-commit/)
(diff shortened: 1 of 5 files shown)


dod.yml
```diff
@@ -1,4 +1,5 @@
 dod:
   - 'Commit history is clean'
+  - 'Commit message adhere to [requirements](docs/git_commit_messages_requirements.md)'
```
packgron, line 3: TODO: update to perma-link once accepted.

lcruz99:
Agree with the guidelines and would like to also see a commit identifier with type and scope, eg:

- natlab: Add six package to deps dockerfile
- telio-wg: Handle interface cleanup
- ci: Add apple tvos build job

packgron:
> Agree with the guidelines and would like to also see a commit identifier with type and scope, eg:
> 
> * natlab: Add six package to deps dockerfile
> * telio-wg: Handle interface cleanup
> * ci: Add apple tvos build job

Something like this: https://github.com/tokio-rs/tokio/commits/master?

I just have few questions :
- Do we always just add id to subject line, to body description paragraphs or mix? (Could you write example how you imagine it)
- Where/how do we describe scopes (components + [ci, nat-lab])?

tomaszklak:
> Agree with the guidelines and would like to also see a commit identifier with type and scope [...]

If we want to do it, I think we should automate verification that this prefix is present.

docs/git_commit_messages_requirements.md
```diff
@@ -0,0 +1,115 @@
...
+Bad
+```
+Fixed buffer overflow
+
+Fixing buffer overflow
+
+Buffer overflow fixed
+```
+
+## 6. Wrap the body at 72 characters
```
tomaszklak, line 76: I'm not gonna count characters in the web ui of github :stuck_out_tongue_closed_eyes: 

We should either drop this requirement or add a checker for it on CI.

docs/git_commit_messages_requirements.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+Bad
+```
+Fix buffer overflow
+
+The buffer overflow was caused by off-by-one error when copying strings. The source string buffer includes the null terminator, while the destination string buffer does not.
+```
+
+## 7. Use the body to explain what and why vs. how
```
Jauler, line 94: It does sound to me, that this part is the most important one, and we had it in our guidelines for quite some time.

The question is: do we want to have this "what" and "why" more in the PR comments, or within every commit?

docs/git_commit_messages_requirements.md
Jauler: Would it be possible to use git hooks for that? Sounds like an "earlier" place to check for this.

lcruz99:
> > Agree with the guidelines and would like to also see a commit identifier with type and scope, eg:
> > 
> > * natlab: Add six package to deps dockerfile
> > * telio-wg: Handle interface cleanup
> > * ci: Add apple tvos build job
> 
> Something like this: https://github.com/tokio-rs/tokio/commits/master?

yes, having a list with all the valid components would help.

> I just have few questions :
> 
>     * Do we always just add id to subject line, to body description paragraphs or mix? (Could you write example how you imagine it)

Only to the subject line. For longer/impact commits a commit description may be included:

natlab: Add six package to deps dockerfile

With the adding of pynat package, six package was missing as it is a dependency of pynat causing docker builds to fail.  [What was the problem]

This commit adds six package installation to the dockerfile. [How problem is fixed]

 From now on, '-six' flag has to passed to docker build command. [If there's any consequences]

PR Review: [https://github.com/NordSecurity/libtelio/pull/30](#30)

>     * Where/how do we describe scopes (components + [ci, nat-lab])?

I think components [crates, nat-lab, ci, ...] would be enough, I'm not sure about having both, it might get confusing:
- fix(nat-lab): Fix container clean up between tests

docs/git_commit_messages_requirements.md
tomaszklak: I think so! But you can skip hooks with `--no-verify` so it would be safer to check them on CI to be sure. If we make it a script it should be easy to have both hooks and CI verification.

docs/git_commit_messages_requirements.md
lcruz99: I suppose the idea is to have a clean commit history so that we can scroll it without having to check every PR/Jira issue to understand the why's/how's/what's of the changes, making everyone's life easier. Also, I think PR description != Commit description, PR is higher level and more related with the Jira issue, but if the dev wants to understand specific changes it would look to file/commit history. So I think making the effort to write a commit description is worth it. 

I believe that with time it'll get natural to write a description for every commit.

tomaszklak:
> A nice article about commit messages: [A note on commit messages](https://bitcrowd.dev/a-note-on-commit-messages/)

For me a good real world example of great commit messages are those by Russ Cox: https://github.com/golang/go/commits?author=rsc

docs/git_commit_messages_requirements.md
```diff
@@ -0,0 +1,115 @@
+# Git Commit Message Requirements
+
+## 1. Use the body to explain what and why vs. how
```
mathiaspeters, line 3: Can we rephrase this line? I feel like it's a bit confusing. Maybe something like `Use the body to explain why and on a high level what`

mathiaspeters (approved):
LGTM! +2.0

PR #48, opened 2023-08-02
Define UniFFI UDL

mathiaspeters:
### Description
This PR is the first introduction of UniFFI into libtelio, but so far UniFFI isn't actually used. That comes in a later PR. 

With UniFFI we can make changes to our FFI code that is incompatible with SWIG and do general cleanup, so UniFFI is a breaking change, but this PR doesn't actually break anything yet. Instead it duplicates out FFI code and then cleans up the duplicate, but fear not, the duplication is not going to be long-lived.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ rand.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["full"] }
 tracing.workspace = true
+uniffi.workspace = true
 uuid.workspace = true
 
 telio-crypto.workspace = true
@@ -87,6 +88,7 @@ telio-wg = { workspace = true, features = ["mockall", "test-adapter"] }
 [build-dependencies]
 anyhow.workspace = true
 cc.workspace = true
+uniffi = { workspace = true, features = ["build"] }
 
 [target.'cfg(windows)'.build-dependencies]
 winres = "0.1"
@@ -151,6 +153,7 @@ time = { version = "0.3.9", features = ["formatting"] }
 tokio = ">=1.22"
 tracing-subscriber = { version = "0.3.17", features = ["local-time"] }
 tracing-appender = "0.2.3"
+uniffi = { git = "https://github.com/NordSecurity/uniffi-rs.git", tag = "v0.25.0-1" }
 url = "2.2.2"
 uuid = { version = "1.1.2", features = ["v4"] }
 winapi = { version = "0.3", features = ["netioapi", "ws2def"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,11 @@ once_cell.workspace = true
 jni = "0.19"
 rustls-platform-verifier = { git = "https://github.com/tomaszklak/rustls-platform-verifier.git", rev = "2db68dd17d76eadb062176b913040341b99ef4f3" }
 
+[target.'cfg(windows)'.dependencies]
+winapi = { workspace = true, features = ["ntdef", "winerror"] }
+
 [dev-dependencies]
+maplit.workspace = true
 slog-async = "2.7"
 slog-term = "2.8"
 
@@ -83,9 +87,6 @@ telio-wg = { workspace = true, features = ["mockall", "test-adapter"] }
 [build-dependencies]
 anyhow.workspace = true
 cc.workspace = true
-
-[target.'cfg(windows)'.dependencies]
-winapi = { workspace = true, features = ["ntdef", "winerror"] }
 
 [target.'cfg(windows)'.build-dependencies]
 winres = "0.1"
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -15,6 +15,9 @@ pub const DEFAULT_PERSISTENT_KEEPALIVE_PERIOD: u32 = 25;
 /// Default keepalive period for direct peers
 pub const DEFAULT_DIRECT_PERSISTENT_KEEPALIVE_PERIOD: u32 = 5;
 
+/// Type alias for UniFFI
+pub type EndpointProviders = HashSet<EndpointProvider>;
+
 /// Configurable persistent keepalive periods for different types of peers
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
 pub struct FeaturePersistentKeepalive {
@@ -183,7 +186,7 @@ pub struct FeatureDirect {
     /// Endpoint providers [default all]
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_providers")]
-    pub providers: Option<HashSet<EndpointProvider>>,
+    pub providers: Option<EndpointProviders>,
     /// Polling interval for endpoints [default 10s]
     pub endpoint_interval_secs: Option<u64>,
     /// Configuration options for skipping unresponsive peers
@@ -244,7 +247,8 @@ pub struct FeatureDerp {
 }
 
 /// Whether to validate keys
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Deserialize)]
+#[serde(transparent)]
 pub struct FeatureValidateKeys(pub bool);
 
 impl Default for FeatureValidateKeys {
@@ -258,7 +262,7 @@ impl Default for FeatureValidateKeys {
 #[serde(transparent)]
 pub struct FeatureBoringtunResetConns(pub bool);
 
-fn deserialize_providers<'de, D>(de: D) -> Result<Option<HashSet<EndpointProvider>>, D::Error>
+fn deserialize_providers<'de, D>(de: D) -> Result<Option<EndpointProviders>, D::Error>
 where
     D: Deserializer<'de>,
 {
@@ -299,7 +303,6 @@ pub struct Features {
     pub direct: Option<FeatureDirect>,
     /// Test environment (natlab) requires binding feature disabled
     /// TODO: Remove it once mac integration tests support the binding mechanism
-    #[cfg(any(target_os = "macos", feature = "pretend_to_be_macos"))]
     pub is_test_env: Option<bool>,
     /// Derp server specific configuration
     pub derp: Option<FeatureDerp>,
@@ -423,7 +426,7 @@ mod tests {
             "boringtun_reset_connections": true
         }"#;
 
-    static EXPECTED_FEATURES: Lazy<Features> = Lazy::new(|| Features {
+    static EXPECTED_FEATURES_WITH_IS_TEST_ENV: Lazy<Features> = Lazy::new(|| Features {
         wireguard: FeatureWireguard {
             persistent_keepalive: FeaturePersistentKeepalive {
                 vpn: None,
@@ -457,7 +460,6 @@ mod tests {
         exit_dns: Some(FeatureExitDns {
             auto_switch_dns_ips: Some(true),
         }),
-        #[cfg(any(target_os = "macos", feature = "pretend_to_be_macos"))]
         is_test_env: Some(true),
         derp: Some(FeatureDerp {
             tcp_keepalive: Some(1),
@@ -471,8 +473,7 @@ mod tests {
         boringtun_reset_connections: FeatureBoringtunResetConns(true),
         flush_events_on_stop_timeout_seconds: None,
     });
-
-    static EXPECTED_FEATURES_WITHOUT_TEST_ENV: Lazy<Features> = Lazy::new(|| Features {
+    static EXPECTED_FEATURES_WITHOUT_IS_TEST_ENV: Lazy<Features> = Lazy::new(|| Features {
         wireguard: FeatureWireguard {
             persistent_keepalive: FeaturePersistentKeepalive {
                 vpn: None,
@@ -504,7 +505,6 @@ mod tests {
         exit_dns: Some(FeatureExitDns {
             auto_switch_dns_ips: None,
         }),
-        #[cfg(any(target_os = "macos", feature = "pretend_to_be_macos"))]
         is_test_env: None,
         derp: None,
         validate_keys: Default::default(),
@@ -677,6 +677,7 @@ mod tests {
             paths: None,
             direct: None,
             exit_dns: None,
+            is_test_env: None,
             derp: None,
             validate_keys: Default::default(),
             ipv6: false,
@@ -703,6 +704,7 @@ mod tests {
             paths: None,
             direct: None,
             exit_dns: None,
+            is_test_env: None,
             derp: None,
             validate_keys: Default::default(),
             ipv6: false,
@@ -724,6 +726,7 @@ mod tests {
             paths: None,
             direct: None,
             exit_dns: None,
+            is_test_env: None,
             derp: None,
             validate_keys: Default::default(),
             ipv6: false,
@@ -764,6 +767,7 @@ mod tests {
             exit_dns: Some(FeatureExitDns {
                 auto_switch_dns_ips: Some(true),
             }),
+            is_test_env: None,
             derp: None,
             validate_keys: Default::default(),
             ipv6: false,
@@ -781,6 +785,7 @@ mod tests {
             exit_dns: Some(FeatureExitDns {
                 auto_switch_dns_ips: None,
             }),
+            is_test_env: None,
             derp: None,
             validate_keys: Default::default(),
             ipv6: false,
@@ -806,7 +811,6 @@ mod tests {
             lana: None,
             paths: None,
             direct: None,
-            #[cfg(any(target_os = "macos", feature = "pretend_to_be_macos"))]
             is_test_env: None,
             exit_dns: None,
             derp: None,
@@ -834,10 +838,9 @@ mod tests {
     fn test_json_without_is_test_env_to_feature_set() {
         let deserialization: Result<Features, _> =
             from_str(CORRECT_FEATURES_JSON_WITHOUT_IS_TEST_ENV);
-
         assert_eq!(
             deserialization.unwrap(),
-            *EXPECTED_FEATURES_WITHOUT_TEST_ENV
+            *EXPECTED_FEATURES_WITHOUT_IS_TEST_ENV
         );
     }
 
@@ -846,7 +849,10 @@ mod tests {
         let deserialization: Result<Features, _> =
             from_str(CORRECT_FEATURES_JSON_WITH_ALL_OPTIONAL);
 
-        assert_eq!(deserialization.unwrap(), *EXPECTED_FEATURES);
+        assert_eq!(
+            deserialization.unwrap(),
+            *EXPECTED_FEATURES_WITH_IS_TEST_ENV
+        );
     }
 
     #[test]
@@ -859,6 +865,7 @@ mod tests {
             paths: None,
             exit_dns: None,
             direct: None,
+            is_test_env: None,
             derp: None,
             validate_keys: Default::default(),
             ipv6: false,
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -499,7 +499,7 @@ impl Device {
         }
     }
 
-    pub fn try_shutdown(mut self, timeout: Duration) -> Result {
+    pub fn try_shutdown(&mut self, timeout: Duration) -> Result {
         let art = self.art.take().ok_or(Error::NotStarted)?;
         let art = Arc::try_unwrap(art);
         match art {
@@ -2595,7 +2595,7 @@ mod tests {
         let (sender, _receiver) = tokio::sync::broadcast::channel(1);
         let features = Features {
             direct: Some(FeatureDirect {
-                providers: Some(HashSet::<telio_model::api_config::EndpointProvider>::new()),
+                providers: Some(HashSet::new()),
                 endpoint_interval_secs: None,
                 skip_unresponsive_peers: Default::default(),
             }),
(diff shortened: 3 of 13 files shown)


crates/telio-model/src/api_config.rs
```diff
@@ -164,7 +164,7 @@ pub struct FeatureDirect {
     /// Endpoint providers [default all]
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_providers")]
-    pub providers: Option<HashSet<EndpointProvider>>,
+    pub providers: Option<Vec<EndpointProvider>>,
```
mathiaspeters, line 167: UniFFI can't deal with hash sets so replaced with vector

crates/telio-model/src/api_config.rs
```diff
@@ -219,12 +219,16 @@ pub struct Features {
     pub direct: Option<FeatureDirect>,
     /// Test environment (natlab) requires binding feature disabled
     /// TODO: Remove it once mac integration tests support the binding mechanism
-    #[cfg(any(target_os = "macos", feature = "pretend_to_be_macos"))]
```
mathiaspeters, line 222: UniFFI doesn't like conditionally including/excluding stuff it exposes so this field will now always be there

crates/telio-model/src/api_config.rs
```diff
@@ -219,12 +219,16 @@ pub struct Features {
...
+    #[serde(default = "default_is_test_env")]
     pub is_test_env: Option<bool>,
     /// Derp server specific configuration
     pub derp: Option<FeatureDerp>,
 }
 
+fn default_is_test_env() -> Option<bool> {
+    Some(true)
+}
```
mathiaspeters, line 230: I don't fully remember why this defaults to true, but I remember trying to just do the type default (back then false, because it wasn't yet an option) and that broke stuff in tests

src/ffi/types.rs
```diff
@@ -3,10 +3,111 @@ use log::Level;
...
+    #[error("LockError")]
+    LockError,
+    #[error("InvalidString")]
+    InvalidString,
+    #[error("AlreadyStarted")]
+    AlreadyStarted,
+}
+
+#[derive(Copy, Clone, Debug)]
+#[allow(clippy::enum_variant_names)] // Otherwise clippy complains about all variants having "Tun" as their suffix
```
mathiaspeters, line 43: Without this line the compilers/linters complain that all variants end with `Tun`

src/libtelio.udl
```diff
@@ -0,0 +1,520 @@
...
+    /// # Parameters
+    /// - `log_level`: Max log level to log.
+    /// - `logger`: Callback to handle logging events.
+    void set_global_logger(TelioLogLevel log_level, TelioLoggerCb logger);
+
+    /// Get default recommended adapter type for platform.
+    TelioAdapterType uniffi_telio_get_default_adapter();
+
+    /// Get current version tag.
+    string uniffi_telio_get_version_tag();
```
mathiaspeters, line 30: Some functions here are prefixed with `uniffi_telio_`, that is temporary and will be removed in a future PR. It's just to prevent name clashes with existing functions until those functions are removed

src/ffi/mod.rs
```diff
@@ -82,9 +113,567 @@ lazy_static::lazy_static! {
...
+        error!("{}", err)
+    }
+
+    log::set_max_level(
+        (<types::TelioLogLevel as Into<Level>>::into(log_level) as Level).to_level_filter(),
+    );
+}
+
+/// Get default recommended adapter type for platform.
+pub fn uniffi_telio_get_default_adapter() -> TelioAdapterType {
```
mathiaspeters, line 131: See my other comment about temporary name prefixes

tomaszklak:
Can you add in this PR something like `ffi/tests` to prove that this is at least working a little bit? As far as I understand this PR lacks any form of testing/verification.

Cargo.toml
```diff
@@ -38,6 +38,7 @@ tokio = { version = ">=1.22", features = ["full"] }
 async-trait = "0.1.51"
 uuid =  { version = "1.1.2", features = ["v4"] }
 wasm-bindgen = "=0.2.83" # Only for compatability with moose wasm version
+uniffi.workspace = true
```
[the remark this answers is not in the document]
jjanowsk: Also as I understand this is done just to remove duplication of `uniffi = { git = "https://github.com/NordSecurity/uniffi-rs.git", branch = "nordsec.0.23.0" }` but I would advise to use official uniffi 0.23.0 so maybe it's not a problem to just have
```
uniffi = "0.23.0"
...
uniffi = { version = "0.23.0", features = ["build"] }
```
I know it still requires updates in two places for version bumps but reduces the complexity of the toml in general. 

Just my 2 cents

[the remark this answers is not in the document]
mathiaspeters: As far as I know, upstream uniffi doesn't support doc comments so probably best to stick to our fork. But I can duplicate the version information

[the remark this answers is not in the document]
jjanowsk: Only generators must support docstrings. Scaffolding part does not need to know about it and I believe upstream just ignores comments (this is to be checked) so it should work with this udl. And we keep compatibility between upstream and our fork. So you should be able to safely use upstream 0.23.0 and use our generators to support docstrings in the generated bindings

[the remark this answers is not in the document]
mathiaspeters: Ah, there's been a misunderstanding then. I asked about this in a slack thread and was told both the rust dependency and the generator needs to be our fork for doc comments to work, but it was probably me being unclear in my question. Good to know we can use upstream here

crates/telio-model/src/api_config.rs
mathiaspeters: I can do that, though that `cfg` attribute doesn't persist across crate boundaries so other crates who depend on `telio-model` would get the `not(test)` branch all the time

jjanowsk: IMO it should default to false. In current form it will be defaulted to true also in production which is incorrect. It should default to false and be changed to true only via config during natlab tests

mathiaspeters: I don't know why it became an optional in the first place, but opinion of defaulting to `None`?

jjanowsk: It should also be fine, and maybe even the cleanest option

crates/telio-model/src/config.rs
```diff
@@ -194,6 +194,18 @@ pub struct Config {
     pub dns: Option<DnsConfig>,
 }
 
+impl Config {
+    /// Check if the config is empty
+    pub fn is_empty(&self) -> bool {
+        self.this.hostname.is_empty()
+            && self.this.identifier.is_empty()
+            && self.this.ip_addresses.is_none()
```
jjanowsk, line 202: Any reason to omit: `self.this.public_key`?

tomaszklak: Public key is non optional and it itself is wrapper over `[u8; KEY_SIZE]` so it never is empty (unless you would count something like `[0; KEY_SIZE]` as empty).

mathiaspeters: I think it was mainly a decision of convenience, influenced by exactly what Tomasz is saying. It will never be empty so the next best thing is to check that it's all 0, but if everything else in the config is empty, we can already say that the config is empty and skip that check. It's not the cleanest solution but it was the most convenient, and one made to keep the interface more similar. Though it's probably better, now that we don't deserialize strings into configs to not have the clients send None instead of an empty config if that's what they mean. After all, the clients should now better in which scenarios the config is empty by intention

tomaszklak:
I've checked the generated code and it looks like it solve the issue we had with using `repr(C)` for enums that @stalowyjez tried to solve here: https://bucket.digitalarsenal.net/low-level-hacks/vpn/client/libtelio-oss/-/merge_requests/144

Could you verify this by adding a test (i think it would have to be a C test) that tries to call new ffi functions with enum values out of range?

src/ffi/types.rs
```diff
@@ -231,6 +332,36 @@ map_enum! {
...
+    Error = Error,
+    Warn = Warning,
+    Info = Info,
+    Debug = Debug,
+    Trace = Trace,
+}
+
+map_enum! {
+    TelioLogLevel -> Level,
+    Critical = Error,
```
jjanowsk, line 357: Can we drop Critical from our interface since we modify it anyway?

mathiaspeters: Done

Cargo.toml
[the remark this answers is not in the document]
mathiaspeters: Done

crates/telio-model/src/api_config.rs
```diff
@@ -187,7 +187,7 @@ where
         Err(_) => return Ok(None),
     };
 
-    let eps: HashSet<_> = eps
+    let eps: Vec<_> = eps
```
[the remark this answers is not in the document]
mathiaspeters: This deserializer block might actually be possible to remove at some point, since we're no longer getting and deserializing strings for domain objects. I added deduplication of endpoint providers to the constructor of `Device`

crates/telio-model/src/api_config.rs
mathiaspeters: I tried using a type alias for `HashSet<EndpointProvider>` and I tried creating a wrapper, and neither worked. I think it should be possible to have a wrapper but so far I feel like it's much more work than it's worth given how little we actually use the endpoint providers. Instead I added deduplication of endpoint providers to the constructor for `Device`

mathiaspeters:
Regarding adding tests for the FFI bindings, as mentioned by Tomasz, in this PR:
When we're done migrating to UniFFI, all official apps will be using high-level languages to interact with libtelio, and the interface will be strongly typed, so adding tests for things like calling a function with an invalid enum value does not make that much sense. As for adding more general tests, a follow-up PR will be migrating the natlab tests to use these bindings instead of `tcli`, and these bindings will not be used before then, so adding dedicated FFI tests like in `ffi/tests` would also not bring much value.

tomaszklak (approved):
+1

crates/telio-model/src/api_config.rs
```diff
@@ -179,16 +191,21 @@ pub struct FeatureDerp {
 }
 
 /// Whether to validate keys
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
-pub struct FeatureValidateKeys(pub bool);
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Deserialize)]
+pub struct FeatureValidateKeys {
+    /// Defaults to true
+    pub validate_keys: bool,
+}
```
mathiaspeters, line 198: Reminder for myself: this change will require a change in the confluence page for libtelio features

crates/telio-model/src/api_config.rs
```diff
@@ -244,19 +249,27 @@ pub struct FeatureDerp {
...
 }
 
 /// Turns on connection resets upon VPN server change
 #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, Deserialize)]
 #[serde(transparent)]
-pub struct FeatureBoringtunResetConns(pub bool);
+pub struct FeatureBoringtunResetConns {
+    /// Defults to false
+    pub reset_connections: bool,
+}
```
mathiaspeters, line 272: Reminder for myself: this change will require a change in the confluence page for libtelio features

tests/uniffi_logger.rs
```diff
@@ -0,0 +1,42 @@
...
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    };
+
+    use telio::ffi_types::TelioLoggerCb;
+
+    use super::*;
+
+    #[test]
+    fn uniffi_test_logger() {
```
mathiaspeters, line 15: `tracing::set_global_default` is necessary to make this test work, but there can't be called twice in the same test file, that's why this test got it's own file instead of also placing it in `logger.rs`. The duplication will be fixed in a later PR

crates/telio-model/src/api_config.rs
mathiaspeters: I managed to fix it with a type alias

crates/telio-model/src/api_config.rs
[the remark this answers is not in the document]
mathiaspeters: No longer relevant

crates/telio-model/src/api_config.rs
tomaszklak: why is that? It seems that the json format is the same 🤔

crates/telio-model/src/api_config.rs
tomaszklak: I think if you add `repr(transparent)` you will keep the json format unchanged.

src/device/mod.rs
```diff
@@ -2728,10 +2728,13 @@ mod tests {
 
         let (sender, _receiver) = tokio::sync::broadcast::channel(1);
 
-        let mut providers = HashSet::<telio_model::api_config::EndpointProvider>::new();
-        providers.insert(telio_model::api_config::EndpointProvider::Stun);
-        providers.insert(telio_model::api_config::EndpointProvider::Upnp);
-        providers.insert(telio_model::api_config::EndpointProvider::Local);
+        let providers = vec![
```
tomaszklak, line 2731: nit: you don't need a vector, an array should also work but you can use `maplit::hashset` instead of creating temporary collection.

src/ffi/mod.rs
```diff
@@ -85,12 +119,604 @@ lazy_static::lazy_static! {
...
+    }
+
+    /// Explicitly deallocate telio object and shutdown async rt.
+    pub fn destroy_hard(&self) -> FFIResult<()> {
+        let mut dev = self
+            .inner
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        let res = catch_ffi_panic(|| {
+            dev.try_shutdown(Duration::from_millis(1000))
```
tomaszklak, line 303: Previously `try_shutdown` was destroying `dev` here (and thus deallocating whole object tree starting in `dev`). Why is this no longer done?

src/ffi/mod.rs
```diff
@@ -1063,8 +1689,70 @@ impl Subscriber for TelioTracingSubscriber {
     }
 }
 
+pub struct UniffiTelioTracingSubscriber {
```
tomaszklak, line 1692: Why is this a copy pasted version of `TelioTracingSubscriber`?

src/ffi/mod.rs
```diff
@@ -85,12 +119,604 @@ lazy_static::lazy_static! {
...
+                        .set(format!("{}", info)),
+                );
+
+                telio_log_debug!("call_once: {:?}", e);
+                events(e);
+            }));
+        });
+
+        // TODO: Update windows ffi to take in void*, for protect
+        #[cfg(not(target_os = "android"))]
```
packgron, line 250: Do we really need this to be gatted under android. I think the option is universal enough, and in the future someone needs to customize this, I see no problem with that.

src/ffi/mod.rs
```diff
@@ -85,12 +119,604 @@ lazy_static::lazy_static! {
...
+        catch_ffi_panic(|| {
+            dev.stop();
+            Ok(())
+        })
+    }
+
+    /// get device luid.
+    pub fn get_adapter_luid(&self) -> u64 {
+        match self.inner.lock() {
+            Ok(mut d) => d.get_adapter_luid(),
```
packgron, line 427: What will be returned for unix platforms ?

src/lib.rs
```diff
@@ -51,3 +53,124 @@ pub use telio_lana;
...
+
+    impl UniffiCustomTypeConverter for PublicKey {
+        type Builtin = String;
+
+        fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
+            Ok(val.parse().map_err(|_| TelioError::InvalidKey)?)
+        }
+
+        fn from_custom(obj: Self) -> Self::Builtin {
+            String::from_utf8_lossy(&obj.0).to_string()
```
tomaszklak, line 80: I don't think this is correct, why not instead convert it to something like base64?

src/lib.rs
```diff
@@ -51,3 +53,124 @@ pub use telio_lana;
 
 /// cbindgen:ignore
 pub use telio_firewall;
+
+pub use uniffi_libtelio::*;
+#[allow(clippy::panic, clippy::unwrap_used, clippy::expect_used, unwrap_check)]
+mod uniffi_libtelio {
```
tomaszklak, line 59: All of the conversion here should have tests.

src/libtelio.udl
```diff
@@ -0,0 +1,572 @@
...
+    /// Derp will send empty messages after this many seconds of not sending/receiving any data [default 60s]
+    u32? derp_keepalive;
+    /// Enable polling of remote peer states to reduce derp traffic
+    boolean? enable_polling;
+    /// Use Mozilla's root certificates instead of OS ones [default false]
+    boolean use_built_in_root_certificates;
+};
+
+/// Whether to validate keys
+dictionary FeatureValidateKeys {
```
packgron, line 450: maybe this just be a boolean?

src/libtelio.udl
```diff
@@ -0,0 +1,572 @@
...
+    /// Defaults to true
+    boolean validate_keys;
+};
+
+/// Turns on connection resets upon VPN server change
+dictionary FeatureBoringtunResetConns {
```
packgron, line 456: maybe this should just be a boolean?

src/libtelio.udl
```diff
@@ -0,0 +1,572 @@
...
+callback interface TelioLoggerCb {
+    void log(TelioLogLevel log_level, string payload);
+};
+
+callback interface TelioProtectCb {
+    void protect(i32 payload);
+};
+
+/// Encompasses all of the possible features that can be enabled
+dictionary Features {
```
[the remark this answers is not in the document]
tomaszklak: I imagine that this can make going back and forth between libtelio versions slightly harder, but that is mostly a concern for the apps dev teams so I think they should also be involved here.

src/ffi/mod.rs
mathiaspeters: Because the old code is still alive, but the new code should still be tested. There will be a cleanup PR later where all of the duplication introduced here is removed (the `uniffi_` and `UniffI` prefixes will be removed at the same time)

src/ffi/mod.rs
mathiaspeters: Because UniFFI doesn't allow you to take ownership of self. It doesn't even allow you to take a mutable reference to self

src/ffi/mod.rs
mathiaspeters: I was following the old logic here, but I'm happy to not conditionally compile this

src/ffi/mod.rs
mathiaspeters: `boringtun` and `linux-native` are hard coded to return 0 in `boring.rs` and `linux_native_wg.rs`

src/ffi/mod.rs
tomaszklak: I think that if you change the `TelioTracingSubscriber` to also use `TelioLoggerCb` there will be no need for `UniffiTelioTracingSubscriber`.

src/ffi/mod.rs
tomaszklak: I understand, but why are we no longer destroying that objects/memory - is that still correct behaviour? Maybe `telio` should store represent `inner` as `Mutex<Option<Device>>` and set it to `None` as part of the `try_shutdown`?

mathiaspeters: That sounds like a valid option. I will make the change

src/libtelio.udl
mathiaspeters: Yes, I updated these to not have to do custom conversion logic for UniFFI but of course it's better to keep the API simple than to avoid such conversions. Fixed

src/libtelio.udl
mathiaspeters: Yes, I updated these to not have to do custom conversion logic for UniFFI but of course it's better to keep the API simple than to avoid such conversions. Fixed

src/lib.rs
mathiaspeters: Added

src/lib.rs
mathiaspeters: I updated these to use `Vec<u8>`instead of `String` since that is what the keys use under the hood. I think I remember getting some errors when I tried to do that in the past, but it worked now

src/ffi/mod.rs
mathiaspeters: I added an enum to wrap the callback

src/device/mod.rs
mathiaspeters: Fixed

crates/telio-model/src/api_config.rs
mathiaspeters: Yes, I updated these to not have to do custom conversion logic for UniFFI but of course it's better to keep the API simple than to avoid such conversions. Fixed

crates/telio-model/src/api_config.rs
mathiaspeters: Yes, I updated these to not have to do custom conversion logic for UniFFI but of course it's better to keep the API simple than to avoid such conversions. Fixed

src/ffi/mod.rs
mathiaspeters: It seems like file descriptors are u64 on windows but i32 on all other platforms. I can either try to dig into the code and add a conversion wherever windows is calling it, or a simpler option, exclude the protect closure on windows. Opinion?

packgron (approved):
+1.0

src/ffi/mod.rs
packgron: so let's this this for unix only

packgron (approved):

PR #50, opened 2023-08-02
Add task monitor

packgron:
### Description
Improving telio-task

### Changes
Split telio_task::task into 3 smaller parts: (see first commit)
- runtime
- error
- wait_response

Added tool to monitor behavior of telio task's, enabled by telio_monitor feature.
This tool will report:
- start of task
- stop of task and reason for that stop
- excessive looping of a task

This feature is now used with all integration test's

To make Errors reported by Task's a bit more useful in `trait Runtime` requires them to implement `std::error::Error` trait.
For task, that does not actually produces failure error, a new std::convert::Infallible should be used

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -263,6 +263,47 @@ where
     Ok(Some(eps))
 }
 
+/// Telio task monitor configuration
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
+pub struct FeatureTaskMonitor {
+    /// The number of consecutive wait iterations that should trigger busy loop detection.
+    /// This typically occurs when erroneous code results in a wait loop without any awaits.
+    #[serde(default = "FeatureTaskMonitor::default_max_polls_immediately")]
+    pub max_polls_immediately: Option<u64>,
+
+    /// The frequency at which the same event can be reported by the task monitor, in milliseconds.
+    #[serde(default = "FeatureTaskMonitor::default_duration_between_reporting_same_event")]
+    pub duration_between_reporting_same_event_ms: Option<u64>,
+}
+
+impl FeatureTaskMonitor {
+    /// Default for [FeatureTaskMonitor::max_polls_immediately]
+    pub const DEFAULT_MAX_POLLS_IMMEDIATELY: u64 = 10_000;
+
+    /// Default for [FeatureTaskMonitor::duration_between_reporting_same_event]
+    pub const DEFAULT_DURATION_BETWEEN_REPORTING_SAME_EVENT: u64 = 1000;
+
+    /// Default for [FeatureTaskMonitor::max_polls_immediately]
+    pub const fn default_max_polls_immediately() -> Option<u64> {
+        Some(Self::DEFAULT_MAX_POLLS_IMMEDIATELY)
+    }
+
+    /// Default for [FeatureTaskMonitor::duration_between_reporting_same_event]
+    pub const fn default_duration_between_reporting_same_event() -> Option<u64> {
+        Some(Self::DEFAULT_DURATION_BETWEEN_REPORTING_SAME_EVENT)
+    }
+}
+
+impl Default for FeatureTaskMonitor {
+    fn default() -> Self {
+        Self {
+            max_polls_immediately: Self::default_max_polls_immediately(),
+            duration_between_reporting_same_event_ms:
+                Self::default_duration_between_reporting_same_event(),
+        }
+    }
+}
+
 #[derive(Clone, Debug, Default, PartialEq, Eq, Deserialize)]
 /// Encompasses all of the possible features that can be enabled
 pub struct Features {
@@ -288,6 +329,11 @@ pub struct Features {
     /// Flag to specify if keys should be validated
     #[serde(default)]
     pub validate_keys: FeatureValidateKeys,
+    /// Avoid sending periodic messages to peers with no traffic reported by wireguard
+    #[serde(default)]
+    pub skip_unresponsive_peers: bool,
+    /// Telio task monitor configuration
+    pub task_monitor: Option<FeatureTaskMonitor>,
 }
 
 impl FeaturePaths {
@@ -341,7 +387,8 @@ mod tests {
                 "force": "relay"
             },
             "direct": {},
-            "exit_dns": {}
+            "exit_dns": {},
+            "task_monitor": {}
         }"#;
 
     const CORRECT_FEATURES_JSON_WITH_ALL_OPTIONAL: &str = r#"
@@ -388,7 +435,9 @@ mod tests {
                 "derp_keepalive": 2,
                 "enable_polling": true
             },
-            "validate_keys": false
+            "validate_keys": false,
+            "skip_unresponsive_peers": true,
+            "task_monitor": {}
         }"#;
 
     static EXPECTED_FEATURES: Lazy<Features> = Lazy::new(|| Features {
@@ -433,6 +482,8 @@ mod tests {
             enable_polling: Some(true),
         }),
         validate_keys: FeatureValidateKeys(false),
+        skip_unresponsive_peers: true,
+        task_monitor: Some(FeatureTaskMonitor::default()),
     });
 
     static EXPECTED_FEATURES_WITHOUT_TEST_ENV: Lazy<Features> = Lazy::new(|| Features {
@@ -471,6 +522,8 @@ mod tests {
         is_test_env: None,
         derp: None,
         validate_keys: Default::default(),
+        skip_unresponsive_peers: Default::default(),
+        task_monitor: Some(FeatureTaskMonitor::default()),
     });
 
     #[test]
@@ -638,6 +691,8 @@ mod tests {
             exit_dns: None,
             derp: None,
             validate_keys: Default::default(),
+            skip_unresponsive_peers: Default::default(),
+            task_monitor: None,
         };
 
         let empty_qos_features = Features {
@@ -660,6 +715,8 @@ mod tests {
             exit_dns: None,
             derp: None,
             validate_keys: Default::default(),
+            skip_unresponsive_peers: Default::default(),
+            task_monitor: None,
         };
 
         let no_qos_features = Features {
@@ -677,6 +734,8 @@ mod tests {
             exit_dns: None,
             derp: None,
             validate_keys: Default::default(),
+            skip_unresponsive_peers: Default::default(),
+            task_monitor: None,
         };
 
         assert_eq!(from_str::<Features>(full_json).unwrap(), full_features);
@@ -713,6 +772,8 @@ mod tests {
             }),
             derp: None,
             validate_keys: Default::default(),
+            skip_unresponsive_peers: Default::default(),
+            task_monitor: None,
         };
 
         let empty_features = Features {
@@ -726,6 +787,8 @@ mod tests {
             }),
             derp: None,
             validate_keys: Default::default(),
+            skip_unresponsive_peers: Default::default(),
+            task_monitor: None,
         };
 
         assert_eq!(from_str::<Features>(full_json).unwrap(), full_features);
@@ -763,6 +826,8 @@ mod tests {
             direct: None,
             derp: None,
             validate_keys: Default::default(),
+            skip_unresponsive_peers: Default::default(),
+            task_monitor: None,
         };
 
         assert_eq!(Features::default(), expected_defaults);
--- a/crates/telio-proxy/src/proxy.rs
+++ b/crates/telio-proxy/src/proxy.rs
@@ -2,6 +2,7 @@ use async_trait::async_trait;
 use futures::future::{pending, select_all, FutureExt};
 use std::{
     collections::{HashMap, HashSet},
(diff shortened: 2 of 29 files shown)


nat-lab/tests/telio_features.py
```diff
@@ -39,7 +39,13 @@ class Nurse:
...
     initial_heartbeat_interval: Optional[int] = None
     qos: Optional[Qos] = None
-    enable_nat_type_collection: bool = False
+
+
+@dataclass_json(undefined=Undefined.EXCLUDE)
+@dataclass
+class TaskMonitor:
+    max_poll_imidiates: Optional[int]
+    duration_between_reporting_same_event: Optional[int]
```
gytsto, line 48: Just curious, does it automatically default to `None`?

crates/telio-task/src/task/runtime.rs
```diff
@@ -24,13 +25,14 @@ pub trait Runtime: Sized {
     const NAME: &'static str;
 
     /// Error that may occur in [Task]
-    type Err: Send + 'static;
+    // TODO: Improve error reporting by adding Error trait
```
tomaszklak, line 28: `Error` is already added, this todo can be removed

crates/telio-task/src/task/runtime.rs
```diff
@@ -51,127 +53,64 @@ pub trait Runtime: Sized {
...
 where
     S: Runtime + Send + 'static,
 {
     /// Start a new task for state
     pub fn start(mut state: S) -> Self {
-        let stop = Arc::new(Notify::new());
+        let Chan {
+            tx: stop,
+            rx: mut stopped,
+        } = Chan::<StopKind>::new(1);
```
tomaszklak, line 78: maybe a `tokio::sync::oneshot` would be better here, since I expect that only one message should be sent over this channel?

crates/telio-task/src/task/runtime.rs
```diff
@@ -200,25 +139,25 @@ where
 
     /// Stop task
     pub async fn stop(mut self) -> StopResult<S::Err> {
-        println!("task stopped - {}", S::NAME);
-        self.stop.notify_one();
+        // We can ignore error, if fail to send stop
+        // it means that inner task was already stopped
+        // and reciever part was dropped
+        let _ = self.stop.send(StopKind::Manual).await;
```
tomaszklak, line 145: maybe a warning here would be useful to see if we have any such cases where we try to explicitly stop tasks no longer running?

crates/telio-traversal/src/endpoint_providers/local.rs
```diff
@@ -270,11 +270,12 @@ impl<T: WireGuard, G: GetIfAddrs> Runtime for State<T, G> {
...
-                self.handle_rx_packet(buf, &addr).await.unwrap_or_else(
-                    |e| {
-                        telio_log_warn!("Failed to handle packet received no local interface endpoint provider {:?}", e);
-                    });
+                if let Some(buf) = rx_buff.get(..len) {
+                    self.handle_rx_packet(buf, &addr).await.unwrap_or_else(
+                        |e| {
+                            telio_log_warn!("Failed to handle packet received no local interface endpoint provider {:?}", e);
+                        });
+                }
```
tomaszklak, line 278: nit: maybe an else branch with `unreachable!`? :)

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -527,11 +528,15 @@ impl<Wg: WireGuard, E: Backoff> Runtime for State<Wg, E> {
...
+                if let Some(buf) = ext_buf.get(..size) {
+                    let _ = self.handle_rx(buf, &src_addr).await;
+                }
             }
             Ok((size, src_addr)) = self.tun_socket.recv_from(&mut tun_buf) => {
                 // We will not pinging through wireguard.
-                let _ = self.try_handle_stun_rx(tun_buf.get(..size).ok_or(())?, &src_addr).await;
+                if let Some(buf) = tun_buf.get(..size) {
+                    let _ = self.try_handle_stun_rx(buf, &src_addr).await;
+                }
```
tomaszklak, line 539: nit: maybe an else branch with unreachable!? :)

crates/telio-utils/src/never.rs
```diff
@@ -0,0 +1,16 @@
+use std::{error::Error, fmt::Display};
+
+/// Error type that is impossible to construct
```
tomaszklak, line 3: That's not correct, it's possible to construct but it will fail at runtime when printed using `Display`.

crates/telio-utils/src/never.rs
```diff
@@ -0,0 +1,16 @@
...
+/// This should be usefull until [never] is stabalized
+///
+/// [nver]: https://doc.rust-lang.org/std/primitive.never.html
+#[derive(Debug)]
+pub struct Never(());
+
+impl Display for Never {
+    fn fmt(&self, _: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        unreachable!("Imposible to construct this type")
```
tomaszklak, line 12: **If** we want keep `Never`, I think that this should be `unimplemented!`

crates/telio-utils/src/never.rs
packgron: How would you construct it outside of this file ?

crates/telio-utils/src/never.rs
```diff
@@ -0,0 +1,16 @@
+use std::{error::Error, fmt::Display};
+
+/// Error type that is impossible to construct
+/// This should be usefull until [never] is stabalized
+///
+/// [nver]: https://doc.rust-lang.org/std/primitive.never.html
```
tomaszklak, line 6: typo, should be `never`

crates/telio-proxy/src/proxy.rs
```diff
@@ -230,12 +230,10 @@ impl Runtime for StateIngress {
...
-                let msg = DataMsg::new(if let Some(buf) = self.read_buf.get(..n) {
-                    buf
-                } else {
-                    return Self::error(());
-                });
-                let _ = permit.send((pk, msg));
+                if let Some(buf) = self.read_buf.get(..n) {
+                    let msg = DataMsg::new(buf);
+                    let _ = permit.send((pk, msg));
+                }
```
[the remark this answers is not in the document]
packgron: So get it self I don't think as an issue, it should not really be possible to overstep n, as it will always be smaller that read_buf.

But probably a better question is what happens if socket for some reason dies.
It may cause a busy loop, and we should probably decide what we should do in such a case.

crates/telio-utils/src/never.rs
tomaszklak: Good point, but I think that a more obviously impossible type empty enum, something like `enum Never {}`.

crates/telio-relay/src/multiplexer/inout.rs
```diff
@@ -25,21 +25,23 @@ where
...
-            _ => None,
-        })
+    pub fn joined(&mut self) -> &mut T {
+        self.try_in_place(Self::into_joined)
+            .expect("InOut always works with same underlying channel and it cannot fail on join");
+
+        match &mut self.0 {
+            Joined(j) => j,
+            _ => unreachable!("into_join enusures self will always be Joined"),
+        }
```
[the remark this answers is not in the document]
packgron: I have interior mutability here, and it's hard to ensure this statically, this entire struct is basically for ensuring this in variance, so it would pretty for upper use in other area :/

crates/telio-task/src/task/monitor.rs
```diff
@@ -0,0 +1,456 @@
...
+    ) -> impl Future<Output = Result<StopKind, R::Err>> + Send + 'static {
+        let handler = self.handler.clone();
+        async move {
+            (handler)(MonitorEvent::TaskStart { name: R::NAME });
+
+            // This is the simplest way to capture panic.
+            // It should be possible to do this with catch_unwind,
+            // but it is way too complex to implement for a functionality that
+            // is primarily used for debugging.
+            let res = tokio::spawn(inner).await;
```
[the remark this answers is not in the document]
packgron: Nice point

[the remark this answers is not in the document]
packgron: Nope it won't work, spawn local would just lock to one thread, not to that upper ant inner task run on the same thread,

crates/telio-utils/src/never.rs
packgron: Yes I think it is more idiomatic way, but I will use Infallible, I somehow did not like at first as it comes from convert and felt like it's inteded for conversions only :sweat: but it really does not matter

crates/telio-relay/src/multiplexer/inout.rs
```diff
@@ -25,21 +25,23 @@ where
...
 
     /// Get references to split halves
-    pub fn split(&mut self) -> Option<(&mut SplitSink<T, I>, &mut SplitStream<T>)> {
+    pub fn split(&mut self) -> (&mut SplitSink<T, I>, &mut SplitStream<T>) {
         self.in_place(Self::into_split);
         match &mut self.0 {
-            Split(tx, rx) => Some((tx, rx)),
-            _ => None,
+            Split(tx, rx) => (tx, rx),
+            _ => unreachable!("After in_place(Self::into_split) self will always be split"),
```
tomaszklak, line 32: What about the `Empty` case? `into_split` is not making it into `Split`.

crates/telio-relay/src/multiplexer/inout.rs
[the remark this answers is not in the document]
tomaszklak: Now that I think about it, I'm worried that this is incorrect. For example in case of `fn split` the assumption expressed in the `unreachable!` is not correct - `self.0` can be still `Empty`.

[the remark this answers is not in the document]
packgron: Well it's something that should not happen, the only time I tun self into empty wile using in_place.
There are no other ways to construct Empty, I think I will remove try join so it would not confuse this.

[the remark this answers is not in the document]
packgron: I will queue property based test for this one :D

crates/telio-proxy/src/proxy.rs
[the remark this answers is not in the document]
packgron: Created LLT-4406 to address this

crates/telio-relay/src/multiplexer/inout.rs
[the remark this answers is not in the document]
packgron: LLT-4407 created for this

crates/telio-task/src/task/monitor.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        if self.immediate_poll_count > max_polls {
+            self.immediate_poll_count = 0;
+            self.monitor.report(MonitorEvent::TaskBusyLoop {
+                name: self.task_name,
+            });
+            task::yield_now().await;
+        }
+
+        tokio::pin!(step);
```
tomaszklak, line 167: We can now use `std::pin::pin`.

tomaszklak (commented):
+0.9

crates/telio-model/src/api_config.rs
```diff
@@ -222,6 +222,47 @@ where
     Ok(Some(eps))
 }
 
+/// Telio task monitor configuration
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
+pub struct FeatureTaskMonitor {
+    /// The number of consecutive wait iterations that should trigger busy loop detection.
+    /// This typically occurs when erroneous code results in a wait loop without any awaits.
+    #[serde(default = "FeatureTaskMonitor::default_max_polls_immediately")]
+    pub max_polls_immediately_ms: Option<u64>,
```
mathiaspeters, line 231: The comment talks about iterations but the naming of the field mentions time (ms), unless ms means something else in this case, in which case that should be clarified

crates/telio-model/src/api_config.rs
```diff
@@ -222,6 +222,47 @@ where
...
+
+    /// The frequency at which the same event can be reported by the task monitor, in milliseconds.
+    #[serde(default = "FeatureTaskMonitor::default_duration_between_reporting_same_event")]
+    pub duration_between_reporting_same_event_ms: Option<u64>,
+}
+
+impl FeatureTaskMonitor {
```
mathiaspeters, line 238: I my opinion the constants should either be removed or made private. The functions should be enough

mathiaspeters:
+0.5

matislovas:
+0.5 leavins the final word to @Jauler  :smiley_cat: 

nit: IMO, this needs to be chopped into several commits.

crates/telio-model/src/api_config.rs
packgron: I need non optional values for place I use them, initialy i designed so that it would just have a default value, but then you have an issue then python test's pass a null. We can probably just fix python and say that they are not allowed to be set to null.

But I will create an improvement task for api_config.rs in general

PR #51, opened 2023-08-02
Stop sending handshake initations to offline peers

baltrus1:
* Implement derp polling for remote peer states
* Add poll request/response message types
* Turn off keepalives when peers are inactive according to derp

### Description
This implements a polling mechanism that asks server if peers with specified keys are connected to derp cluster or not. If we get information, that some peers are offline, we turn off persistent keepalives for those peers locally, so after REKEY_ATTEMPT_TIME when handshake timer expires passive keepalive packet would not reset that timer.
Poll message is send every DERP_KEEPALIVE seconds, this value is configurable and defaults to 60 seconds.
The polling itself needs to be turned on to use this mechanism via features:derp:enable_polling boolean.

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-relay/src/derp/proto.rs
```diff
@@ -122,6 +122,9 @@ enum FrameType {
     Ping = 0x12,
     /// 8 byte payload, the contents of the ping being replied to
     Pong = 0x13,
+    /// controll message for communication with derp. Since these messages are not
+    /// for communication with other peers through derp, they don't contain public_key
+    ControlMessage = 0x14,
```
packgron, line 127: This may need a longer discussion together with @Jauler.
There is kindof two approaches here:
- Current one where we basically have 3 types on channels with relay:
    - Node to Relay, 
    - Node to Relay to Node
    - Node to Node
   And inner telio-proto type specifies the actual type, but it's a bit weird, because it's not that clear from telio-proto side witch now goes where.
   We already have some confusion with Node to Node vs Node to Relay to Node, and now we are adding more.
   
 - Second approch would be to add Request, Response directly here, without extra wrapping layer, it would be a bit cleaner from protocol side, but I think it will complicate message passing to other telio components :/

packgron (changes requested):

crates/telio-relay/src/derp/proto.rs
baltrus1: I did use the second approach at the start, but it looked pretty messy as the only distinction for the new derp message types was that they don't have public keys, so I wanted to add some generic message type for Node to Relay messages

crates/telio-relay/src/derp/http.rs
```diff
@@ -55,9 +56,12 @@ pub type Protect = Arc<dyn Fn(i32) + Send + Sync + 'static>;
 const MAX_TCP_PACKET_SIZE: usize = u16::MAX as usize;
 
 pub struct DerpConnection {
-    pub comms: Chan<(PublicKey, Vec<u8>)>,
+    pub comms: Chan<(Option<PublicKey>, Vec<u8>)>,
```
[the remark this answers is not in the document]
baltrus1: Yep this is for derp module -> derp connection communication, so since we don't have public key in messages directly with derp, made this into optional

crates/telio-relay/src/derp/mod.rs
```diff
@@ -103,6 +106,10 @@ struct State {
     rng: StdRng,
     /// Used to get external sockets
     socket_pool: Arc<SocketPool>,
+    /// Used to match derp poll request with response
+    derp_poll_session: Session,
```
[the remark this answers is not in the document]
baltrus1: There isn't really "no session", I just made first session start with zero, but the point of this number is to match request with response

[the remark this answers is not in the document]
packgron: I guess it does not really matter then

crates/telio-relay/src/derp/proto.rs
Jauler: Would it make sense to use one channel from telio-relay for `Node -> Relay -> Node` messages and another for `Node -> Relay` messages? 

`Node -> Node` case does not go through telio-relay, so I am thinking, that probably there is no need to distinguish it into its own separate type.

crates/telio-proto/src/packet/derppoll.rs
```diff
@@ -0,0 +1,399 @@
...
+/// ```
+#[derive(Debug, PartialEq, Clone)]
+pub struct DerpPollResponseMsg {
+    session: Session,
+    msg: DerpPollResponse,
+}
+
+impl DerpPollResponseMsg {
+    /// Creates new request message from `peers`.
+    pub fn new(peer_states: HashMap<PublicKey, bool>, session: Session) -> Self {
```
packgron, line 172: NITPICK: I personaly would find (session, peer_states) param order a bit more ergonomic.

crates/telio-proto/src/packet/mod.rs
packgron, line 1: It feels to me that it would make sence to split packet/ into either

```
packet/
    relayed/
    control/
```

or 

```
packet_relayed/
packet_control/
```

crates/telio-relay/src/derp/http.rs
```diff
@@ -185,24 +190,33 @@ async fn connect_and_start<RW: AsyncRead + AsyncWrite + Send + 'static>(
...
         }),
         join_receiver: tokio::spawn(async move {
-            start_write(writer, rx, addr)
+            start_write(writer, receiver_relayed, receiver_direct, addr)
                 .await
                 .map_err(|err| IoError::new(ErrorKind::Other, err.to_string()))
         }),
+        poll_timer: {
+            let poll_interval = Duration::from_secs(server_keepalives.derp_keepalive as u64);
+            interval_at(tokio::time::Instant::now() + poll_interval, poll_interval)
```
packgron, line 218: Set missed tick behavior to [Delay](https://docs.rs/tokio/1.32.0/tokio/time/enum.MissedTickBehavior.html#variant.Delay)

Default behavior is Burst, meaning that after sleep you would fire sleep / poll_interval event's. This would be an unnecessary DDOS to relay servers :D

crates/telio-relay/src/derp/http.rs
```diff
@@ -55,9 +56,13 @@ pub type Protect = Arc<dyn Fn(i32) + Send + Sync + 'static>;
...
 
 pub struct DerpConnection {
-    pub comms: Chan<(PublicKey, Vec<u8>)>,
+    pub comms_relayed: Chan<(PublicKey, Vec<u8>)>,
+    pub comms_direct: Chan<Vec<u8>>,
     pub join_sender: JoinHandle<Result<(), IoError>>,
     pub join_receiver: JoinHandle<Result<(), IoError>>,
+
+    /// For polling derp about remote peers states
+    pub poll_timer: Interval,
```
packgron, line 65: This is this together with DerpConnection

crates/telio-model/src/api_config.rs
```diff
@@ -176,6 +176,8 @@ pub struct FeatureDerp {
     pub tcp_keepalive: Option<u32>,
     /// Derp will send empty messages after this many seconds of not sending/receiving any data [default 60s]
     pub derp_keepalive: Option<u32>,
+    /// Enable polling of remote peer states to reduce derp traffic
+    pub enable_polling: Option<bool>,
```
packgron, line 180: Is there a hard limit on what poll timer should be, or would it make sence to make it remote configurable ?

packgron (changes requested):

crates/telio-model/src/api_config.rs
baltrus1: It's already configurable, poll timer will use derp_keepalive interval as this polling is a replacement for derp keepalives

packgron:
+2.0

packgron (approved):

PR #52, opened 2023-08-03
Use synthetic time in firewall tests

tomaszklak:
Instead of using real time/clock which can be imprecise and cause flakiness, use simulated time that we have full control over in the tests.

Dependencies in Cargo.toml also have been sorted.


baltrus1 (approved):
+1

lcruz99:
+1

lcruz99 (approved):
+1

PR #53, opened 2023-08-03
nat-lab: Add tests parametrization

lcruz99:
### Description
Some of the tests are only running on linux, excluding windows, macos, besides every possible adapter.

This PR adds `@pytest.mark.parametrize` decorator for each test to run for every platform/adapter combination, where it makes sense.

This changes add 174 tests (221->395), significantly increasing the duration of the nat-lab job.

- [x] changelog.md is updated
- [x] Commit history is clean
- [x] README.md is updated

### DIFF
--- a/nat-lab/pyproject.toml
+++ b/nat-lab/pyproject.toml
@@ -52,7 +52,8 @@ disable = [
     "global-statement",
     "bare-except",
     "too-many-branches",
-    "duplicate-code"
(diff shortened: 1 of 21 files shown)


lcruz99:
Due to the big number of tests this PR adds it was decided with @gytsto to put it on hold while we don't have a low natlab flakyness rate.

PR #59, opened 2023-08-07
LLT-4023: Validate meshnet config public key

mathiaspeters:
### Description
Validate that the public key associated with a requested meshnet config matches the private key of the device the config will be applied to

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -90,6 +90,8 @@ pub enum Error {
     AdapterConfig(String),
     #[error("Private key does not match meshnet config's public key.")]
     BadPrivateKey,
+    #[error("Public key in requested config does not match the device's private key.")]
+    BadPublicKey,
     #[error("Invalid node configuration")]
     InvalidNode,
     #[error("Dublicate exit node")]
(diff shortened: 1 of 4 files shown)


arvydasv (commented):
+1

mathiaspeters:
> Do not forget to update changelog ;)

Fixed!

src/device/mod.rs
```diff
@@ -1163,6 +1165,12 @@ impl Runtime {
     }
 
     async fn set_config(&mut self, config: &Option<Config>) -> Result {
+        if let Some(cfg) = config {
+            if cfg.this.public_key != self.get_private_key().await.map(|key| key.public())? {
+                return Err(Error::BadPublicKey);
```
Jauler, line 1170: Could you also add `telio_log_err()` statement here? If we hit this case within the apps -> it will help diagnose it immediately.

mathiaspeters: It will already be logged in `ffi/mod.rs L905` so I don't think that's necessary

Jauler: true

Jauler (dismissed):
+2.0

src/device/mod.rs
```diff
@@ -1043,6 +1045,12 @@ impl Runtime {
     async fn set_private_key(&mut self, private_key: &SecretKey) -> Result {
         // TODO: create a global controll state to consolidate all entities
 
+        let should_validate_keys = self.features.validate_keys.0;
+        let meshnet_is_on = self.requested_state.meshnet_config.is_some();
+        if should_validate_keys && meshnet_is_on {
```
Jauler, line 1050: This is probably not too relevant, but if the `set_private_key` is called with the key, which is already set, probably no error should be thrown, as the main point of the validation is to ensure, that public key in meshmap and private key in WG are from the same pair. This would throw error even if they are of the same pair.

Jauler: Other than this, +2.0

mathiaspeters: I skipped adding that at first because it doesn't make sense to try to set the same key again, but you are right. It's quite the unnecessary error to return, so I fixed it

Jauler (approved):
+2.0

PR #63, opened 2023-08-08
Add scheduled gitlab runs

tomaszklak:
### Description

Start gitlab pipelines every 30 minutes during the night. This is very similar to the old gitlab based scheduled pipelines. We use the new functionality of the action to mark our scheduled pipelines. This will be used later on to collect them and generate summaries and statistics of the failing nat-lab tests.

The schedule is slightly different from the old gitlab one, to avoid the crowded full hour mark.

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -3,12 +3,19 @@ on:
   pull_request_target:
     types: [labeled]
   push:
+  schedule:
+    # Avoid running on 0 since a lot of other workflows on github start at that
+    # time and this can cause delays or even dropping of jobs:
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
+    - cron: 18,48 18-23 * * *
+    - cron: 17,47 0-5 * * *
 permissions: {}
 
 jobs:
   trigger-gitlab-pipeline:
     runs-on: [self-hosted, libtelio]
     if: |
+      github.event_name == 'schedule' ||
       github.event_name == 'push' ||
       (
         github.event_name == 'pull_request_target' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/gitlab.yml
```diff
@@ -3,22 +3,30 @@ on:
   pull_request_target:
     types: [labeled]
   push:
+  schedule:
+    # Avoid running on 0 since a lot of other workflows on github start at that
+    # time and this can cause delays or even dropping of jobs:
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
+    - cron: 18,48 21-23 * * *
+    - cron: 17,47 0-8 * * *
```
gytsto, line 11: is it UTC, like documentation says? If yes, it will be running from 8:00 - 11:00 Lithuanian time, which is not that good. Also will start late. :smile:

tomaszklak: Good point! I copied the cron config from gitlab and didn't realized that it was set to be in LT timezone. I will move it back by 3 hours

gytsto: also weird configuration of minutes, can't you just start write something like `15-59`

.github/workflows/gitlab.yml
```diff
@@ -3,22 +3,30 @@ on:
...
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
+    - cron: 18,48 18-23 * * *
+    - cron: 17,47 0-5 * * *
 permissions: {}
 
 jobs:
   trigger-gitlab-pipeline:
     runs-on: [self-hosted, libtelio]
     if: |
+      github.event.schedule ||
```
jjanowsk, line 18: can you please use: `github.event_name == 'schedule'` to be consistent with the rest of the `if`? Same in the second place. Sometimes different events can have the same payload type so this is clearer IMO

.github/workflows/gitlab.yml
tomaszklak: The doc linked in the comment explains that a little. Basically github can delay or even drop scheduled workflows when there are too many workflows scheduled at a given time. They specifically mention full hours but as you can imagine round number of minutes (0,15,30,45) probably also are more crowded. To avoid that I've selected those hopefully less crowded minutes to start this workflow.

gytsto: ok, this is kinda reasonable. I probably missunderstood abit, cuz  I tought `schedule: ${{ github.event.schedule != null }}`, this would not allow to queue pipeline if one is already queued

.github/workflows/gitlab.yml
tomaszklak: Good point! Done :)

gytsto (approved):
lgtm +1

jjanowsk (approved):
+2

PR #65, opened 2023-08-10
LLT-4013: Add conntrack logic for ICMP

mathiaspeters:
### Description
So far there has been no conntrack logic for ICMP. The firewall flow for an inbound ICMP packet has been:
1. If peer is on whitelist, allow
2. If ICMP packet type is blocked, block
3. Allow

With these changes, outbound ICMP connections create cache entries, and the firewall flow becomes:
1. If peer is on whitelist, allow
2. If peer connection is in the cache, allow
3. If ICMP packet type is blocked, block
4. Allow

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -13,6 +13,7 @@ use pnet_packet::{
     Packet,
 };
 use std::{
+    convert::TryInto,
     fmt::{Debug, Formatter},
     net::{IpAddr as StdIpAddr, Ipv4Addr as StdIpv4Addr, Ipv6Addr as StdIpv6Addr},
     sync::{Mutex, RwLock},
@@ -218,6 +219,8 @@ pub struct StatefullFirewall {
     udp: Mutex<LruCache<IpConnWithPort, UdpConnectionInfo>>,
     /// Recent tcp connections
     tcp: Mutex<LruCache<IpConnWithPort, TcpConnectionInfo>>,
+    /// Recent icmp connections
+    icmp: Mutex<LruCache<IcmpConn, ()>>,
     /// Whitelist of networks/peers allowed to connect
     whitelist: RwLock<Whitelist>,
 }
@@ -275,7 +278,26 @@ impl Debug for IpConnWithPort {
             .field("remote_addr", &StdIpAddr::from(self.remote_addr))
             .field("remote_port", &self.remote_port)
             .field("local_addr", &StdIpAddr::from(self.local_addr))
-            .field("remote_addr", &self.local_port)
+            .field("local_port", &self.local_port)
+            .finish()
+    }
+}
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+struct IcmpConn {
+    remote_addr: IpAddr,
+    local_addr: IpAddr,
+    response_type: u8,
+    identifier_sequence: u32,
+}
+
+impl Debug for IcmpConn {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("IcmpConn")
+            .field("remote_addr", &StdIpAddr::from(self.remote_addr))
+            .field("local_addr", &StdIpAddr::from(self.local_addr))
+            .field("response_type", &self.response_type)
+            .field("identifier_sequence", &self.identifier_sequence)
             .finish()
     }
 }
@@ -322,11 +344,12 @@ impl StatefullFirewall {
     }
 
     /// Constructs firewall with custom capacity and timeout in ms (for testing only).
-    pub fn new_custom(capacity: usize, ttl: u64) -> Self {
+    fn new_custom(capacity: usize, ttl: u64) -> Self {
         let ttl = Duration::from_millis(ttl);
         Self {
             tcp: Mutex::new(LruCache::new(ttl, capacity)),
             udp: Mutex::new(LruCache::new(ttl, capacity)),
+            icmp: Mutex::new(LruCache::new(ttl, capacity)),
             whitelist: RwLock::new(Whitelist::default()),
         }
     }
@@ -363,6 +386,12 @@ impl StatefullFirewall {
             }
             IpNextHeaderProtocols::Tcp => {
                 self.handle_outbound_tcp(&ip);
+            }
+            IpNextHeaderProtocols::Icmp => {
+                self.handle_outbound_icmp(&ip);
+            }
+            IpNextHeaderProtocols::Icmpv6 => {
+                self.handle_outbound_icmp(&ip);
             }
             _ => (),
         };
@@ -468,6 +497,16 @@ impl StatefullFirewall {
                 let TcpConnectionInfo { tx_alive, .. } = e.get_mut();
                 *tx_alive = false;
             }
+        }
+    }
+
+    fn handle_outbound_icmp<'a>(&self, ip: &impl IpPacket<'a>) {
+        let key = unwrap_option_or_return!(Self::build_icmp_key(ip, false));
+        let mut icmp_cache = unwrap_lock_or_return!(self.icmp.lock());
+        // If key already exists, dont change the value, just update timer with entry
+        if let Entry::Vacant(e) = icmp_cache.entry(key) {
+            telio_log_trace!("Inserting new ICMP conntrack entry {:?}", e.key());
+            e.insert(());
         }
     }
 
@@ -609,15 +648,85 @@ impl StatefullFirewall {
     }
 
     fn handle_inbound_icmp<'a, P: IpPacket<'a>>(&self, peer: &PublicKey, ip: &P) -> bool {
-        let icmp_packet = unwrap_option_or_return!(P::Icmp::new(ip.payload()), false);
+        let icmp_packet = unwrap_option_or_return!(IcmpPacket::new(ip.payload()), false);
 
-        if P::Icmp::BLOCKED_TYPES.contains(&icmp_packet.get_type()) {
+        if P::Icmp::BLOCKED_TYPES.contains(&icmp_packet.get_icmp_type().0) {
             telio_log_trace!("Dropping ICMP packet {:?} {:?}", ip, peer);
             return false;
         }
 
-        telio_log_trace!("Accepting ICMP packet {:?} {:?}", ip, peer);
-        true
+        let mut icmp_cache = unwrap_lock_or_return!(self.icmp.lock(), false);
+        let key = unwrap_option_or_return!(Self::build_icmp_key(ip, true), false);
+        if icmp_cache.get(&key).is_some() {
+            telio_log_trace!("Matched ICMP conntrack entry {:?}", key,);
+            telio_log_trace!("Removing ICMP conntrack entry {:?}", key);
+            icmp_cache.remove(&key);
+            telio_log_trace!("Accepting ICMP packet {:?} {:?}", ip, peer);
+            true
+        } else {
+            false
+        }
+    }
+
+    fn build_icmp_key<'a, P: IpPacket<'a>>(ip: &P, inbound: bool) -> Option<IcmpConn> {
+        let icmp_packet = unwrap_option_or_return!(IcmpPacket::new(ip.payload()), None);
+        let response_type = {
+            use IcmpTypes as v4;
+            use Icmpv6Types as v6;
+            let it = icmp_packet.get_icmp_type().0;
+            // Request messages get the icmp type of the corresponding reply message
+            // Reply messages get their own icmp type, to match that of its corresponding request
+            // We only create keys for outbound requests and inbound replies
+            if inbound {
+                if it == v4::EchoReply.0
+                    || it == v4::TimestampReply.0
+                    || it == v4::InformationReply.0
+                    || it == v4::AddressMaskReply.0
+                    || it == v6::EchoReply.0
+                {
+                    it
+                } else {
+                    return None;
+                }
+            } else if it == v4::EchoRequest.0 {
+                v4::EchoReply.0
+            } else if it == v4::Timestamp.0 {
+                v4::TimestampReply.0
+            } else if it == v4::InformationRequest.0 {
+                v4::InformationReply.0
+            } else if it == v4::AddressMaskRequest.0 {
+                v4::AddressMaskReply.0
+            } else if it == v6::EchoRequest.0 {
+                v6::EchoReply.0
+            } else {
+                return None;
+            }
+        };
+
+        let identifier_sequence = {
+            let payload = icmp_packet.payload();
+            if payload.len() < 4 {
+                return None;
+            }
+            #[allow(index_access_check)]
+            let bytes = payload[0..4].try_into();
+            if let Ok(bytes) = bytes {
+                u32::from_ne_bytes(bytes)
+            } else {
+                return None;
+            }
+        };
+        let (remote_addr, local_addr) = if inbound {
+            (ip.get_source().into(), ip.get_destination().into())
+        } else {
+            (ip.get_destination().into(), ip.get_source().into())
+        };
+        Some(IcmpConn {
+            remote_addr,
+            local_addr,
+            response_type,
+            identifier_sequence,
+        })
     }
 }
 
@@ -1007,6 +1116,20 @@ pub mod tests {
         ip.set_destination(dst.parse().expect("ICMP: Bad dst IP"));
 
         raw
+    }
+
+    fn make_icmp_with_id_and_seq<F>(id: u8, seq: u8, create_packet: F) -> Vec<u8>
+    where
+        F: Fn() -> Vec<u8>,
+    {
+        let mut packet = create_packet();
+        let len = packet.len();
+        // Packets created by `make_icmp4` and `make_icmp6` have bodies that are 14 bytes long
+        // The first two bytes of the body is the identifier and the following two bytes is the sequence number
+        // The following "magic" numbers lets us set the first (little endian) byte in the identifier and sequence numbers
+        packet[len - 14] = id;
+        packet[len - 12] = seq;
+        packet
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


jjanowsk (commented):
+1

crates/telio-firewall/src/firewall.rs
```diff
@@ -325,6 +342,7 @@ impl StatefullFirewall {
         Self {
             tcp: Mutex::new(LruCache::new(ttl, capacity)),
             udp: Mutex::new(LruCache::new(ttl, capacity)),
+            icmp: Mutex::new(LruCache::new(ttl, capacity)),
```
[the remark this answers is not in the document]
mathiaspeters: Fixed

crates/telio-firewall/src/firewall.rs
```diff
@@ -1415,7 +1498,8 @@ pub mod tests {
             TestInput { src1: "2001:4860:4860::8888",src2: "2001:4860:4860::8844", dst: "::1",       make_icmp: &make_icmp6, is_v4: false},
         ];
         for TestInput { src1, src2, dst, make_icmp, is_v4 } in test_inputs {
-            let fw = StatefullFirewall::new_custom(3, LRU_TIMEOUT);
+            // Set capacity to 0 to test only blocking of icmp type
```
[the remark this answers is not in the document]
mathiaspeters: Good suggestion! Added

jjanowsk (approved):

lcruz99:
Do we expect to accept inbound ICMP packet even when `allow_incoming_connections=False`?

mathiaspeters:
@lcruz99 @jjanowsk Can you have another look at this?

crates/telio-firewall/src/firewall.rs
```diff
@@ -610,6 +656,20 @@ impl StatefullFirewall {
...
 
+        let key = IcmpConn {
+            remote_addr: ip.get_source().into(),
+            local_addr: ip.get_destination().into(),
+            identifier: 0,
+        };
+        let mut icmp_cache = unwrap_lock_or_return!(self.icmp.lock(), false);
+        if icmp_cache.get(&key).is_some() {
+            telio_log_trace!("Matched ICMP conntrack entry {:?}", key,);
+            telio_log_trace!("Removing ICMP conntrack entry {:?}", key);
```
jjanowsk, line 667: I don't think it makes sense now. With the previous logic we allowed all incoming traffic, so basically we've been opening the firewall for all ICMP traffic. If the logic is that we should only allow one response packet, then it only  makes sense to let the packet that really is the response, not any packet. So the only candidate for me that should be placed in the cache is an echo req, and it should be removed by a matching echo response. That's it. With this implementation by sending echo req we allow the other side to send an echo req to us which seems strange.

crates/telio-firewall/src/firewall.rs
```diff
@@ -275,7 +277,24 @@ impl Debug for IpConnWithPort {
...
+            .field("local_port", &self.local_port)
+            .finish()
+    }
+}
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+struct IcmpConn {
+    remote_addr: IpAddr,
+    local_addr: IpAddr,
+    identifier: u16,
```
jjanowsk, line 289: I think identifier should be u32. ICMP identifier + seq number. Now if one ping process sends two ping requests they will have the same identifier but different seq number. In this implementation the second response will not get through, because there will be only one entry in the cache which will get removed by the first response.

crates/telio-firewall/src/firewall.rs
```diff
@@ -610,6 +656,20 @@ impl StatefullFirewall {
...
+        let mut icmp_cache = unwrap_lock_or_return!(self.icmp.lock(), false);
+        if icmp_cache.get(&key).is_some() {
+            telio_log_trace!("Matched ICMP conntrack entry {:?}", key,);
+            telio_log_trace!("Removing ICMP conntrack entry {:?}", key);
+            icmp_cache.remove(&key);
+            telio_log_trace!("Accepting ICMP packet {:?} {:?}", ip, peer);
+            return true;
+        }
+
         if P::Icmp::BLOCKED_TYPES.contains(&icmp_packet.get_type()) {
```
jjanowsk, line 673: And maybe we want to think about letting BLOCKED_TYPES through in case of whitelisted peers? Not sure about that one. Just opening a discussion

crates/telio-firewall/src/firewall.rs
```diff
@@ -610,6 +656,20 @@ impl StatefullFirewall {
     fn handle_inbound_icmp<'a, P: IpPacket<'a>>(&self, peer: &PublicKey, ip: &P) -> bool {
         let icmp_packet = unwrap_option_or_return!(P::Icmp::new(ip.payload()), false);
 
+        let key = IcmpConn {
+            remote_addr: ip.get_source().into(),
+            local_addr: ip.get_destination().into(),
+            identifier: 0,
```
jjanowsk, line 662: Also this should be filled in this function as well in case of echo response I guess

crates/telio-firewall/src/firewall.rs
jjanowsk: Never mind. I can see now that the whitelisting by peer is done earlier before this function

crates/telio-firewall/src/firewall.rs
jjanowsk: Ok more candidates for the cache: Basically all from BLOCKED_TYPES and ICMPV6_BLOCKED_TYPES, but only those.

jjanowsk: And also because now we have more packet types in the cache, not only echo, the ICMP type and maybe even code, should also be part of the cache key.

jjanowsk: And one additional note: The key in cache should probably have repose type. So the logic would go: when we have outbound echo request we put the key in cache with the same code, identifier and seq number and with echo response type. And when this response comes we remove this key. That's it. I've also  checked that all ipv4 req/resp packets that we have in BLOCKED_TYPES have identifier and seq number

crates/telio-firewall/src/firewall.rs
```diff
@@ -608,16 +644,73 @@ impl StatefullFirewall {
...
+            134 => (u8::MAX, false), // router advert message
+            135 => (136, false),     // neighbor solicit message
+            136 => (u8::MAX, false), // neighbor advert message
+            // No other types go into the cache
+            _ => return None,
+        };
+        let identifier_sequence = if has_identifier_sequence {
+            let payload = icmp_packet.payload();
+            if icmp_packet.payload().len() >= 4 {
+                u32::from_ne_bytes(payload[0..4].try_into().unwrap())
```
lcruz99, line 695: lint nagger is complaining about the `unwrap`

lcruz99 (commented):
+1.0

crates/telio-firewall/src/firewall.rs
```diff
@@ -608,16 +644,73 @@ impl StatefullFirewall {
...
+            // ICMPv6 types
+            128 => (129, true),      // echo request message
+            129 => (u8::MAX, true),  // echo reply message
+            133 => (134, false),     // router solicit message
+            134 => (u8::MAX, false), // router advert message
+            135 => (136, false),     // neighbor solicit message
+            136 => (u8::MAX, false), // neighbor advert message
+            // No other types go into the cache
+            _ => return None,
+        };
```
jjanowsk, line 691: There is no test for this, if matching replies to requests work. You should add it. But I believe it does not work. Request and replies should produce the same key in the cache. Here outgoing echo request produces different key then incoming echo response, unless I'm missing something. I think it should be sth like:
```
            0 => (0, true),     // echo reply message
            8 => (0, true),     // echo request message
            13 => (14, true),   // timestamp message
            14 => (14, true),   // timestamp reply message
            ...
```
Additionally incoming requests should be just blocked without looking into the cache.

jjanowsk: But now thinking about it, we only have requests on `ICMP_BLOCKED_TYPES` lists, so actually the logic that we wanted is already implemented on main. The requests do not go through unless the peer is whitelisted and the responses do come through. What am I missing? :D

mathiaspeters: You're right about the first comment here, I will update it and add a test.

Regarding the second comment here, the logic on main is: 
1. If whitelisted, allow
2. If on block list, reject
3. allow
and with the changes in this PR, the logic is:
1. if whitelisted, allow
2. if connection is a direct response, allow
3. if on block list, reject
4. allow

crates/telio-firewall/src/firewall.rs
mathiaspeters: I updated a function to create this key, that handles this and makes sure it's populated when it should be

crates/telio-firewall/src/firewall.rs
mathiaspeters: Fixed

crates/telio-firewall/src/firewall.rs
```diff
@@ -470,6 +493,29 @@ impl StatefullFirewall {
         }
     }
 
+    fn handle_outbound_icmp<'a>(&self, ip: &impl IpPacket<'a>) {
+        let icmp_packet = unwrap_option_or_return!(IcmpPacket::new(ip.payload()));
+        let icmp_type = icmp_packet.get_icmp_type().0;
+        // icmp type 0 is echo reply message and icmp type 8 is echo message
+        let identifier = if icmp_type == 0 || icmp_type == 8 {
+            let body = icmp_packet.payload();
+            u16::from_ne_bytes([body[0], body[1]])
```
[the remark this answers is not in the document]
mathiaspeters: Fixed

crates/telio-firewall/src/firewall.rs
mathiaspeters: All good points, and all addressed

mathiaspeters:
@lcruz99 @jjanowsk I've addressed your comments, please have another look

crates/telio-firewall/src/firewall.rs
jjanowsk: TBH this seems strange to me :) Now the only added value I see is that we do not allow ICMP responses if the request was not initiated by this peer. But we allow all other ICMP packets in anyway, so I'm not sure what's so special about responses that we want to block them, but not block other icmp packets that are not req/resp. Also now the implementation contains a bug. I've hinted this in my first msg by: "Additionally incoming requests should be just blocked without looking into the cache." Because now the remote peer can send echo request to us when we send the request to him, and that for sure should be not allowed if the peer is not whitelisted.

crates/telio-firewall/src/firewall.rs
```diff
@@ -1402,6 +1499,64 @@ pub mod tests {
...
+
+            // Add src1, src2 and src3 to the cache, and dropping src1
+            assert_eq!(fw.process_outbound_packet(&make_peer(), &make_icmp(dst, src1, IcmpTypes::EchoRequest.into())), true);
+            assert_eq!(fw.process_outbound_packet(&make_peer(), &make_icmp(dst, src2, IcmpTypes::EchoRequest.into())), true);
+            assert_eq!(fw.process_outbound_packet(&make_peer(), &make_icmp(dst, src3, IcmpTypes::EchoRequest.into())), true);
+
+            // src2 and src3 are in the cache so they pass, but src1 fails as it is not in the cache
+            assert_eq!(fw.process_inbound_packet(&make_peer(), &make_icmp(src1, dst, IcmpTypes::EchoRequest.into())), false);
+            assert_eq!(fw.process_inbound_packet(&make_peer(), &make_icmp(src2, dst, IcmpTypes::EchoRequest.into())), true);
+            assert_eq!(fw.process_inbound_packet(&make_peer(), &make_icmp(src3, dst, IcmpTypes::EchoRequest.into())), true);
```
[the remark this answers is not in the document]
mathiaspeters: Fixed

crates/telio-firewall/src/firewall.rs
mathiaspeters: This should be fixed now. As before, all outbound packets are allowed, and for inbound packets, the logic is now:
1. If whitelisted, allow
2. If on block list, reject (the block list contains the request types we put in the cache, but not the reply types, so we block `EchoRequest` for example)
3. If direct response, allow
4. reject

mathiaspeters:
@lcruz99 @jjanowsk Hopefully my last ping for review here :smile: 
@Jauler please have a look as well

jjanowsk:
So again I have a general question to this :) Do we really need it at this point. We've come to the working implementation which looks and works ok IMO, but the only difference to the `main` is that we filter out echo responses and advert messages. Is it worth doing? Is there any harm in letting in an ICMP echo response? I'll leave some nitpick comments anyway in case we go for it :)

crates/telio-firewall/src/firewall.rs
```diff
@@ -608,15 +644,71 @@ impl StatefullFirewall {
...
+        }
+    }
+
+    fn build_icmp_key<'a, P: IpPacket<'a>>(ip: &P, inbound: bool) -> Option<IcmpConn> {
+        let icmp_packet = unwrap_option_or_return!(IcmpPacket::new(ip.payload()), None);
+        let (response_type, has_identifier_sequence) = match icmp_packet.get_icmp_type().0 {
+            // Request messages get the icmp type of the corresponding reply message
+            // Reply messages get their own icmp type, to match that of its corresponding request
+            // ICMP types
+            0 => (0, true),   // echo reply message
```
jjanowsk, line 673: Could those be replaced by constants from `IcmpTypes` instead of relying on comments?

crates/telio-firewall/src/firewall.rs
```diff
@@ -608,15 +644,71 @@ impl StatefullFirewall {
...
+            129 => (129, true),  // echo reply message
+            133 => (134, false), // router solicit message
+            134 => (134, false), // router advert message
+            135 => (136, false), // neighbor solicit message
+            136 => (136, false), // neighbor advert message
+            // No other types go into the cache
+            _ => return None,
+        };
+        let identifier_sequence = if has_identifier_sequence {
+            let bytes = icmp_packet.payload()[0..4].try_into();
```
jjanowsk, line 692: I think you've again forgot about the check for the length of the payload?

mathiaspeters:
> So again I have a general question to this :) Do we really need it at this point. We've come to the working implementation which looks and works ok IMO, but the only difference to the `main` is that we filter out echo responses and advert messages. Is it worth doing? Is there any harm in letting in an ICMP echo response? I'll leave some nitpick comments anyway in case we go for it :)

With the current solution we're filtering out anything that isn't a direct response to the message types `BLOCKED_TYPES` (4 types for v4 and 4 types for v6), so there is a lot more filtering now than there was before. You can see if the function `build_icmp_key` which ICMP types we create cache keys for, any other type we reject. On top of that, since checking `BLOCKED_TYPES` now happens before checking the cache, `handle_inbound_icmp` only allows the few reply types that get cache keys.

crates/telio-firewall/src/firewall.rs
mathiaspeters: `try_into` returns a `Result` with an `Err` if the body is not long enough, which then gets rejected by the if-else below this line

jjanowsk: It does not protect you from the panic. `try_into` in this example converts from slice to array. It converts to 4 element array because that's what expected by `u32::from_ne_bytes` and it will fail if you give it a slice of the wrong size. E.g. `icmp_packet.payload()[0..2].try_into()` will return None because the slice is too short. But it does nothing to check that your slice itself is correct, and in that case it might not be correct. So if your slice points to unallocated memory this will panic.

jjanowsk: Here's the playground to verify: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=e8bcc5f5b43d0e32c7f62afb6d8522c6

mathiaspeters: I didn't consider the slicing operation, only the conversion. You are of course correct, and I will add a check

jjanowsk:
> > So again I have a general question to this :) Do we really need it at this point. We've come to the working implementation which looks and works ok IMO, but the only difference to the `main` is that we filter out echo responses and advert messages. Is it worth doing? Is there any harm in letting in an ICMP echo response? I'll leave some nitpick comments anyway in case we go for it :)
> 
> With the current solution we're filtering out anything that isn't a direct response to the message types `BLOCKED_TYPES` (4 types for v4 and 4 types for v6), so there is a lot more filtering now than there was before. You can see if the function `build_icmp_key` which ICMP types we create cache keys for, any other type we reject. On top of that, since checking `BLOCKED_TYPES` now happens before checking the cache, `handle_inbound_icmp` only allows the few reply types that get cache keys.

You're right, I've missed that. Previous logic was that if the function creating the key returned None the packet was let through.

Jauler:
> > > So again I have a general question to this :) Do we really need it at this point. We've come to the working implementation which looks and works ok IMO, but the only difference to the `main` is that we filter out echo responses and advert messages. Is it worth doing? Is there any harm in letting in an ICMP echo response? I'll leave some nitpick comments anyway in case we go for it :)
> > 
> > 
> > With the current solution we're filtering out anything that isn't a direct response to the message types `BLOCKED_TYPES` (4 types for v4 and 4 types for v6), so there is a lot more filtering now than there was before. You can see if the function `build_icmp_key` which ICMP types we create cache keys for, any other type we reject. On top of that, since checking `BLOCKED_TYPES` now happens before checking the cache, `handle_inbound_icmp` only allows the few reply types that get cache keys.
> 
> You're right, I've missed that. The logic of the previous version was that if the function creating the key returned None the packet was let through. But now I'm worried that by blocking all icmp traffic we can do more harm then good. E.g. Path MTU Discovery procedure can be broken because of blocking "Fragmentation required" message. There are probably more examples like this. So in my opinion blocking just blacklisted packet types like we do on main should be enough. But I would like some additional pair of eyes to review this and participate in this discussion. Sorry for dragging this review ;)

That is a good point. IMHO, the "best" way to filter ICMP error packets, would be according to the IP header stored in the payload of the message (e.g. Normally it starts 8 bytes into the ICMP message, but please double-check this on every message). At first I personally I somehow thought that properly supporting ICMP messages is somewhat low on the priority list. But now that you mention MTU discovery, I am kinda thinking, that it would make sense to be on the safe side and add support for them.

@jjanowsk, would you agree with the approach of taking the IP header of the original message from ICMP error packet and matching it against either TCP or UDP or ICMP [caches](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-firewall/src/firewall.rs#L218)?

crates/telio-firewall/src/firewall.rs
```diff
@@ -364,6 +387,9 @@ impl StatefullFirewall {
             IpNextHeaderProtocols::Tcp => {
                 self.handle_outbound_tcp(&ip);
             }
+            IpNextHeaderProtocols::Icmp => {
```
Jauler, line 390: If I remember correctly, Next Header protocol for IPv6 ICMP is different from IPv4, so we should have a case for `IpNextHeaderProtocols::Icmpv6` (refernece: https://docs.rs/pnet/latest/pnet/packet/ip/IpNextHeaderProtocols/constant.Icmpv6.html)

crates/telio-firewall/src/firewall.rs
```diff
@@ -608,15 +644,78 @@ impl StatefullFirewall {
...
+    }
+
+    fn build_icmp_key<'a, P: IpPacket<'a>>(ip: &P, inbound: bool) -> Option<IcmpConn> {
+        let icmp_packet = unwrap_option_or_return!(IcmpPacket::new(ip.payload()), None);
+        let (response_type, has_identifier_sequence) = {
+            use IcmpTypes as v4;
+            use Icmpv6Types as v6;
+            let it = icmp_packet.get_icmp_type().0;
+            // Request messages get the icmp type of the corresponding reply message
+            // Reply messages get their own icmp type, to match that of its corresponding request
```
Jauler, line 674: :thinking: In this case I believe we do not expect replies when an outbound "reply" packet is sent, therefore we probably just need to populate cache only if outbound request packet passes.

Or are there any cases which I am missing?

crates/telio-firewall/src/firewall.rs
mathiaspeters: In Matas' PR he handles this case, but it's essentially just generics and calling the same function

crates/telio-firewall/src/firewall.rs
mathiaspeters: Would we ever send an outbound reply? I can add a special case for that if you think it's relevant

mathiaspeters:
@jjanowsk There is a follow-up ticket to add handling of ICMP error types that I will start with as soon as this is okay to merge. So disregarding that, is there anything left here that is a blocker or needs fixing in your opinion?

crates/telio-firewall/src/firewall.rs
mathiaspeters: Good point. I was thinking that we don't have outbound replies but of course we do. I will add special handling of that

jjanowsk:
> @jjanowsk There is a follow-up ticket to add handling of ICMP error types that I will start with as soon as this is okay to merge. So disregarding that, is there anything left here that is a blocker or needs fixing in your opinion?

Yes, my question stays: What do we gain by complicating ICMP rules so much? Now if we accept this we need to fix error messages by complicating it even further. What is so wrong with the current approach we have on main where we just filter out requests but let all other ICMP traffic through. Maybe it would be better to first think if there are any other ICMP codes that we want to filter out?

mathiaspeters:
> > @jjanowsk There is a follow-up ticket to add handling of ICMP error types that I will start with as soon as this is okay to merge. So disregarding that, is there anything left here that is a blocker or needs fixing in your opinion?
> 
> Yes, my question stays: What do we gain by complicating ICMP rules so much? Now if we accept this we need to fix error messages by complicating it even further. What is so wrong with the current approach we have on main where we just filter out requests but let all other ICMP traffic through. Maybe it would be better to first think if there are any other ICMP codes that we want to filter out?

@Jauler I'll leave this question to you

Jauler:
> > @jjanowsk There is a follow-up ticket to add handling of ICMP error types that I will start with as soon as this is okay to merge. So disregarding that, is there anything left here that is a blocker or needs fixing in your opinion?
> 
> Yes, my question stays: What do we gain by complicating ICMP rules so much? Now if we accept this we need to fix error messages by complicating it even further. What is so wrong with the current approach we have on main where we just filter out requests but let all other ICMP traffic through. Maybe it would be better to first think if there are any other ICMP codes that we want to filter out?

Generally the issue is that with stateless approach we do let "unexpected" replies (and other type packets) through firewall e.g. imagine a situation where node A does not allow incoming connections from node B, all ICMP packets except for request ones will be let through the firewall in this situation. And this "incoming connection" permission does give the expectation that only packets belonging to the communication initiated by the node A will be let through firewall.

While I would not consider this issue to be high priority, with this PR - we already have all the required state tracking implemented, and ICMP error handling does not require additional state tracking, only additional allow condition for inbound packets.

mathiaspeters:
@jjanowsk Thoughts about :point_up:?

jjanowsk:
Sorry, I've missed the reply. As I wrote. I don't have strong opinion on that, I just wanted to start a discussion: is it really worth adding the complexity. If you think it is worth it then I'm fine with that. I'll quicky glance over the changes again now and approve it if I won't find anything

jjanowsk:
Ohh, actually I see from comments that you still want to do some changes, so I'll wait with final review for those. Ping me when it's done

mathiaspeters:
@jjanowsk Please have another look

crates/telio-firewall/src/firewall.rs
```diff
@@ -608,15 +644,79 @@ impl StatefullFirewall {
...
+            // Outbound reply messages don't get a key since we're not expecting a response
+            if it == v4::EchoRequest.0 || (it == v4::EchoReply.0 && inbound) {
```
jjanowsk, line 676: I think you have this logic backwards. You don't really care about replies. You care about not letting in requests. So I would imagine having

crates/telio-firewall/src/firewall.rs
```diff
@@ -1401,7 +1501,195 @@ pub mod tests {
...
+            dst: &'static str,
+            make_icmp: MakeIcmp,
+        }
+
+        let test_inputs = vec![
+            TestInput { src1: "8.8.8.8",              src2: "8.8.4.4",              src3: "4.4.4.4",              dst: "127.0.0.1", make_icmp: &make_icmp4 },
+            TestInput { src1: "2001:4860:4860::8888", src2: "2001:4860:4860::8844", src3: "2001:4860:4860::4444", dst: "::1",       make_icmp: &make_icmp6 },
+        ];
+        for TestInput{ src1, src2, src3, dst, make_icmp } in test_inputs {
+            let make_packet = |src, dst, icmp_type, id, seq| -> Vec<u8> {
```
jjanowsk, line 1518: this helper is copied 3 times. I think it can be extracted to the separate function

crates/telio-firewall/src/firewall.rs
```diff
@@ -1401,7 +1501,195 @@ pub mod tests {
...
+            }
+        }
+
+        for (i, request) in V6_REQUEST_TYPES.iter().enumerate() {
+            for (j, reply) in V6_REPLY_TYPES.iter().enumerate() {
+                let fw = StatefullFirewall::new();
+                let outbound = make_packet(&make_icmp6, V6_SRC_IP, V6_DST_IP, (*request).into(), 1, 1);
+                let inbound = make_packet(&make_icmp6, V6_DST_IP, V6_SRC_IP, (*reply).into(), 1, 1);
+                assert_eq!(fw.process_outbound_packet(&make_peer(), &outbound), true);
+                assert_eq!(fw.process_inbound_packet(&make_peer(), &inbound), i == 0 && j == 0);
```
jjanowsk, line 1661: Why `i == 0 && j == 0`? Advert should also be a valid response for solicit

jjanowsk: I think it should be `i == j` as for v4 messages. If it's not then there is a bug somewhere. It might be that you overwrite some bytes in `make_packet` function for those messages even though they do not have id and seq fields.

mathiaspeters: The complicated thing here, which I haven't really thought about how to approach is that `BLOCKED_TYPES` for IPv6 contains `RouterAdvert`, `NeighborSolicit` and `NeighborAdvert`. Since those types are in `BLOCKED_TYPES` I added them to the cache handling like the other types, but they are essentially unusable given that they're in `BLOCKED_TYPES` so it would probably make more sense to remove them from the cache handling completely. That would reject any inbound ICMP packet with those type, but then again, that is already happening

mathiaspeters: I removed handling of those ICMP types from the cache since they will be blocked anyways, and I simplified this test

jjanowsk (dismissed):
+1

jjanowsk (approved):

PR #66, opened 2023-08-10
LLT-4100: natlab fix exception handling during failure

gytsto:
### Description
Problem:  
- During failed tests, most of the exception are suppressed by `asyncio_util::run_async`. When exception occurs and it is not `asyncio.CancelledError`, it goes to general exception handler. There it calls `sys.exit(1)`, which mostly does nothing when exception is raised and basically works as `pass`. This behavior suppresses exceptions and doesn't allow program to handle them, which might lead to improper cleanup, causing instability in next running tests.

Solution:  
- Remove `sys.exit(1)` from `asyncio_util` and let exceptions through.
- Also ignore exit code from processes when they are cancelled.

**Note:** Log output was not affected like the comments said.

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_network_switch.py
+++ b/nat-lab/tests/test_network_switch.py
@@ -220,7 +220,12 @@ async def test_vpn_network_switch(
             AdapterType.LinuxNativeWg,
             marks=[
                 pytest.mark.linux_native,
-                pytest.mark.xfail(reason="the test is flaky - JIRA issue: LLT-4105"),
+                pytest.mark.xfail(
+                    reason=(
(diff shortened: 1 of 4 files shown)


mathiaspeters:
Code looks good to me. Do you have a pipeline that shows what an error looks like after these changes?

gytsto:
> Code looks good to me. Do you have a pipeline that shows what an error looks like after these changes?

gitlab pipeline id: 1481514, check under natlab-tests, there is one that is cancelled and on failed, there you will see

lcruz99 (approved):
+1

PR #67, opened 2023-08-10
Add temporary cron schedule for testing

tomaszklak:
This adds new cron schedule to start a new pipeline every 60 minutes. It's only temporary to make it possible to test some changes outside of normal nightly runs and reduce feedback loop while I develop it.

- [x] Commit history is clean
- [x] README.md is updated
- [x] changelog.md is updated


jjanowsk (approved):
+2

PR #73, opened 2023-08-14
Add support for IPv6 packet handling in telio-dns

dfetti:
### Problem
We need support for IPv6 DNS requests

### Solution
Handle IPv6 packets in a similar way as we handle IPv4 packets

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ alpha-events.db
 beta-events.db
 gamma-events.db
 nat-lab/report.html
+events-moose.log
--- a/crates/telio-dns/src/dns.rs
+++ b/crates/telio-dns/src/dns.rs
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 use boringtun::crypto::x25519::{X25519PublicKey, X25519SecretKey};
 use boringtun::noise::Tunn;
 use ipnetwork::IpNetwork;
-use std::net::{IpAddr, Ipv4Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::{net::SocketAddr, sync::Arc};
 use telio_crypto::{PublicKey, SecretKey};
 use telio_wg::uapi::Peer;
@@ -161,17 +161,25 @@ impl DnsResolver for LocalDnsResolver {
         vec![
             IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into(),
             IpAddr::V4(Ipv4Addr::new(100, 64, 0, 3)).into(),
+            IpAddr::V6(Ipv6Addr::new(0xfc74, 0x656c, 0x696f, 0, 0, 0, 0, 2)).into(),
+            IpAddr::V6(Ipv6Addr::new(0xfc74, 0x656c, 0x696f, 0, 0, 0, 0, 3)).into(),
         ]
     }
 
     #[allow(unwrap_check)]
     fn get_exit_connected_dns_allowed_ips(&self) -> Vec<IpNetwork> {
-        vec![IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into()]
+        vec![
+            IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into(),
+            IpAddr::V6(Ipv6Addr::new(0xfc74, 0x656c, 0x696f, 0, 0, 0, 0, 2)).into(),
+        ]
     }
 
     #[allow(unwrap_check)]
     fn get_default_dns_servers(&self) -> Vec<IpAddr> {
-        vec![IpAddr::V4(Ipv4Addr::new(100, 64, 0, 3))]
+        vec![
+            IpAddr::V4(Ipv4Addr::new(100, 64, 0, 3)),
+            IpAddr::V6(Ipv6Addr::new(0xfc74, 0x656c, 0x696f, 0, 0, 0, 0, 3)),
+        ]
     }
 }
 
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -65,3 +65,4 @@ alpha-events.db
 beta-events.db
 gamma-events.db
 nat-lab/report.html
+events-moose.log
```
mathiaspeters, line 68: :tada:

crates/telio-dns/tests/nameserver.rs
```diff
@@ -217,17 +245,85 @@ impl WGClient {
...
+    }
+
     fn client_address(&self) -> SocketAddr {
         self.client_address
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 enum DnsTestType {
```
mathiaspeters, line 300: If we're adding a `Ipv6` suffix to the IPv6 lines then I think we should add a `Ipv4` suffix to the other lines

crates/telio-dns/tests/nameserver.rs
```diff
@@ -311,6 +407,25 @@ async fn dns_request_local() {
     .expect("Test timeout");
 }
 
+#[tokio::test]
```
mathiaspeters, line 410: It feels like there only negative tests but no positive tests. Is there a reason for that?

crates/telio-dns/src/nameserver.rs
```diff
@@ -64,6 +66,309 @@ impl LocalNameServer {
...
+                    Ok(false)
+                } else {
+                    Err(String::from("Invalid IP version"))
+                }
+            })?;
+
+        let resolver = Resolver::new();
+        let (dns_response, udp_destination, udp_source) = {
+            let payload;
+            let (udp_request, dns_source) = if is_ipv4 {
```
tomaszklak, line 207: This is no what I had in mind. Technically there is now one function but it doesn't solve the code duplication issue. I was hoping for something like [process_inbound_ip_packet](https://github.com/NordSecurity/libtelio/blob/93780b99b588d6cc66b1c1d25e89cb1ae3114905/crates/telio-firewall/src/firewall.rs#L374). This way you will have common code in only one place. This doesn't have to be done at compile time since I doubt dns responses need to be maximally fast but however you chose to do it please make sure that the common parts are not duplicated.

crates/telio-dns/src/nameserver.rs
```diff
@@ -64,6 +66,423 @@ impl LocalNameServer {
...
+            .ok_or_else(|| String::from("Inexistent DNS request"))?;
+        let dns_request = Request::new(dns_request, request_info.dns_source(), Protocol::Udp);
+        zones.lookup(&dns_request, resolver.clone()).await;
+
+        telio_log_debug!("DNS request: {:?}", &dns_request);
+
+        let dns_response = resolver.0.lock().await;
+        telio_log_debug!("Nameserver response: {:?}", &dns_response);
+
+        Ok(dns_response.clone())
```
mathiaspeters, line 206: This clone doesn't seem like it should be needed

mathiaspeters:
+1

crates/telio-dns/src/dns.rs
```diff
@@ -161,16 +161,21 @@ impl DnsResolver for LocalDnsResolver {
         vec![
             IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into(),
             IpAddr::V4(Ipv4Addr::new(100, 64, 0, 3)).into(),
+            // TODO: [LLT-4127] Add IPv6 when we decide what we will use for DNS
```
Jauler, line 164: You can add `fc00:0074:656c:696f/64` range here ;).

crates/telio-dns/src/dns.rs
```diff
@@ -161,16 +161,21 @@ impl DnsResolver for LocalDnsResolver {
...
         ]
     }
 
     #[allow(unwrap_check)]
     fn get_exit_connected_dns_allowed_ips(&self) -> Vec<IpNetwork> {
-        vec![IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into()]
+        vec![
+            IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into(),
+            // TODO: [LLT-4127] Add IPv6 when we decide what we will use for DNS
```
Jauler, line 172: You can add `fc00:0074:656c:696f/64` range here ;).

crates/telio-dns/tests/nameserver.rs
```diff
@@ -152,14 +154,29 @@ impl WGClient {
...
+
+                let ip_response = Ipv6Packet::new(response).expect("Failed to parse ip response");
+                let udp_response =
+                    UdpPacket::new(ip_response.payload()).expect("Failed to parse udp response");
+                dns_parser::Packet::parse(udp_response.payload())
+                    .expect("Failed to parse dns response");
+            }
             TunnResult::Err(e) => panic!("Decapsulate error: {:?}", e),
-            _ => println!("Unexpected TunnResult while receiving the dns response"),
+            _ => panic!("Unexpected TunnResult while receiving the dns response"),
```
Jauler, line 179: This panic seems dangerous, can we handle this in less dangerous way, e.g. `telio_log_err()` or similar. 

Also `TunnResult::Err(e)` case is probably dangerous in a similar manner. Can we handle it in some less destructive way? :thinking:

Jauler: Oh wait, I misread this and missed the fact that this is test code.

mathiaspeters: It's in a test so should not be a problem

Jauler (approved):
+1.0

PR #75, opened 2023-08-16
add tests for ssh/docker process implementattion

gytsto:
### Problem

### Solution


lcruz99 (approved):
+1

PR #76, opened 2023-08-16
Fixing issues in telio-firewall

jjanowsk:
1. Bugfix for LLT-4119 - consolidate_firewall function was removing exit_node from firewall. This commit fixes it.
2. consolidate_firewall now uses whitlists from firewall itself as a current state insted of relying on wg interface
3. connect_exit_node was fixed. Previous implementation contained a bug when it was called twice with the same vpn server as an exit_node. In that case the function assumed it is not a vpn server because it already had such peer in wg interface. Fixed it by differentiating between vpn server and mesh peer by just looking at the endpoint field.
4. consolidate_firewall was not removing peers from firewall's whitelists when one of their permissions was stripped. E.g. if the peer was previously allowed for incomming connections and to send files and later it was stripped from one of those the firewall rules were not updated correctly. Now fixed.
5. Adding exit_node to firewall was moved from connect_exit_node to consolidate_firewall to keep the logic in one place.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -45,7 +45,6 @@ use wg::uapi::{self, PeerState};
 
 use std::{
     collections::HashSet,
-    convert::TryInto,
     future::Future,
     io::{Error as IoError, ErrorKind},
     net::{IpAddr, Ipv4Addr, SocketAddr},
@@ -94,8 +93,8 @@ pub enum Error {
     BadPublicKey,
     #[error("Invalid node configuration")]
     InvalidNode,
-    #[error("Dublicate exit node")]
-    DuplicateExitNode,
+    #[error("Configured exit node is not a meshnet node and does not have an endpoint set")]
+    EndpointNotProvided,
     #[error("Deleting non-existant node")]
     InvalidDelete,
     #[error("Meshnet IP is not set for the node")]
@@ -1316,38 +1315,30 @@ impl Runtime {
     }
 
     async fn connect_exit_node(&mut self, exit_node: &ExitNode) -> Result {
-        let mut node: Node = exit_node.try_into()?;
+        let exit_node = exit_node.clone();
 
         // dns socket for macos should only be bound to tunnel interface when connected to exit,
         // otherwise with no exit dns peer will try to forward packets through tunnel and fail
         bind_tun::set_should_bind(true);
 
-        let wg_peer = self
-            .entities
-            .wireguard_interface
-            .get_interface()
-            .await?
-            .peers
-            .get(&node.public_key)
-            .cloned();
+        let is_meshnet_exit_node = self
+            .requested_state
+            .meshnet_config
+            .as_ref()
+            .and_then(|config| config.peers.as_deref())
+            .map(|peers| peers.iter().any(|p| p.public_key == exit_node.public_key))
+            .unwrap_or_default();
 
-        if let Some(wg_peer) = wg_peer {
-            node.is_vpn = false;
-            node.endpoint = wg_peer.endpoint;
-
-            //  forward dns traffic to exit peer's dns resolver
+        if is_meshnet_exit_node {
             if let Some(dns) = &self.entities.dns.lock().await.resolver {
                 self.reconfigure_dns_peer(dns, &dns.get_default_dns_servers())
                     .await?;
             }
-        } else {
-            node.is_vpn = true;
-            self.entities
-                .firewall
-                .add_to_peer_whitelist(node.public_key);
+        } else if exit_node.endpoint.is_none() {
+            return Err(Error::EndpointNotProvided);
         }
 
-        self.requested_state.exit_node = Some(exit_node.clone());
+        self.requested_state.exit_node = Some(exit_node);
         wg_controller::consolidate_wg_state(&self.requested_state, &self.entities).await
     }
 
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -10,7 +10,7 @@ use telio_crypto::PublicKey;
 use telio_dns::DnsResolver;
 use telio_firewall::firewall::{Firewall, FILE_SEND_PORT};
 use telio_model::EndpointMap;
-use telio_model::{mesh::Node, SocketAddr};
+use telio_model::SocketAddr;
 use telio_proxy::Proxy;
 use telio_traversal::{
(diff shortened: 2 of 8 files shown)


src/device/mod.rs
```diff
@@ -1316,38 +1315,22 @@ impl Runtime {
...
-            .peers
-            .get(&node.public_key)
-            .cloned();
-
-        if let Some(wg_peer) = wg_peer {
-            node.is_vpn = false;
-            node.endpoint = wg_peer.endpoint;
-
+        // if exit_node is not a vpn server
+        if exit_node.endpoint.is_none() {
```
Jauler, line 1325: I know I mentioned that the the way to decide whether `exit_node` is meshnet peer and we are traffic routing or whether `exit_node` is VPN server is by endpoint being either `None` or `Some(...)` respectively.

But there is alternative way, by checking whether a node with the same public key exists amongst meshnet peers, e.g. node such that `requested_state.meshnet_config.peers.base.public_key == exit_node.public_key`.

We already use similar check in our WG consolidation here: https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L311 

I am thinking that having single thing being checked would be safer.

src/device/wg_controller.rs
```diff
@@ -250,27 +246,52 @@ fn check_allowed_ips_correctness(peers: &BTreeMap<PublicKey, RequestedPeer>) ->
...
         .iter()
         .flat_map(|c| c.peers.iter())
         .flatten()
+        .map(|p| (p.public_key, p.into()))
         .collect();
-    let to_keys: HashSet<PublicKey> = to_peers.iter().map(|p| p.public_key).collect();
-    let delete_keys = &from_keys - &to_keys;
+    if let Some(exit_node) = &requested_state.exit_node {
+        if let Some(peer) = to_peers.get_mut(&exit_node.public_key) {
+            peer.is_exit = true;
```
Jauler, line 266: Just to double-check. This assignment will not affect the "real" `requested_state.meshnet_config` as `to_peers` is a copy of meshnet peer list?

src/device/wg_controller.rs
```diff
@@ -250,27 +246,52 @@ fn check_allowed_ips_correctness(peers: &BTreeMap<PublicKey, RequestedPeer>) ->
...
+            firewall.add_to_peer_whitelist(node.public_key);
+        } else {
+            firewall.remove_from_peer_whitelist(node.public_key);
+        }
+
+        if node.allow_peer_send_files {
+            firewall.add_to_port_whitelist(node.public_key, FILE_SEND_PORT);
+        } else {
+            firewall.remove_from_port_whitelist(node.public_key);
+        }
```
Jauler, line 294: @jjanowsk What are you thoughts about this:

Instead of building `to_peers` which is a list of Nodes, we build two sets of peers which need to be peer-whitelisted or port-whitelisted directly. 

It could look somewhat similar to this:
```rust
    let from_keys_peer_whitelist: HashSet<PublicKey> =
        firewall.get_peer_whitelist().iter().copied().collect();
    let from_keys_ports_whitelist: HashSet<PublicKey> =
        firewall.get_port_whitelist().keys().copied().collect();

    // Build a list of peers expected to be peer-whitelisted according
    // to allow_incoming_connections permission
    let to_keys_peer_whitelist: HashSet<PublicKey> = requested_state
        .meshnet_config
        .iter()
        .flat_map(|c| c.peers.iter())
        .flatten()
        .filter(|p| p.allow_incoming_connections)
        .map(|p| p.public_key)
        .collect();

    // VPN peer must always be peer-whitelisted
    if let Some(exit_node) = &requested_state.exit_node {
        //TODO
    }


    // Build a list of peers expected to be port-whitelisted according
    // to allow_peer_send_files permission
    let to_keys_ports_whitelist: HashSet<PublicKey> = requested_state
        .meshnet_config
        .iter()
        .flat_map(|c| c.peers.iter())
        .flatten()
        .filter(|p| p.allow_peer_send_files)
        .map(|p| p.public_key)
        .collect();

    // Consolidate peer-whitelist
    let delete_keys = &from_keys_peer_whitelist - &to_keys_peer_whitelist;
    for key in delete_keys {
        firewall.remove_from_peer_whitelist(key);
    }
    let add_keys = &to_keys_peer_whitelist - &from_keys_peer_whitelist;
    for key in add_keys {
        firewall.add_to_peer_whitelist(key);
    }

    // Consolidate port-whitelist
    let delete_keys = &from_keys_ports_whitelist - &to_keys_ports_whitelist;
    for key in delete_keys {
        firewall.remove_from_port_whitelist(key);
    }
    let add_keys = &to_keys_ports_whitelist - &from_keys_ports_whitelist;
    for key in add_keys {
        firewall.add_to_port_whitelist(key, FILE_SEND_PORT);
    }
```

jjanowsk: I agree. It looks better. I'll apply this

src/device/mod.rs
packgron: Hmm it would be also nice to have a check that node that is not in meshnet nodes (assumed to be vpn node).
Has endpoint. nobody curretly forbit's from API point of view, to try to set exit_node, before setting meshnet config.

jjanowsk: So how about:
```    
    async fn connect_exit_node(&mut self, exit_node: &ExitNode) -> Result {
        let exit_node = exit_node.clone();

        // dns socket for macos should only be bound to tunnel interface when connected to exit,
        // otherwise with no exit dns peer will try to forward packets through tunnel and fail
        bind_tun::set_should_bind(true);

        let mut is_meshnet_exit_node = false;

        if let Some(config) = &self.requested_state.meshnet_config {
            if let Some(peers) = &config.peers {
                is_meshnet_exit_node = peers
                    .iter()
                    .any(|peer| peer.base.public_key == exit_node.public_key)
            }
        }

        if is_meshnet_exit_node {
            if let Some(dns) = &self.entities.dns.lock().await.resolver {
                    .await?;
            }
        } else {
            if exit_node.endpoint.is_none() {
                return Err(Error::EndpointNotProvided);
            }
        }

        self.requested_state.exit_node = Some(exit_node);
        wg_controller::consolidate_wg_state(&self.requested_state, &self.entities).await
    }
```

Jauler: This looks nice to me!

Jauler (approved):
+1.5

packgron (dismissed):
+1.

gytsto (approved):

PR #79, opened 2023-08-18
Resolve LLT-4149: Fix process test flakyness and add test for general exception

gytsto:
### Problem
Processes were killed before actually raising `ProcessExecError` when failed

### Solution
Move process existence checking before moving out scope, prolonging killing of the porcess.


Jauler (approved):
+2.0

PR #81, opened 2023-08-21
Resolve LLT-4053: improve testing framework

gytsto:
### Problem
- general improvements need

### Solution
- Introduce setups functions
- Parameterize tests with setup configurations
- Introduce timeout args in some `telio::Client` functions
- Start waiting for events before actual action (on some race condition places)

### TODO
- Move connection tracker to function decorator somehow (not sure if needed)
- Introduce basic parameter sets / decorators to remove redundant parametrization code (reduces enormous parametrizations)

- [x] README.md is updated
- [x] changelog.md is updated


nat-lab/tests/test_connection_states.py
```diff
@@ -37,52 +36,45 @@
...
-        api = API()
-
-        (alpha, beta) = api.default_config_two_nodes()
-
-        (connection_alpha, alpha_conn_tracker) = await exit_stack.enter_async_context(
-            new_connection_with_conn_tracker(
-                alpha_connection_tag,
-                generate_connection_tracker_config(
-                    alpha_connection_tag, derp_1_limits=ConnectionLimits(1, 1)
+        env = await setup_environment(
```
Jauler, line 49: In my personal opinion:

I like the `setup_environment()` idea, as, at least for me, it is more clear. But in my opinion the biggest benefit could be gained from having common sets of parameters, like:
```python
env = await setup_meshnet_three_connected_nodes(...);
```

This function may accept some parameters (like `connection_tag` or similar), but if those parameters are passed, it would be kinda useful if they would only come from test parameterization. This would allow to nicely decouple environment from test scenario. 

And one more thing, there would be very nice, if this helper function would also include waiting for DERP connection as well as (if needed) meshnet nodes to be reported as connected.

This way this test would look like:

```python
env = await setup_meshnet_two_connected_nodes(...);
# <<< At this moment we not we have meshnet nodes and that they are connected

# Start routing
await testing.wait_long(env.nodes[0].get_router().create_exit_node_route())
await testing.wait_long(env.nodes[1].connect_to_exit_node(beta.public_key))

# Test that event with "status=connected" has been raised:
await testing.wait_long(env.nodes[1].wait_for_event_peer(beta.public_key, [telio.State.Connected]))

# Stop routing
await testing.wait_long(client_alpha.disconnect_from_exit_nodes())

# Test that event with "status=connected" has been raised again
 await testing.wait_long(client_alpha.wait_for_event_peer(beta.public_key, [telio.State.Connected]))

# Perform ping:
async with Ping(connection_alpha, beta.ip_addresses[0]).run() as ping:

# Assert connection tracker limits
assert alpha_conn_tracker and alpha_conn_tracker.get_out_of_limits() is None
assert beta_conn_tracker and beta_conn_tracker.get_out_of_limits() is None
```

If it would be possible to somehow tie connection tracker limits to the destruction of `env` it would be even nicer, as tests would not need to repeat it everytime.

``` 

*

Comment (Jauler) [reply nat-lab/tests/test_connection_states.py:49]: Also `await testing.wait_long(...)` is repeated a lot of times, maybe it would be possible to somehow add timeout value for all the methods (e.g. `create_exit_node_route()`, `connect_to_exit_node()` as an function parameter with default value. 

such that the test would become:
```python
# Acquire meshnet with two connected nodes
env = await setup_meshnet_two_connected_nodes(...);

# Start routing
await env.nodes[0].get_router().create_exit_node_route()
await env.nodes[1].connect_to_exit_node(beta.public_key)

# Test that event with "status=connected" has been raised:
await env.nodes[1].wait_for_event_peer(beta.public_key, [telio.State.Connected])

# Stop routing
await client_alpha.disconnect_from_exit_nodes()

# Test that event with "status=connected" has been raised again
await client_alpha.wait_for_event_peer(beta.public_key, [telio.State.Connected])

# Perform ping:
async with Ping(connection_alpha, beta.ip_addresses[0]).run() as ping:
    await ping.wait_for_next_ping()
```

nat-lab/tests/test_tcli.py
```diff
@@ -1,95 +0,0 @@
-import asyncio
```
arg0d, line 1: Is there another simple test that only checks DERP connection? I found this test very useful to check if basic functionality is working.

gytsto: any test from `test_derp_connect.py` and also `test_telio_tasks_with_all_features` will be doing the same thing, when I implement usage of `setup_mesh_nodes`

Jauler (dismissed):
+1.0

matislovas:
+1 but needs a rebase IMO

matislovas (approved):

PR #85, opened 2023-08-22
Natlab police: address nightly flakyness

lcruz99:
### Problem
During nightly runs it was found a few flaky tests.

### Solution

For the ones that the flakyness cause was simple enough, a fix is also provided in this PR.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_events.py
+++ b/nat-lab/tests/test_events.py
@@ -301,6 +301,7 @@ async def test_event_content_vpn_connection(
 
 @pytest.mark.asyncio
 @pytest.mark.derp
+@pytest.mark.xfail(reason="test is flaky - Jira issue: LLT-4162")
 @pytest.mark.parametrize(
     "alpha_connection_tag,adapter_type",
     [
--- a/nat-lab/tests/test_events.py
+++ b/nat-lab/tests/test_events.py
@@ -301,7 +301,6 @@ async def test_event_content_vpn_connection(
 
 @pytest.mark.asyncio
 @pytest.mark.derp
-@pytest.mark.xfail(reason="test is flaky - Jira issue: LLT-4162")
 @pytest.mark.parametrize(
     "alpha_connection_tag,adapter_type",
     [
@@ -398,10 +397,11 @@ async def test_event_content_exit_through_peer(
 
         await testing.wait_long(client_beta.get_router().create_exit_node_route())
 
-        await testing.wait_long(client_alpha.connect_to_exit_node(beta.public_key))
-
         await testing.wait_long(
-            client_alpha.wait_for_event_peer(beta.public_key, [State.Connected])
+            asyncio.gather(
+                client_alpha.connect_to_exit_node(beta.public_key),
+                client_alpha.wait_for_event_peer(beta.public_key, [State.Connected]),
+            )
         )
 
         ip_alpha: str = await testing.wait_long(
--- a/nat-lab/tests/test_reconnections.py
+++ b/nat-lab/tests/test_reconnections.py
@@ -16,6 +16,7 @@ from utils.ping import Ping
 
 @pytest.mark.asyncio
 @pytest.mark.derp
+@pytest.mark.xfail(reason="test is flaky - Jira issue: LLT-4161")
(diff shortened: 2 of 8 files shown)


nat-lab/tests/test_reconnections.py
```diff
@@ -16,6 +16,7 @@
 
 @pytest.mark.asyncio
 @pytest.mark.derp
+@pytest.mark.xfail(reason="test is flaky - Jira issue: LLT-4161")
```
gytsto, line 19: Not sure about this test

dfetti:
+1.0

nat-lab/tests/test_reconnections.py
lcruz99: If is flaky?

gytsto: yes

stalowyjez: Hmm, if I found which pipeline it was, it looks like some random fail, the logs look ok.

lcruz99: @stalowyjez it's on the linked Jira issue: https://bucket.digitalarsenal.net/low-level-hacks/vpn/client/libtelio-build/-/jobs/8645474

lcruz99: On these logs above, the expected behaviour is printed to stdout but it timesout when checking for the events. 

On this more recent pipeline the test failed again: https://bucket.digitalarsenal.net/low-level-hacks/vpn/client/libtelio-build/-/jobs/8682968#L2913

It failed also with different containers/params but the failure is the same: `wait_for_event_on_any_derp` timesout, even though the derp connection is logged to stdout. So my suspicion is the usual race condition between the action and `wait_for_event` functions. It can be reproduced introducing a `asyncio.sleep(1)` after `set_meshmap()`. I think one fix could be launching `wait_for_event` tasks asynchronously before restarting the device, or checking for the state instead of event.. I think the former fix suits better as the `wait_for_state()` function only checks the last reported state.

lcruz99: @gytsto this was the reason for including an `asyncio.Event()` as parameter of `wait_for_event()` functions, to guarantee we are waiting for the event before doing the action (restarting device in this case).

nat-lab/tests/test_events.py
```diff
@@ -398,10 +398,11 @@ async def test_event_content_exit_through_peer(
...
         await testing.wait_long(client_beta.get_router().create_exit_node_route())
 
-        await testing.wait_long(client_alpha.connect_to_exit_node(beta.public_key))
-
         await testing.wait_long(
-            client_alpha.wait_for_event_peer(beta.public_key, [State.Connected])
+            asyncio.gather(
+                client_alpha.connect_to_exit_node(beta.public_key),
+                client_alpha.wait_for_event_peer(beta.public_key, [State.Connected]),
+            )
```
stalowyjez, line 405: Hmm, maybe it would be more intuitive to first connect to beta node and then make it an exit node?

lcruz99: you mean `client_alpha.connect_to_exit_node(beta.public_key)` before `client_beta.get_router().create_exit_node_route()`?

gytsto: I think he meant something like this:

```
async with run_async_context(client_alpha.wait_for_event_peer(beta.public_key, [State.Connected])) as future:
    await testing.wait_long(client_alpha.connect_to_exit_node(beta.public_key))
    await testing.wait_long(future)
```

But putting them to `asyncio.gather` is already enough

lcruz99:
@dfetti @stalowyjez @gytsto Added one last final commit to use 60s as default timeout for direct connections, it should fix random fails due to slower executions, mainly on Windows VMs.

nat-lab/tests/test_reconnections.py
gytsto: Ok, let's leave it as it is for now

gytsto (dismissed):

stalowyjez:
+1.0

PR #86, opened 2023-08-22
LLT-3913: Deduplicate IPs of external meshent peers

mathiaspeters:
### Problem
Wireguard mandates that every connected peer has a unique IP address. So far we haven't validated that that is the case in our meshnet config. 

### Solution
We have two types of peers, internal and external. Internal peers will never have IP collisions with each other, but external peers might have collisions with internal and/or external peers. Added here is a deduplication step where overlapping IPs are removed from external peers. If two or more external peers have the same IP, only the first one specified in the config will keep the IP.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -16,7 +16,7 @@ use telio_traversal::{
     cross_ping_check::CrossPingCheckTrait, SessionKeeperTrait, UpgradeSyncTrait,
     WireGuardEndpointCandidateChangeEvent,
 };
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_info};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn};
 use telio_wg::{uapi::Peer, WireGuard};
 use thiserror::Error as TError;
 use tokio::sync::Mutex;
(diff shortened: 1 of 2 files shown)


src/device/wg_controller.rs
```diff
@@ -546,6 +550,42 @@ async fn build_requested_meshnet_peers_list<
...
+    let mut peer_ips = HashMap::new();
+    let mut occupied_ips = vec![];
+    let mut external_peers = vec![];
+    let mut process_ips_for_peer = |peer: &telio_model::config::Peer| {
+        if let Some(ips) = &peer.ip_addresses {
+            let ips = ips
+                .iter()
+                .filter_map(|ip| {
+                    if occupied_ips.contains(ip) {
+                        None
```
Jauler, line 566: Could we add some `info` or `warn` level log message here if peer IP duplicate has been detected? This seems to be very useful information when analyzing logs in case of issues

stalowyjez: Maybe it would be nice to print some more detailed info about the collision? Also, if two peers would have the same sets of allowed IPs the current implementation drops the latter, shouldn't the algo try to resolve such situation? (Maybe not at any cost, not sure if the problem is easy.)

stalowyjez:
+1

stalowyjez (approved):

PR #87, opened 2023-08-23
Use generated wg keys instead of hardcoded

gytsto:
### Problem
- vpn / wg-stun servers works unstable when peers with same public_keys are trying to connect alot of times in a row

### Solution
- generate wg key pair when registering to mesh_api and update vpn-server config

**Note:** This didn't help much for nordderper (wg-stun) somehow. Therefore restart of nordderper is still in `conftest.py`

- [x] README.md is updated
- [x] changelog.md is updated


nat-lab/tests/config.py
```diff
@@ -59,15 +59,21 @@ def get_root_path(path: str) -> str:
     "ipv4": "10.0.100.1",
     "port": 51820,
     "public_key": "N2Kuejq5Gd553kGYKY4iwtnCuk9MJjAPpg7C04FOdDw=",
+    "private_key": "IK88LLAc0zr0iEGZ3bZaXMYb2nFFNUBYOpEOyL6vtl0=",
```
Jauler, line 62: Even though we used a shared key between two VPN servers previsouly, could we add different private keys to both testing VPN servers (this will have to change corresponding public keys in the clients code)?

I think this key sharing may have unintended consequences, and introduce some flakyness due to WG roaming.

nat-lab/tests/mesh_api.py
```diff
@@ -163,76 +195,90 @@ def get_meshmap(
...
         if node_id not in self.nodes:
             raise MissingNodeError(node_id)
         return self.nodes[node_id]
+
+    @classmethod
+    def get_allowed_ip_list(cls, addrs: List[str]) -> List[str]:
+        return [
+            ip + "/32" if get_ip_address_type(ip) == IPProto.IPv4 else ip + "/128"
+            for ip in addrs
+            if get_ip_address_type(ip) in (IPProto.IPv4, IPProto.IPv6)
```
Jauler, line 237: Is it possible for `get_ip_address_type(...)` to return something that is not IPv4 nor IPv6?

gytsto: hm, actually no, exception will be thrown if anything else appears, so I guess this line is not needed

gytsto: ah sorry, it can return None, so it is needed actually

gytsto: improved this place abit

nat-lab/tests/config.py
gytsto: also added key generation

nat-lab/tests/mesh_api.py
```diff
@@ -163,76 +195,90 @@ def get_meshmap(
...
+        )
+
+        for node in node_list:
+            wg_conf += generate_peer_config(
+                node, ", ".join(cls.get_allowed_ip_list(node.ip_addresses))
+            )
+
+        full_command = (
+            f"docker exec {server_config['container']} bash -c 'echo \"{wg_conf}\" >"
+            " /wg_temp.conf; wg setconf wg0 /wg_temp.conf'"
```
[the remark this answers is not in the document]
gytsto: done

Jauler (approved):
+2.0

PR #90, opened 2023-08-23
telio-relay: Bump tokio-rustls version

lcruz99:
### Problem
There's a new DoS vulnerability in `webpki` crate reported by RustSec. (see https://rustsec.org/advisories/RUSTSEC-2023-0052)

`webpki` is an indirect dependency of `telio-relay`and this caused deny linter job to fail.

`telio-relay`-> `tokio-rustls` -> `rustls` -> `webpki`

### Solution

Update `tokio-rustls`.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.19.0"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
+checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -90,3 +90,6 @@ members = [
 exclude = [
 	"wireguard-go-rust-wrapper"
 ]
+
+[patch.crates-io]
+tokio-rustls = { git = 'https://github.com/rustls/tokio-rustls', branch = 'main' }
```
lcruz99, line 95: Maybe it would be better to provide a commit hash to avoid fetching a different version in the future?

lcruz99: cargo deny will throw an error if that happens though

Jauler: I do not think that this `patch.crates-io` entry is needed :thinking:

deny.toml
```diff
@@ -56,7 +56,9 @@ skip-tree = []
 unknown-registry = "warn"
 unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
-allow-git = []
+allow-git = [
+    "https://github.com/rustls/tokio-rustls?branch=main#b4fd5aa9391822320b4d0e3e2ed0b4ec877afa58",
+]
```
Jauler, line 61: I do not think that this allowed-git entry is needed :thinking:

Jauler (approved):
+2.0

PR #91, opened 2023-08-23
[LLT-3875] Fix derp disconnections

stalowyjez:
### Problem
Telio sometimes tries to reconnect with Derp server when already stopped

### Solution
Ensure that Derp doesn't try to reconnect when the Multiplexer channel is closed

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-relay/src/derp/mod.rs
+++ b/crates/telio-relay/src/derp/mod.rs
@@ -93,7 +93,7 @@ pub struct DerpRelay {
 
 struct State {
     /// Channel, that communicates with upper multiplexer module
-    channel: Chan<(PublicKey, PacketRelayed)>,
+    channel: Option<Chan<(PublicKey, PacketRelayed)>>,
     /// Derp configuration
     config: Option<Config>,
     /// Connection object
@@ -260,7 +260,7 @@ impl DerpRelay {
 
         Self {
             task: Task::start(State {
-                channel,
+                channel: Some(channel),
                 config: None,
                 conn: None,
                 event,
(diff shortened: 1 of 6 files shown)


crates/telio-relay/src/derp/mod.rs
```diff
@@ -565,11 +565,18 @@ impl Runtime for State {
 
         match &mut self.conn {
             Some(c) => {
+                let (upper_read, chan_tx) = match self.channel.as_mut() {
+                    Some(c) => (c.rx.recv(), &c.tx),
```
lcruz99, line 569: can you rename outer/inner c's for readability? also causing unneeded shadowing

crates/telio-relay/src/derp/mod.rs
```diff
@@ -565,11 +565,18 @@ impl Runtime for State {
...
+                    None => {
+                        // Similar case to config, but the log is already printed
+                        return (update.await)(self).await;
+                    }
+                };
```
lcruz99, line 574:
```suggestion
               if self.channel.is_none() {
                    return (update.await)(self).await;
                }
```

isn't this enough?

stalowyjez: It could be replaced by a flag and the option from channel could be removed, or probably in some other way as well, I thought that it is more clear this way, but I'm not really attached to it, so if you would like some other solution more I can change it.

crates/telio-relay/src/derp/mod.rs
```diff
@@ -565,11 +565,18 @@ impl Runtime for State {
 
         match &mut self.conn {
             Some(c) => {
```
lcruz99, line 567: Can we have Some(self.conn) && None(self.channel)?

Also, on the Some(conn) case there's no connection attempt, it is done only on the None(conn) case where you added the check ([line 616)](https://github.com/NordSecurity/libtelio/pull/91/files#diff-51a48b71eedfdd3893868ac7fb36eb992316673851e6a379ee230b84058a1843R616), isn't that enough?

if self.channel.is_none() {
                    return (update.await)(self).await;
                }

crates/telio-relay/src/derp/mod.rs
lcruz99: the channel was still defined later, but it's ok this is nitpicking

crates/telio-relay/src/derp/mod.rs
lcruz99: nitpicking

crates/telio-relay/src/derp/mod.rs
stalowyjez: >Can we have Some(self.conn) && None(self.channel)?

I don't think so, but we need a channel here, not an option, so I need to get it somehow. As I said, if you don't like this solution, I can move it to the large `match` statement, or replace with a flag, or refactor it in some other way

lcruz99 (commented):
+1

lcruz99 (approved):
nitpicks aside +2.

I took some time to understand this PR as found hard to relate the solution/problem you stated.. I suggest to add a better PR description next time.

gytsto (approved):

PR #94, opened 2023-08-25
Remove nordderper restart

gytsto:
### Problem

### Solution

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/mesh_api.py
+++ b/nat-lab/tests/mesh_api.py
@@ -1,5 +1,6 @@
 import os
 import pprint
+import random
 import time
 import uuid
 from config import DERP_SERVERS, WG_SERVERS
@@ -246,7 +247,8 @@ class API:
 
         wg_conf = (
             f"[Interface]\nPrivateKey = {server_config['private_key']}\nListenPort ="
-            f" {server_config['port']}\nAddress = 100.64.0.1/10, fd00::1/64\n\n"
+            f" {server_config['port']}\nAddress = 100.64.0.1/10,"
+            " fc74:656c:696f::1/64\n\n"
         )
 
         for node in node_list:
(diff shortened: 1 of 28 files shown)


nat-lab/tests/mesh_api.py
```diff
@@ -273,9 +275,14 @@ def config_dynamic_nodes(self, node_configs: List[bool]) -> Tuple[Node, ...]:
                 public_key=str(public),
                 is_local=is_local,
             )
-            # TODO correct subnet when we'll decide about the range
-            self.assign_ip(node.id, f"100.64.33.{node_idx}")
-            self.assign_ip(node.id, f"fd00::dead:{node_idx}")
+            ipv4 = (
+                f"100.{random.randint(65, 127)}.{random.randint(0, 255)}.{random.randint(8, 255)}"
```
matislovas, line 279: Not including `64` in second octet, and last octet should exclude `255` which is broadcast, but these are nitts/edge cases.

nat-lab/tests/mesh_api.py
```diff
@@ -273,9 +275,14 @@ def config_dynamic_nodes(self, node_configs: List[bool]) -> Tuple[Node, ...]:
...
+            )
+            ipv6 = (
+                f"fc74:656c:696f:0:{format(random.randint(0, 0xFFFF), 'x')}:{format(random.randint(0, 0xFFFF), 'x')}:{format(random.randint(0, 0xFFFF), 'x')}:{format(random.randint(0, 0xFFFF), 'x')}"
```
matislovas, line 282: `:0:` can be changed to `::` , last segment should have range of  `(8, 0xFFFF)` to account for reserved addresses.

Hasan6979:
lgtm! +1

nat-lab/tests/mesh_api.py
gytsto: updated

nat-lab/tests/mesh_api.py
gytsto: updated, but left `:0:` since there is no proper ipv6 string comparison... will leave it for future implementation

matislovas:
+0.5 for corrected IPv6 ranges

Jauler (approved):
+2.0

PR #95, opened 2023-08-25
Natlab police fix flaky tests

lcruz99:
### Problem
`test_event_content_meshnet_node_upgrade_direct` and `test_lana_with_second_node_joining_later_meshnet_id_can_change` are flaky.

### Solution
`test_event_content_meshnet_node_upgrade_direct` failed due to a unhandled utf-8 decode error, which this PR fixes.

`test_lana_with_second_node_joining_later_meshnet_id_can_change` flakyness cause is unknown and might be a caught bug by the test. This PR marks it as flaky.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_lana.py
+++ b/nat-lab/tests/test_lana.py
@@ -718,7 +718,7 @@ async def test_lana_with_disconnected_node() -> None:
 
 @pytest.mark.moose
 @pytest.mark.asyncio
-@pytest.mark.xfail
+@pytest.mark.xfail(reason="test is flaky - LLT-4187")
 async def test_lana_with_second_node_joining_later_meshnet_id_can_change() -> None:
     async with AsyncExitStack() as exit_stack:
         api = API()
--- a/nat-lab/tests/utils/process/docker_process.py
+++ b/nat-lab/tests/utils/process/docker_process.py
@@ -106,7 +106,9 @@ class DockerProcess(Process):
                     continue
                 decodeable_lines = lines[:-1]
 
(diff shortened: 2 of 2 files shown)


Jauler (approved):
+2.0

arg0d:
In which cases broken utf8 is produced?

lcruz99:
> In which cases broken utf8 is produced?

Not sure, but they occasionally happen, when they happen test fails because we didn't chose how to handle them.

arg0d:
I think it would be very good to understand when this actually happens. I agree that this will fix flakiness in the short them, but in the long term this is going to mask the real problem and eventually send someone on very hard journey trying to understand why a process is producing corrupt output.

gytsto:
> I think it would be very good to understand when this actually happens. I agree that this will fix flakiness in the short them, but in the long term this is going to mask the real problem and eventually send someone on very hard journey trying to understand why a process is producing corrupt output.

I agree with you 100%, but this is not the main problem. I think, whenever we get exception, buffer is cut in random places, causing those wrong utf characters. Which is basically aftermath of some other failure.

Jauler:
> > I think it would be very good to understand when this actually happens. I agree that this will fix flakiness in the short them, but in the long term this is going to mask the real problem and eventually send someone on very hard journey trying to understand why a process is producing corrupt output.
> 
> I agree with you 100%, but this is not the main problem. I think, whenever we get exception, buffer is cut in random places, causing those wrong utf characters. Which is basically aftermath of some other failure.

UniFFI should fix this issue.

PR #100, opened 2023-08-31
Update nordderper configs

stalowyjez:
Updates nordderper configs, so they are compliant with the new version of nordderper.

### Problem
New nordderper version requires updated config files.

### Solution
Update config files.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -29,4 +29,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           token: ${{ secrets.TOKEN }}
           schedule: ${{ github.event_name == 'schedule' }}
-          ref: v0.1.10
+          ref: v0.1.11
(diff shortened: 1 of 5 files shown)


Jauler (approved):
+2.0

gytsto (commented):
Add `limits` section to configs

gytsto (commented):
Also dont forget to update ref to build repo

gytsto (commented):
`tests/test_derp_client_message_forward.py::DERP_SERVER` and `DERP_SERVER_2` maybe needs an update also?

stalowyjez:
@gytsto it looks that tests/test_derp_client_message_forward.py::DERP_SERVER should use the user port, which is the same as previous port, so I think we should leave it as it is. About limits - I guess default ones are ok?

gytsto:
> @gytsto it looks that tests/test_derp_client_message_forward.py::DERP_SERVER should use the user port, which is the same as previous port, so I think we should leave it as it is. About limits - I guess default ones are ok?

Just wanted to add limits to config, so people know that there is an option of configuring it without looking at nordderper repo/source

jjanowsk (approved):

gytsto (approved):

PR #102, opened 2023-09-01
Move wireguard-go and norddeper into docker image 

jjanowsk:
Wireguard-go and nordderper were built on every natlab run. We can move
them to the docker image to speed up natlab startup time.


.github/workflows/gitlab.yml
```diff
@@ -29,4 +29,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           token: ${{ secrets.TOKEN }}
           schedule: ${{ github.event_name == 'schedule' }}
-          ref: v0.1.10
+          ref: LLT-4169-move-derper-and-wireguard-go
```
stalowyjez, line 32: Probably this should be changed to some tag before merging?

stalowyjez (commented):
Looks good, +1

Jauler (approved):
+1.0

gytsto (approved):

nat-lab/Dockerfile
```diff
@@ -5,14 +5,3 @@ FROM ${CONTAINER_REGISTRY}/low-level-hacks/vpn/client/libtelio-build/natlab-deps
 LABEL org.opencontainers.image.authors="info@nordsec.com"
 
 COPY --chmod=0755 bin/ /opt/bin/
-COPY data/nordderper.deb /tmp/
```
arg0d, line 8: How does the binary end up in the container? There is no mention of it in Dockerfile.

PR #103, opened 2023-09-01
natlab: Fix test teardown

lcruz99:
### Problem
Until now, if an exception was thrown during test execution, whether for timing out or any other reason, context managers were not able to properly exit. For instance, when timing out from pytest-timeout, the plugin throws a TimeOutError exception and it doesn't allow the user to catch it. Instead it handles it internally and replaces it by a pytest.fail() call, which in its turn raises a Fail exception.

Our actual implementation doesn't handle every kind of exception, naturally, and in every test whether we have an active ssh or docker connection the teardown needs to be carefully handled so that the cleanup takes place.

There was two main issues:
 - After exception was thrown, event loop would close before cleanup takes place (mainly for docker/ssh connections)
 - Client instance would hangup during teardown, trying to send a `dev stop` to a already closed `tcli` process.

### Solution
 - Adding a custom event loop, where all the tasks are manually cancelled when the test execution is going to stop.
 - Checking if tcli process is active before cleaning up, otherwise the Process concrete class will take care of it during teardown.
  
 PS: This not the ideal solution as the cleanup we design is not really happening, instead we cleanup by killing the processes. Nevertheless, this fixes the actual issue where configuration leftovers are not cleaned from the container, making other tests to fail.

With the ongoing improvement of the framework, we are going to be able to make pytest.fixtures from setup and teardown functions, isolating the setup/teardown phase from the test itself, avoiding these type of issues.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -1,0 +1,40 @@
+import asyncio
+import pytest
+
+
+def _cancel_all_tasks(loop: asyncio.AbstractEventLoop):
+    to_cancel = asyncio.tasks.all_tasks(loop)
+    if not to_cancel:
+        return
+
+    for task in to_cancel:
+        task.cancel()
+
+    loop.run_until_complete(asyncio.tasks.gather(*to_cancel, return_exceptions=True))
+
+    for task in to_cancel:
+        if task.cancelled():
+            continue
+        if task.exception() is not None:
+            loop.call_exception_handler(
+                {
+                    "message": "unhandled exception during asyncio.run() shutdown",
+                    "exception": task.exception(),
+                    "task": task,
+                }
+            )
+
+
+@pytest.fixture(scope="function")
+def event_loop():
+    loop = asyncio.get_event_loop_policy().new_event_loop()
+    try:
+        yield loop
+    finally:
+        try:
+            _cancel_all_tasks(loop)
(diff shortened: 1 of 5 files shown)


nat-lab/tests/conftest.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+import asyncio
+import signal
```
gytsto, line 3: unused `signal` and `logging` imports, run `autoflake`, will remove unused imports automagically :smile:

nat-lab/tests/test_direct_connection.py
```diff
@@ -595,7 +595,7 @@ async def test_direct_connection_loss_for_infinity(
 
 @pytest.mark.timeout(90)
 @pytest.mark.asyncio
-@pytest.mark.xfail(reason="test is flaky - LLT-4115")
+# @pytest.mark.xfail(reason="test is flaky - LLT-4115")
```
gytsto: Not sure about this, let's just leave it for natlab police to handle this?

gytsto (approved):
Besides python-linter issues, +2

nat-lab/tests/test_direct_connection.py
gytsto: oh, this is gone already :laughing:

gytsto (approved):

PR #105, opened 2023-09-04
Add integration test for IPv6 DNS service

dfetti:
In the end I decided to parameterize the existent DNS integration tests. It's not worth to duplicate the code.

- [x] README.md is updated
- [x] changelog.md is updated


nat-lab/tests/test_dns.py
```diff
@@ -16,14 +17,30 @@
...
+    IPV4 = auto()
+    IPV6 = auto()
 
 
 @pytest.mark.asyncio
-async def test_dns() -> None:
+@pytest.mark.parametrize(
+    "ip_version",
+    [
+        pytest.param(IpVersion.IPV4),
```
gytsto, line 33: Why not just parametrize ip itself, instead of ip version, cuz `ip_version` is only used to determine dns_server_address value. For exmaple:

```
@pytest.mark.parametrize("dns_server_address", [DNS_IPV4_SERVER_ADDRESS, DNS_IPV6_SERVER_ADDRESS])
```

dfetti: Look at the description. If you think is better with ip itself, I will change it.

gytsto: I guess it is just a matter of preference :smile:
I think you could also do something like this:
```
@pytest.mark.parametrize(
    "dns_server_address",
    [
        pytest.param(DNS_IPV4_SERVER_ADDRESS, id="IPv4"),
        pytest.param(DNS_IPV6_SERVER_ADDRESS, id="IPv6"),
    ],
)
```
So you wouldn't need introduce additional enum class and you will get your wanted representation

dfetti: Oh, cool. I'll do that.

gytsto (dismissed):
other than one nitpick, +2.0

gytsto (dismissed):

gytsto (approved):

PR #107, opened 2023-09-06
Mark test_dns as flaky

Hasan6979:
### Problem

### Solution

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_dns.py
+++ b/nat-lab/tests/test_dns.py
@@ -28,6 +28,7 @@ DNS_IPV6_SERVER_ADDRESS = config.LIBTELIO_DNS_IPV6
         pytest.param(DNS_IPV6_SERVER_ADDRESS, id="IPv6"),
     ],
 )
+@pytest.mark.xfail(reason="Test is flaky - JIRA issue LLT-4242")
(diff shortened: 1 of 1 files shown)


gytsto (approved):

nat-lab/tests/test_dns.py
```diff
@@ -28,6 +28,7 @@
         pytest.param(DNS_IPV6_SERVER_ADDRESS, id="IPv6"),
     ],
 )
+@pytest.mark.xfail(reason="Test is flaky - JIRA issue LLT-4242")
```
stalowyjez, line 31: Maybe I overlooked something, but I see this test failing only once and only for IPv6 setting, so maybe we could mark only IPv6 part, or even wait a bit if this failure reoccurs?

Hasan6979: We could see this entire week and then decide ?

Jauler (approved):
+1.0

Jauler:
Lets close this pull request and cancel the jira task because it seems like this flakyness is not really flakyness-enough :D

PR #108, opened 2023-09-06
Migrate natlab VMs to libvirt

arg0d:
MacOS VM has been updated to Catalina (previously Big-Sierra). Catalina
has different directory structure, and `/workspace` is not writeable.
Instead, copy the requires files to `/var/root/workspace`.

For some reason with libvirt Windows network interface names are not
consistent across different machines, i.e. the names are different when
running on my computer and when running on CI. Add some code to detect
the names of interfaces by their IP address.

`test_process_run_fail` was failing because of different error messgage
from ping. Rewrite the test to use `bash -c exit 77`.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/config.py
+++ b/nat-lab/tests/config.py
@@ -18,17 +18,25 @@ MAC_VM_IP = "10.55.0.12"
 LINUX_VM_PRIMARY_GATEWAY = "10.55.0.10"
 LINUX_VM_SECONDARY_GATEWAY = "10.66.0.10"
 
+PRIMARY_VM_NETWORK_PREFIX = "10.55.0"
+SECONDARY_VM_NETWORK_PREFIX = "10.66.0"
+
+# https://vagrant-libvirt.github.io/vagrant-libvirt/configuration.html#management-network
+LIBVIRT_MANAGEMENT_NETWORK_PREFIX = "192.168.121"
+
 # Same as defined in `libtelio/nat-lab/docker-compose.yml`, 10.0.0.0/16
 DOCKER_NETWORK_IP = "10.0.0.0"
 DOCKER_NETWORK_MASK = "255.255.0.0"
 STUN_SERVER = "10.0.1.1"
 
 # Same as defined in `libtelio/nat-lab/provision_windows.ps1`
-STUN_BINARY_PATH_WINDOWS = "C:/workspace/stunserver/release/stunclient.exe"
-STUN_BINARY_PATH_MAC = "/Users/vagrant/stunserver/stunclient"
+STUN_BINARY_PATH_WINDOWS = "C:/workspace/stunserver/release/stunclient.exe".replace(
+    "/", "\\"
+)
+STUN_BINARY_PATH_MAC = "/var/root/stunserver/stunclient"
 
-IPERF_BINARY_MAC = "/Users/vagrant/iperf3/iperf3"
-IPERF_BINARY_WINDOWS = "C:\\workspace\\iperf3\\iperf3.exe"
+IPERF_BINARY_MAC = "/var/root/iperf3/iperf3"
+IPERF_BINARY_WINDOWS = "C:/workspace/iperf3/iperf3.exe".replace("/", "\\")
 
 # JIRA issue: LLT-1664
 # The directories between host and Docker container are shared via
@@ -46,7 +54,8 @@ LIBTELIO_BINARY_PATH_DOCKER = (
 )
 
 # Libtelio binary path inside Windows and Mac VMs
-LIBTELIO_BINARY_PATH_VM = "/workspace/binaries/"
+LIBTELIO_BINARY_PATH_WINDOWS_VM = "C:/workspace/binaries/".replace("/", "\\")
+LIBTELIO_BINARY_PATH_MAC_VM = "/var/root/workspace/binaries/"
 
 LIBTELIO_DNS_IPV4 = "100.64.0.2"
 LIBTELIO_DNS_IPV6 = "fc74:656c:696f::2"
--- a/nat-lab/tests/test_process.py
+++ b/nat-lab/tests/test_process.py
@@ -89,46 +89,34 @@ async def test_process_run_success(
 
 
 @pytest.mark.parametrize(
-    "connection_tag,ping_command,error_msg",
+    "connection_tag,ping_command",
     [
         pytest.param(
             ConnectionTag.DOCKER_CONE_CLIENT_1,
-            ["ping", "asdfg"],
-            "ping: asdfg: Temporary failure in name resolution",
+            ["bash", "-c", "exit 77"],
         ),
         pytest.param(
             ConnectionTag.WINDOWS_VM,
-            ["ping", "-t", "asdfg"],
-            "Ping request could not find host asdfg",
+            ["cmd.exe", "/c", "exit 77"],
             marks=pytest.mark.windows,
         ),
         pytest.param(
             ConnectionTag.MAC_VM,
-            ["ping", "asdfg"],
-            "ping: cannot resolve asdfg: Unknown host",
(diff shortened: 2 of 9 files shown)


nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -1,16 +1,68 @@
...
+        matches = re.findall(
+            r"Configuration for interface \"(.*)\"[\s\S]*?IP Address:\s*([\d.]*)",
+            process.get_stdout(),
+        )
+
+        result: List[Interface] = []
+        for match in matches:
+            result.append(Interface(match[0], match[1]))
+
+        return result
```
stalowyjez, line 54: IMHO it looks more like `Interface` static method?

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -1,16 +1,68 @@
...
+            for interface in interfaces:
+                if interface.ipv4.startswith(prefix):
+                    return interface.name
+            assert False, f"interface not found with prefix `{prefix}`, {interfaces}"
+
+        return ConfiguredInterfaces(
+            find_interface(config.LIBVIRT_MANAGEMENT_NETWORK_PREFIX),
+            find_interface(config.PRIMARY_VM_NETWORK_PREFIX),
+            find_interface(config.SECONDARY_VM_NETWORK_PREFIX),
+        )
```
stalowyjez, line 37: It looks just like constructor

nat-lab/tests/config.py
```diff
@@ -18,17 +18,25 @@ def get_root_path(path: str) -> str:
 LINUX_VM_PRIMARY_GATEWAY = "10.55.0.10"
 LINUX_VM_SECONDARY_GATEWAY = "10.66.0.10"
 
+PRIMARY_VM_NETWORK_PREFIX = "10.55.0"
+SECONDARY_VM_NETWORK_PREFIX = "10.66.0"
+
+# https://vagrant-libvirt.github.io/vagrant-libvirt/configuration.html#management-network
+LIBVIRT_MANAGEMENT_NETWORK_PREFIX = "192.168.121"
+
```
stalowyjez, line 26: As I see these consts are used only in one file, maybe it would be more convenient to keep them there?

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -1,16 +1,68 @@
...
+        for match in matches:
+            result.append(Interface(match[0], match[1]))
+
+        return result
 
 
 class NetworkSwitcherWindows(NetworkSwitcher):
     def __init__(self, connection: Connection) -> None:
         self._connection = connection
+        self._interfaces: Optional[ConfiguredInterfaces] = None
```
stalowyjez, line 60: Are these interfaces supposed to change during the test run? Maybe you could initialize them in the constructor and then just use?

stalowyjez (commented):
It looks like a bunch of various refactors, are they all connected to libvirt?

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
arg0d: Nope, they aren't going to change during test run. I wish I could initialize them, but constructors can't be async :( Doing this lazy initialization is one of solutions, and the other solution would be to add "post init" method. Now that I think about it, adding post init method should be possible without changing the abstract network switcher interface.

arg0d: But I'm not sure what would be gained from that, because `_interfaces` would still be optional, and would need to be asserted before each use to satisfy the type checker. So the end result would be the same :(

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
arg0d: I didn't understand what you mean. `get_network_interfaces` should be renamed to `Interface`?

jjanowsk:
lgtm +1

nat-lab/tests/config.py
arg0d: I think the idea for `config.py` is to keep all Vagrant, Docker constants in a single place, to make it easier to make changes without having to change these magic constants throughout the code.

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
stalowyjez: Well, that's a pain. Maybe it could be initialized by with a list of Interfaces?

nat-lab/tests/test_process.py
```diff
@@ -89,46 +89,34 @@ async def test_process_run_success(
 
 
 @pytest.mark.parametrize(
-    "connection_tag,ping_command,error_msg",
+    "connection_tag,ping_command",
```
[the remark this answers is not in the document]
arg0d: fixed

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
stalowyjez: No, I meant that it doesn't really do anything with `ConfiguredInterfaces` objects, but operates on `Interface` objects, so I guess it would have more reason to make it a static method of `Interface` class

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
arg0d: I think that would work, I will work something out and you can take look at it again.

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
arg0d: Ok I see your point. I can move `get_network_interfaces` into `Interface` class. What do you think about getting rid of `Interface` class altogether and simply using named tuple? I haven't tried Python's named tuples yet, but if they work well I could use that instead.

arg0d: Nevermind the tuple, its defined the same way as a dataclass, so it still looks like a separate class. I will follow your suggestion.

arg0d:
> It looks like a bunch of various refactors, are they all connected to libvirt?

These changes are connected because of PR libtelio-build/128

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
stalowyjez: Hmm, in my imagination we could pass result of `get_network_interfaces` to it and make it not `async` and thus make it just a constructor, but maybe passing `ConfiguredInterfaces` is somehow nicer than `List[Interface]`? I'm not sure which one would be better TBH

stalowyjez:
+1

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
arg0d: Are you talking about something like this?
```
class ConfiguredInterfaces:
    def __init__(self, interface: List[Interface]) -> None:
        def find_interface(prefix: str) -> str:
            for interface in interfaces:
                if interface.ipv4.startswith(prefix):
                    return interface.name
            assert False, f"interface not found with prefix `{prefix}`, {interfaces}"

        self.default = find_interface(config.LIBVIRT_MANAGEMENT_NETWORK_PREFIX)
        self.primary = find_interface(config.PRIMARY_VM_NETWORK_PREFIX)
        self.secondary = find_interface(config.SECONDARY_VM_NETWORK_PREFIX)
```

I find this to be harder to read than the static method. I think at this point its a matter of personal preference.

Jauler (approved):
+1.0

PR #109, opened 2023-09-06
lana: Add moose init callback validation

lcruz99:
### Problem
Libtelio updated moose to a new async version which report’s it’s errors not by return but by callback.
One of the callbacks, `moose::InitCallback`, should be waited for and in case of error, analytics should be disabled entirely.

### Solution
Wait for `moose::InitCallback` value with a 2s timeout and in case there's an error, lana initialization fails with just an error message, continuing libtelio startup.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-lana/src/event_log_file.rs
+++ b/crates/telio-lana/src/event_log_file.rs
@@ -229,7 +229,7 @@ pub mod moose {
             match state {
                 TrackerState::Ready => Result::Success,
                 TrackerState::AlreadyStarted => Result::AlreadyInitiated,
-                _ => unreachable!("All other cases are not covered by Result enum"),
+                _ => unimplemented!("All other cases are not covered by Result enum"),
             }
         }
     }
--- a/crates/telio-lana/src/event_log_file.rs
+++ b/crates/telio-lana/src/event_log_file.rs
@@ -224,6 +224,25 @@ pub mod moose {
         FileDeleted,
     }
 
+    impl From<TrackerState> for Result {
+        fn from(state: TrackerState) -> Result {
+            match state {
+                TrackerState::Ready => Result::Success,
+                TrackerState::AlreadyStarted => Result::AlreadyInitiated,
+                _ => unreachable!("All other cases are not covered by Result enum"),
+            }
+        }
+    }
+
+    impl From<MooseError> for Error {
+        fn from(moose_error: MooseError) -> Error {
+            match moose_error {
+                MooseError::NotInitiated => Error::NotInitiatedError,
+                _ => Error::EventLogError,
+            }
+        }
+    }
+
     pub trait InitCallback: Send {
         fn after_init(&self, result: &std::result::Result<TrackerState, MooseError>);
     }
@@ -252,7 +271,7 @@ pub mod moose {
         init_cb: Box<(dyn InitCallback + 'static)>,
         error_cb: Box<(dyn ErrorCallback + Sync + std::marker::Send + 'static)>,
     ) -> std::result::Result<Result, Error> {
-        match super::event_log(
+        let result = match super::event_log(
             "init",
             Some(vec![
                 event_path.as_str(),
@@ -262,8 +281,15 @@ pub mod moose {
                 prod.to_string().as_str(),
             ]),
         ) {
-            Ok(_) => Ok(Result::Success),
-            _ => Err(Error::EventLogError),
+            Ok(_) => Ok(TrackerState::Ready),
+            _ => Err(MooseError::NotInitiated),
+        };
+
+        init_cb.after_init(&result);
+
+        match result {
+            Ok(r) => Ok(Result::from(r)),
+            Err(e) => Err(Error::from(e)),
         }
     }
 
--- a/crates/telio-lana/src/lib.rs
+++ b/crates/telio-lana/src/lib.rs
@@ -1,21 +1,31 @@
 #![deny(missing_docs)]
 //! Contains macros to simplify using Moose functions from within libtelio
 
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::{
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        mpsc::sync_channel,
+    },
+    time::Duration,
+};
 
-pub use telio_utils::telio_log_warn;
+pub use telio_utils::{telio_log_error, telio_log_warn};
 
 #[cfg_attr(all(feature = "moose", not(docrs)), path = "event_log_moose.rs")]
 #[cfg_attr(any(not(feature = "moose"), docsrs), path = "event_log_file.rs")]
 pub mod event_log;
+
 /// Module containing moose callbacks
 pub mod moose_callbacks;
+
 pub use event_log::*;
 
 /// App name used to initialize moose with
 pub const LANA_APP_NAME: &str = "libtelio";
 /// Version of the tracker used, should be updated everytime the tracker library is updated
 pub const LANA_MOOSE_VERSION: &str = "0.8.0";
+/// Timeout for moose to send init error, if any, through callback
+pub const LANA_MOOSE_MAX_INIT_TIME: u8 = 2;
 
 static MOOSE_INITIALIZED: AtomicBool = AtomicBool::new(false);
 const DEFAULT_ORDERING: Ordering = Ordering::SeqCst;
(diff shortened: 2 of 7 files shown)


gytsto (approved):
+1

crates/telio-lana/src/lib.rs
```diff
@@ -62,26 +68,39 @@ pub fn init_lana(
...
+            Box::new(moose_callbacks::MooseErrorCallback),
+        ) {
+            telio_log_warn!("[Moose] Error: {} on call to `{}`", err, "moose_init");
+            return Err(err);
+        };
 
-        if result.is_ok() {
-            MOOSE_INITIALIZED.store(true, DEFAULT_ORDERING);
-            init_context_info();
+        match rx.recv_timeout(Duration::from_secs(2)) {
```
tomaszklak, line 87: please make a named constant out of that duration, maybe something like `MOOSE_MAX_INIT_TIME` or something similar to make it clear what is it's purpose.

Also, how was this 2s value selected? Do we know what is the typical time it takes on low(est) powered devices we support?

tomaszklak (dismissed):
+1 if we at least get a nicer name for the timeout value :)

crates/telio-lana/src/lib.rs
lcruz99: Agree. I came up with this magic number of 2s inspired by the libdrop implementation, which seemed reasonable.

> Do we know what is the typical time it takes on low(est) powered devices we support?
 I have no idea 🤔

crates/telio-lana/src/moose_callbacks.rs
```diff
@@ -1,19 +1,28 @@
 pub use crate::event_log::*;
+use std::sync::mpsc::SyncSender;
 pub use telio_utils::{telio_log_error, telio_log_info, telio_log_warn};
 
-/// Struct for Moose callbacks
-pub struct MooseCallbacks;
+/// Struct for moose::InitCallback
+pub struct MooseInitCallback {
+    /// Channel tx half to send init result
+    pub init_tx: SyncSender<Result<moose::TrackerState, moose::MooseError>>,
```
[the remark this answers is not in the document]
lcruz99: Moose side implementation is expecting a blocking send call, and besides that we set the the channel buffer to 1, is there any extra reason to choose oneshot over the actual `SyncSender`?

> Also, we don't really read the value inside the result, so maybe we only need to send a bool and rename the field to something like init_success.

Agreed, will change.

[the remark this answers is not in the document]
tomaszklak: The main reason would be to express the fact that the is not an ordinary channel but only a single value will ever be transmitted over it. When someone reads this in future it will be obvious from the type that this will never be used to send multiple value. Like I said, it's a nitpick :)

crates/telio-lana/src/lib.rs
tomaszklak: Maybe moose devs know something about it and/or there are some metrics that track that? Additionally you could consider doing something like reducing your cpu frequency to the lowest possible value and measure it yourself in such conditions (I've mentioned how to do this on linux in the comment of LLT-3626).

tomaszklak (dismissed):

lcruz99:
@lcruz99

gytsto (approved):

crates/telio-lana/src/lib.rs
lcruz99: Limited cpu/stressed and the values were always between 10-20ms.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +140,50 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
 pub fn is_lana_initialized() -> bool {
     MOOSE_INITIALIZED.load(DEFAULT_ORDERING)
 }
+
+#[cfg(test)]
+mod test {
```
tomaszklak, line 145: Please add two more tests. One to verify behaviour when `moose::init` fails and a second one to verify behaviour when the `moose::init` doesn't call `init_cb`.

tomaszklak: In fact it should be 3 tests, since `init_cb` can either return an error or no be called at all.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +135,196 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
 pub fn is_lana_initialized() -> bool {
     MOOSE_INITIALIZED.load(DEFAULT_ORDERING)
```
tomaszklak, line 136: One thing I missed when we talked is this global flag. It requires us to do 2 things:
- restore it to false at the end of each test
- since we can't/won't change it to thread local, we need to mark all tests in this file to run sequentially (for example with `serial_test` crate)
The second point means that we might not need to put mock in thread local storage and a single global mutex protected variable might be enough.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +135,196 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
...
+        fn default() -> Self {
+            MooseMock {
+                result_fn: true,
+                result_cb: true,
+                timeout_cb: false,
+            }
+        }
+    }
+
+    fn set_custom_mock(i_mock: Option<MooseMock>) {
```
tomaszklak, line 250: there should be a matching call after each test that uses this function that resets the mock to `None`. The best way to do this is a wrapper that does this in the `Drop::drop`. This way we will handle both the normal exit and a panic.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +135,196 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
...
+            }
+        }
+    }
+
+    impl Default for MooseMock {
+        fn default() -> Self {
+            MooseMock {
+                result_fn: true,
+                result_cb: true,
+                timeout_cb: false,
```
tomaszklak, line 245: I think that both `_cb` and `_fn` suffixes are usually used for fields that store something callable, so some better names would be nice here. I would go with (in order): `return_success`, `init_cb_status` and `should_call_init_cb`. Additionally I would change the type of `init_cb_status` to be `Result<TrackerState, MooseError>`.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +135,196 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
...
+#[cfg(test)]
+mod test {
+    use super::event_log::moose;
+    use super::moose::{ErrorCallback, InitCallback};
+    use super::RefCell;
+    use super::{deinit_lana, init_lana, is_lana_initialized};
+
+    thread_local!(pub static MOCK: RefCell<Option<MooseMock>> = RefCell::new(None));
+
+    pub struct MooseMock {
```
tomaszklak, line 206: nit: Since we are not using any mocking library I think I would go with a `MooseStub` to make it more clear to readers not to expect eg. `mockall`.

crates/telio-lana/src/lib.rs
lcruz99: Added

crates/telio-lana/src/event_log_file.rs
```diff
@@ -224,6 +224,25 @@ pub mod moose {
         FileDeleted,
     }
 
+    impl From<TrackerState> for Result {
+        fn from(state: TrackerState) -> Result {
+            match state {
+                TrackerState::Ready => Result::Success,
+                TrackerState::AlreadyStarted => Result::AlreadyInitiated,
+                _ => unreachable!("All other cases are not covered by Result enum"),
```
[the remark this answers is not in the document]
lcruz99: Changed

crates/telio-lana/src/lib.rs
lcruz99: - Restored to false at the end of each test on teardown().
- Tests now run sequentially.

crates/telio-lana/src/lib.rs
lcruz99: Done on teardown()

crates/telio-lana/src/lib.rs
lcruz99: I agree, names changed.

crates/telio-lana/src/lib.rs
lcruz99: Changed.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +132,207 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
...
+    use std::sync::Mutex;
+
+    use super::{
+        deinit_lana,
+        event_log::moose,
+        init_lana, is_lana_initialized,
+        moose::{ErrorCallback, InitCallback},
+    };
+
+    pub static MOCK: Mutex<Option<MooseStub>> = Mutex::new(None);
```
tomaszklak, line 190: Now that we use the Stub suffix this global should also be renamed to STUB

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +132,207 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
...
+                should_call_init_cb: true,
+            }
+        }
+    }
+
+    fn set_custom_mock(mock: Option<MooseStub>) {
+        *MOCK.try_lock().unwrap() = mock;
+    }
+
+    fn teardown() {
```
tomaszklak, line 235: This approach doesn't handle exiting via panic, please handle it also.

crates/telio-lana/src/lib.rs
```diff
@@ -115,3 +132,207 @@ pub fn deinit_lana() -> Result<moose::Result, moose::Error> {
...
+/// * app_version - Indicates the semantic version of the application.
+/// * prod - whether the events should be sent to production or not
+/// * tx - channel tx half for the InitCallback instance
+pub fn init_moose(
+    event_path: String,
+    app_version: String,
+    prod: bool,
+    tx: SyncSender<bool>,
+) -> std::result::Result<moose::Result, moose::Error> {
+    cfg_if! {
```
tomaszklak, line 149: I don't think we need `cfg_if!` in this case.

crates/telio-lana/src/lib.rs
lcruz99: Changed

crates/telio-lana/src/lib.rs
lcruz99: Leftover... changed.

crates/telio-lana/src/lib.rs
lcruz99: Added wrapper to drop global on panic.

tomaszklak: Please also use it in all tests that require stub customization and remove all explicit calls to the `teardown` function.

lcruz99: We should still call `teardown` in the end of tests that do not need the stub, no?

tomaszklak (approved):

PR #110, opened 2023-09-08
Do not answer to DNS request if forward lookup fails

dfetti:
### Problem
The initial problem was that if the forward lookup from trust-dns fails, the failure is not indicated to libtelio. Instead the response buffer was filled with an empty lookup response. This, combined with the doubled retry mechanism (OS and trust-dns) produced some issues on IOS.

### Solution
Patched the trust-dns so now it indicates the failure and from libtelio side we try to copy the server behavior and do not respond to this requests.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-dns/Cargo.toml
+++ b/crates/telio-dns/Cargo.toml
@@ -8,11 +8,10 @@ publish = false
 
 [dependencies]
 rand = "0.8"
-trust-dns-client = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0" }
-trust-dns-proto = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0" }
-trust-dns-resolver = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0" }
-trust-dns-server = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0", features = ["resolver"] }
-
+hickory-client = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0" }
+hickory-proto = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0" }
+hickory-resolver = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0" }
+hickory-server = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0", features = ["resolver"] }
 async-trait.workspace = true
 base64.workspace = true
 boringtun.workspace = true
--- a/crates/telio-dns/src/zone.rs
+++ b/crates/telio-dns/src/zone.rs
@@ -1,18 +1,19 @@
 use async_trait::async_trait;
-use std::{
-    collections::{HashMap, HashSet},
-    net::IpAddr,
-    str::FromStr,
-};
-use trust_dns_client::rr::{rdata::SOA, DNSClass, LowerName, Name, RData, Record, RecordType};
-use trust_dns_resolver::config::{NameServerConfigGroup, ResolverOpts};
-use trust_dns_server::{
+use hickory_client::rr::{rdata::SOA, DNSClass, Name, RData, Record, RecordType};
+use hickory_proto::rr::{rdata, LowerName};
+use hickory_resolver::config::{NameServerConfigGroup, ResolverOpts};
+use hickory_server::{
     authority::{
         Authority, AuthorityObject, Catalog, LookupError, LookupOptions, MessageRequest,
         UpdateResult, ZoneType,
     },
     server::{Request, RequestInfo, ResponseHandler, ResponseInfo},
     store::{forwarder::ForwardConfig, in_memory::InMemoryAuthority},
+};
+use std::{
+    collections::{HashMap, HashSet},
+    net::IpAddr,
+    str::FromStr,
 };
 
 use crate::forward::ForwardAuthority;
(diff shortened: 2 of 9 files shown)


crates/telio-dns/Cargo.toml
```diff
@@ -8,11 +8,10 @@ publish = false
...
 rand = "0.8"
-trust-dns-client = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0" }
-trust-dns-proto = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0" }
-trust-dns-resolver = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0" }
-trust-dns-server = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v2.0.0", features = ["resolver"] }
-
+hickory-client = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0-test" }
+hickory-proto = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0-test" }
+hickory-resolver = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0-test" }
+hickory-server = { git = "https://github.com/NordSecurity/trust-dns.git", tag = "v3.0.0-test", features = ["resolver"] }
```
mathiaspeters, line 14: Just curious, what's with these name changes?

mathiaspeters:
Code looks good to me but it only needs to be a single commit so please fix the history

crates/telio-dns/tests/nameserver.rs
```diff
@@ -123,7 +123,7 @@ impl WGClient {
         }
 
         let result = timeout(
-            Duration::from_secs(10),
+            Duration::from_secs(11),
```
matszczygiel, line 126: Why?

crates/telio-dns/Cargo.toml
matszczygiel: The `trust-dns` author changed the crate name to `hickory`

matszczygiel:
+1

crates/telio-dns/Cargo.toml
tomaszklak: It's a nitpick but I would keep the old name here to make the diff smaller/simpler and make a follow up pr with a name change.

crates/telio-dns/src/zone.rs
```diff
@@ -95,13 +105,17 @@ impl AuthoritativeZone {
...
+        Ok(AuthoritativeZone {
+            zone: Arc::new(zone),
+        })
     }
 }
 
 #[async_trait]
-impl Authority for AuthoritativeZone {
-    type Lookup = <InMemoryAuthority as Authority>::Lookup;
+impl AuthorityObject for AuthoritativeZone {
```
tomaszklak, line 115: `AuthorityObject` is already implemented for anything that implements `Authority` (and is inside of an `Arc`). Maybe instead of putting an `Arc` inside of our `AuthoritativeZone` we should just wrap it in an `Arc`?

crates/telio-dns/tests/nameserver.rs
```diff
@@ -433,7 +434,7 @@ async fn dns_request_forward_to_slow_server() {
     // Start a query that will run for several seconds
     tokio::spawn(dns_test_with_server(
         "google.com",
-        DnsTestType::CorrectIpv4,
+        DnsTestType::NonRespondingForwardServer,
```
tomaszklak, line 437: Is this test still testing something? It should verify that we can modify the nameserver while the query **is still running**

crates/telio-dns/Cargo.toml
matszczygiel: Without the name change, it wouldn't find the package in the repo thus failing the compilation

crates/telio-dns/src/zone.rs
dfetti: I fixed it. I hope I understand right. Please correct me if not.

crates/telio-dns/Cargo.toml
dfetti: Is there a way we can work it around?

crates/telio-dns/tests/nameserver.rs
dfetti: It was a mistake. Sorry for that.

crates/telio-dns/tests/nameserver.rs
dfetti: As far as I understand, it validates that the nameserver is still available while having still running queries. Please, correct me if I'm wrong.
And yes, it makes sense to me to also validate if we can modify the nameserver while the query is still running, but I would prefer to do it in a different PR (for extending tests on telio-dns).

crates/telio-dns/Cargo.toml
matszczygiel: You can rename dependency in `Cargo.toml` for example:
```bar = { git = "https://github.com/example/project.git", package = "foo" }```
However IMHO it isn't worth it

crates/telio-dns/tests/nameserver.rs
tomaszklak: Sorry if I wasn't clear. I'm slightly worried that after the change to the underlying trust-dns library we no longer have a long running query here since it is 'immediately' fails with that new error. If you are confident that this is not the case feel free to resolve this conversation :)

tomaszklak (approved):
+1.0

tomaszklak (approved):

PR #114, opened 2023-09-11
natlab: Compare fp based on pk

lcruz99:
### Problem
`test_lana_with_second_node_joining_later_meshnet_id_can_change` was missing the logic of [`resolve_current_meshnet_id`](https://github.com/NordSecurity/libtelio/blob/f0d1dba72c9118529c8840b77fcbea4d3dc3dcde/crates/telio-nurse/src/heartbeat.rs#L181): when different meshnet IDs are used by the same number of nodes, it lexicographically compares the PK (in binary: base64->base2) of those nodes and chooses the smallest.

### Solution
Update test assert logic.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_lana.py
+++ b/nat-lab/tests/test_lana.py
@@ -1,4 +1,5 @@
 import asyncio
+import base64
 import pytest
 import subprocess
 import telio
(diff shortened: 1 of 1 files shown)


packgron (approved):
+2.0

gytsto (approved):

PR #115, opened 2023-09-11
Fix wg-stun behavior

stalowyjez:
### Problem
WG STUN to the first server was failing because it was connecting before the server was added as a WG peer.

### Solution
Sending STUN requests when STUN server is already added to wireguard peers

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/stun.rs
+++ b/crates/telio-traversal/src/endpoint_providers/stun.rs
@@ -92,7 +92,7 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
                 last_candidates: Vec::new(),
                 ping_pong_tracker: ping_pong_handler,
                 stun_peer_publisher,
-                stun_state: StunState::SearchingForServer,
+                stun_state: StunState::WaitingForWg,
             }),
         }
     }
@@ -116,7 +116,8 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
                 s.current_server_index = 0;
 
                 if s.stun_peer_publisher
-                    .try_send(s.servers.get(s.current_server_index).cloned())
+                    .send(s.servers.get(s.current_server_index).cloned())
+                    .await
                     .is_err()
                 {
                     telio_log_error!("Could not publish the STUN peer");
@@ -125,10 +126,9 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
 
             s.exponential_backoff.reset();
 
-            // Start STUN session, but only if we have at least one STUN server configured
-            if !s.servers.is_empty() && s.start_stun_session().await.is_err() {
-                telio_log_error!("STUN session could not be started");
-            }
+            s.transition_to_wait_for_wg();
+
+            s.try_transition_to_searching_for_server().await;
 
             Ok(())
         })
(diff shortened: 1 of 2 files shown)


crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -125,10 +126,7 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
 
             s.exponential_backoff.reset();
 
-            // Start STUN session, but only if we have at least one STUN server configured
-            if !s.servers.is_empty() && s.start_stun_session().await.is_err() {
-                telio_log_error!("STUN session could not be started");
-            }
+            s.transition_to_wait_for_wg().await;
```
ThrasherLT, line 129: Why is checking if there's at least one STUN server configured no longer necessary?

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -384,6 +403,21 @@ impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
             .await
     }
 
+    async fn transition_to_wait_for_wg(&mut self) {
+        self.stun_state = StunState::WaitingForWg;
+        self.current_timeout = PinnedSleep::new(STUN_TIMEOUT, ());
+
+        if self.is_wg_ready().await {
+            self.transition_to_searching_for_server();
```
ThrasherLT, line 411: Why is a function which is called "transition to wait for WG" doing a transition to searching for server?

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -294,7 +312,8 @@ impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
 
         if self
             .stun_peer_publisher
-            .try_send(self.servers.get(self.current_server_index).cloned())
+            .send(self.servers.get(self.current_server_index).cloned())
```
ThrasherLT, line 315: Why was this change made?

stalowyjez: This is changed to be consistent with channel usage in other endpoint providers. I can leave `try_send`, just thought it would be nicer to do it in the same way.

ThrasherLT: I wonder if originally `try_send()` was used instead of `send()` because there is some reason why the code can't await here if the channel buffer is full. I also wonder what are the implications if the channel won't send the message when the buffer is full.

stalowyjez: Yeah, because `next_server` was not `async`, so we couldn't wait here for blocking send.

crates/telio-traversal/src/endpoint_providers/stun.rs
stalowyjez: You're right, this is probably not a good idea

crates/telio-traversal/src/endpoint_providers/stun.rs
stalowyjez: Good point - I guess it should be moved a bit though

gytsto (approved):
+1

ThrasherLT:
+1

stalowyjez:
> Could you assign yourself to this task [LLT-4179](https://nordsec.atlassian.net/browse/LLT-4179)

@arvydasv sure

PR #118, opened 2023-09-14
Reenable RepeatedActions test

tomaszklak:
### Problem

It was previously disabled due to failing on tarpaulin builds. No more information was available but when run without tarpaulin it also seemingly randomly failed. It turns out that the loop run 4 iterations and the last one was for the 300ms which should have triggered both the 100ms and 150ms action. Since it selected them in a random order, the test failed.

### Solution

This change removes the loop + spawn which only made it less clear what is happening. Instead we explicitly spell out each assertion for each action we expect to happen. The last part where both actions are expected is verified using sets to make sure that the order doesn't matter.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3987,6 +3987,7 @@ dependencies = [
  "hashlink",
  "log",
  "lru_time_cache",
+ "maplit",
  "mockall",
  "parking_lot 0.12.1",
  "proptest",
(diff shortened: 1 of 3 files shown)


stalowyjez:
lgtm, +1

PR #119, opened 2023-09-18
Llt 4266 meshnet node which is exit node wrong statuses during meshnet shutdown

ThrasherLT:
### Problem
*LLT-4266 meshnet node which is also `exit_node` fires incorrect statuses during meshnet shutdown sequence, because exit node wasn't being cleared*

### Solution
*Add exit node clearing after meshnet disable while making sure that VPN node isn't cleared instead.*

- [x] README.md is updated
- [x] changelog.md is updated


baltrus1 (approved):
+1

changelog.md
```diff
@@ -20,6 +20,7 @@
 * LLT-4128: Add integration tests for IPv6 MagicDNS service
 * LLT-3316: Improve usage of peer-reflexive endpoints
 * LLT-4246: Add self nordname as well as self IP address to the DNS records
+* LLT-4266: Add exit node clearing when disabling meshenet
```
baltrus1, line 23: nitpick: meshnet spelling

ThrasherLT: Fixed

nat-lab/tests/test_mesh_exit_through_peer.py
```diff
@@ -111,3 +111,17 @@ async def test_mesh_exit_through_peer(
...
         assert alpha_conn_tracker.get_out_of_limits() is None
         assert beta_conn_tracker.get_out_of_limits() is None
+
+        # Testing if the exit node is cleared after disabling meshnet. See LLT-4266 for more details.
+        await client_alpha.set_mesh_off()
+        await testing.wait_long(
+            client_alpha.wait_for_event_peer(
+                beta.public_key, [State.Disconnected], list(telio.PathType)
+            )
+        )
```
mathiaspeters, line 121: Will this fail if there is a `connecting` state before the `disconnected` state`? Please correct me if I'm wrong but I don't think it does, and I think the next block that expects a `TimeoutError` will not verify either that there's no `connecting event` before `disconnecting` since this first block will clear those events

ThrasherLT: Fixed

ThrasherLT: Fixed.

nat-lab/tests/test_mesh_plus_vpn.py
```diff
@@ -365,6 +365,21 @@ async def test_vpn_plus_mesh(
         async with Ping(connection_alpha, beta.ip_addresses[0]).run() as ping:
             await testing.wait_long(ping.wait_for_next_ping())
 
+        # Testing if the VPN node is not cleared after disabling meshnet. See LLT-4266 for more details.
```
mathiaspeters, line 368: Question: does this also need to validate that other connection states are not fired? I can't tell, so just checking

ThrasherLT: No, this test is specifically needed just in case something's wrong with the code and it clears the VPN node when it shouldn't

mathiaspeters (approved):

PR #124, opened 2023-09-20
LLT-3451: Use workspace dependencies

mathiaspeters:
### Problem
There are a lot of dependencies that are shared between our crates, making it easy to miss something when updating those shared dependencies. Also, our dependencies were not very well organized everywhere.

### Solution
In this PR, workspace dependencies are used for any dependency that's used at least twice, with the exception of our fuzzer crates and `wireguard-go-rust-wrapper`. The dependencies are now also reorganized, where the sections are ordered `dependencies, dev-dependencies, build-dependencies`, with platform-specific dependencies coming after that, with their internal ordering the same. Within each dependency section, the dependencies are now group as `non-workspace deps, workspace deps, git deps, internal crates`, and each such group has been alphabetized.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,69 +17,71 @@ crate-type = ["staticlib", "cdylib", "lib"]
 pretend_to_be_macos = ["telio-model/pretend_to_be_macos"]
 
 [dependencies]
-anyhow = "1.0.64"
-base64 = "0.13.0"
 cfg-if = "1.0.0"
-crypto_box = { version = "0.8.2", features = ["std"] }
 ffi_helpers = "0.3.0"
-futures = "0.3"
-ipnetwork = "0.18"
-lazy_static = "1.4.0"
-libc = "0.2.99"
-log = {version = "0.4.14", features = ["release_max_level_debug"]}
-modifier = "0.1.0"
 num_cpus = "1.15.0"
-parking_lot = "0.12"
-serde = { version = "1.0", features = ["derive"] }
-serde_with = "1.10"
-serde_json = "1.0"
-thiserror = "1.0"
-tokio = { version = ">=1.22", features = ["full"] }
-async-trait = "0.1.51"
-uuid =  { version = "1.1.2", features = ["v4"] }
 wasm-bindgen = "=0.2.83" # Only for compatability with moose wasm version
 
-telio-crypto = { version = "0.1.0", path = "./crates/telio-crypto" }
-telio-dns = { version = "0.1.0", path = "./crates/telio-dns" }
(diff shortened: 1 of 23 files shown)


matislovas:
lgtm +2

Cargo.toml
```diff
@@ -90,3 +92,73 @@ members = [
...
+telio-nurse = { version = "0.1.0", path = "./crates/telio-nurse" }
+telio-proto = { version = "0.1.0", path = "./crates/telio-proto" }
+telio-proxy = { version = "0.1.0", path = "./crates/telio-proxy" }
+telio-relay = { version = "0.1.0", path = "./crates/telio-relay" }
+telio-sockets = { version = "0.1.0", path = "./crates/telio-sockets" }
+telio-task = { version = "0.1.0", path = "./crates/telio-task" }
+telio-test = { version = "1.0.0", path = "./crates/telio-test" }
+telio-traversal = { version = "0.1.0",  path = "./crates/telio-traversal" }
+telio-utils = { version = "0.1.0", path = "./crates/telio-utils" }
+telio-wg = { version = "0.1.0", path = "./crates/telio-wg" }
```
packgron, line 164: Missing newline

Cargo.toml
```diff
@@ -90,3 +92,73 @@ members = [
 exclude = [
 	"wireguard-go-rust-wrapper"
 ]
+
+[workspace.dependencies]
+anyhow = "1"
```
packgron, line 97: I personally feel like it would make sense to only leave library dependencies here. As it we only thing we truly ship. 
So something like clap, and anyhow, that is only build by clis, would not be added here.

But I don't have to strong of an opinion.

packgron:
Hmm I like reordering, but it would nice to add doc to know about this, for now maybe just add it to docs/cargo_depency_order.md

And mention in comment, for future, maybe queue a lint implementation?

packgron:
> Hmm I like reordering, but it would nice to add doc to know about this, for now maybe just add it to docs/cargo_depency_order.md
> 
> And mention in comment, for future, maybe queue a lint implementation?

Aslo it would probably good idea to say what is the ordering format in git commit :D

Cargo.toml
mathiaspeters: As far as I can tell, specifying a dependency in that section doesn't change the resulting artifacts. If it did, I would agree with you, but since it doesn't, I think the simplified maintenance makes it worth it

mathiaspeters:
@packgron `docs/cargo_dependency_ordering.md` and a follow-up ticket (LLT-4295) has been created

packgron (approved):
+2.0

packgron (approved):

PR #125, opened 2023-09-22
Build libtelio for apple tvOS target on arm64 

lcruz99:
### Problem
Apple team is evaluating needed work for having NordVPN app for apple TV. And for that libtelio should be compiled for that target as well.
For the time being, aarch64-apple-tvos is a [Tier 3 rustc target](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-tvos.html) so it doesn't have official rust support, this means there's no rust std lib available and many crates doesn't support it neither.

### Solution
This PR makes the necessary changes to make libtelio compatible with this target, and adds a patch section to cargo manifest that is used only when this target is used, as demonstrated in: [libtelio-build!104 (diffs)](https://bucket.digitalarsenal.net/low-level-hacks/vpn/client/libtelio-build/-/merge_requests/104/diffs).

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.cargo/config_apple_tvos
+++ b/.cargo/config_apple_tvos
@@ -1,0 +1,19 @@
+
+[patch.crates-io]
+cc = { git = 'https://github.com/nordsecurity/cc-rs.git', branch = 'add_apple_tvos_support' }
+tokio = { git = 'https://github.com/nordsecurity/tokio.git', branch = 'add_apple_tvos_support_1_32' }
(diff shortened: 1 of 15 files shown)


packgron:
lgtm +2.0

Small question,  ci changes are only in build repo ?

packgron (approved):

mathiaspeters (approved):
Seems like a mammoth job to update all those dependencies. Well done! +2

lcruz99:
> lgtm +2.0
> 
> Small question, ci changes are only in build repo ?

Yes, build repo and helper scripts.

PR #132, opened 2023-09-26
Clarify usage of hard-coded secrets

tomaszklak:
### Problem

There are multiple places where we use secrets such as passwords or keys to configure infrastructure. 

### Solution

All of those secrets are used in the context of tests. Most have been converted to be dynamically generated for each test run. All already reported cases have been fixed and some additional places have been modified (found by detect-secrets tool).

In case of nordderper configuration files, only clarification via comments has been added since rendering of fresh config file for nat-labs docker compose would needlessly complicate already non trivial process of starting nat-lab. If we will still receive reports about those secrets we can revisit this issue and make them dynamic.


packgron (commented):
lgtm +2.0

packgron (approved):

PR #133, opened 2023-09-27
LLT-4306 Fix infinite loop in STUN handler

mathiaspeters:
### Problem
An infinite loop was created in our STUN handler because a timeout was set too late. The timeout determines if we should try to start a STUN session or wait, but it was set after trying to start the STUN session. In cases where starting the STUN session failed, the timeout would never be updated, making the code think that we should start a STUN session, even though we just tried, leading to an infinite loop.

### Solution
Add error handling to the affected call site to transition to the right state

### Note
There will be a follow-up PR to add tests for this case. For this PR, there is just manual verification that it solves the issue

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/stun.rs
+++ b/crates/telio-traversal/src/endpoint_providers/stun.rs
@@ -60,6 +60,15 @@ impl<Wg: WireGuard> StunEndpointProvider<Wg> {
             stun_peer_publisher,
         ))
     }
+
+    /// Force STUN endpoint provider to reconnect
+    pub async fn reconnect(&self) {
+        let _ = task_exec!(&self.task, async move |s| {
+            let _ = s.reconnect();
+            Ok(())
+        })
+        .await;
+    }
 }
 
 impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
@@ -212,6 +221,7 @@ impl<Wg: WireGuard, E: Backoff + 'static> EndpointProvider for StunEndpointProvi
             .await))
         .await?
     }
+
     async fn get_current_endpoints(&self) -> Option<Vec<EndpointCandidate>> {
         task_exec!(&self.task, async move |s| Ok(Some(
             s.last_candidates.clone()
@@ -296,6 +306,14 @@ impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
                 public_key,
             )
             .await
+    }
+
+    async fn reconnect(&mut self) {
+        if let StunState::BackingOff = self.stun_state {
+            self.transition_to_wait_for_wg();
+            self.exponential_backoff.reset();
+            self.try_transition_to_searching_for_server().await;
+        }
     }
 
     /// Get wg port identified by stun
@@ -516,6 +534,7 @@ impl<Wg: WireGuard, E: Backoff> Runtime for State<Wg, E> {
             }
             // We are waiting either for stun session to timeout, or for the end of penalty
             _ = &mut self.current_timeout => {
+                telio_log_debug!("Processing timeout in {:?} state", self.stun_state);
                 match self.stun_state {
                     StunState::SearchingForServer | StunState::WaitingForWg => {
                         // This is a stun timeout. We should back off.
@@ -533,7 +552,11 @@ impl<Wg: WireGuard, E: Backoff> Runtime for State<Wg, E> {
                             self.transition_to_backing_off_state().await;
                         } else {
                             // This is a poll interval. We're still in HasEndpoints state.
-                            let _ = self.start_stun_session().await;
+                            let res = self.start_stun_session().await;
+                            if let Err(err) = res {
+                                telio_log_error!("Starting STUN session failed with error: {:?}", err);
+                                self.transition_to_backing_off_state().await;
+                            }
                         }
                     },
                 }
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -147,6 +147,14 @@ async fn consolidate_wg_peers<
         telio_log_info!("Inserting peer: {:?}", requested_peers.get(key));
         let peer = requested_peers.get(key).ok_or(Error::PeerNotFound)?;
         wireguard_interface.add_peer(peer.peer.clone()).await?;
+
+        if let Some(stun) = stun_ep_provider {
+            if let Some(wg_stun_server) = requested_state.wg_stun_server.as_ref() {
+                if wg_stun_server.public_key == *key {
+                    stun.trigger_endpoint_candidates_discovery().await?;
+                }
+            }
+        }
     }
 
     for key in update_keys {
@@ -164,14 +172,6 @@ async fn consolidate_wg_peers<
             wireguard_interface
                 .add_peer(requested_peer.peer.clone())
                 .await?;
-        }
-
-        if let Some(stun) = stun_ep_provider {
-            if let Some(wg_stun_server) = requested_state.wg_stun_server.as_ref() {
-                if wg_stun_server.public_key == *key {
-                    stun.trigger_endpoint_candidates_discovery().await?;
-                }
-            }
         }
 
         match (
--- a/changelog.md
+++ b/changelog.md
@@ -26,6 +26,7 @@
 * LLT-4266: Add exit node clearing when disabling meshnet
 * LLT-4159: Clarify purpose of hardcoded secrets in repo
 * LLT-3451: Use workspace dependencies
+* LLT-4306: Prevent infinite loop in STUN handler on socket error
 
 <br>
 
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -1106,6 +1106,12 @@ impl Runtime {
             .drop_connected_sockets()
             .await?;
 
+        if let Some(direct) = &self.entities.direct {
+            if let Some(stun) = &direct.stun_endpoint_provider {
+                stun.reconnect().await;
+            }
+        }
+
         self.entities.derp.reconnect().await;
         self.log_nat().await;
         Ok(())


matislovas:
seems trivial +2

packgron:
Hey @Jauler, I have one, though. If we get socket unreachable, should we transition to backing off state, and switch server?
This would trigger exponential backoff also.

I'm a bit torn.
From one side, it's a fail state, and previously would do exponential backoff + server switching. This definitely makes sence if you are pinging servers. As you don't want to accidentally flood in case of an outage.

From the other side, this case will typically happen due to local device's network issues. As such we are not spamming some servers, and backoff would just increase the time to recovery after a device is off network.

"Best" solution to me feels like it would be a combination of exponential backoff + reset by network_changed call.
But it's not reliably used by all platforms.

In conclusion I think that current solution is best for now, but what are your thoughts @Jauler

Jauler:
> Hey @Jauler, I have one, though. If we get socket unreachable, should we transition to backing off state, and switch server? This would trigger exponential backoff also.
> 
> I'm a bit torn. From one side, it's a fail state, and previously would do exponential backoff + server switching. This definitely makes sence if you are pinging servers. As you don't want to accidentally flood in case of an outage.
> 
> From the other side, this case will typically happen due to local device's network issues. As such we are not spamming some servers, and backoff would just increase the time to recovery after a device is off network.
> 
> "Best" solution to me feels like it would be a combination of exponential backoff + reset by network_changed call. But it's not reliably used by all platforms.
> 
> In conclusion I think that current solution is best for now, but what are your thoughts @Jauler

Well in my humble opinion, in this specific case, where there is no route towards our servers it is ok to retry every timeout without exponential backoff, but the error handling mechanism here is a general one used not just for this specific problem, but it could be triggered in different situations, I would imagine, that ICMP port unreachable messages could trigger socket errors under some circumstances and in such a case we are indeed sending outbound traffic, meaning that exponential backoff is required. I would consider this to be a rare edge case, but IMHO it is safer to apply exponential backoff anyways, even if it means, that recovery time may grow to a minute or two.

packgron:
> > Hey @Jauler, I have one, though. If we get socket unreachable, should we transition to backing off state, and switch server? This would trigger exponential backoff also.
> > I'm a bit torn. From one side, it's a fail state, and previously would do exponential backoff + server switching. This definitely makes sence if you are pinging servers. As you don't want to accidentally flood in case of an outage.
> > From the other side, this case will typically happen due to local device's network issues. As such we are not spamming some servers, and backoff would just increase the time to recovery after a device is off network.
> > "Best" solution to me feels like it would be a combination of exponential backoff + reset by network_changed call. But it's not reliably used by all platforms.
> > In conclusion I think that current solution is best for now, but what are your thoughts @Jauler
> 
> Well in my humble opinion, in this specific case, where there is no route towards our servers it is ok to retry every timeout without exponential backoff, but the error handling mechanism here is a general one used not just for this specific problem, but it could be triggered in different situations, I would imagine, that ICMP port unreachable messages could trigger socket errors under some circumstances and in such a case we are indeed sending outbound traffic, meaning that exponential backoff is required. I would consider this to be a rare edge case, but IMHO it is safer to apply exponential backoff anyways, even if it means, that recovery time may grow to a minute or two.
> 
> In order to transition to backoff, one just needs to call `self.transition_to_backing_off_state()` in error case here: https://github.com/NordSecurity/libtelio/pull/133/files#diff-f52b9bdf9a9389d5ebb2456346414b982fc5fd532c9cf08e23284a430ff06003R536

@mathiaspeters Let's go with exponential backoff and reset backoff, when we get notify_network_change  then sun is in BackingOffState

Jauler:
I think we are still missing error handling here: https://github.com/NordSecurity/libtelio/pull/133/files#diff-f52b9bdf9a9389d5ebb2456346414b982fc5fd532c9cf08e23284a430ff06003R552

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -60,6 +60,15 @@ impl<Wg: WireGuard> StunEndpointProvider<Wg> {
             stun_peer_publisher,
         ))
     }
+
+    ///
```
packgron, line 64: Write full comment :D

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -270,10 +280,10 @@ struct State<Wg: WireGuard, E: Backoff> {
 impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
     async fn start_stun_session(&mut self) -> Result<(), Error> {
         if self.stun_session.is_none() {
+            self.current_timeout = PinnedSleep::new(self.exponential_backoff.get_backoff(), ());
```
packgron, line 283: This should be left as it was STUN_TIMEOUT, backoff should happen in case of failure

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -298,6 +308,12 @@ impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
             .await
     }
 
+    async fn reconnect(&mut self) {
+        self.transition_to_wait_for_wg();
```
packgron, line 312: This should only trigger if in backoff state

crates/telio-traversal/src/endpoint_providers/stun.rs
mathiaspeters: Put it there to silence the editor warnings and forgot about it :see_no_evil: Fixed!

mathiaspeters:
@Jauler @packgron I've updated this PR, please have another look

stalowyjez:
@packgron @Jauler would it be a bad idea to just trigger endpoint candidates discovery on network change?

Jauler:
> @packgron @Jauler would it be a bad idea to just trigger endpoint candidates discovery on network change?

So it is pretty much what we are going with ;)

We do the exponential backoff to prevent infinite loop with `BackingOff` state and on `notify_network_change()` we reset the states, meaning, that if `notify_network_change()` is called when needed - recovery times would not suffer.

packgron (approved):
+2.0

Jauler:
Lets not merge just yet, I am getting weird lockups on this branch :(

Jauler:
@packgron there seems to be another issue, where `stun.trigger_endpoint_candidates_discovery()` is called here: https://github.com/NordSecurity/libtelio/blame/main/src/device/wg_controller.rs#L169, I think it should be in the loop which is processing `insert_peers` instead of `update_peers`.

packgron:
> @packgron there seems to be another issue, where `stun.trigger_endpoint_candidates_discovery()` is called here: https://github.com/NordSecurity/libtelio/blame/main/src/device/wg_controller.rs#L169, I think it should be in the loop which is processing `insert_peers` instead of `update_peers`. Would it make sense to fix it in this MR too?

Yeh so that one is done to notify reaction to updated peer.

There is basicaly two processes that happen on different tast's and we need to sync that.

stun.rs changes server, and report that new wg peer should be added.

then at some later time, device wait loop, recieves it, and reconsolidates.

at this reconsolidation we need to somehow report back to stun.rs, that is now save to send new wg_stun request.

We have introduced WaitingForWg there to solve this. and this triggerring was used to notify

I think it's probably problematic that we have double duty for that triggering


Wen we have config update, and then we update just peer.

Jauler:
> > @packgron there seems to be another issue, where `stun.trigger_endpoint_candidates_discovery()` is called here: https://github.com/NordSecurity/libtelio/blame/main/src/device/wg_controller.rs#L169, I think it should be in the loop which is processing `insert_peers` instead of `update_peers`. Would it make sense to fix it in this MR too?
> 
> Yeh so that one is done to notify reaction to updated peer.
> 
> There is basicaly two processes that happen on different tast's and we need to sync that.
> 
> stun.rs changes server, and report that new wg peer should be added.
> 
> then at some later time, device wait loop, recieves it, and reconsolidates.
> 
> at this reconsolidation we need to somehow report back to stun.rs, that is now save to send new wg_stun request.
> 
> We have introduced WaitingForWg there to solve this. and this triggerring was used to notify
> 
> I think it's probably problematic that we have double duty for that triggering
> 
> Wen we have config update, and then we update just peer.

Yeah, I understood why it is there, just that I think it should be in different place in `wg_controller`, currently it is here: https://github.com/NordSecurity/libtelio/blame/main/src/device/wg_controller.rs#L169 and this means, that it is called on every wg_controller consolidation event. We probably want to call it only when WG-STUN is inserted. (note that when we are rotating servers, there will be remove and insert events pair, as public key is changing).

Jauler (dismissed):
+1.0 @packgron @mathiaspeters we may need reapprovals, due to wg_controller change

src/device/wg_controller.rs
```diff
@@ -147,6 +147,15 @@ async fn consolidate_wg_peers<
         telio_log_info!("Inserting peer: {:?}", requested_peers.get(key));
         let peer = requested_peers.get(key).ok_or(Error::PeerNotFound)?;
         wireguard_interface.add_peer(peer.peer.clone()).await?;
+
+        if let Some(stun) = stun_ep_provider {
+            if let Some(wg_stun_server) = requested_state.wg_stun_server.as_ref() {
+                if wg_stun_server.public_key == *key {
+                    telio_log_debug!("OSCDEBUG: triggering EP disco disco party party");
```
packgron, line 154: Cleanup ?

Jauler: :laughing: oopsie, leftover some debugging logs :D

packgron (approved):
+1.0

Jauler (approved):
+1.0

PR #134, opened 2023-09-27
Wait some time before getting process list in test_process_run_not_found

tomaszklak:
### Problem

Sometimes process list seems to contain processes that have already ended. This causes `test_process_run_not_found` to fail sometimes.

### Solution

To account for above behaviour we wait 1 second before getting process list. This is not a 100% reliable solution but I'm not sure if there is a better one. Alternatively we could just remove the check of process listing since we already know that executing not existing process has failed via the exception. For now let's keep it this way, especially since few other similar tests are already performing sleep before getting the process list.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_process.py
+++ b/nat-lab/tests/test_process.py
@@ -120,7 +120,6 @@ async def test_process_run_fail(connection_tag: ConnectionTag, command: list[str
         assert " ".join(command) not in await _get_running_process_list(connection)
 
 
-@pytest.mark.xfail(reason="test is flaky - LLT-4311")
 @pytest.mark.parametrize(
     "connection_tag",
     [
@@ -148,6 +147,7 @@ async def test_process_run_not_found(connection_tag: ConnectionTag):
                 await process.wait_stdin_ready()
                 assert command in await _get_running_process_list(connection)
         assert e.value.cmd == [command]
+        await asyncio.sleep(1)
         assert command not in await _get_running_process_list(connection)
 
 


arg0d:
Will this actually fix the CI failures?
```
[2023-09-24 06:13:23]     async def test_process_run_not_found(connection_tag: ConnectionTag):
[2023-09-24 06:13:23]         async with AsyncExitStack() as exit_stack:
[2023-09-24 06:13:23]             connection = await exit_stack.enter_async_context(
[2023-09-24 06:13:23]                 new_connection_by_tag(connection_tag)
[2023-09-24 06:13:23]             )
[2023-09-24 06:13:23]             command = "non_existing_binary"
[2023-09-24 06:13:23]             with pytest.raises(ProcessExecError) as e:
[2023-09-24 06:13:23]                 async with connection.create_process([command]).run() as process:
[2023-09-24 06:13:23]                     await process.wait_stdin_ready()
[2023-09-24 06:13:23] >                   assert command in await _get_running_process_list(connection)
[2023-09-24 06:13:23] E                   AssertionError: assert 'non_existing_binary' in 
```

tomaszklak:
> Will this actually fix the CI failures?

Is there something specific you are worried about?

arg0d:
It doesn't look like this change will fix the error from CI. The `sleep(1)` is put after the error has already occurred.

tomaszklak:
> It doesn't look like this change will fix the error from CI. The `sleep(1)` is put after the error has already occurred.

Your, right - I was looking at the wrong process list yesterday... thanks for pointing this out. Will close this.

arg0d:
I think the test is racey at its core. Simply removing the checks for finding the process should be fine. It looks like the point of the test is to check that an exception is thrown, and searching for the process is not required to achieve that.

tomaszklak:
> I think the test is racey at its core. Simply removing the checks for finding the process should be fine. It looks like the point of the test is to check that an exception is thrown, and searching for the process is not required to achieve that.

It is but for a different reasons than I thought. I will submit a fix with explanation soon.

PR #135, opened 2023-09-27
Remove assertion which always fails but races with the rest of the test

tomaszklak:
The assertion which checks if the not existing executable is on the list of running processes, would **always** fails. But it's almost never reached in the test since at almost the same time the subprocess creation is throwing expected exception which most of the time is slightly faster/earlier. This is because we only wait for the stdout ready event which happens before we check the exit code and trigger an exception.

Since the assertion would always fail when reached and is occasionally reached, it is safe to remove it since it is incorrect.

This has been confirmed with additional logging and also by adding a sleep before the `raise` in the `DockerProcess.execute` which causes consistent test failures due to wrong exception being thrown.

Snippet of run with additional logging:

```
1695807387.393364 WAIT STDIN_READY <utils.process.docker_process.DockerProcess object at 0x7f2b8ad00b50>
1695807387.395445 READ LOOP ENTER ['non_existing_binary']
1695807387.3954818 GET PROCESS LIST (process is executing True)
1695807387.3974044 READ LOOP ENTER ['ps', 'aux']
1695807387.4835947 RAISE ['non_existing_binary'] 126
1695807387.5035448 process is executing False
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   3924  2944 ?        Ss   Sep26   0:00 bash /opt/bin/client
root          15  0.0  0.0   2484  1280 ?        S    Sep26   0:00 sleep infinity
root       48951  0.0  0.0   8088  4096 ?        Rs   09:36   0:00 ps aux
```

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_process.py
+++ b/nat-lab/tests/test_process.py
@@ -120,7 +120,6 @@ async def test_process_run_fail(connection_tag: ConnectionTag, command: list[str
         assert " ".join(command) not in await _get_running_process_list(connection)
 
 
-@pytest.mark.xfail(reason="test is flaky - LLT-4311")
 @pytest.mark.parametrize(
     "connection_tag",
     [
@@ -144,9 +143,8 @@ async def test_process_run_not_found(connection_tag: ConnectionTag):
         )
         command = "non_existing_binary"
         with pytest.raises(ProcessExecError) as e:
-            async with connection.create_process([command]).run() as process:
-                await process.wait_stdin_ready()
-                assert command in await _get_running_process_list(connection)
+            async with connection.create_process([command]).run():
+                await asyncio.sleep(1)
         assert e.value.cmd == [command]
         assert command not in await _get_running_process_list(connection)
 
--- a/nat-lab/tests/test_process.py
+++ b/nat-lab/tests/test_process.py
@@ -143,8 +143,8 @@ async def test_process_run_not_found(connection_tag: ConnectionTag):
         )
         command = "non_existing_binary"
         with pytest.raises(ProcessExecError) as e:
-            async with connection.create_process([command]).run():
-                await asyncio.sleep(1)
+            async with connection.create_process([command]).run() as process:
+                await process.is_done()
         assert e.value.cmd == [command]
         assert command not in await _get_running_process_list(connection)
 
--- a/nat-lab/tests/utils/process/docker_process.py
+++ b/nat-lab/tests/utils/process/docker_process.py
@@ -17,6 +17,7 @@ class DockerProcess(Process):
     _stderr: str
     _allowed_exit_codes: List[int]
     _stdin_ready: asyncio.Event
+    _is_done: asyncio.Event
     _stream: Optional[Stream]
     _execute: Optional[Exec]
 
@@ -26,6 +27,7 @@ class DockerProcess(Process):
         self._stdout = ""
         self._stderr = ""
         self._stdin_ready = asyncio.Event()
+        self._is_done = asyncio.Event()
         self._stream = None
         self._execute = None
 
@@ -63,7 +65,13 @@ class DockerProcess(Process):
         stdout_callback: Optional[StreamCallback] = None,
         stderr_callback: Optional[StreamCallback] = None,
     ) -> AsyncIterator["DockerProcess"]:
-        async with run_async_context(self.execute(stdout_callback, stderr_callback)):
+        async def mark_as_done():
+            try:
+                await self.execute(stdout_callback, stderr_callback)
+            finally:
+                self._is_done.set()
+
+        async with run_async_context(mark_as_done()):
             try:
                 yield self
             finally:
@@ -146,3 +154,6 @@ class DockerProcess(Process):
 
     def is_executing(self) -> bool:
         return self._stream is not None
+
+    async def is_done(self) -> None:
+        await self._is_done.wait()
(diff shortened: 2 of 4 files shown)


jjanowsk (approved):
+2

arg0d:
+2, but I think it might still be racey. `await process.wait_stdin_ready()` does not directly wait for the process to start. It only waits until our internal process abstraction is ready to start accepting stdin. This means the Docker process might not have had enough time to execute (and fail) when `await process.wait_stdin_ready()` exits.

tomaszklak:
> +2, but I think it might still be racey. `await process.wait_stdin_ready()` does not directly wait for the process to start. It only waits until our internal process abstraction is ready to start accepting stdin. This means the Docker process might not have had enough time to execute (and fail) when `await process.wait_stdin_ready()` exits.

You might be right, but I don't feel confident enough about python async to be sure. I could add some sort of wait on event that process is finished, which will never be reached since it would be set just before `return self` in `DockerProcess.execute` - what do you think?

tomaszklak:
Actually it just might be simpler to sleep indefinitely instead of waiting for anything.

arg0d:
I think you can just put `await asyncio.sleep(1)`. The process is going to fail almost immediately, so sleeping for 1 second should be more than enough time for the exception to occur. Also, because the tests expects an exception, it will never wait for the full 1 second. It will continue (and stop waiting) as soon the exception is raised. So I think in this case a simple sleep is a good enough synchronization measure.

tomaszklak:
I will add that 1s sleep since this should be enough but have to say I'm still not 100% sure I understand this :D

arg0d:
I somehow missed the fact that the process is running asynchronously. In that case the exception will only be thrown once `async with connection.create_process([command]).run() as process:` block exits. This is because the exception actually occurs in another coroutine, and it's only propagated to the main coroutine once the block is finished.

jjanowsk:
Sorry, it took me some time to go through all of this async code and understanding it, since I've been reading it for the first time. I have a question: what is the purpose of this test? :) Because I guess it's some kind of meta test to make sure our machinery when it comes to connections, processes etc. works as expected. And in that case I think resorting to a kind of "hack" is not the best. But I think it can be achieved by first waiting for `wait_stdin_ready` and then waiting until `is_executing` returns false. We can also think if those couldn't be wrapped into a function like: `executed` so you can `await process.executed()` to have a way to know when the process is completed. In current implementation just waiting till `is_executing` returns False is not enough because it is possible that you check it before the process even started to execute. So it would be nice to wrap it together with `wait_stdin_ready`. Just my 2 cents. Let me know what you think

tomaszklak:
I thought so too, but that's not going to work. When you wait for `False` from `is_exectuting()` like this:
```
diff --git a/nat-lab/tests/test_process.py b/nat-lab/tests/test_process.py
index c3fb7e5..eec07dc 100644
--- a/nat-lab/tests/test_process.py
+++ b/nat-lab/tests/test_process.py
@@ -143,8 +143,10 @@ async def test_process_run_not_found(connection_tag: ConnectionTag):
         )
         command = "non_existing_binary"
         with pytest.raises(ProcessExecError) as e:
-            async with connection.create_process([command]).run():
-                await asyncio.sleep(1)
+            async with connection.create_process([command]).run() as process:
+                await process.wait_stdin_ready()
+                while process.is_executing():
+                    pass
         assert e.value.cmd == [command]
         assert command not in await _get_running_process_list(connection)
```
What you will get is a kind of deadlock. The other coorutine that tries to run `non_existing_binary` will be stuck in the first iteration of a loop in `DockerProcess._read_loop` waiting for `message = await stream.read_out()`. Only when the body of `async with` in the test end, will `read_out()` complete returning `Message(stream=1, data=b'OCI runtime exec failed: exec failed: unable to start container process: exec: "non_existing_binary": executable file not found in $PATH: unknown\r\n')`.

arg0d:
That is a busy loop, it would fail to yield control to other coroutines. Using a short sleep instead of `pass` should work. The other coroutine is executing asynchronously, but the result of that coroutine (exception in this case) is only visible after joining it (exiting the `async with` block).
```
                while process.is_executing():
                    await asyncio.sleep(0.1)
```

jjanowsk:
Yes, that would require `sleep`. I'm aware of that and that's also why I would prefer to wrap it in the function. Additionally when you wrap it you do not have to use loop with sleep. You could add additional signal to the `Ssh/DockerProcess`  implementations and just wait for it inside this wrapper function.

tomaszklak:
> That is a busy loop, it would fail to yield control to other coroutines. Using a short sleep instead of `pass` should work. The other coroutine is executing asynchronously, but the result of that coroutine (exception in this case) is only visible after joining it (exiting the `async with` block).
> 
> ```
>                 while process.is_executing():
>                     await asyncio.sleep(0.1)
> ```

I thought I've tested that exact thing yesterday and it didn't work, but it does - thanks for pointing this out. I just pushed new version with synchronization based on the `asyncio.Event` but to be honest I'm not sure that it solves this completely or if there is still a short period of time when we exit async with block and subprocess has not raised/propagated the exception.

jjanowsk:
I think it does solve the issue. My understanding of the problem is that the exception was not raised at all. That is because when the `async with connection.create_process([command]).run() as process:` scope was exited it has waited for the process to get a pid and the process was killed by us in `DockerProcess/run/finally` before it had time to crash. Later the exception was not rasied because of `if exit_code and exit_code not in [0, 137]:` So that I think it is solved now. But even if not we'll find out easily if it starts to fail again.

jjanowsk (approved):
+2

PR #137, opened 2023-09-28
Fix test building inside of the telio-sockets crate

tomaszklak:
Before this change, when building this crate from `crates/telio-sockets` directory the compilation would fail. Because cargo would select socket2 in the 0.4.x version which lacks functions used in this crate. When build from top level libtelio directory cargo would select newer version 0.5.x which would work correctly. Which is why bumping to 0.5.x is safe since it is eitherway the only version that compiles and one which is selected when whole libtelio is being built.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3896,7 +3896,7 @@ dependencies = [
  "objc",
  "objc-foundation",
  "parking_lot 0.12.1",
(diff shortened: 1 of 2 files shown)


jjanowsk (approved):
+2

PR #139, opened 2023-09-29
LLT-4106: Skip ticks on analytics interval after telio was paused

arvydasv:
After a few analytic interval long pause, telio sends analytic data one after another to catch up. Setting the tick to skip the missed ticks.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/heartbeat.rs
+++ b/crates/telio-nurse/src/heartbeat.rs
@@ -23,7 +23,7 @@ use telio_task::{
 };
 use telio_utils::{map_enum, telio_log_debug, telio_log_error, telio_log_trace, telio_log_warn};
 use telio_wg::uapi::PeerState;
-use tokio::time::{interval_at, sleep, Duration, Instant, Interval, Sleep};
+use tokio::time::{interval_at, sleep, Duration, Instant, Interval, MissedTickBehavior, Sleep};
 use uuid::Uuid;
 
 use crate::config::HeartbeatConfig;
@@ -356,13 +356,15 @@ impl Analytics {
             // This way, the interval between events will be constant.
             Instant::now() + config.collect_interval - config.collect_answer_timeout
         };
+        let mut interval: Interval = interval_at(start_time, config.collect_interval);
+        interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
         let mut config_nodes = HashMap::new();
         // Add self in config_nodes hashset
         config_nodes.insert(public_key, true);
 
         Self {
-            task_interval: interval_at(start_time, config.collect_interval),
+            task_interval: interval,
             collect_period: Box::pin(sleep(FAR_FUTURE)),
             config,
             io,
--- a/crates/telio-nurse/src/heartbeat.rs
+++ b/crates/telio-nurse/src/heartbeat.rs
@@ -356,6 +356,7 @@ impl Analytics {
             // This way, the interval between events will be constant.
             Instant::now() + config.collect_interval - config.collect_answer_timeout
         };
+
         let mut interval: Interval = interval_at(start_time, config.collect_interval);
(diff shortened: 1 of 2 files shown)


tomaszklak:
Please add tests that verify this change.

arvydasv:
Tests would only check `tokio` tick behavior and not our code. Precisely this:
```interval.set_missed_tick_behavior(MissedTickBehavior::Skip);```

Don't think tests are necessary, and would require significant effort adding the `suspend` function for natlab.

**Comment by tomaszklak** [comment]
You introduced observable change in behaviour - it should be accompanied by a test that would fail before and now passes. Maybe a rust test, maybe a nat-lab test but there should be something that will prevent anyone from restoring previous behaviour.

File: crates/telio-nurse/src/heartbeat.rs [hunk 1]
```diff
@@ -1199,4 +1208,19 @@ mod tests {
...
+        // First tick should not fail. After this task_interval should catch up with the skipped ticks.
+        if let Err(_) = timeout(Duration::from_millis(200), analytics.task_interval.tick()).await {
+            assert_eq!(true, false);
+        }
+
+        // Ticks should have caught up and this should timeout
+        if let Ok(_) = timeout(Duration::from_millis(200), analytics.task_interval.tick()).await {
+            assert_eq!(true, false);
+        }
+    }
```
tomaszklak, line 1225: This doesn't check the change in bevaiour of sending messages, only that some property changed. Please write a test that verifies the change in the bevaiour.

If you are unsure if it is possible to observe this in test, here is a **hacked** together test that does that: https://github.com/NordSecurity/libtelio/commit/8e51a57dbc8c183142a87457c62179145ad72b39

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -1199,4 +1208,19 @@ mod tests {
         assert_eq!(true, meshnet_enabled);
         assert_eq!("vpn:7600617f9f9db5691a8c2768bd9d8110:2", external_links);
     }
+
+    #[tokio::test]
+    async fn test_send_analytics_report_once_on_skipped_ticks() {
```
matislovas, line 1213: Maybe use a paused time and advance the time [manually](https://docs.rs/tokio/latest/tokio/time/fn.advance.html)?

crates/telio-nurse/src/heartbeat.rs
arvydasv: Done

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -1199,4 +1204,43 @@ mod tests {
...
+
+        let rt = Task::start(analytics);
+
+        // Should succeed to get one analytic meesage
+        assert_eq!(
+            true,
+            timeout(Duration::from_secs(1), analytics_channel.recv())
+                .await
+                .is_ok()
+        );
```
tomaszklak, line 1234: nit: you can simply this to `assert!(timeout(...).await.is_ok());`

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -1199,4 +1204,43 @@ mod tests {
...
+
+        // Should timeout trying to get second analytic message
+        assert_eq!(
+            false,
+            timeout(Duration::from_secs(1), analytics_channel.recv())
+                .await
+                .is_ok()
+        );
```
tomaszklak, line 1242: nit: you can simplify this to `assert!(timeout(...).await.is_err());`

tomaszklak (approved):
+1, just a few optional nits, please don't forget to squash the commits into one with proper description.

crates/telio-nurse/src/heartbeat.rs
arvydasv: Good point. After my latest changes does not seems necessary now.

PR #140, opened 2023-10-02
add cat, because why not

haliucinas:
### Problem
No cats

### Solution
Add cats


Jauler:
It is super cute, but not really necessary for libtelio. Thanks for your help.

PR #146, opened 2023-10-03
Add tarpaulin based code coverage job

tomaszklak:
This is a port from previous version of CI and uses the same command as previously.

There might seem to be multiple other and better ways to do it but they are all broken in some ways.

We could use the action https://github.com/actions-rs/tarpaulin but that one is no longer maintained. The last version of tarpaulin it works with is 0.22 from October 2022 and the fix is still not merged in after multiple months: https://github.com/actions-rs/tarpaulin/pull/23 . Additionally there are discussions to deprecate this action: https://github.com/actions-rs/tarpaulin/issues/6

Alternatively we could have used the 'official' docker image of tarpaulin: https://hub.docker.com/r/xd009642/tarpaulin . This will not work since when github action runs in a docker, the container is not started with enough privileges which causes tarpaulin to crash, see as an example of this here:
https://github.com/xd009642/tarpaulin/issues/146

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,9 @@ jobs:
   benchmarks:
     needs: trigger
     uses: ./.github/workflows/benchmarks.yml
+  coverage:
+    needs: trigger
+    uses: ./.github/workflows/coverage.yml
 
   check-all-green:
     needs:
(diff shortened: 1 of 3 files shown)


packgron (dismissed):
lgtm +2.0

packgron (dismissed):

packgron (approved):

PR #152, opened 2023-10-06
LLT-4306: Add regression test for STUN infinite loop

mathiaspeters:
### Problem
There was a bug in our STUN handler that would cause an infinite loop in some scenarios recently. We prioritized getting a fix out, leaving the task of adding a regression test for later. This is that regression test.

The bug was triggered when there was a socket error to the STUN server after already establishing a STUN session.

### Solution
1. Establish direct connection between two nodes
2. Break UDP connection to derp servers to make STUN unreachable
3. Verify that there aren't many STUN requests being made

To fail the test, comment out line `558` in `stun.rs` and run the test. 

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/utils/router/mac_router.py
+++ b/nat-lab/tests/utils/router/mac_router.py
@@ -147,3 +147,9 @@ class MacRouter(Router):
         self, address: str  # pylint: disable=unused-argument
     ) -> AsyncIterator:
         yield
+
+    @asynccontextmanager
+    async def break_udp_conn_to_host(
+        self, address: str  # pylint: disable=unused-argument
+    ) -> AsyncIterator:
+        yield
--- a/nat-lab/tests/test_direct_connection.py
+++ b/nat-lab/tests/test_direct_connection.py
@@ -1,11 +1,12 @@
 import asyncio
+import config
 import pytest
 import telio
 from config import DERP_SERVERS
 from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass
 from mesh_api import API, Node
-from telio import PathType, State, AdapterType
+from telio import PathType, State, AdapterType, Client
 from telio_features import TelioFeatures, Direct
 from typing import List, AsyncIterator, Tuple, Optional
 from utils import testing
@@ -13,6 +14,7 @@ from utils.connection import Connection
 from utils.connection_util import (
     ConnectionTag,
     new_connection_with_tracker_and_gw,
+    new_connection_by_tag,
     ConnectionTracker,
     generate_connection_tracker_config,
     ConnectionLimits,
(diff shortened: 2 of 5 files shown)


nat-lab/tests/test_client_basic_stun.py
```diff
@@ -24,3 +30,70 @@ async def test_client_basic_stun(connection_tag: ConnectionTag, public_ip: str)
...
+                ["tcpdump", "-i", "any", "(", "port", "3478", "or", "3479", ")"]
+            ).run()
+        )
+        await asyncio.sleep(5)
+        
+        stun_requests = tcpdump.get_stdout().splitlines()
+        # There seems to be some delay when getting stdout from a process 
+        # Without this delay, `stun_requests` is empty even if tcpdump reports traffic
+        await asyncio.sleep(0.5)
+        assert len(stun_requests) < 20
```
baltrus1, line 99: why 20?

mathiaspeters: Semi-random number that's small enough to verify the issue isn't happening, but big enough that there shouldn't be any false positives

nat-lab/tests/test_client_basic_stun.py
```diff
@@ -24,3 +30,70 @@ async def test_client_basic_stun(connection_tag: ConnectionTag, public_ip: str)
...
+                alpha_client.get_router().break_udp_conn_to_host(str(server["ipv4"]))
+            )
+            await exit_stack.enter_async_context(
+                beta_client.get_router().break_udp_conn_to_host(str(server["ipv4"]))
+            )
+
+        # 3478 and 3479 are STUN ports in natlab containers
+        tcpdump = await exit_stack.enter_async_context(
+            alpha_connection.create_process(
+                ["tcpdump", "-i", "any", "(", "port", "3478", "or", "3479", ")"]
```
baltrus1, line 90: if you reject outgoing packets, do they not show up on tcpdump?

nat-lab/tests/test_client_basic_stun.py
```diff
@@ -24,3 +30,70 @@ async def test_client_basic_stun(connection_tag: ConnectionTag, public_ip: str)
...
+            ).run()
+        )
+        await asyncio.sleep(5)
+        
+        stun_requests = tcpdump.get_stdout().splitlines()
+        # There seems to be some delay when getting stdout from a process 
+        # Without this delay, `stun_requests` is empty even if tcpdump reports traffic
+        await asyncio.sleep(0.5)
```
baltrus1, line 98: how sure are you, that this will not introduce flakiness? 😢

mathiaspeters: Not very sure at all, but I first noticed that if I added a simple `print(...)` there the test worked and without it, it didn't, so it should not be flaky (famous last words)

nat-lab/tests/test_client_basic_stun.py
mathiaspeters: They're not really rejected, the firewall just responds with `icmp-host-unreachable`. And it does indeed show up in tcpdump

mathiaspeters: The point here of checking `tcpdump` is that the bug would make the code spam the STUN server in an infinite loop. The requests are supposed to show up in `tcpdump` (and they do), there just shouldn't be that many requests

baltrus1: riiight, thanks for explanation 👍

nat-lab/tests/test_client_basic_stun.py
baltrus1: surely it will be fine

nat-lab/tests/test_client_basic_stun.py
```diff
@@ -24,3 +29,71 @@ async def test_client_basic_stun(connection_tag: ConnectionTag, public_ip: str)
     async with new_connection_by_tag(connection_tag) as connection:
         ip = await testing.wait_long(stun.get(connection, config.STUN_SERVER))
         assert ip == public_ip, f"wrong public ip for the client {ip}"
+
+
+@pytest.mark.asyncio
+# Regression test for LLT-4306
+async def test_infinite_loop_in_stun_when_nonet() -> None:
```
stalowyjez, line 36: Two things here:
 - I'm not sure if this is the right file to place it (it looks a bit obsolete TBH) - I would either move it to a separate file, or just to direct connections,
 - What do you think about replacing `_when_nonet` with something more verbose? Or just getting rid of it?

nat-lab/tests/test_client_basic_stun.py
```diff
@@ -24,3 +29,71 @@ async def test_client_basic_stun(connection_tag: ConnectionTag, public_ip: str)
...
+            60,
+        )
+
+        for server in config.DERP_SERVERS:
+            await exit_stack.enter_async_context(
+                alpha_client.get_router().break_udp_conn_to_host(str(server["ipv4"]))
+            )
+            await exit_stack.enter_async_context(
+                beta_client.get_router().break_udp_conn_to_host(str(server["ipv4"]))
+            )
```
stalowyjez, line 85: Do we need to break connection for both clients?

nat-lab/tests/test_client_basic_stun.py
stalowyjez: I think it should be commented somehow in the code, if someone will read it in the feature it might not be familiar with this comment ^^'

mathiaspeters: Good point. Comment added

nat-lab/tests/test_client_basic_stun.py
mathiaspeters: No we don't. I removed the break for `beta_client`

nat-lab/tests/utils/router/mac_router.py
```diff
@@ -147,3 +147,9 @@ async def break_tcp_conn_to_host(
         self, address: str  # pylint: disable=unused-argument
     ) -> AsyncIterator:
         yield
+
+    @asynccontextmanager
+    async def break_udp_conn_to_host(
+        self, address: str  # pylint: disable=unused-argument
+    ) -> AsyncIterator:
+        yield
```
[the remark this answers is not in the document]
mathiaspeters: I think the current setup makes sense. Eventually we should implement this for the other platforms too, but for this is a good solution that makes it easy to add it to other platforms later

stalowyjez:
+1

baltrus1 (approved):
+2

PR #153, opened 2023-10-06
A special branch for testing derp client with self signed certs

Jauler:
### Problem
Branch for testing. Do not merge.

### Solution


Jauler:
This was not intended to be merged

PR #156, opened 2023-10-09
gitlab.yml: change runner label

edvyno1:
### Problem
*We are switching our Gitlab trigger runner for a different one*

### Solution
*Change label to new runner's label*


lcruz99:
What's the impact/need of these changes? Will different runners be triggered or only the one that triggers CI?

edvyno1:
> What's the impact/need of these changes? Will different runners be triggered or only the one that triggers CI?

We are turning off our old runner that was used to triggers and using this one to trigger gitlab pipelines instead. The runner switch was initiated by SRE.

lcruz99 (approved):
+2

You need a fully checked DoD template to be on the PR description to pass DoD job. Besides that, as you modified the `.github` directory you'll also need code owner's review.

jjanowsk (approved):
+2

PR #157, opened 2023-10-09
Make stun module rfc5389 compliant

stalowyjez:
### Problem
There are two problems with our current STUN module:
 - it is inconsistent - while on Windows and Mac we use stunman which is RFC5389 compliant, on Linux we use default Ubuntu package which is RFC3489 compliant, which is obsolete and
 - our nordderper implementation requires RFC5389 compliance from the connecting STUN clients.

### Solution
Instead of the default STUN Ubuntu client we can use `turnutils_stunclient` which is RFC5389 and also add a possibility of choosing STUN port, so we can use it to test nordderper.

- [x] README.md is updated
- [x] changelog.md is updated


Jauler (approved):
+2.0

PR #159, opened 2023-10-09
LLT-4286: Add logic for ICMP error types in firewall

mathiaspeters:
### Problem
A previous PR added logic to the firewall for dealing with ICMP packets. That PR didn't include handling of ICMP error types, in order to keep the PRs simple. This PR handles that.

### Solution
An ICMP error packet contains the original packet as it's body. The ICMP cache key builder will now, when it sees an error type, unpack the body and try to build a cache key from that.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -291,6 +291,16 @@ struct IcmpConn {
     identifier_sequence: u32,
 }
 
+type IcmpKey = Result<IcmpConn, IcmpErrorKey>;
+
+#[derive(Clone, Debug)]
+enum IcmpErrorKey {
+    Udp(Option<IpConnWithPort>),
+    Tcp(Option<IpConnWithPort>),
+    Icmp(Option<IcmpConn>),
+    None,
+}
+
 impl Debug for IcmpConn {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("IcmpConn")
@@ -433,13 +443,7 @@ impl StatefullFirewall {
     }
 
     fn handle_outbound_udp<'a>(&self, ip: &impl IpPacket<'a>) {
-        let udp_packet = unwrap_option_or_return!(UdpPacket::new(ip.payload()));
-        let key = IpConnWithPort {
-            remote_addr: ip.get_destination().into(),
-            remote_port: udp_packet.get_destination(),
-            local_addr: ip.get_source().into(),
-            local_port: udp_packet.get_source(),
-        };
+        let key = unwrap_option_or_return!(Self::build_udp_key(ip, false));
         let mut udp_cache = unwrap_lock_or_return!(self.udp.lock());
         // If key already exists, dont change the value, just update timer with entry
         if let Entry::Vacant(e) = udp_cache.entry(key) {
@@ -452,18 +456,10 @@ impl StatefullFirewall {
     }
 
     fn handle_outbound_tcp<'a>(&self, ip: &impl IpPacket<'a>) {
-        let tcp_packet = unwrap_option_or_return!(TcpPacket::new(ip.payload()));
-        let flags = tcp_packet.get_flags();
+        let (key, flags) = unwrap_option_or_return!(Self::build_tcp_key(ip, false));
+        let mut tcp_cache = unwrap_lock_or_return!(self.tcp.lock());
 
         if flags & TCP_FIRST_PKT_MASK == TcpFlags::SYN {
-            let key = IpConnWithPort {
-                remote_addr: ip.get_destination().into(),
-                remote_port: tcp_packet.get_destination(),
-                local_addr: ip.get_source().into(),
-                local_port: tcp_packet.get_source(),
-            };
-
-            let mut tcp_cache = unwrap_lock_or_return!(self.tcp.lock());
             telio_log_trace!("Inserting TCP conntrack entry {:?}", key);
             tcp_cache.insert(
                 key,
@@ -474,24 +470,9 @@ impl StatefullFirewall {
                 },
             );
         } else if flags & TcpFlags::RST == TcpFlags::RST {
-            let key = IpConnWithPort {
-                remote_addr: ip.get_destination().into(),
-                remote_port: tcp_packet.get_destination(),
-                local_addr: ip.get_source().into(),
-                local_port: tcp_packet.get_source(),
-            };
-
             telio_log_trace!("Removing TCP conntrack entry {:?}", key);
-            unwrap_lock_or_return!(self.tcp.lock()).remove(&key);
+            tcp_cache.remove(&key);
         } else if flags & TcpFlags::FIN == TcpFlags::FIN {
-            let key = IpConnWithPort {
-                remote_addr: ip.get_destination().into(),
-                remote_port: tcp_packet.get_destination(),
-                local_addr: ip.get_source().into(),
-                local_port: tcp_packet.get_source(),
-            };
-
-            let mut tcp_cache = unwrap_lock_or_return!(self.tcp.lock());
             telio_log_trace!("Connection {:?} closing", key);
             if let Entry::Occupied(mut e) = tcp_cache.entry(key) {
                 let TcpConnectionInfo { tx_alive, .. } = e.get_mut();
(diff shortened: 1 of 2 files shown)


crates/telio-firewall/src/firewall.rs
```diff
@@ -685,6 +685,27 @@ impl StatefullFirewall {
                     || it == v6::EchoReply.0
                 {
                     it
+                } else if (it == v4::DestinationUnreachable.0
+                    || it == v4::TimeExceeded.0
+                    || it == v4::ParameterProblem.0)
+                    && ip.get_next_level_protocol() == IpNextHeaderProtocols::Icmp
+                {
+                    let packet = unwrap_option_or_return!(
+                        Ipv4Packet::try_from(&icmp_packet.payload()[4..]),
```
Jauler, line 694: we should avoid using `[]` operator here (wonder why linter is not crying bout it, because it will panic if `payload` is shorter than 4 bytes, which effectively could create potential DoS.

crates/telio-firewall/src/firewall.rs
```diff
@@ -685,6 +685,27 @@ impl StatefullFirewall {
...
+                    );
+                    return Self::build_icmp_key(&packet, false);
+                } else if (it == v6::DestinationUnreachable.0
+                    || it == v6::PacketTooBig.0
+                    || it == v6::ParameterProblem.0
+                    || it == v6::TimeExceeded.0)
+                    && ip.get_next_level_protocol() == IpNextHeaderProtocols::Icmpv6
+                {
+                    let packet = unwrap_option_or_return!(
+                        Ipv6Packet::try_from(&icmp_packet.payload()[4..]),
```
Jauler, line 705: Same as https://github.com/NordSecurity/libtelio/pull/159/files#r1360433998

crates/telio-firewall/src/firewall.rs
```diff
@@ -685,6 +685,27 @@ impl StatefullFirewall {
...
+                } else if (it == v4::DestinationUnreachable.0
+                    || it == v4::TimeExceeded.0
+                    || it == v4::ParameterProblem.0)
+                    && ip.get_next_level_protocol() == IpNextHeaderProtocols::Icmp
+                {
+                    let packet = unwrap_option_or_return!(
+                        Ipv4Packet::try_from(&icmp_packet.payload()[4..]),
+                        None
+                    );
+                    return Self::build_icmp_key(&packet, false);
```
Jauler, line 697: I believe the call to build_icmp_key(...) here will almost always return None, because the packet contained in icmp_packet.payload() is generally not gonna be ICMP one.

crates/telio-firewall/src/firewall.rs
```diff
@@ -685,6 +685,27 @@ impl StatefullFirewall {
...
+                    || it == v6::PacketTooBig.0
+                    || it == v6::ParameterProblem.0
+                    || it == v6::TimeExceeded.0)
+                    && ip.get_next_level_protocol() == IpNextHeaderProtocols::Icmpv6
+                {
+                    let packet = unwrap_option_or_return!(
+                        Ipv6Packet::try_from(&icmp_packet.payload()[4..]),
+                        None
+                    );
+                    return Self::build_icmp_key(&packet, false);
```
Jauler, line 708: Same as https://github.com/NordSecurity/libtelio/pull/159/files#r1360652683

crates/telio-firewall/src/firewall.rs
```diff
@@ -1678,6 +1715,59 @@ pub mod tests {
...
+                false
+            );
+            assert_eq!(
+                fw.process_inbound_packet(&make_peer(), &actual_response),
+                true
+            );
+        }
+    }
+
     #[rustfmt::skip]
```
Jauler, line 1771: Keeping in mind https://github.com/NordSecurity/libtelio/pull/159/files#r1360652683, maybe it would make sense to add two tests where ICMP in its own payload contains UDP packet in one case and then TCP packet in another case?

crates/telio-firewall/src/firewall.rs
```diff
@@ -704,30 +743,115 @@ impl StatefullFirewall {
...
             local_addr,
             response_type,
             identifier_sequence,
+        };
+        Some(IcmpKey::IcmpConn(key))
+    }
+
+    fn build_icmp_error_key(icmp_packet: IcmpPacket, is_v4: bool) -> Option<IcmpKey> {
+        let inner_packet = unwrap_option_or_return!(icmp_packet.payload().get(4..), None);
+        let (udp_key, tcp_key, icmp_key) = if is_v4 {
```
mathiaspeters, line 769: I would love to make this generic so that the if and else blocks don't need to be duplicated when the only difference is a single type change, `Ipv4Packet` and `Ipv6Packet`, but due to lifetimes I couldn't make it happen without resorting to a macro, and I don't want to introduce that for this small case

crates/telio-firewall/src/firewall.rs
```diff
@@ -1090,14 +1118,30 @@ pub mod tests {
             IPV4_HEADER_MIN,
             ip_len,
         );
+        //ip.set_payload(body);
```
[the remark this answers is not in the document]
mathiaspeters: Good catch, thanks!

crates/telio-firewall/src/firewall.rs
```diff
@@ -291,6 +291,16 @@ struct IcmpConn {
     identifier_sequence: u32,
 }
 
+#[derive(Clone, Debug)]
+enum IcmpKey {
+    IcmpConn(IcmpConn),
+    Error {
+        udp_key: Option<IpConnWithPort>,
+        tcp_key: Option<IpConnWithPort>,
+        icmp_key: Option<IcmpConn>,
```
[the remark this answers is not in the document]
mathiaspeters: I think I was for some most likely bad reason trying to not create extra structs/enum but this is of course the way to go

Jauler (approved):
+1.0. Very nice! Tested live and looks very nice.

crates/telio-firewall/src/firewall.rs
```diff
@@ -704,29 +743,113 @@ impl StatefullFirewall {
...
+                }
+                IpNextHeaderProtocols::Tcp => {
+                    IcmpErrorKey::Tcp(Self::build_tcp_key(&packet, false).map(|(key, _)| key))
+                }
+                IpNextHeaderProtocols::Icmp => IcmpErrorKey::Icmp(
+                    Self::build_icmp_key(&packet, false).and_then(|key| match key {
+                        IcmpKey::IcmpConn(key) => Some(key),
+                        IcmpKey::Error { .. } => None,
+                    }),
+                ),
```
tomaszklak, line 783: Seems to me that all udp packets that fail to parse will have the same key. Similarly for the tcp packets (and possibly icmp). I would expect that it would be an error of some kind, not just an ordinary key value.

That is why I was expecting the `IcmpErrorKey` variants not to contain options.

crates/telio-firewall/src/firewall.rs
```diff
@@ -501,12 +486,13 @@ impl StatefullFirewall {
...
-        if let Entry::Vacant(e) = icmp_cache.entry(key) {
-            telio_log_trace!("Inserting new ICMP conntrack entry {:?}", e.key());
-            e.insert(());
+        if let Some(IcmpKey::IcmpConn(key)) = Self::build_icmp_key(ip, false) {
+            // If key already exists, dont change the value, just update timer with entry
+            if let Entry::Vacant(e) = icmp_cache.entry(key) {
+                telio_log_trace!("Inserting new ICMP conntrack entry {:?}", e.key());
+                e.insert(());
+            }
         }
```
tomaszklak, line 496: maybe an else case with a trace log here? Seems to me that failing to build the key and completely skipping processing of this packet (a change to previous behaviour) could be something interesting to see in logs.

mathiaspeters: We are still processing the packet if building the key fail. All code in this file determines if we allow or reject inbound and outbound packets, and we allow all outbound packets, so it's really about determining if we allow or reject inbound packets. If we simplify that, it becomes allow if it's in the whitelist or cache, reject otherwise. This block of code here just determines which ICMP packets end up in the cache, but that doesn't affect whether we allow the outbound packet. 

Adding an else with a log statement would only really be helpful if we are worried about sending malformed ICMP packets. But if we are, we aren't allowing responses to those packets since it's not added to the cache. I can add it if you still think it's useful, but to me it isn't.

crates/telio-firewall/src/firewall.rs
mathiaspeters: Technically yes but practically no. Failing to parse a packet that is nested in an ICMP error shouldn't really produce an error, it just signals that we should reject the packet. If we can't parse a packet the key becomes `None`, which is automatically rejected.

crates/telio-firewall/src/firewall.rs
tomaszklak: That kind of log might be useful if it turns out that parsing of packet is subtly wrong and we are failing to parse correct packet. I think it's worth adding trace log for that.

crates/telio-firewall/src/firewall.rs
tomaszklak: You are right but that made me realize that the key is not the key and I was thinking all this time tha it is and it ends up the the hashmap. As far as I can see `IcmpKey` should be an alias to `Result<IcmpConn, IcmpError>` where `IcmpError` is renamed `IcmpErrorKey`.

mathiaspeters: Good idea. I simplified some other stuff as well to rely on `IcmpErrorKey::None` to simplify `Option<IcmpKey>` to just `IcmpKey` as well.

crates/telio-firewall/src/firewall.rs
mathiaspeters: I don't think it makes sense to add that log statement here, but I added log statements in a few other places that would be useful for such diagnostic information

tomaszklak (approved):
+1.0

PR #160, opened 2023-10-10
Suspend CPC and SessionKeeper for unresponsive peers.

tomaszklak:
### Problem
When peer is not responsive we still send messages to it from CrossPingCheck which causes increased energy usage.

### Solution
Suspend CPC for unresponsive peers.
    
When the last handshake time for a given peer is above 180s we should suspend the CrossPingCheck tasks for that peer.

I've run the whole nat-lab with this feature turned on and it all passed. Separate MR will add nat-lab test(s) with this feature explicitly tested.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -247,6 +247,9 @@ pub struct Features {
     /// Flag to specify if keys should be validated
     #[serde(default)]
     pub validate_keys: FeatureValidateKeys,
+    /// Avoid sending periodic messages to peers with no traffic reported by wireguard
+    #[serde(default)]
+    pub skip_unresponsive_peers: bool,
 }
 
 impl FeaturePaths {
@@ -303,7 +306,7 @@ mod tests {
             "exit_dns": {}
         }"#;
 
-    const CORRECT_FEATURES_JSON_WITH_IS_TEST_ENV: &str = r#"
+    const CORRECT_FEATURES_JSON_WITH_ALL_OPTIONAL: &str = r#"
         {
             "wireguard":
             {
@@ -326,9 +329,24 @@ mod tests {
                 "priority": ["relay", "direct"],
                 "force": "relay"
             },
-            "direct": {},
-            "exit_dns": {},
-            "is_test_env": true
+            "direct": 
+            {
+                "providers": 42,
+                "endpoint_interval_secs": 10   
+            },
+            "exit_dns":
+            {
+                "auto_switch_dns_ips": true
+            },
+            "is_test_env": true,
+            "derp":
+            {
+                "tcp_keepalive": 1,
+                "derp_keepalive": 2,
+                "enable_polling": true
+            },
+            "validate_keys": false,
+            "skip_unresponsive_peers": true
         }"#;
 
     static EXPECTED_FEATURES: Lazy<Features> = Lazy::new(|| Features {
@@ -357,15 +375,20 @@ mod tests {
         }),
         direct: Some(FeatureDirect {
             providers: None,
-            endpoint_interval_secs: None,
+            endpoint_interval_secs: Some(10),
         }),
         exit_dns: Some(FeatureExitDns {
-            auto_switch_dns_ips: None,
+            auto_switch_dns_ips: Some(true),
         }),
         #[cfg(any(target_os = "macos", feature = "pretend_to_be_macos"))]
         is_test_env: Some(true),
-        derp: None,
-        validate_keys: Default::default(),
+        derp: Some(FeatureDerp {
+            tcp_keepalive: Some(1),
+            derp_keepalive: Some(2),
+            enable_polling: Some(true),
+        }),
+        validate_keys: FeatureValidateKeys(false),
+        skip_unresponsive_peers: true,
     });
 
     static EXPECTED_FEATURES_WITHOUT_TEST_ENV: Lazy<Features> = Lazy::new(|| Features {
@@ -403,6 +426,7 @@ mod tests {
         is_test_env: None,
         derp: None,
         validate_keys: Default::default(),
+        skip_unresponsive_peers: Default::default(),
     });
 
     #[test]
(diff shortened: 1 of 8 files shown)


crates/telio-traversal/src/session_keeper.rs
```diff
@@ -219,14 +235,23 @@ impl Runtime for State {
             Ok((pk, action)) = self.keepalive_actions.select_action() => {
                 let pk = *pk;
 
-                telio_log_trace!("({}) name: \"{}\", action = actions.select_action()", Self::NAME, pk);
+                if let Some(last_handshake_time_provider) = &self.last_handshake_time_provider {
```
mathiaspeters, line 238: How come we use the old logic in case `last_handshake_time_provider` is `None`? Just trying to understand

tomaszklak: The `last_handshake_time_provider` is set in the device/mod.rs` based on the feature config flag `skip_unresponsive_peers`. If it is set no `None` we should not skip such peers and always send ping, otherwise we should check respective last handshake time and decide based on that if we should send the ping.

tomaszklak: Your are right, ping sending logic should have been duplicated in the outer else branch.

tomaszklak: Fixed the code and expanded the relevant test to cover both old and new way of working.

mathiaspeters:
+1

Jauler:
Oh, I just realized. I think SessionKeeper was already tied to succesfully established direct connection, by adding a node here:
* https://github.com/NordSecurity/libtelio/blob/LLT-4152-disable-cpc-and-session-keeper/src/device/wg_controller.rs#L223
And removing here: 
* https://github.com/NordSecurity/libtelio/blob/LLT-4152-disable-cpc-and-session-keeper/src/device/wg_controller.rs#L189

Which makes the LLT-4152 task specification maybe not super strictly clear, as SessionKeeper is already disabled when device is not connected :thinking: .

crates/telio-traversal/src/last_handshake_time_provider.rs
```diff
@@ -0,0 +1,65 @@
...
+impl LastHandshakeTimeProvider for DynamicWg {
+    async fn last_handshake_time(&self, public_key: &PublicKey) -> Result<Duration, Error> {
+        let interface = self.get_interface().await?;
+        match interface
+            .peers
+            .get(public_key)
+            .map(|peer| peer.time_since_last_handshake)
+        {
+            Some(Some(t)) => Ok(t),
+            Some(None) => Err(Error::UnknownLastHandshakeTime(*public_key)),
```
Jauler, line 36: I would recommend using `Result<Option<Duration>, Error>` as a return type. It is possible, that no handshake occured at all (e.g. we have never been connected to the peer), thus the `Duration` would be `None` in that case. It seems like it would represent the situation more naturally, when compared to `Err(...)`.

crates/telio-traversal/src/last_handshake_time_provider.rs
```diff
@@ -0,0 +1,65 @@
...
+    last_handshake_time_provider: &dyn LastHandshakeTimeProvider,
+    public_key: &PublicKey,
+) -> bool {
+    match last_handshake_time_provider
+        .last_handshake_time(public_key)
+        .await
+    {
+        Ok(time) => time < MAX_PEER_HANDSHAKE_TIMEOUT,
+        // If we don't know the handshake time, let's assume that peer is alive
+        Err(Error::UnknownLastHandshakeTime(_)) => true,
```
Jauler, line 52: If there was no handshake established at all it does mean, that we are not connected, therefore I think we should treat such a peer as dead.

crates/telio-traversal/src/last_handshake_time_provider.rs
tomaszklak: done

crates/telio-traversal/src/last_handshake_time_provider.rs
tomaszklak: done

crates/telio-model/src/api_config.rs
```diff
@@ -247,6 +247,9 @@ pub struct Features {
     /// Flag to specify if keys should be validated
     #[serde(default)]
     pub validate_keys: FeatureValidateKeys,
+    /// Avoid sending periodic messages to peers with no traffic reported by wireguard
+    #[serde(default)]
+    pub skip_unresponsive_peers: bool,
```
Jauler, line 252: Damn, missed this yesterday. 

Can we move this `bool` into "FeatureDirect" as CPC is fully dependent on direct support. In a way it is nice, that this bool becomes "unsettable" if direct connections are disabled :).

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -808,17 +855,27 @@ impl<E: Backoff> EndpointConnectivityCheckState<E> {
...
-                if duration_in_state > self.exponential_backoff.get_backoff() {
+                if self
+                    .should_resend_call_me_maybe_request(duration_in_state)
+                    .await
+                {
                     self.exponential_backoff.next_backoff();
                     self.send_call_me_maybe_request(session, intercoms).await?;
                     do_state_transition!(m, SendCallMeMaybeRequest, self);
+                } else {
+                    telio_log_debug!("Skipping unresponsive peer {}", self.public_key);
```
Jauler, line 866: Could we change these logs to `trace`?

crates/telio-model/src/api_config.rs
tomaszklak: It is aleady moved in the branch with natlab tests: https://github.com/NordSecurity/libtelio/blob/LLT-4152-disable-cpc-and-session-keeper-natlab-test/crates/telio-model/src/api_config.rs#L181-L193 and we agreed just agreed that there is no need to backport it here. I will make it on by default on that branch also.

Jauler (approved):
+2.0

Jauler (approved):
+2.0

PR #162, opened 2023-10-10
Add mesh file-config <meshmap.json> command to tcli

Jauler:
Sometimes one needs to specify completely custom meshmap JSON to check out various conditions manually. While we do have `mesh config <meshmap JSON>` command it is not very convenient to use when doing manual tests. This PR adds aditional command which behaves in a exactly the same manner as `mesh config` behaves except for the fact that meshmap is read from file specified as an argument instead of required to specify the whole json as an argument to the command.


clis/tcli/src/cli.rs
```diff
@@ -239,6 +239,10 @@ enum MeshCmd {
         #[clap(default_value = "")]
         mesh_config: String,
     },
+    /// Read meshmap from file
+    FileConfig {
+        filename: String,
```
tomaszklak, line 244: nit: use `std::path::PathBuf`

tomaszklak (approved):
+2

tomaszklak (approved):

PR #164, opened 2023-10-12
[LLT-4124] Add IPv6 feature

stalowyjez:
### Problem
There is no feature flag for the introduced IPv6 changes which are meant to be disabled by default, but which should be tested with all of the new changes.

### Solution
Add IPv6 feature flag, so we will be able to use battle-tested IPv4 implementation on production, but also a possibility to test our new changes with newly introduced IPv6 support.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -288,6 +288,9 @@ pub struct Features {
     /// Flag to specify if keys should be validated
     #[serde(default)]
     pub validate_keys: FeatureValidateKeys,
+    /// IPv6 support
+    #[serde(default)]
+    pub ipv6: bool,
 }
 
 impl FeaturePaths {
@@ -388,7 +391,8 @@ mod tests {
                 "derp_keepalive": 2,
                 "enable_polling": true
             },
-            "validate_keys": false
+            "validate_keys": false,
+            "ipv6": true
         }"#;
 
     static EXPECTED_FEATURES: Lazy<Features> = Lazy::new(|| Features {
@@ -433,6 +437,7 @@ mod tests {
             enable_polling: Some(true),
         }),
         validate_keys: FeatureValidateKeys(false),
+        ipv6: true,
     });
 
     static EXPECTED_FEATURES_WITHOUT_TEST_ENV: Lazy<Features> = Lazy::new(|| Features {
@@ -471,6 +476,7 @@ mod tests {
         is_test_env: None,
         derp: None,
         validate_keys: Default::default(),
+        ipv6: false,
     });
 
     #[test]
@@ -638,6 +644,7 @@ mod tests {
             exit_dns: None,
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         let empty_qos_features = Features {
@@ -660,6 +667,7 @@ mod tests {
             exit_dns: None,
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         let no_qos_features = Features {
@@ -677,6 +685,7 @@ mod tests {
             exit_dns: None,
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         assert_eq!(from_str::<Features>(full_json).unwrap(), full_features);
@@ -713,6 +722,7 @@ mod tests {
             }),
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         let empty_features = Features {
@@ -726,6 +736,7 @@ mod tests {
             }),
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         assert_eq!(from_str::<Features>(full_json).unwrap(), full_features);
@@ -763,6 +774,7 @@ mod tests {
             direct: None,
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         assert_eq!(Features::default(), expected_defaults);
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -698,7 +698,7 @@ impl Runtime {
         features: Features,
         protect: Option<Protect>,
     ) -> Result<Self> {
-        let firewall = Arc::new(StatefullFirewall::new());
+        let firewall = Arc::new(StatefullFirewall::new(features.ipv6));
         let firewall_filter_inbound_packets = {
             let fw = firewall.clone();
             move |peer: &[u8; 32], packet: &[u8]| fw.process_inbound_packet(peer, packet)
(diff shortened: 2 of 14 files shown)


stalowyjez:
> I know it might sound a bit crazy, but anyways - shouldn't we make this flag as a double-bit flag? Looking into a future we _might_ disable IPv4 for certain user (because of small IPv4 range). So maybe this would make sense to add a double switch (`ipv4: bool, ipv6: bool`) ..?

That sounds to me a bit crazy indeed, @Jauler what do you think?

crates/telio-dns/src/forward.rs
```diff
@@ -32,21 +32,23 @@ use trust_dns_server::{
 use crate::bind_tun;
 
 #[derive(Clone, Copy)]
-pub struct TelioRuntime;
-impl RuntimeProvider for TelioRuntime {
+pub struct TelioRuntime<const USE_IPV6: bool>;
+impl<const USE_IPV6: bool> RuntimeProvider for TelioRuntime<USE_IPV6> {
```
dfetti, line 36: Do we really care if the forward authority uses IPv6 or not for resolving the DNS?

Jauler: There is this theoretical edge-case where forward server may potentially be IPv6. At some point we were working on (LLT-909) and one of the solutions was attempt to use another meshnet node's DNS server when routing traffic through it. This has not been released, and does not look like we are going to be releasing it any time soon. But at least theoretically in that case forward server could be IPv6. Now whether it makes sense to introduce IPv6 support in `forward.rs` for this edge-case, I do not know, but it seems like we already have some amount of it, so maybe removing is not really beneficial. What is more, looks like safe changes for me.

Jauler: Oh wait, but `forward.rs` had IPv6 support previously and it is already in prod. In that case I would agree with @dfetti that forward.rs does not need to be feature flagged here.

crates/telio-dns/src/nameserver.rs
```diff
@@ -53,15 +53,17 @@ pub trait NameServer {
 pub struct LocalNameServer {
```
Jauler, line 53: As `nameserver.rs` already have IPv6 support in prod, lets maybe skip the feature flag here.

crates/telio-dns/src/zone.rs
```diff
@@ -29,11 +29,11 @@ pub type Records = HashMap<String, Vec<IpAddr>>;
```
Jauler, line 29: As `zone.rs` already have IPv6 support in prod, lets maybe skip the feature flag here.

crates/telio-model/src/api_config.rs
```diff
@@ -763,6 +773,7 @@ mod tests {
             direct: None,
             derp: None,
             validate_keys: Default::default(),
+            ipv6: false,
         };
 
         assert_eq!(Features::default(), expected_defaults);
```
Jauler, line 779: Could we add one test where `ipv6: true`? :pray:

src/device/mod.rs
```diff
@@ -191,6 +191,9 @@ pub struct RequestedState {
 
     // Requested keepalive periods
     pub(crate) keepalive_periods: FeaturePersistentKeepalive,
+
+    // Indicates whether we want to use IPv6 addresses
+    pub use_ipv6: bool,
```
Jauler, line 196: I think this is duplicated, as you should have the same value in `self.features.ipv6`

src/device/mod.rs
```diff
@@ -336,13 +339,24 @@ struct Runtime {
 }
 
 impl Device {
+    fn check_ipv6_availability() -> bool {
```
Jauler, line 342: Oh there is no need to base this on the availability (fallback mechanisms are designed for that). Therefore it is enough to base it on `self.features.ipv6`

src/device/mod.rs
```diff
@@ -336,13 +339,24 @@ struct Runtime {
...
+
     // Incoming changes should fix
     #[allow(unwrap_check)]
     pub fn new<F: EventCb>(
-        features: Features,
+        mut features: Features,
         event_cb: F,
         protect: Option<Protect>,
     ) -> Result<Self> {
+        if features.ipv6 {
```
Jauler, line 356: There is no need to base this on the actual availability (fallback mechanisms are designed for that). Therefore it is enough to base it on `self.features.ipv6` instead of checking interface availability.

Also it is extremely likely, that when calling `device::new` the IP will not be set into the interface itself.

nat-lab/tests/test_mesh_exit_through_peer.py
```diff
@@ -91,13 +91,18 @@ async def test_mesh_exit_through_peer(
         )
 
         client_alpha = await exit_stack.enter_async_context(
-            telio.Client(connection_alpha, alpha, adapter_type).run_meshnet(
-                api.get_meshmap(alpha.id)
-            )
+            telio.Client(
```
Jauler, line 94: Do we really need these reformattings?

src/device/mod.rs
stalowyjez: Yes, but actually no. I.e. I need to pass this flag somehow to wg-controller - it seems to me that we don't pass features in any objects and adding additional argument is not an option because clippy won't let us additional arguments. Of course, we could consider wrapping wg function arguments with some structs, but I'm not sure if that's worth it.

crates/telio-dns/src/nameserver.rs
stalowyjez: Soo, generally I can remove changes from the whole `telio-dns` crate?

nat-lab/tests/test_mesh_exit_through_peer.py
stalowyjez: Just my python formatter just did it. I can undo them.

Jauler:
> > I know it might sound a bit crazy, but anyways - shouldn't we make this flag as a double-bit flag? Looking into a future we _might_ disable IPv4 for certain user (because of small IPv4 range). So maybe this would make sense to add a double switch (`ipv4: bool, ipv6: bool`) ..?
> 
> That sounds to me a bit crazy indeed, @Jauler what do you think?

Generally feature flags makes sense for optional features/functionality, but is kinda weird for mandatory things. In this case, IPv4 support is mandatory for out-of-tunnel communications (e.g. to the relaying servers, VPNs etc.) thus disabling it does not really make sense. For in-tunnel communication, disabling IPv4 does not really solve much, and requires quite a bit of effort. So IMHO, while the symmetry is nice, I do not think it is worth the effort.

nat-lab/tests/test_mesh_exit_through_peer.py
Jauler: :thinking: probably would be nice to minimze the `diff`, but this is more of a nitpick kind of thing, so up to you to decide ;)

crates/telio-dns/src/nameserver.rs
Jauler: As Far As I can see - yes

Jauler:
One request, will you be able to update confluence doc on FeatureConfig after merging this? :)

Jauler (dismissed):
+1.0

dfetti:
+1.0

dfetti (approved):

dfetti (approved):

PR #168, opened 2023-10-17
Stop stun and upnp when not needed

baltrus1:
### Problem
Stun and upnp providers should not send packets when there is no peer that wants to upgrade to direct connection. That happens when all peers are either offline or already directly connected.

### Solution
Pause Stun/Upnp providers in that case.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -192,14 +192,18 @@ pub struct FeatureDirect {
     /// Configuration options for skipping unresponsive peers
     #[serde(default = "FeatureDirect::default_skip_unresponsive_peers")]
     pub skip_unresponsive_peers: Option<FeatureSkipUnresponsivePeers>,
+    /// Parameters to optimize battery lifetime
+    #[serde(default)]
+    pub endpoint_providers_optimization: Option<FeatureEndpointProvidersOptimization>,
 }
 
 impl Default for FeatureDirect {
     fn default() -> Self {
         Self {
-            skip_unresponsive_peers: Self::default_skip_unresponsive_peers(),
             providers: Default::default(),
             endpoint_interval_secs: Default::default(),
+            skip_unresponsive_peers: Self::default_skip_unresponsive_peers(),
+            endpoint_providers_optimization: Default::default(),
         }
     }
 }
@@ -254,6 +258,24 @@ pub struct FeatureValidateKeys(pub bool);
 impl Default for FeatureValidateKeys {
     fn default() -> Self {
         Self(true)
+    }
+}
+
+/// Control which battery optimizations are turned on
+#[derive(Clone, Debug, Default, PartialEq, Eq, Deserialize)]
+pub struct FeatureEndpointProvidersOptimization {
+    /// Controls whether Stun endpoint provider should be turned off when there are no proxying peers
+    #[serde(default = "FeatureEndpointProvidersOptimization::default_optimization_variant")]
+    pub optimize_direct_upgrade_stun: bool,
+    /// Controls whether Upnp endpoint provider should be turned off when there are no proxying peers
+    #[serde(default = "FeatureEndpointProvidersOptimization::default_optimization_variant")]
+    pub optimize_direct_upgrade_upnp: bool,
+}
+
+impl FeatureEndpointProvidersOptimization {
+    /// Turning optimizations on by default if feature itself is enabled
+    pub fn default_optimization_variant() -> bool {
+        true
     }
 }
 
@@ -473,6 +495,7 @@ mod tests {
             skip_unresponsive_peers: Some(FeatureSkipUnresponsivePeers {
                 no_rx_threshold_secs: 50,
             }),
+            endpoint_providers_optimization: None,
         }),
         exit_dns: Some(FeatureExitDns {
             auto_switch_dns_ips: Some(true),
@@ -519,6 +542,7 @@ mod tests {
             providers: None,
             endpoint_interval_secs: None,
             skip_unresponsive_peers: Some(Default::default()),
+            endpoint_providers_optimization: None,
         }),
         exit_dns: Some(FeatureExitDns {
             auto_switch_dns_ips: None,
@@ -559,12 +583,14 @@ mod tests {
             skip_unresponsive_peers: Some(FeatureSkipUnresponsivePeers {
                 no_rx_threshold_secs: 42,
             }),
+            endpoint_providers_optimization: None,
         };
 
         let partial_features = FeatureDirect {
             providers: Some(vec![EndpointProvider::Local].into_iter().collect()),
             endpoint_interval_secs: None,
             skip_unresponsive_peers: Some(Default::default()),
+            endpoint_providers_optimization: None,
         };
 
         assert_eq!(from_str::<FeatureDirect>(full_json).unwrap(), full_features);
--- a/crates/telio-traversal/src/endpoint_providers/stun.rs
+++ b/crates/telio-traversal/src/endpoint_providers/stun.rs
@@ -25,6 +25,7 @@ use super::{EndpointCandidate, EndpointCandidatesChangeEvent, EndpointProvider, 
 
 pub type StunServer = telio_model::config::Server;
 
+const STUN_TIMEOUT_PAUSED: Duration = Duration::from_secs(600);
 #[cfg(not(test))]
 const STUN_TIMEOUT: Duration = Duration::from_secs(5);
 #[cfg(test)]
@@ -60,12 +61,14 @@ impl<Wg: WireGuard> StunEndpointProvider<Wg> {
         exponential_backoff_bounds: ExponentialBackoffBounds,
         ping_pong_handler: Arc<Mutex<PingPongHandler>>,
         stun_peer_publisher: chan::Tx<Option<StunServer>>,
+        is_battery_optimization_on: bool,
     ) -> Result<Self, Error> {
         Ok(Self::start_with_exp_backoff(
             wg,
             ExponentialBackoff::new(exponential_backoff_bounds)?,
             ping_pong_handler,
             stun_peer_publisher,
+            is_battery_optimization_on,
         ))
     }
 
@@ -85,6 +88,7 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
         exponential_backoff: E,
         ping_pong_handler: Arc<Mutex<PingPongHandler>>,
         stun_peer_publisher: chan::Tx<Option<StunServer>>,
+        is_battery_optimization_on: bool,
     ) -> Self {
         telio_log_info!("Starting stun endpoint provider");
         Self {
@@ -93,15 +97,16 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
                 current_server_index: 0,
                 current_proto: IpProto::IPv6,
                 wg,
+                ping_pong_tracker: ping_pong_handler,
                 change_event: None,
                 pong_event: None,
                 stun_session: None,
                 exponential_backoff,
                 current_timeout: PinnedSleep::new(STUN_TIMEOUT, ()),
                 last_candidates: Vec::new(),
-                ping_pong_tracker: ping_pong_handler,
-                stun_peer_publisher,
                 stun_state: StunState::WaitingForWg,
+                is_battery_optimization_on,
+                stun_peer_publisher,
                 sockets: None,
             }),
         }
@@ -228,6 +233,10 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
                 }
             }
 
+            if s.stun_state == StunState::Paused {
+                return Ok(());
+            }
+
             // Update STUN state
             s.exponential_backoff.reset();
             s.transition_to_wait_for_wg();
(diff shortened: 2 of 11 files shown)


gytsto (dismissed):
+1

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -229,6 +238,32 @@ impl<Wg: WireGuard, E: Backoff + 'static> EndpointProvider for StunEndpointProvi
...
+            }
+            s.stun_state = StunState::Paused;
+            s.current_timeout = PinnedSleep::new(STUN_TIMEOUT_PAUSED, ());
+            Ok(())
+        })
+        .await;
+    }
+
+    async fn unpause(&self) {
+        let _ = task_exec!(&self.task, async move |s| {
```
Jauler, line 258: I believe, we should reset exponential backoff here on `unpause` event

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -229,6 +238,32 @@ impl<Wg: WireGuard, E: Backoff + 'static> EndpointProvider for StunEndpointProvi
...
+                s.transition_to_wait_for_wg();
+                s.try_transition_to_searching_for_server().await;
+            }
+            Ok(())
+        })
+        .await;
+    }
 }
 
 //                                -------------
```
Jauler, line 269: Could you update the state diagram with "paused" state :)?

crates/telio-model/src/api_config.rs
```diff
@@ -240,6 +244,17 @@ impl Default for FeatureValidateKeys {
     }
 }
 
+/// Control which battery optimizations are turned on
+#[derive(Clone, Debug, Default, PartialEq, Eq, Deserialize)]
+pub struct FeatureBatteryOptimization {
+    /// Controls whether Stun endpoint provider should be turned off when there are no proxying peers
+    #[serde(default)]
+    pub optimize_direct_upgrade_stun: Option<bool>,
```
Jauler, line 252: We should enable these optimizations by default I believe :)

crates/telio-model/src/api_config.rs
```diff
@@ -240,6 +244,17 @@ impl Default for FeatureValidateKeys {
...
+    /// Controls whether Upnp endpoint provider should be turned off when there are no proxying peers
+    #[serde(default)]
+    pub optimize_direct_upgrade_upnp: Option<bool>,
```
Jauler, line 255: Same as https://github.com/NordSecurity/libtelio/pull/168/files#r1370007726

Jauler:
Even though you marked the box, I do not believe that there are sufficient tests for this functionality (e.g. unit or integration)

src/device/mod.rs
```diff
@@ -912,6 +912,15 @@ impl Runtime {
                     },
                     ping_pong_tracker.clone(),
                     stun_server_events.tx,
+                    true,
+                    // features
```
packgron, line 916: leftovers?

src/device/mod.rs
```diff
@@ -935,6 +944,15 @@ impl Runtime {
                         maximal: Some(Duration::from_secs(120)),
                     },
                     ping_pong_tracker.clone(),
+                    true,
+                    // features
```
packgron, line 948: leftovers?

src/device/wg_controller.rs
```diff
@@ -583,13 +603,35 @@ async fn build_requested_meshnet_peers_list<
                     // If peer is offline according to derp, we turn off keepalives.
                     None
                 } else {
+                    // Side effect: if peer is proxying but not marked offline by derp,
+                    // means it might want to upgrade to direct connection
+                    is_any_peer_proxying = true;
```
Jauler, line 608: I think we should have two conditions:
* Peer is proxying, just like you did here and peer has active WG session (last handshake < 180s)

Additionally I would recommend to move this pause/unpause logic to `consolidate_wg_peers()` logic, as generally `build_requested_peers_list()` might not be really a expected place to do actual state mutations, while `consolidate_wg_peers()` is designed to do exactly that :).

baltrus1:
> Even though you marked the box, I do not believe that there are sufficient tests for this functionality (e.g. unit or integration)

I wanted to add integration test but a test that does exactly what I would want already exists(test_direct_working_paths_with_skip_unresponsive_peers) - it creates connection, removes remote peer, and after coming online checks if direct connection is made okay(because this optimization might not allow direct connection to be made if it is buggy). So I just enabled feature for all direct tests basically, I think that should be enough, I will still have to measure battery impact manually

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -551,14 +589,15 @@ impl<Wg: WireGuard, E: Backoff> Runtime for State<Wg, E> {
...
-                            // This is a poll interval. We're still in HasEndpoints state.
+                            // This is a poll interval. We're still in HasEndpoints state
                             let res = self.start_stun_session().await;
                             if let Err(err) = res {
                                 telio_log_error!("Starting STUN session failed with error: {:?}", err);
                                 self.transition_to_backing_off_state().await;
                             }
                         }
                     },
+                    StunState::Paused => (),
```
Jauler, line 600: In paused state this will form infinite loop

Jauler: Much better place to handle "pause" state would be: https://github.com/NordSecurity/libtelio/pull/168/files#diff-f52b9bdf9a9389d5ebb2456346414b982fc5fd532c9cf08e23284a430ff06003R555

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -229,6 +238,33 @@ impl<Wg: WireGuard, E: Backoff + 'static> EndpointProvider for StunEndpointProvi
...
         .unwrap_or(None)
     }
+
+    async fn pause(&self) {
+        let _ = task_exec!(&self.task, async move |s| {
+            if !s.is_battery_optimization_on {
+                return Ok(());
+            }
+            if s.stun_state != StunState::Paused {
+                s.servers = Vec::new();
```
Jauler, line 248: I believe that after resetting servers here - will not allow us to restore them on unpause, meaning that stun endpoint provider will be left stuck here: https://github.com/NordSecurity/libtelio/pull/168/files#diff-f52b9bdf9a9389d5ebb2456346414b982fc5fd532c9cf08e23284a430ff06003R555

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -557,11 +582,13 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> Runtime for State<Wg, I, E> {
                 let _ = self.handle_ping_rx(&buff[..len], &addr).await;
             }
             _ = &mut self.upnp_interval => {
-                if self.check_endpoint_candidate().await.is_ok(){
-                    self.upnp_interval =  PinnedSleep::new(self.exponential_backoff.get_backoff(), ());
-                } else {
-                    self.exponential_backoff.next_backoff();
-                };
+                if !self.is_endpoint_provider_paused {
```
Jauler, line 585: Can instead of polling we do similar thing like STUN Endpoint provider has, where it simply sleeps indefinately in a `tokio_select()`. It has `update` and `pending()`  arms where one is sleeping indefinately, and other processes `task_exec!`'s. Because pause/unpause is `task_exec()`. This would allow us to avoid unnecessary polling.

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -255,6 +258,8 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> UpnpEndpointProvider<Wg, I, E
                 epc_event_tx: None,
                 exponential_backoff,
                 upnp_interval: PinnedSleep::new(UPNP_INTERVAL, ()),
+                is_battery_optimization_on,
+                is_endpoint_provider_paused: false,
```
Jauler, line 262: Could we start UPnP provider in paused mode?

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -93,15 +97,16 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
...
+                ping_pong_tracker: ping_pong_handler,
                 change_event: None,
                 pong_event: None,
                 stun_session: None,
                 exponential_backoff,
                 current_timeout: PinnedSleep::new(STUN_TIMEOUT, ()),
                 last_candidates: Vec::new(),
-                ping_pong_tracker: ping_pong_handler,
-                stun_peer_publisher,
                 stun_state: StunState::WaitingForWg,
```
Jauler, line 107: Can we start STUN endpoint provider in paused mode?

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -547,24 +573,38 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> Runtime for State<Wg, I, E> {
...
                 let buff = self.rx_buff.clone();
                 let _ = self.handle_ping_rx(&buff[..len], &addr).await;
             }
             _ = &mut self.upnp_interval => {
-                if self.check_endpoint_candidate().await.is_ok(){
-                    self.upnp_interval =  PinnedSleep::new(self.exponential_backoff.get_backoff(), ());
-                } else {
-                    self.exponential_backoff.next_backoff();
-                };
+                if !self.is_endpoint_provider_paused {
```
Jauler, line 598: This branch is technically unreachabpe here due to the above if now. While functionally this is correct, I am thinking, maybe we should remove the `if !self.is_endpoint_provider_paused` if to make the select logic slightly simpler?

src/device/wg_controller.rs
```diff
@@ -247,6 +257,34 @@ async fn consolidate_wg_peers<
...
+            actual_peer.time_since_last_handshake
+        );
+        if is_actual_peer_proxying && actual_peer.state() == Connected {
+            is_any_peer_eligible_for_upgrade = true;
+        }
+    }
+
+    let ep_control = |ep: Arc<dyn EndpointProvider>| async move {
+        if is_any_peer_eligible_for_upgrade {
+            telio_log_debug!("Unpausing {} provider", ep.name());
```
Jauler, line 274: Could you make these logs `Info` level?

Jauler (commented):
Could we add integration and/or unit tests covering optimization? 

E.g. Ensuring, that STUN requests are not performed if there is no peers eligible for upgrade?

src/device/wg_controller.rs
baltrus1: it is spamming every 5 seconds, not sure if this is needed.

src/device/wg_controller.rs
```diff
@@ -247,6 +257,34 @@ async fn consolidate_wg_peers<
...
+            ep.unpause().await;
+        } else {
+            telio_log_debug!("Pausing {} provider", ep.name());
```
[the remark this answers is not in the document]
baltrus1: same as above

baltrus1:
> Could we add integration and/or unit tests covering optimization?
> 
> E.g. Ensuring, that STUN/UPnP requests are not performed if there is no peers eligible for upgrade?

How would we check if these requests are made or not? tcpdump feels kinda hacky and might introduce flakyness

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
baltrus1: right I forgot the state is replaced and not updated right away

crates/telio-model/src/api_config.rs
```diff
@@ -189,14 +189,18 @@ pub struct FeatureDirect {
     /// Configuration options for skipping unresponsive peers
     #[serde(default = "FeatureDirect::default_skip_unresponsive_peers")]
     pub skip_unresponsive_peers: Option<FeatureSkipUnresponsivePeers>,
+    /// Parameters to optimize battery lifetime
+    #[serde(default)]
+    pub battery_optimization: Option<FeatureBatteryOptimization>,
```
Jauler, line 194: Could we name this value slightly more specific? As it will be in a somewhat "public" interface.

For example, `endpoint_providers_usage_optimization`.

baltrus1: renamed to endpoint_providers_optimization

nat-lab/tests/test_direct_connection.py
```diff
@@ -663,3 +669,85 @@ async def test_infinite_stun_loop(setup_params: List[SetupParameters]) -> None:
...
+            ):
+                stun_enabled = True
+
+            if (
+                param.features.direct.providers is not None
+                and "upnp" in param.features.direct.providers
+            ):
+                upnp_enabled = True
+
+            param.features.wireguard = Wireguard(proxying=5, direct=5)
```
Jauler, line 701: :thinking: is there a reason to change default keepalive periods?

Jauler (changes requested):
+1.0 if the lints are fixed :P

Jauler (approved):
+1.0

PR #175, opened 2023-10-20
Llt 3950 wg stun ipv6 support

stalowyjez:
### Problem
Our wg-stun implementation does not support IPv6.

### Solution
Add support for IPv6 in the wg-stun implementation, bump libtelio-build to version which uses Nordderper supporting that feature as well.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/stun.rs
+++ b/crates/telio-traversal/src/endpoint_providers/stun.rs
@@ -1,3 +1,4 @@
+use std::net::{Ipv4Addr, Ipv6Addr};
 use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use std::ops::Deref;
@@ -8,6 +9,7 @@ use stun_codec::TransactionId;
 use telio_crypto::PublicKey;
 use telio_model::config::Server;
 use telio_proto::{Session, WGPort};
+use telio_sockets::SocketPool;
 use telio_sockets::{native::AsNativeSocket, External};
 use telio_task::{io::chan, task_exec, BoxAction, Runtime, Task};
 use telio_utils::{
@@ -44,16 +46,12 @@ impl<Wg: WireGuard> StunEndpointProvider<Wg> {
     /// - `exponential_backoff_bounds` - Config for exponential backoff
     ///   controlling intervals between consequtive queries to the STUN server
     pub fn start(
-        tun_socket: UdpSocket,
-        ext_socket: External<UdpSocket>,
         wg: Arc<Wg>,
         exponential_backoff_bounds: ExponentialBackoffBounds,
         ping_pong_handler: Arc<Mutex<PingPongHandler>>,
         stun_peer_publisher: chan::Tx<Option<StunServer>>,
     ) -> Result<Self, Error> {
         Ok(Self::start_with_exp_backoff(
-            tun_socket,
-            ext_socket,
             wg,
             ExponentialBackoff::new(exponential_backoff_bounds)?,
             ping_pong_handler,
@@ -73,25 +71,16 @@ impl<Wg: WireGuard> StunEndpointProvider<Wg> {
 
 impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
     fn start_with_exp_backoff(
-        tun_socket: UdpSocket,
-        ext_socket: External<UdpSocket>,
         wg: Arc<Wg>,
         exponential_backoff: E,
         ping_pong_handler: Arc<Mutex<PingPongHandler>>,
         stun_peer_publisher: chan::Tx<Option<StunServer>>,
     ) -> Self {
         telio_log_info!("Starting stun endpoint provider");
-        telio_log_debug!(
-            "tun_socket: {:?}, ext_socket: {}",
-            tun_socket.as_native_socket(),
-            ext_socket.deref().as_native_socket()
-        );
         Self {
             task: Task::start(State {
                 servers: vec![],
                 current_server_index: 0,
-                tun_socket,
-                ext_socket,
                 wg,
                 change_event: None,
                 pong_event: None,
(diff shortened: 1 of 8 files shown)


crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -357,9 +360,22 @@ impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
                 .get(&server.public_key)
                 .ok_or(Error::NoStunPeer)?;
 
+            let use_ipv6 = self
```
packgron, line 363: Ideally then we have both ipv6 and ipv4, we should try to do stun request first with ipv6, and if that fails we should try with ipv4. I like this approach as it's simpler, but I think that we cannot truly relay that ipv6 will always work. @Jauler please comment

stalowyjez: But how should it behave? Should it change server, how it should cooperate with exponential backoff?

Jauler: generally the retry logic can be based on two things:
* IPv6 should be preferred if it is available, due to conflict-free nature of it.
* We should not assume, that IPv6 is bullet-proof, bug-free or other impossible things :D 


So my recommendation would be:
1. Try IPv6 if it is enabled
1. If IPv6 attempt failed or it is disabled -> attempt IPv4
1. If IPv4 attempt failed, increment exponential backoff time.
1. Wait for exponential backoff to pass
1. Go to step 1

This way we still have exponential backoff increasing and increasing time between retries. We do attempt to use both IPv6 and IPv4 (meaning that in case we have some bug preventing one from working, but allowing other -> meshnet would still be operational), and we prefer IPv6 due to its conflict-free nature :).

Jauler: Ok, after a bit more discussions with @stalowyjez I realized that the approach above requires rather non-trivial amount of refactoring for WG-STUN module, which may be more risky, than then risk of WG-STUN servers not responding via IPv6 inside of the tunnel. What is more, there seems to be growing need for refactoring IPv4/IPv6 fallback logic into a common place, such that different modules would behave the same in terms of selecting IP protocol to use. But that is a task for another Jira issue and pull-request

packgron (dismissed):
+2.0

Jauler (dismissed):
+1.0

crates/telio-traversal/src/endpoint_providers/stun.rs
Jauler: After some more discussions with @stalowyjez I just realized the problem. The STUN EndpointProvider is instantiated at the `start()` method. And in that context we do not yet know whether IPv6 address has been actually assigned to this meshnet node by the API. Meaning, that selecting IPv6 in that situation would be unsafe.

src/device/mod.rs
```diff
@@ -1306,8 +1304,36 @@ impl Runtime {
             // Refresh the lists of servers for STUN endpoint provider
             if let Some(direct) = self.entities.direct.as_ref() {
                 if let Some(stun_ep) = direct.stun_endpoint_provider.as_ref() {
+                    let use_ipv6 = self.features.ipv6 && {
```
Jauler, line 1307: This causes on problem.

Now on every call to `telio_set_config()` new sockets will be allocated for `STUN`. This means, that new source port is allocated for STUN. While it is less relevant for WG-STUN requests, as those get encapsulated to the same WG tunnel anyways, the plaintext STUN requests will change its port (and thus `xor-mapped-address` attribute on every `telio-config` call. IMHO, changing ports here may cause problems for direct connections.

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -102,12 +89,22 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
...
+    pub async fn configure(&self, mut servers: Vec<Server>, sockets: Option<StunSockets>) {
         let _ = task_exec!(&self.task, async move |s| {
+            if let Some(stun_sockets) = sockets.as_ref() {
+                telio_log_debug!(
+                    "tun_socket: {:?}, ext_socket: {}",
+                    stun_sockets.tun_socket.as_native_socket(),
+                    stun_sockets.ext_socket.deref().as_native_socket()
+                );
+            }
+            s.sockets = sockets;
```
Jauler, line 106: The override of the sockets, means that any previous STUN session which we may have had are now invalid. Thus we should invalidate existing session and change the state of the STUN to the initial state. Just we must be careful not to do that on _every_ `configure()` call, but only on those, where we are actually transitioning from ipv6->ipv4 or vice-versa.

This applies to the sockets override as well. As new sockets will add new source port, meaning that new NAT.mapping will be created for that.

LukasPukenis (approved):
I see that this PR already has +3 so just approving

LukasPukenis (approved):
I see that this PR already has +3 so just approving

LukasPukenis (approved):
Seeing it's already +3, just approving

stalowyjez:
@LukasPukenis there were a few changes since these approvals, currently we need to fix one more issue with Windows tests, before this PR can be merged

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -295,17 +373,21 @@ impl<Wg: WireGuard, E: Backoff> State<Wg, E> {
...
-            .await
-            .send_ping(
-                addr,
-                self.get_wg_port(),
-                &self.ext_socket,
-                session_id,
-                public_key,
-            )
-            .await
+        if let Some(config) = self.sockets.as_ref() {
```
Jauler, line 376:
```suggestion
        if let Some(sockets) = self.sockets.as_ref() {
```

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -512,62 +618,62 @@ impl<Wg: WireGuard, E: Backoff> Runtime for State<Wg, E> {
 
         pin!(updated);
 
-        //If no STUN servers are configured -> nothing to do.
-        //NOTE: this will get cancelled on reconfiguration attempt
-        if self.servers.is_empty() {
+        if let (Some(sockets), false) = (self.sockets.as_ref(), self.servers.is_empty()) {
```
Jauler, line 621: This is a small style thing, but maybe it would be slightly easier to read if we would reverse the `if` condition:
```rust
        // We must ensure that sockets are created and that we have a populated
        // server list, before we can continue doing anything.
        let sockets = match (self.sockets.as_ref(), self.servers.is_empty()) {
            (Some(sockets), false) => {sockets},
            (_, _) => {
                //If no STUN servers or sockets are configured -> nothing to do.
                //NOTE: this will get cancelled on reconfiguration attempt
                tokio::select! {
                    _ = pending() => {},
                    update = &mut updated => {
                        return update(self).await;
                    }
                }
                return Ok(());
            }
        };
```


And then leave `tokio::select!` untouched? This would remove one indentation level for it. @stalowyjez What are you thoughts about this?

Jauler (approved):
+1.0

Jauler (dismissed):
Clean rebase on `main`

crates/telio-traversal/src/endpoint_providers/stun.rs
```diff
@@ -102,12 +91,74 @@ impl<Wg: WireGuard, E: Backoff> StunEndpointProvider<Wg, E> {
...
+                        .tun_socket
+                        .local_addr()
+                        .map(|addr| addr.is_ipv6() != use_ipv6)
+                        .ok()
+                })
+                .unwrap_or(false);
+
+            if s.sockets.is_none() || ip_version_changed {
+                s.sockets = match (
+                    if use_ipv6 {
```
packgron, line 131: NIT: i would like to see these moved to var's above match, It would personaly for me look less busy

packgron (approved):
+2.0, one small nit pick

crates/telio-traversal/src/endpoint_providers/stun.rs
Jauler: Updated

jjanowsk:
I can see that Michał forgot to update wireguard-nt-rust-wrapper version in telio-wg's Cargo.toml file. I've made a fix there to fix one of the tests that started failing after his changes in this MR. It was fixed in this PR: https://github.com/NordSecurity/wireguard-nt-rust-wrapper/pull/3 and the version should be bumped from v1.0.2 to v1.0.3. @Jauler Can you please update it?

Jauler:
> I can see that Michał forgot to update wireguard-nt-rust-wrapper version in telio-wg's Cargo.toml file. I've made a fix there to fix one of the tests that started failing after his changes in this MR. It was fixed in this PR: [NordSecurity/wireguard-nt-rust-wrapper#3](https://github.com/NordSecurity/wireguard-nt-rust-wrapper/pull/3) and the version should be bumped from v1.0.2 to v1.0.3. @Jauler Can you please update it?

:heavy_check_mark: done

packgron (dismissed):

packgron (approved):

PR #178, opened 2023-10-23
Use OS certificates instead of hardcoded mozilla list

tomaszklak:
### Problem
Libtelio is using hardcoded list of root certificate authorities (from Mozilla) to verify https certificates when connecting to derp. This can cause out of date / insecure / revoked certificates to be used if we don't update the hardcoded list on time.

### Solution
Us operating system, built in mechanisms to validate certificates, since those are automatically (from our pov) updated when needed.

Tested on:

- [x] macos
- [x] ios
- [x] linux
- [x] windows
- [x] android  

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v0.4.8
+          triggered-ref: v0.4.9
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,7 +7,7 @@ libtelio-build-pipeline:
     LIBTELIO_PROJECT_PATH: $CI_PROJECT_PATH
     LIBTELIO_DEFAULT_BRANCH: $CI_DEFAULT_BRANCH
 
-    # Don't checkout libtelio submodule when using GIT_SUBMODULE_STRATEGY.
+    # Don't checkout libtelio submodule when using GIT_SUBMODULE_STRATEGY. 
     # This ensures that all jobs that use `libtelio` submodule are calling
     # `ci/checkout.sh`, which uses `LIBTELIO_COMMIT_SHA` to
     # checkout the correct libtelio REF.
@@ -15,5 +15,5 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v0.4.8
+    branch: v0.4.9
     strategy: depend
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,11 @@ telio-task.workspace = true
 telio-traversal.workspace = true
 telio-utils.workspace = true
 telio-wg.workspace = true
+once_cell.workspace = true
+
+[target.'cfg(target_os = "android")'.dependencies]
+jni = "0.19"
+rustls-platform-verifier = { git = "https://github.com/tomaszklak/rustls-platform-verifier.git", rev = "1eeed2dc3a4a7f437220875feb31e50cdec0bf07" }
 
 [dev-dependencies]
 slog-async = "2.7"
--- a/linkers/dld
+++ b/linkers/dld
@@ -7,7 +7,7 @@ LINKER=${LINKER#"d-"}
 ARGS=$@
 
 while (( "$#" )); do
-	if [[ $1 == *"-Wl,--version-script="* ]]; then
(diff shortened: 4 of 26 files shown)


.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v0.4.8
+          triggered-ref: rust-platform-verifier
```
tomaszklak, line 34: This will need to be changed to version with https://bucket.digitalarsenal.net/low-level-hacks/vpn/client/libtelio-build/-/merge_requests/156 merged

.gitlab-ci.yml
```diff
@@ -7,13 +7,13 @@ libtelio-build-pipeline:
...
+    # Don't checkout libtelio submodule when using GIT_SUBMODULE_STRATEGY. 
     # This ensures that all jobs that use `libtelio` submodule are calling
     # `ci/checkout.sh`, which uses `LIBTELIO_COMMIT_SHA` to
     # checkout the correct libtelio REF.
     GIT_SUBMODULE_PATHS: ':(exclude)libtelio'
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v0.4.8
+    branch: rust-platform-verifier
```
tomaszklak, line 18: This will need to be changed to version with https://bucket.digitalarsenal.net/low-level-hacks/vpn/client/libtelio-build/-/merge_requests/156 merged

Cargo.toml
```diff
@@ -57,6 +57,10 @@ telio-traversal.workspace = true
 telio-utils.workspace = true
 telio-wg.workspace = true
 
+[target.'cfg(target_os = "android")'.dependencies]
+jni = "0.19"
+rustls-platform-verifier = { git = "https://github.com/tomaszklak/rustls-platform-verifier.git", rev = "1eeed2dc3a4a7f437220875feb31e50cdec0bf07" }
```
tomaszklak, line 62: fork of rustls-platform-verifier is used, since original doesn't yet officially support tvos

linkers/dld
```diff
@@ -7,7 +7,7 @@ LINKER=${LINKER#"d-"}
 ARGS=$@
 
 while (( "$#" )); do
-	if [[ $1 == *"-Wl,--version-script="* ]]; then
+	if [[ $1 == *"-Wl,--version-script="* && ! $ARGS =~ -o.*librustls_platform_verifier.* ]]; then
```
tomaszklak, line 10: This is to make sure that what is happening in the body of 'if' is only executed for object files from libtelio, and not for those from librustls_platform_verifier.

src/ffi/mod.rs
```diff
@@ -151,6 +151,27 @@ fn deserialize_features(features: *const c_char) -> Result<Features, telio_resul
...
+/// Initialize OS certificate store, should be called only once. Without call to telio_init_cert_store
+/// telio will not be able to verify https certificates in the system certificate store.
+/// # Params
+/// - `env`:    see https://developer.android.com/training/articles/perf-jni#javavm-and-jnienv
+/// - `ctx`:    see https://developer.android.com/reference/android/content/Context
+pub extern "C" fn telio_init_cert_store(env: *mut jni::sys::JNIEnv, ctx: jni::sys::jobject) {
+    static INIT: Once = Once::new();
+    INIT.call_once(|| match unsafe { jni::JNIEnv::from_raw(env) } {
+        Ok(env) => {
+            if let Err(err) = rustls_platform_verifier::android::init_hosted(&env, ctx.into()) {
```
matszczygiel, line 165: Do you think reporting the error with a status code or something like that makes sense?
It seems the `init_hosted()` already does the `Once` logic for you

matszczygiel (dismissed):
+1, with a minor nitpick

src/ffi/mod.rs
tomaszklak: > Do you think reporting the error with a status code or something like that makes sense?

Good idea - I've added a fallback feature flag for hardcoded list so, when this fails application can actually try to 'recover'.

> It seems the init_hosted() already does the Once logic for you

That's an implementation detail of `init_hosted` which is not guaranteed to be there in future. On the other hand readme says:

> This only needs to be called once and the verifier will be valid for the lifetime of your app's process.

Which doesn't *require* one call but clearly states that only one call is *needed*, and I think it's better to make sure on our side that we don't make unnecessary calls.

matszczygiel: Yeah, makes sense to guard from multiple calls on our side :+1:

Regarding the error code, it would be nice to have it plus maybe an error for the case when this function is called more than once. However, this will complicate the `Once` logic a bit, so IDK, your call.

packgron (approved):
+1

matszczygiel (approved):

packgron (approved):

PR #181, opened 2023-10-23
[LLT-4432] Fix iOS linking issue

stalowyjez:
### Problem
There are “undefined refernce to kSCNetworkInterfaceTypeIrDA” errors when one tries to build iOS application with libtelio v4.1.1

### Solution
We need to use a bit modified version (our fork) of system-configuration crate


Jauler (approved):
+2.0

PR #185, opened 2023-10-24
Add custom peer names to meshnet config

lcruz99:
### Problem
We are able to read received configuration and we have input fields to prepare magic DNS configuration.
Configuration fields may be not presented, and this scenario needs to be supported by code.

### Solution
Add meshnet config support to custom peer names.

- [x] README.md is updated
- [x] changelog.md is updated


packgron (changes requested):
I think it would be a good idea to expand mesh_api.py also

packgron (approved):
+2.0

PR #186, opened 2023-10-24
Add a note that bug bounty only applies to nordsec owned repos

Jauler:
In a lot of forks now there is a nordsecurity bug bounty links in SECURITY.md file, therefore adding a note for future forks to have that bug bounty is only applicable if the vulnerability being reported is present in code repositories owned by NordSecurity as well.


LukasPukenis (approved):

packgron:
+2

packgron (approved):

PR #189, opened 2023-10-25
[LLT-4427] Fix exit node events

stalowyjez:
Make sure that exit node events are marked as exit node ones.

### Problem
When a node stops begin exit node, an event is presented 

### Solution
Just base "is_exit" field in event on the ip_addresses, not the existence of the ExitNode in the requested state.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -1478,28 +1478,29 @@ impl Runtime {
         Ok(self.entities.socket_pool.clone())
     }
 
-    async fn peer_to_node(&self, peer: &uapi::Peer, state: Option<PeerState>) -> Option<Node> {
+    async fn peer_to_node<'a>(
+        &'a self,
+        peer: &uapi::Peer,
+        state: Option<PeerState>,
+    ) -> Option<Node> {
         let endpoint = peer.endpoint;
 
         // Check if ExitNode in requested state matches the node being reported by public
         // key
-        let exit_node = if let Some(exit_node) = self.requested_state.exit_node.clone() {
-            Some(exit_node)
-        } else {
-            self.requested_state.last_exit_node.clone()
-        }
-        .filter(|node| node.public_key == peer.public_key);
+        let exit_node = self.requested_state.exit_node.as_ref();
+        let exit_node = exit_node
+            .or(self.requested_state.last_exit_node.as_ref())
+            .filter(|node| node.public_key == peer.public_key);
 
         // Find a peer with matching public key in meshnet_config and retrieve the needed
         // information about it from there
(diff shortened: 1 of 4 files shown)


src/device/mod.rs
```diff
@@ -1492,7 +1492,7 @@ impl Runtime {
             .meshnet_config
             .as_ref()
             .and_then(|config| config.peers.clone())
-            .and_then(|config_peers| {
```
matszczygiel, line 1495: This whole logic can be simplified for sure, possibly without cloning. However, since it's not the goal of this PR I'm not pressing on this one

matszczygiel:
+1

src/device/mod.rs
stalowyjez: Eh, unfortunately we have a lot of stuff like this in our code. It would be nice to merge it without rerunning CI, as it is extra fragile in the last days (due to 500 errors which we receive from some external servers). But if I need to introduce some other changes anyway I will be happy to fix it as well.

stalowyjez: @matszczygiel @LukasPukenis tried to simplify this function a bit, please check if you like the result.

stalowyjez: Well, and here are server errors: https://github.com/NordSecurity/libtelio/actions/runs/6773518206/job/18408574733

src/device/mod.rs
```diff
@@ -1487,64 +1487,30 @@ impl Runtime {
...
-                        .collect::<Vec<Peer>>()
-                        .first()
-                        .cloned()
-                }),
+        let get_config_peer = |config: Option<&Config>| {
+            config
+                .and_then(|cfg| cfg.peers.as_ref())?
+                .iter()
+                .find(|&p| p.base.public_key == peer.public_key)
+                .cloned()
```
matszczygiel, line 1495: Do you need it cloned?

matszczygiel:
Nice, since @LukasPukenis probably forgot to review and the change is pretty small, +2

src/device/mod.rs
Jauler: Just would be super nice to do them in a separate PR ;). As it makes reviewers life slightly harder.

stalowyjez: @Jauler sure, as a rule of course, but as reviewer requested them here, I just added them, the change is still quite small (and split it to separate commits to make reviewer life easier 😁 )

Jauler (dismissed):
+1.0

I have requested in-branch testing from windows team, to check if this resolves their issue. So lets wait for their response :).

Jauler (approved):
+2.0

PR #196, opened 2023-11-07
LLT-4280: Kill existing connections on mobile platform upon silent reconnect (TCP)

matszczygiel:
### Problem
Upon changing the VPN server for the TCP connection the packet originating from the remote peer fails to reach the client.
This is happening because the client's public IP address has changed.

### Solution
The solution is to force the client to reconnect to the remote peer. In order to achieve that we need to fake the TCP RST packet and inject it into the tunnel on behalf of the remote peer. This resets the connection on the client's end and forces it to reconnect.

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-firewall/src/firewall.rs
```diff
@@ -1653,22 +1768,22 @@ pub mod tests {
...
             assert_eq!(fw.process_outbound_packet(&make_peer(), &make_tcp(us, them, TcpFlags::FIN)), true);
             assert_eq!(fw.tcp.lock().unwrap().len(), 1);
 
             // update tcp cache entry timeout
             assert_eq!(fw.tcp.lock().unwrap().get(&conn_key), Some(&TcpConnectionInfo{
-                tx_alive: false, rx_alive: false, conn_remote_initiated: false
+                tx_alive: false, rx_alive: false, conn_remote_initiated: false, remote_seq: 0
             }));
 
             // process inbound packet (should not update ttl, because not ACK, but entry should still exist)
```
Jauler, line 1789: Could you add some unit tests covering `reset_connections()` function? It is much easier to cover various cases in FW via unit-tests then integration.

crates/telio-firewall/src/firewall.rs
```diff
@@ -371,26 +384,32 @@ impl StatefullFirewall {
...
-                    "Outbound IP packet is for whitelisted peer, forwarding: {:?}",
-                    ip
-                );
-                return true;
+        let proto = ip.get_next_level_protocol();
+
+        // whitelist read-lock scope
+        let whitelist = unwrap_lock_or_return!(self.whitelist.read(), false);
+
+        // Fasttrack, if peer is whitelisted - skip any conntrack and allow immediately
```
Jauler, line 392: A small suggestion for updating the comment
```suggestion
        // If peer is whitelisted - allow immediately
```

crates/telio-firewall/src/firewall.rs
```diff
@@ -371,26 +384,32 @@ impl StatefullFirewall {
...
+        if whitelist.peer_whitelist.contains(&peer) {
+            telio_log_trace!(
+                "Outbound IP packet is for whitelisted peer, forwarding: {:?}",
+                ip
+            );
+
+            // We still need to track the state of TCP connections
+            if matches!(proto, IpNextHeaderProtocols::Tcp) {
```
Jauler, line 400: Due to the fact that this call will be acquiring some mutexes it should only be done if feature flag has enabled this killing of the connections.

crates/telio-firewall/src/firewall.rs
```diff
@@ -371,26 +384,32 @@ impl StatefullFirewall {
...
-                    ip
-                );
-                return true;
+        let proto = ip.get_next_level_protocol();
+
+        // whitelist read-lock scope
+        let whitelist = unwrap_lock_or_return!(self.whitelist.read(), false);
+
+        // Fasttrack, if peer is whitelisted - skip any conntrack and allow immediately
+        if whitelist.peer_whitelist.contains(&peer) {
```
Jauler, line 393:
```suggestion
if !feature_flag_enabled && whitelist.peer_whitelist.contains(&peer) {
    telio_log_trace!("Outbound IP packet is for whitelisted peer, forwarding: {:?}", ip);
    return true;
}

if !ip.check_valid() {
    telio_log_trace!("Outbound IP packet is not valid, dropping: {:?}", ip);
    return false;
}

match proto {
    IpNextHeaderProtocols::Udp => {
        self.handle_outbound_udp(&ip);
    }
    IpNextHeaderProtocols::Tcp => {
        self.handle_outbound_tcp(&ip);
    }
.............................
};

telio_log_trace!("Accepting packet {:?} {:?}", ip, peer);
true            
```

crates/telio-firewall/src/firewall.rs
```diff
@@ -425,6 +446,12 @@ impl StatefullFirewall {
                 "Inbound IP packet is for whitelisted peer, forwarding: {:?}",
                 ip
             );
+
+            // We still need to track the state of TCP connections
+            if matches!(proto, IpNextHeaderProtocols::Tcp) {
```
Jauler, line 451: This handle_inbound_tcp function should only be called for whitelisted peers if connection killing is enabled in feature configs, as this would allow as to not pay its performance penalty for platforms where connection killing is not required.

crates/telio-firewall/src/firewall.rs
matszczygiel: Yeah, if we're going to make firewall aware of the feature flag then sure

crates/telio-firewall/src/firewall.rs
```diff
@@ -696,7 +732,8 @@ impl StatefullFirewall {
             }
         };
         let flags = tcp_packet.get_flags();
-        Some((key, flags))
+        let seq = tcp_packet.get_sequence();
```
Jauler, line 735: Correct me if I am wrong, but I think this is not strictly correct, as the sequence number used in the TCP RST packet must be should be something like this:
```
TCP_RST_seq_num = Last_TCP_pkt_seq_num + TCP_payload_length
```

And the payload when sending ACK may very well be larger than 0 or 1.

crates/telio-firewall/src/firewall.rs
matszczygiel: Actually, the thing is that you still need to allow packets for whitelisted peers even when the `check_valid()` fails. At least that it was before my changes

matszczygiel: Anyway, I can simply not call `handle_tcp_packet()` when the feature flag is off

crates/telio-wg/src/adapter/mod.rs
```diff
@@ -190,6 +197,7 @@ pub(crate) fn start(
     socket_pool: Arc<SocketPool>,
     firewall_process_inbound_callback: FirewallCb,
     firewall_process_outbound_callback: FirewallCb,
+    firewall_icmp_kill_conns_callback: FirewallResetConnsCb,
```
Jauler, line 200: As we are not using ICMP only for killing of connections, maybe it would be better to:
```suggestion
    firewall_kill_conns_callback: FirewallResetConnsCb,
```

matszczygiel: Oh, yeah. It's a leftover

src/device/mod.rs
```diff
@@ -1376,7 +1389,14 @@ impl Runtime {
         }
 
         self.requested_state.exit_node = Some(exit_node);
-        wg_controller::consolidate_wg_state(&self.requested_state, &self.entities).await
+        wg_controller::consolidate_wg_state(&self.requested_state, &self.entities).await?;
+
+        self.entities
```
Jauler, line 1394: In this case we can probably attempt reseting existing connections only in reconnection state. 

Which would mean, that we should take `self.requested_state.exit_node` before assignment in line 1391, and only if it is `Some(...)` do the call to reset_existing_connections.

crates/telio-firewall/src/firewall.rs
matszczygiel: Yeah, you're right. It was overly simplified by me. Let me correct this one

crates/telio-firewall/src/firewall.rs
```diff
@@ -950,6 +989,82 @@ impl Firewall for StatefullFirewall {
...
+                MutableIpv4Packet::new(&mut ipv4buf).expect("IPv4 buffer should not be too small");
+            ipv4pkg.set_version(4);
+            ipv4pkg.set_header_length(5);
+            ipv4pkg.set_total_length((IPV4_LEN + TCP_LEN) as _);
+            ipv4pkg.set_flags(Ipv4Flags::DontFragment);
+            ipv4pkg.set_ttl(0xFF);
+            ipv4pkg.set_next_level_protocol(IpNextHeaderProtocols::Tcp);
+
+            let mut ipv4pkgbuf = [0u8; IPV4_LEN + TCP_LEN];
+            for (key, val) in tcp.iter() {
```
Jauler, line 1027: I realized that we missed one thing in the RFC LLT-0042, when killing the connections, we should not be killing all of the connections, but only the ones which were created over `exit_node` peer.

The clearest manifestation of this would be that when user is using meshnet and VPN at the same time, reconnecting from one VPN server to another - should not kill meshnet connections. IMHO, attaching public key of the peer to the connection state or event implementing per-peer conntrack as suggested in LLT-4502 would resolve this issue naturally. But when killing the connections, peer's public key will have to be passed all the way from src/device/mod.rs :/

nat-lab/tests/test_vpn.py
```diff
@@ -219,3 +223,143 @@ async def test_vpn_reconnect(
...
+                        testing.unpack_optional(alpha.get_ip_address(ip_proto)),
+                        serv_ip,
+                        str(80),
+                    ]
+                ).run(stdout_callback=on_stdout, stderr_callback=on_stdout)
+            )
+            await testing.wait_normal(sender_start_event.wait())
+
+            await testing.wait_long(
+                client.disconnect_from_vpn(
```
Jauler, line 353: AFAIK, you should not need to disconnect from the existing VPN client, just issue new connect call.

Jauler (commented):
Very nice work!

In general I think we need to address the comment: https://github.com/NordSecurity/libtelio/pull/196/files#r1386351949 here before we can request in-branch testing from android team.

crates/telio-firewall/src/firewall.rs
matszczygiel: done

crates/telio-firewall/src/firewall.rs
matszczygiel: done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
```diff
@@ -950,6 +989,82 @@ impl Firewall for StatefullFirewall {
...
     }
+
+    fn reset_connections(
+        &self,
+        sink4: &mut dyn io::Write,
+        _: &mut dyn io::Write,
+    ) -> io::Result<()> {
+        telio_log_debug!("Constructing connetion reset packets");
+
+        if let Ok(mut tcp) = self.tcp.lock() {
```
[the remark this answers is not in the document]
matszczygiel: Done

crates/telio-wg/src/adapter/mod.rs
matszczygiel: Done

src/device/mod.rs
matszczygiel: I agree with checking if there's an old node. However, the flag check should be internal to the adapter itself IMHO.

nat-lab/tests/test_vpn.py
matszczygiel: Actually python test framework fails without disconnecting first because it sets up interfaces

src/device/mod.rs
matszczygiel: I've added the check if there is a current exit node or if there is a `last_exit_node` set up in case `disconnect_exit_node()` was called beforehand.
However, I've omitted the feature flag check

crates/telio-firewall/src/firewall.rs
matszczygiel: I've added the fix and expanded the test case for the client which uses an already synchronized sequence number so that  both paths are tested

crates/telio-firewall/src/firewall.rs
LukasPukenis: Does this work other way around as well? Like when user starts routing traffic through another node, it should not kill VPN connections I suppose as well. If `exit_node` is reused for meshnet and for VPN I think the problem is solved in one place

crates/telio-firewall/src/firewall.rs
matszczygiel: Unit test added

crates/telio-firewall/src/firewall.rs
Jauler: Generally routing through meshnet exit node and routing through VPN server has the same problem, so the resolution would be the same. 

You want to kill connections which has been established via the old exit node (whether it is VPN or meshnet exit node) and to avoid touching connections to/via other peers.

Jauler: This diagram shows "more full" picture.

We want to kill only dashed connections:
Stale-TCP-conns-when-VPN+meshnet

matszczygiel: @Jauler I've pushed the relevant changes, only for TCP though as this is the scope of the PR

crates/telio-wg/src/adapter/boring.rs
```diff
@@ -81,4 +86,49 @@ impl Adapter for BoringTun {
...
+                Ok(())
+            }
+        }
+
+        let dev = self.device.read().await;
+        let dev = dev.device.read();
+        let tun = dev.iface();
+
+        let mut sink4 = Sink4 { tun };
+        let mut sink6 = Sink6 { tun };
```
matislovas, line 130: Not sure, if this would change something, but maybe we should we limit IPv6 socket reset with IPv6 feature flag?

matszczygiel: Do you mean guarding IPv6 resets under the feature flag? Honestly, I doubt it would be useful, the IPv6 isn't implemented yet and probably never will

matislovas: The IPv6 is already implemented and ipv6-through-exit-node is still under consideration.

matszczygiel: I meant this feature for IPv6, not the IPv6 as a whole

matislovas:
+0.5 nice implementation

src/device/mod.rs
```diff
@@ -1422,9 +1440,23 @@ impl Runtime {
             return Err(Error::EndpointNotProvided);
         }
 
-        self.requested_state.exit_node = Some(exit_node);
+        let old_exit_node = self.requested_state.exit_node.replace(exit_node);
         wg_controller::consolidate_wg_state(&self.requested_state, &self.entities, &self.features)
-            .await
+            .await?;
+
+        let old_exit_pubkey = old_exit_node
```
Jauler, line 1447: I think in this situation we will be killing connections not just on reconnect, but also on disconnect.

E.g. if `exit_node` is None, but `self.requested_state.last_exit_node` is `Some(...)` then we still call `.reset_existing_connections(...)`. In super practical terms it would mean, that on android the killing will happen twice, once per OS mechanism, and once in ours.

But IMHO, this may be ok todo for "extra safety". I will request Dragan's opinion too :D

crates/telio-firewall/src/firewall.rs
```diff
@@ -460,18 +500,23 @@ impl StatefullFirewall {
...
         if flags & TCP_FIRST_PKT_MASK == TcpFlags::SYN {
-            telio_log_trace!("Inserting TCP conntrack entry {:?}", key);
+            telio_log_trace!("Inserting TCP conntrack entry {:?}", key.link);
             tcp_cache.insert(
                 key,
                 TcpConnectionInfo {
                     tx_alive: true,
                     rx_alive: true,
                     conn_remote_initiated: false,
+                    next_seq: 0,
```
Jauler, line 519: Due to the fact that in this specific moment, we do not know what is the actual `next_seq` number, maybe it would be better represented by Option<u32> with the value `None` in this situation? 

While it does not seem to be strictly speaking incorrect, having `TcpConnectionInfo` record in this stage, if `.reset_existing_connections()` gets called during this state, it  would mean, that TCP RST packet will be generated for not yet established connection with incorrect seq number (`0` in our case). I am thinking, that it would be a bit better if we would not trigger `TCP RST` in such case...

src/device/mod.rs
dragan-cecavac-nordsec: I'm curious in which cases should `exit_node` be None here?

On disconnect we currently stop telio, so it should retain `last_exit_node`, if I got it right.
But even if we change this eventually, it shouldn't be much of an overhead.

crates/telio-firewall/src/firewall.rs
```diff
@@ -285,6 +302,12 @@ impl Debug for IpConnWithPort {
     }
 }
 
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+struct TcpConn {
+    link: IpConnWithPort,
+    pubkey: PublicKey,
```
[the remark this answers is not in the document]
LukasPukenis: Instead of the comment, maybe it's better to extract into an enum?

src/device/mod.rs
matszczygiel: Ok, so I will wait for the decision

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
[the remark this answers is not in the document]
matszczygiel: Done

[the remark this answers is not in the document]
matszczygiel: @LukasPukenis I think it doesn't make sense to branch on the enum just for information purposes

Jauler (dismissed):
Amazing work! :rocket: :rocket: 

+1.5

nat-lab/tests/test_vpn.py
```diff
@@ -253,3 +257,147 @@ async def test_vpn_reconnect(
...
+                        str(80),
+                    ]
+                ).run(stdout_callback=on_stdout, stderr_callback=on_stdout)
+            )
+
+            await proc.wait_stdin_ready()
+            await asyncio.sleep(2.0)
+            # Without this sleep nc get EOF on stdin for some reason
+            await proc.write_stdin("GET")
+
```
LukasPukenis, line 386: non-functional: It is not necessary to push data through the TCP connection(as opposed to UDP) as it already did the 3-way handshake before it get's to the STDIN part at which point we already have all the building blocks for RST packet.

The existing code checks for more common case which is great but I felt the comment might be a bit misleading

LukasPukenis: Adding the comment on the first client as well would be an improvement. Though all of this is minor you can feel free to ignore it

LukasPukenis (approved):

Jauler (approved):
+1.0

PR #199, opened 2023-11-08
Changes for running nat-lab from macos host

tomaszklak:
Changes:

1. python-wireguard hardcodes opening of `*.so` file which obviously fails on macos. To fix that alternative, pure python implementation of the key derivation is provided.
2. `conntrack` cli is not available on macos, so on macos conntrack python wrapper always return no out of limits
3. the official architecture name of arm on macos is `arm64` but the corresponding name in the rust world is `aarch64` so some translation is added.

This results in most of the tests passing, example end result:
`47 failed, 177 passed, 1 skipped, 181 deselected, 15 xfailed, 36 xpassed`

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/mac_wg.py
+++ b/nat-lab/tests/mac_wg.py
@@ -1,0 +1,19 @@
+import codecs
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric.x25519 import X25519PrivateKey
+
+
+def key_pair():
+    private_key = X25519PrivateKey.generate()
+    bytes_ = private_key.private_bytes(
+        encoding=serialization.Encoding.Raw,
+        format=serialization.PrivateFormat.Raw,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+    sk = codecs.encode(bytes_, "base64").decode("utf8").strip()
+
+    pubkey = private_key.public_key().public_bytes(
+        encoding=serialization.Encoding.Raw, format=serialization.PublicFormat.Raw
+    )
+    pk = codecs.encode(pubkey, "base64").decode("utf8").strip()
+    return (sk, pk)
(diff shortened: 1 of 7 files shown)


lcruz99:
Is conntrack out of limits the reason for the 47 tests failures?

lcruz99:
Can you update the natlab docs and add an _howto_ for this platform?

tomaszklak:
> Is conntrack out of limits the reason for the 47 tests failures?
> 
> If so, maybe we could remove the assert on this platform

I don't think so. The limits thing is already disabled on mac here: https://github.com/NordSecurity/libtelio/pull/199/files#diff-84fc76d8957b236a1123ffd561b2efc50424f1a042ee6f3c371e1629261f214fR117

The vast majority of the failures (40 out of 47) are variants of `test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_client_side` which uses conntrack in a more involved way. Once this is in, I will create separate issues for all failing tests to investigate how to fix them on mac.

nat-lab/tests/mac_wg.py
```diff
@@ -0,0 +1,19 @@
+import codecs
```
gytsto, line 1: maybe this can be moved to utils and used instead of python_wireguard for all platforms, since only `Key` is used from it

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -113,6 +114,8 @@ async def execute(self) -> None:
         await self._process.execute(stdout_callback=self.on_stdout)
 
     def get_out_of_limits(self) -> Optional[Dict[str, int]]:
+        if platform.system() == "Darwin":
+            return None
```
gytsto, line 118: add simillar line to `execute` method, to prevent from even trying to start conntrack

nat-lab/tests/mac_wg.py
tomaszklak: I was thinking about that, but decided not to do this - I think that having real wireguard key generation being used on linux is valuable.

baltrus1 (approved):
+1

gytsto (approved):

PR #206, opened 2023-11-13
[Backport] lana: Add nordvpn app version to moose context

lcruz99:
### Problem

When receiving heartbeat events on analytics there is no way to filter/group/analyze those events by NordVPN application versions. Thus some of the dashboard filters cannot be implemented.

### Solution

Copy NordVPN application version context field from nordvpnapp tracker.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,17 @@
+### UNRELEASED
+---
+* LLT-4515: Add nordvpn app version to moose context
+
+<br>
+
+### v4.1.2
+### **Bruschetta**
+---
+### Changelog
+* LLT-4432: Use forked system-configuration crate to prevent iOS linking errors
+
+<br>
+
 ### v4.1.1
 ### **Bruschetta**
 ---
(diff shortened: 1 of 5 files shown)


LukasPukenis (commented):
I noticed that it's the same what we just reviewed, thus +1.5

changelog.md
```diff
@@ -1,3 +1,17 @@
+### UNRELEASED
```
Jauler, line 1: Could you add v4.1.3 to this, so that separate pipeline run would not be required just for version change in changelog?

lcruz99: My mistake, I've forgot this.

crates/telio-lana/src/event_log_moose.rs
```diff
@@ -98,109 +99,167 @@ pub fn init_context_info() {
...
+                {
+                    tracker_info
+                        .application
+                        .config
+                        .current_state
+                        .nordvpnapp_version = get_string_field(
+                        foreign_app_config_current_state_info,
+                        "nordvpnapp_version",
+                    );
+                    tracker_info
```
Jauler, line 199: None of the below values
* tracker_info.application.config.current_state.meshnet_enabled
* tracker_info.application.config.current_state.internal_meshnet

Will exist in `nordvpnapp` tracker, meaning that it will always fail to parse them, and thus probably not useful to attempt to parse them.

lcruz99: fixed.

Jauler (approved):
+1.0

PR #209, opened 2023-11-15
Add check to verify if the setup for nat-lab is correct

tomaszklak:
### Problem
It's possible to run nat-lab in an incompatible environment.

### Solution
Check if the version of outer environment matches the one which is expected (`triggered-ref` in github workflows).

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/run_local.py
+++ b/nat-lab/run_local.py
@@ -4,6 +4,7 @@ import argparse
 import os
 import subprocess
 import sys
+import time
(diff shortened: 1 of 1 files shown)


matislovas:
+2, if you promise, that it won't be transformed to `sys.exit(1)` on failed check :smile:

matislovas (approved):

PR #210, opened 2023-11-16
Draft: Add no-link detection mechanism

dfetti:
### Problem
A hinting mechanism for no-link detection is required to improve reaction time.

### Solution
Provide a link_state field in Node events consisting in a hint based on the time elapsed since last_rx_timestamp.


crates/telio-wg/src/wg.rs
```diff
@@ -528,9 +549,22 @@ impl State {
...
+                    .unwrap_or_default();
+
                 #[allow(clippy::collapsible_if)]
-                if !old.is_same_event(new) || old_state != new_state {
-                    if !try_send(new_state, new.clone()).await {
+                if !old.is_same_event(new)
+                    || old_state != new_state
+                    || old_link_state != new_link_state
+                {
+                    if !try_send(new_state, new_link_state, new.clone()).await {
```
LukasPukenis, line 567: I think we should only ever use `link_state` when we consider the peer as `Connected`. This way it should reduce the confusion on API usage part(what does it mean that Link is `up` but State is `down`) for example, as it might happen.

I suggest wrapping `link_state` into an Option and have it as `None` if `State != Connected`.

It should then also only be emitted when the peer is considered as connected, thus not confusing the API user. This should work sort of automatically as a side effect of what's written above :)

crates/telio-wg/src/wg.rs
```diff
@@ -635,33 +674,6 @@ impl State {
                 .await
                 .map(|r| r.errno == 0)
                 .unwrap_or_default();
```
LukasPukenis, line 676: In the original logic `push` dictates if it pushes uapi request or syncs everything. The current way will do the second part regardless which is change in behavior

crates/telio-wg/src/wg.rs
```diff
@@ -470,30 +477,36 @@ impl State {
...
+                    peer,
+                }))
                 .map(|r| r.is_ok())
         };
 
         // Notify all disconnects
         for key in &diff_keys.delete_keys {
             if !try_send(
                 PeerState::Disconnected,
+                LinkState::Down,
```
LukasPukenis, line 509: I think we should have an `Option` with `None` in this case as we should only care about the `LinkState` when we're connected

crates/telio-wg/src/wg.rs
```diff
@@ -717,8 +729,83 @@ impl State {
...
+            self.additional_peer_info = new_additional_peer_info;
+        }
+
         success
     }
+
+    fn compute_new_additional_peer_info(
+        &self,
+        to: &uapi::Interface,
+    ) -> HashMap<PublicKey, AdditionalPeerInfo> {
```
LukasPukenis, line 743: `AdditionalPeerInfo` is correct that it's additional info, but it only takes part of the state, for example I would expect to find `last_endpoint_change` in there as well in that case.

My personal take would be to add `last_link_state` or something similar nearby as it was and not wrap it into this additional structure :). Or wrap it fully.

LukasPukenis (commented):
A good draft on the idea, it might even work out as-is. I left few comments about how I think to make it better.

It's also obviously missing:
- feature flag
- any sort of test. The test should be fairly simple to devise as we can just disconnect the peer and expect within keepalive+rtt to receive event with link=down

crates/telio-wg/src/uapi.rs
```diff
@@ -249,6 +251,17 @@ pub struct AnalyticsEvent {
     pub timestamp: Instant,
 }
 
+/// TODO ADD DOCS
+#[derive(Clone)]
+pub struct AdditionalPeerInfo {
+    /// TODO ADD DOCS
+    pub last_rx_timestamp: tokio::time::Instant,
```
matszczygiel, line 258: Why tokio's `Instant`?

crates/telio-model/src/api_config.rs
```diff
@@ -268,6 +268,13 @@ where
     Ok(Some(eps))
 }
 
+/// Turns on the no link detection mechanism
+#[derive(Default, Clone, Debug, PartialEq, Eq, Deserialize)]
+pub struct FeatureNoLinkDetection {
+    /// Configurable rtt
+    pub rtt: Option<u64>,
```
matszczygiel, line 275: It's quite confusing, to be honest. Can you use `Duration`?

crates/telio-model/src/mesh.rs
```diff
@@ -90,6 +92,17 @@ pub enum NodeState {
     Connected,
 }
 
+/// This is a hint state computed based on the last_rx_timestamp
+#[derive(Copy, Clone, Default, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LinkState {
+    /// Link is Down
+    #[default]
```
matszczygiel, line 100: Does it make sense to provide the default?

crates/telio-wg/src/wg.rs
```diff
@@ -30,6 +31,9 @@ use crate::{
 
 use std::{collections::HashSet, future::Future, io, sync::Arc, time::Duration};
 
+const WG_KEEPALIVE_SECONDS: u64 = 10;
+const DEFAULT_CONFIGURABLE_RTT: u64 = 15;
```
matszczygiel, line 35: `Duration`?

crates/telio-wg/src/wg.rs
```diff
@@ -110,13 +114,18 @@ struct State {
...
     analytics_tx: Option<mc_chan::Tx<Box<AnalyticsEvent>>>,
 
     // Detecting unexpected driver failures, such as a malicious removal
     // We won't be notified of any errors, but a periodic call to get_config_uapi() will return a Win32 error code != 0.
     uapi_failed_last_call: bool,
     uapi_fail_counter: i32,
+    /// Enable no link detection mechanism
+    pub enable_no_link_detection: bool,
+    /// Configurable rtt. Default value is 5 seconds.
+    pub configurable_rtt: Option<u64>,
```
matszczygiel, line 128: `Duration`?

crates/telio-wg/src/wg.rs
```diff
@@ -202,21 +211,45 @@ impl DynamicWg {
...
+    ///         None,
     ///     );
     /// }
     /// ```
-    pub fn start(io: Io, cfg: Config) -> Result<Self, Error>
+    pub fn start(
+        io: Io,
+        cfg: Config,
+        enable_no_link_detection: bool,
+        configurable_rtt: Option<u64>,
```
matszczygiel, line 223: `Duration`?

crates/telio-wg/src/wg.rs
```diff
@@ -202,21 +211,45 @@ impl DynamicWg {
...
+            configurable_rtt,
+        ));
     }
 
-    fn start_with(io: Io, adapter: Box<dyn Adapter>, #[cfg(unix)] cfg: Config) -> Self {
+    fn start_with(
+        io: Io,
+        adapter: Box<dyn Adapter>,
+        enable_no_link_detection: bool,
+        configurable_rtt: Option<u64>,
```
matszczygiel, line 250: Can we make it more type-safe? For example:
```rust
pub enum NoLinkDetection {
    Disabled,
    Enabled {
        rtt: Duration,
    }
}
```

crates/telio-wg/src/wg.rs
```diff
@@ -544,24 +585,78 @@ impl State {
...
+            if new_rx != old_rx {
+                if new_rx.is_some() {
+                    Some(LinkState::Up)
+                } else {
+                    None
+                }
+            } else {
+                let configurable_rtt = self.configurable_rtt.unwrap_or(DEFAULT_CONFIGURABLE_RTT);
+                self.last_rx_timestamp.get(peer).map(|ts| {
+                    let time_since_last_rx = Instant::now() - ts.0;
```
matszczygiel, line 617:
```suggestion
                    let time_since_last_rx = ts.0.elapsed();
```

crates/telio-wg/src/wg.rs
```diff
@@ -544,24 +585,78 @@ impl State {
...
+                let should_notify_link_state_change = |new_state, old, new: Option<LinkState>| {
+                    if !self.enable_no_link_detection {
+                        return false;
+                    }
+
+                    if new_state == PeerState::Connected && old != new && new.is_some() {
+                        return true;
+                    }
+
+                    false
```
matszczygiel, line 645:
```suggestion
                    new_state == PeerState::Connected && old != new && new.is_some()
```

crates/telio-wg/src/wg.rs
```diff
@@ -747,9 +818,56 @@ impl State {
...
+                        if self.last_rx_timestamp.remove(&peer).is_some() {
+                            telio_log_debug!("removed peer last rx timestamp {:?}", peer);
+                        }
+                    }
+                }
+            }
+
+            match self.last_rx_timestamp.get(&peer) {
+                Some((t, _)) => {
+                    let time_since_last_rx = (Instant::now() - *t).as_secs();
```
matszczygiel, line 854:
```suggestion
                    let time_since_last_rx = t.elapsed().as_secs();
```

crates/telio-model/src/api_config.rs
dfetti: Every duration in feature structures are defined like this and I would prefer to keep it as it is for consistency reasons. I converted it in Duration lower in the lib as you suggested.

matszczygiel: Oh, yeah. I haven't noticed it's being deserialized. You're right on this one :+1:

crates/telio-wg/src/wg.rs
```diff
@@ -544,24 +592,76 @@ impl State {
...
+
+            if new_rx != old_rx {
+                if new_rx.is_some() {
+                    Some(LinkState::Up)
+                } else {
+                    None
+                }
+            } else {
+                let configurable_rtt = match self.no_link_detection {
+                    NoLinkDetection::Disabled => DEFAULT_CONFIGURABLE_RTT,
```
matszczygiel, line 623: This branch is never executed because of `match!()` at the beginning of the closure. Can you merge botch matches and do the logic only when `NoLinkDetection` is `Enabled`?

crates/telio-model/src/api_config.rs
packgron: You should add _ms or unit in field name to make it clear to user.

changelog.md
```diff
@@ -47,6 +47,7 @@
 * LLT-4124: Add IPv6 feature flag
 * LLT-3950: Enable IPv6 for wg-stun.
 * LLT-4502: Implement ICMP, UDP and TCP conntrack tracking for each peer separately
+* LLT-4545: Add no-link detection mechanismm
```
LukasPukenis, line 50: double m typo :)

crates/telio-model/src/mesh.rs
```diff
@@ -90,6 +92,16 @@ pub enum NodeState {
...
 }
 
+/// This is a hint state computed based on the last_rx_timestamp
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LinkState {
+    /// Link is Down
+    Down,
+    /// Link is Up
+    Up,
```
LukasPukenis, line 102: nitpick: these are obvious because the names are good

crates/telio-wg/src/wg.rs
```diff
@@ -544,24 +592,76 @@ impl State {
...
 
+                let old_link_state = self.last_link_state.get(key).copied();
+                let new_link_state = current_link_state(key);
+
+                let should_notify_link_state_change = |new_state, old, new: Option<LinkState>| {
+                    if matches!(self.no_link_detection, NoLinkDetection::Disabled) {
+                        return false;
+                    }
+
+                    new_state == PeerState::Connected && old != new && new.is_some()
```
LukasPukenis, line 650: This is correct logic, however if everything is as we expect, `new` should always be `Some(T)` when `new_state == PeerState::Connected`. For this I suggest of removing this additional check and adding a warning if that's the case instead. This way we could catch that it's happening and fix it in the future.

crates/telio-wg/src/wg.rs
```diff
@@ -544,24 +592,76 @@ impl State {
...
         // Check for updates, and notify
         for key in &diff_keys.update_keys {
             if let (Some(old), Some(new)) = (from.peers.get(key), to.peers.get(key)) {
                 let old_state = old.state();
                 let new_state = new.state();
 
+                let old_link_state = self.last_link_state.get(key).copied();
+                let new_link_state = current_link_state(key);
+
+                let should_notify_link_state_change = |new_state, old, new: Option<LinkState>| {
```
LukasPukenis, line 645: The naming is not that clear, `old` and `new` make sense if the function would accept only two arguments, but now it's very ambiguous as it's `new_state` and then `new`. My suggestion would be to give those variables a proper name.

I can clearly see that `new_state` is `PeerState` down below but it's this extra effort which could be reduces and make it more pleasurable :)

crates/telio-wg/src/wg.rs
```diff
@@ -747,9 +823,57 @@ impl State {
...
+                            telio_log_debug!("removed peer last rx timestamp {:?}", peer);
+                        }
+                    }
+                }
+            }
+
+            match self.last_rx_timestamp.get(&peer) {
+                Some((t, _)) => {
+                    let configurable_rtt = match self.no_link_detection {
+                        NoLinkDetection::Disabled => DEFAULT_CONFIGURABLE_RTT,
```
LukasPukenis, line 860: Could you clarify a bit:

- `start_with()` function short-circuits when it encounters disabled link-state feature - here we provide a default value even if it's disabled and modify the state

As per my understanding we should choose to either do nothing about it if it's disabled, or provide a default value and always work.

The feature doesn't interfer with inner workings of anything else so it's pretty safe to have it always on to be fair, this way it could be even simpler.

crates/telio-model/src/mesh.rs
dfetti: I am forced to add documentation :(

crates/telio-wg/src/wg.rs
dfetti: start_with doesn't have anything to do with this. There I just save the no_link_detection config in the State.
The short-circuit happens only when we generate events. 
As you said maintaining the last_link_state even when the feature is disabled should not infer with anything, might actually be useful I think if we will want to enable it during runtime at some point. But now that you pointed it out and I'm thinking more about it, I think we shouldn't do that due to performance reasons. This update function from telio-wg is one of the most called function in the lib. What do you think? Should I leave it as it is? Or should I skip the maintaining of last_link_state when the feature is disabled?

matszczygiel:
LGTM :+1:

crates/telio-model/src/mesh.rs
LukasPukenis: oh..

crates/telio-wg/src/wg.rs
```diff
@@ -544,24 +594,78 @@ impl State {
...
+                let new_link_state = current_link_state(key);
+
+                let should_notify_link_state_change =
+                    |new_node_state, old_link_state, new_link_state: Option<LinkState>| {
+                        if matches!(self.no_link_detection, NoLinkDetection::Disabled) {
+                            return false;
+                        }
+
+                        if new_link_state.is_none() {
+                            telio_log_warn!("new_link_state is None and it shouldn't");
```
LukasPukenis, line 651: nit: there could be an early return in here

nat-lab/tests/test_events.py
```diff
@@ -679,3 +679,289 @@ async def test_event_content_meshnet_node_upgrade_direct(
...
+        # Kill beta node
+        await client_beta.stop_device()
+
+        # As we have configured the rtt to 30 reconds we should not receive any event
+        # in 10 (WG_KEEPALIVE) + 30 (configured_rtt) seconds
+        with pytest.raises(asyncio.TimeoutError):
+            await client_alpha.wait_for_link_state_event(
+                beta.public_key, [LinkState.Down], timeout=39
+            )
+
```
LukasPukenis, line 963: Just curious, can't we make the configured rtt to like 1 second and expect the event in ~11-12 seconds? Should yield quicker test time. It doesn't matter if the test runner is parallelized that much though.

LukasPukenis (dismissed):
+1 nice!

LukasPukenis (approved):

PR #212, opened 2023-11-16
LLT-4052: add natlab guidelines

gytsto:
### Descripton
Add guidelines/tips on how to write natlab tests so they would be less flaky or in general more stable.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/guidelines.md
+++ b/nat-lab/guidelines.md
@@ -1,0 +1,204 @@
+### Natlab Guidelines:
+
+1. **Use Parametrization:**
+   - Test across all operating systems when possible.
+   - Ensure parametrization for each node.
+
+   **Why:**
+   - Covers more areas.
+   - Better readability.
+   - Wider usage of a test.
+
+   **Example:**
+   ```
+   @pytest.mark.parametrize("alpha_node", [ConnectionTag.MAC_VM, ConnectionTag.WINDOWS_VM, ConnectionTag.DOCKER_CONE_CLIENT_1])
+   @pytest.mark.parametrize("beta_node", ConnectionTag.MAC_VM)
+   def test_params(alpha_node, beta_node):
+   ```
+
+2. **Assertions Clarity:**
+   - Assertions must be explicit and easily visible in the test.
+   - Avoid assertions in different tasks, keep it within test, raise an exception instead if needed.
+
+   **Why:**
+   - Easier to track which test asserted if assert is within a test.
+   - Easier to read stacktrace.
+
+3. **Start Event Checks Promptly:**
+   - Initiate event checks before the corresponding action begins.
+
+   **Why:**
+   - Starting to wait for event after action, might result in missing the event.
+
+   **Example**:
+   - DON'T:
+   ```
+      await connect_to_vpn()
+      # event might occur event before starting to wait
+      await wait_for_event_peer(State.Connected)
+   ```
+   - DO:
+   ```
+      async with asyncio_util.run_async_context(wait_for_event_peer(State.Connected)) as event:
+            await asyncio.wait_for(
+                asyncio.gather(
+                    *[
+                        connect_to_vpn()
+                        event,
+                    ]
+                ),
+                timeout
+            )
+   ```
+
+4. **Utilize Provided Helpers:**
+   - Employ provided helper functions to establish the test environment (helpers.py and etc.).
+
+5. **Timeouts for Potential Hangs:**
+   - Apply timeouts to actions susceptible to hanging.
+
+   **Example:**
+   - As previously mentioned, event might occur before starting to wait for it. To avoid natlab hanging, wrap coroutine or future to complete with timeout.
+   ```
+      await connect_to_vpn()
+      await asyncio.wait_for(wait_for_event_peer(State.Connected), timeout=5)
+   ```
+
+6. **VM/Docker Reversion:**
+   - Ensure any modifications to VM/Docker are reverted during test teardown.
+   - Tests should leave no residual configurations even in case of failure.
+
+   **Why:**
+   - Some tests tinker with iptables / routing tables and some other significant stuff, which might lead to failure of other tests if not properly cleaned.
+
+   **Example**:
+   - DON'T:
+   ```
+   def test_1():
+      # for the purpose of example, break_tcp_conn_to_host is not context manager and doesnt have a proper cleanup
+
+      await connect_to(DERP_1)
+
+      async with asyncio_util.run_async_context(wait_for_event_peer(State.Disconnected)) as event:
+            await asyncio.wait_for(
+                asyncio.gather(
+                    *[
+                        break_tcp_conn_to_host(DERP_1)
+                        event,
+                    ]
+                ),
+                timeout
+            )
+
+   # this test will fail, because previous test didn't cleanup properly
+   def test_2():
+      async with asyncio_util.run_async_context(wait_for_event_peer(State.Connected)) as event:
+            await asyncio.wait_for(
+                asyncio.gather(
+                    *[
+                        connect_to(DERP_1)
+                        event,
+                    ]
+                ),
+                timeout
+            )
+   ```
+
+   - DO:
+   ```
+   @contextmanager
+   def tmp_break_conn()
+      try:
+         break_tcp_conn_to_host(DERP_1)
+         yield
+      finally:
+         restore_tcp_conn_to_host(DERP_1)
+
+
+   def test():
+      await connect_to(DERP_1)
+      async with asyncio_util.run_async_context(wait_for_event_peer(State.Disconnected)) as event:
+         with tmp_break_conn():
+            await asyncio.wait_for(event, timeout)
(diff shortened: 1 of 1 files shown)


nat-lab/guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+   **Example:**
+   ```
+   @pytest.mark.parametrize("alpha_node", [ConnectionTag.MAC_VM, ConnectionTag.WINDOWS_VM, ConnectionTag.DOCKER_CONE_CLIENT_1])
+   @pytest.mark.parametrize("beta_node", ConnectionTag.MAC_VM)
+   def test_params(alpha_node, beta_node):
+   ```
+
+2. **Assertions Clarity:**
+   - Assertions must be explicit and easily visible in the test.
+   - Avoid assertions in different threads, keep it within test, raise an exception instead if needed.
```
tomaszklak, line 21: Correct me if I'm wrong but I think that 'thread' is not correct here - I think 'task' is the python name for what you mean here.

nat-lab/guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+                        connect_to_vpn()
+                        event,
+                    ]
+                ),
+                timeout
+            )
+   ```
+
+4. **Utilize Provided Helpers:**
+   - Employ provided helper functions to establish the test environment.
```
tomaszklak, line 55: It's not clear which helpers are those. If you are referring to those in `helpers.py` in addition to mentioning this file, I think it would be good to add documentation to each helper.

nat-lab/guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+
+5. **Timeouts for Potential Hangs:**
+   - Apply timeouts to actions susceptible to hanging.
```
tomaszklak, line 58: I don't think this is actionable - how do I decide if something is "susceptible to hanging"?

nat-lab/guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+      await connect_to_vpn()
+      await asyncio.wait_for(wait_for_event_peer(State.Connected), timeout=5)
+   ```
+
+6. **VM/Docker Reversion:**
+   - Ensure any modifications to VM/Docker are reverted during test teardown.
+   - Tests should leave no residual configurations even in case of failure.
+
+   **Why:**
+   - Some tests tinker with iptables / routing tables and some other significant stuff, which might lead to failure of other tests if not properly cleaned.
```
tomaszklak, line 72: I think that instead of making this a guideline, such modifications can be automatically reverted using python context managers (`with` statement).

nat-lab/guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+   ```
+
+8. **Comprehensive Node State Checking:**
+   - Examine the states of all related nodes, not just a single one.
+
+   **Why:**
+   - By checking only one node, it is not made sure, that other node will be the same.
+
+   **Example:**  
+   To check if two nodes are connected via meshnet PING is used.
```
tomaszklak, line 151: This could be automated with a function that takes both nodes and verifies ping in both directions instead of asking devs to remember to do this.

nat-lab/guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+   process = connection_alpha.create_process("nc -nluv -4 100.64.0.100 12345", on_stdout = output_notifier.handle_output())
+   output_notifier.notify_output(f"Bound on 100.64.0.100 12345", listening_start_event)
+   await process.execute()
+   await listening_start_event.wait()
+
+   # listening started for sure
+   ```
+
+10. **Distinguish Between State and Event:**
+    - Differentiate between state and event, considering where to appropriately apply each.
```
tomaszklak, line 205: I don't think this is clear how/when to apply this guideline.

tomaszklak (commented):
I think some of the identified guidelines could be good ideas for framework improvements.

nat-lab/guidelines.md
gytsto: true that, will change

nat-lab/guidelines.md
gytsto: At the moment, there is only `helpers.py`, but in the future there might appear more stuff, so didn't want to specify. And about documentation on `helpers.py` - will do in the future

nat-lab/guidelines.md
gytsto: You wouldn't know all of the places, but for those, like in the example, you can add timeouts.

nat-lab/guidelines.md
gytsto: It is already done with context managers, but I just wanted to mention that in guidelines, so that developers wouldn't forget that. Also it is not all about iptables/routing tables, I took them as an example. Other places might be changed and not restored properly

nat-lab/guidelines.md
gytsto: Ofc it could, but as if now, it is just a guideline. In the future, code quality will be improved accordingly and such function might appear.

nat-lab/guidelines.md
gytsto: This one yes, couldn't remember the reason why I wrote it at first :smile: Will remove it for now

nat-lab/guidelines.md
tomaszklak: Yeah, but the correct example you provided is not using context manager (so I assume it can permanently change the state if the exception is raised before call to `restore_tcp_conn_to_host`). I think it would be better to show in the example the best possible version :)

nat-lab/guidelines.md
tomaszklak: I understand that this is just an example but the positive example can be used to show the reader that the correct approach is to use higher level abstraction (like `verify_connectivity_between_nodes([alpha, beta])`) instead of teaching that it's fine to copy&paste snippets of code (even if they do what they should be doing).

nat-lab/guidelines.md
```diff
@@ -0,0 +1,212 @@
...
+   ```
+
+10. **Prefer Wrapper Classes:**
+    - Opt for wrapper classes when interacting with external tools .
+
+    **Why:**
+    - Improved encapsulation
+    - Maintainability
+    - Uniformity
+    - Multiple OS support
```
stalowyjez, line 211: "tools."

Also, you mean here our wrappers, like `stun` one? Asking because python offers a lot of wrappers for different things as well, but they aren't particularly helpful when running things on the different containter.

nat-lab/guidelines.md
stalowyjez: I think that some notice about `helpers.py` file would be helpful

nat-lab/guidelines.md
tomaszklak: Currently this is (imho) almost like "don't make mistakes". First of all we don't have per test timeout so you can't rely on overall test timeout to detect a failure (right, or am I mistaken?). Second, we often expect to reach certain states in specified time so adding a timeout can be used to verify that we are within the required time. Finally, adding timeout to commands that perform network transmissions (eg. nslookup) might also be a good idea since we expect networks to be sometimes slow/broken.

In general adding some sort of heuristic of at least few distinct examples will help people to understand when this is good idea.

nat-lab/guidelines.md
tomaszklak: Fine by me, just make sure that at least a ticket to track this is created :)

gytsto: LLT-4557

nat-lab/guidelines.md
gytsto: was talking about our own wrappers like ping.py; stun.py and etc. Updated doc

nat-lab/guidelines.md
gytsto: Updated

nat-lab/guidelines.md
gytsto: Updated

nat-lab/guidelines.md
gytsto: Also added note about `helpers.py`

gytsto:
> I think some of the identified guidelines could be good ideas for framework improvements.

Yes, And it will be improved accordingly in the future. I tried to apply all of these in #81 , but was not possible to do that with only one PR :smile:

nat-lab/guidelines.md
gytsto: 1. We have timeout
2. Maybe I should just rephrase `potential hangs` to `wanted state not being reached`

tomaszklak: I think that few distinct examples of situations when to add timeout would be useful, but maybe I'm overthinking this and this is fine as it is. Consider this a nitpick and feel free to skip it :)

tomaszklak (approved):
+1.0

tomaszklak:
+2.0

tomaszklak (approved):
+2.0

PR #214, opened 2023-11-16
LLT-4505: Kill existing connections on mobile platform upon silent reconnect (UDP)

matszczygiel:
### Problem
Upon changing the VPN server for the UDP connection the packet originating from the remote peer fails to reach the client.
This is happening because the client's public IP address has changed.

### Solution
The solution is to force the client to reconnect to the remote peer. In order to achieve that we need to fake the ICMP Destination Unreachable (Port Unreachable) packet and inject it into the tunnel on behalf of the remote peer. This resets the `read()` OS calls on the client's end and forces it to reconnect.

- [x] README.md is updated
- [x] changelog.md is updated


matszczygiel:
Blocked by PR #196

LukasPukenis (approved):

crates/telio-firewall/src/firewall.rs
```diff
@@ -476,29 +509,43 @@ impl StatefullFirewall {
...
+        match udp_cache.entry(key) {
+            Entry::Vacant(e) => {
+                let conninfo = UdpConnectionInfo {
+                    is_remote_initiated: false,
+                    last_out_pkg_headers: Some(headers),
+                };
+                telio_log_trace!("Inserting new UDP conntrack entry {:?}", e.key());
+                e.insert(conninfo);
+            }
+            Entry::Occupied(mut o) => o.get_mut().last_out_pkg_headers = Some(headers),
```
LukasPukenis, line 548: Question: how does the logic of whitelisting work..? My question is also targeted to the original code as it follows the same logic.

My confusion comes from the fact we set `is_remote_initiated` when we receive inbound UDP connection not in cache and then later we check for it and check if port is whitelisted, what is the reason for that?

As from my understanding we should do the port whitelisting on the initial request and it's presence in the cache indicates it already passed the firewall successfully.

crates/telio-firewall/src/firewall.rs
```diff
@@ -1118,6 +1170,101 @@ impl Firewall for StatefullFirewall {
             telio_log_error!("TCP cache poisoned");
         };
 
+        if let Ok(mut udp_conn_cache) = self.udp.lock() {
```
LukasPukenis, line 1173: Can we split `reset_connections()` into two functions for TCP and UDP? This should make the `reset_connections()` be the orchestractor of reset logic with each function doing only it's part, would be easier on the eyes

crates/telio-firewall/src/firewall.rs
```diff
@@ -1118,6 +1170,101 @@ impl Firewall for StatefullFirewall {
...
+            #[allow(index_access_check)]
+            let mut ipv4pkg = MutableIpv4Packet::new(&mut ipv4pkgbuf[..IPV4_HEADER_LEN])
+                .expect("IPv4 buffer should not be too small");
+            ipv4pkg.set_version(4);
+            ipv4pkg.set_header_length(5);
+            ipv4pkg.set_total_length((IPV4_HEADER_LEN + ICMP_HEADER_LEN + ICMP_DATA_LEN) as _);
+            ipv4pkg.set_flags(Ipv4Flags::DontFragment);
+            ipv4pkg.set_ttl(0xFF);
+            ipv4pkg.set_next_level_protocol(IpNextHeaderProtocols::Icmp);
+            ipv4pkg.set_source(StdIpv4Addr::LOCALHOST);
```
LukasPukenis, line 1196: This should be the tunnel address instead of localhost to make it clear that meshnet's interface emited the stuff

crates/telio-firewall/src/firewall.rs
```diff
@@ -1118,6 +1170,101 @@ impl Firewall for StatefullFirewall {
...
+                        ipv4pkg.set_destination(dst);
+                        ipv4pkg.set_source(src);
+                        ipv4pkg.set_checksum(0);
+                        ipv4pkg.set_checksum(pnet_packet::ipv4::checksum(&ipv4pkg.to_immutable()));
+                        drop(ipv4pkg);
+
+                        telio_log_debug!("Injecting IPv4 ICMP (for UDP) packet {key:#?}");
+                        sink4.write_all(&ipv4pkgbuf)?;
+                    }
+                    (IpAddr::Ipv6(_), IpAddr::Ipv6(_)) => (), // TODO(msz): implement this pice when IPv4 will be fully supported
```
LukasPukenis, line 1258:
```suggestion
                    (IpAddr::Ipv6(_), IpAddr::Ipv6(_)) => (), // TODO(msz): implement this piece when IPv4 will be fully supported
```

crates/telio-firewall/src/firewall.rs
```diff
@@ -1118,6 +1170,101 @@ impl Firewall for StatefullFirewall {
...
+                    _ => telio_log_warn!(
+                        "Local and remote IP addrs version missmatch, this should never happen"
```
LukasPukenis, line 1260:
```suggestion
                        "Local and remote IP addrs version mismatch, this should never happen"
```

crates/telio-firewall/src/firewall.rs
```diff
@@ -2886,4 +3036,62 @@ pub mod tests {
...
+
+            fn flush(&mut self) -> io::Result<()> {
+                Ok(())
+            }
+        }
+
+        let mut sink4 = Sink4::default();
+        fw.reset_connections(&PublicKey(peer), None, &mut sink4, &mut std::io::sink()); // We do not test IPv6 yet
+
+        assert_eq!(sink4.pkgs.len(), 1); //3 packets in random order
```
LukasPukenis, line 3069: the comment says 3 packets but the code asserts for 1. Comment looks stale :)

crates/telio-firewall/src/firewall.rs
```diff
@@ -2886,4 +3036,62 @@ pub mod tests {
         assert_eq!(tcppkgs[2].get_destination(), 12345);
         assert_eq!(tcppkgs[2].get_sequence(), 1);
     }
+
+    #[test]
+    fn firewall_udp_conns_reset() {
```
LukasPukenis, line 3041: This unit test :man_cook: :kiss:

LukasPukenis (commented):
Overall I really like the solution, it has everything. Left several comments.

crates/telio-firewall/src/firewall.rs
```diff
@@ -240,12 +243,18 @@ pub struct StatefullFirewall {
...
 
 #[derive(Debug)]
 struct UdpConnectionInfo {
     is_remote_initiated: bool,
+    last_out_pkg_headers: Option<[u8; Self::LAST_PKG_HEADER]>,
+}
+
+impl UdpConnectionInfo {
+    /// UDP packet are always at least of this size
+    const LAST_PKG_HEADER: usize = 28;
```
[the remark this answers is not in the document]
matszczygiel: This is the size of the packet chunk included in the IMCP payload, it must be at least 28 bytes. The coincidence is that the UDP packet is always at least 28 bytes, hence the comment

crates/telio-firewall/src/firewall.rs
```diff
@@ -476,29 +509,43 @@ impl StatefullFirewall {
...
-        let key = unwrap_option_or_return!(Self::build_udp_key(ip, false));
+    fn handle_outbound_udp<'a>(&self, pubkey: PublicKey, ip: &impl IpPacket<'a>) {
+        let link = unwrap_option_or_return!(Self::build_udp_key(ip, false));
+        let key = UdpConn { link, pubkey };
+
+        let Some(headers) = ip
+            .packet()
+            .chunks_exact(UdpConnectionInfo::LAST_PKG_HEADER)
+            .next()
+            .and_then(|c| c.try_into().ok())
```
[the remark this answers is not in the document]
matszczygiel: Yeah, plus the `packet()` method returns `&[u8]`. If the compiler failed to not infer that, it would not compile

crates/telio-firewall/src/firewall.rs
matszczygiel: Yeah, that's a good idea :+1:

crates/telio-firewall/src/firewall.rs
matszczygiel: I'm not entirely sure. I tried not to change the whitelisting logic. However, port whitelisting is for the incoming packets only it seems

crates/telio-firewall/src/firewall.rs
matszczygiel: Oh yeah, it's a leftover as the real source setup is 30 lines below. Thanks for pointing this out

crates/telio-firewall/src/firewall.rs
matszczygiel: It is. Thanks, I will change the comment

crates/telio-firewall/src/firewall.rs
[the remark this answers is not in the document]
LukasPukenis: Oh dang! it's 8 bytes of payload and 20 for the header of UDP :D

crates/telio-firewall/src/firewall.rs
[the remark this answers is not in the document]
LukasPukenis: I am fully aware that this compiles, just wondered how it works, thanks for confirming

crates/telio-firewall/src/firewall.rs
LukasPukenis: Fair enough, this PR doesnt look like it changes that behaviour. @Jauler @packgron could you shed some light on the whitelisting?

LukasPukenis (commented):
Overall +1

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-wg/src/wg.rs
```diff
@@ -388,9 +389,12 @@ impl WireGuard for DynamicWg {
         let _ = self.task.stop().await.resume_unwind();
     }
 
-    async fn reset_existing_connections(&self, exit: PublicKey) -> Result<(), Error> {
+    async fn reset_existing_connections(&self, exit: &ExitNode) -> Result<(), Error> {
+        let pubkey = exit.public_key;
+        let addr = exit.endpoint.map(|addr| addr.ip());
```
Jauler, line 394: There is a small problem with using exit_node's endpoing like this. So generally `ExitNode` struct, which is provided to `connect_exit_node(...)` structure may potentially refer to VPN server as well as Meshnet exit node. While for VPN apps will know, and will provide endpoint to connect to, this is not the case for meshnet exit nodes.

What is slightly annoying for meshnet nodes - the endpoint is fundamentally rather dynamic, and can change during runtime at any point. Therefore my recommendation would be to resolve the endpoint in the `device/mod.rs` method `connect_exit_node()` just before calling `.reset_existing_connections()`. The resolution would look like this. If `exit_node` public key is amongst the `self.requested_state_meshnet_config.peers` or even better -> in wireguard peer list, which you can acquire from `self.entities.wireguard_interface`, then acquire endpoint from that peers endpoint provided by `self.entities.wireguard_interface`. Otherwise take endpoint from `exit_node`.

crates/telio-firewall/src/firewall.rs
```diff
@@ -240,12 +243,18 @@ pub struct StatefullFirewall {
     allow_ipv6: bool,
     /// Wheter to still keep track of whitelisted TCP connections.
```
Jauler, line 244:
```suggestion
    /// Wheter to still keep track of whitelisted TCP/UDP connections.
```

crates/telio-firewall/src/firewall.rs
```diff
@@ -310,6 +319,12 @@ struct TcpConn {
     pubkey: PublicKey,
 }
 
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+struct UdpConn {
```
Jauler, line 323: Maybe it makes sense to merge `UdpConn` and `TcpConn` to one struct, as both seems to be identical by now :).

crates/telio-firewall/src/firewall.rs
```diff
@@ -476,29 +509,43 @@ impl StatefullFirewall {
...
 
-    fn handle_outbound_udp<'a>(&self, ip: &impl IpPacket<'a>) {
-        let key = unwrap_option_or_return!(Self::build_udp_key(ip, false));
+    fn handle_outbound_udp<'a>(&self, pubkey: PublicKey, ip: &impl IpPacket<'a>) {
+        let link = unwrap_option_or_return!(Self::build_udp_key(ip, false));
+        let key = UdpConn { link, pubkey };
+
+        let Some(headers) = ip
+            .packet()
+            .chunks_exact(UdpConnectionInfo::LAST_PKG_HEADER)
```
Jauler, line 529: If I understand correctly, this may fail for IP packets containing options, as then the IP header will be longer, and UDP header would get cut out?

refering to [IPv4Packet::get_header_length()](https://docs.rs/pnet/latest/pnet/packet/ipv4/struct.Ipv4Packet.html#method.get_header_length) could figure out the length of the header. and then adding 8 bytes as per [ICMP RFC](https://datatracker.ietf.org/doc/html/rfc792) probably would resolve the issue.

crates/telio-firewall/src/firewall.rs
Jauler: > My confusion comes from the fact we set is_remote_initiated when we receive inbound UDP connection not in cache and then later we check for it and check if port is whitelisted, what is the reason for that?

For non-whitelisted peers we want to allow responses for locally initiated UDP flows (or TCP connections), but drop any packet for remotely initiated ones. Therefore we remember the 5-tuple (proto, src ip, src port, dst ip, dst port) in the LRU cache, which indicates that the connection is locally initiated to later allow responses on the same tuple (just of course with swapped src and dst).

**Note:** before stale connection killing there was no reason to do connection tracking for whitelisted peers (thus it was not done), as the packets would get accepted without consulting the connection tracking state. Right now we need that in order to have enough information when injecting TCP RST's or ICMP destination unreachable upon ExitNode reconnect.

crates/telio-firewall/src/firewall.rs
```diff
@@ -310,6 +319,12 @@ struct TcpConn {
     pubkey: PublicKey,
 }
 
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+struct UdpConn {
+    link: IpConnWithPort,
+    pubkey: PublicKey,
```
Jauler, line 325: Interestingly enough, I think adding key here covers LLT-4502 for UDP, and TCP is already covered by previous PR. Thus in order to finish LLT-4502, only ICMP seems to be left :).

crates/telio-firewall/src/firewall.rs
```diff
@@ -559,9 +610,7 @@ impl StatefullFirewall {
                 key,
                 connection_info
             );
-            if connection_info.is_remote_initiated
-                && !whitelist.is_port_whitelisted(peer, key.local_port)
-            {
+            if connection_info.is_remote_initiated && !is_port_whitelisted(key.link.local_port) {
                 telio_log_trace!("Removing UDP conntrack entry {:?}", key);
                 udp_cache.remove(&key);
```
Jauler, line 615: This part seems to have depended on the fact, that for whitelisted peers this functions was not called previously. As it removes the `is_remote_initiated` immediately. This has changed right now, so we need to check whitelist and remove entry from udp_cache only if both are true:
* peer is not whitelisted
* peer is talking over not whitelisted port

Jauler: Unfortunately, it seems like I have missed this during review on TCP too. Could we:
* Add a unit or integration test where `record_whitelisted == true` and some inbound connections (both, for TCP and UDP) are initiated to check if they can succesfully communicate past first response packet. (e.g. one outgoing packet and later two/three response packets).
* Fix inbound packet handling to also check whether peer is whitelisted (not just port) when doing cache removals

crates/telio-firewall/src/firewall.rs
```diff
@@ -572,22 +621,19 @@ impl StatefullFirewall {
...
-        telio_log_trace!(
-            "Updating UDP conntrack entry {:?} {:?} {:?}",
-            key,
-            peer,
-            conninfo
-        );
-        udp_cache.insert(key, conninfo);
+        telio_log_trace!("Updating UDP conntrack entry {:?} {:?}", key, peer,);
+        match udp_cache.entry(key) {
+            Entry::Occupied(mut occ) => occ.get_mut().is_remote_initiated = true,
```
Jauler, line 631: It seems like this `Entry::Occupied` ARM cannot really be executed ever, as if the if statement at line 607 was false (otherwise it would have returned by here), means, that it must be Vacant. 

Maybe it would be clearer to convert that `if` into `match` statement, to avoid such an arm?

crates/telio-firewall/src/firewall.rs
```diff
@@ -938,6 +986,202 @@ impl StatefullFirewall {
...
+        #[allow(index_access_check)]
+        let mut icmppkg = MutableIcmpPacket::new(
+            &mut ipv4pkgbuf[IPV4_HEADER_LEN..(IPV4_HEADER_LEN + ICMP_HEADER_LEN)],
+        )
+        .expect("ICMP buffer should not be too small");
+        icmppkg.set_icmp_type(IcmpTypes::DestinationUnreachable);
+        icmppkg.set_icmp_code(IcmpCodes::DestinationPortUnreachable);
+        drop(icmppkg);
+
+        let iter = udp_conn_cache.iter().filter_map(|(k, v)| {
```
Jauler, line 1129: If we have that `continue` on line 1140, maybe this filtering is not really necessary? (also applies to TCP).

crates/telio-firewall/src/firewall.rs
```diff
@@ -938,6 +986,202 @@ impl StatefullFirewall {
...
+            if k.pubkey == *pubkey {
+                Some((&k.link, v))
+            } else {
+                None
+            }
+        });
+
+        for (key, val) in iter {
+            let Some(last_headers) = &val.last_out_pkg_headers else {
```
Jauler, line 1138: Maybe it would be nice to filter these in the same  filter (e.g. at line 1129)? (also applies for TCP).

crates/telio-firewall/src/firewall.rs
```diff
@@ -938,6 +986,202 @@ impl StatefullFirewall {
...
+        };
+
+        telio_log_debug!(
+            "Inspecting {} UDP connections for reset",
+            udp_conn_cache.len()
+        );
+
+        const IPV4_HEADER_LEN: usize = 20;
+        const ICMP_HEADER_LEN: usize = 8;
+        const ICMP_DATA_LEN: usize = UdpConnectionInfo::LAST_PKG_HEADER;
```
Jauler, line 1102: here we must take care to see whether (`last_out_pkg_headers` did not have any IP options set. As otherwise it may be longer than 20 bytes, meaning, that full ICMP payload may be larger then 28 bytes.

crates/telio-wg/src/wg.rs
LukasPukenis: Why it's not the case for the meshnet node?

matszczygiel: @Jauler I've proved fallback in case it's `None`. However, your description provides a more complete solution :+1:

crates/telio-firewall/src/firewall.rs
matszczygiel: Yeaaaah, however, it may not last long :). I kind of prefer keeping things type-safe

crates/telio-firewall/src/firewall.rs
matszczygiel: I wanted to keep things on a stack array for performance. However, it seems to me that the length should be variable. In that case, let me store much more of a packet, I believe one of the RFCs recommends attaching the whole message to the ICMP packet

crates/telio-firewall/src/firewall.rs
matszczygiel: Yep, correct. That's why I marked LLT-4502 as blocked by this feature

crates/telio-firewall/src/firewall.rs
LukasPukenis: Thanks for the explanation.

I think the logic is quite convoluted around the `is_remote_initiated` flag, the state machine is quite complicated that given several re-reads I still have trouble understanding it fully.

I think this part in `handle_inbound_udp:613` is not needed:
```
!is_port_whitelisted(key.link.local_port)
```

As there's no way for `is_remote_initiated` to become `true` but not have port whitelisted unless the whitelisting was modified mid-flight, is this possible..?

crates/telio-firewall/src/firewall.rs
matszczygiel: Ok, let me fix these

crates/telio-firewall/src/firewall.rs
matszczygiel: Actually, I cannot replace it with entry altogether since `entry()` does not invalidate outdated values

matszczygiel: Or I will try to change the `entry()` API so that it invalidates the keys

crates/telio-firewall/src/firewall.rs
matszczygiel: @Jauler actually I think the current implementation ticks all boxes. Can you take a look again? if I'm missing something? Note that `is_port_whitelisted()` is a closure which for whitelisted peer always returns `true`

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done, I've used the max IP header len (60) + UDP header len (8)

crates/telio-firewall/src/firewall.rs
matszczygiel: I made the changes, but I'm leaving this discussion open as I'm not sure if that's the correct approach

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

dfetti:
+0.5

crates/telio-wg/src/wg.rs
```diff
@@ -388,9 +389,18 @@ impl WireGuard for DynamicWg {
         let _ = self.task.stop().await.resume_unwind();
     }
 
-    async fn reset_existing_connections(&self, exit: PublicKey) -> Result<(), Error> {
+    async fn reset_existing_connections(&self, exit: &ExitNode) -> Result<(), Error> {
+        let pubkey = exit.public_key;
+        let exit_addr = exit.endpoint.map(|addr| addr.ip());
```
Jauler, line 394: Damn, I just realized, that we are mixing out-of-tunnel IP addressses and in-tunnel IP addresses here. Got confused myself previously too.

I am thinking, maybe using our own tunnel interface IP address would be better? This means, that setting source IP from endpoint is not really correct :(.

Therefore maybe it would be best to take local tunnel interface IP address (IPv4 one) and use that from this: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-model/src/config.rs#L185 and if this is None (the whole config in `Device::requested_config.meshnet_config`, then default to `10.5.0.2`.

crates/telio-firewall/src/firewall.rs
Jauler: > unless the whitelisting was modified mid-flight, is this possible..?

Yes, it is possible.

crates/telio-firewall/src/firewall.rs
Jauler: Oh yeah, you are right.

But in that case, maybe lets rename the closure? Generally this closure now checks whether the connection is whitelisted (and this means, that if `true` is returned it might be not port-whitelisted, but simply all incoming connections allowed). So maybe lets call it `is_connection_whitelisted` hopefully to avoid some misunderstandings.

crates/telio-utils/src/lru_cache.rs
```diff
@@ -157,6 +171,8 @@ impl<Key: Clone + Eq + Hash, Value> LruCache<Key, Value> {
     /// Gets the given key’s corresponding entry in the map for in-place manipulation.
     #[inline(always)]
     pub fn entry(&mut self, key: Key) -> Entry<'_, Key, Value> {
+        self.remove_expired();
```
tomaszklak, line 174: @Jauler pointed out this change here to me. Nice find! I've just fixed this (and updated tests to detect this) here: https://github.com/NordSecurity/libtelio/pull/222/files

crates/telio-wg/src/wg.rs
matszczygiel: I did that and it doesn't work :(

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: The conclusion is that we still need to check for the port being whitelisted

crates/telio-wg/src/wg.rs
matszczygiel: After a new round of trying it works finally

tomaszklak:
Please run criterion benchmarks of telio-firewall and compare with the result from master to make sure we don't introduce performance regression.

matszczygiel:
@tomaszklak Yeah, there are some regressions up to 30%. However, there are some improvements as well. 
It's somewhat expected I guess

src/device/mod.rs
```diff
@@ -1458,15 +1460,41 @@ impl Runtime {
         wg_controller::consolidate_wg_state(&self.requested_state, &self.entities, &self.features)
             .await?;
 
-        let old_exit_pubkey = old_exit_node
+        if let Some(last_exit) = old_exit_node
             .as_ref()
             .or(self.requested_state.last_exit_node.as_ref())
-            .map(|node| node.public_key);
+        {
```
Jauler, line 1466: nice :+1:

crates/telio-firewall/src/firewall.rs
```diff
@@ -922,7 +988,9 @@ impl StatefullFirewall {
                 }
                 is_in_cache
             }
-            IcmpErrorKey::Udp(Some(udp_key)) => {
+            IcmpErrorKey::Udp(Some(link)) => {
```
Jauler, line 991: Could you also rename `is_port_whitelisted` to `is_conn_whitelisted` in TCP too? For the same reason.

Jauler (dismissed):
+1.0, just rename that `is_port_whitelisted` in `handle_inbound_tcp()`. Thanks :pray: .

crates/telio-firewall/src/firewall.rs
```diff
@@ -938,6 +1006,203 @@ impl StatefullFirewall {
...
+
+                    #[allow(index_access_check)]
+                    ipv4pkgbuf[..IPV4_LEN].copy_from_slice(ipv4pkg.packet());
+
+                    #[allow(index_access_check)]
+                    ipv4pkgbuf[IPV4_LEN..].copy_from_slice(tcppkg.packet());
+
+                    sink4.write_all(&ipv4pkgbuf)?;
+                }
+                (IpAddr::Ipv6(_), IpAddr::Ipv6(_)) => (), // TODO(msz): implement this pice when IPv4 will be fully supported
```
tomaszklak, line 1094: I think the comment should refer to IPv6

tomaszklak:
> @tomaszklak Yeah, there are some regressions up to 30%. However, there are some improvements as well. It's somewhat expected I guess

I think this might be a good reason to add a separate task to investigate if the real world performance changed after those changes.

tomaszklak (commented):
+0.5, conditional on addressing comments :)

crates/telio-firewall/src/firewall.rs
matszczygiel: Oh, yeah. You're totally right

crates/telio-firewall/src/firewall.rs
```diff
@@ -238,14 +241,22 @@ pub struct StatefullFirewall {
...
+    /// Wheter to still keep track of whitelisted TCP/UDP connections.
     /// Used for connection reset mechanism
-    tcp_record_whitelisted: bool,
+    record_whitelisted: bool,
 }
 
 #[derive(Debug)]
 struct UdpConnectionInfo {
     is_remote_initiated: bool,
+    last_out_pkg_chunk: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
matszczygiel: Yeah, I was thinking about it however I try to allocate the buffer only once at the beginning.

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
```diff
@@ -476,29 +511,57 @@ impl StatefullFirewall {
...
             IpNextHeaderProtocols::Icmp => self.handle_inbound_icmp(peer, &ip),
             IpNextHeaderProtocols::Icmpv6 if self.allow_ipv6 => self.handle_inbound_icmp(peer, &ip),
             _ => false,
         }
     }
 
-    fn handle_outbound_udp<'a>(&self, ip: &impl IpPacket<'a>) {
-        let key = unwrap_option_or_return!(Self::build_udp_key(ip, false));
+    fn handle_outbound_udp<'a>(&self, pubkey: PublicKey, ip: &impl IpPacket<'a>) {
+        let link = unwrap_option_or_return!(Self::build_udp_key(ip, false));
```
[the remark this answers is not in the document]
matszczygiel: Done

crates/telio-firewall/src/firewall.rs
matszczygiel: Done

tomaszklak (approved):

PR #216, opened 2023-11-17
LLT-4006: Restrict nat-lab containers capabilities

lcruz99:
### Problem
At the moment nat-lab docker containers are being ran with root privileges, thus have all the kernel capabilities enabled, which is unsafe and not necessary. 

### Solution
This PR restricts these containers capabilities, following the least privilege principle, and also adds other unrelated changes:
- Remove Windows VM firewall disablement for every test (done now during VM startup).
- Remove previously added windows VM debug logs which are no longer needed.
- Add custom peer names integration test that asserts the nodes can have their nicknames changed/removed.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v0.4.5
+          triggered-ref: v0.4.6
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -11,29 +11,12 @@ services:
     profiles:
       - base
 
-  # Create gateway for first cone network
-  cone-gw-01: &common-gw
-    hostname: cone-gw-01
-    image: nat-lab:base
-    entrypoint: /opt/bin/cone-gw
-    cap_drop:
-      - ALL
-    cap_add:
-      - NET_ADMIN
-      - NET_RAW
-    sysctls:
-      - net.ipv4.ip_forward=1
-    security_opt:
-      - no-new-privileges:true
-    networks:
-      internet:
-        ipv4_address: 10.0.254.1
-        priority: 1000
-      cone-net-01:
-        ipv4_address: 192.168.101.254
-        priority: 900
-  # Create one client for cone network, this client will use
-  # the above gateway as its route to the "internet"
+
+  ###########################################################
+  #                         CLIENTS                         #
+  ###########################################################
+  # Clients for cone network, they'll use cone gateways
+  # as their route to the open internet
   cone-client-01: &common-client
     hostname: cone-client-01
     image: nat-lab:base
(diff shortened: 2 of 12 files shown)


.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v0.4.4
+          triggered-ref: LLT-4006_fix_libvirt
```
lcruz99, line 34: To be updated to eventual v0.4.5 tag before merging.

nat-lab/docker-compose.yml
```diff
@@ -11,157 +11,94 @@ services:
...
+    hostname: cone-client-01
     image: nat-lab:base
-    entrypoint: /opt/bin/vpn-server
+    entrypoint: /opt/bin/client
+    devices:
+      - "/dev/net/tun"
+    cap_drop:
+      - ALL
+    cap_add:
+      - NET_RAW
```
lcruz99, line 29: NET_XXX capabilities are need for iptables/tcpdump/tcli

nat-lab/docker-compose.yml
```diff
@@ -11,157 +11,94 @@ services:
...
+  ###########################################################
+  # Clients for cone network, they'll use cone gateways
+  # as their route to the open internet
+  cone-client-01: &common-client
+    hostname: cone-client-01
     image: nat-lab:base
-    entrypoint: /opt/bin/vpn-server
+    entrypoint: /opt/bin/client
+    devices:
+      - "/dev/net/tun"
```
lcruz99, line 25: allows boringtun to use tun/tap interfaces

nat-lab/docker-compose.yml
```diff
@@ -11,157 +11,94 @@ services:
...
+    cap_drop:
+      - ALL
+    cap_add:
+      - NET_RAW
+      - NET_ADMIN
+      - NET_BIND_SERVICE
+      - SETUID
```
lcruz99, line 32: SETUID/SETGID are needed for applications that change to a non-root user (privilege deescalation). This could be exploited by other actors to do the opposite but if containers runs as root it's trivial..

nat-lab/tests/test_dns.py
```diff
@@ -860,3 +860,111 @@ async def test_dns_nickname() -> None:
         )
         for ip in beta.ip_addresses:
             assert ip in beta_response.get_stdout()
+
+
+@pytest.mark.asyncio
+async def test_dns_change_nickname() -> None:
```
tomaszklak, line 866: I'm not sure how this test relates to the restriction of nat-lab containers 🤔

nat-lab/tests/utils/vm/windows_vm_util.py
```diff
@@ -37,32 +37,13 @@ async def new_connection() -> AsyncIterator[Connection]:
...
-            print(stdout)
-
-        await connection.create_process(["route", "print"]).execute(
-            stdout_callback=on_stdout,
-            stderr_callback=on_stdout,
-        )
-        await connection.create_process(["ipconfig/all"]).execute(
-            stdout_callback=on_stdout,
-            stderr_callback=on_stdout,
-        )
```
tomaszklak, line 52: Is having those logs causing some problems? If not maybe we can keep them.

tomaszklak (commented):
Based on the commit messages I assume you didn't hide anything in the reformatting ;) and I think that at least that one commit should be kept separate if you plan to squash everything before merging.

nat-lab/tests/test_dns.py
lcruz99: you're right it's unrelated, I should have mentioned it

nat-lab/tests/utils/vm/windows_vm_util.py
lcruz99: I've added them a while ago to try to debug a problem with Windows VMs and are no longer needed. I would keep them out to avoid polluting the logs

nat-lab/tests/test_dns.py
tomaszklak: If that is the case, can you extract it to separate review?

lcruz99: ✅

nat-lab/docker-compose.yml
```diff
@@ -11,168 +11,94 @@ services:
...
+      - "derp-02:10.0.10.2"
+      - "derp-03:10.0.10.3"
+      - "derp-00:10.0.10.245"
     volumes:
       - ../:/libtelio
-  vpn-02:
-    hostname: vpn-02
-    image: nat-lab:base
-    entrypoint: /opt/bin/vpn-server
+  cone-client-02:
```
tomaszklak, line 60: Why doesn't this client have the same cap_*, devices, security and sysctls sections as the `cone-client-01`?

lcruz99: It has, it inherits its values through the `common-client` anchor.

tomaszklak: Oops, right. I didn't understand anchors...

tomaszklak (approved):
+1.0

nat-lab/docker-compose.yml
tomaszklak: Oops, right. I didn't understand anchors...

LukasPukenis (commented):
I am not fully sold on the idea it's necessary to limit container's, as the harm is really limited to the virtual machine alone.

However, I assume that if it works, means the refactor didn't break anything so obligatory +1 :)

LukasPukenis (dismissed):

LukasPukenis (approved):

PR #220, opened 2023-11-22
LLT-4502: Perform connection state tracking in telio firewall for each peer separately

matszczygiel:
### Problem
The TCP connection can be dropped by spoofing ICMP error packets. It's because the peer public key is not matched to the TCP connection conntrack thus allowing different peers to affect the connection

### Solution
This PR adds a validation of the peer's public key before any action on the TCP connection is done 

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -333,6 +333,9 @@ struct IcmpConn {
     local_addr: IpAddr,
     response_type: u8,
     identifier_sequence: u32,
+    // This public key refers to source peer for inbound connections and
+    // destination peer for outbound connections
+    pubkey: PublicKey,
 }
 
 type IcmpKey = Result<IcmpConn, IcmpErrorKey>;
@@ -458,10 +461,10 @@ impl StatefullFirewall {
                 self.handle_outbound_tcp(peer, &ip);
             }
             IpNextHeaderProtocols::Icmp => {
-                self.handle_outbound_icmp(&ip);
+                self.handle_outbound_icmp(peer, &ip);
             }
             IpNextHeaderProtocols::Icmpv6 => {
-                self.handle_outbound_icmp(&ip);
+                self.handle_outbound_icmp(peer, &ip);
             }
             _ => (),
         };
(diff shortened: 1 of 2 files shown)


matszczygiel:
Blocked by #214

crates/telio-firewall/src/firewall.rs
```diff
@@ -3137,4 +3142,158 @@ pub mod tests {
...
+            them: &'static str,
+            random: &'static str,
+        }
+
+        let test_inputs = [TestInput {
+            us: "127.0.0.1:1111",
+            them: "8.8.8.8:8888",
+            random: "192.168.0.1:7777",
+        }];
+        for test_input @ TestInput { us, them, random } in test_inputs {
```
mathiaspeters, line 3209:
```suggestion
        for TestInput { us, them, random } in test_inputs {
```

crates/telio-firewall/src/firewall.rs
```diff
@@ -3137,4 +3142,158 @@ pub mod tests {
...
+                src1: "127.0.0.1:1111",
+                src2: "127.0.0.1:2222",
+                src3: "127.0.0.1:3333",
+                src4: "127.0.0.1:4444",
+                src5: "127.0.0.1:5555",
+                dst1: "8.8.8.8:8888",
+                dst2: "8.8.8.8:7777",
+                make_udp: &make_udp,
+            },
+            // TODO(msz): Implement a fix for IPv6
```
mathiaspeters, line 3252: If there is a separate ticket for fixing this then I think you should add the ticket number here, otherwise, I think it should be fixed in this PR

mathiaspeters:
Two of the tests are specifying a list of test inputs even though there's just a single tested case. Is there a plan to add more test cases? If not, I would remove the loop and whoever adds extra cases in the future can introduce the loop again.

Also, commits are clean and well-described, but in my opinion this should be just one commit

matszczygiel:
@mathiaspeters the test cases are directly copied from the VCP tickets in a pretty much unchanged state. Let me improve them.
Regarding commits, do you mean that you'd like to see one commit for all tests and the fix in a separate one?

crates/telio-firewall/src/firewall.rs
matszczygiel: I've cleaned the test code. However since each test case corresponds to a different VCP ticket, I prefer to keep them in separate commits

mathiaspeters:
We can wait for input from the other reviewer before you make any changes to the commits, but in my opinion the tests and fix should be in the same commit since it's such a small and self-contained change

matislovas:
I'm not sure that it is a separate task, but I would add some tests to `nat-lab` as well.

matislovas:
> @matislovas The unit tests cover 100% of the problem. If they are passing, that means the vulnerability is gone. What's the point of nat-lab tests in that case?

Ok than, +1

mathiaspeters:
@matislovas Any opinion on the commits here? You will be the tie breaker on keeping as is vs squashing to a single commit

mathiaspeters (approved):

matszczygiel:
@mathiaspeters After a short discussion with @LukasPukenis he's opinion is that he prefers separate commits in that case, so I'm merging it as is.
Thanks for the review :+1:

PR #221, opened 2023-11-22
Handle egress errors

Hasan6979:
### Problem
Add error handling in proxy egress. Log change of state and error only once.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-proxy/src/proxy.rs
+++ b/crates/telio-proxy/src/proxy.rs
@@ -2,6 +2,7 @@ use async_trait::async_trait;
 use futures::future::{pending, select_all, FutureExt};
 use std::{
     collections::{HashMap, HashSet},
+    io::ErrorKind,
     net::{Ipv4Addr, SocketAddr},
     sync::Arc,
     time::Duration,
@@ -19,7 +20,7 @@ use telio_task::{
 };
 use tokio::{net::UdpSocket, sync::mpsc::error::SendTimeoutError};
 
-use telio_utils::{telio_log_debug, PinnedSleep};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, PinnedSleep};
 
 type SocketMap = HashMap<PublicKey, Arc<UdpSocket>>;
 type SocketMuteMap = HashMap<PublicKey, (Arc<UdpSocket>, Option<PinnedSleep<PublicKey>>)>;
@@ -44,6 +45,39 @@ pub enum Error {
     /// Peer not found
     #[error("Peer not found error")]
     PeerNotFound,
+}
+
+#[derive(Debug, thiserror::Error)]
+/// Mapping proxy errors into 3 broad categories
+pub enum ErrorType {
+    /// Error will recover automatically
+    #[error("Recoverable Error")]
+    RecoverableError,
+    /// Error cannot be resolved in this module and
+    /// has to be handled externally
+    #[error("Unrecoverable Error")]
+    UnrecoverableError,
+    /// Socket IO error. Restart the WG socket and continue
+    #[error("Socket IO Error")]
+    SocketIOError,
+}
+
+impl From<std::io::Error> for ErrorType {
+    fn from(f: std::io::Error) -> ErrorType {
+        match f.kind() {
+            ErrorKind::PermissionDenied | ErrorKind::InvalidInput | ErrorKind::InvalidData => {
+                ErrorType::SocketIOError
+            }
+            ErrorKind::ConnectionRefused
+            | ErrorKind::ConnectionReset
+            | ErrorKind::ConnectionAborted
+            | ErrorKind::TimedOut
+            | ErrorKind::WriteZero
+            | ErrorKind::AddrNotAvailable => ErrorType::RecoverableError,
+            ErrorKind::BrokenPipe => ErrorType::UnrecoverableError,
+            _ => ErrorType::RecoverableError,
+        }
+    }
 }
 
 /// Proxy links incoming and outgoing WG Packets to
@@ -97,6 +131,7 @@ struct StateEgress {
     sockets: SocketMuteMap,
     input: Rx<(PublicKey, DataMsg)>,
     wg_addr: Option<SocketAddr>,
+    conn_state: Result<(), std::io::ErrorKind>,
 }
 
 impl UdpProxy {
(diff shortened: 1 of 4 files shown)


crates/telio-proxy/src/proxy.rs
```diff
@@ -39,6 +40,48 @@ pub enum Error {
     /// Send failed
     #[error(transparent)]
     Send(#[from] SendTimeoutError<(PublicKey, DataMsg)>),
+    /// WG addr is null
+    #[error("Empty wg Addr")]
+    WgAddrNull,
```
lcruz99, line 45: Are these new Errors WgAddrNull and RecvChanError necessary? If not, use ErrorKind::AddrNotAvailable/ErrorKind::BrokenPipe instead

crates/telio-proxy/src/proxy.rs
```diff
@@ -191,6 +243,40 @@ impl StateEgress {
...
+    }
+
+    async fn handle_error(&mut self, err: Error) {
+        if self.conn_state != OutboundConnState::Error {
+            Self::log_error(&err);
+
+            match ErrorType::from(err) {
+                ErrorType::RecoverableError => (),
+                ErrorType::UnrecoverableError => {
+                    Self::sleep_forever().await;
```
lcruz99, line 261: If there's an UnrecoverableError `conn_state` should be `OutboundConnState::Error`, move its assignment to the beginning of this function.

crates/telio-proxy/src/proxy.rs
```diff
@@ -191,6 +243,40 @@ impl StateEgress {
...
+                            tx_buf_size: Some(SOCK_BUF_SZ),
+                        })),
+                    )
+                    .await
+                    {
+                        telio_log_error!("Cannot open socket. ERR :{}", err);
+                    };
+                }
+            }
+            self.conn_state = OutboundConnState::Error;
```
lcruz99, line 277: If there's an `UnrecoverableError`, this line won't be reached as task is suspended. Move this assignment to the beginning of the function.

lcruz99, line 277: If we have a Recoverable/SocketIO-Unrecoverable error sequence it will still call recv(). I don't think `conn_state` is needed here

lcruz99 (changes requested):
conn_state logic is flawed, either remove it or fix the case when UnrecoverableError happens coming from any other Error.

crates/telio-proxy/src/proxy.rs
```diff
@@ -39,6 +40,48 @@ pub enum Error {
...
+            Error::IoError(k) => match k.kind() {
+                ErrorKind::PermissionDenied => ErrorType::SocketIOError,
+                ErrorKind::ConnectionRefused => ErrorType::RecoverableError,
+                ErrorKind::ConnectionReset => ErrorType::RecoverableError,
+                ErrorKind::ConnectionAborted => ErrorType::RecoverableError,
+                ErrorKind::InvalidInput => ErrorType::SocketIOError,
+                ErrorKind::InvalidData => ErrorType::SocketIOError,
+                ErrorKind::TimedOut => ErrorType::RecoverableError,
+                ErrorKind::WriteZero => ErrorType::RecoverableError,
+                _ => ErrorType::RecoverableError,
```
matislovas, line 80: You can combine multiple match arms to one line:

e.g.:
```
ErrorKind::InvalidInput | ErrorKind::InvalidData => ErrorType::SocketIOError,
```

crates/telio-proxy/src/proxy.rs
Hasan6979: Yes. recv() would be called again and again and until the error is resolved it will keep logging error. conn_state is added to avoid spamming the logs with error. That applies for when everything is working as well!

matislovas:
+1

crates/telio-proxy/src/proxy.rs
lcruz99: IMO `ErrorKind::AddrNotAvailable` and `ErrorKind::BrokenPipe` looks appropriate for those cases when the address (or ifc) is not available or when the channel is closed (broken).

> I also feel that reusing the ErrorKind errors would not propagate the error type properly.

Why do you think that?

crates/telio-proxy/src/proxy.rs
lcruz99: If the channel is broken, it doesn't matter how many times you call `recv()`, it won't fix it.

Let's say you have a Recoverable/SocketIO error (`conn_state` -> `OutboundConnState::Error`) and then, right on the next call to `recv()` the channel is broken and an Unrecoverable error is thrown. How are you going to handle the error, because if `conn_state` is already on the `OutboundConnState::Error` state it won't enter the if clause from [line 239](https://github.com/NordSecurity/libtelio/pull/221/files#diff-9c4504f1cf7958587fd86df0679204014595a9d76b01c8420e61c442cb4a1dd7R239).

Hasan6979: Mmhm. Makes sense!
I changed the logic a bit. Made conn_state as result.

crates/telio-proxy/src/proxy.rs
```diff
@@ -191,6 +227,39 @@ impl StateEgress {
...
+    }
+
+    async fn handle_error(&mut self, err: std::io::Error) {
+        match self.conn_state {
+            Ok(()) => self.log_error(&err),
+            Err(e) if e != err.kind() => self.log_error(&err),
+            Err(_) => (),
+        }
+
+        match ErrorType::from(err) {
```
lcruz99, line 243:
```suggestion
     async fn handle_error(&self, err: ErrorKind) {
        telio_log_error!("Unable to send. {}", err);

        match ErrorType::from(err) {
```
`log_error` can be removed and change `handle_error` input to `ErrorKind` to simplify things

lcruz99: just to add that, following suggestion above `conn_state` could either be assigned here or before this function calls

lcruz99 (dismissed):

lcruz99:
Looks good! Can you just add some tests?

lcruz99:
+2

lcruz99 (dismissed):

lcruz99 (approved):

PR #225, opened 2023-11-27
natlab: Specify destination on macos VPN routing rule

lcruz99:
### Problem

Both VPN and linux gateways were being added to the routing table as default destinations, so when switching to the secondary network, deleting the default route would also delete the VPN route.

### Solution

Specify on the routing table the subnet destination of the packets that should be routed through the VPN interface. This way we can delete the default route while switching networks without touching the VPN route.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/utils/router/mac_router.py
+++ b/nat-lab/tests/utils/router/mac_router.py
@@ -1,10 +1,5 @@
 from .router import Router, IPStack, IPProto
-from config import (
-    LIBTELIO_IPV6_WG_SUBNET,
-    LINUX_VM_PRIMARY_GATEWAY,
-    DERP_SERVERS,
-    VPN_SERVER_SUBNET,
-)
+from config import LIBTELIO_IPV6_WG_SUBNET
 from contextlib import asynccontextmanager
 from typing import AsyncIterator, List
 from utils.connection import Connection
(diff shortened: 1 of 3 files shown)


dfetti (commented):
+1.0
I'm not very familiar with routes, but seems ok.

nat-lab/tests/utils/router/mac_router.py
```diff
@@ -89,15 +74,14 @@ async def create_meshnet_route(self) -> None:
...
-                ["route", "delete", "-inet", "default"]
-            ).execute()
-
-            await self._connection.create_process(
-                ["route", "add", "-inet", "default", "-interface", self._interface_name]
-            ).execute()
-
-            await self._connection.create_process(
-                ["route", "add", "-inet", VPN_SERVER_SUBNET, LINUX_VM_PRIMARY_GATEWAY]
+                [
```
baltrus1, line 77: what is the point of this route? it redirects 10.0/16 to tunnel interface, meanwhile you need to simulate all traffic being sent to tunnel interface, except for vpn server ip

baltrus1: did you mean to do "default" instead of 10.0.0.0/16 maybe?

baltrus1: so maybe 0/1 + 128/1 is what you need? this subnet 10.0/16 is too small to capture all traffic(as the vpn route should do)

lcruz99: That seems about right, fixed

nat-lab/tests/utils/router/mac_router.py
```diff
@@ -89,15 +74,26 @@ async def create_meshnet_route(self) -> None:
...
+                    "route",
+                    "add",
+                    "-inet",
+                    "0/1",
+                    "-interface",
+                    self._interface_name,
+                ]
             ).execute()
 
+        if self.ip_stack in [IPStack.IPv4, IPStack.IPv4v6]:
```
baltrus1, line 87: same if() as above

baltrus1 (dismissed):
lgtm +1

baltrus1 (approved):

PR #227, opened 2023-11-28
natlab: Add change peer nickname test

lcruz99:
### Problem
There was a test case missing for nicknames feature to ensure that a peer nickname change is done with success.

### Solution
Add that test case.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_dns.py
+++ b/nat-lab/tests/test_dns.py
@@ -826,8 +826,8 @@ async def test_dns_nickname() -> None:
             conn.connection for conn in env.connections
         ]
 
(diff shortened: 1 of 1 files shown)


LukasPukenis (commented):
+1 the test looks solid from high level overview. I wish however that the _meat_ of the test would be in just a few lines so we could easily see that:
- we assign a name
- we check for IP
- we rename
- we check for IP
- we reset
- we check for IP
Now each step is wrapped around few lines of code making it an extra effort to parse

PR #228, opened 2023-11-29
Migrate Telio fom log crate to tracing crate

ThrasherLT:
### Problem
*It's difficult to find root causes of bugs on Libtelio and one of the solutions would be to use spans from the tracing crate, but before we can use them, we need to migrate Telio to use the tracing crate instead of log.*

### Solution
*Migrate Telio from using the Log crate to using Tracing crate*

>NOTE: With the current solution the output format for TCLI logs will change. The logs for the FFI bindings will stay the same as they were before the migration tho.

- [x] README.md is updated
- [x] changelog.md is updated


matislovas:
Does this react to build configuration (`--debug` and `--release`)?

crates/telio-sockets/src/protector/linux.rs
```diff
@@ -44,7 +44,6 @@ impl Protector for NativeProtector {
 }
 
 fn set_fwmark(fd: i32, fwmark: u32) -> io::Result<()> {
-    let fwmark = fwmark;
```
Hasan6979, line 47: Was there any specific reason why fwmark was added redundantly here ? Or just something that was missed during reviews ?

src/ffi/mod.rs
```diff
@@ -1039,6 +1096,7 @@ mod tests {
             log_level,
             telio_logger,
         );
+        println!("what {} == {}; {}", res, TELIO_RES_OK, !telio_dev.is_null());
```
Hasan6979, line 1099: Maybe turn this into log! or remove it ?

src/ffi/mod.rs
```diff
@@ -1068,6 +1126,8 @@ mod tests {
             telio_logger,
         );
 
+        println!("what {} == {}; {}", res, TELIO_RES_OK, !telio_dev.is_null());
```
Hasan6979, line 1129: This too.

src/ffi/mod.rs
```diff
@@ -867,29 +863,90 @@ fn filter_log_message(msg: String) -> Option<String> {
...
-        if !self.enabled(record.metadata()) {
+impl Subscriber for TelioTracingSubscriber {
+    fn enabled(&self, metadata: &tracing::Metadata<'_>) -> bool {
+        metadata.level() <= &tracing::level_filters::STATIC_MAX_LEVEL
+            && metadata.level() <= &self.max_level
+    }
+
+    fn new_span(&self, _span: &tracing::span::Attributes<'_>) -> tracing::span::Id {
+        // TODO using a placeholder for now
+        tracing::span::Id::from_u64(1337)
```
matislovas, line 911: Is this actually used? If yes, put the number to `const` variable.

src/ffi/types.rs
```diff
@@ -216,21 +216,21 @@ map_enum! {
...
 
 map_enum! {
     telio_log_level -> Level,
-    TELIO_LOG_CRITICAL = Error,
-    TELIO_LOG_ERROR = Error,
-    TELIO_LOG_WARNING = Warn,
-    TELIO_LOG_INFO = Info,
-    TELIO_LOG_DEBUG = Debug,
-    TELIO_LOG_TRACE = Trace,
+    TELIO_LOG_CRITICAL = ERROR,
```
Hasan6979, line 228: log_critical was used when we were using slog crate. Does it makes sense for us to remove this entirely now ?

clis/tcli/src/cli.rs
```diff
@@ -318,48 +322,20 @@ macro_rules! cli_try {
...
-            }
-        }
+        let (non_blocking_writer, _tracing_worker_guard) =
+            tracing_appender::non_blocking(File::create("tcli.log")?);
+        tracing_subscriber::fmt()
+            .with_writer(non_blocking_writer)
+            .with_ansi(false)
+            .with_line_number(true)
+            .with_level(true)
+            .init();
```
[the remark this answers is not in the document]
ThrasherLT: The module information is printed by default. The ".with_file(true)" function enables actual file path to be printed which was not present in the original logger.
I couldn't find a convenient way to format the time logging, but the current format looks like this: `2023-12-01T08:18:14.160963Z`. I could implement a custom tracing subscriber so that the log format is identical to the old one, but that would take time and I don't know if it's necessary.

crates/telio-wg/src/adapter/linux_native_wg.rs
```diff
@@ -157,9 +157,8 @@ impl LinuxNativeWg {
             Err(e) => {
                 telio_log_error!("LinuxNativeWg: [SET01] {}", e);
                 1
-            }
-            // Err(err::SetDeviceError::NlSerError(_e)) => 1,
-            // Err(err::SetDeviceError::NlError(_e)) => 1,
+            } // Err(err::SetDeviceError::NlSerError(_e)) => 1,
```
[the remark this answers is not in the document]
ThrasherLT: formatter updated this line and I don't even know why was this left here, but OK, I'll remove it :D

src/ffi/mod.rs
ThrasherLT: Yeah, this is my leftover debug code, that I forgot to remove... So I'll clean it up.

crates/telio-sockets/src/protector/linux.rs
ThrasherLT: I can't figure out any reason why this was done, since this isn't a pointer or anything. The function call already creates a new instance of that value on the stack, so there shouldn't be any way that this provides some special functionality. Clippy threw an error here, so that's why I removed it in the first place. I assume that even if this had some special side effect, there should have been at least a comment here.

ThrasherLT:
> Does this react to build configuration (`--debug` and `--release`)?

No, but as far as I could tell, the previous implementation over log didn't react to build configs neither. If there is any need for it or if I missed some functionality, I can add it, but I'm not sure what are the requirements for the functionality

matislovas:
> @matislovas
> 
> > Does this react to build configuration (`--debug` and `--release`)?
> 
> No, but as far as I could tell, the previous implementation over log didn't react to build configs neither. If there is any need for it or if I missed some functionality, I can add it, but I'm not sure what are the requirements for the functionality

IIRC, we cannot see `Debug` and `Trace` if we do not build `--debug` build.

matislovas:
Other that, +1

matislovas (dismissed):

PR #230, opened 2023-11-29
natlab: Refine test_mesh_network_switch_direct flakyness

lcruz99:
### Problem
Flakyness of `test_mesh_network_switch_direct` was random and completely unknown, and further investigation was needed.

### Solution
There were two distinct issues found and the tickets filed: LLT-4600 and LLT-4605.

Also, on the last ping, after switching networks, the timeout was previously increased from the default 5s to 30s to pass the test which ended up hiding the macOS bug. This PR reverts it back to the default timeout.


matislovas (approved):
+2

PR #231, opened 2023-11-30
Flush libmoose changes upon sending heartbeat event

Jauler:
libtelio heartbeat event is generally a once an hour occurance, therefore it is expected that event throughput is so low, that the moose queue will never contain more than one element. What is more, in tests there is an expectation that after triggering the analytics, database will contain events relatively soon. For some reason this does not happen in about 10% of the cases. This change is an attempt to verify whether flushing has impact on the event queueing.

### DIFF
--- a/crates/telio-lana/src/event_log_file.rs
+++ b/crates/telio-lana/src/event_log_file.rs
@@ -461,6 +461,11 @@ pub mod moose {
         }
     }
 
+    /// Mocked moose function
+    pub fn flush_changes() -> std::result::Result<Result, Error> {
+        Ok(Result::Success)
+    }
+
     /// Mocked moose function.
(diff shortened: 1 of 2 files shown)


LukasPukenis:
I have a few questions:
- is there any reason the `flush` is public in moose and was never used so far? I just find it a bit interesting.
- This PR doesn't really explain why this flushing is needed, it says that it's so the heartbeat would be flushed to the database, however what's so special about the heartbeat event?

Sorry, I am not familiar, just interested :)

LukasPukenis:
Thanks for explaining. Does flushing mean that moose collects even in memory and only occasionally flushes them to the disk(sqlite)? I assumed the tests would use file implementation instead so moose events would immediately appear in the file.

ThrasherLT (approved):
+2.0

PR #239, opened 2023-12-04
traversal: Fix libtelio block on upnp gateway search

lcruz99:
### Problem
UPnP started blocking whole libtelio due to blocking sockets used in igd crate. In this case, a call to search_gateway() in third party igd crate blocks processing loop for UPnP endpoint provider, blocking all task_exec!()'s as well, including cross_ping_check pinging and its processing loop, then all CPC's functions using task_exec!() are blocked too and eventually blocks the device processing loop that uses CPC for endpoint validation. From there the whole libtelio is blocked.

### Solution
Spawn a thread to run `search_gateway()` call on a dedicated pool for blocking tasks, using `tokio::task::spawn_blocking()`.

Also, during implementation it was found that the backing off was not working as expected, after the first iteration when the `PinnedSleep` completed a new instance was never being reassigned. This PR also tackles it.

EDIT: It was decided to enable async feature for the igd crate (feature = ["aio"]), avoiding calling sync on async code.

EDIT2: Removed most of the search task logic and added the igd `search_gateway()` call directly on the `tokio::select!` from the main loop

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
+++ b/crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
@@ -5,11 +5,16 @@ use crate::endpoint_providers::{
 use crate::ping_pong_handler::PingPongHandler;
 use async_trait::async_trait;
 use futures::prelude::*;
-use igd::{search_gateway, Gateway, PortMappingProtocol};
+use igd::{
+    aio::{search_gateway, Gateway},
+    PortMappingProtocol,
+};
 use ipnet::Ipv4Net;
 use rand::Rng;
 use rupnp::Error::HttpErrorCode;
+use std::fmt::Debug;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
 use telio_crypto::PublicKey;
@@ -25,45 +30,55 @@ use tokio::{net::UdpSocket, sync::Mutex};
 
 #[cfg(test)]
 use mockall::automock;
-#[cfg(test)]
-const UPNP_INTERVAL: Duration = Duration::from_millis(100);
 
 const MAX_SUPPORTED_PACKET_SIZE: usize = 1500;
-#[cfg(not(test))]
-const UPNP_INTERVAL: Duration = Duration::from_secs(5);
-const GET_INTERFACE_TIMEOUT: u64 = 2;
+const GET_INTERFACE_TIMEOUT_S: Duration = Duration::from_secs(2);
 
 type Result<T> = std::result::Result<T, Error>;
 
-#[async_trait]
 #[cfg_attr(test, automock)]
-pub trait UpnpEpCommands: Send + Sync + Default + 'static {
-    fn check_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<bool>;
-    fn add_any_endpoint_routes(
+#[async_trait]
+pub trait UpnpEpCommands: Send + Default + 'static {
+    async fn check_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<bool>;
+    async fn add_any_endpoint_routes(
         &self,
         ip_addr: Ipv4Addr,
         proxy_port: u16,
         wg_port: u16,
     ) -> Result<(u16, u16)>;
-    fn extend_endpoint_duration(
+    async fn extend_endpoint_duration(
         &self,
         ip_addr: Ipv4Addr,
         proxy_port: PortMapping,
         wg_port: PortMapping,
     ) -> Result<()>;
-    fn delete_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<()>;
-    fn get_igd_gateway(&mut self) -> Result<()>;
-    fn get_external_ip(&self) -> Result<Ipv4Addr>;
+    async fn delete_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<()>;
+    async fn get_external_ip(&self) -> Result<Ipv4Addr>;
+    async fn ensure_igd_gateway(&mut self) -> Result<()>;
+    fn has_igd_gateway(&self) -> bool;
+    fn drop_igd_gateway(&mut self);
 }
 
-#[derive(Default, Debug, Clone)]
+type GatewaySearch = Pin<Box<dyn Future<Output = Result<Gateway>> + Send + Sync + 'static>>;
+
+#[derive(Default)]
 pub struct IgdGateway {
+    search: Option<GatewaySearch>,
     gw: Option<Gateway>,
+}
+
+impl Debug for IgdGateway {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("IgdGateway")
+            .field("search", &self.search.is_some())
+            .field("gw", &self.gw)
+            .finish()
+    }
 }
 
 #[async_trait]
 impl UpnpEpCommands for IgdGateway {
-    fn check_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<bool> {
+    async fn check_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<bool> {
         let mut i = 0;
         let mut map_ok: (bool, bool) = (false, false);
         let mut extend_timeout: bool = false;
@@ -72,7 +87,7 @@ impl UpnpEpCommands for IgdGateway {
             None => return Err(Error::NoIGDGateway),
         };
 
-        while let Ok(resp) = gw.get_generic_port_mapping_entry(i) {
+        while let Ok(resp) = gw.get_generic_port_mapping_entry(i).await {
             i += 1;
 
             if resp.lease_duration < 10 * 60 {
@@ -94,7 +109,7 @@ impl UpnpEpCommands for IgdGateway {
         ))
     }
 
-    fn extend_endpoint_duration(
+    async fn extend_endpoint_duration(
         &self,
         ip_addr: Ipv4Addr,
         proxy_port: PortMapping,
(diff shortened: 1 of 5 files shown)


jjanowsk:
Before I start the review. Have you tried enabling `feature = "aio"` for igd crate and using `igd::aio::search_gateway()` instead?

lcruz99:
> igd::aio::search_gateway()

Will try, also given that I have some last details to change will put this review back to draft for now.

jjanowsk:
I don't think it's true. it uses `tokio::net::UdpSocket` instead of `std::net::UdpSocket` which is set as nonblocking by default. Otherwise having `recv_from` method being `async` would not make sense in `tokio::net::UdpSocket`

lcruz99:
You're right on the socket being non-blocking, however the tokio::udpsocket::recv_from() is not:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3e9fe645290bea261f1306621f8cad4c

So the gist is igd behaviour where I was expecting to have a WOULDBLOCK error, which is the case of tokio::udpsocket::try_recv_from()

mathiaspeters:
+1.0

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -35,35 +41,71 @@ const GET_INTERFACE_TIMEOUT: u64 = 2;
...
+        }
+
+        igd_gateway
+    }
+
+    fn start_gw_search(&mut self) {
+        let tx = self.tx.clone();
+        self.search_handle = Some(tokio::task::spawn(async move {
+            let result = Box::pin(search_gateway(Default::default())).await;
+            let _ = tx.send(result).await;
```
packgron, line 101: Look to me like you can simply use task return instead of channel.
Then rewriting to something like this:
```rust
    async fn get_igd_gateway(&mut self) -> Result<()> {
        match &mut self.search_handle {
            Some(handle) if handle.is_finished() => {
                self.gw = handle.await.ok().and_then(std::result::Result::ok);
            }
            None => self.start_gw_search(),
            _ => (),
        };

        // Gateway search in progress, check again on the following iteration
        Err(Error::NoIGDGateway)
    }
```
Maybe adding warn log's for failure case

packgron (changes requested):

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -35,35 +41,71 @@ const GET_INTERFACE_TIMEOUT: u64 = 2;
...
+            igd_gateway.start_gw_search();
+        }
+
+        igd_gateway
+    }
+
+    fn start_gw_search(&mut self) {
+        let tx = self.tx.clone();
+        self.search_handle = Some(tokio::task::spawn(async move {
+            let result = Box::pin(search_gateway(Default::default())).await;
```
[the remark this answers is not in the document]
lcruz99: dylint was throwing an error about this await, "large future with a size of 2176 bytes", it suggested to Box pin it and store it on the heap instead

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -18,52 +21,83 @@ use telio_sockets::External;
...
 };
 use telio_wg::{DynamicWg, WireGuard};
-use tokio::{net::UdpSocket, sync::Mutex};
+use tokio::{net::UdpSocket, sync::Mutex, time::sleep};
 
 #[cfg(test)]
 use mockall::automock;
 #[cfg(test)]
-const UPNP_INTERVAL: Duration = Duration::from_millis(100);
+const UPNP_INTERVAL_MS: u64 = 100;
```
jjanowsk, line 32: Why to go from Duration to u64  here?

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -18,52 +21,83 @@ use telio_sockets::External;
...
 
 const MAX_SUPPORTED_PACKET_SIZE: usize = 1500;
 #[cfg(not(test))]
-const UPNP_INTERVAL: Duration = Duration::from_secs(5);
-const GET_INTERFACE_TIMEOUT: u64 = 2;
+const UPNP_INTERVAL_MS: u64 = 5000;
```
jjanowsk, line 36: Same for this one

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -132,43 +168,69 @@ impl UpnpEpCommands for IgdGateway {
...
         Ok(())
     }
 
-    fn get_igd_gateway(&mut self) -> Result<()> {
-        self.gw = Some(search_gateway(Default::default())?);
-        Ok(())
+    async fn get_igd_gateway(&mut self) -> Result<()> {
+        if let Some(handle) = &mut self.search_handle {
+            tokio::select! {
+                _ = sleep(Duration::from_secs(GET_IGW_GATEWAY_TIMEOUT_MS)) => {},
```
jjanowsk, line 203: I think you have a bug here, at least judging by the names (secs vs. MS). That's precisely why we should avoid keeping time as u64 :D

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -557,11 +632,10 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> Runtime for State<Wg, I, E> {
                 let _ = self.handle_ping_rx(&buff[..len], &addr).await;
             }
             _ = &mut self.upnp_interval => {
-                if self.check_endpoint_candidate().await.is_ok(){
-                    self.upnp_interval =  PinnedSleep::new(self.exponential_backoff.get_backoff(), ());
-                } else {
+                if self.check_endpoint_candidate().await.is_err() {
```
jjanowsk, line 635: I think this is wrong. You treat 100ms timeout on getting a gateway as an error that will trigger a backoff (If I read it correctly). It works now because of the other bug that makes it 100 second timeout instead of 100ms,  but in general this logic seems to be wrong.

jjanowsk (changes requested):
I think the issue with the timeout and handling of it is a blocker. I need a little bit more time to think how it  should be done correctly, unless I'm missing something. So I'm posting a review already in order for you to be able to respond to my comments in case I've missed sth

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
lcruz99: Added flexibility

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
lcruz99: To match the type defined by the other one.

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
lcruz99: Oh you're right, missed that

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
lcruz99: The triggering of the next backoff on error is intentional, when the search times out (100ms) it gets a `NoIGDGateway` error and the EP check is tried again after the next backoff period.

Even with the 100s timeout, the worst case would be to wait 10s (internal igd::search_gateway() timeout), but as soon as the upnp gw is found (~5ms on tests) it wakes the task and returns it.

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -21,49 +24,80 @@ use telio_utils::{
...
 
 #[cfg(test)]
 use mockall::automock;
 #[cfg(test)]
-const UPNP_INTERVAL: Duration = Duration::from_millis(100);
+const UPNP_INTERVAL_MS: Duration = Duration::from_millis(100);
 
 const MAX_SUPPORTED_PACKET_SIZE: usize = 1500;
 #[cfg(not(test))]
-const UPNP_INTERVAL: Duration = Duration::from_secs(5);
```
lcruz99, line 33: TODO: Fix initial 5s delay caused by this timeout. (STUN EP is facing the same issue)

jjanowsk:
I think there is an underlying issue with the original implementation that this current fix is trying to overcome and hide, but I would consider it a hack instead of the real fix. So the problem I think is that in this place:

lcruz99:
Removed most of the wrapped search task logic from `IgdGateway` struct and added the igd `search_gateway()` call directly to the `tokio::select!` from the main loop.

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -556,12 +587,22 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> Runtime for State<Wg, I, E> {
                 let buff = self.rx_buff.clone();
                 let _ = self.handle_ping_rx(&buff[..len], &addr).await;
             }
-            _ = &mut self.upnp_interval => {
-                if self.check_endpoint_candidate().await.is_ok(){
+            result = self.igd_gw.search_igd_gateway(), if !self.igd_gw.has_igd_gateway() => {
```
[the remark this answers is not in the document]
lcruz99: I imagined that when coding it. I think it's not that important to pause/resume the gw search, if there's no function to execute (`update`), the search will execute while`upnp_interval` doesn't trigger, which might be [5, 10, 20, 40,..] seconds, that is more than reasonable considering we provide default params to the search which has a 10s timeout and waiting in the worst case 5s should be enough. However in the case where we have a function to execute on `update` it's also fine to interrupt the search as once we restart it it just sends another broadcast msg, waiting for one upnp gw to reply..

crates/telio-traversal/src/endpoint_providers/upnp/mod.rs
```diff
@@ -132,47 +135,59 @@ impl UpnpEpCommands for IgdGateway {
...
-        gw.remove_port(PortMappingProtocol::UDP, proxy_port)?;
+        gw.remove_port(PortMappingProtocol::UDP, wg_port).await?;
+        gw.remove_port(PortMappingProtocol::UDP, proxy_port).await?;
         Ok(())
     }
 
-    fn get_igd_gateway(&mut self) -> Result<()> {
-        self.gw = Some(search_gateway(Default::default())?);
+    async fn search_igd_gateway(&mut self) -> Result<()> {
+        self.gw = Some(Box::pin(search_gateway(Default::default())).await?);
```
packgron, line 168: Why is this inner future pinned ?

lcruz99: clippy complains about having a future too large

packgron (approved):
+2.0

packgron (dismissed):

jjanowsk (approved):
+2 on behalf of @packgron

PR #240, opened 2023-12-04
LLT-4583: Don't add IPv6 addresses when IPv6 features is disabled

mathiaspeters:
### Problem
There was an issue where even if the IPv6 feature was disabled, we were still adding the IPv6 addresses of peers and WG-STUN to the wireguard interface.

### Solution
If IPv6 feature is disabled, IPv6 addresses are no longer added to wireguard interface


gytsto (dismissed):
+1

mathiaspeters:
@tomaszklak Good idea. Test added

src/device/mod.rs
```diff
@@ -2670,4 +2670,87 @@ mod tests {
...
+                    p.allowed_ips.iter().any(|ip| match ip {
+                        ipnetwork::IpNetwork::V4(_) => false,
+                        ipnetwork::IpNetwork::V6(_) => true,
+                    })
+                });
+
+            match ipv6 {
+                true => assert!(has_ipv6_address),
+                false => assert!(!has_ipv6_address),
+            }
```
tomaszklak, line 2750: nit: this can be simplified to `assert_eq!(ipv6, has_ipv6_address);`.

src/device/mod.rs
```diff
@@ -2670,4 +2670,87 @@ mod tests {
...
+        }
+
+        run_test(true).await;
+        run_test(false).await;
```
tomaszklak, line 2754: nit: maybe convert this to `rstest` and create separate `case`es?

src/device/mod.rs
```diff
@@ -2517,7 +2517,7 @@ mod tests {
             None,
         )
         .await
-        .unwrap();
+        .unwrap(); // here
```
tomaszklak, line 2520: I think this comment can be removed :)

tomaszklak (approved):
+1.0

src/device/wg_controller.rs
```diff
@@ -360,6 +360,7 @@ async fn build_requested_peers_list<
         upgrade_sync,
         proxy_endpoints,
         remote_peer_states,
+        features,
     )
     .await?;
     let mut exit_node_exists = false;
```
matislovas, line 366: I would add a `ipv6` check for adding `IpNetwork::V6("::/0".parse()?),` to exit node's allowed ips.

mathiaspeters: @matislovas do you mean on line 375?

matislovas: Oh damn! Missed that :smile:  yeah, lgtm than

PR #241, opened 2023-12-05
[LLT-4543] log ffi calls

Hasan6979:
### Problem
To have a better idea on during debugging, it is better if we have an idea of the ffi calls.
#instrument from tracing crate is not used as there is a separate task to add span ids which isn't implemented yet.

### Solution
Log most of ffi calls and their parameters.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-model/src/config.rs
+++ b/crates/telio-model/src/config.rs
@@ -3,6 +3,7 @@
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use serde_json::{from_value, Error, Value};
+use telio_utils::Hidden;
 
 use std::{
     net::{IpAddr, Ipv4Addr},
@@ -19,7 +20,7 @@ pub struct PeerBase {
     /// Public key of the peer
     pub public_key: PublicKey,
     /// Hostname of the peer
-    pub hostname: String,
+    pub hostname: Hidden<String>,
     /// Ip address of peer
     pub ip_addresses: Option<Vec<IpAddr>>,
     /// Nickname for the peer
@@ -177,7 +178,6 @@ impl PartialConfig {
 
 /// Rust representation of [meshnet map]
 /// A network map of all the Peers and the servers
-/// [meshnet map]: https://docs.nordvpn.com/client-api/#get-map
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct Config {
     #[serde(flatten)]
@@ -332,7 +332,7 @@ mod tests {
                 public_key: "qj1pru+cP0mU9K0FrU8e0JYtTaPo0YiQG8O2NbFHeH4="
                     .parse()
                     .unwrap(),
-                hostname: "everest-alice.nord".to_owned(),
+                hostname: telio_utils::Hidden("everest-alice.nord".to_owned()),
                 ip_addresses: Some(vec!["198.51.100.42".parse().unwrap()]),
                 nickname: Some("bunnyg".to_owned()),
             },
@@ -343,7 +343,7 @@ mod tests {
                         public_key: "LRrbraNJXOrVdnpXy6gA/XcpmxymE0oMZlzP5Pqi20I="
                             .parse()
                             .unwrap(),
-                        hostname: "everest-bob.nord".to_owned(),
+                        hostname: telio_utils::Hidden("everest-bob.nord".to_owned()),
                         ip_addresses: Some(vec!["198.51.100.43".parse().unwrap()]),
                         nickname: Some("".to_owned()),
                     },
@@ -357,7 +357,7 @@ mod tests {
                         public_key: "LRrbraNJXOrVdnpXy6gA/XcpmxymE0oMZlzP5Pqi20I="
                             .parse()
                             .unwrap(),
-                        hostname: "everest-alice.nord".to_owned(),
+                        hostname: telio_utils::Hidden("everest-alice.nord".to_owned()),
                         ip_addresses: Some(vec!["198.51.100.43".parse().unwrap()]),
                         nickname: None,
                     },
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -4,7 +4,8 @@ use base64::{decode as base64decode, encode as base64encode};
 use ffi_helpers::{error_handling, panic as panic_handling};
 use ipnetwork::IpNetwork;
 use libc::c_char;
-use telio_crypto::SecretKey;
+use rand::Rng;
+use telio_crypto::{PublicKey, SecretKey};
 use telio_wg::AdapterType;
 use tracing::{error, trace, Subscriber};
 
@@ -30,11 +31,12 @@ use std::{
 
 use self::types::*;
 use crate::device::{Device, DeviceConfig, Result as DevResult};
-use telio_model::{config::PartialConfig, event::*, mesh::ExitNode};
+use telio_model::{api_config::Features, config::PartialConfig, event::*, mesh::ExitNode};
 
 // debug tools
 use telio_utils::{
-    commit_sha, telio_log_debug, telio_log_error, telio_log_trace, telio_log_warn, version_tag,
+    commit_sha, telio_log_debug, telio_log_error, telio_log_info, telio_log_trace, telio_log_warn,
+    version_tag,
 };
 
 const DEFAULT_PANIC_MSG: &str = "libtelio panicked";
@@ -84,7 +86,10 @@ lazy_static::lazy_static! {
 }
 
 #[allow(non_camel_case_types)]
-pub struct telio(Mutex<Device>);
+pub struct telio {
+    inner: Mutex<Device>,
+    id: usize,
+}
 
 /// cbindgen:ignore
 static PANIC_HOOK: Once = Once::new();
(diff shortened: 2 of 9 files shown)


src/ffi/mod.rs
```diff
@@ -681,9 +733,9 @@ pub extern "C" fn telio_disconnect_from_exit_nodes(dev: &telio) -> telio_result
 /// # Parameters
 /// - `cfg`: Output of GET /v1/meshnet/machines/{machineIdentifier}/map
 ///
-pub extern "C" fn telio_set_meshnet(dev: &telio, cfg: *const c_char) -> telio_result {
+pub extern "C" fn telio_set_meshnet(telio: &telio, cfg: *const c_char) -> telio_result {
```
mathiaspeters, line 736: Nit: I think this should go back to `dev`. The `&telio` param in all other functions here is called `dev`, I think we should be consistent, and since we're (finally) getting UniFFI soon, it makes sense to leave as much as possible the same

mathiaspeters:
+1.0

I think it would be nice to add a small description of why this was added to the commit message

src/ffi/mod.rs
```diff
@@ -131,6 +136,9 @@ pub extern "C" fn telio_new(
         TELIO_RES_OK => telio_log_debug!("Operation was succesful (telio_new)"),
     };
 
+    unsafe {
+        telio_log_info!("telio_new entry with instance id: {}. features: {:?}. Log level: {:?}. Event_ptr: {:?}. Logger_ptr: {:?}. Return value: {}", (*(*dev)).id, features, log_level, events, logger, ret);
```
tomaszklak, line 140: I think that if `telio_new_common` returns an error, this will dereference null pointer.

crates/telio-model/src/config.rs
```diff
@@ -26,6 +27,17 @@ pub struct PeerBase {
     pub nickname: Option<String>,
 }
 
+impl fmt::Debug for PeerBase {
```
tomaszklak, line 30: I think that wrapping private fields in `telio_utils::Hidden` is less error prone. It's harder to print log them accidentally and you don't have to write custom debug impls :)

src/ffi/mod.rs
```diff
@@ -401,6 +415,15 @@ pub extern "C" fn telio_get_adapter_luid(dev: &telio) -> u64 {
     }
 }
 
+fn char_ptr_to_type<T: std::str::FromStr>(private_key: *const c_char) -> Result<T, telio_result> {
+    let cstr = unsafe { CStr::from_ptr(private_key) };
+    let private_key = cstr.to_str().map_err(|_| TELIO_RES_INVALID_STRING);
```
tomaszklak, line 420: nit: Adding `?` at the end of this line will make the code simpler since you will get the `key` directly.

src/ffi/mod.rs
```diff
@@ -401,6 +415,15 @@ pub extern "C" fn telio_get_adapter_luid(dev: &telio) -> u64 {
...
+    match private_key {
+        Ok(key) => key.parse().map_err(|_| TELIO_RES_INVALID_STRING),
```
tomaszklak, line 422: nit: Why not at least print the error that we ignore here (and 2 lines above)? This might make any debugging easier :)

src/ffi/mod.rs
```diff
@@ -401,6 +415,15 @@ pub extern "C" fn telio_get_adapter_luid(dev: &telio) -> u64 {
...
+        Err(e) => Err(e),
+    }
+}
+
```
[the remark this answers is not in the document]
Hasan6979: Yep. Initially had written the method just for private_key but after going over the code realized it could be reused for other places as well. Leftover refactoring.

src/ffi/mod.rs
Hasan6979: Doesn't adding `?` resolve that ? Since it is called in a ffi_try! macro which prints if the method returns an error.

tomaszklak: `ffi_try!` applies `?` to the result of `char_ptr_to_type`. As seen in this line the `parse()` error is ignored (bound to `_` and not used in any way) and always `TELIO_RES_INVALID_STRING` is used as a resulting error. This the `?` will print `TELIO_RES_INVALID_STRING`.

src/ffi/mod.rs
```diff
@@ -448,7 +475,12 @@ pub extern "C" fn telio_get_private_key(dev: &telio) -> *mut c_char {
 ///
 pub extern "C" fn telio_set_fwmark(dev: &telio, fwmark: c_uint) -> telio_result {
     ffi_catch_panic!({
-        let dev = ffi_try!(dev.0.lock().map_err(|_| TELIO_RES_LOCK_ERROR));
+        telio_log_info!(
```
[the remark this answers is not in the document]
tomaszklak: The LLT-4543 in point iv say "Log return value of the function", if this is no longer desired please update the task description.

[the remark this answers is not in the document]
Hasan6979: Done!

src/ffi/mod.rs
tomaszklak: This is not solved, you are now checking 'outer' pointer which is passed to us from applications (a place where we should store the pointer to telio instance) but you are not checking the inner pointer which we do set. Considering how error prone this is, please add a test that verifies this situation - that we will not crash if `telio_new_common` returns an error (and doesn't set the address where telio instance is located).

src/ffi/mod.rs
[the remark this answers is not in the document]
tomaszklak: Not sure how it is done, can you clarify?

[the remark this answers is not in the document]
Hasan6979: I updated the ticket.

src/ffi/mod.rs
```diff
@@ -610,12 +677,17 @@ pub extern "C" fn telio_enable_magic_dns(
     dev: &telio,
     forward_servers: *const c_char,
 ) -> telio_result {
+    let servers_str = ffi_try!(unsafe { CStr::from_ptr(forward_servers) }
```
tomaszklak, line 680: do we know that `forward_servers` is not null  here?

src/ffi/mod.rs
```diff
@@ -1136,10 +1231,20 @@ mod tests {
...
+
+    #[test]
+    fn test_logging_when_telio_dev_empty() {
+        let telio_dev: *mut *mut telio = ptr::null_mut();
+        let res = get_instance_from_ptr(telio_dev);
+        assert_eq!(res, None);
+        let telio_dev: *mut *mut telio = Box::into_raw(Box::new(ptr::null_mut()));
+        let res = get_instance_from_ptr(telio_dev);
+        assert_eq!(res, None);
+    }
```
tomaszklak, line 1249: Please also add a positive case when we do get an id stored inside.

src/ffi/mod.rs
```diff
@@ -111,30 +116,37 @@ pub extern "C" fn telio_new(
...
-        TELIO_RES_INVALID_STRING => telio_log_debug!("Cannot parse a string (telio_new)"),
-        TELIO_RES_ERROR => telio_log_debug!("Unknown error (telio_new)"),
-        TELIO_RES_OK => telio_log_debug!("Operation was succesful (telio_new)"),
-    };
-
+    log_entry(features, events, log_level, logger, ret, dev);
     ret
 }
 
+fn char_to_str(features: *const c_char) -> Result<&'static str, telio_result> {
```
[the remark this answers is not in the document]
Hasan6979: Fixed!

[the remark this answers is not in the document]
Hasan6979: Fixed!

src/ffi/mod.rs
```diff
@@ -152,12 +164,47 @@ pub extern "C" fn telio_new_with_protect(
...
+        None
+    } else {
+        let inner = unsafe { *dev };
+        if inner.is_null() {
+            None
+        } else {
+            Some(unsafe { (*inner).id })
+        }
+    }
+}
```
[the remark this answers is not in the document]
Hasan6979: Done

src/ffi/mod.rs
```diff
@@ -152,12 +164,47 @@ pub extern "C" fn telio_new_with_protect(
...
     protect: telio_protect_cb,
 ) -> telio_result {
-    telio_new_common(dev, features, events, log_level, logger, Some(protect))
+    let features = ffi_try!(deserialize_features(features));
+    let ret = telio_new_common(dev, &features, events, log_level, logger, Some(protect));
+    log_entry(features, events, log_level, logger, ret, dev);
+    ret
+}
+
+fn get_instance_from_ptr(dev: *mut *mut telio) -> Option<usize> {
```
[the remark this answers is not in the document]
Hasan6979: typo. Missed writing ID. Fixed it

src/ffi/mod.rs
Hasan6979: We don't. I had just changed the order. Added a check.

src/ffi/mod.rs
Hasan6979: Done

tomaszklak (approved):
+1.0

PR #244, opened 2023-12-08
natlab: Remove alpha asserts after network switch

lcruz99:
### Problem
`test_mesh_network_switch_direct` triggers network switch on node alpha and then strictly requires node events transitioning direct->relay and later vice-versa relay->direct, but due to the fact that beta node has not changed its endpoint, alpha node may trigger roaming in beta (by sending any packet to beta) restoring direct connection without the need for direct->relay and relay->direct cycle. The presence of this cycle is inherently racy.

### Solution
Remove alpha's assertions after network switch, as beta's are enough.

- [x] README.md is updated
- [x] changelog.md is updated


dfetti (approved):
+2.0

PR #246, opened 2023-12-11
Fix crash in wireguard-go

jjanowsk:
Wireguard-go could crash if winrio initization fails. That is because in that case StdNetBind is returned from NewDefaultBind function instead of WinRingBind. Later this object was casted to the WinRingBind in order to call OpenOnLocalhost method. This commit adds new method OpenOnLocalhost to our interface FullBind and implements is for both WinRingBind and StdNetBind. It requires us to copy StdNetBind implementation for the upstream.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/wireguard-go-rust-wrapper/wireguard-go/bind_windows.go
+++ b/wireguard-go-rust-wrapper/wireguard-go/bind_windows.go
@@ -82,11 +82,12 @@ type WinRingBind struct {
 	isOpen uint32
 }
 
-func NewDefaultBind() conn.Bind { return NewWinRingBind() }
+func NewDefaultBind() FullBind { return NewWinRingBind() }
 
-func NewWinRingBind() conn.Bind {
+func NewWinRingBind() FullBind {
 	if !winrio.Initialize() {
-		return conn.NewStdNetBind()
+		warningf("Failed to initialize winrio. Falling back to StdNetBind implementation.")
+		return NewStdNetBind()
 	}
 	return new(WinRingBind)
 }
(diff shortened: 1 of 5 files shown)


wireguard-go-rust-wrapper/wireguard-go/platform_specific_windows.go
```diff
@@ -73,7 +74,7 @@ func (b Binder) Open(port uint16) (fns []conn.ReceiveFunc, actualPort uint16, er
 		return recv_fns, actual_port, err
 	}
 
-	recv_fns_local, actual_port, err := b.local_bind.(*WinRingBind).OpenOnLocalhost(port)
+	recv_fns_local, actual_port, err := b.local_bind.OpenOnLocalhost(port)
```
LukasPukenis, line 77: I am wondering if this can be simplified without copy-pasting in the code from the upstream. A check can be made in Go at runtime to see if `local_bind` is `WinRingBind`, assuming it is not then it's an errorneous case where the initialization has failed and thus we could handle it(instead of crashing)?

jjanowsk: But we want it to work in case of failed initialization. Not just fail gracefully. That's why OpenOnLocalhost is added to `StdNetBind`. And in order to add this implementation we need a copy of it unfortunately.

LukasPukenis: I guess it makes sense then, if it still wants to do `OpenOnLocalhost`  then these changes make sense.

+1.0

wireguard-go-rust-wrapper/wireguard-go/bind_windows.go
```diff
@@ -82,11 +82,11 @@ type WinRingBind struct {
...
 }
 
-func NewDefaultBind() conn.Bind { return NewWinRingBind() }
+func NewDefaultBind() FullBind { return NewWinRingBind() }
 
-func NewWinRingBind() conn.Bind {
+func NewWinRingBind() FullBind {
 	if !winrio.Initialize() {
-		return conn.NewStdNetBind()
+		return NewStdNetBind()
```
Jauler, line 89: :thinking: could we log somehow which bind is being used? WinRingBind or StdNetBind? 

I believe this may have performance implications, and may cause headaches trying to understand them. If we would have them logged -> this would help immensely :). AFAIK, go code logging is configured so :crossed_fingers: should not be difficult to log it from here.

Jauler (dismissed):
+2.0 assuming, the logs can be sorted out :)

Jauler (approved):
+2.0

PR #247, opened 2023-12-11
LLT-4614: implement wg post quantum handshake in libtelio codebase

matszczygiel:
### Problem
We want to make the VPN tunnel post-quantum safe.

### Solution
The solution performs an initial wireguard connection during which a preshared key is established as the shared secret of the kyber768 key exchange mechanism.
The preshared key is used for the subsequent wireguard tunnel being created just after the  initial handshake

There isn't any unit or integration test because that's a separate task and @LukasPukenis is responsible for this one.
The PR doesn't contain the rekey mechanism for similar reasons, it's going to be implemented in subsequent PRs.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-crypto/src/lib.rs
+++ b/crates/telio-crypto/src/lib.rs
@@ -81,7 +81,8 @@ impl SecretKey {
         Self::gen_with(&mut rand::rngs::StdRng::from_entropy())
     }
 
-    pub(crate) fn gen_with(rng: &mut (impl RngCore + CryptoRng)) -> Self {
+    /// Generates a new random SecretKey with provided RNG
+    pub fn gen_with(rng: &mut (impl RngCore + CryptoRng)) -> Self {
         let mut key = SecretKey([0u8; KEY_SIZE]);
         rng.fill_bytes(&mut key.0);
         // Key clamping
--- a/crates/telio-crypto/src/lib.rs
+++ b/crates/telio-crypto/src/lib.rs
@@ -27,7 +27,7 @@ use std::{convert::TryInto, fmt};
 use rand::prelude::*;
 use serde_with::{DeserializeFromStr, SerializeDisplay};
 
-/// Secret/Public key size in bytes
+/// Secret, Public and Wireguard Preshared key size in bytes
 pub const KEY_SIZE: usize = 32;
 
 /// Secret key type
@@ -41,6 +41,12 @@ pub struct SecretKey([u8; KEY_SIZE]);
     Default, PartialOrd, Ord, PartialEq, Eq, Hash, Copy, Clone, DeserializeFromStr, SerializeDisplay,
 )]
 pub struct PublicKey(pub [u8; KEY_SIZE]);
+
+/// Preshared key type
+#[derive(
+    Default, PartialOrd, Ord, PartialEq, Eq, Hash, Copy, Clone, DeserializeFromStr, SerializeDisplay,
+)]
+pub struct PresharedKey(pub [u8; KEY_SIZE]);
 
 /// Error returned when parsing fails for SecretKey or PublicKey.
 #[derive(Debug, thiserror::Error)]
@@ -123,6 +129,13 @@ impl SecretKey {
 }
 
 impl PublicKey {
+    /// Create new key from bytes
+    pub const fn new(bytes: [u8; 32]) -> Self {
+        Self(bytes)
+    }
+}
+
+impl PresharedKey {
     /// Create new key from bytes
     pub const fn new(bytes: [u8; 32]) -> Self {
         Self(bytes)
@@ -282,7 +295,7 @@ macro_rules! gen_common {
         gen_common!($($tt),+);
     };
 }
-gen_common!(SecretKey, PublicKey);
+gen_common!(SecretKey, PublicKey, PresharedKey);
 
 #[cfg(test)]
 mod tests {
--- a/crates/telio-wg/src/pq.rs
+++ b/crates/telio-wg/src/pq.rs
@@ -190,6 +190,12 @@ async fn handshake(
     Ok(TunnelSock { tunn, sock })
 }
 
+/// Public function exposed for fuzzing framework
+#[cfg(feature = "fuzzing")]
+pub fn parse_get_response_fuzz(pkgbuf: &[u8]) -> Result<kyber768::Ciphertext, Error> {
+    parse_get_response(pkgbuf)
+}
+
 fn parse_get_response(pkgbuf: &[u8]) -> Result<kyber768::Ciphertext, Error> {
     let mut cipherbuf = [0; CIPHERTEXT_LEN as usize];
 
--- a/crates/telio-wg/src/pq.rs
+++ b/crates/telio-wg/src/pq.rs
@@ -1,0 +1,354 @@
+use std::{
+    collections::btree_map::Range,
+    convert::TryInto,
+    io::{self, Read, Write},
+    net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
+    ops::RangeInclusive,
+    time::Duration,
+};
+
+use boringtun::noise;
+use pnet_packet::{
+    ip::{self, IpNextHeaderProtocols},
+    ipv4::{self, Ipv4Flags, Ipv4Packet, MutableIpv4Packet},
+    udp::{self, MutableUdpPacket, UdpPacket},
+    Packet,
+};
+use pqcrypto_kyber::kyber768;
+use pqcrypto_traits::kem::{Ciphertext, PublicKey, SharedSecret};
+use rand::{prelude::Distribution, SeedableRng};
+use tokio::net::{ToSocketAddrs, UdpSocket};
+
+const SERVICE_PORT: u16 = 6480;
+const LOCAL_PORT_RANGE: RangeInclusive<u16> = 49152..=u16::MAX; // dynamic port range
+const LOCAL_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 2);
+const REMOTE_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 1);
+const HANDSHAKE_VERSION: u32 = 1;
+const CIPHERTEXT_LEN: u32 = kyber768::ciphertext_bytes() as _;
+
+const IPV4_HEADER_LEN: usize = 20;
+const UDP_HEADER_LEN: usize = 8;
+
+/// The PQ module error type
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// IO error
+    #[error("IO: {0:?}")]
+    Io(#[from] io::Error),
+    /// Generic unrecoverable error
+    #[error("Generic: {0}")]
+    Generic(String),
+}
+
+impl From<String> for Error {
+    fn from(value: String) -> Self {
+        Self::Generic(value)
+    }
+}
+
+impl From<&str> for Error {
+    fn from(value: &str) -> Self {
+        Self::Generic(value.to_string())
+    }
+}
+
+impl From<tokio::time::error::Elapsed> for Error {
+    fn from(value: tokio::time::error::Elapsed) -> Self {
+        Self::Io(io::Error::new(io::ErrorKind::TimedOut, value))
+    }
+}
+
+/// Post quantum keys retrived from hanshake
+#[derive(Clone, Copy)]
+pub struct PqKeys {
+    /// Kyber shared secret
+    pub pq_shared: telio_crypto::PresharedKey,
+    /// X25519 secret key
+    pub wg_secret: telio_crypto::SecretKey,
+}
+
+struct TunnelSock {
+    tunn: Box<noise::Tunn>,
+    sock: telio_sockets::External<UdpSocket>,
+}
+
+/// Get PQ keys from the VPN server
+/// The packet sequence is as follows
+///
+/// 1) C -> S             : Client sends WG hansake packet                           
+/// 2) S -> C             : Server responds to handsake                              
+/// 3) C -> S (in tunnel) : Client sends a UDP PQ GET packet with keys               
+/// 4) S -> C (in tunnel) : Server responds with UDP packet containing the ciphertext
+///
+/// The shared key is established which should be used as,
+/// a WG preshared key for subsequent connection
+pub async fn fetch_keys(
+    sock_pool: &telio_sockets::SocketPool,
+    endpoint: impl ToSocketAddrs,
+    secret: &telio_crypto::SecretKey,
+    peers_pubkey: &telio_crypto::PublicKey,
+) -> Result<PqKeys, Error> {
+    let TunnelSock { tunn, sock } = handshake(sock_pool, endpoint, secret, peers_pubkey).await?;
+
+    let mut rng = rand::rngs::StdRng::from_entropy();
+
+    // Generate keys
+    let wg_secret = telio_crypto::SecretKey::gen_with(&mut rng);
+    let (pq_public, pq_secret) = kyber768::keypair();
+    let wg_public = telio_crypto::PublicKey::from(&wg_secret);
+
+    // Send GET packet
+    let pkgbuf = create_get_packet(&wg_public, &pq_public, &mut rng); // 4 KiB
+
+    let mut recvbuf = [0u8; 2048]; // 2 KiB buffer should suffice
+    match tunn.encapsulate(&pkgbuf, &mut recvbuf) {
+        noise::TunnResult::Err(err) => {
+            return Err(format!("Failed to encapsulate PQ keys message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToNetwork(buf) => {
+            sock.send(buf).await?;
+        }
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    // Receive response
+    let read = sock.recv(&mut recvbuf).await?;
+
+    let mut msgbuf = [0u8; 2048]; // 2 KiB buffer should shuffice
+
+    #[allow(index_access_check)]
+    let ciphertext = match tunn.decapsulate(None, &recvbuf[..read], &mut msgbuf) {
+        noise::TunnResult::Err(err) => {
+            return Err(format!("Failed to decapsulate PQ keys message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToTunnelV4(buf, _) => parse_get_response(buf)?,
+        _ => return Err("Unexpected WG tunnel output".into()),
+    };
+
+    // Extract the shared secret
+    let pq_shared = kyber768::decapsulate(&ciphertext, &pq_secret);
+    let pq_shared = telio_crypto::PresharedKey::new(
+        pq_shared
+            .as_bytes()
+            .try_into()
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))?,
+    );
+
+    Ok(PqKeys {
+        pq_shared,
+        wg_secret,
+    })
+}
+
+async fn handshake(
+    sock_pool: &telio_sockets::SocketPool,
+    endpoint: impl ToSocketAddrs,
+    secret: &telio_crypto::SecretKey,
+    peers_pubkey: &telio_crypto::PublicKey,
+) -> Result<TunnelSock, Error> {
+    let sock = sock_pool
+        .new_external_udp((Ipv4Addr::UNSPECIFIED, 0), None)
+        .await?;
+    sock.connect(endpoint).await?;
+
+    let tunn = noise::Tunn::new(
+        secret.into_bytes().into(),
+        peers_pubkey.0.into(),
+        None,
+        None,
+        0,
+        None,
+    )?;
+
+    let mut pkgbuf = [0u8; 2048];
+    match tunn.encapsulate(&[], &mut pkgbuf) {
+        noise::TunnResult::Err(err) => {
+            return Err(format!("Failed to encapsulate handshake message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToNetwork(buf) => {
+            sock.send(buf).await?;
+        }
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    // The response should be 92, so the buffer is sufficient
+    let read = sock.recv(&mut pkgbuf).await?;
+
+    let mut msgbuf = [0u8; 2048];
+
+    #[allow(index_access_check)]
+    match tunn.decapsulate(None, &pkgbuf[..read], &mut msgbuf) {
+        noise::TunnResult::Err(err) => {
+            return Err(format!("Failed to decapsulate handshake message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToNetwork(_) => {
+            // This is a outgoing keep alive message, we can skip sending it for the hanshake
+        }
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    Ok(TunnelSock { tunn, sock })
+}
+
+fn parse_get_response(pkgbuf: &[u8]) -> Result<kyber768::Ciphertext, Error> {
+    let mut cipherbuf = [0; CIPHERTEXT_LEN as usize];
+
+    let ip = Ipv4Packet::new(pkgbuf).ok_or(io::Error::new(
+        io::ErrorKind::InvalidData,
+        "Invalid PQ keys IP packet received",
+    ))?;
+
+    let udp = UdpPacket::new(ip.payload()).ok_or(io::Error::new(
+        io::ErrorKind::InvalidData,
+        "Invalid PQ keys UDP packet received",
+    ))?;
+
+    let mut data = io::Cursor::new(udp.payload());
+
+    let mut version = [0u8; 4];
+    data.read_exact(&mut version)?;
+    let version = u32::from_le_bytes(version);
+    if version != HANDSHAKE_VERSION {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            "Server responded with invalid PQ handshake version",
+        )
+        .into());
+    }
+
+    let mut ciphertext_len = [0u8; 4];
+    data.read_exact(&mut ciphertext_len)?;
+    let ciphertext_len = u32::from_le_bytes(ciphertext_len);
+
+    if ciphertext_len != CIPHERTEXT_LEN {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            "Server responded with invalid PQ handshake ciphertext lenght",
+        )
+        .into());
+    }
+
+    data.read_exact(&mut cipherbuf)?;
+
+    let ct = kyber768::Ciphertext::from_bytes(&cipherbuf)
+        .map_err(|err| format!("Invalid PQ ciphertext received: {err:?}"))?;
+
+    Ok(ct)
+}
+
+fn create_get_packet(
+    wg_public: &telio_crypto::PublicKey,
+    pq_public: &kyber768::PublicKey,
+    rng: &mut impl rand::Rng,
+) -> Vec<u8> {
+    let mut pkgbuf = Vec::with_capacity(1024 * 4); // 4 KiB
+    pkgbuf.resize(IPV4_HEADER_LEN + UDP_HEADER_LEN, 0);
+
+    push_get_method_udp_payload(&mut pkgbuf, wg_public, pq_public);
+    fill_get_packet_headers(pkgbuf.as_mut_slice(), rng);
+
+    pkgbuf
+}
+
+/// The payload looks as follows:
+///
+/// ---------------------------------
+///  version           , u32le , = 1
+/// ---------------------------------
+///  method            , u32le , = 0
+/// ---------------------------------
+///  WG pubkey len     , u32le , = 32
+/// ---------------------------------
+///  WG pubkey bytes   , [u8]
+/// ---------------------------------
+///  Kyber pubkey len  , u32le, = 1184
+/// ---------------------------------
+///  Kyber pubkey bytes, [u8]
+/// ---------------------------------
+fn push_get_method_udp_payload(
+    pkgbuf: &mut Vec<u8>,
+    wg_public: &telio_crypto::PublicKey,
+    pq_public: &kyber768::PublicKey,
+) {
+    let method = 0u32; // get
+
+    // UDP packet payload
+    pkgbuf.extend_from_slice(&HANDSHAKE_VERSION.to_le_bytes());
+    pkgbuf.extend_from_slice(&method.to_le_bytes());
+    pkgbuf.extend_from_slice(&(wg_public.len() as u32).to_le_bytes());
+    pkgbuf.extend_from_slice(wg_public);
+    pkgbuf.extend_from_slice(&(pq_public.as_bytes().len() as u32).to_le_bytes());
+    pkgbuf.extend_from_slice(pq_public.as_bytes());
+}
+
+/// Sets up UDP and IP headers in the provided buffer
+///
+/// # Panics
+///
+/// Panics if the buffer size is less than IP + UDP headers bytes.
+fn fill_get_packet_headers(pkgbuf: &mut [u8], rng: &mut impl rand::Rng) {
+    let pkg_len = pkgbuf.len();
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut udppkg = MutableUdpPacket::new(&mut pkgbuf[IPV4_HEADER_LEN..])
+        .expect("UDP buffer should not be too small");
+    udppkg.set_source(random_port(rng));
+    udppkg.set_destination(SERVICE_PORT);
+    udppkg.set_length((pkg_len - IPV4_HEADER_LEN) as _);
+    udppkg.set_checksum(udp::ipv4_checksum(
+        &udppkg.to_immutable(),
+        &LOCAL_IP,
+        &REMOTE_IP,
+    ));
+    drop(udppkg);
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut ippkg = MutableIpv4Packet::new(&mut pkgbuf[..IPV4_HEADER_LEN])
+        .expect("IPv4 buffer should not be too small");
+    ippkg.set_version(4);
+    ippkg.set_header_length(5);
+    ippkg.set_total_length(pkg_len as _);
+    ippkg.set_flags(Ipv4Flags::DontFragment);
+    ippkg.set_ttl(0xFF);
+    ippkg.set_next_level_protocol(IpNextHeaderProtocols::Udp);
+    ippkg.set_source(LOCAL_IP);
+    ippkg.set_destination(REMOTE_IP);
+    ippkg.set_checksum(ipv4::checksum(&ippkg.to_immutable()));
+}
+
+fn random_port(rng: &mut impl rand::Rng) -> u16 {
+    rand::distributions::Uniform::new_inclusive(LOCAL_PORT_RANGE.start(), LOCAL_PORT_RANGE.end())
+        .sample(rng)
+}
+
+#[cfg(test)]
+mod tests {
+    use pqcrypto_kyber::kyber768;
+    use pqcrypto_traits::kem::{Ciphertext, SecretKey, SharedSecret};
+
+    #[test]
+    // The Kyber KEM has two competing algorithms and the library we're using changed the algorithm in their new version.
+    // The test here is for capturing the eventual regression when the library is updated for example.
+    fn decapsulate_kyber_ciphertext() {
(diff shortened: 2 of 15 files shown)


src/device/mod.rs
```diff
@@ -1460,12 +1463,37 @@ impl Runtime {
             .map(|peers| peers.iter().any(|p| p.public_key == exit_node.public_key))
             .unwrap_or_default();
 
+        self.requested_state.postquantum_wg.take();
```
tomaszklak, line 1466: Not sure why is this `.take();` and not just ` = None;` - since you are not using return value this seems to be needlessly confusing.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    let mut msgbuf = [0u8; 2048];
+
+    #[allow(index_access_check)]
+    match tunn.decapsulate(None, &pkgbuf[..read], &mut msgbuf) {
+        noise::TunnResult::Err(err) => {
+            return Err(format!("Failed to decapsulate handshake message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToNetwork(_) => {
+            // This is a keep alive message, we can ignore it for the hanshake
```
tomaszklak, line 259: handshake i a data massage with length 0, we are not checking length here so it seems that either comment is imprecise (it should mention any data message, not only keepalive) or there should be a change to match/condition.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+    udp::{self, MutableUdpPacket, UdpPacket},
+    Packet,
+};
+use pqcrypto_kyber::kyber768;
+use pqcrypto_traits::kem::{Ciphertext, PublicKey, SharedSecret};
+use rand::{prelude::Distribution, SeedableRng};
+use tokio::net::{ToSocketAddrs, UdpSocket};
+
+const SERVICE_PORT: u16 = 6480;
+const LOCAL_PORT_RANGE: RangeInclusive<u16> = 1024..=u16::MAX; // Unprivileged port range
```
tomaszklak, line 23: nit: maybe this should be limited to dynamic port range (49152 to 65535)?

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    let payload = std::iter::empty()
+        .chain(HANDSHAKE_VERSION.to_le_bytes())
+        .chain(method.to_le_bytes())
+        .chain((wg_public.as_bytes().len() as u32).to_le_bytes())
+        .chain(wg_public.to_bytes())
+        .chain((pq_public.as_bytes().len() as u32).to_le_bytes())
+        .chain(pq_public.as_bytes().iter().copied());
+
+    pkgbuf.extend(payload);
```
tomaszklak, line 114: nit: I think this is inefficient - it would be better to use `extend_from_slice`

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+            return Err(format!("Failed to encapsulate handshake message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToNetwork(buf) => {
+            sock.send(buf).await?;
+        }
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    // The response should be 92, so the buffer is sufficient
+    let read = sock.recv(&mut pkgbuf).await?;
```
tomaszklak, line 249: should this be wrapped in timeout?

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
...VhafidbPe4tFDFJdtDhQhNtjCK/VS0vDrKmSErWaKBhYSgKxYDtCDB/0CdKnC6EJwmoJUSLMtw/eFNCll3KFOE51EKkXg6KuYQiOzCIoE5P8w3LyZotUm9gohhXvsl8pZD9oFsl0YQUEd6Qqkbrqa8gSdL7ABWwpI+b1IdTYYB68QgU6OkpnMt7wXEGaxsnYY1aYWJARk+vkdmaLNFsCoxcOnFv6Z24cmfGBEe0LB/HsV2DREmzbUSIcVwjZsGOLMqcsCq+bCD8eWEeIu8wamiwUGmGVMTwFOxMjMPEBKhvLa2UthpEeEuPLwNjbg0AtyCL4p0/5Qo2tS0YphM+5GlrCJkjVwbfHmAnHivXlzCK9hNchqo5URXlHJOjehD2WSvBIvPQCi9IYAgSMAd/BNrq1eDCCzAOoali5QswxPD22AbHfrBkBWXy9JTyZJCExwRP5ZVTjIPSGKC++uVBce8aDZ+kVKPP1GOsXqxVikC0xVOfyMTCDN2/qSUMAcBMCaiEgSTFZdyyvgrHsVntJOHw9h1HaupMxuXuSNo5ms2zOCaTOiKR4tZYSMCxTaRLNRlLnPEHvsRxyFE1DpmIIPKn3KUt9kaUWM+ovwJy4tHu4nMUPJfsGem+ccsQdDE80uwJtLFeUC845FQC+HLoLecoNzOfhghxVcGhsyEUUIbRKmlOgEtI7OpmuSWB9xqocirBTx05ZI/f5lzqtgUXRiRx9wxCgoFZlY/33o2qieSLxlV5afEYCuKqIxE93aQmvcNkDwZbyNOD/IJV2g+3Ks4JEqlR0YLqPNyj4qomzp+IOY/hDgeftoJUKcxTUWDZXYmuIAjH1CSBgm4l+o9hHWGl1pl3LIGn1KNMhK5A1yNX3AqCHl6J8gjAcw3XKSmMMRZPEdMlFgVTLcicWSd6iZ1NyvO+fM8bkOKR7Gn6ekPzspztVtwl1ayzYXLMWRUdOhqtWtukCarIEXFyFG8ByKlsVQCRvKMSUCjp0eRHoIP/SnMGUufs1ZCM5ejwXZR/EK6XoNmyNQUrrVNGuB/2sm9Z/FLEEwShFiVlYMXvGFcZubPr03TqybQWtkXQ0KQxvxhLgLsk495B3i13F342yT5m1OX3Wor1JNyV6nAolgf/IDQGyXh9fBzETTooRcbl8jLfllhVDimPrZMw7488eXXw9qet8DIjUCP4UQ3B4R/jsSx";
```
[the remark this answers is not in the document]
matszczygiel: Um, why?

src/device/mod.rs
matszczygiel: Yeah, I can change this one. However, it doesn't matter IMHO

crates/telio-wg/src/pq.rs
matszczygiel: Sure, let me change this one

crates/telio-wg/src/pq.rs
matszczygiel: I wanted to have a clear book-like prescription of how the packet looks like. But I can change this one as well

crates/telio-wg/src/pq.rs
matszczygiel: Yeah, I was thinking about it. However, I decided to wrap the whole method in the timeout on the higher level

crates/telio-wg/src/pq.rs
matszczygiel: I meant that since we haven't provided any data to the `tunn`, there can only be a keepalive message available to send, which we don't need for the (post-quantum) handshake.

crates/telio-wg/src/pq.rs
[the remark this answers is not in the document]
tomaszklak: it's basically random blob of bytes with no obvious significance. Furthermore if the test fails I won't be able to understand why by reading that string - I will have to use some tool (maybe a rust fn or an external tool) on it, which I think would be simpler if if was included from an external file. If you feel strongly that this is not the case, feel free to ignore this comment, I forgot to mark it as nitpick :)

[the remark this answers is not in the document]
matszczygiel: The thing is that these are the test vectors. I wouldn't include this, but the Kyber kem has two competing algorithms and the library I'm using changed the algorithm in their new version.
The test here is for capturing the eventual regression when the library is updated for example.
So to sum up, I don't think you would need to inspect the test vectors, IMHO cluttering the file structure of the project is worse than having quite a long string inlined, thus I will keep it as is :)

crates/telio-wg/src/pq.rs
tomaszklak: I think this is still going to be pretty clear:
```rust
pkgbuf.extend_from_slice(&HANDSHAKE_VERSION.to_le_bytes());
pkgbuf.extend_from_slice(&method.to_le_bytes());
pkgbuf.extend_from_slice(&(wg_public.as_bytes().len() as u32).to_le_bytes());
pkgbuf.extend_from_slice(&wg_public.to_bytes());
pkgbuf.extend_from_slice(&(pq_public.as_bytes().len() as u32).to_le_bytes());
pkgbuf.extend_from_slice(&pq_public.as_bytes());
```

src/device/mod.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// Get PQ keys from the VPN server
+pub async fn fetch_keys(
+    endpoint: impl ToSocketAddrs,
+    secret: &telio_crypto::SecretKey,
+    peers_pubkey: &telio_crypto::PublicKey,
+) -> Result<PqKeys, Error> {
+    let (tunn, sock) = handshake(endpoint, secret, peers_pubkey).await?;
+
+    let mut rng = rand::rngs::StdRng::from_entropy();
```
stalowyjez, line 92: I think that adding some helper functions, or splitting this in some other way would improve the readability

crates/telio-wg/src/pq.rs
matszczygiel: I've reworded the comment a bit

crates/telio-crypto/src/lib.rs
```diff
@@ -42,6 +42,12 @@ pub struct SecretKey([u8; KEY_SIZE]);
 )]
 pub struct PublicKey(pub [u8; KEY_SIZE]);
 
+/// Preshared key type
+#[derive(
+    Default, PartialOrd, Ord, PartialEq, Eq, Hash, Copy, Clone, DeserializeFromStr, SerializeDisplay,
+)]
+pub struct PresharedKey(pub [u8; KEY_SIZE]);
```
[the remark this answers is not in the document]
matszczygiel: This is the preshared key in the wireguard sense https://www.wireguard.com/papers/wireguard.pdf#subsection.5.2 and I wanted to have a strongly typed equivalent of this one.
I'll update the `KEY_SIZE` comment accordingly.

crates/telio-wg/src/uapi.rs
```diff
@@ -81,6 +88,7 @@ impl From<set::Peer> for Peer {
                 .map(|ip| IpNetwork::new(ip.ipaddr, ip.cidr_mask))
                 .collect::<Result<Vec<IpNetwork>, _>>()
                 .unwrap_or_default(),
+            preshared_key: item.preshared_key.map(PresharedKey),
```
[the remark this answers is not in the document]
matszczygiel: No, preshared key is always optional. It was unused by the libtelio implementation, however, for the post-quantum tunnel it is required.

src/device/wg_controller.rs
```diff
@@ -765,6 +777,7 @@ fn compare_peers(a: &telio_wg::uapi::Peer, b: &telio_wg::uapi::Peer) -> bool {
         && a.endpoint == b.endpoint
         && a.persistent_keepalive_interval == b.persistent_keepalive_interval
         && a.allowed_ips == b.allowed_ips
+        && a.preshared_key == b.preshared_key
```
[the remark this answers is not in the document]
matszczygiel: Wireguard interface is always using it but reserves special value for when it's not used by the tunnel. The special value is all zeros

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+            let sk_bytes = base64::decode(secret_key).unwrap();
+            let secret_key = kyber768::SecretKey::from_bytes(&sk_bytes).unwrap();
+
+            let ct_bytes = base64::decode(ciphertext).unwrap();
+            let ciphertext = kyber768::Ciphertext::from_bytes(&ct_bytes).unwrap();
+
+            let shared = kyber768::decapsulate(&ciphertext, &secret_key);
+
+            base64::encode(shared.as_bytes())
+        };
```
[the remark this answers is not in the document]
matszczygiel: I've explained the reason behind this test here https://github.com/NordSecurity/libtelio/pull/247#discussion_r1424034213.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+        &udppkg.to_immutable(),
+        &LOCAL_IP,
+        &REMOTE_IP,
+    ));
+    drop(udppkg);
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut ippkg = MutableIpv4Packet::new(&mut pkgbuf[..IPV4_HEADER_LEN])
+        .expect("IPv4 buffer should not be too small");
```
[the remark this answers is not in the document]
matszczygiel: I disagree. The code statically guarantees the minimum buffer size

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+    ippkg.set_version(4);
+    ippkg.set_header_length(5);
+    ippkg.set_total_length(pkg_len as _);
+    ippkg.set_flags(Ipv4Flags::DontFragment);
+    ippkg.set_ttl(0xFF);
+    ippkg.set_next_level_protocol(IpNextHeaderProtocols::Udp);
+    ippkg.set_source(LOCAL_IP);
+    ippkg.set_destination(REMOTE_IP);
+    ippkg.set_checksum(ipv4::checksum(&ippkg.to_immutable()));
+    drop(ippkg);
```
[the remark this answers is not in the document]
matszczygiel: Um, no. What do you propose instead?

crates/telio-wg/src/pq.rs
matszczygiel: What do you mean exactly? The `handshake()` is already extracted as a helper function, do have something specific in mind?

crates/telio-crypto/src/lib.rs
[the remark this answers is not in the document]
matszczygiel: Comment updated

tomaszklak (dismissed):
+1.0

crates/telio-wg/src/pq.rs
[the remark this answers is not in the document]
LukasPukenis: Chaining like `new().set_version(4).set_...;` should eliminate the `drop`

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    let mut recvbuf = [0u8; 2048]; // 2 KiB buffer should shuffice
```
LukasPukenis, line 147:
```suggestion
    let mut recvbuf = [0u8; 2048]; // 2 KiB buffer should suffice
```

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+    // Pepare payload
+    let method = 0u32; // get
+
+    // UDP packet payload
+    pkgbuf.extend_from_slice(&HANDSHAKE_VERSION.to_le_bytes());
+    pkgbuf.extend_from_slice(&method.to_le_bytes());
+    pkgbuf.extend_from_slice(&(wg_public.as_bytes().len() as u32).to_le_bytes());
+    pkgbuf.extend_from_slice(&wg_public.to_bytes());
+    pkgbuf.extend_from_slice(&(pq_public.as_bytes().len() as u32).to_le_bytes());
+    pkgbuf.extend_from_slice(pq_public.as_bytes());
```
LukasPukenis, line 114: Would it be possible to draw a ASCII diagram of the payload structure? I know it can be inferred from the code, but still would feel nice(sizes and meanings would be obvious), maybe even extract it into a separate function as it's crucial part and pretty fragile, would be nice to give it special treatment

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    let pkg_len = pkgbuf.len();
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut udppkg = MutableUdpPacket::new(&mut pkgbuf[IPV4_HEADER_LEN..])
+        .expect("UDP buffer should not be too small");
+    udppkg.set_source(random_port(&mut rng));
```
LukasPukenis, line 122: `random_port()` is not used anywhere, could be inlined here

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+    pkgbuf.extend_from_slice(&wg_public.to_bytes());
+    pkgbuf.extend_from_slice(&(pq_public.as_bytes().len() as u32).to_le_bytes());
+    pkgbuf.extend_from_slice(pq_public.as_bytes());
+
+    let pkg_len = pkgbuf.len();
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut udppkg = MutableUdpPacket::new(&mut pkgbuf[IPV4_HEADER_LEN..])
+        .expect("UDP buffer should not be too small");
```
LukasPukenis, line 121: We should eliminate unwraps and expects. It probably means that something went very bad however we can handle it easily by wrappig it in an error and allowing graceful retry with non-pq for example

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,298 @@
...
+        noise::TunnResult::Err(err) => {
+            return Err(format!("Failed to encapsulate handshake message: {err:?}").into())
+        }
+        noise::TunnResult::WriteToNetwork(buf) => {
+            sock.send(buf).await?;
+        }
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    // The response should be 92, so the buffer is sufficient
```
LukasPukenis, line 248: What does `92` represent? Regarding the comment it looks like the response size, however it's not checked anywhere. This comment is a continuation of the previous one about the ascii diagram - it would be great to have the diagram of some sort in the codebase in the comment of how libtelio connets to the PQ server. With the responses it expects and the meanings of those :)

LukasPukenis (commented):
+0.75 I really like the implementation, it's easy to follow(code), left some comments mainly about documenting things for future readers.

Tests are missing, but it was agreed that it's because currently it's not available to implement those(this is work in progress by me).

src/device/mod.rs
```diff
@@ -1460,12 +1463,37 @@ impl Runtime {
...
-        } else if exit_node.endpoint.is_none() {
+        } else if let Some(addr) = exit_node.endpoint {
+            if let Some(pq_conf) = &self.features.post_quantum_vpn {
+                telio_log_debug!("Initializing PQ hanshake");
+
+                let fetch_keys = Box::pin(wg::pq::fetch_keys(
+                    addr,
+                    &self.requested_state.device_config.private_key,
+                    &exit_node.public_key,
+                )); // The future is large, let's move it onto the heap
```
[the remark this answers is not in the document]
matszczygiel: The dylint stage fails with an error that the future is large. This is a way to avoid the red CI

crates/telio-wg/src/pq.rs
matszczygiel: Yeah, it makes sense to inline it but it makes sense to keep it as a separate function as well since it's quite a self-contained unit. The same goes with the `handshake()` which is used only in one place and the proposed extraction of payload generation into a separate function.

crates/telio-wg/src/pq.rs
matszczygiel: This is statically guaranteed to never panic

crates/telio-wg/src/pq.rs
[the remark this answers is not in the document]
matszczygiel: The `set_*()` methods don't return `&mut Self`

crates/telio-wg/src/pq.rs
matszczygiel: Yeah, it's 92 bytes long message. The `handshake()` function performs a normal WG handshake, so nothing PQ-related here. However, I will provide the diagram for the whole key exchange mechanism

crates/telio-wg/src/pq.rs
matszczygiel: So IDK, if you have a strong opinion on this one let me inline it, but if not then let me keep it as is

src/device/mod.rs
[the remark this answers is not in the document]
tomaszklak: In the past we got some crashes due to stack exhaustion caused by a lot of state being stored on stack by future, which is why we now try to guard against it with that lint.

crates/telio-wg/src/pq.rs
LukasPukenis: Could you elaborate a bit more on this?

crates/telio-wg/src/pq.rs
LukasPukenis: no strong opinion on this one

src/device/mod.rs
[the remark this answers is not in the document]
LukasPukenis: That was unexpected really, thanks for explaining.

src/device/mod.rs
```diff
@@ -1460,12 +1463,37 @@ impl Runtime {
...
             if let Some(dns) = &self.entities.dns.lock().await.resolver {
                 self.reconfigure_dns_peer(dns, &dns.get_default_dns_servers())
                     .await?;
             }
-        } else if exit_node.endpoint.is_none() {
+        } else if let Some(addr) = exit_node.endpoint {
+            if let Some(pq_conf) = &self.features.post_quantum_vpn {
+                telio_log_debug!("Initializing PQ hanshake");
+
+                let fetch_keys = Box::pin(wg::pq::fetch_keys(
```
Jauler, line 1477: :thinking: 

Are there any retry mechanisms for PQ handshake?

crates/telio-wg/src/pq.rs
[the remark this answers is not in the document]
stalowyjez: Yeah, I just think that it would be nice to add some comment in the code as well, so it is clear why it is here and someone won't just remove it in the future.

crates/telio-wg/src/pq.rs
stalowyjez: For example, maybe it would be nice to extract the creation of the packet?

crates/telio-wg/src/pq.rs
matszczygiel: According to the `MutableUdpPacket::new` docs, it panics only when the buffer provided is smaller than the UDP header size. The buffer used has at least the required length (the `UDP_HEADER`) variable

src/device/mod.rs
matszczygiel: No :D, what do you think it should look like? a loop with the timeout?

crates/telio-wg/src/pq.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
matszczygiel: done

src/device/mod.rs
Jauler: Generally I see two viable options:
* Simpler one: a constant-period retry, this is how WG itself solves this problem. E.g. every 5s attempt to do the PQ-handshake. If the handshake progresses somewhere (e.g. first static WG is done, but PQ handshake does not respond, or if PQ handshake responds, but the WG one, where preshared key is already set - does not) and fails in some later step -> just restart it from the beginning
* Slightly fancier one: Do the same as above, but instead of constant-period retry, use exponential backoff strategy.

matszczygiel: Is it OK for `connect_exit_node()` to have a potentially infinite loop? The thing is that we need to set the PQ things before the wireguard interface state is consolidated

matszczygiel: So the solution to the problem is clear right now. However, it's going to be done in a follow-up PR based on LLT-4657

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,331 @@
...
+
+impl PqKeys {
+    /// Retrieve the PQ preshared key
+    pub fn pq_shared(&self) -> telio_crypto::PresharedKey {
+        #[allow(clippy::expect_used)]
+        telio_crypto::PresharedKey(
+            self.pq_shared
+                .as_bytes()
+                .try_into()
+                .expect("Preshared WG keys should have the same lenght as PQ preshared key"),
```
stalowyjez, line 77: Ehh, so we generally just don't use `expect` in Telio and this doesn't look like some very good place to make an exception, constructor should just return a result IMHO. Also, it looks that it is +/- implementation of `TryFrom` trait, maybe you could turn it into one?

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,331 @@
...
+        )
+    }
+
+    /// Retrieve the WG private key
+    pub fn wg_secret(&self) -> telio_crypto::SecretKey {
+        telio_crypto::SecretKey::new(self.wg_secret.to_bytes())
+    }
```
stalowyjez, line 84: Also looks like `From` trait IMHO.

crates/telio-wg/src/pq.rs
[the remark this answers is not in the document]
stalowyjez: So, what about this comment?

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,331 @@
...
+    // Generate keys
+    let wg_secret = x25519_dalek::StaticSecret::random_from_rng(&mut rng);
+    let (pq_public, pq_secret) = kyber768::keypair();
+    let wg_public = x25519_dalek::PublicKey::from(&wg_secret);
+
+    let mut pkgbuf = Vec::with_capacity(1024 * 4); // 4 KiB
+
+    pkgbuf.resize(IPV4_HEADER_LEN + UDP_HEADER_LEN, 0);
+    push_get_method_udp_payload(&mut pkgbuf, &wg_public, &pq_public);
+    fill_get_packet_headers(pkgbuf.as_mut_slice(), &mut rng);
```
stalowyjez, line 115: Much nicer! But isn't it a bit C-style? Maybe we could just return this `pkgbuf` from a function and avoid making it mutable?

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,331 @@
...
+                .into());
+            }
+
+            data.read_exact(&mut cipherbuf)?;
+
+            kyber768::Ciphertext::from_bytes(&cipherbuf)
+                .map_err(|err| format!("Invalid PQ ciphertext received: {err:?}"))?
+        }
+        _ => return Err("Unexpected WG tunnel output".into()),
+    };
```
stalowyjez, line 180: I would consider moving this to some helper function, like `handle_response` or sth.

stalowyjez (commented):
Overall it looks quite good, just a few comments, to maybe improve readability a bit

crates/telio-wg/src/pq.rs
matszczygiel: I'm afraid I sadly have to disagree, both the WG preshared key and Kyber shared secret are by definition 32 bytes long. The `except` is here only because `as_bytes()` return `&[u8]` instead of `&[u8;32]` which is it's internal representation.
I find creating yet another type of error type and handling logic for errors that will never happen more cumbersome and unnecessary than simply unwrapping

crates/telio-wg/src/pq.rs
matszczygiel: You can't implement `From` for foreign create types, only `Into`. Moreover `Into` doesn't say which key it is, only that it's secret. I wanted to explicit

crates/telio-wg/src/pq.rs
matszczygiel: Let, me try

crates/telio-wg/src/pq.rs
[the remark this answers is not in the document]
matszczygiel: Commented

crates/telio-wg/src/pq.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
matszczygiel: done

crates/telio-wg/src/pq.rs
stalowyjez: Is it used anywhere as `kyber768::SharedSecret` key? Maybe we could just keep it stored as `telio_crypto::PresharedKey`?

Jauler: I have to agree with @stalowyjez this does introduce a sneaky hidden dependency. Can we somehow ensure, that this dependency (e.g. 32 bytes of length) is checked during compile time, so in a situation where this hiddend depenedncy of 32 bytes length is not true anymore -> we would be notified by failing build, instead of runtime crash?

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+    pkgbuf.extend_from_slice(pq_public.as_bytes());
+}
+
+fn fill_get_packet_headers(pkgbuf: &mut [u8], rng: &mut impl rand::Rng) {
+    let pkg_len = pkgbuf.len();
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut udppkg = MutableUdpPacket::new(&mut pkgbuf[IPV4_HEADER_LEN..])
+        .expect("UDP buffer should not be too small");
```
Jauler, line 292: Please make this compile-time failure instead of runtime

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+        &udppkg.to_immutable(),
+        &LOCAL_IP,
+        &REMOTE_IP,
+    ));
+    drop(udppkg);
+
+    #[allow(clippy::expect_used)]
+    #[allow(index_access_check)]
+    let mut ippkg = MutableIpv4Packet::new(&mut pkgbuf[..IPV4_HEADER_LEN])
+        .expect("IPv4 buffer should not be too small");
```
Jauler, line 306: Please make this compile-time failure instead of runtime

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+        wg_secret,
+    })
+}
+
+async fn handshake(
+    endpoint: impl ToSocketAddrs,
+    secret: &telio_crypto::SecretKey,
+    peers_pubkey: &telio_crypto::PublicKey,
+) -> Result<(Box<noise::Tunn>, UdpSocket), Error> {
+    let sock = UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0)).await?;
```
Jauler, line 153: Please use `telio_sockets::new_external_udp()` for socket creation, as this will form infinite routing loop on one platform, and will work on others, depending on default routing configuration.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    // The response should be 92, so the buffer is sufficient
+    let read = sock.recv(&mut pkgbuf).await?;
+
+    let mut msgbuf = [0u8; 2048];
+
+    #[allow(index_access_check)]
+    match tunn.decapsulate(None, &pkgbuf[..read], &mut msgbuf) {
```
Jauler, line 182: Please do not just `#[allow(index_access_check)]` as if this got merged, it would introduce Denial-Of-Service vulnerability in networks with higher MTU than 2048.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+        _ => return Err("Unexpected WG tunnel output".into()),
+    }
+
+    // Receive response
+    let read = sock.recv(&mut recvbuf).await?;
+
+    let mut msgbuf = [0u8; 2048]; // 2 KiB buffer should shuffice
+
+    #[allow(index_access_check)]
+    let ciphertext = match tunn.decapsulate(None, &recvbuf[..read], &mut msgbuf) {
```
Jauler, line 131: Please do not just #[allow(index_access_check)]!

As if this would get merged, it would introduce Denial-Of-Service vulnerability in networks with higher MTU than 2048.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+    endpoint: impl ToSocketAddrs,
+    secret: &telio_crypto::SecretKey,
+    peers_pubkey: &telio_crypto::PublicKey,
+) -> Result<PqKeys, Error> {
+    let (tunn, sock) = handshake(endpoint, secret, peers_pubkey).await?;
+
+    let mut rng = rand::rngs::StdRng::from_entropy();
+
+    // Generate keys
+    let wg_secret = x25519_dalek::StaticSecret::random_from_rng(&mut rng);
```
Jauler, line 107: I think x25519_dalek does not clamp the keys, therefore please use telio-crypto key generation function ;), to ensure clamping.

crates/telio-wg/src/pq.rs
```diff
@@ -0,0 +1,351 @@
...
+            let ciphertext = kyber768::Ciphertext::from_bytes(&ct_bytes).unwrap();
+
+            let shared = kyber768::decapsulate(&ciphertext, &secret_key);
+
+            base64::encode(shared.as_bytes())
+        };
+
+        assert_eq!(shared, expected_shared);
+    }
+}
```
Jauler, line 351: Could you add some fuzzing for protocol section in a similar manner like, for example, telio-crypto has: https://github.com/NordSecurity/libtelio/tree/main/crates/telio-crypto/fuzz.

src/device/mod.rs
Jauler: Also keep in mind, that we will not be able to block external libtelio methods until handshake is done. We will need to do that in somewhat async manner, but spinning of separate module/task like we discussed is one way to go about it :).

src/device/mod.rs
```diff
@@ -1465,12 +1468,37 @@ impl Runtime {
...
 
+        self.requested_state.postquantum_wg = None;
+
         if is_meshnet_exit_node {
             if let Some(dns) = &self.entities.dns.lock().await.resolver {
                 self.reconfigure_dns_peer(dns, &dns.get_default_dns_servers())
                     .await?;
             }
-        } else if exit_node.endpoint.is_none() {
+        } else if let Some(addr) = exit_node.endpoint {
```
Jauler, line 1478: Also note, that we should handle meshnet<->PQ-WG incompatabilities, by either erroring out in an incompatible situation, or reconnecting. It would be nice to sync sync with Linux-app devs on exact solution.

Jauler (changes requested):
Hi, I think we do need to adjust a few things before merging :).

crates/telio-wg/src/pq.rs
matszczygiel: I believe, `UdpSocket::recv()` would not report more bytes read than the `recvbuf.len()`, thus it will not panic. I've seen this pattern multiple times in libtelio codebase btw

crates/telio-wg/src/pq.rs
matszczygiel: The same comment applies https://github.com/NordSecurity/libtelio/pull/247#discussion_r1431019907

crates/telio-wg/src/pq.rs
matszczygiel: How? https://docs.rs/pnet_packet/latest/pnet_packet/udp/struct.MutableUdpPacket.html does not have API with compile time checks

crates/telio-wg/src/pq.rs
matszczygiel: Same comment applies https://github.com/NordSecurity/libtelio/pull/247#discussion_r1431022384

src/device/mod.rs
matszczygiel: Yeah, since PQ is a feature, I decided to throw an error from `set_config()` when PQ is enabled https://github.com/NordSecurity/libtelio/pull/259/commits/11dcb11051004f339f7947e19b965aa64f2e74c5
I've planned to merge it with the follow-up PR, I can port this commit to this one

crates/telio-wg/src/pq.rs
matszczygiel: Let me try, I guess I will create a custom key type for this one

crates/telio-wg/src/pq.rs
Jauler: yeah, seems like I am mistaken, you are right. The `read()` will not read more than buffer

crates/telio-wg/src/pq.rs
Jauler: Arrrgh, seems like I was distracted yesterday when looking through the code. I was thinking about the `pkgbuf` access, and thought it may be dangerous, but the `expect()` is called upon UDP packet, not the other places

In any case, this would crash the whole tunnel if we pass invalid/unparsable UDP packet, therefore these `expect()` must be handled in a proper way (e.g. by returning an error or similar).

src/device/mod.rs
Jauler: If it is planned in follow up PR -> it is ok, we can add it there.

crates/telio-wg/src/pq.rs
matszczygiel: I did move this conversion into the `handshake()` function, returning io error instead of `expect()`

crates/telio-wg/src/pq.rs
matszczygiel: Done

crates/telio-wg/src/pq.rs
matszczygiel: Done

src/device/mod.rs
matszczygiel: I've added it here

crates/telio-wg/src/pq.rs
matszczygiel: Done :D

src/device/wg_controller.rs
```diff
@@ -382,10 +387,16 @@ async fn build_requested_peers_list<
...
+        let preshared_key = requested_state
+            .postquantum_wg
+            .as_ref()
+            .map(|pq| pq.pq_shared);
+
         if let Some(meshnet_peer) = requested_peers.get_mut(&exit_node.public_key) {
             // Exit node is meshnet peer, so just promote already existing node to be exit node
             // with allowed ips change
             meshnet_peer.peer.allowed_ips = allowed_ips;
+            meshnet_peer.peer.preshared_key = preshared_key;
```
Jauler, line 399: This does not really make sense for meshnet peer, as PQ and meshnet is not compatible. But it will not hurt also, lets just remember to remove it in the rekey PR ;).

matszczygiel: Yeah, let me do it in the followup

Jauler (approved):
+1.0

src/device/wg_controller.rs
matszczygiel: Already done: https://github.com/NordSecurity/libtelio/pull/259/commits/cb5a90c22e310d821be689face0cc61dc189a808

PR #249, opened 2023-12-11
Fix crash in wireguard-go - port to release/v4.1

jjanowsk:
Wireguard-go could crash if winrio initization fails. That is because in that case StdNetBind is returned from NewDefaultBind function instead of WinRingBind. Later this object was casted to the WinRingBind in order to call OpenOnLocalhost method. This commit adds new method OpenOnLocalhost to our interface FullBind and implements is for both WinRingBind and StdNetBind. It requires us to copy StdNetBind implementation for the upstream.

This is port of: https://github.com/NordSecurity/libtelio/pull/246

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+### v4.1.4
+### **Bruschetta**
+---
+### Changelog
+* LLT-4586: Fix wireguard-go initialization crash
+
+<br>
+
 ### v4.1.3
 ### **Bruschetta**
 ---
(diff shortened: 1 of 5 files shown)


Jauler (approved):
+2.0

changelog.md
```diff
@@ -1,3 +1,11 @@
+### v4.1.4
```
Jauler, line 1: Maybe it is worth updating version in `cargo.toml` too, so it will not require separate push

PR #250, opened 2023-12-11
Add wildcarded subdomains to peers dns records

lcruz99:
### Problem
Currently it is not possible to use <something>.<nickname/hostname>.nord as a domain name. It is quite useful when single server in meshnet hosts multiple services.

### Solution
Add wildcarded entries to the dns records list.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -1119,13 +1119,19 @@ impl Runtime {
 
     async fn upsert_dns_peers(&self) -> Result {
         if let Some(dns) = &self.entities.dns.lock().await.resolver {
-            // hostnames have priority over nicknames (override if there's any conflict)
             let mut peers: Records = HashMap::new();
+            // Hostnames have priority over nicknames (override if there's any conflict)
             if self.features.nicknames {
                 peers = self.requested_state.collect_dns_nickname_records();
             }
-
             peers.extend(self.requested_state.collect_dns_records());
+
+            // Insert wildcard for subdomains
+            let wildcarded_peers: Records = peers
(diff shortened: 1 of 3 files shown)


Jauler (approved):
+1.5

src/device/mod.rs
```diff
@@ -1124,9 +1124,16 @@ impl Runtime {
...
             }
-
             peers.extend(self.requested_state.collect_dns_records());
 
+            // Insert wildcard for subdomains
+            let wildcarded_peers: Records = peers
+                .clone()
+                .into_iter()
+                .map(|(name, ip)| (format!("*.{}", name), ip))
+                .collect();
```
mathiaspeters, line 1134: It might not be that performance critical, but I think we could reduce the number of allocations slightly this way:

```suggestion
            let wildcarded_peers: Records = peers
                .iter()
                .map(|(name, ip)| (format!("*.{}", name), ip.clone()))
                .collect();
```

mathiaspeters: And if I'm not mistaken, `HashMap::extend` takes an iterator, so the `collect` might be unnecessary here

baltrus1 (dismissed):
0.5

src/device/mod.rs
lcruz99: Nice, thanks for pointing that out. It still needs to collect() though, otherwise compiler refuses to extend `peers` as the iterator borrowed peers already.

mathiaspeters (approved):

PR #251, opened 2023-12-12
Add no-link detection mechanism

dfetti:
Add a hinting mechanism for no-link detection. More info about this can be found in RFC-LLT-0045.

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-model/src/api_config.rs
```diff
@@ -268,6 +268,13 @@ where
     Ok(Some(eps))
 }
 
+/// Turns on the no link detection mechanism
+#[derive(Default, Copy, Clone, Debug, PartialEq, Eq, Deserialize)]
+pub struct FeatureNoLinkDetection {
+    /// Configurable rtt in seconds
+    pub rtt_seconds: Option<u64>,
```
matszczygiel, line 275: So if the `FeatureNoLinkDetection` is `None` in the main config then it's disabled. What does it mean when the feature is `Some()` but the `rtt_seconds` is `None`? Does it make sense for it to be `None`?
If you want to provide default value, use `#[serde(default = "function")`

crates/telio-wg/src/wg.rs
```diff
@@ -544,12 +630,39 @@ impl State {
             } else {
                 return false;
             };
-            if !try_send(PeerState::Connecting, peer.clone()).await {
+
+            // Node is new and default LinkState is down. Save it before sending the event
+            self.no_iink_detection_info
+                .insert(*key, NoLinkDetectionInfo::default());
+            if !self
```
matszczygiel, line 637: These `if`s are a bit strange, can we return an error with `?` instead?

matszczygiel (changes requested):
Some minor nits need to be settled but overall great effort

crates/telio-wg/src/wg.rs
```diff
@@ -747,9 +873,72 @@ impl State {
...
+
+                let (rx_bytes, rx_ts) = if new_rx > old_info.last_rx_bytes {
+                    (new_rx, now)
+                } else {
+                    (old_info.last_rx_bytes, old_info.last_rx_timestamp)
+                };
+
+                let (tx_bytes, tx_ts) = if new_tx > old_info.last_tx_bytes {
+                    // We should not bother about passive keepalives, they will not modify
+                    // the tx_bytes count received from wg
```
LukasPukenis, line 912: Hmmm.. Is this actually true? Sending the keepalive(passive or persistent) does not increase `tx_bytes`? The payload is null however the headers still take up space. Without detecting the keepalive which is outgoing we will falsely assume that we need to wait for a response where we don't actually

LukasPukenis (commented):
The state machine is simpler and much easier to follow than previously.

I have however questions about keepalives, left those in the comments.

crates/telio-wg/src/wg.rs
```diff
@@ -102,21 +108,76 @@ pub struct Io {
...
 
     // Detecting unexpected driver failures, such as a malicious removal
     // We won't be notified of any errors, but a periodic call to get_config_uapi() will return a Win32 error code != 0.
     uapi_failed_last_call: bool,
     uapi_fail_counter: i32,
+
+    // No link detection mechanism config
+    no_link_detection: NoLinkDetection,
+    // Information used for no-link detection mechanism
+    no_iink_detection_info: HashMap<PublicKey, NoLinkDetectionInfo>,
```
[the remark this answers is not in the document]
dfetti: I wouldn't have noticed it :)) Good eyes

crates/telio-wg/src/wg.rs
```diff
@@ -559,9 +672,49 @@ impl State {
...
+                    }
+                }
+
                 #[allow(clippy::collapsible_if)]
-                if !old.is_same_event(new) || old_state != new_state {
-                    if !try_send(new_state, new.clone()).await {
+                if !old.is_same_event(new)
+                    || old_state != new_state
+                    || should_notify_link_state_change
+                {
```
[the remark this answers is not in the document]
dfetti: I think it makes sense to be like this, because is_same_event checks if the endpoint or allowed_ips have changed. And we want to notify this changes too.

crates/telio-model/src/api_config.rs
dfetti: There are 3 cases:
- no_link_detection is None which means feature is Disabled
- no_link_detection is Some with rtt None which means feature is Enabled with default rtt value
- no_link_detection is Some with rtt Some which means feature is Enabled with custom rtt value

I did not used `#[serde(default = "function")]` because I placed this logic in the From trait implementation of NoLinkDetection config struct from telio-wg. If I had to do that I found useless defaulting rtt here.

```
/// No link detection mechanism config
#[derive(Clone)]
pub enum NoLinkDetection {
    /// No link detection mechanism is disabled
    Disabled,
    /// No link detection mechanism is enabled
    /// Which means that events with link_state will be emitted
    Enabled {
        /// Configurable rtt
        rtt: Duration,
    },
}

impl From<Option<FeatureNoLinkDetection>> for NoLinkDetection {
    fn from(value: Option<FeatureNoLinkDetection>) -> Self {
        match value {
            None => NoLinkDetection::Disabled,
            Some(FeatureNoLinkDetection { rtt_seconds: None }) => NoLinkDetection::Enabled {
                rtt: DEFAULT_CONFIGURABLE_RTT,
            },
            Some(FeatureNoLinkDetection {
                rtt_seconds: Some(rtt),
            }) => NoLinkDetection::Enabled {
                rtt: Duration::from_secs(rtt),
            },
        }
    }
}
```
What do you think?

crates/telio-wg/src/wg.rs
```diff
@@ -360,7 +428,7 @@ impl WireGuard for DynamicWg {
         Ok(task_exec!(&self.task, async move |s| {
             let mut to = s.interface.clone();
             to.peers.remove(&key);
-            s.update(&to, true).await;
+            let _ = s.update(&to, true).await;
```
LukasPukenis, line 431: Just a tip for the future: unrelated changes could go into a separate commit, no need to squash everything as it inflates the commit's value :)

crates/telio-wg/src/wg.rs
```diff
@@ -501,55 +569,92 @@ impl State {
...
+        peer: Peer,
+    ) -> Result<(), Error> {
+        // If the feature is disabled, the link_state field must be None
+        let link_state = if matches!(self.no_link_detection, NoLinkDetection::Disabled) {
+            None
+        } else if state != PeerState::Connected {
+            // If the peer is not connected, the link state must always be down
+            Some(LinkState::Down)
+        } else {
+            link_state
```
LukasPukenis, line 590: Hmmm, curious if we need to pass `link_state` at all, judging from the glance at the code we can deduce it from `state` alone(conneted = link_state::up, else down)

crates/telio-model/src/api_config.rs
matszczygiel: I prefer the default value in the config itself, but I'm not pressing this one. If you think it is better as it is, then let's leave it

crates/telio-wg/src/wg.rs
```diff
@@ -748,7 +867,53 @@ impl State {
...
+        let now = Instant::now();
+
+        let (rx_bytes, rx_ts) = if new_rx > old_info.last_rx_bytes {
+            (new_rx, now)
+        } else {
+            (old_info.last_rx_bytes, old_info.last_rx_timestamp)
+        };
+
+        let (tx_bytes, tx_ts) = if new_tx > old_info.last_tx_bytes {
+            // Ignore the keepalive messages
```
LukasPukenis, line 893: It could be stated that this only works in a limited fashion: because we poll every second, it means the detection might not work as intended if there are more packets within that time window as the difference is the sum of packets.

On the sender side persistent-keepalives might interfere as those are emitted regardless the activity, meaning for example if persistent-keepalive is set at 5seconds then it might interfere with the passive-keepalive which is a constant of 10seconds.

IMHO **This actually could be a testcase: set persistent-keepalive to 5seconds and observe that it reports link as up when receiver is offline. RTT might be 0 to speed it up**. What are your thoughts on the subject? :)

crates/telio-wg/src/wg.rs
```diff
@@ -748,7 +867,53 @@ impl State {
...
+    fn compute_new_node_info(
+        &self,
+        old_info: &NoLinkDetectionInfo,
+        new_tx: u64,
+        new_rx: u64,
+    ) -> NoLinkDetectionInfo {
+        let configurable_rtt = match self.no_link_detection {
+            NoLinkDetection::Disabled => DEFAULT_CONFIGURABLE_RTT,
+            NoLinkDetection::Enabled { rtt } => rtt,
+        };
```
matszczygiel, line 882: IMHO this doesn't make sense when the `NoLinkDetection` is disabled, maybe `rtt` should be passed as a parameter (then self is unused) and the `match thing should be higher in the call stack

nat-lab/tests/telio.py
```diff
@@ -27,6 +27,18 @@ class State(Enum):
...
+class LinkState(Enum):
+    Down = "down"
+    Up = "up"
+
+
+@dataclass
+class LinkStateEventsCount:
+    none: int = 0
+    up: int = 0
+    down: int = 0
```
LukasPukenis, line 39: It's not bad as it is but we have deterministic results(given perfect network conditions) so I would prefer having an explicit ordering of events as an improvement(up, down, then up again) for example. Without order it might hide subtle bugs.

matszczygiel (dismissed):

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,489 @@
+import asyncio
```
LukasPukenis, line 1: Nice moving these tests into a separate file:)

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,489 @@
...
+        )
+        alpha, beta = env.nodes
+        client_alpha, client_beta = env.clients
+
+        # Expect no link event while peers are idle
+        await asyncio.sleep(20)
+        alpha_counts = client_beta.get_link_state_events_count(alpha.public_key)
+        beta_counts = client_alpha.get_link_state_events_count(beta.public_key)
+
+        # 1 down when node is Connecting and 1 up when node is Connected
```
LukasPukenis, line 80: piggy backing on the previous comment about the order - due to a bug it might be  that events come in as `up` and then `down` yet this test would pass.

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,489 @@
...
+        env = await setup_mesh_nodes(
+            exit_stack, [alpha_setup_params, beta_setup_params]
+        )
+        alpha, beta = env.nodes
+        client_alpha, client_beta = env.clients
+        connection_alpha, connection_beta = [
+            conn.connection for conn in env.connections
+        ]
+
+        for _ in range(0, 5):
```
LukasPukenis, line 137: Correct me if I am wrong but the test executes 5 pings every second? This isn't a very long time(the testname says it is). I would have them chatting for at least a minute so to be sure that it's at least longer than the passive-keepalive(10s)+rtt

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,489 @@
...
+        connection_alpha, connection_beta = [
+            conn.connection for conn in env.connections
+        ]
+
+        async with Ping(connection_alpha, beta.ip_addresses[0]).run() as ping:
+            await testing.wait_long(ping.wait_for_next_ping())
+        async with Ping(connection_beta, alpha.ip_addresses[0]).run() as ping:
+            await testing.wait_long(ping.wait_for_next_ping())
+
+        await client_beta.stop_device()
```
LukasPukenis, line 342: Are we sure in this case that the ping was successful before we stop the device? I mean `await` might resolve if the ping was just sent, in that case we might stop the device and enter a race condition where:
- peer's response might have reached us
- peer's device was shut down and the response was not even generated

Overall I wonder why in this testcase the alpha doesn't receive `down` event twice as well? A few lines below it issues a ping and then sleeps, this should guarantee it :thinking:

LukasPukenis (commented):
Great effort, I really like the testcases, those really show exactly what we expect and when, thus making future debugging much easier.

I have left several questions and ideas on the testcases and have these additional testcase ideas:
- Bring up only 1 peer, check that it really never connects to another one.
- I would prefer to have at least some testcases with persistent keepalive enabled at short interval just to see that it doesn't alter the testcases. In theory, setting it should not alter the results at all, doesn't matter if it's alpha, beta or both with short persistent keepalive. This is more for a piece of mind. What is your take on this one?
- Would be fun to have a testcase with persistent keepalive that's a divisor of passive-keepalive constant. The setup would be that _beta_ would stream data to _alpha_ and _alpha_ would not respond, meaning it's WireGuard would generate passive-keepalives every 10 seconds. If _alpha_ would also have persistent-keepalives at say 10 or 5 seconds, it would then interfere with the keepalive check on `tx_ts` update. However I am not sure if this scenario actually tests anything as it requires receiving data(for passive keepalive to happen) which means that the link is up.

@Jauler please take a look

matszczygiel:
+1

nat-lab/tests/test_events_link_state.py
dfetti: The `wait_for_next_ping` actually waits for the response so it should be all ok.

crates/telio-wg/src/wg.rs
dfetti: Consider if the peer state is connected and hinting mechanism detected possible link down. We want to send link down event.

matszczygiel (approved):

matislovas:
+1

PR #254, opened 2023-12-12
Llt 4225 migrate telio from log crate to tracing crate 

ThrasherLT:
### Problem
In order to help with finding the root cause of various bugs, it was suggested to start using the tracing crate for it's various features like spans

### Solution
Migrate Telio to use the tracing crate instead of the log crate

- [x] README.md is updated
- [x] changelog.md is updated


src/ffi/mod.rs
```diff
@@ -867,29 +863,90 @@ fn filter_log_message(msg: String) -> Option<String> {
...
+
+    fn new_span(&self, _span: &tracing::span::Attributes<'_>) -> tracing::span::Id {
+        // TODO using a placeholder for now
+        tracing::span::Id::from_u64(1337)
     }
 
-    fn log(&self, record: &Record) {
-        if !self.enabled(record.metadata()) {
+    fn record(&self, _span: &tracing::span::Id, _values: &tracing::span::Record<'_>) {
+        // TODO
```
LukasPukenis, line 915: what should these functions do?

LukasPukenis (commented):
+1.0 more or less how I would expect it - log functions changed to tracing functions, however I think the main benefit of tracing is that it supports spans/nesting but we're using telio_log_* macros heavily, are there plans to migrate those to plain tracing with spans?

src/ffi/mod.rs
ThrasherLT: These will be implemented when spans are implemented, which is scheduled for a separate task/PR to make the migration a bit simpler.

ThrasherLT:
> +1.0 more or less how I would expect it - log functions changed to tracing functions, however I think the main benefit of tracing is that it supports spans/nesting but we're using telio_log_* macros heavily, are there plans to migrate those to plain tracing with spans?

Yes, spans are planned to be implemented separately to make the migration a bit simpler

src/ffi/mod.rs
LukasPukenis: Fair enough, Thanks for explaining.

stalowyjez (commented):
Looks good, +1

src/ffi/mod.rs
```diff
@@ -162,19 +163,14 @@ fn telio_new_common(
...
-    } else if let Err(err) =
-        tracing::subscriber::set_global_default(telio_utils::tracing::TracingToLogConverter)
-    {
-        error!("{}", err);
+    let tracing_subscriber = TelioTracingSubscriber {
+        callback: logger,
+        max_level: log_level.into(),
+    };
+    if tracing::subscriber::set_global_default(tracing_subscriber).is_err() {
+        telio_log_warn!("Could not set logger, because logger had already been set by previous libtelio instance");
```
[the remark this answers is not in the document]
ThrasherLT: It's the only documented error on that function. Also we need to warn the app about it, because this case is very likely due to the fact that there's no way to unset the subscriber. Some apps destroy the entire telio instance and create another one, but the subscriber will stay the same.

stalowyjez (approved):

PR #259, opened 2023-12-14
LLT-4620: implement wg post quantum rekey mechanism in libtelio codebase

matszczygiel:
### Problem
We want to make the VPN tunnel post-quantum safe. Followup of PR #247

### Solution
The preshared key obtained during the PQ handshake must be periodically changed. This PR implements a periodic rekey procedure.

There isn't any unit or integration test because that's a separate task and @LukasPukenis is responsible for this one.

- [x] README.md is updated
- [x] changelog.md is updated


matszczygiel:
Blocked by PR #247

crates/telio-wg/src/pq.rs
```diff
@@ -16,19 +16,21 @@ use pnet_packet::{
...
 const LOCAL_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 2);
 const REMOTE_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 1);
-const HANDSHAKE_VERSION: u32 = 1;
+const PQ_PROTO_VERSION: u32 = 1;
 const CIPHERTEXT_LEN: u32 = kyber768::ciphertext_bytes() as _;
 
 const IPV4_HEADER_LEN: usize = 20;
 const UDP_HEADER_LEN: usize = 8;
 
+type RekeyStatus = u8;
```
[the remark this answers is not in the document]
matszczygiel: Yeah, it's an internal (private) helper. And to be honest I don't know how many different values it takes, because it's the u8 field in the response packet. Will leave it as is

src/device/mod.rs
```diff
@@ -230,6 +231,8 @@ pub struct Entities {
 
     // Nurse
     nurse: Option<Arc<Nurse>>,
+
+    postquantum_wg: Option<pq::Entity>,
```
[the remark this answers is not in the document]
matszczygiel: I think the name is self-explanatory, look at the `nurse` comment above - I don't want to do the same just for the sake of having a comment

src/device/mod.rs
```diff
@@ -1475,36 +1482,25 @@ impl Runtime {
...
-                    fetch_keys,
-                )
-                .await
-                .map_err(|_| {
-                    telio_log_warn!("PQ hanshake timeout");
-                    wg::pq::Error::Generic("Fetching PQ keys timeout".into())
-                })??;
+                    self.requested_state.device_config.private_key,
+                    exit_node.public_key,
+                    pq_conf,
```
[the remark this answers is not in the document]
matszczygiel: Um, why? It's closely tied to a single VPN connection. The task (it's not a thread) does the PQ handshake and then maintains key rotation with the VPN server

src/device/pq.rs
```diff
@@ -0,0 +1,94 @@
+use std::{net::SocketAddr, sync::Arc, time::Duration};
```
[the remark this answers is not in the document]
matszczygiel: IDK, I've placed it here because it's the device that maintains the connection logic. Maybe doing a separate crate for PQ? I think I will wait for more input from the reviewer because it's quite subjective IMHO

src/device/wg_controller.rs
```diff
@@ -387,22 +393,22 @@ async fn build_requested_peers_list<
...
             // Exit node is a fresh node, therefore - insert create new peer
             let public_key = exit_node.public_key;
             let endpoint = exit_node.endpoint;
             let persistent_keepalive_interval = requested_state.keepalive_periods.vpn;
+            // If the PQ VPN is set up we need to configure the preshared key
+            let preshared_key = requested_state
+                .postquantum_keys
+                .as_ref()
+                .map(|pq| pq.pq_shared);
+
```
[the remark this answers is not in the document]
matszczygiel: Yeah, sadly it's incompatible with meshnet

changelog.md
```diff
@@ -13,7 +13,7 @@
 * LLT-4225: Migrate Telio to use the tracing crate instead of log crate
 * LLT-4594: Fix session keeper's bug, to enable ICMP6 packets
 * LLT-4597: Fix QoS' IPv6 pinging
-* LLT-4614: Implement the post quantum VPN handshake
+* LLT-4614: Implement the post quantum VPN handshake and rekey mechanism
```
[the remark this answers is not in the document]
matszczygiel: Done

crates/telio-wg/src/pq.rs
```diff
@@ -279,14 +346,35 @@ fn push_get_method_udp_payload(
...
     pkgbuf.extend_from_slice(&(wg_public.len() as u32).to_le_bytes());
     pkgbuf.extend_from_slice(wg_public);
     pkgbuf.extend_from_slice(&(pq_public.as_bytes().len() as u32).to_le_bytes());
     pkgbuf.extend_from_slice(pq_public.as_bytes());
 }
 
+/// The REKEY payload looks as follows:
+///
+/// ---------------------------------
+///  version           , u32le , = 1
```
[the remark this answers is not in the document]
matszczygiel: Fixed

src/device/mod.rs
```diff
@@ -278,6 +281,7 @@ pub struct EventListeners {
     derp_event_subscriber: mc_chan::Rx<Box<DerpServer>>,
     endpoint_upgrade_event_subscriber: chan::Rx<UpgradeRequestChangeEvent>,
     stun_server_subscriber: chan::Rx<Option<StunServer>>,
+    post_quantum: Chan<pq::Event>,
```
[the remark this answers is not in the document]
matszczygiel: Done

src/device/mod.rs
```diff
@@ -1786,6 +1787,34 @@ impl TaskRuntime for Runtime {
...
+                        if let Some(keys) = &mut self.requested_state.postquantum_keys {
+                            // Check if we are still talking to the same exit node
+                            if keys.wg_secret == wg_secret {
+                                // and only then update the preshared key,
+                                // otherwise we're connecting to different node already
+                                keys.pq_shared = pq_shared;
+                            }
+                        }
+                    },
+                }
```
LukasPukenis, line 1807: This is a solid approach but I think adding a logging statement of low severity might aid in debuggin if the keys don't match.

src/device/wg_controller.rs
```diff
@@ -54,6 +54,12 @@ pub async fn consolidate_wg_state(
     entities: &Entities,
     features: &Features,
 ) -> Result {
+    if features.post_quantum_vpn.is_some() && requested_state.postquantum_keys.is_none() {
+        // The post quantum state is not ready, we don't want to set up quantum
+        // unsafe tunnel
+        return Ok(());
+    }
+
```
LukasPukenis, line 62: Please clarify my understanding on this one - `requested_state` is the request we send to WG driver and it will try it's best to match it?

In this case if there's a request for VPN connection but PQ keys are not present it will result in this call being prematurely exited and thus WG will not be set up for the spcified VPN server? TLDR: having no PQ keys doesn't allow for VPN connection?

matszczygiel: The requested_state structure is specific to telio device and holds the full information of current configuration. It's a big mutable structure and when something changes then a certain part of it is modified so that other parts of the system are aware of the change by consulting the structure.

So yes, when the PQ keys are not set up then the wg interface is not configured. However that's expected, because configuring the interface means setting up a quantum unsafe tunnel and we don't want that. Keep in mind that this call is actively executed when the keys are available.

LukasPukenis: Does it still allow for non-pq vpn connections?

src/device/mod.rs
matszczygiel: Done

src/device/wg_controller.rs
matszczygiel: No. And it shouldn't.

LukasPukenis: Oh my bad. After looking again and again I saw that it's
```
if features.post_quantum_vpn.is_some
```
and not
```
if features.vpn.is_some
```

:D

src/device/pq.rs
```diff
@@ -0,0 +1,94 @@
...
+                    Ok(key) => {
+                        keys.pq_shared = key;
+
+                        // The channel is allways open during the library operation.
+                        // It can only be closed on library shutdown, in that case we
+                        // may not care since the task itself will be killed soon
+                        let _ = chan.send(Event::Rekey(keys)).await;
+                    }
+                    Err(err) => telio_log_warn!("Failed to perform PQ rekey: {err}"),
+                }
```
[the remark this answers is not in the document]
matszczygiel: Nah, the task is aborted and a new one is started on the VPN switch since we need to establish the handshake first

src/device/mod.rs
```diff
@@ -1786,6 +1787,36 @@ impl TaskRuntime for Runtime {
...
+                        self.requested_state.postquantum_keys = Some(keys);
+                    },
+                    pq::Event::Rekey(pq::Keys { wg_secret, pq_shared }) => {
+                        if let Some(keys) = &mut self.requested_state.postquantum_keys {
+                            // Check if we are still talking to the same exit node
+                            if keys.wg_secret == wg_secret {
+                                // and only then update the preshared key,
+                                // otherwise we're connecting to different node already
+                                keys.pq_shared = pq_shared;
+                            }else {
```
dfetti, line 1804: nit: Format here

src/device/pq.rs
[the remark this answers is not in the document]
dfetti: I would also suggest to move this into a separate crate with all the PQ logic.

dfetti (commented):
+1.0

src/device/mod.rs
matszczygiel: Done

src/device/pq.rs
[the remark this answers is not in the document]
matszczygiel: Ok, let me do it then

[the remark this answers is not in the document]
matszczygiel: Done

LukasPukenis (commented):
+1

src/device/wg_controller.rs
```diff
@@ -54,6 +54,12 @@ pub async fn consolidate_wg_state(
     entities: &Entities,
     features: &Features,
 ) -> Result {
+    if features.post_quantum_vpn.is_some() && requested_state.postquantum_keys.is_none() {
+        // The post quantum state is not ready, we don't want to set up quantum
+        // unsafe tunnel
+        return Ok(());
```
Jauler, line 60: If I understand the idea correclty, IMHO, in such a case we should still consolidate, but against empty list of WG peers.

This `if` now basically means, that
```
if PQ is enabled, but keys are not yet established, then please leave the WG configuration unchanged
```

And I believe we would like to clear any WG peers in such a case? Or am I mistaken?

src/device/mod.rs
```diff
@@ -201,7 +201,7 @@ pub struct RequestedState {
     // Requested keepalive periods
     pub(crate) keepalive_periods: FeaturePersistentKeepalive,
 
-    pub postquantum_wg: Option<wg::pq::PqKeys>,
+    pub postquantum_keys: Option<telio_pq::Keys>,
```
Jauler, line 204: While not a strict rule or anything, the `requested_state` struct generally represents configuration requests coming from outside of libtelio. And internally derived/calculated/communicated things are stored directly in the entities themselves, which may be made public by the choice of said component.

We do have exception, like `wg_stun_server`, but it is generally too small to contain its own active component, and thus got here, even though does not really fit the purpose of the struct.

But in `telio-pq` case, it is already active component, therefore maybe some getters would make it fit better?

One example of such getter in a rather small active component is [upgrade_sync](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/upgrade_sync.rs), and the getter is defined [here](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/upgrade_sync.rs#L95), and example call site is [here](https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L553)

@matszczygiel what do you think about moving keys to the telio-pq crate itself instead of adding them into requested state?

src/device/mod.rs
```diff
@@ -281,6 +283,7 @@ pub struct EventListeners {
     derp_event_subscriber: mc_chan::Rx<Box<DerpServer>>,
     endpoint_upgrade_event_subscriber: chan::Rx<UpgradeRequestChangeEvent>,
     stun_server_subscriber: chan::Rx<Option<StunServer>>,
+    post_quantum: chan::Rx<telio_pq::Event>,
```
Jauler, line 286:
```suggestion
    post_quantum_subscriber: chan::Rx<telio_pq::Event>,
```

Maybe lets name it a subscriber, so it more clearly states that this is not the post_quantum entity itself?

src/device/mod.rs
```diff
@@ -291,6 +294,9 @@ pub struct EventPublishers {
 
     /// Used to manually trigger a nurse event collection
     nurse_collection_trigger_publisher: Option<Tx<Box<()>>>,
+
+    /// Passed down the PQ key rotation task
+    post_quantum: chan::Tx<telio_pq::Event>,
```
Jauler, line 299: Nit, you can ignore this if you think it does not make sense:
```suggestion
    post_quantum_publisher: chan::Tx<telio_pq::Event>,
```

Maybe lets name it a publisher, so it more clearly states that this is not the post_quantum entity itself?

src/device/mod.rs
```diff
@@ -1484,36 +1494,25 @@ impl Runtime {
...
-                    wg::pq::Error::Generic("Fetching PQ keys timeout".into())
-                })??;
+                    self.requested_state.device_config.private_key,
+                    exit_node.public_key,
+                    pq_conf,
+                );
 
-                self.requested_state.postquantum_wg = Some(keys);
-                telio_log_debug!("PQ hanshake finished succesfully");
+                self.entities.postquantum_wg = Some(entt);
```
Jauler, line 1515: What would happen if `telio_connect_to_exit_node()` is called while `self.entities.postquantum_wg` is already `Some(...)`.

Dont we have to kill/stop etc. tokio task spawned on "previous" entity?

src/device/mod.rs
matszczygiel: I agree about moving keys from `RequestedState` however I don't like moving it to telio-pq because the key handling logic is (and must be) implemented in `device/mod.rs`. Thus even if we move keys storage there, they still need to be set up from the device making a weird dependency loop.
I think I will try to move them into the entity stored in the device but it would be a different entity than now, something like
```rust
struct PqEntity {
    keys: telio_pq::Keys,
    conn: telio_pq::Entity, // I'll probably change the type name in this case
}
```

src/device/mod.rs
matszczygiel: It will be killed in the `impl Drop for Entity`, so this case if handled

src/device/wg_controller.rs
matszczygiel: I fixed this one by not adding a peer to the list when it's set as an exit node but keys are not available yet. The rest of the consolidation is proceeding as usual

src/device/mod.rs
matszczygiel: Done as described, but moved the entity nonetheless into `telio-pq` crate

src/device/mod.rs
matszczygiel: Done

src/device/mod.rs
matszczygiel: Done

matszczygiel:
I've cleaned the commits a bit after addressing the review comments

tomaszklak (approved):
+1.0

PR #264, opened 2023-12-15
Llt 4490 start endpoint providers after meshnet

stalowyjez:
### Problem
Endpoint providers, CPC, SessionKeeper, upgrade_sync proxy and relay should be started only after the meshnet is started and configured. Also they should be stopped when the meshnet config is set to `None`.

### Solution
Move Meshnet entities start to `set_config` method.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/heartbeat.rs
+++ b/crates/telio-nurse/src/heartbeat.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use bitflags::bitflags;
+use futures::FutureExt;
 use nat_detect::NatType;
 use std::collections::BTreeSet;
 use std::fmt::Write;
@@ -19,6 +20,7 @@ use telio_model::{
 };
 use telio_nat_detect::nat_detection::{retrieve_single_nat, NatData};
 use telio_proto::{HeartbeatMessage, HeartbeatNatType, HeartbeatStatus, HeartbeatType};
+use telio_relay::multiplexer::Multiplexer;
 use telio_relay::DerpRelay;
 use telio_task::{
     io::{chan, mc_chan, Chan},
@@ -85,9 +87,9 @@ impl Default for MeshLink {
 /// Input/output channel for Heartbeat
 pub struct Io {
     /// Channel sending and receiving heartbeat data
-    pub chan: Chan<(PublicKey, HeartbeatMessage)>,
+    pub chan: Option<Chan<(PublicKey, HeartbeatMessage)>>,
     /// Event channel to gather derp events
-    pub derp_event_channel: mc_chan::Rx<Box<Server>>,
+    pub derp_event_channel: Option<mc_chan::Rx<Box<Server>>>,
     /// Event channel to gather wg events
     pub wg_event_channel: mc_chan::Rx<Box<Event>>,
     /// Event channel to gather local nodes from mesh config updates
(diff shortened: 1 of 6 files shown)


crates/telio-nurse/src/heartbeat.rs
```diff
@@ -262,6 +264,21 @@ impl Runtime for Analytics {
     type Err = ();
 
     async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        let derp_event_future: Pin<Box<dyn Future<Output = Result<Box<Server>, _>> + Send>> =
```
packgron, line 267: futures crate has [BoxFuture](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html) for this.
And also .boxed() method to nicely convert to it, in FutureExt. Use it to remove some boilder plate

packgron: Also you can use [Either](https://docs.rs/futures/latest/futures/future/enum.Either.html) to avoid heap allocation entirely. and use [left/right_future extension methods](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.left_future)

packgron (changes requested):
With this solution then meshent is disabled, we will still wait for 5s until meshnet nodes "reply" with heartbeat. But as that will never really happen, maybe it makes some sense that we report heartbeat info instantly?

packgron (dismissed):
+2.0

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -262,6 +264,18 @@ impl Runtime for Analytics {
     type Err = ();
 
     async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        let derp_event_future = if let Some(channel) = self.io.derp_event_channel.as_mut() {
```
Jauler, line 267: would'nt this mean, that we will not receive anything from multiplexer until we have received DERP event?

crates/telio-nurse/src/nurse.rs
```diff
@@ -163,6 +152,22 @@ impl State {
...
+    /// Inform Nurse that Derp service started.
+    ///
+    /// # Arguments
+    ///
+    /// * `derp` - The new Derp service pointer to use.
+    pub async fn configure_meshnet(&self, meshnet_entities: Option<MeshnetEntities>) {
+        let _ = task_exec!(&self.heartbeat, async move |state| {
+            state.configure_meshnet(meshnet_entities).await;
+
+            telio_log_debug!("Updated meshnet config");
```
Jauler, line 164: Nit: it would be so nice to log whether we are enabling meshnet analytics or disabling here :thinking:

src/device/mod.rs
```diff
@@ -748,6 +774,56 @@ impl RequestedState {
...
+                stop_entity!(Arc::try_unwrap($entity), $name)
+            }};
+        }
+
+        if let Some(direct) = self.direct.take() {
+            // Arc dependency on endpoint providers
+            stop_arc_entity!(direct.cross_ping_check, "CrossPingCheck");
+            drop(direct.endpoint_providers);
+
+            // Arc dependency on wrireguard
```
Jauler, line 806:
```suggestion
            // Arc dependency on wireguard
```

Jauler (approved):
+1.0

Jauler (approved):
+1.0

PR #271, opened 2023-12-20
Revert apple's socket protector api

matislovas:
### Problem
For some unknown reason, `socket2` API for binding sockets to interfaces does not work (Apple's case).

### Solution
Reverting to _old_ method.

- [x] README.md is updated
- [x] changelog.md is updated


tomaszklak (approved):
+2.0, thanks for a quick fix. Independently, maybe create a jira task to try to write a test that would catch this?

PR #272, opened 2023-12-21
LLT-4633 bind to interfaces which has router property only

Jauler:
When libtelio is selecting interfaces candidates to bind to for
underlying VPN configuration, multiple candidates may be available, in
certain circumstances libtelio would select interface without access to
the internet meaning, that VPN connection fails to be established. This
Pull requests adds additional requirement for interface to have "Router"
property when building candidate list.

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -26,7 +26,7 @@ use crate::native::interface_index_from_name;
 use crate::native::NativeSocket;
 use crate::Protector;
 use nix::sys::socket::{getsockname, AddressFamily, SockaddrLike, SockaddrStorage};
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn};
 
 #[cfg(any(target_os = "ios", target_os = "tvos"))]
 use objc::{class, msg_send, runtime::Object, sel, sel_impl};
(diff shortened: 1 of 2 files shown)


crates/telio-sockets/src/protector/apple.rs
```diff
@@ -331,12 +331,15 @@ fn get_primary_interface_names() -> Vec<String> {
             let primary_service_dictionary = store
                 .get(format!("State:/Network/Service/{}/IPv4", service).as_str())
                 .and_then(CFPropertyList::downcast_into::<CFDictionary>);
+
```
Jauler, line 334: Lets add some logs of the interfaces being considered and selected.

crates/telio-sockets/src/protector/apple.rs
```diff
@@ -331,17 +331,27 @@ fn get_primary_interface_names() -> Vec<String> {
...
+                        .map(|ptr| unsafe { CFString::wrap_under_get_rule(*ptr as CFStringRef) });
+                    if let Some(interface_name) = interface_name {
+                        primary_interface_names.push(interface_name.to_string());
+                    }
                 }
             }
         }
     }
 
+    telio_log_info!(
```
baltrus1, line 351: maybe debug would be better? to not spam info logs

baltrus1 (approved):
looks good +2

crates/telio-sockets/src/protector/apple.rs
Jauler: 🤔 Hmm, I suspect this should not be super spammy, and debug will not appear in real usage. Only in debug builds, which may cause finding issues with interface selection harder... Therefore I would be in favor of Info.

twardakm: In my opinion it's OK to have `info` here. This log will be printed while establishing the connection and when interface changes, so not that often

PR #277, opened 2024-01-02
Backport to main: LLT-4633 bind to interfaces which has router property only

Jauler:
When libtelio is selecting interfaces candidates to bind to for underlying VPN configuration, multiple candidates may be available, in certain circumstances libtelio would select interface without access to the internet meaning, that VPN connection fails to be established. This Pull requests adds additional requirement for interface to have "Router" property when building candidate list.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -26,7 +26,7 @@ use crate::native::interface_index_from_name;
 use crate::native::NativeSocket;
 use crate::Protector;
 use nix::sys::socket::{getsockname, AddressFamily, SockaddrLike, SockaddrStorage};
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn};
 
 #[cfg(any(target_os = "ios", target_os = "tvos"))]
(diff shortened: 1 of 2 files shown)


crates/telio-sockets/src/protector/apple.rs
```diff
@@ -334,17 +334,27 @@ fn get_primary_interface_names() -> Vec<String> {
...
+                        .map(|ptr| unsafe { CFString::wrap_under_get_rule(*ptr as CFStringRef) });
+                    if let Some(interface_name) = interface_name {
+                        primary_interface_names.push(interface_name.to_string());
+                    }
                 }
             }
         }
     }
 
+    telio_log_info!(
```
baltrus1, line 354: optional: make debug log

baltrus1 (approved):
+2

crates/telio-sockets/src/protector/apple.rs
Jauler: :thinking: Hmm, I suspect this should not be super spammy, and `debug` will not appear in real usage. Only in debug builds, which may cause finding issues with interface selection harder...

Jauler: I have chatted with Marcin, and they too are kinda in favor of `Info`, so I will leave it as `Info`.

baltrus1 (approved):

PR #278, opened 2024-01-03
Use boringtun with saturating duration it timers.rs

tomaszklak:
This should fix rare crash during startup caused by negative duration.

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (approved):
+2

PR #280, opened 2024-01-03
LLT-4616: Make QoS ping asynchronous

lcruz99:
### Problem
A few lana tests were failing as events were not being listed on the fetched database.

Initial investigation shown that if we have 2 nodes, where alpha supports only ipv6 and beta both ipv4 and ipv6, alpha's QoS fails to ping beta addresses (expectedly) as it doesn't support it. This way QoS will be busy pinging ([see](https://github.com/NordSecurity/libtelio/blob/b9fd30ad227195d46d8a0206d67857d09641220c/crates/telio-nurse/src/qos.rs#L186)), blocking the sending of the heartbeat event [here](https://github.com/NordSecurity/libtelio/blob/de92d11233a084645698ec0b04f5f183dd49c15d/crates/telio-nurse/src/nurse.rs#L228). 

### Solution
Make QoS ping asynchronous, launching non-blocking parallel pings.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/qos.rs
+++ b/crates/telio-nurse/src/qos.rs
@@ -2,8 +2,10 @@ use async_trait::async_trait;
 use histogram::Histogram;
 use serde::Deserialize;
 use std::collections::{BTreeSet, HashMap};
+use std::sync::Arc;
 use std::time::Instant;
 
+use tokio::sync::mpsc;
 use tokio::time::{interval_at, Duration, Interval};
 
 use telio_crypto::PublicKey;
@@ -14,7 +16,7 @@ use telio_utils::{telio_log_debug, DualTarget};
 
 use crate::config::QoSConfig;
 
-use crate::rtt::ping::Ping;
+use crate::rtt::ping::{DualPingResults, Ping};
 
 /// Enum denoting ways to calculate RTT.
 #[derive(Eq, PartialEq, Debug, Deserialize)]
@@ -107,7 +109,7 @@ impl From<AnalyticsEvent> for NodeInfo {
 }
 
 /// QoS data.
-#[derive(Default)]
+#[derive(Debug, Default, PartialEq)]
 pub struct OutputData {
     pub rtt: String,
     pub rtt_loss: String,
@@ -119,6 +121,20 @@ pub struct OutputData {
 }
 
 impl OutputData {
+    /// Create an empty `OutputData`
+    #[allow(dead_code)]
+    pub fn new(buckets: u32) -> Self {
+        OutputData {
+            rtt: Analytics::empty_data(buckets),
+            rtt_loss: Analytics::empty_data(buckets),
+            rtt6: Analytics::empty_data(buckets),
+            rtt6_loss: Analytics::empty_data(buckets),
+            tx: Analytics::empty_data(buckets),
+            rx: Analytics::empty_data(buckets),
+            connection_duration: String::from("0"),
+        }
+    }
+
     /// Create a new `OutputData` containing the contents of two `OutputData`
     pub fn merge(a: OutputData, b: OutputData) -> Self {
         OutputData {
@@ -161,7 +177,9 @@ pub struct Analytics {
     rtt_interval: Interval,
     io: Io,
     nodes: HashMap<PublicKey, NodeInfo>,
-    ping_backend: Option<Ping>,
+    ping_backend: Arc<Option<Ping>>,
+    ping_channel_rx: mpsc::Receiver<(PublicKey, DualPingResults)>,
+    ping_channel_tx: mpsc::WeakSender<(PublicKey, DualPingResults)>,
     buckets: u32,
 }
 
@@ -181,9 +199,14 @@ impl Runtime for Analytics {
                 })
             },
 
-            _ = self.rtt_interval.tick() => {
+            _ = self.rtt_interval.tick(), if self.ping_channel_tx.upgrade().is_none() => {
+                self.perform_ping();
+                Self::next()
+            },
+
+            Some(node_ping_res) = self.ping_channel_rx.recv() => {
                 Self::guard(async move {
-                    self.perform_ping().await;
+                    self.process_node_ping_results(node_ping_res);
                     Ok(())
                 })
             }
@@ -203,10 +226,12 @@ impl Analytics {
     ///
     /// A new `Analytics` instance with the given configuration but with no nodes.
     pub fn new(config: QoSConfig, io: Io) -> Self {
+        let (ping_channel_tx, ping_channel_rx) = mpsc::channel(1);
+
         let ping_backend = if config.rtt_types.contains(&RttType::Ping) {
-            Ping::new(config.rtt_tries).ok()
+            Arc::new(Ping::new(config.rtt_tries).ok())
         } else {
-            None
+            Arc::new(None)
         };
 
         Self {
@@ -214,6 +239,8 @@ impl Analytics {
             io,
             nodes: HashMap::new(),
             ping_backend,
+            ping_channel_rx,
+            ping_channel_tx: ping_channel_tx.downgrade(),
             buckets: config.buckets,
         }
     }
@@ -244,7 +271,6 @@ impl Analytics {
         sorted_public_keys: &BTreeSet<PublicKey>,
     ) -> OutputData {
         let mut output = OutputData::default();
-
         for key in sorted_public_keys {
             if let Some(node) = nodes.get(key) {
                 // RTT
@@ -324,9 +350,8 @@ impl Analytics {
         output
     }
 
-    /// Clear cached data
+    /// Clear cached data.
     pub fn reset_cached_data(&mut self) {
-        // Clear cached data
         for node in self.nodes.values_mut() {
             // Keep the nodes, but clear the histograms
             node.rtt_histogram = Histogram::new();
@@ -339,6 +364,10 @@ impl Analytics {
         }
     }
 
+    /// Update peer data from a received WG event.
+    /// # Arguments
+    ///
+    /// * `event` - Received WG event.
     async fn handle_wg_event(&mut self, event: AnalyticsEvent) {
         telio_log_debug!("WG event: {:?}", event);
         self.nodes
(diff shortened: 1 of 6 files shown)


crates/telio-nurse/src/qos.rs
```diff
@@ -203,8 +232,10 @@ impl Analytics {
     ///
     /// A new `Analytics` instance with the given configuration but with no nodes.
     pub fn new(config: QoSConfig, io: Io) -> Self {
+        let (ping_channel_tx, ping_channel_rx) = mpsc::channel(16);
```
lcruz99, line 235: As this channel won't be used I can reduce the buffer size (or even wrap it in an Option 🤔)

crates/telio-nurse/src/qos.rs
```diff
@@ -348,28 +388,36 @@ impl Analytics {
...
     }
 
-    async fn perform_ping(&mut self) {
-        if let Some(pinger) = &self.ping_backend {
-            for (_, node) in self.nodes.iter_mut() {
-                pinger.perform(node).await;
-            }
+    /// Launch ping task for every node.
+    fn perform_ping(&mut self) {
+        let (ping_channel_tx, ping_channel_rx) = mpsc::channel(32);
```
lcruz99, line 403: TODO: Change buffer size to self.nodes.len()

crates/telio-nurse/src/qos.rs
```diff
@@ -324,9 +356,13 @@ impl Analytics {
         output
     }
 
-    /// Clear cached data
+    #[cfg(test)]
+    pub fn get_nodes_hashmap(&self) -> HashMap<PublicKey, NodeInfo> {
```
lcruz99, line 360: TODO: Remove this fn as it's not needed

nat-lab/tests/test_lana.py
```diff
@@ -105,13 +105,17 @@ def get_moose_db_file(container_tag, container_path, local_path):
...
     start_time = asyncio.get_event_loop().time()
-
+    events = []
     while asyncio.get_event_loop().time() - start_time < DEFAULT_CHECK_TIMEOUT:
         get_moose_db_file(container, CONTAINER_EVENT_PATH, events_path)
         events = fetch_moose_events(events_path)
         if len(events) == nr_events:
             return events
         await asyncio.sleep(DEFAULT_CHECK_INTERVAL)
+    print(
```
dfetti, line 115: Is this print visible anywhere?

lcruz99: Yes, printed during pytest when a test fail to aid debugging

dfetti: Oh, I did not know that prints are visible when a test fails

crates/telio-nurse/src/qos.rs
dfetti: This is actually a semaphore count, not the actual buffer size as far as I know

crates/telio-nurse/src/qos.rs
```diff
@@ -181,9 +199,20 @@ impl Runtime for Analytics {
...
+                })
+            },
+
+            Some(node_ping_res) = self.ping_channel_rx.recv() => {
                 Self::guard(async move {
-                    self.perform_ping().await;
+                    self.nodes
+                        .entry(node_ping_res.public_key)
+                        .and_modify(|n| {
+                            *n = node_ping_res;
```
dfetti, line 214: I think making the pinger send only the ping results is safer.
This thing can destroy good data if something bad happens in the pinger.

crates/telio-nurse/src/qos.rs
lcruz99: wdym by semaphore count? 🤔 

https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html

crates/telio-nurse/src/qos.rs
lcruz99: Agreed, will change

crates/telio-nurse/src/qos.rs
dfetti: https://docs.rs/tokio/latest/src/tokio/sync/mpsc/bounded.rs.html#144-156

tomaszklak (commented):
I'm not sure I understand how exactly this is fixing the problem, mostly because I don't 100% understand why the issue is sporadic. For example, running the `test_lana_with_second_node_joining_later_meshnet_id_can_change[IPStack.IPv6-IPStack.IPv4v6]` locally, always succeeds for me.

crates/telio-nurse/src/qos.rs
```diff
@@ -181,9 +199,16 @@ impl Runtime for Analytics {
                 })
             },
 
-            _ = self.rtt_interval.tick() => {
+            _ = self.rtt_interval.tick(), if self.ping_channel_tx.upgrade().is_none() => {
```
tomaszklak, line 202: Can you clarify what kind of situation is this guard preventing?

crates/telio-nurse/src/qos.rs
```diff
@@ -181,9 +199,16 @@ impl Runtime for Analytics {
...
+                Self::guard(async move {
+                    self.perform_ping();
+                    Ok(())
+                })
+            },
+
+            Some(node_ping_res) = self.ping_channel_rx.recv() => {
                 Self::guard(async move {
-                    self.perform_ping().await;
+                    self.process_node_ping_results(node_ping_res);
```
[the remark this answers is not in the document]
lcruz99: that's correct

tomaszklak:
> I'm not sure I understand how exactly this is fixing the problem, mostly because I don't 100% understand why the issue is sporadic. For example, running the `test_lana_with_second_node_joining_later_meshnet_id_can_change[IPStack.IPv6-IPStack.IPv4v6]` locally, always succeeds for me.

I've managed to reproduce this with those additional logs:

```
$ git diff crates/telio-nurse/src/
diff --git a/crates/telio-nurse/src/nurse.rs b/crates/telio-nurse/src/nurse.rs
index fdba49d..aed9ea5 100644
--- a/crates/telio-nurse/src/nurse.rs
+++ b/crates/telio-nurse/src/nurse.rs
@@ -230,7 +230,9 @@ impl State {
         let internal_sorted_public_keys = info.internal_sorted_public_keys;
         let external_sorted_public_keys = info.external_sorted_public_keys;
         let (internal_qos_data, external_qos_data) = if let Some(qos) = self.qos.as_ref() {
-            task_exec!(qos, async move |state| {
+println!("sending of the heartbeat event start");
+            let tmp= task_exec!(qos, async move |state| {
+
                 let result = (
                     state.get_data(&internal_sorted_public_keys),
                     state.get_data(&external_sorted_public_keys),
@@ -239,7 +241,9 @@ impl State {
                 Ok(result)
             })
             .await
-            .unwrap_or_default()
+            .unwrap_or_default();
+println!("sending of the heartbeat event end");
+tmp
         } else {
             (QoSData::default(), QoSData::default())
         };
diff --git a/crates/telio-nurse/src/qos.rs b/crates/telio-nurse/src/qos.rs
index 44ab7e3..88905b8 100644
--- a/crates/telio-nurse/src/qos.rs
+++ b/crates/telio-nurse/src/qos.rs
@@ -183,7 +183,9 @@ impl Runtime for Analytics {

             _ = self.rtt_interval.tick() => {
                 Self::guard(async move {
+                    println!("perform_ping start");
                     self.perform_ping().await;
+                    println!("perform_ping end");
                     Ok(())
                 })
             }
```

An it looks like this: [lana.txt](https://github.com/NordSecurity/libtelio/files/13887022/lana.txt) and as far as I can see, everything that starts, also ends - there is no task hanging/waiting for something:
```
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
[alpha]: stdout: - started telio with BoringTun:YG/GFjc6dOSs8tqShSboDj3C0b7723uX84qjbBHzo1Y=...
[alpha]: stdout: sending of the heartbeat event start
[alpha]: stdout: sending of the heartbeat event end
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
[alpha]: stdout: sending of the heartbeat event start
[alpha]: stdout: sending of the heartbeat event end
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
[alpha]: stdout: perform_ping start
[alpha]: stdout: perform_ping end
```
And yet the test failed with
```
>           assert alpha_events
E           assert None
```

crates/telio-nurse/src/qos.rs
lcruz99: I didn't think much about it honestly and followed the same pattern, in fact I'm wondering if one of these select branches could be interrupted since it we're implementing wait() fn and not wait_with_update()

crates/telio-nurse/src/qos.rs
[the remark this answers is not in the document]
lcruz99: > mostly because I don't 100% understand why the issue is sporadic.

Pinging invalid endpoints would block the task on perform_ping(), since it's guarded it wouldn't be interrupted..

Reading you're question about the guard I now realize that probably removing the guard would help.. wdyt?

lcruz99:
@tomaszklak thanks right away for the double check.

Could you check if the other nodes also succeed sending the event?

lcruz99:
Just realized you attached the full logs 😅

The logic is the other way around, alpha.db has the events that alpha received from beta, and looking at the logs, beta didn't managed to send them.

tomaszklak:
> Just realized you attached the full logs 😅
> 
> The logic is the other way around, alpha.db has the events that alpha received from beta, and looking at the logs, beta didn't managed to send them.

Yeah, you're right! It didn't occur to me to check the beta side 😓

crates/telio-nurse/src/qos.rs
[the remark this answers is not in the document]
tomaszklak: Yup, I think that the guard is not really needed.

tomaszklak (approved):
+1

PR #282, opened 2024-01-04
Make it possible to run natlab on non x86-64 platforms

tomaszklak:
Installing python-wireguard on eg. linux arm64 is resulting in a not working installation - the hardcoded `py-wireguard.so` (already incorrect name on macos) is always for x86-64, regardless of what the host architecture is.

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (approved):
+2 neat

PR #283, opened 2024-01-04
Remove link state testcase scenarios with short keepalive periods

dfetti:
Sometimes the nodes fail to connect when the persistent keepalive periods are set to a low value, like 5 seconds. As this has nothing to do with the no-link detection mechanism, I removed them until the LLT-4718 will be fixed.

- [x] README.md is updated
- [x] changelog.md is updated


gytsto (approved):
+2

PR #287, opened 2024-01-05
Add go wrap symbols to dynamic library

Savolro:
When building Golang wrapper, modified linker `gold` adds `--whole-archive` flag to the build, resulting in all go wrapper symbols being exported in final dynamic library.

### Problem
Symbols of `go_wrap.c` are embedded into `libtelio.a`, therefore, static linking works. However, when linking `libtelio.so`, all of the wrapper symbols are stripped resulting in failure when linking dynamically. It is complicated to determine Go dependency file location during linking time of an user application.

### Solution
Considering the complications, the fact that `go_wrap.c` is already built by default, and the similar workaround being done for Java bindings, this PR contains a new linker wrapper `gold` (standing for Golang ld) which forces `-lgo_wrap` flag to be wrapped with flags `--whole-archive` and `--no-whole-archive`

- [x] README.md is updated
- [x] changelog.md is updated


linkers/gold
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
tomaszklak, line 1: seems this file and `linkers/dld` only differ in two lines/variables. Could you make it so that we have one common linker wrapper script, that is parametrized with needed changes?

packgron: To do this correctly we need to get language as a parameter, (is it GO or JAVA).
We can do this via https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env, in build.rs. I think we should do this even if we will replace this whole area this quarter.

packgron (approved):
+2.0

PR #292, opened 2024-01-10
[LLT-4557] Document helpers.py

stalowyjez:
### Problem
Functions from helpers.py, which are widely used in our nat-lab tests are undocumented

### Solution
Add some description and, if feasible, an example to each of them


nat-lab/tests/helpers.py
```diff
@@ -19,6 +19,21 @@
 
 @dataclass
 class SetupParameters:
+    """
+    Setup parameters for a single Telio node
+
+    # Attributes
```
mathiaspeters, line 25: Adding the default value for these would be nice, and maybe taking it a step further than just what is specified in the parameter list. For example, the default value for `derp_servers` as listed in the parameter list is `None`, but if you dig more it's actually `config.DERP_SERVERS`

mathiaspeters (commented):
LGTM! +1

ThrasherLT:
LGTM +1

gytsto (approved):
+2

gytsto (approved):

PR #293, opened 2024-01-10
LLT-4528 Mute proxy channels during EP upgrade

matislovas:
### Problem
During the connection upgrade process, peers sometimes begins oscillate between `Relay` and `Direct` EPs. TL;DR this happens because of WG's roaming feature. 

### Solution
When upgrading to `Direct` EP, mute peer in `Proxy` module (only the Egress (telio --> WG) part) for 15 secs. 

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-proxy/src/proxy.rs
```diff
@@ -187,10 +208,61 @@ impl StateEgress {
...
+                    (
+                        socket,
+                        old_sockets
+                            .get(&pk)
+                            .map(|(_, muted)| *muted)
+                            .unwrap_or(false),
+                    ),
+                )
+            })
+            .collect();
```
matszczygiel, line 227: Too much cloning IMHO
```suggestion
        let old_sockets = std::mem::take(&mut self.sockets);
        self.sockets = sockets
            .into_iter()
            .map(|(pk, socket)| {
                (
                    pk,
                    (
                        socket,
                        old_sockets
                            .get(&pk)
                            .map_or(false, |(_, muted)| *muted),
                    ),
                )
            })
            .collect();
```

crates/telio-proxy/src/proxy.rs
```diff
@@ -187,10 +208,61 @@ impl StateEgress {
...
+
+    fn try_mute(&mut self, pk: &PublicKey, dur: Duration) -> Result<(), Error> {
+        if let Some((_, muted)) = self.sockets.get_mut(pk) {
+            telio_log_debug!("Muting peer: {:?} for {:?}", pk, dur);
+            self.mutes.insert(*pk, PinnedSleep::new(dur, *pk));
+            *muted = true;
+            return Ok(());
+        }
+
+        Err(Error::PeerNotFound)
```
matszczygiel, line 240:
```suggestion
        let Some((_, muted)) = self.sockets.get_mut(pk) else {
             return Err(Error::PeerNotFound);
        };
        telio_log_debug!("Muting peer: {:?} for {:?}", pk, dur);
        self.mutes.insert(*pk, PinnedSleep::new(dur, *pk));
        *muted = true;
        Ok(());
```

crates/telio-proxy/src/proxy.rs
```diff
@@ -187,10 +208,61 @@ impl StateEgress {
...
+    }
+
+    fn try_unmute(&mut self, pk: &PublicKey) -> Result<(), Error> {
```
matszczygiel, line 243: Same as above

crates/telio-proxy/src/proxy.rs
```diff
@@ -187,10 +208,61 @@ impl StateEgress {
 
         self.wg_addr = wg_port.map(|p| (Ipv4Addr::LOCALHOST, p).into());
 
-        self.sockets = sockets;
+        let old_sockets = self.sockets.clone();
+        self.sockets = sockets
```
matszczygiel, line 212: Shouldn't you clean up the `self.mutes` of the sockets that are no longer present in `self.sockets` as well?

crates/telio-proxy/src/proxy.rs
```diff
@@ -187,10 +208,61 @@ impl StateEgress {
...
+        for (pk, mute) in self.mutes.iter() {
+            if mute.is_elapsed() {
+                // Cannot try_remove() here because of borrow checker
+                if let Some((_, muted)) = self.sockets.get_mut(pk) {
+                    telio_log_debug!("Cleaning mute from peer {:?}", pk);
+                    *muted = false;
+                }
+            }
+        }
+        self.mutes.retain(|_, mute| !mute.is_elapsed());
```
matszczygiel, line 264: `HashMap::retain` gives you mutable access to the values, so you can write it in a single loop

```suggestion
        for (pk, mute) in self.mutes.iter() {
            if mute.is_elapsed() {
                // Cannot try_remove() here because of borrow checker
                if let Some((_, muted)) = self.sockets.get_mut(pk) {
                    telio_log_debug!("Cleaning mute from peer {:?}", pk);
                    *muted = false;
                }
            }
        }
        self.mutes.retain(|_, mute| if mute.is_elapsed() {
                if let Some((_, muted)) = self.sockets.get_mut(pk) {
                    telio_log_debug!("Cleaning mute from peer {:?}", pk);
                    *muted = false;
                }
                true
            } else {
               false
        });
```

crates/telio-proxy/src/proxy.rs
```diff
@@ -259,10 +331,35 @@ impl Runtime for StateEgress {
...
+                        telio_log_debug!("Peer {:?} is muted", pk);
+                    }
+                }
+            },
+            (pk, _, _) = select_all(mute_futures) => {
+                if let Some((_, muted)) = self.sockets.get_mut(&pk) {
+                    telio_log_debug!("Unmuting peer: {:?}", pk);
+                    *muted = false;
+                }
             }
```
matszczygiel, line 363: I somewhat feel that you should just implement a check when the packet is about to send, meaning that there should be no `select_all()` but a simple check just before the `socket.send_to()` if the peer is muted or even better there should be no `muted` hash map but as hash map with `PublickKey -> Instant (last time muted)` and you should check here if the peer was muted more than 15s ago, then send. It would be much simpler to implement and more reliable IMHO

crates/telio-proxy/src/proxy.rs
lcruz99: look's good

crates/telio-proxy/src/proxy.rs
lcruz99: look's good

crates/telio-proxy/src/proxy.rs
lcruz99: look's good

lcruz99: Here I don't think we can use hashmap::retain() in the way you suggested, self.mutes closure borrows it as immutable and then we borrow self.sockets as mutable. Borrow checker doesn't allows this

matszczygiel: Yeah, but this compiles fine:
```rust
        let socks = &mut self.sockets;

        self.mutes.retain(|pk, mute| {
            if mute.is_elapsed() {
                if let Some((_, muted)) = socks.get_mut(pk) {
                    telio_log_debug!("Cleaning mute from peer {:?}", pk);
                    *muted = false;
                }
                true
            } else {
                false
            }
        });
```

crates/telio-proxy/src/proxy.rs
lcruz99: I partially agree, that simplifies a lot if we know the default sleep would never change, in the future if we accept custom mute durations it will add complexity. So I would prefer to just wait for the mute sleeps to expire

crates/telio-proxy/src/proxy.rs
lcruz99: I've just removed this function on the last push anw

matszczygiel (approved):
+1

lcruz99 (approved):
+1

PR #295, opened 2024-01-11
LLT-4616: Fix lana tests

lcruz99:
### Problem

### Solution


crates/telio-lana/src/lib.rs
```diff
@@ -25,7 +25,7 @@ pub const LANA_APP_NAME: &str = "libtelio";
 /// Version of the tracker used, should be updated everytime the tracker library is updated
 pub const LANA_MOOSE_VERSION: &str = "0.11.0";
 /// Timeout for moose to send init error, if any, through callback
-pub const LANA_MOOSE_MAX_INIT_TIME: u8 = 2;
```
tomaszklak, line 28: Shouldn't this be investigated by the lana/moose team instead? If all that is being done in initialization is some local I/O, than this shouldn't take that long (at least most of the time). Alternatively, if there is no problem in moose, maybe there is something wrong with how our runners are configured if few I/O operations are taking that long.

lcruz99: moose team didn't have an answer for this, it needs further investigation. As we agreed, I'll create a follow-up ticket on this and increase the timeout for the sake of stability.

LukasPukenis (approved):
Looks OK. The delay increase makes sense in test environment, though no idea what LANA does that it takes so long :)

+2

PR #296, opened 2024-01-12
Revert "Merge pull request #251 from NordSecurity/LLT-1113-no-link-de…

Jauler:
There was a small mistake in the connection liveness tracking, which resulted in unstable direct connections, root cause:

The flow is this, during initialization, the peer's NoLinkDetectionInfo is inserted here:
https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/wg.rs#L633
But at first, there is no connection, meaning, that boths last_rx and last_tx is None. as there is no connection at all at that moment. Meaning, that we hit this condition: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/wg.rs#L682 and return None as new_node_info. which later causes removal of the peer from the no_link_detection_info here: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/wg.rs#L712. now all subsequent iterations of that  function will always get None when trying to get NoLinkDetectionInfo for that peer here: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/wg.rs#L673 which causes last_rx to be returned as None from this getter: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/wg.rs#L572

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (approved):
+2

It looks like at least a testcase is needed afterwards and possibly better abstractions in the codebase

PR #298, opened 2024-01-15
Fix CPC unresponsive time threshold

Jauler:
CPC implemented an optimization to avoid sending CMMs and pings to nodes which are offline. The "offline" in the context of this optimization is determined by the time since last handshake. Wireguard attempts new handshake every 120-180s, but somehow the value in the tests for CPC "offline" threshold has been set to 10s (single tests are rarely allowed to run for longer, than single wireguard sessions). This means, that every time, when direct connection is not formed in the first 10s - it will not be attempted anymore. And this introduces non-deterministic failures in the tests.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_direct_connection.py
+++ b/nat-lab/tests/test_direct_connection.py
@@ -37,7 +37,7 @@ def _generate_setup_parameter_pair(
                 direct=Direct(
                     providers=endpoint_providers,
                     skip_unresponsive_peers=SkipUnresponsivePeers(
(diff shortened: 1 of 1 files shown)


matislovas (approved):
+2, let's see if this bad boy fixes these tests

Jauler:
Integration tests showed that in order to fix the issue, https://github.com/NordSecurity/libtelio/pull/238 is needed

PR #299, opened 2024-01-15
LLT-4674: Stabilize DNS queries in natlab tests

mathiaspeters:
### Problem
`test_dns_through_exit` would sometimes fail. Local testing showed that `nslookup` was sometimes not returning any results. 

### Solution
`test_dns.py` had abstracted calls to `nslookup` to a dedicated function but that function was not used in `test_dns_through_exit.py`. That function is now moved to a dedicated utils file to make it easier to share. Additionally, the function was made more stable by slightly increasing how long the test waits for a DNS response, removing the retry option passed to `nslookup` to allow it to retry infinitely, and lastly by setting the internal timeout for `nslookup` to only a second, which means it will retry over and over instead of just waiting.

Before the fix I could reliably reproduce the flakiness of `test_dns_through_exit`, with one of the variants failing on average every 5th time or so. After the fix, I ran every variant for 50 times without a single failure.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_dns.py
+++ b/nat-lab/tests/test_dns.py
@@ -8,11 +8,10 @@ from config import LIBTELIO_DNS_IPV4, LIBTELIO_DNS_IPV6
 from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_api, setup_environment, setup_mesh_nodes
 from telio import AdapterType, TelioFeatures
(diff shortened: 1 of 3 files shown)


dfetti:
+1.0

LukasPukenis (approved):
+1 PR description :100:

PR #300, opened 2024-01-15
natlab: Add test parameters identification

lcruz99:
### Problem
The introduction of helper functions in natlab tests has hidden parameters ids making debugging less transparent.  

### Solution
Add identification function to relevant tests so that devs can identify the parameters for each test.

Test previously described as:
`test_upnp_without_support[beta_setup_params0-alpha_setup_params4]`
with this PR becomes:
`test_upnp_without_support[DOCKER_CONE_CLIENT_2-AdapterType.BoringTun-['upnp']-MAC_VM-AdapterType.BoringTun-['upnp']]`

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -1,5 +1,8 @@
 import asyncio
 import pytest
+from helpers import SetupParameters
+from typing import List, Tuple
+from utils.router import IPStack
 
 
 def _cancel_all_tasks(loop: asyncio.AbstractEventLoop):
@@ -38,3 +41,29 @@ def event_loop():
         finally:
             asyncio.events.set_event_loop(None)
             loop.close()
+
+
+def pytest_make_parametrize_id(config, val):
+    param_id = ""
+    if isinstance(val, (List, Tuple)):
+        for v in val:
+            res = pytest_make_parametrize_id(config, v)
+            if isinstance(res, str):
+                param_id += f"-{res}"
+        param_id = f"{param_id[1:]}"
+    elif isinstance(val, (SetupParameters,)):
+        param_id = f"{val.connection_tag.name }-{val.adapter_type}"
+        if val.features.direct is not None:
+            param_id += f"-{val.features.direct.providers}"
+    elif isinstance(val, IPStack):
+        if val == IPStack.IPv4:
+            param_id = "IPv4"
+        elif val == IPStack.IPv4v6:
+            param_id = "IPv4v6"
+        elif val == IPStack.IPv6:
+            param_id = "IPv6"
+    elif isinstance(val, str):
+        param_id = f"{val}"
+    else:
+        return None
+    return param_id


nat-lab/tests/conftest.py
```diff
@@ -38,3 +40,27 @@ def event_loop():
         finally:
             asyncio.events.set_event_loop(None)
             loop.close()
+
+
+def idfn(val):
```
gytsto, line 45: Can't this be applied for all tests during a collection or using this `pytest_make_parametrize_id` hook, instead of manually importing and assigning this function to every parameter id? Just an idea, not 100% sure if possible

LukasPukenis (commented):
+1 nice and welcome change

matszczygiel:
Good effort overall. However, I'd like to see @gytsto commend addressed first

nat-lab/tests/conftest.py
```diff
@@ -38,3 +40,27 @@ def event_loop():
...
+            param_id = f"'{param_id[1:]}'"
+    elif isinstance(val, (SetupParameters,)):
+        param_id = f"{val.connection_tag.name }-{val.adapter_type}"
+        if val.features.direct is not None:
+            param_id += f"-{val.features.direct.providers}"
+    elif isinstance(val, str):
+        param_id = f"'{val}'"
+    else:
+        return None
+    return param_id
```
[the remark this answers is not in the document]
lcruz99: I've checked and for now is covering every test correctly. In the case the val type is not covered it just returns None and the parameter name is used raw.

> I am afrait some testcases might clash into one for example. 
Only the parameters are being processed not the tests ID, so not possible to happen ^^.

nat-lab/tests/conftest.py
```diff
@@ -38,3 +40,27 @@ def event_loop():
         finally:
             asyncio.events.set_event_loop(None)
             loop.close()
+
+
+def idfn(val):
+    param_id = ""
+    if isinstance(val, (List,)):
+        if isinstance(val[0], SetupParameters):
```
arg0d, line 48: Could this be a recursive call?
```
if isinstance(val, (List,)):
  for v in val:
    param_id += "-" + idfn(v)
```

nat-lab/tests/conftest.py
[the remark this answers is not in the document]
arg0d: It looks like there is some type of deduplication going on under the hood. Producing identical test names creates additional default suffix:
```
import pytest
from dataclasses import dataclass

def idfn(val):
    return "0"

@pytest.mark.parametrize("value", [0, 1], ids=idfn)
def test_aaa(value):
    pass
```

Produces:
```
tests/aaa/test_aaa.py::test_aaa[00] PASSED                                                  [ 50%]
tests/aaa/test_aaa.py::test_aaa[01] PASSED                                                  [100%]
```

nat-lab/tests/conftest.py
arg0d: Oh, if returning `None` uses the default mangling logic, then there should be no issues.

nat-lab/tests/conftest.py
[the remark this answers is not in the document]
lcruz99: Oh I've missed that situation.. That can happen if two tests have the same connection_tag and adapter_type 🤔

nat-lab/tests/conftest.py
lcruz99: Added hook

nat-lab/tests/conftest.py
lcruz99: Added

matszczygiel (approved):
+1

nat-lab/tests/conftest.py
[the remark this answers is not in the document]
lcruz99: However uncovered parameters on the `idfn` will be used raw so identical test names/parameters should not happen

PR #301, opened 2024-01-16
Use pure python internal dns server instead of 1.1.1.1 and 8.8.8.8

tomaszklak:
### Problem
Some of the nat-lab dns related tests forwarded queries to external dns servers (1.1.1.1 and 8.8.8.8). This can lead to transient failures when those external servers are briefly unreachable.

### Solution
This pr adds pure python dns server (based on the twisted framework) that handles small subset of the queries that currently are generated by the nat-lab dns tests. If needed it should be easy to extend the supported queries and/or hostnames.


LukasPukenis (approved):
+2 that's a very welcome change

PR #305, opened 2024-01-18
natlab: Reduce parameters id length

lcruz99:
### Problem
After adding parameterization id function, some tests start to fail due to the long names the logs output file would have. Unix standard length is 256 b/chars.

### Solution
Reduce some parameters length by shortening them.

eg:
`AdapterType.BoringTun` -> `BoringTun`
`DOCKER_CONE_CLIENT_1` -> `CONE_CLIENT_1`
`ConnectionTag.DOCKER_CONE_CLIENT_1` -> `CONE_CLIENT_1`
`2001:db8:85a4::adda:edde:7` -> `2001:db8:85a4:..`


baltrus1:
+1

lcruz99:
For the test we have at the moment it does, in the future it depends on the number of pytest.params. It doesn't guarantee, but I can add that to save_logs() fn in telio.py.

tomaszklak:
> I can add that check to save_logs() fn in telio.py.

Yeah, I think it would be best to handle it now.

PR #306, opened 2024-01-18
[LLT-4420] Fix killing docker processes

stalowyjez:
It improves the experience of running the tests locally in two ways:
 - on some setups you can't just kill the docker process from host (even with sudo)
 - some tests won't need entering sudoer password anymore

### Problem
While running locally tests have sometimes problems with killing docker processes

### Solution
We can map the host PID to the PID inside the container and then kill it inside, which is done in this commit

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/bin/kill_process_by_natlab_id
+++ b/nat-lab/bin/kill_process_by_natlab_id
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+if [[ "$#" -ne 1 ]]; then
+    echo "Wrong number of parameters"
+    exit 1
+fi
+
+NATLAB_ID=$1
+
+for pid in $(ps -e -o pid=); do
+    if grep --null-data --text KILL_ID=${NATLAB_ID} /proc/${pid}/environ; then
+        echo "Killing ${pid}"
+        kill "${pid}"
+        exit 0
+    fi
+done
+
+echo "The process to kill not found"
--- a/nat-lab/tests/utils/process/docker_process.py
+++ b/nat-lab/tests/utils/process/docker_process.py
@@ -1,5 +1,7 @@
 import asyncio
-import os
+import random
+import string
+import subprocess
 import sys
 from .process import Process, ProcessExecError, StreamCallback
 from aiodocker.containers import DockerContainer
@@ -20,6 +22,9 @@ class DockerProcess(Process):
     _is_done: asyncio.Event
     _stream: Optional[Stream]
     _execute: Optional[Exec]
+    _kill_id: (
+        str  # Private ID added to find the process easily when it needs to be killed
+    )
 
     def __init__(self, container: DockerContainer, command: List[str]) -> None:
         self._container = container
@@ -30,6 +35,9 @@ class DockerProcess(Process):
         self._is_done = asyncio.Event()
         self._stream = None
         self._execute = None
+        self._kill_id = "".join(
+            random.choices(string.ascii_uppercase + string.digits, k=12)
+        )
 
     async def execute(
         self,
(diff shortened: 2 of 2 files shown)


nat-lab/tests/utils/process/docker_process.py
```diff
@@ -158,3 +161,30 @@ def is_executing(self) -> bool:
...
+    # for the other platforms we fallback to the old solution
+    if sys.platform not in ("linux", "linux2"):
+        os.system(f"sudo kill -9 {pid}")
+        return
+
+    # It may happen that the process already exited at this point, in which situation
+    # we may expect an error
+    try:
+        docker_pid = (
+            subprocess.check_output(f"cat /proc/{pid}/status | grep NSpid", shell=True)
```
matszczygiel, line 181: I'd like to see it implemented in Python itself, meaning reading and inspecting the `/proc/<pid>/status` file, because it adds a hidden dependency for `cat` and `grep` being installed on the host machine

nat-lab/tests/utils/process/docker_process.py
```diff
@@ -158,3 +161,30 @@ def is_executing(self) -> bool:
...
+
+
+def kill_docker_process_by_host_pid(pid: str, container_id: str):
+    """This function may look a bit weird, because we could just kill the process
+    using its PID on host, but, unfortunately, such kill is often denied due to some
+    security measures (even while using sudo)"""
+
+    # Currenlty the gentle way of killing the docker process is supported only on linux,
+    # for the other platforms we fallback to the old solution
+    if sys.platform not in ("linux", "linux2"):
```
arg0d, line 173:
```suggestion
    if "linux" in sys.platform:
```

or

```suggestion
    if sys.platform.startswith("linux"):
```

nat-lab/bin/kill_process_by_natlab_id
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+NATLAB_ID=$1
+
+for pid in $(ps -e -o pid=); do
```
arg0d, line 5: How many processes are there typically in a container? This looks like `O(N)` algorithm, I wonder if its a problem or not.

stalowyjez: Two are run in the beginning and then we start another few - it looks that it should be less than 10

nat-lab/bin/kill_process_by_natlab_id
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
matszczygiel, line 1: This script introduces even more hidden dependencies: bash, grep, ps.
Are they installed on the nat-lab containers? I would recommend at least using POSIX sh instead of bash

stalowyjez: There are a few things here:
 1. All of the our scripts use bash, I think we should keep this consistent
 2. ps and grep are quite basic tools which are installed on the Ubuntu based docker containers and both of them are already used is our scripts
 3. I think we should strongly prefer extending docker images over adding complexity to the scripts

matszczygiel (dismissed):
+1

nat-lab/tests/utils/process/docker_process.py
```diff
@@ -30,14 +35,19 @@ def __init__(self, container: DockerContainer, command: List[str]) -> None:
         self._is_done = asyncio.Event()
         self._stream = None
         self._execute = None
+        self._kill_id = "".join(
+            random.choices(string.ascii_uppercase + string.digits, k=12)
+        )
 
     async def execute(
```
gytsto, line 42: update process killing in `async def execute` exception handler

gytsto (approved):

PR #307, opened 2024-01-18
LLT-4799: wg: Fix deadlock on interface callbacks when stopping

lcruz99:
### Problem
There are a few places that can deadlock on `dev stop` due to incorrect use of [CancelMibChangeNotify2](https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-cancelmibchangenotify2#remarks).

Specifically:
-  https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/windows/tunnel/interfacewatcher.rs#L108
- https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/windows/tunnel/interfacewatcher.rs#

and

- https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/windows/tunnel/mtumonitor.rs#L99
- https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/windows/tunnel/mtumonitor.rs#L110
- https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/windows/tunnel/mtumonitor.rs#L138
- https://github.com/NordSecurity/libtelio/blob/main/crates/telio-wg/src/windows/tunnel/mtumonitor.rs#L174

The deadlock can happen when the stop function lock’s callback’s handle, at the same time network change triggers the callback’s code on another thread, that tries to acquire the same lock to the callback’s handle. However, [CancelMibChangeNotify2](https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-cancelmibchangenotify2#remarks) will not be complete until that callback is done, both the stopping thread and callback thread are deadlocked.

### Solution
Add an atomic boolean to sinchronize between threads and sinalize that we're stopping telio and no more callbacks should be called.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
+++ b/crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
@@ -340,13 +340,9 @@ impl InterfaceWatcher {
         );
 
         let self_ptr = CallerContext as *mut InterfaceWatcher;
-        if (*self_ptr).iface_cb_handle.clone().lock().is_ok() {
-            if NotificationType == MibAddInstance {
-                (*self_ptr).mib_add_instance(Row);
-            };
-        } else {
-            telio_log_error!("error obtaining lock");
-        }
+        if NotificationType == MibAddInstance {
+            (*self_ptr).mib_add_instance(Row);
+        };
 
         telio_log_trace!(
             "--- InterfaceWatcher::interface_change_callback: CallerContext {:p}, Row {:p}, NotificationType {}",
--- a/crates/telio-wg/src/windows/tunnel/mtumonitor.rs
+++ b/crates/telio-wg/src/windows/tunnel/mtumonitor.rs
@@ -135,18 +135,14 @@ impl MtuMonitor {
         );
 
         let self_ptr = CallerContext as *mut MtuMonitor;
-        if (*self_ptr).route_cb_handle.clone().lock().is_ok() {
-            if 0 == (*Row).DestinationPrefix.PrefixLength {
-                // Result can be ignored
-                match (*self_ptr).do_it() {
-                    Ok(_) => {}
-                    Err(err) => {
-                        telio_log_trace!("MtuMonitor::do_it returned error {}", err);
-                    }
+        if 0 == (*Row).DestinationPrefix.PrefixLength {
+            // Result can be ignored
+            match (*self_ptr).do_it() {
+                Ok(_) => {}
+                Err(err) => {
+                    telio_log_trace!("MtuMonitor::do_it returned error {}", err);
                 }
             }
-        } else {
-            telio_log_error!("error obtaining lock");
         }
 
         telio_log_trace!(
@@ -171,18 +167,14 @@ impl MtuMonitor {
         );
 
         let self_ptr = CallerContext as *mut MtuMonitor;
-        if (*self_ptr).iface_cb_handle.clone().lock().is_ok() {
-            if NotificationType == MibParameterNotification {
-                // Result can be ignored
-                match (*self_ptr).do_it() {
-                    Ok(_) => {}
-                    Err(err) => {
-                        telio_log_trace!("MtuMonitor::do_it returned error {}", err);
-                    }
+        if NotificationType == MibParameterNotification {
+            // Result can be ignored
+            match (*self_ptr).do_it() {
+                Ok(_) => {}
+                Err(err) => {
+                    telio_log_trace!("MtuMonitor::do_it returned error {}", err);
                 }
             }
-        } else {
-            telio_log_error!("error obtaining lock");
         }
 
         telio_log_trace!(


matszczygiel:
The field you're deadlocking on is `Arc<Mutex<usize>>`. Is it possible to change it to `Arc<AtomicUsize>`? Thus reducing mutex overhead and solving the deadlock issue at the same time?

tomaszklak:
The way the flag is used, looks like a job for https://doc.rust-lang.org/std/sync/struct.Once.html :)

jjanowsk:
I think there are more problems here. 

Firstly I think this additional flag may work in this shape (for a deadlock problem) but I think it can be completely avoided. Most likely we do not need to lock `iface_cb_handle` and `route_cb_handle` at all inside callbacks. I don't know what was the reasoning behind that and maybe we can find the original author of this code to explain it, but I spent some time yesterday analyzing it and it seems useless. Maybe the idea was to just serialize the callbacks so two of them do not run in parallel, but they are serialized automatically by the kernel so there is no need to do that. 

> The invocation of the callback function specified in the Callback parameter is serialized
> 
> Source: https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-notifyipinterfacechange#remarks 

That said there is another problem of concurrent access to the shared state (callback's context). While the calls to route and interface callbacks are serialized individually there is nothing stopping one route and one interface callback to be run in parallel. And because they lock separate mutexes they will run in parallel overriding the context. 
I've taken a look at the original implementation in Go and there they use closure as a callback (wrapped of course somehow in a callback function) The point is that every closure will have separate state and in our case they both share the same. 

So in my opinion we should:
1. Remove locking of `iface_cb_handle` and `route_cb_handle` inside our callbacks. This will already solve the deadlock issue.
2. Either keep somewhere small separate contexts for the callback with just `last_luid` `last_mtu` and maybe few more fields that are needed or add separate mutex to our classes that will serialize access across the callbacks. I would lean into the first solution

jjanowsk:
And congrats on the investigation 🍾

Jauler:
> I think there are more problems here.
> 
> Firstly I think this additional flag may work in this shape (for a deadlock problem) but I think it can be completely avoided. Most likely we do not need to lock `iface_cb_handle` and `route_cb_handle` at all inside callbacks. I don't know what was the reasoning behind that and maybe we can find the original author of this code to explain it, but I spent some time yesterday analyzing it and it seems useless. Maybe the idea was to just serialize the callbacks so two of them do not run in parallel, but they are serialized automatically by the kernel so there is no need to do that.
> 
> > The invocation of the callback function specified in the Callback parameter is serialized
> > Source: https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-notifyipinterfacechange#remarks
> 
> That said there is another problem of concurrent access to the shared state (callback's context). While the calls to route and interface callbacks are serialized individually there is nothing stopping one route and one interface callback to be run in parallel. And because they lock separate mutexes they will run in parallel overriding the context. I've taken a look at the original implementation in Go and there they use closure as a callback (wrapped of course somehow in a callback function) The point is that every closure will have separate state and in our case they both share the same.
> 
> So in my opinion we should:
> 
>     1. Remove locking of `iface_cb_handle` and `route_cb_handle` inside our callbacks. This will already solve the deadlock issue.
> 
>     2. Either keep somewhere small separate contexts for the callback with just `last_luid` `last_mtu` and maybe few more fields that are needed or add separate mutex to our classes that will serialize access across the callbacks. I would lean into the first solution
> 
> 
> Edit: Of course point 2 is only relevant for mtumonitor. Interfacewatcher has only one callback. I could also see you creating a JIRA item for the second step and merging just a fix for the deadlock with this PR if we want to have it merged before the weekend so we can see how our tests are doing over the weekend

I would love to collect test failure statistics information over the weekend, and this PR does look functionally correct, even if, @jjanowsk suggestion seems to be better.

@jjanowsk @lcruz99 What do you think about merging this, and removing the unnecessary locking and separating the context for MTU/Interface in a separate task?

jjanowsk:
> > I think there are more problems here.
> > Firstly I think this additional flag may work in this shape (for a deadlock problem) but I think it can be completely avoided. Most likely we do not need to lock `iface_cb_handle` and `route_cb_handle` at all inside callbacks. I don't know what was the reasoning behind that and maybe we can find the original author of this code to explain it, but I spent some time yesterday analyzing it and it seems useless. Maybe the idea was to just serialize the callbacks so two of them do not run in parallel, but they are serialized automatically by the kernel so there is no need to do that.
> > > The invocation of the callback function specified in the Callback parameter is serialized
> > > Source: https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-notifyipinterfacechange#remarks
> > 
> > 
> > That said there is another problem of concurrent access to the shared state (callback's context). While the calls to route and interface callbacks are serialized individually there is nothing stopping one route and one interface callback to be run in parallel. And because they lock separate mutexes they will run in parallel overriding the context. I've taken a look at the original implementation in Go and there they use closure as a callback (wrapped of course somehow in a callback function) The point is that every closure will have separate state and in our case they both share the same.
> > So in my opinion we should:
> > ```
> > 1. Remove locking of `iface_cb_handle` and `route_cb_handle` inside our callbacks. This will already solve the deadlock issue.
> > 
> > 2. Either keep somewhere small separate contexts for the callback with just `last_luid` `last_mtu` and maybe few more fields that are needed or add separate mutex to our classes that will serialize access across the callbacks. I would lean into the first solution
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Edit: Of course point 2 is only relevant for mtumonitor. Interfacewatcher has only one callback. I could also see you creating a JIRA item for the second step and merging just a fix for the deadlock with this PR if we want to have it merged before the weekend so we can see how our tests are doing over the weekend
> 
> I would love to collect test failure statistics information over the weekend, and this PR does look functionally correct, even if, @jjanowsk suggestion seems to be better.
> 
> @jjanowsk @lcruz99 What do you think about merging this, and removing the unnecessary locking and separating the context for MTU/Interface in a separate high prio task?

As I said in the edit, I'm all for it. Second issue can be fixed next week

Jauler (approved):
+1.0

@lcruz99 Please create issue for @jjanowsk suggestions ;).

jjanowsk:
But I would like to see removing those locks in callback instead of having this flag already in this review. It's a 2 min change and it will guarantee no deadlock. Current flag solution still contains a race. Imagine two threads running in parallel one executing callback and one stopping function. You can still have a situation when the thread running callback function will read `is_stopping` flag as false, then stopping thread will set this flag to true and lock the mutex first before callback thread can do this. The deadlock window is now narrower then before but it still exists. Removing locks inside callbacks will make it fully functional
@Jauler @lcruz99

Jauler:
> But I would like to see removing those locks in callback instead of having this flag already in this review. It's a 2 min change and it will guarantee no deadlock. Current flag solution still contains a race. Imagine two threads running in parallel one executing callback and one stopping function. You can still have a situation when the thread running callback function will read `is_stopping` flag as false, then stopping thread will set this flag to true and lock the mutex first before callback thread can do this. The deadlock window is now narrower then before but it still exists. Removing locks inside callbacks will make it fully functional @Jauler @lcruz99

Do we have any sizeable number of local runs with locks removed? I can try to attempt them.

jjanowsk (approved):
+2

PR #309, opened 2024-01-19
LLT-4799 Disable management interface during tests

Jauler:
### Describtion
libvirt uses management interface to configure and/or provision the virtual machines, unfortunately this interface is completely controlled by libvirt and can issue/renew DHCP leases any time, especially if default route to it has been deleted (we used to do that in tests). This introduces competition against primary/secondary interfaces default routes which is used for testing. During tests - management interface should never be used, therefore this PR ensures, that this interface is disabled during tests.
    
Note that this will cause `vagrant ssh` to not work during tests, but one can always use `ssh vagrant@10.55.0.11` to connect to windows machines.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection/docker_connection.py
+++ b/nat-lab/tests/utils/connection/docker_connection.py
@@ -2,14 +2,17 @@ from .connection import Connection, TargetOS
 from aiodocker.containers import DockerContainer
 from typing import List
 from utils.process import Process, DockerProcess
+from datetime import datetime
 
 
 class DockerConnection(Connection):
     _container: DockerContainer
 
-    def __init__(self, container: DockerContainer):
+    def __init__(self, container: DockerContainer, container_name: str):
         super().__init__(TargetOS.Linux)
+        self._name = container_name
         self._container = container
 
     def create_process(self, command: List[str]) -> "Process":
+        print(datetime.now(), "Executing", command, "on", self._name)
         return DockerProcess(self._container, command)
--- a/nat-lab/tests/utils/connection/docker_connection.py
+++ b/nat-lab/tests/utils/connection/docker_connection.py
@@ -1,8 +1,8 @@
 from .connection import Connection, TargetOS
 from aiodocker.containers import DockerContainer
+from datetime import datetime
 from typing import List
 from utils.process import Process, DockerProcess
-from datetime import datetime
 
 
 class DockerConnection(Connection):
--- a/nat-lab/tests/utils/network_switcher/network_switcher_windows.py
+++ b/nat-lab/tests/utils/network_switcher/network_switcher_windows.py
@@ -2,7 +2,7 @@ import config
 import re
 from .network_switcher import NetworkSwitcher
 from dataclasses import dataclass
-from typing import List
+from typing import List, Optional
 from utils.connection import Connection
 from utils.process import ProcessExecError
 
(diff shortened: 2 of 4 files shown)


gytsto:
What if we were to make NetworkSwitcher as context manager and when we are exiting we just enable Management interface?

Jauler:
> What if we were to make NetworkSwitcher as context manager and when we are exiting we just enable Management interface? This would allow to use `vagrant ssh` if I understand correctly. 🤔

Yeah, I was thinking about that. But maybe lets create separate issue and PR for that, as I want to avoid DHCP bug over the weekend nightlies, and I am afraid not to finish everything with context manager.

Jauler:
> > > What if we were to make NetworkSwitcher as context manager and when we are exiting we just enable Management interface? This would allow to use `vagrant ssh` if I understand correctly. 🤔
> > 
> > 
> > Yeah, I was thinking about that. But maybe lets create separate issue and PR for that, as I want to avoid DHCP bug over the weekend nightlies, and I am afraid not to finish everything with context manager today
> 
> OK, waiting for ticket number in replies 😄

LLT-4812

nat-lab/tests/utils/connection/docker_connection.py
```diff
@@ -1,15 +1,18 @@
...
     _container: DockerContainer
 
-    def __init__(self, container: DockerContainer):
+    def __init__(self, container: DockerContainer, container_name: str):
         super().__init__(TargetOS.Linux)
+        self._name = container_name
         self._container = container
 
     def create_process(self, command: List[str]) -> "Process":
+        print(datetime.now(), "Executing", command, "on", self._name)
```
LukasPukenis, line 17: :+1:

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -131,3 +134,16 @@ async def _delete_route(self, interface_name: str) -> None:
                 pass
             else:
                 raise exception
+
+    async def _disable_management_interface(self) -> None:
+        if self._interfaces.default is not None:
```
LukasPukenis, line 139: is it worth adding an assert so `self._interfaces.default` would be present or do we have cases where that's expected?

Jauler: Yes, if the interface is disabled at the start of the tests, `self._interfaces.default` will be `None`.

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -41,17 +41,20 @@ def __init__(self, default: str, primary: str, secondary: str) -> None:
...
-            assert False, f"interface not found with prefix `{prefix}`, {interfaces}"
+            return None
 
-        return ConfiguredInterfaces(
-            find_interface(config.LIBVIRT_MANAGEMENT_NETWORK_PREFIX),
-            find_interface(config.PRIMARY_VM_NETWORK_PREFIX),
-            find_interface(config.SECONDARY_VM_NETWORK_PREFIX),
-        )
+        # Allow management interface to be shut down
+        management_itf = find_interface((config.LIBVIRT_MANAGEMENT_NETWORK_PREFIX))
```
LukasPukenis, line 51: I think it would be a light improvement to assert on management interface being present if the target OS is Windows and MacOS

LukasPukenis:
+1

LukasPukenis (approved):

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
Jauler: Yeah, I was thinking about MacOS too, but on MacOS routing is different, and there is no competition with DHCP advertised routes versus windows. So right now I have changed only how windows is being configured.

LukasPukenis (approved):

PR #311, opened 2024-01-22
natlab: Add dig timeout to allow process failure

lcruz99:
### Problem
On `test_dns_port`, `dig` command is used which has by default 5 seconds timeout for each retry (3 tries by default). 
Enabling magic_dns can take sometimes more than we expect and the 1s timeout from `wait_normal` might not be enough.

### Solution
Increase command timeout from `wait_normal`->`wait_long` and assert we get process error when the 3 tries are finished.

There might be still an issue with libtelio, however as I could not reproduce the issue locally and the logs are no longer available these changes should place us in a better position to debug it if the test fails again in the future.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_dns.py
+++ b/nat-lab/tests/test_dns.py
@@ -8,10 +8,9 @@ from config import LIBTELIO_DNS_IPV4, LIBTELIO_DNS_IPV6
 from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_api, setup_environment, setup_mesh_nodes
 from telio import AdapterType, TelioFeatures
-from utils import testing
 from utils.connection_tracker import ConnectionLimits
 from utils.connection_util import ConnectionTag, generate_connection_tracker_config
-from utils.dns import query_dns
+from utils.dns import query_dns, query_dns_port
 from utils.process import ProcessExecError
 from utils.router import IPStack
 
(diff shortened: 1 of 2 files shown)


nat-lab/tests/test_dns.py
```diff
@@ -190,36 +190,58 @@ async def test_dns_port(alpha_ip_stack: IPStack) -> None:
...
+                    [
+                        "dig",
+                        "+timeout=1",
+                        "@" + dns_server_address_alpha,
+                        "-p",
+                        "53",
+                        "google.com",
+                    ]
                 ).execute()
             )
```
mathiaspeters, line 205: What do you think about extracting this to its own function, either in this file or `dns.py`?

lcruz99: I think is not that benefitial like with `nslookup` on `query_dns`, it wouldn't hurt though..  will do it

nat-lab/tests/test_dns.py
```diff
@@ -190,36 +190,58 @@ async def test_dns_port(alpha_ip_stack: IPStack) -> None:
...
                 connection_alpha.create_process(
-                    ["dig", "@" + dns_server_address_alpha, "-p", "53", "google.com"]
+                    [
+                        "dig",
+                        "+timeout=1",
+                        "@" + dns_server_address_alpha,
+                        "-p",
+                        "53",
+                        "google.com",
+                    ]
```
LukasPukenis, line 203: As a possible improvement I think we could use `dig` with one of our own services inside of the docker, what do you think about that?

LukasPukenis (commented):
+1

nat-lab/tests/test_dns.py
mathiaspeters: It's already doing that. `dig` will communicate with the node's internal DNS address, which will delegate to, in this case, the DNS docker container that Tomasz created for anything that is not looking up the IP of a meshnet peer

nat-lab/tests/test_dns.py
mathiaspeters: For sure not as beneficial since it's only used in one test, but if we need to change how we're calling `dig`, it's better to just have to change in one place

nat-lab/tests/test_dns.py
lcruz99: As an improvement I agree however for this PR might be already out of scope

LukasPukenis: Ooops! Google.com is just a placeholder, `dns_server_address_alpha` is already there :facepalm:

PR #314, opened 2024-01-24
Resolve LLT-4812: Reenable windows management interface on natlab.

gytsto:
### Problem
Windows management interface is beind disabled during tests, but never reenabled, causing `vagrant ssh windows` to not work. 

### Solution
Reenable windows management interface once test is done.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_network_switch.py
+++ b/nat-lab/tests/test_network_switch.py
@@ -52,12 +52,13 @@ async def test_network_switcher(
         )
 
         assert conn_mngr.network_switcher
-        await conn_mngr.network_switcher.switch_to_secondary_network()
-
-        assert (
-            await testing.wait_long(stun.get(conn_mngr.connection, config.STUN_SERVER))
-            == secondary_ip
-        )
+        async with conn_mngr.network_switcher.switch_to_secondary_network():
+            assert (
+                await testing.wait_long(
+                    stun.get(conn_mngr.connection, config.STUN_SERVER)
+                )
+                == secondary_ip
+            )
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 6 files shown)


LukasPukenis (commented):
It seems that this PR is squashed to one commit and heavily inflated with formatting changes making it much harder to track down what were the functional changes. For the future I suggest doing so.

The change seems to be symmetrical to the interface disabling which was made before so it makes perfect sense but maybe it's possible to do the disabling and enabling a level or two above? What I mean is that `switch_to_[primary|secondary]_network` look  a bit too low level for interface manipulation, would be nice to have it somewhere separated as such:
```
disable_management_interface()
run_tests()
enable_management_interface()
```

gytsto:
> It seems that this PR is squashed to one commit and heavily inflated with formatting changes making it much harder to track down what were the functional changes. For the future I suggest doing so.

If you were to look more closely, there is no unnecessary formatting changes. Some of the parts were converted to `asynccontextmanagers`, therefore, every code below that were affected by it had to be indented.

gytsto:
> The change seems to be symmetrical to the interface disabling which was made before so it makes perfect sense but maybe it's possible to do the disabling and enabling a level or two above? What I mean is that `switch_to_[primary|secondary]_network` look a bit too low level for interface manipulation, would be nice to have it somewhere separated as such:
> 
> ```
> disable_management_interface()
> run_tests()
> enable_management_interface()
> ```

I could do that, but we already have this setup for manipulating routes/interfaces. I don't see a need move this particular part to a higher level for now. Possibly would introduce too many changes and maybe instabilities.

Jauler (approved):
+1.5

dfetti (approved):
+0.5

LukasPukenis:
Re: formatting - my bad, `async with` slipped through my eyes and so it looked like some sort of formatting changes at a glance.

Still I think they have nothing to do with the interface enable/disable flow and could go into a separate commit for easier understanding :)

PR #315, opened 2024-01-24
Dont notify socket waiter after cleanup

Hasan6979:
### Problem
After turning off meshnet on macOs, there are a series of "Unable to bind" error messages.

### Solution
This happens because during cleanup sockets waiters are notified which tries to rebind to all the sockets again. As it is during cleanup, default interface sockets have already been closed and it is not able to bind hence the error message.

Removing the notify fixes the problem. 

- [x] README.md is updated
- [x] changelog.md is updated


packgron:
+2.0
It makes sense, we don't need to rebind all if just one was deleted.

packgron (dismissed):

packgron (approved):

PR #316, opened 2024-01-25
[LLT-4729] Fix upload-maven ci/cd job

packgron:
### Problem
Gradle invokes project in to ci/cd jobs
package-aar and upload-maven and in both of them
gradle project is in different dirs.

Using absolute path from build_libetlio script
causes an error as both of them ar run in
different jobs and as such path also changes.

### Solution
This implementation switches to using file(..)
function that get paths relative to project dir.
They can be either android_aar or android_aar/main
due to gradle weirdness :/.
findManifest checks 4 possible locations.

- [x] README.md is updated
- [x] changelog.md is updated


gytsto (commented):
+1.0

Jauler:
We should probably remove the templating logic in here: https://github.com/NordSecurity/rust_build_utils/blob/main/rust_build_utils/android_build_utils.py#L88 

But it would be separate thing, do not block this MR on that one ;).

Jauler (approved):
+1.0

PR #317, opened 2024-01-25
[LLT-4802] Fix the connection lag after restart

stalowyjez:
### Problem
Original task was already done in the solution for LLT-4981. We need to additionaly remove timeout from network_switch_direct test which should now be obsolete.

### Solution
In this PR we remove the mentioned timeout, it also simplifies select_endpoint_for_peer function a bit

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_network_switch.py
+++ b/nat-lab/tests/test_network_switch.py
@@ -335,8 +335,5 @@ async def test_mesh_network_switch_direct(
             await relay
             await direct
 
-        # TODO: workaround for LLT-4802, remove after that is fixed
-        await asyncio.sleep(25)
-
         async with Ping(alpha_connection, beta.ip_addresses[0]).run() as ping:
             await testing.wait_long(ping.wait_for_next_ping())
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -246,13 +246,15 @@ async fn consolidate_wg_peers<
                 // local side endpoint, such that the other node can do this upgrade too.
                 let public_key = requested_peer.peer.public_key;
 
-                if let (Some(remote_endpoint), Some(local_direct_endpoint)) = (
-                    requested_peer.peer.endpoint,
-                    requested_peer.local_direct_endpoint,
-                ) {
-                    if let Some(us) = upgrade_sync {
+                if let Some(us) = upgrade_sync {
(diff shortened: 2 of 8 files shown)


packgron (approved):
+1.0

nat-lab/tests/test_network_switch.py
```diff
@@ -320,14 +318,12 @@ async def test_mesh_network_switch_direct(
         async with run_async_contexts(
             [
                 derp_connected_future,
-                peers_connected_relay_future,
```
Jauler, line 323: Are we not getting relayed connection event anymore?

src/device/wg_controller.rs
```diff
@@ -642,14 +642,21 @@ async fn build_requested_meshnet_peers_list<
...
-            &upgrade_request_endpoint,
-        )
-        .await?;
+            checked_endpoint,
+            proxy_endpoint,
+            upgrade_request_endpoint.as_ref(),
+        );
+
+        // If checked endpoint is consumed, we need to remove it from CPC
+        if selected_remote_endpoint == checked_endpoint.map(|ep| ep.remote_endpoint) {
```
Jauler, line 654: maybe it would be better to call `consume_verified_endpoint()` in the place where we in-fact advertise it to the other side via the `request_upgrade()` call?

I am afraid, that there are several steps between here and the endpoint being actually requested for upgrade, that some condition may override it, and then we would be consuming it incorrectly.

nat-lab/tests/test_network_switch.py
stalowyjez: Well, after the last change I guess I get them again, but they seems to be unnecessary - I got just Direct -> Direct endpoint change recently, which seems to be fine as well, but doesn't pass the test in its original form

nat-lab/tests/test_network_switch.py
```diff
@@ -312,27 +312,23 @@ async def test_mesh_network_switch_direct(
...
+        ]) as (derp, direct):
             await exit_stack.enter_async_context(
                 network_switcher.switch_to_secondary_network()
             )
             await alpha_client.notify_network_change()
             await derp
-            await relay
             await direct
 
         # TODO: workaround for LLT-4802, remove after that is fixed
```
gytsto, line 334: don't forget this TODO

PR #318, opened 2024-01-25
Llt 4827 wireguard go version check

jjanowsk:
### Problem
1. There is a bug somewhere in our build system. Because of that it might happen that the old version of wireguard-go is linked into the release package. The bug is not found yet but this should protect us from releasing incorrect package.
2. Wireguard-go logs from panic are lost when telio.dll is run inside a windows service.
### Solution
1. Temporary solution is to add an export with a version string in it to the wireguard-go library. Because it requires manual effort to update it a github actions job is added to monitor that the version is updated every time there are changes done to the wireguard-go lib itself. 
2. Print stacktrace to our logger in case of a panic

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/wg_go_version_check.yml
+++ b/.github/workflows/wg_go_version_check.yml
@@ -1,0 +1,23 @@
+name: Wireguard-go version update check
+on: [pull_request]
+permissions: {}
+
+jobs:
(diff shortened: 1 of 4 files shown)


packgron (approved):
+2.0

PR #319, opened 2024-01-25
Add network info to Mac VM logs

ThrasherLT:
### Problem
It is currently difficult to reproduce LLT-4600 and a couple of more Mac VM related flaky NAT-Lab tests.

### Solution
Added network interface name and IP, routing table and the last 20 lines of syslog to a separate log file for MAC VM.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -1,24 +1,25 @@
 import asyncio
+import datetime
 import json
 import os
 import re
 import shlex
-import datetime
 from collections import Counter
-from config import DERP_PRIMARY, DERP_SERVERS
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
-from dataclasses_json import dataclass_json, DataClassJsonMixin
 from enum import Enum
-from mesh_api import Node, Meshmap
+from typing import AsyncIterator, List, Optional, Set
+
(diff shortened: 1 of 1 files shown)


nat-lab/tests/telio.py
```diff
@@ -870,7 +885,7 @@ async def save_logs(self) -> None:
         os.makedirs(log_dir, exist_ok=True)
 
         log_content = await self.get_log()
-
+        network_info_info = await self.get_network_info()
```
LukasPukenis, line 888: I think it would be a slight improvement by moving it nearby where the writing happens

LukasPukenis (approved):
+2

packgron:
I think together with this let's add system logs.
There is log command in macos, and you can specify --start and --end to see what happened at via time of test's specifically
https://www.manpagez.com/man/1/log/

nat-lab/tests/telio.py
ThrasherLT: Done

ThrasherLT:
> I think together with this let's add system logs. There is log command in macos, and you can specify --start and --end to see what happened at via time of test's specifically https://www.manpagez.com/man/1/log/

Done

packgron (approved):

PR #325, opened 2024-01-26
Add no-link detection mechanism

dfetti:
Add a hinting mechanism for no-link detection. The event generation can be configured using Features and the field in the event is optional. More info about this can be found in RFC-LLT-0045.

This commit also includes nat-lab tests for this detection mechanism and a fix in is_same_event() discovered while working on it.

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-model/src/api_config.rs
```diff
@@ -342,6 +349,8 @@ pub struct Features {
     /// Flag to turn on post quantum VPN tunnel
     #[serde(default)]
     pub post_quantum_vpn: Option<FeaturePostQuantumVPN>,
+    /// No link detection mechanism
+    pub no_link_detection: Option<FeatureNoLinkDetection>,
```
matszczygiel, line 353:
```suggestion
    #[serde(default)]
    pub no_link_detection: Option<FeatureNoLinkDetection>,
```
?

crates/telio-wg/src/no_link_detection.rs
```diff
@@ -0,0 +1,522 @@
...
+        };
+
+        let (tx_bytes, tx_ts) = if new_tx > self.last_tx_bytes {
+            // Ignore the keepalive messages
+            // This only works in a limited fashion because we poll every second and
+            // the detection might not work as intended if there are more packets within that time window.
+            // On the sender side persistent-keepalives might interfere as those are emitted regardless the activity,
+            // meaning for example if persistent-keepalive is set at 5seconds then it might interfere with the
+            // passive-keepalive which is a constant of 10seconds.
+            let tx_ts = if new_tx - self.last_tx_bytes == 32 {
```
matszczygiel, line 236: Can you elaborate on what `32` means here?

crates/telio-wg/src/wg.rs
```diff
@@ -501,55 +515,80 @@ impl State {
...
-            } else {
-                return false;
-            };
-            if !try_send(PeerState::Connecting, peer.clone()).await {
-                return false;
-            }
+            let peer = to
+                .peers
+                .get(key)
+                .cloned()
```
matszczygiel, line 572:
```suggestion
                .cloned()
```

crates/telio-wg/src/wg.rs
```diff
@@ -559,16 +598,26 @@ impl State {
                 let old_state = old.state();
                 let new_state = new.state();
 
+                let no_link_detection_update_result =
+                    self.no_link_detection
+                        .update(key, new.rx_bytes, new.tx_bytes, new_state, push);
+
                 #[allow(clippy::collapsible_if)]
```
matszczygiel, line 605:
```suggestion
```

crates/telio-model/src/api_config.rs
```diff
@@ -303,6 +303,13 @@ where
     Ok(Some(eps))
 }
 
+/// Turns on the no link detection mechanism
+#[derive(Default, Copy, Clone, Debug, PartialEq, Eq, Deserialize)]
+pub struct FeatureNoLinkDetection {
+    /// Configurable rtt in seconds
+    pub rtt_seconds: Option<u64>,
```
matszczygiel, line 310: As in the previous node, I'm voting for using `#[serde(default = "...")]` here. It would save you a couple of `match`es down the road

crates/telio-wg/src/no_link_detection.rs
dfetti: 32 is the size of a keepalive packet. I'll make it a constant.

dfetti: We want to ignore them from the sending perspective

crates/telio-wg/src/no_link_detection.rs
```diff
@@ -0,0 +1,522 @@
...
+    api_config::FeatureNoLinkDetection,
+    mesh::{LinkState, NodeState},
+};
+
+/// Default wireguard keepalive duration
+const WG_KEEPALIVE: Duration = Duration::from_secs(10);
+/// Default configurable rtt value (15 sec)
+const DEFAULT_CONFIGURABLE_RTT: Duration = Duration::from_secs(15);
+
+pub struct NoLinkDetection {
```
matszczygiel, line 15: This structure seems to serve a dual purpose, depending on whether no link detection is enabled or disabled. My suggestion is to have two private structures
```rust
struct NoLinkDetectionEnabled{
    ...
}
````
and 
```rust
struct NoLinkDetectionEnabled {
     ...
}
```
Then putting these into the common structure with public API:
```rust
pub enum NoLinkDetecction {
    Enabled(NoLinkDetectionEnabled),
    Disabled(NoLinkDetectionDisabled),
}
```
alternatively (I don't like this) to have a `Box<dyn NoLinkDetection>` where `NoLinkDetection` is a trait implemented by both private structures.
Why? Because it will save you a ton of state consistency checks in the methods like `if self.is_disabled()` and stuff.

matszczygiel, line 15: BTW `LinkDetection` would be a better name than `NoLinkDetection` IMHO

crates/telio-wg/src/wg.rs
```diff
@@ -202,21 +206,31 @@ impl DynamicWg {
...
     ///         },
+    ///         None,
     ///     );
     /// }
     /// ```
-    pub fn start(io: Io, cfg: Config) -> Result<Self, Error>
+    pub fn start(
+        io: Io,
+        cfg: Config,
+        no_link_detection: Option<FeatureNoLinkDetection>,
```
matszczygiel, line 216: Maybe we can put it into `Config`?

crates/telio-wg/src/no_link_detection.rs
matszczygiel: Ok. But is it possible that some other packet has 32 bytes? Thus breaking our logic here?

crates/telio-wg/src/no_link_detection.rs
```diff
@@ -0,0 +1,524 @@
...
+            let rtt = self.rtt();
+            let new_info = old_info.compute_new_node_info(rx_bytes, tx_bytes, rtt);
+            let should_notify = self.should_notify(
+                node_state,
+                old_info.last_link_state,
+                new_info.last_link_state,
+            );
+            let link_state = new_info.last_link_state;
+
+            self.info.insert(*public_key, new_info);
```
LukasPukenis, line 89: The comment speaks like it shouldn't happen but it feels like a natural cause of calling update() the first time?

crates/telio-wg/src/no_link_detection.rs
```diff
@@ -0,0 +1,524 @@
...
+
+        time::advance(Duration::from_secs(1)).await;
+
+        nld.insert(&key, None, None, Some(LinkState::Up));
+        check_node_info(
+            nld.info.get(&key).unwrap(),
+            0,
+            0,
+            Instant::now(),
+            Instant::now(),
```
LukasPukenis, line 349: What is the reason of advancing time in here? I assumed because we might call `update()` but we basically insert item into the table?

crates/telio-wg/src/no_link_detection.rs
```diff
@@ -0,0 +1,524 @@
...
+        assert_eq!(nld.time_since_last_rx(&key), Some(last_change.elapsed()));
+
+        time::advance(Duration::from_secs(1)).await;
+        nld.update(&key, Some(15), Some(15), NodeState::Connected, false);
+        assert_eq!(nld.time_since_last_rx(&key), Some(Instant::now().elapsed()));
+        let last_change = Instant::now();
+
+        time::advance(Duration::from_secs(1)).await;
+        nld.update(&key, Some(15), Some(15), NodeState::Connected, true);
+        assert_eq!(nld.time_since_last_rx(&key), Some(last_change.elapsed()));
```
LukasPukenis, line 415: Not exactly sure what these asserts check to be honest, as I understand we advance time, insert and then check so the time of update would show to now(time of insert) but that could be only one small testcase. Could you explain why do we need so many?

crates/telio-wg/src/no_link_detection.rs
```diff
@@ -0,0 +1,524 @@
...
+
+        time::advance(Duration::from_secs(1)).await;
+        nld.update(&key, None, None, NodeState::Connected, true);
+        assert_eq!(nld.time_since_last_rx(&key), Some(last_change.elapsed()));
+    }
+
+    #[test]
```
LukasPukenis, line 422: In cases where the parameters are of same type it makes sense to have a separate Enum describing the state explicitly or having a structure that can be passed. It's pretty easy to misplace them by mistake. Same goes for `update()`.

If you think it's a good idea I suggest doing a followup MR for improvements only

nat-lab/tests/telio_features.py
```diff
@@ -3,6 +3,10 @@
 from typing import Optional, List
 
 
+def ExcludeIfNone(value) -> bool:
```
LukasPukenis, line 6: This does not have a call site, nor can I find it anywhere in the codebase, could you explain it? thanks!

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,480 @@
...
+        ]
+
+        for _ in range(0, 40):
+            await asyncio.sleep(1)
+            async with Ping(connection_alpha, beta.ip_addresses[0]).run() as ping:
+                await testing.wait_long(ping.wait_for_next_ping())
+            async with Ping(connection_beta, alpha.ip_addresses[0]).run() as ping:
+                await testing.wait_long(ping.wait_for_next_ping())
+
+        # Expect no link event while peers are active
```
LukasPukenis, line 135: no `nolink` event? :D

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,480 @@
...
+                ),
+                features=TelioFeatures(
+                    no_link_detection=NoLinkDetection(rtt_seconds=1),
+                    wireguard=long_persistent_keepalive_periods(),
+                ),
+            )
+        ),
+    ],
+)
+async def test_event_link_state_peers_exchanging_data_then_idling_then_resume(
```
LukasPukenis, line 390: What was the fix so this testcase would pass?

LukasPukenis: Hmmm... So at this state after communication goes silent and then continues it will keep repeating Down as far as I remember? If so we need to ticket it and fix it afterwards as it will be hit all the time in real life unless I understood something wrong.(might not be the case as libtelio has persistent keepalives enabled always I think)

dfetti: We have LLT-4753 for that. And yes, the flicker will happen if the peer will not reply in one second, the wg update interval.

crates/telio-model/src/api_config.rs
matszczygiel: @dfetti what with this one?

matszczygiel:
+1

crates/telio-model/src/api_config.rs
dfetti: Done

nat-lab/tests/test_events_link_state.py
```diff
@@ -0,0 +1,412 @@
...
+        alpha_events = client_beta.get_link_state_events(alpha.public_key)
+        beta_events = client_alpha.get_link_state_events(beta.public_key)
+
+        # 1 down when node is Connecting and 1 up when node is Connected
+        assert alpha_events == [LinkState.Down, LinkState.Up]
+        # beta will have 2 down events: 1 when is Connecting and 1 detected
+        assert beta_events == [LinkState.Down, LinkState.Up, LinkState.Down]
+
+
+# TODO: Add the scenario where peers exchange data then idle then resume
```
mathiaspeters, line 351: Will this be addressed in a follow-up PR?

dfetti: Yes, for sure.
Also the comments left unresolved from this PR.

LukasPukenis:
+1 assuming direct peer connections are being tested and not breaking

crates/telio-wg/src/link_detection.rs
```diff
@@ -0,0 +1,477 @@
...
+
+        time::advance(Duration::from_secs(1)).await;
+
+        // Update with different rx_bytes
+        ld.update(&key, Some(10), None, NodeState::Connected, false);
+        assert_eq!(ld.time_since_last_rx(&key), Some(Instant::now().elapsed()));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_compute_new_node_info_with_no_link_state_updates() {
```
mathiaspeters, line 417: This test function doesn't do anything async so it can be a standard `#[test]`. I also think it makes sense to define `check_node_info` inside this fn since it's the only place it's called, and I think if you define it as a closure instead of a function and do that after creating `old`, you don't even need to pass old to the closure. That explanation might be a bit unclear so:
```
let name = "telio";
let speak = || println!("Hello, {name}!");
speak();
```

dfetti: The test is async because I need time to be paused, If you look down I assert with Instant::now().
I moved check_node_info into a closure. I will think for a better solution in the next PR.

mathiaspeters: My bad about the async!

mathiaspeters (dismissed):
+1

mathiaspeters (approved):

mathiaspeters (approved):

PR #331, opened 2024-01-30
LLT-4874: Add FFI functions to create Features and Config objects from JSON strings

mathiaspeters:
### Problem
The FFI functions for Telio currently take strongly typed objects, but consumers might have a platform-specific representation of those types and would have an easier time passing json-encoded strings. 

### Solution
To make it easier for those consumers, this commit introduces functions to create objects from such strings. At this time we only see a need to do that for Features and Config, so the conversion functions have only been implemented for those types

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4885,8 +4885,8 @@ checksum = "fc72304796d0818e357ead4e000d19c9c174ab23dc11093ac919054d20a6a7fc"
 
 [[package]]
 name = "uniffi"
-version = "0.25.0"
-source = "git+https://github.com/NordSecurity/uniffi-rs.git?tag=v0.25.0-1#de1e773204d0ddcda58faac0a77570f1f8056679"
+version = "0.3.1+v0.25.0"
+source = "git+https://github.com/NordSecurity/uniffi-rs.git?tag=v0.3.1+v0.25.0#1a5dc527165589456c3d7233107917e065192b03"
 dependencies = [
  "anyhow",
  "uniffi_build",
@@ -4896,8 +4896,8 @@ dependencies = [
 
 [[package]]
 name = "uniffi_bindgen"
-version = "0.25.0"
-source = "git+https://github.com/NordSecurity/uniffi-rs.git?tag=v0.25.0-1#de1e773204d0ddcda58faac0a77570f1f8056679"
+version = "0.3.1+v0.25.0"
+source = "git+https://github.com/NordSecurity/uniffi-rs.git?tag=v0.3.1+v0.25.0#1a5dc527165589456c3d7233107917e065192b03"
 dependencies = [
  "anyhow",
  "askama",
@@ -4919,8 +4919,8 @@ dependencies = [
 
 [[package]]
 name = "uniffi_build"
-version = "0.25.0"
-source = "git+https://github.com/NordSecurity/uniffi-rs.git?tag=v0.25.0-1#de1e773204d0ddcda58faac0a77570f1f8056679"
+version = "0.3.1+v0.25.0"
+source = "git+https://github.com/NordSecurity/uniffi-rs.git?tag=v0.3.1+v0.25.0#1a5dc527165589456c3d7233107917e065192b03"
 dependencies = [
  "anyhow",
  "camino",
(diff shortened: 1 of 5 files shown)


gytsto (commented):
+1.0 lgtm

LukasPukenis:
Could you give a bit more of an explanation what it does and why, I tried to follow but find it hard? FFI functions are very simplistic C-like functions, I am speaking about `no_mangle` ones, in Libtelio those accept just char pointers, so maybe the description is a bit ambiguous...

JSON configs are already being passed if I understand libtelio correctly.

A more in depth explanation would be very appreciated!

mathiaspeters:
@LukasPukenis Sorry, this is the PR description being bad. The change is for UniFFI. It's true that in the older swig FFI code, the functions take simple pointers that we deserialize into json-strings and then strongly typed objects, but in UniFFI world out FFI functions already take those strongly typed objects. This is how we want those function to look like, but there are cases like on apple platforms where our apps have a different representation for some of those structures, and it would be better for them to be able to pass a json string. To enable their case, a utility function is introduced which does that conversion so they can then call the relevant UniFFI-specified FFI functions with strongly typed objects

LukasPukenis:
@mathiaspeters thanks for explaining that this is about UniFFI, this really brings understanding :) . Isn't then this going from JSON strings to UniFFI + Types and now doing it backwards UniFFI + Types -> JSON(again)?

mathiaspeters:
@LukasPukenis I'm not sure I fully understand what you mean, but I'll try to explain in more details, sorry if I'm explaining something other than what you're asking about.

With the old swig bindings, we could have a function like this: `fn telio_set_meshnet(dev: &telio, cfg: *const c_char)`, where the consumer can basically just pass their json string as-is and we do all the work inside of the function. With UniFFI, that function would look like `fn set_meshnet(&self, cfg: Config)` where a complete `Config` object is passed, which for some apps will mean that they will either need to refactor their whole code base to use our `Config` definition, or maintain a mapping whenever they need to interact with libtelio, which they kinda had before except that the mapping was to json. This is a simple pseudocode example of before and after. 
```
let json = ...;

// Before
set_meshnet(json);

// After
let cfg = string_to_config(json);
set_meshnet(cfg);
```

I hope that clears up the intention, and if it doesn't, feel free to keep asking

LukasPukenis:
Thanks @mathiaspeters :) you explained it clear. I personally think that allowing multiple inputs(in this case direct Config parameter or build with some helper) increases the surface area for bugs to appear, for example now if platform X complains that libtelio doesn't work properly and we inspect the logs and see the config is weird.

Earlier we had only 2 places to look at:
- either app passed from JSON
- either libtelio does something wonky

Now we will have 4 places to look at:
- either apps passed bad config to libtelio
- either libtelio does something wonky
- either apps pass wrong json to the converter function
- or the converter function is buggy itself

Jauler (approved):
+1.0

mathiaspeters:
@LukasPukenis I disagree with the surface area for bugs increasing, I would say it stays the same, and in some cases even decreases.

With the old code there was only one way to call the function, that always had the same possible points of failure:
1. Is the pointer valid?
2. Is the json valid?
3. Did the conversion json->cfg work?
4. Did libtelio handle the config properly?

With the new UniFFI code that same list would now be:
1. Is the json valid?
2. Did the conversion json->cfg work?
3. Did libtelio handle the config properly?
But steps 1 and 2 are now optional. 

Additionally, if the app passes invalid json, we would get conversion errors, so we don't really need to care too much about that. What we do need to worry about is when libtelio behaves incorrectly, is it doing the wrong thing for the given config, and that isn't changing now. And as long as we know if the platform is using the conversion function, we can quickly find out if we need to look at potentially parsing the json incorrectly, but that json parsing can also be very easily unit tested if we want to increase our confidence in it.

LukasPukenis:
@mathiaspeters I have no doubts UniFFI is the right approach to decrease the surface as you mention, but having UniFFI and utility that converst from json to UniFFI types increases the surface :)

Basically what I am saying is - if we provide two ways for apps to do the same thing, then we have two things to worry about instead of just one.

I may, however, lack the context maybe there's more to it, maybe apps complained about the transition to using UniFFIied-Telio and this is a warranted change. Sorry about the introduced friction :)

mathiaspeters:
@LukasPukenis Discussions are good! While you are correct that two things may have a bigger bug surface than just one thing, we're just splitting the old functionality into two different functions :) And there's wasn't a specific complain about that but it was brought up by Rytis or Povilas that one of the apps has their own way to represent `Features` or `Config` (I don't remember which), and providing these conversion functions (which again, are just extracting the old functionality to a separate function) means that they can reuse their json serialization logic. It is meant to make it easier to integrate, and easier to upgrade, for example, if we add optional feature config values that an app doesn't care about, or doesn't care about yet.

LukasPukenis:
@mathiaspeters thanks for explaining. I really enjoyed discussing it :)

LukasPukenis (approved):

PR #338, opened 2024-02-05
Fix wireguard-go linking issue

jjanowsk:
### Problem
When #[link] attribute is used cargo and sccache are not aware of this dependency. Because of that old version of wireguard-go-rust-wrapper can be fetched from cache. It is a problem because this fetched .rlib contains wireguard-go static lib embedded inside (which might be out of date). 
### Solution
Instead lets use cargo:rustc-link-lib command. This way cargo is aware of this dependency and sccache will take wireguard-go.lib as part of dependency hash.
This commit also removes safety check put in place before the rootcause for linking with old version of wireguard-go was found.

- [x] README.md is updated
- [x] changelog.md is updated


matszczygiel:
@jjanowsk FYI we've disabled the usage of sccache in the build repo

packgron (approved):

PR #341, opened 2024-02-05
Improve no-link detection mechanism

dfetti:
Convert the entire mechanism into a state machine. This should fix the flicker detailed in LLT-4753.


LukasPukenis:
It would be better if PR descriptions would contain the rationale and reasoning behind the changes. Even better if the commits inside the PR would have that. "flicker" is very undescriptive without context

crates/telio-wg/src/link_detection.rs
```diff
@@ -118,70 +107,57 @@ impl LinkDetectionEnabled {
...
+            .peers
+            .get(public_key)
+            .map(|p| p.current_link_state())
+            .unwrap_or(LinkState::Down);
+
+        let link_state = if matches!(node_state, NodeState::Connected) {
             link_state
+        } else {
+            LinkState::Down
         };
```
matszczygiel, line 157:
```suggestion
        let link_state = if matches!(node_state, NodeState::Connected) {
            self
            .peers
            .get(public_key)
            .map(|p| p.current_link_state())
            .unwrap_or(LinkState::Down)
        } else {
            LinkState::Down
        };
```

crates/telio-wg/src/link_detection.rs
```diff
@@ -240,75 +216,170 @@ pub struct LinkDetectionUpdateResult {
...
+        stats: BytesAndTimestamps,
+    },
+    PossibleDown {
+        stats: BytesAndTimestamps,
+        count: usize,
+    },
+    Up {
+        stats: BytesAndTimestamps,
+    },
 }
```
matszczygiel, line 230: IMHO would be nicer to do something like this:
```suggestion
struct State {
    stats: BytesAndTimestamps,
    variant: StateVariant,
}
enum StateVariant {
    Down,
    PossibleDown {
        count: usize,
    },
    Up,
}
```
It may reduce noise in the `update()` function

crates/telio-wg/src/link_detection.rs
```diff
@@ -240,75 +216,170 @@ pub struct LinkDetectionUpdateResult {
...
+    }
+
+    fn current_link_state(&self) -> LinkState {
+        match self {
+            State::Up { .. } | State::PossibleDown { .. } => LinkState::Up,
+            State::Down { .. } => LinkState::Down,
+        }
+    }
+
+    fn is_link_up(stats: &BytesAndTimestamps, rtt: Duration) -> bool {
```
matszczygiel, line 333: It should be `BytesAndTimestampls` method IMHO

crates/telio-wg/src/link_detection.rs
dfetti: Good idea <3 I did it

crates/telio-wg/src/link_detection.rs
```diff
@@ -240,75 +216,170 @@ pub struct LinkDetectionUpdateResult {
...
+        node_state: NodeState,
         rtt: Duration,
-    ) -> Self {
-        let new_rx = new_rx.unwrap_or(self.last_rx_bytes);
-        let new_tx = new_tx.unwrap_or(self.last_tx_bytes);
+    ) -> LinkDetectionUpdateResult {
+        // Guard if node_state is not connected
+        // self -> new with down and return
+        if !matches!(node_state, NodeState::Connected) {
+            *self = State::new(rx_bytes, tx_bytes, node_state);
```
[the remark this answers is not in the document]
dfetti: NodeState != Connected is intended to act as a reset "button" on the fsm. 
It's ok to reset the timestamps as when the node comes back to the connected state we want it to do a natural transition to link_state up.

matszczygiel (dismissed):
+1

crates/telio-wg/src/link_detection.rs
```diff
@@ -240,77 +213,167 @@ pub struct LinkDetectionUpdateResult {
...
-#[derive(Clone, Copy, Debug)]
-struct NodeInfo {
-    last_tx_timestamp: Instant,
-    last_rx_timestamp: Instant,
-    last_tx_bytes: u64,
-    last_rx_bytes: u64,
-    last_link_state: LinkState,
+enum StateVariant {
+    Down,
+    PossibleDown { count: usize },
```
LukasPukenis, line 218: What does `count` refer to in the `PossibleDown` variant?

LukasPukenis (commented):
I see no nat-lab test related the new functionality, however as I am aware nolink tests are removed, is there a plan to bring them back + new tests?

crates/telio-wg/src/link_detection.rs
dfetti: Do you think I should add a diagram? As this one https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/endpoint_providers/stun.rs#L383

dfetti:
Yes, testing is what I'm working on now. Do you think I should do it in a different PR?

LukasPukenis:
It would be nice for tests to appear in the same PR as the functionality as else it's very easy to miss it.. Of course this is a special case so the next PR hopefully would be the testcases :)

crates/telio-wg/src/link_detection.rs
LukasPukenis: Diagram is nice and it _might_ be beneficial but it might get outdated pretty quickly if the code changes. I suggest on improving the `count` parameter. The issue I am having with it is that it's hard to deduce how long does the `count` work for. As we now have context we know it's most probably seconds as we update once per second. For this I would suggest adding a timestamp in the future when it will be expired and set it to `now + 3 seconds` or something like that :)

By the way, I am not dissing the diagram, I think it's a great tool so if you see a quick way of doing so, it would be nice to have.

dfetti:
I'll add the tests in a different commit in this PR. I can't merge it now anyway due to nat-lab issues.

matszczygiel (approved):

PR #345, opened 2024-02-07
LLT-4724 implement PMTU discovery mechanism for Linux/Android

matszczygiel:
### Problem
We want to optimize the MTU of the VPN interface

### Solution
Probe the PMTU to the VPN server to see if it's possible to increase the hardcoded MTU values

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,377 @@
...
+
+            telio_log_debug!(
+                "Sending ICMP Echo Request, dst: {} seq: {seq}, size: {pmtu}",
+                self.host
+            );
+
+            if let Err(err) = self.sock.send(pkg.bytes(pmtu)).await {
+                // The kernel might retun this error when the packet size == kernel's estimate of PMTU.
+                // But the packet is sent anyway
+                if err.raw_os_error() != Some(libc::EMSGSIZE) {
```
LukasPukenis, line 117: Suggestion: instead of the comment, we can leave the URL to https://man7.org/linux/man-pages/man7/udp.7.html
and copy an excerpt:
```
       By default, Linux UDP does path MTU (Maximum Transmission Unit)
       discovery.  This means the kernel will keep track of the MTU to a
       specific target IP address and return EMSGSIZE when a UDP packet
       write exceeds it.  When this happens, the application should
       decrease the packet size.
```

crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,377 @@
...
+                PmtuResult::Ok => {
+                    pmtu_range.start = pmtu;
+                    (pmtu_range.start + pmtu_range.end) / 2
+                }
+                PmtuResult::Dropped => {
+                    pmtu_range.end = pmtu;
+                    (pmtu_range.start + pmtu_range.end) / 2
+                }
+                PmtuResult::FragNeeded => {
+                    let mtu = read_error_pmtu(&self.sock, self.host.is_ipv6())?;
```
LukasPukenis, line 67: Just wondering. How reliable is it trusting the kernel error message PMTU in favour of our own?
There are huge benefit from our own in this case as it's totally reliable, kernel doesn't really know more than we do.

nat-lab/docker-compose.yml
```diff
@@ -508,6 +509,24 @@ services:
         ipv4_address: 10.0.80.83
         ipv6_address: 2001:db8:85a4::adda:edde:8
 
+  # A host which drop all packets with size larger than 1300
```
LukasPukenis, line 512: a tiny tiny improvement: could we pass 1300 via entrypoint?

nat-lab/tests/test_pmtu.py
```diff
@@ -0,0 +1,199 @@
...
+from helpers import SetupParameters, setup_environment, setup_connections
+from telio import AdapterType
+from telio_features import TelioFeatures
+from utils import testing, stun
+from utils.connection_util import (
+    ConnectionTag,
+    generate_connection_tracker_config,
+    ConnectionLimits,
+)
+from utils.ping import Ping
```
LukasPukenis, line 13: Hmmm... How do we know our ipv6 implementation works then? are there ways to fix Docker for that or maybe we should reconsider having ipv6 support for MTU. Does it make sense to support it for ipv6 at all?

nat-lab/tests/test_pmtu.py
```diff
@@ -0,0 +1,199 @@
...
+from utils.router import IPStack
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "setup_params",
+    [
+        pytest.param(
+            SetupParameters(
+                ip_stack=IPStack.IPv4,
```
LukasPukenis, line 23: was it intended to be named `black_hole`? :)

nat-lab/tests/test_pmtu.py
```diff
@@ -0,0 +1,199 @@
...
+        #            causing `connect()` call to fail.
+        # pytest.param(
+        #     SetupParameters(
+        #         ip_stack=IPStack.IPv6,
+        #         connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_1,
+        #     )
+        # ),
+    ],
+)
+async def test_pmtu_with_nexthop(setup_params: SetupParameters) -> None:
```
LukasPukenis, line 76: I wonder if it would make sense to wrap `inject-icmp-host-unreachable` into a docker container as well? It might be a bit more bloaty but then again - very easy to debug(you can just ssh into it and run tcpdump for example and inspect iptables) and also it would be more uniform with the existing setup

LukasPukenis (commented):
+1

I really like the implementation. It's very nice to read. I left some comments here and there about possible improvements and some questions

crates/telio-lana/src/event_log_file.rs
```diff
@@ -473,4 +491,27 @@ pub mod moose {
             _ => Err(Error::EventLogError),
         }
     }
+
+    #[allow(non_snake_case)]
+    /// Mocked moose function.
+    pub fn send_developer_logging_log(
```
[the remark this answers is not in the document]
matszczygiel: Yeah, it seems it didn't

crates/telio-model/src/api_config.rs
```diff
@@ -343,6 +343,17 @@ impl FeatureLinkDetection {
...
 }
 
+/// Whether to enable PMTU discovery for VPN connection
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Deserialize)]
+#[serde(transparent)]
+pub struct FeaturePmtuDiscovery(pub bool);
+
+impl Default for FeaturePmtuDiscovery {
+    fn default() -> Self {
+        Self(true)
```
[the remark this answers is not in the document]
matszczygiel: Yeah, there's no code to execute on other platforms so no problem either

src/device/mod.rs
```diff
@@ -708,6 +712,36 @@ impl Device {
...
+                            rt.entities.socket_pool.make_external(sock.as_raw_fd());
+                            sock.probe_pmtu().await
+                        };
+
+                        Ok(future.await.map_err(Error::PmtuProbe))
+                    })
+                    .await?
+                }
+                Err(_) => {
+                    // The device is not started. Let us create runtime then
```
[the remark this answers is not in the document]
matszczygiel: This call is not exposed to ffi layer but is only used by tcli in nat-lab. Erroring out would break the nat-lab tests for PMTU. Besides what's wrong with creating ephemeral runtime here?

crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,377 @@
...
+const TIMEOUT: Duration = Duration::from_secs(1);
+const RETRIES: u32 = 3;
+
+impl PMTUSocket {
+    pub fn new(host: IpAddr) -> io::Result<Self> {
+        let sock = crate::create_icmp_socket(&host)?;
+        set_pmtu_on(&sock, host.is_ipv6())?;
+        Ok(Self { sock, host })
+    }
+    pub async fn probe_pmtu(&self) -> io::Result<u32> {
```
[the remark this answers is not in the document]
matszczygiel: Yeah, it was done (CI would warn). Newline is not required according to rustfmt

crates/telio-pmtu/src/linux.rs
matszczygiel: This passage is not relevant here and we operate on ICMP protocol

crates/telio-pmtu/src/linux.rs
matszczygiel: Um, I don't quite get what you mean. There are cases when kernel produces way wrong answer here, though no by its fault. For example when the router responds with `nexthop = 0` then kernel assumes pmtu of 552. Which might not be true. That's why we don't set bound on kernel response but treat it as a hint

nat-lab/docker-compose.yml
matszczygiel: Um, probably. The packets are dropped with the help of iptables configuration. Making it not hardocded would increase the complexity without benefit IMHO

nat-lab/tests/test_pmtu.py
matszczygiel: We do have IPv6 VPN server as far as I know so it would be nice to have PMTU support for it. As for docker, it's not really docker's fault but the way local networks of IPv6 are constructed. The problem is that the IP ranges overlap between different network interfaces and kernel doesn't know to which interface bind the socket

nat-lab/tests/test_pmtu.py
matszczygiel: Um, yeah. The term refers to router not responding with ICMP errors

nat-lab/tests/test_pmtu.py
matszczygiel: Um, it is running inside the clients container

crates/telio-pmtu/src/linux.rs
LukasPukenis: My bad. The correct excerpt seems to be this one:
```
    If the message is too long to pass atomically through the
       underlying protocol, the error EMSGSIZE is returned, and the
       message is not transmitted.
```

from the https://man7.org/linux/man-pages/man2/send.2.html

My point being that comment tells me that despite the error it was ommited but there doesn't seem to be a manual entry guarding that. Linking an appropriate documentation would help a lot in this case.

Of course if what's happening is not what's described in the manual then we have another problem.

crates/telio-pmtu/src/linux.rs
LukasPukenis: Sorry for being too ambiguous maybe - I mean you have a manual calculation just above these lines of code doing binary bisect - can we just use that instead and ignore the kernel altogether? This way it's gonna be much simpler :)

nat-lab/docker-compose.yml
LukasPukenis: Wouldn't it be just passing `$1` instead of `1300` in the entrypoint script?

nat-lab/tests/test_pmtu.py
LukasPukenis: I would love to learn more about it - is this inherent problem for ipv6 or is this somehow related to our setup?

nat-lab/tests/test_pmtu.py
LukasPukenis: I get it, but ther'es a typo, it's `back` hole instead of `black` :)

nat-lab/tests/test_pmtu.py
LukasPukenis: Hmm... Shouldn't it run on the server's container instead?

LukasPukenis: I mean it might work both ways, injecting packets directly on the client works but having it on the server would be a cleaner approach. A docker-container-per-case

crates/telio-pmtu/src/linux.rs
LukasPukenis: That makes things a bit more complicated but I guess that's just how it is. Thanks for explaining

crates/telio-pmtu/src/linux.rs
matszczygiel: It might seem so but sometimes it's hard to devise criteria for considering a packet being too big to be transmitted. The kernel reports errors according to its own calculations regardless of whether the packet was transmitted. So it's easier to track down what the kernel does than blindly guess what had happened.
My first implementation was at one point ignoring the kernel reports but still, it was erroring out on `send()` and `recv()` calls sometimes without a reason. The current implementation is more reliable in that sense and is based on the `ping` command code and still, it treats the kernel-tracked pmtu as a hint.
So long story short, it's easier to cooperate with the kernel that fight with it

LukasPukenis: Fair enough, thanks for explaining.

My main concern is about having two solutions where both are not very realiable it seems but look like they do the same thing.

Might I offer adding the reasoning on top for the future readers so they would be confused less?

src/device/mod.rs
```diff
@@ -1639,6 +1680,10 @@ impl Runtime {
                     exit_node.public_key,
                 );
             }
+
+            if let Some(pmtud) = &mut self.entities.pmtu_detection {
+                pmtud.run(addr.ip()).await;
```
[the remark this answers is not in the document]
matszczygiel: I've added log inside that `start()` call to not pollute log on platforms not supporting PMTU discovery

Cargo.toml
```diff
@@ -68,6 +69,7 @@ rustls-platform-verifier = { git = "https://github.com/tomaszklak/rustls-platfor
 [target.'cfg(windows)'.dependencies]
 winapi = { workspace = true, features = ["ntdef", "winerror"] }
```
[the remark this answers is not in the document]
matszczygiel: Done

src/device/mod.rs
```diff
@@ -1726,6 +1771,10 @@ impl Runtime {
             pq_entt.stop();
         }
 
+        if let Some(pmtud) = &mut self.entities.pmtu_detection {
+            pmtud.stop().await;
```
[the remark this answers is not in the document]
matszczygiel: I've added it in the stop() implementation to not pollute logs for other platforms

nat-lab/docker-compose.yml
matszczygiel: Yeah, but what's the point? Presumably, no one will ever change this

nat-lab/tests/test_pmtu.py
matszczygiel: The thing is that this address in present on more than one network https://stackoverflow.com/questions/12260003/connect-returns-invalid-argument-with-ipv6-address

nat-lab/tests/test_pmtu.py
matszczygiel: Oh, haven't noticed that. Let me fix this

nat-lab/tests/test_pmtu.py
matszczygiel: This script simply doesn't work on server because of iptables rules. I tried this approach but failed.

nat-lab/bin/inject-icmp-host-unreachable
```diff
@@ -0,0 +1,33 @@
...
+from scapy.all import *
+
+NEXTHOP = 1300
+
+
+def callback(pkt):
+    if not pkt.haslayer(ICMP):
+        return
+
+    if pkt[ICMP].type == 8:
```
Jauler, line 13: A super small thing, but maybe it is worth adding some sort of constant or comment explaining, that `8` is the `Echo Request` type?

nat-lab/bin/pmtu-probe
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+iptables -I INPUT -m length --length 1301:1500 -j DROP
```
Jauler, line 5: UUu nice way of simulating MTU bottleneck link :+1:

nat-lab/bin/inject-icmp-host-unreachable
matszczygiel: Done

crates/telio-pmtu/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn drop(&mut self) {
+        if let Some(task) = self.task.take() {
+            task.abort();
+        }
+    }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(unix)]
+fn create_icmp_socket(host: &std::net::IpAddr) -> std::io::Result<tokio::net::UdpSocket> {
```
Jauler, line 108: Note that this socket does need to be "taken care of" in order to prevent it from going inside of a tunnel. 

* For linux, one needs to [setup fwmark](https://github.com/NordSecurity/libtelio/blob/v4.2.4/crates/telio-sockets/src/protector/linux.rs#L46)
* For android one needs to call [VPNService.protect()](https://developer.android.com/reference/android/net/VpnService#protect(int)) we already have a mechanism to do that via callbacks established.
* If other platforms would be needed, platform-specific mechanisms would be required too.

Keeping :arrow_up: in mind, I would recommend to move socket creation logic to `telio-sockets` module, as that module is responsible for exactly that, and have platform-dependant mechanisms (called "protectors") already implemented :).

Jauler: Oh wait, you are using `socket_pool.make_external(...)` which does mean, that all the required configuration for socket is done.

So this is purely structural thing, not functional, maybe that function would still belong close to `new_external_tcp_v4` or `new_internal_udp` or all other socket creation functions. As it is similar in purpose? What do you think?

crates/telio-pmtu/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+// We assume that 1000 is always supported.
+//1500 is the Ethernet MTU so we're not expecting PMTU to be higher than that.
+#[allow(unused)]
+const PMTU_RANGE: Range<u32> = 1000..1501;
+
+#[allow(unused)]
+const ICMP_ECHO_REQ_IDENT: u16 = 0x6C74; // "lt" (libtelio) in hex
+
+#[allow(unused)]
+const INTERVAL: Duration = Duration::from_secs(10 * 60); // 10 mins
```
Jauler, line 21: I am thinking, that this may be slightly too frequent for our needs, and can generate a bit too much quality metrics traffic when is really needed. Maybe lets make it 1h?

crates/telio-pmtu/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+            socket_pool.make_external(sock.as_raw_fd());
+            sock.probe_pmtu().await
+        }
+
+        let pool = self.socket_pool.clone();
+        let task = tokio::spawn(async move {
+            let mut timer = tokio::time::interval(INTERVAL);
+
+            loop {
+                timer.tick().await;
```
Jauler, line 73: Could you remind me what is `tokio::time::interval(...)` behavior in terms of sleep? E.g. if the process/device is sleeping, will "tick()" drop missed events, or will it try to "catch-up"?

crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,378 @@
...
+        Ok(Self { sock, host })
+    }
+
+    pub async fn probe_pmtu(&self) -> io::Result<u32> {
+        telio_log_debug!("Connecting to: {}", self.host);
+        self.sock.connect((self.host, 0)).await?;
+
+        telio_log_debug!("Connected to: {}", self.host);
+
+        let mut pmtu = read_pmtu(&self.sock, self.host.is_ipv6())?;
```
Jauler, line 28: If I am not mistaken, in case `read_mtu()` returns `Err(...)` we will silently fail. Maybe it is worth considering some `warn` or `info` level log in case of `read_pmtu()` failure in order to make debugging easier.

nat-lab/tests/test_pmtu.py
matszczygiel: Fixed

crates/telio-pmtu/src/lib.rs
matszczygiel: TBH I'd prefer keeping the PMTU module decoupled from the VPN itself. The primary reason is not to require PMTU nat-lab tests to set up the VPN things and the second one is to allow usage of PMTU for manual tests without the telio device configuration beforehand.
The `make_external` call does the job well IMHO, so all the requirements are ticked but the convenience and decoupling are preserved for the current implementation

crates/telio-pmtu/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub use platform::PMTUSocket;
+
+// We assume that 1000 is always supported.
+//1500 is the Ethernet MTU so we're not expecting PMTU to be higher than that.
+#[allow(unused)]
+const PMTU_RANGE: Range<u32> = 1000..1501;
+
+#[allow(unused)]
+const ICMP_ECHO_REQ_IDENT: u16 = 0x6C74; // "lt" (libtelio) in hex
```
Jauler, line 18: Could we generate this ID randomly instead of hardcoding it?

crates/telio-pmtu/src/lib.rs
matszczygiel: It will catch up, although it's configurable. Sleep just speels and `tick()` does more or less `sleep(INTERVAL - <time it takes for loop to turn once>)`

crates/telio-pmtu/src/linux.rs
matszczygiel: I don't think it can fail just like that. But you're right, we cannot guarantee that it won't fail so let's handle this cleanly :+1:

crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,378 @@
...
+use pnet_packet::{icmp, icmpv6, Packet};
+use telio_utils::telio_log_debug;
+use tokio::net::UdpSocket;
+
+pub struct PMTUSocket {
+    sock: UdpSocket,
+    host: IpAddr,
+}
+
+const TIMEOUT: Duration = Duration::from_secs(1);
```
Jauler, line 12: IMHO, I would recommend bumping this to 5s.

crates/telio-pmtu/src/linux.rs
Jauler: well, we probably will not be able to do anything too much about that error, than just report that PMTU detection has failed. But it would be nice to see in the logs what was exact reason for it failing, and here the logs come in :).

crates/telio-pmtu/src/lib.rs
Jauler: :thinking: I think it would be nice to configure it to skip the events, as otherwise we will be spamming many PMTU detection events after waking up from sleep.

crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,378 @@
...
+                            err.ee_errno,
+                            err.ee_origin,
+                            err.ee_type,
+                            err.ee_code,
+                            err.ee_pad,
+                            err.ee_info,
+                            err.ee_data,
+                        );
+
+                        if err.ee_errno == libc::EMSGSIZE as u32 {
```
Jauler, line 368: I might be missing some part, but I wonder whether it would make sense to take the last error of type EMSGSIZE in this case instead of first one? could it be possible, that some errors would be queued from either previous runs, or from this run, but different MTU attempt making the error stale?

nat-lab/tests/test_pmtu.py
```diff
@@ -0,0 +1,201 @@
...
+        client, *_ = env.clients
+        connection, *_ = [conn.connection for conn in env.connections]
+
+        host = (
+            config.PMTU_PROBE_HOST_IP4
+            if setup_params.ip_stack in [IPStack.IPv4, IPStack.IPv4v6]
+            else config.PMTU_PROBE_HOST_IP6
+        )
+
+        proc = connection.create_process(["/opt/bin/inject-icmp-host-unreachable"])
```
Jauler, line 90: I wonder if we are sure that there is no race against "normal" kernel generated Echo responses and the injected ICMP error message.

crates/telio-pmtu/src/linux.rs
matszczygiel: Not necessarily, we can use some kind of default value if it fails. The PMTU read here is to provide a reasonable starting point for the search

nat-lab/tests/test_pmtu.py
matszczygiel: There shouldn't be since we have iptables rule to drop ICMP over 1400 bytes

crates/telio-pmtu/src/linux.rs
matszczygiel: I don't think it is possible. We do the queue check every time we get an `recv()` error. Of course the kernel might decide whatever it wants but since we treat this as a hint rather the true PMTU then I don't see any problems with that.

src/device/mod.rs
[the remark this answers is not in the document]
matszczygiel: Oh, I'm sorry. It was created here but probably was lost due to constant rebases. I'll fix this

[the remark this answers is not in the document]
matszczygiel: Fixed it. It was simply a leftover because the runtime is created on device creation.

crates/telio-pmtu/src/linux.rs
matszczygiel: I've added some comments clarifying

crates/telio-pmtu/src/lib.rs
matszczygiel: Done

crates/telio-pmtu/src/lib.rs
matszczygiel: Done

crates/telio-pmtu/src/lib.rs
matszczygiel: Done

crates/telio-pmtu/src/linux.rs
matszczygiel: Done

crates/telio-pmtu/src/linux.rs
matszczygiel: I did just that

crates/telio-lana/src/event_log_file.rs
```diff
@@ -105,6 +105,20 @@ pub mod moose {
         pub config: MeshnetappContextApplicationConfig,
     }
 
+    /// Mock of moose::MeshnetappLogLevel.
+    #[derive(Debug, Clone, Deserialize, PartialEq, Serialize)]
+    /// value of 'log_level'
+    pub enum LibtelioappLogLevel {
```
mathiaspeters, line 111: Might be cleaner to apply `#[serde(rename_all(serialize = "lowercase"))]` to the struct rather than manually renaming each variant

matszczygiel: Done

crates/telio-pmtu/src/linux.rs
```diff
@@ -0,0 +1,387 @@
...
+                PmtuResult::Ok => {
+                    pmtu_range.start = pmtu;
+                    (pmtu_range.start + pmtu_range.end) / 2
+                }
+                PmtuResult::Dropped => {
+                    pmtu_range.end = pmtu;
+                    (pmtu_range.start + pmtu_range.end) / 2
+                }
+                PmtuResult::FragNeeded => {
+                    // Read the kernel PMTU estimate, since the kernel has the knowledge of error ICMP packet.
```
Jauler, line 75: :thinking: Maybe it is worth considering totreat `PmtuResult::FragNeeded` as simply `Pmtu:Result::Dropped`? 

It seems like here the only benefit we gain from using ICMP next_hop MTU field is a slightly faster search of the PMTU. But on the other hand, non-cooperating node (e.g. non-complient by bug or malicious) has ability to cause this loop to spin infinitely. Which seems like not worth in terms of only few packets saved per hour :thinking:

matszczygiel: I do agree that we can't fully trust the kernel. But its hints are beneficial for the speed of PMTU discovery.
So I see a different solution. Let's do the kernel estimate but break the loop after reaching some number of turns. In that way, we get the best of both worlds

matszczygiel: I've added the cap of the number of iterations

Jauler: :heavy_check_mark:

Jauler (approved):
+1.0

PR #346, opened 2024-02-07
[LLT-4855] Create connectivity events aggregation data model

stalowyjez:
### Problem
We need to prepare some API for NAT-traversal and Relay connections data gathering, which supports:
 - collecting events about:
   - changes in the connectivity to Relay servers
   - current state of the connections with the other Meshnet peers
 - processing the events of both types and represent the gathered information as ready-to-serialize format, as it was described in the corresponding RFC.

### Solution
In this PR I introduce an aggregator struct, which allows to store Relay (`change_relay_state` method) and NAT-traversal connection events (`change_peer_state_relayed` for relayed peers and `change_peer_state_direct` for the peers connected directly) and provides an API for `telio-nurse` module, so it can just download almost-serialized data and then effortlessly send it to the analytics server.

For the events collection we reuse the `AnalyticsEvents` struct to reduce the implementation overhead in the rest of the codebase.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -106,19 +106,19 @@ pub struct FeatureNurse {
     pub enable_nat_type_collection: Option<bool>,
     /// Enable/disable Relay connection data
     #[serde(default = "FeatureNurse::get_default_enable_relay_conn_data")]
-    pub enable_relay_conn_data: bool,
+    pub enable_relay_conn_data: Option<bool>,
     /// Enable/disable NAT-traversal connections data
     #[serde(default = "FeatureNurse::get_default_enable_nat_traversal_conn_data")]
-    pub enable_nat_traversal_conn_data: bool,
+    pub enable_nat_traversal_conn_data: Option<bool>,
 }
 
 impl FeatureNurse {
-    fn get_default_enable_relay_conn_data() -> bool {
-        true
+    fn get_default_enable_relay_conn_data() -> Option<bool> {
+        Some(true)
     }
 
-    fn get_default_enable_nat_traversal_conn_data() -> bool {
-        true
+    fn get_default_enable_nat_traversal_conn_data() -> Option<bool> {
+        Some(true)
     }
 }
 
@@ -555,8 +555,8 @@ mod tests {
             initial_heartbeat_interval: None,
             qos: None,
             enable_nat_type_collection: None,
-            enable_relay_conn_data: true,
-            enable_nat_traversal_conn_data: true,
+            enable_relay_conn_data: Some(true),
+            enable_nat_traversal_conn_data: Some(true),
         }),
         lana: Some(FeatureLana {
             event_path: "path/to/some/event/data".to_string(),
@@ -613,8 +613,8 @@ mod tests {
             initial_heartbeat_interval: None,
             qos: None,
             enable_nat_type_collection: None,
-            enable_relay_conn_data: true,
-            enable_nat_traversal_conn_data: true,
+            enable_relay_conn_data: Some(true),
+            enable_nat_traversal_conn_data: Some(true),
         }),
         lana: Some(FeatureLana {
             event_path: "path/to/some/event/data".to_string(),
@@ -809,8 +809,8 @@ mod tests {
                     buckets: Some(5),
                 }),
                 enable_nat_type_collection: None,
-                enable_relay_conn_data: true,
-                enable_nat_traversal_conn_data: true,
+                enable_relay_conn_data: Some(true),
+                enable_nat_traversal_conn_data: Some(true),
             }),
             lana: None,
             paths: None,
@@ -843,8 +843,8 @@ mod tests {
                     buckets: None,
                 }),
                 enable_nat_type_collection: None,
-                enable_relay_conn_data: true,
-                enable_nat_traversal_conn_data: true,
+                enable_relay_conn_data: Some(true),
+                enable_nat_traversal_conn_data: Some(true),
             }),
             lana: None,
             paths: None,
@@ -872,8 +872,8 @@ mod tests {
                 initial_heartbeat_interval: None,
                 qos: None,
                 enable_nat_type_collection: None,
-                enable_relay_conn_data: false,
-                enable_nat_traversal_conn_data: false,
+                enable_relay_conn_data: Some(false),
+                enable_nat_traversal_conn_data: Some(false),
             }),
             lana: None,
             paths: None,
--- a/crates/telio-model/src/api_config.rs
+++ b/crates/telio-model/src/api_config.rs
@@ -104,6 +104,22 @@ pub struct FeatureNurse {
     pub qos: Option<FeatureQoS>,
     /// Enable/disable collecting nat type
     pub enable_nat_type_collection: Option<bool>,
+    /// Enable/disable Relay connection data
+    #[serde(default = "FeatureNurse::get_default_enable_relay_conn_data")]
+    pub enable_relay_conn_data: bool,
+    /// Enable/disable NAT-traversal connections data
+    #[serde(default = "FeatureNurse::get_default_enable_nat_traversal_conn_data")]
+    pub enable_nat_traversal_conn_data: bool,
+}
+
+impl FeatureNurse {
+    fn get_default_enable_relay_conn_data() -> bool {
+        true
+    }
+
+    fn get_default_enable_nat_traversal_conn_data() -> bool {
+        true
+    }
 }
 
 /// Configurable features for Lana module
@@ -539,6 +555,8 @@ mod tests {
             initial_heartbeat_interval: None,
             qos: None,
             enable_nat_type_collection: None,
+            enable_relay_conn_data: true,
+            enable_nat_traversal_conn_data: true,
         }),
         lana: Some(FeatureLana {
             event_path: "path/to/some/event/data".to_string(),
@@ -595,6 +613,8 @@ mod tests {
             initial_heartbeat_interval: None,
             qos: None,
             enable_nat_type_collection: None,
+            enable_relay_conn_data: true,
+            enable_nat_traversal_conn_data: true,
         }),
         lana: Some(FeatureLana {
             event_path: "path/to/some/event/data".to_string(),
@@ -770,7 +790,9 @@ mod tests {
         let no_qos_json = r#"
         {
             "nurse": {
-                "fingerprint": "fingerprint_test"
+                "fingerprint": "fingerprint_test",
+                "enable_relay_conn_data": false,
+                "enable_nat_traversal_conn_data": false
             }
         }"#;
 
@@ -787,6 +809,8 @@ mod tests {
                     buckets: Some(5),
                 }),
                 enable_nat_type_collection: None,
+                enable_relay_conn_data: true,
+                enable_nat_traversal_conn_data: true,
             }),
             lana: None,
             paths: None,
@@ -819,6 +843,8 @@ mod tests {
                     buckets: None,
                 }),
                 enable_nat_type_collection: None,
+                enable_relay_conn_data: true,
+                enable_nat_traversal_conn_data: true,
             }),
             lana: None,
             paths: None,
@@ -846,6 +872,8 @@ mod tests {
                 initial_heartbeat_interval: None,
                 qos: None,
                 enable_nat_type_collection: None,
+                enable_relay_conn_data: false,
+                enable_nat_traversal_conn_data: false,
             }),
             lana: None,
             paths: None,
--- a/crates/telio-nurse/src/aggregator.rs
+++ b/crates/telio-nurse/src/aggregator.rs
@@ -1,8 +1,6 @@
 use std::{borrow::Borrow, collections::hash_map::Entry, mem, time::Duration};
 
-use tokio::time::Instant;
-
-use parking_lot::RwLock;
+use tokio::{sync::Mutex, time::Instant};
 
 use serde::{ser::SerializeTuple, Serialize};
 
@@ -191,7 +189,7 @@ struct AggregatorData {
 // A container to store the connectivity data for our meshnet peers
 #[allow(dead_code)]
 pub struct ConnectivityDataAggregator<W: WireGuard> {
-    data: RwLock<AggregatorData>,
+    data: Mutex<AggregatorData>,
 
     aggregate_relay_events: bool,
     aggregate_nat_traversal_events: bool,
@@ -204,7 +202,7 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
     #[allow(dead_code)]
     pub fn new(nurse_features: &FeatureNurse, wg_interface: W) -> ConnectivityDataAggregator<W> {
         ConnectivityDataAggregator {
-            data: parking_lot::RwLock::new(AggregatorData {
+            data: Mutex::new(AggregatorData {
                 current_relay_event: None,
                 current_peer_events: HashMap::new(),
                 unacknowledged_segments: Vec::new(),
@@ -216,7 +214,7 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
     }
 
     #[allow(dead_code)]
-    pub fn change_peer_state_direct(
+    pub async fn change_peer_state_direct(
         &mut self,
         event: AnalyticsEvent,
         initiator_ep: EndpointProvider,
@@ -229,10 +227,11 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
                 responder_ep: responder_ep.into(),
             },
         )
+        .await
     }
 
     #[allow(dead_code)]
-    pub fn change_peer_state_relayed(&mut self, event: AnalyticsEvent) {
+    pub async fn change_peer_state_relayed(&mut self, event: AnalyticsEvent) {
         self.change_peer_state_common(
             event,
             PeerEndpointTypes {
@@ -240,14 +239,19 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
                 responder_ep: EndpointType::Relay,
             },
         )
+        .await
     }
 
-    fn change_peer_state_common(&mut self, event: AnalyticsEvent, endpoints: PeerEndpointTypes) {
+    async fn change_peer_state_common(
+        &mut self,
+        event: AnalyticsEvent,
+        endpoints: PeerEndpointTypes,
+    ) {
         if !self.aggregate_nat_traversal_events {
             return;
         }
 
-        let mut data_guard = self.data.write();
+        let mut data_guard = self.data.lock().await;
         let event_entry = data_guard.current_peer_events.entry(event.public_key);
         let new_segment = match event_entry.borrow() {
             Entry::Occupied(entry)
@@ -276,7 +280,7 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
     }
 
     #[allow(dead_code)]
-    pub fn change_relay_state(
+    pub async fn change_relay_state(
         &mut self,
         event: AnalyticsEvent,
         reason: RelayConnectionChangeReason,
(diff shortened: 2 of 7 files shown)


crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,113 @@
...
+    Relayed,
+    Local(Option<EndpointProvider>),
+    UPnP(Option<EndpointProvider>),
+    UDPHolePunch(Option<EndpointProvider>),
+}
+
+// Possible disconnection reasons
+#[derive(Copy, Clone, Debug)]
+pub enum DisconnectionReason {
+    Initial,
```
Jauler, line 23: Is `initial` meant to be used in that startup phase, where we do not have connection yet. If that is indeed the purpose, maybe it is worth skipping the report for it at all?

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,113 @@
...
+    DisabledByUser,
+    NetworkError,
```
Jauler, line 25: Maybe it would be nice to split `NetworkError` at least in a few subtypes:
* Forced connection termination by the server
* Timeouts
* Some OS errors (like no route to host or similar)

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,113 @@
...
+        if let Entry::Vacant(e) = storage_guard.entry((event.timestamp, event.peer)) {
+            e.insert(event.info);
+        } else {
+            telio_log_warn!("Discarding connectivity info - duplicate connectivity event");
+        }
+    }
+}
+
+impl ConnectivityDataStorageTrait for ConnectivityDataAggregator {
+    fn get_unacknowledged_states(&self) -> Vec<ConnectivityChangeEvent> {
```
Jauler, line 89: IMHO, in my opinion, it would be nice to handle the "unfinished states" in the aggregator. What I mean by that is:
* `telio-traversal` and `telio-relay` sends raw "started/finished" events
* Aggregator gets them and stores them
* Whenever heartbeat occurs, aggregator finds all of the events, converts them from two separate events (start/finish) into a span of the state (e.g. what state, how long it continued, endpoint provider types and/or reasons for state changes etc.)
* If there is a start event, but no finish event for longer then 24h - aggregator converts it into a "finished" state of 24h and new one started.

crates/telio-nurse/src/aggregator.rs
Jauler: Another option would be to simply include `errno` value :thinking:

stalowyjez: I guess it would be nicer to have it as an enum here? I mean, having an errno value forces us to maintain another mapping

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
+use std::{collections::HashSet, time::Duration};
+
+use tokio::time::Instant;
+
+use parking_lot::RwLock;
+use serde::{ser::SerializeTuple, Serialize, Serializer};
```
tomaszklak, line 6: Serialization is not used or tested - is it really needed? If it will be needed in future, please add some tests for it. Especially for the custom implementation.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+        ConnectivityDataAggregator::new(&nurse_features, wg_interface)
+    }
+
+    fn create_basic_event(n: u8) -> AnalyticsEvent {
+        AnalyticsEvent {
+            public_key: PublicKey(*b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"),
+            endpoint: DualTarget::new((Some(Ipv4Addr::from([n, n, n, n])), None)).unwrap(), // Whatever
+            tx_bytes: 0, // We don't need them for Relay event
+            rx_bytes: 0,
+            peer_state: PeerState::Connected, // This should be the same as in the
```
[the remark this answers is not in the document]
stalowyjez: Uh, some leftover, I'll just remove it

tomaszklak:
I think both PR description and commits lack some high level description. As it is right now, without referring to LLT-4855 it is 'impossible' to understand the purpose of this code.

crates/telio-nurse/src/aggregator.rs
stalowyjez: Riight, I think I will remove it for now

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+            EndpointProvider::Stun => EndpointType::Stun,
+            EndpointProvider::Upnp => EndpointType::UPnP,
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, Serialize)]
+pub enum RelayConnectionState {
+    Connecting = 16,
+    Connected = 17,
```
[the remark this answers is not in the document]
stalowyjez: I agree, but I think that they should be the same as in the RFC so, we should change them there first.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+
+        self.unacknowledged_segments
+            .read()
+            .clone()
+            .into_iter()
+            .collect()
+    }
+
+    #[allow(dead_code)]
+    pub fn acknowledge_segments_transmission<T>(&mut self, segment_set: T)
```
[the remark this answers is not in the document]
stalowyjez: I think making it a part of that would be more intuitive - this is how it was presented in the task description and I guess the purpose of that was to be able to get them again when we are not able to send them due to some connection issue without storing them somewhere outside.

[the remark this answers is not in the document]
stalowyjez: @Jauler what do you think about making it a one method?

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+
+// Possible disconnection reasons
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, Serialize)]
+#[allow(dead_code)]
+pub enum RelayConnectionChangeReason {
+    // 1XX reason numbers for configuration changes
+    ConfigurationChange = 101,
+    DisabledByUser = 102,
+
+    // 2XX reason numbers for network problems
```
Jauler, line 73: I wonder, maybe it is worth incorporating some sort of OS error (e.g. errno) when `NetworkError` occurs? :thinking:

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+}
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+#[allow(dead_code)]
+pub enum ConnectionData {
+    Relay(RelayConnectionData),
+    Peer(PeerConnectionData),
+}
+
+impl Serialize for ConnectionData {
```
Jauler, line 102: I really like this serialization way :+1:

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+            return;
+        }
+
+        match self.current_events.write().entry(event.public_key) {
+            std::collections::hash_map::Entry::Occupied(mut entry) => {
+                let (old_event, old_state) = entry.get();
+
+                let segment_connection_data = match old_state {
+                    ConnectionState::Relay(reason) => {
+                        if old_event.peer_state == event.peer_state {
```
Jauler, line 252: Note that server_id may change which would substitute a new state.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+                    }
+                };
+
+                // Just create a new segment here
+                self.unacknowledged_segments
+                    .write()
+                    .insert(ConnectionSegmentData {
+                        node: event.public_key,
+                        start: old_event.timestamp.into(),
+                        length: event.timestamp - old_event.timestamp,
```
Jauler, line 218: :thinking: This would produce invalid length in case of device sleeping, as we would simple stretch the state during sleep, while in reality, we should consider it as being offline. But it is rather hard to make it "correct". In order to avoid next task blocking, this sleeping behavior may be solved in different task/PR.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+                    > Duration::from_secs(60 * 60 * 24)
+                {
+                    match peer_state {
+                        ConnectionState::Relay(reason) => {
+                            Some(ConnectionData::Relay(RelayConnectionData {
+                                state: peer_event.peer_state.into(),
+                                reason: *reason,
+                            }))
+                        }
+                        ConnectionState::Peer(initiator_ep_type, reciever_ep_type) => wg_peers
```
Jauler, line 301:
```suggestion
                        ConnectionState::Peer(initiator_ep_type, receiver_ep_type) => wg_peers
```

stalowyjez: Yeah, this should be `responder` anyway

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Jauler: Yeah, during the initial discussions we have somehow separated collection and acknowledgement (e.g. we would like to acknowledge only after we have successfully submitted events to libmoose), but in practice, this part does not require such strict consistency, and it _can_ be done in the `collect_unacknowledged_segments()` as well.

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Jauler: Initially those numbers did not have any sort of ordering or so, therefore in such a setting, there is no difference what exact number is selected and even if additional endpoint types are added as higher IDs then relaying ones.

But after looking at the code, I see that there is some benefits to be had if we can concatenate endpoint types bitwise for initiator and responder. Therefore... What about reserving 4 bits for each `EndpointType`, this gives 16 options, which seems plenty for this kind of thing, and thus 8 bits would be used to encode endpoint types for initiator and responder individually. If @tomaszklak and @stalowyjez agree, I can update RFC with this information.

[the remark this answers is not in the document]
tomaszklak: 4bits per type sounds good to me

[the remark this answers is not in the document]
stalowyjez: @Jauler sounds good to me

crates/telio-nurse/src/aggregator.rs
Jauler: Ok, lets leave them for future improvements if we will indeed need them.

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Jauler: I updated the RFC. Could you check it out?

Hasan6979 (commented):
Why are there so many Allow dead code macros ?

crates/telio-model/src/api_config.rs
```diff
@@ -104,6 +104,22 @@ pub struct FeatureNurse {
     pub qos: Option<FeatureQoS>,
     /// Enable/disable collecting nat type
     pub enable_nat_type_collection: Option<bool>,
+    /// Enable/disable Relay connection data
+    #[serde(default = "FeatureNurse::get_default_enable_relay_conn_data")]
+    pub enable_relay_conn_data: bool,
```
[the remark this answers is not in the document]
stalowyjez: Why? It's either enabled or disabled, there is no third option...

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,689 @@
...
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, Serialize)]
+pub struct RelayConnectionData {
+    state: RelayConnectionState,
+    reason: RelayConnectionChangeReason,
+}
+
+// Possible disconnection reasons
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, Serialize)]
+#[allow(dead_code)]
+pub enum RelayConnectionChangeReason {
```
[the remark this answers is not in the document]
stalowyjez: It is basically kind of an internal errno, I don't think that splitting it into a few more types would be beneficial in any way

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,808 @@
...
+                public_key: event.public_key,
+                peer_state: event.peer_state,
+                timestamp: event.timestamp.into(),
+                reason,
+            })
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn collect_unacknowledged_segments(&self) -> Vec<ConnectionSegmentData> {
```
tomaszklak, line 313: How is the `ConnectivityDataAggregator` supposed to be used by libtelio? The way I understand this design, there will be one instance of it to which we will push events (both nat and relay) when something changes. At heartbeat time we will call `collect_unacknowledged_segments` and send the result via the 2 new fields `NAT_monitoring` and `derp_monitoring`. If my understanding is correct I see one issue. There are 2 separate fields but we get one collection from `collect_unacknowledged_segments`. Since I don't think we have any use for such a collection, can we return tuple with separate collections for nat and relay events?

Also, I think it would be simlper to debuging to sort those collections based on timestamp before returning them.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,808 @@
...
+    peer_state: PeerState,
+    timestamp: Instant,
+    reason: RelayConnectionChangeReason,
+}
+
+struct AggregatorData {
+    current_relay_event: Option<RelayEvent>,
+    current_peer_events: HashMap<PublicKey, (AnalyticsEvent, PeerEndpointTypes)>,
+    unacknowledged_segments: Vec<ConnectionSegmentData>,
+}
```
tomaszklak, line 189: What will happen with this state when application is going to be shutdown?

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,808 @@
...
+                reason,
+            })
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn collect_unacknowledged_segments(&self) -> Vec<ConnectionSegmentData> {
+        // Getting wg_interface is first, so it won't await while mutex is locked
+        let wg_interface = self.wg_interface.get_interface().await;
+        let mut data_guard = self.data.write();
```
tomaszklak, line 316: Non-async lock used in the async method can block tokio - please change the lock to the one from tokio.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,808 @@
...
+struct AggregatorData {
+    current_relay_event: Option<RelayEvent>,
+    current_peer_events: HashMap<PublicKey, (AnalyticsEvent, PeerEndpointTypes)>,
+    unacknowledged_segments: Vec<ConnectionSegmentData>,
+}
+
+// A container to store the connectivity data for our meshnet peers
+#[allow(dead_code)]
+pub struct ConnectivityDataAggregator<W: WireGuard> {
+    data: RwLock<AggregatorData>,
```
tomaszklak, line 194: As far as I can see, this is always locked in write mode - so it's simpler and cheaper to use mutex.

crates/telio-nurse/src/aggregator.rs
Jauler: After a little discussion, we have decided to offload handling of sleeping behavior into separate task

crates/telio-model/src/api_config.rs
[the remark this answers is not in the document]
Hasan6979: When I had added `enable_nat_type_collection` which was also an enable/disable, there was a problem when sending telio config (I don't recall the exact problem tbh) but because of which I had changed it to option<bool>.

crates/telio-nurse/src/aggregator.rs
stalowyjez: @Jauler what do you think about this one? It looks to me that while it should be handled in a bit different manner, it is kinda similar to the sleeping one 🤔

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Hasan6979: In future as the number of errors might grow, I just feel it would be better readability if there is some separation.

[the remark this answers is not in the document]
stalowyjez: @Jauler what do you think?

crates/telio-model/src/api_config.rs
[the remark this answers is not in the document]
stalowyjez: I feel bad for this kind of solution, because usually `Some(false) == None` which is kinda weird, but I can do it if that is needed.

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Jauler: We were thinking a little bit about representing errors in more detailed manner, but decided to make it a future improvement, to keep scope in control.

crates/telio-nurse/src/aggregator.rs
Jauler: There will be a responsibility of telio-nurse, to generate "meshnet disable" event, and collect those events from the aggregator.

Jauler (approved):
+1.0

Jauler (approved):
+1.0

tomaszklak (dismissed):
+1, remember to clean up the commits :)

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,999 @@
...
+                current_timestamp.duration_since(relay_event.timestamp) > DAY_DURATION
+                    || force_close
+            })
+        {
+            new_relay_segments.push(ConnectionSegmentData::new_relay(event, current_timestamp));
+            event.timestamp = current_timestamp;
+        }
+
+        if force_close {
+            data_guard.current_relay_event.take();
```
tomaszklak, line 353: nit: I think `clear` which you use later on is more clear.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,999 @@
...
+            }
+        } else {
+            telio_log_warn!("Getting wireguard interfaces failed");
+        }
+
+        if force_close {
+            data_guard.current_peer_events.clear();
+        }
+
+        println!("Newo peer segments: {:?}", new_peer_segments);
```
tomaszklak, line 400: please remove :)

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,999 @@
...
+        {
+            new_relay_segments.push(ConnectionSegmentData::new_relay(event, current_timestamp));
+            event.timestamp = current_timestamp;
+        }
+
+        if force_close {
+            data_guard.current_relay_event.take();
+        }
+
+        println!("Newo relay segments: {:?}", new_relay_segments);
```
tomaszklak, line 356: Remove :)

crates/telio-nurse/src/aggregator.rs
```diff
@@ -0,0 +1,999 @@
...
+        &self,
+        force_close: bool,
+    ) -> AggregatorCollectedSegments {
+        // Getting wg_interface is first, so it won't await while mutex is locked
+        let wg_interface = self.wg_interface.get_interface().await;
+        let mut data_guard = self.data.lock().await;
+
+        let current_timestamp = Instant::now();
+
+        println!("Relay event: {:?}", data_guard.current_relay_event);
```
tomaszklak, line 336: Remove :)

tomaszklak (approved):

crates/telio-nurse/src/aggregator.rs
stalowyjez: unfortunately in case of option it seems that `take` is everything we have

tomaszklak (approved):

PR #347, opened 2024-02-08
LLT-4870 bring back post quantum

matszczygiel:
It turned out that the problem was not with the PQ implementation but with the usage of sccahce in CI/CD which breaks `libtelio.a` in case it's built with a native C static library.

### Problem
The `libteliio.a` static lib failed to link with with executable on Mac and Linux. The problem is solved by now in the build repo.

### Solution
Revert the revert

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (approved):

packgron (approved):
+0.5

LukasPukenis:
+1.5

PR #350, opened 2024-02-08
Add configurable TTL value for DNS

ThrasherLT:
### Problem
The current TTL value is too high for use with nicknames, but it would be more convenient to have it as a configurable feature.

### Solution
Made TTL value configurable

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-dns/src/zone.rs
```diff
@@ -48,17 +49,19 @@ impl AuthoritativeZone {
...
+                .set_ttl(ttl_value)
                 .set_rr_type(RecordType::SOA)
                 .set_dns_class(DNSClass::IN)
                 .set_data(Some(RData::SOA(SOA::new(
                     Name::parse("mesh.nordsec.com.", None)?,
                     Name::parse("support.nordsec.com.", None)?,
                     2015082403,
                     7200,
-                    3600,
+                    ttl_value.try_into().map_err(|e| {
```
packgron, line 60: Just pass a default and log a warn.

packgron (changes requested):
We also need change in telio_features.py *pain*

packgron (approved):
+2.0

PR #351, opened 2024-02-09
Nordlynx support in NatLab

LukasPukenis:
### Problem
Current VPN tests only targeted raw WG server. Our infrastructure uses nordlynx

### Solution
Introduce `nlx` docker image and boxes in docker-compose where they run:
- radius server
- fakefm server(authenticates all users)

Docker image is separated from main libtelio image due to being too time consuming to be iterated on.

There's also updated Vagrant box with dkms support out there.

- [x] README.md is updated
- [x] changelog.md is updated


nat-lab/tests/utils/connection_util.py
```diff
@@ -323,6 +329,26 @@ def generate_connection_tracker_config(
...
     lan_addr = LAN_ADDR_MAP[connection_tag]
     ctc_list = [
+        ConnectionTrackerConfig(
+            "nlx_1",
+            nlx_1_limits,
+            FiveTuple(
+                protocol="udp",
+                src_ip=lan_addr,
+                dst_ip=str(config.NLX_SERVER.get("ipv4")),
+                dst_port=51820,
```
stalowyjez, line 339: Is it really "five"?

nat-lab/tests/test_vpn.py
```diff
@@ -228,7 +458,7 @@ async def test_vpn_connection(
         ),
     ],
 )
-async def test_vpn_reconnect(
+async def test_vpn_reconnect_wireguard(
```
stalowyjez, line 461: Couldn't that be parameterized with nordlynx and wireguard servers?

nat-lab/tests/mesh_api.py
```diff
@@ -353,12 +354,54 @@ def generate_peer_config(node: Node, allowed_ips: str) -> str:
...
-            f' "{wg_conf}" > /etc/wireguard/wg0.conf; wg-quick down'
-            " /etc/wireguard/wg0.conf; wg-quick up /etc/wireguard/wg0.conf'"
-        )
-        os.system(full_command)
+            def exec_docker_commands(
+                container_id: str, commands: typing.List[str]
+            ) -> None:
+                for command in commands:
+                    cmd = f"docker exec -d --privileged {container_id} bash -c '{command}'"
+                    os.system(cmd)
```
stalowyjez, line 362: Does it make any sense to add this as a function instead just moving this loop to the proper `if` branch? (Because it looks that it is used only once?)

nat-lab/tests/mesh_api.py
```diff
@@ -353,12 +354,54 @@ def generate_peer_config(node: Node, allowed_ips: str) -> str:
...
+                        "iptables -A FORWARD -i nordlynx0 -j ACCEPT; iptables -A"
+                        " FORWARD -o nordlynx0 -j ACCEPT; iptables -t nat -A"
+                        " POSTROUTING -o eth0 -j MASQUERADE"
+                    ),
+                    # courtesy sleep so it would be really all up and running
+                    "sleep 2",
+                    "ip addr s > /var/log/ipaddr.show",
+                    "nlx > /var/log/nlx.log",
+                    "/usr/bin/fakefm/fakefm -s testing123 2>&1 > /var/log/fakefm.log",
+                    "/opt/nlx-radius/nlx-radius --log=stdout --verb=DEBUG 2>&1 > /var/log/nordlynx.log",
```
stalowyjez, line 393: I would consider moving this commands to some script

stalowyjez (commented):
This `a.out` file looks a bit like some leftover. Besides that, a bunch of suggestions, overall looks ok.

nat-lab/tests/mesh_api.py
gytsto: :100: agree with @stalowyjez

nat-lab/tests/test_vpn.py
```diff
@@ -35,6 +35,22 @@ async def _connect_vpn(
     assert ip == wg_server["ipv4"], f"wrong public IP when connected to VPN {ip}"
 
 
+async def _connect_vpn_nordlynx(
```
gytsto, line 38: is there actually a need for this function?

nat-lab/tests/test_vpn.py
```diff
@@ -144,6 +158,222 @@ async def test_vpn_connection(
...
+                    stun_limits=ConnectionLimits(1, 1),
+                ),
+                is_meshnet=False,
+            ),
+            "10.0.254.7",
+            marks=pytest.mark.mac,
+        ),
+    ],
+)
+async def test_vpn_connection_nordlynx(
```
gytsto, line 240: can be easily parametrized in original `test_vpn_connection`

nat-lab/tests/test_vpn.py
gytsto: :100: agree with @stalowyjez

nat-lab/tests/utils/connection_util.py
gytsto: :question:

nat-lab/tests/test_vpn.py
matszczygiel: Agree

nat-lab/tests/utils/connection_util.py
LukasPukenis: It's 4 elements, but it was modelled after the VPN setup which uses the same FiveTuple for 4 elements. Didn't want to touch that yet

stalowyjez: OK, then

nat-lab/tests/test_vpn.py
LukasPukenis: This is a good suggestion, after some back and worth I have it working and it's much nicer than it was. Thanks for the suggestion :)

nat-lab/tests/mesh_api.py
LukasPukenis: So it's tricky. I've tried doing that however nordlynx is finicky for order of setup so it would need to be like 2 separate parts in the separate shell script and some steps inside of the python. While it would make it a bit simpler I think debugging it would become harder. I could at the very least move it to a separate function so not to pollute this one as much but then this whole `setup_vpn_servers` function would become just a simple wrapper around it

nat-lab/tests/test_vpn.py
LukasPukenis: So basically it was duplicated because existing tests pinged the client from the server which I am not sure how to configure so it would work on nordlynx server. So I duplicated the function and removed one ping instead. Now as I parametrizes everything there's no need for this function anymore I suppose but as a simplification I also removed the offending ping from the code

nat-lab/tests/test_vpn.py
LukasPukenis: done

nat-lab/tests/mesh_api.py
LukasPukenis: it's not needed thus i inlined it

nat-lab/tests/mesh_api.py
stalowyjez: Hmm, I don't get this tbh, here you have just a list of commands which you run one by one with `docker exec` - why couldn't you run just `docker exec -d --privileged server_container /opt/bin/some_script.sh` instead?

nat-lab/tests/test_vpn.py
```diff
@@ -15,27 +15,21 @@
...
 ) -> None:
     await client.connect_to_vpn(
         wg_server["ipv4"], wg_server["port"], wg_server["public_key"]
     )
 
-    async with Ping(connection, config.PHOTO_ALBUM_IP).run() as ping:
+    async with Ping(client_conn, config.PHOTO_ALBUM_IP).run() as ping:
         await testing.wait_long(ping.wait_for_next_ping())
 
-    async with Ping(vpn_connection, client_meshnet_ip).run() as ping:
```
gytsto, line 31: this ping has to stay here, it was added for a reason. See https://github.com/NordSecurity/libtelio/pull/76

nat-lab/tests/mesh_api.py
LukasPukenis: Actually it might be as easy as you describe. I was a bit misguided, let me try :D

LukasPukenis: @stalowyjez @gytsto so it worked pretty nicely. I moved the whole script to a separate shellscript referred in docker-compose and now python code only has private key output to a file and `nlx set ... private-key` command

nat-lab/tests/test_vpn.py
LukasPukenis: I've restored it and placed it as an optional parameter so nordlynx servers wouldn't be pinged. I didn't investigate further just talked with @Jauler some time ago about it and he mentioned that we can skip the pinging so doing exactly that just for my introduced server.

dfetti (commented):
+1.0

nat-lab/bin/nordlynx
```diff
@@ -0,0 +1,26 @@
...
+
+# courtesy sleep so it would be really all up and running
+sleep 2
+
+ip addr s > /var/log/ipaddr.show
+nlx > /var/log/nlx.log
+/usr/bin/fakefm/fakefm -s testing123 2>&1 > /var/log/fakefm.log &
+/opt/nlx-radius/nlx-radius --log=stdout --verb=DEBUG 2>&1 > /var/log/nordlynx.log
+
+sleep infinity
```
stalowyjez, line 26: Oh yeah

nat-lab/tests/test_vpn.py
```diff
@@ -105,43 +123,60 @@ async def _connect_vpn(
...
         ),
     ],
 )
+@pytest.mark.parametrize(
+    "vpn_conf",
+    [
+        pytest.param([
+            VpnConfig(config.WG_SERVER, ConnectionTag.DOCKER_VPN_1, True),
+            VpnConfig(config.NLX_SERVER, ConnectionTag.DOCKER_NLX_1, False),
+        ])
```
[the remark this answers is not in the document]
LukasPukenis: I gave it a go, it works, it's now separated which I like even more but I don't like what the testcase code has became. Though it was fun.

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -40,6 +40,9 @@ class ConnectionTrackerConfig:
     limits: ConnectionLimits
     target: FiveTuple
 
+    def get_key(self) -> str:
```
gytsto, line 43: what's the purpose of this method, when you can simple read key?

nat-lab/tests/test_vpn.py
```diff
@@ -104,43 +117,75 @@ async def _connect_vpn(
...
         env = await exit_stack.enter_async_context(
             setup_environment(exit_stack, [alpha_setup_params])
         )
 
         alpha, *_ = env.nodes
-        connection, *_ = [conn.connection for conn in env.connections]
+        client_conn, *_ = [conn.connection for conn in env.connections]
         client_alpha, *_ = env.clients
 
         vpn_connection, *_ = await setup_connections(
```
gytsto, line 174: not needed when test is being run with `nlx`, also could `DOCKER_VPN_1` should be replaced with `vpn_conf.conn_tag`

nat-lab/tests/test_vpn.py
```diff
@@ -104,43 +117,75 @@ async def _connect_vpn(
...
             exit_stack, [ConnectionTag.DOCKER_VPN_1]
         )
 
-        ip = await testing.wait_long(stun.get(connection, config.STUN_SERVER))
+        ip = await testing.wait_long(stun.get(client_conn, config.STUN_SERVER))
         assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
 
+        await setup_connections(exit_stack, [vpn_conf.conn_tag])
```
gytsto, line 181: this is not needed at all

nat-lab/tests/test_vpn.py
gytsto: I'd suggest doing:
```
vpn_connection_mngr = None
if vpn_conf.should_ping:
    vpn_connection_mngr, *_ = await setup_connections(exit_stack, [vpn_conf.conn_tag])
```

nat-lab/tests/test_vpn.py
```diff
@@ -104,43 +117,75 @@ async def _connect_vpn(
...
+    public_ip: str,
 ) -> None:
+    def find_cfg_by_key(
+        params: SetupParameters, key: str
+    ) -> Optional[ConnectionTrackerConfig]:
+        if params is not None and params.connection_tracker_config is not None:
+            for cfg in params.connection_tracker_config:
+                if cfg.get_key() == key:
+                    return cfg
+        return None
```
gytsto, line 148: `params` is not `Optional` type, should never be `None` afaik

nat-lab/tests/test_vpn.py
```diff
@@ -104,43 +117,75 @@ async def _connect_vpn(
...
+
     async with AsyncExitStack() as exit_stack:
+        alpha_setup_params = copy.deepcopy(alpha_setup_params)
+
+        if vpn_conf.conn_tag == ConnectionTag.DOCKER_VPN_1:
+            cfg = find_cfg_by_key(alpha_setup_params, "vpn_1")
+            assert cfg is not None
+            cfg.limits = ConnectionLimits(1, 1)
```
gytsto, line 156: does it actually write to `alpha_setup_params`

nat-lab/tests/test_vpn.py
LukasPukenis: good catch, slipped through my eyes

nat-lab/tests/test_vpn.py
gytsto: damn I pointed to wrong line...

nat-lab/tests/test_vpn.py
```diff
@@ -104,43 +117,73 @@ async def _connect_vpn(
...
 async def test_vpn_connection(
-    alpha_setup_params: SetupParameters, public_ip: str
+    alpha_setup_params: SetupParameters,
+    vpn_conf: VpnConfig,
+    public_ip: str,
 ) -> None:
+    def find_cfg_by_key(
+        params: SetupParameters, key: str
+    ) -> Optional[ConnectionTrackerConfig]:
+        if params is not None and params.connection_tracker_config is not None:
```
gytsto, line 144: params is not Optional type, should never be None afaik

nat-lab/tests/utils/connection_tracker.py
LukasPukenis: I am paranoid reading the values directly, I think it's too fragile due to type safety, though I cannot exactly tell if having a method is stronger in that regard, having a function also allows for finding call sites easily which I prefer a lot with dynamic typing

nat-lab/tests/test_vpn.py
LukasPukenis: Yep, it actually overwrites the original, that's why there's deepcopy above it. I caught it redhanded when it was failing with both tests but succeeded with each one separately

nat-lab/tests/test_vpn.py
```diff
@@ -104,43 +117,75 @@ async def _connect_vpn(
...
+            for cfg in params.connection_tracker_config:
+                if cfg.get_key() == key:
+                    return cfg
+        return None
+
     async with AsyncExitStack() as exit_stack:
+        alpha_setup_params = copy.deepcopy(alpha_setup_params)
```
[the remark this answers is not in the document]
LukasPukenis: same as above

nat-lab/tests/test_vpn.py
LukasPukenis: Good idea, I did that, it's better now

nat-lab/tests/test_vpn.py
```diff
@@ -104,44 +118,88 @@ async def _connect_vpn(
...
         assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
 
-        await _connect_vpn(
-            connection,
-            vpn_connection.connection,
-            client_alpha,
-            alpha.ip_addresses[0],
-            config.WG_SERVER,
-        )
+        if vpn_conf.should_ping_client:
```
gytsto, line 184: - bad to hardcode `DOCKER_VPN_1`, use `vpn_conf.conn_tag` instead
- duplicate `_connect_vpn` can be reduced

suggestion:
```
vpn_connection = None
if vpn_conf.should_ping_client:
    conn_mngr, *_ = await setup_connections(
        exit_stack, [vpn_conf.conn_tag]
    )
    vpn_connection = conn_mngr.connection

await _connect_vpn(
    client_conn,
    vpn_connection,
    client_alpha,
    alpha.ip_addresses[0],
    vpn_conf.server_conf,
)
```

nat-lab/tests/utils/connection_tracker.py
gytsto: I guess it's just a personal preference. :+1:

nat-lab/tests/test_vpn.py
LukasPukenis: Removed the hardcoded part. Leaving the duplicate as I find it cleaner and more explicit

nat-lab/tests/utils/connection_tracker.py
LukasPukenis: Sort of yeah...

nat-lab/tests/test_vpn.py
gytsto: I hardly disagree with `cleaner` part when duplicating code, but I guess it is personal preference again :laughing:

gytsto (approved):
+2

matszczygiel (approved):

PR #353, opened 2024-02-09
Update moose tracker

dfetti:
Update to the new moose tracker v1.0.0-libtelioApp.
This tracker has a shared context so we don't need to populate it by ourselves using data from a foreign context.

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-lana/src/lib.rs
```diff
@@ -161,16 +156,25 @@ pub fn init_moose(
...
             init_success_tx: tx,
         }),
         Box::new(moose_callbacks::MooseErrorCallback),
-    )
+    )?;
+    moose::set_context_application_version(app_version)?;
+    moose::set_context_application_name(LANA_APP_NAME.to_string())
 }
+
+/// Fetches context string from moose
```
LukasPukenis, line 169: I am slightly confused here:
- why do we need to fetch the context anymore at all
- this `String` I think is JSON, shouldn't it be JSON structure then?

nat-lab/tests/utils/analytics/event_loader.py
```diff
@@ -80,6 +76,11 @@ def fetch_moose_events(database_name):
         for event in events:
             event_json = json.loads(event[0])
             flatten_event_json = flatten(event_json)
+
+            # Skip init event
+            if flatten_event_json["event_name"] == "init":
+                continue
+
```
LukasPukenis, line 83: Is this needed? I assume the new moose tracker emits this event from now on, however the python code is used for the mocked moose code which might not emit this event(i might be wrong just trying to be sure)

crates/telio-nurse/src/nurse.rs
```diff
@@ -267,24 +264,22 @@ impl State {
...
-                        fp
-                    );
-                    Uuid::new_v4()
-                })
-            })
-            .unwrap_or_else(|| {
-                telio_log_trace!(
-                    "Could not find stored meshnet id by moose, generating a fresh one"
+        telio_lana::fetch_context_string(String::from(
+            "contest.application.libtelioapp.config.internal_meshnet.fp",
```
LukasPukenis, line 268: is the new moose tracker working by providing string keys for it? I find it pretty fragile as there's no type safety, we can easily produce libtelio binaries with incorrect strings in here :/

ci/moose_utils.py
```diff
@@ -169,12 +169,12 @@ def unset_cargo_dependencies():
                 )
             _write_file(f"{PROJECT_ROOT}/Cargo.toml", replaced_moose)
 
-    # remove moosemeshnetapp dependency from telio-lana/Cargo.toml
+    # remove mooselibtelioapp dependency from telio-lana/Cargo.toml
```
LukasPukenis, line 172: Maybe you know the reason why they renamed the tracker?

LukasPukenis (commented):
Left some questions regarding the upgrade :)

ci/moose_utils.py
dfetti: No idea

crates/telio-lana/src/lib.rs
dfetti: Because we store meshnet_id in moose context.
And it is a string not a JSON, because this function receives the path through json to meshnet_id string.

crates/telio-nurse/src/nurse.rs
dfetti: As I actually did "contest" :))
The way it works is that you call fetch_context_string and you give it a path and if the path is valid and there is a string, it will return it to you.

nat-lab/tests/utils/analytics/event_loader.py
dfetti: No, nat-lab tests use the real moose tracker.
And yes as far as I saw this event always appear with the new tracker.

LukasPukenis: NatLab using real tracker is very good. Might I offer an improvement for actually testing that this event is present and if not - then raising a fault?

crates/telio-lana/src/lib.rs
LukasPukenis: Where do we store the meshnet ID in moose by the way(and why do we need to re-retrieve it if we already have it)? I want to understand the flow better :) thanks!

dfetti: To be able to store it persistently. If the we restart meshnet we want to use the same meshnet id, not to generate a new one.

nat-lab/tests/utils/analytics/event_loader.py
dfetti: Hmm, could be a good idea

crates/telio-nurse/src/nurse.rs
dfetti: Yes, I think it would be nice. But another update, would break again the compatibility with all the integrations of moose. So I don't think the guys will enjoy to do that.

ci/build_libtelio.py
```diff
@@ -247,6 +247,15 @@ def exec_build(args):
...
         moose_utils.set_cargo_dependencies()
+        # We are using an outdated library in telio-traversal called sm
+        # It uses some old versions of quote, proc-macro2 and syn.
+        # When the moose dependency is injected by the build script it tries to
+        # use these old versions of quote, proc-macro2 and syn. A cargo update
+        # should fix it for now, but it's prone to errors, so we should find an
+        # alternative solution for the sm library.
+        rutils.run_command(["cargo", "update", "-p", "quote@0.6.13"])
+        rutils.run_command(["cargo", "update", "-p", "proc-macro2@0.4.30"])
+        rutils.run_command(["cargo", "update", "-p", "syn@0.15.44"])
```
tomaszklak, line 258: nit: this could be moved to `set_cargo_dependencies`

tomaszklak (commented):
+0.5

ci/build_libtelio.py
dfetti: I placed them here because I wanted them to be more visible. They must be reverted when we get rid of the sm library.

matszczygiel: @dfetti Just add one line `#TODO: remove when we get rid of sm crate`

matszczygiel:
+1

ci/build_libtelio.py
tomaszklak: Please add a jira task id in that `TODO` comment

LukasPukenis (approved):

LukasPukenis (approved):

PR #357, opened 2024-02-12
[LLT-4908] backport configurable ttl value

packgron:
### Problem
TTL value is to big in telio-dns

### Solution
Backport Juozapas Solution.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-dns/src/zone.rs
+++ b/crates/telio-dns/src/zone.rs
@@ -12,9 +12,12 @@ use hickory_server::{
 };
 use std::{
     collections::{HashMap, HashSet},
+    convert::TryInto,
     net::IpAddr,
     str::FromStr,
 };
+use telio_model::api_config::TtlValue;
+use telio_utils::telio_log_warn;
 
 use crate::forward::ForwardAuthority;
 
@@ -35,7 +38,11 @@ pub(crate) struct AuthoritativeZone {
 }
 
 impl AuthoritativeZone {
-    pub(crate) async fn new(name: &str, records: &Records) -> Result<Self, String> {
+    pub(crate) async fn new(
+        name: &str,
+        records: &Records,
+        ttl_value: TtlValue,
+    ) -> Result<Self, String> {
         // TODO: rewrite code so that this assert is not needed.
         for domain in records.keys() {
             if !domain.contains(name) {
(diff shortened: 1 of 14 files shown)


nat-lab/tests/test_dns.py
```diff
@@ -843,3 +844,64 @@ async def test_dns_wildcarded_records() -> None:
...
+                    ),
+                ),
+            ],
+            provided_api=api,
+        )
+        client_alpha, *_ = env.clients
+        connection_alpha, *_ = [conn.connection for conn in env.connections]
+
+        await client_alpha.enable_magic_dns([FIRST_DNS_SERVER, SECOND_DNS_SERVER])
+        await asyncio.sleep(1)
```
tomaszklak, line 885: Will this not become a source of flakiness in the future?

nat-lab/tests/test_dns.py
```diff
@@ -843,3 +844,64 @@ async def test_dns_wildcarded_records() -> None:
...
+                    "dig",
+                    "+noall",
+                    "+nocmd",
+                    "+answer",
+                    "alpha.nord",
+                    "@100.64.0.2",
+                ]
+            ).run()
+        )
+        await asyncio.sleep(1)
```
tomaszklak, line 899: similarly, can't we wait for the process end to avoid sleeping here?

crates/telio-dns/src/zone.rs
```diff
@@ -44,21 +51,27 @@ impl AuthoritativeZone {
         }
         let zone_name = Name::from_str(name)?;
         let zone = InMemoryAuthority::empty(zone_name.clone(), ZoneType::Primary, false);
-
+        let ttl_value_signed: i32 = match ttl_value.0.try_into() {
+            Ok(ttl_value) => ttl_value,
+            Err(_) => {
+                telio_log_warn!("TTL value could not be converted from u32 to i32 without data loss, so using default value");
```
tomaszklak, line 57: I think this warning would be more useful if it contained the actual value which will be used instead of only the word "default".

tomaszklak (approved):
+1
Since this is a backport, may comments probably make more sense in the context of main, so if you agree with them, can you apply fixes to main (or maybe create a ticket to do them in future)?

tomaszklak:
Please squash before merging.

nat-lab/tests/test_dns.py
```diff
@@ -843,3 +844,64 @@ async def test_dns_wildcarded_records() -> None:
...
+        dig_stdout = process.get_stdout()
+        dig_stderr = process.get_stderr()
+
+        actual_ttl_value = int(dig_stdout.strip().split()[1])
+
+        assert (
+            actual_ttl_value == EXPECTED_TTL_VALUE
+        ), f"tcpdump stdout:\n{dig_stdout}\ntcpdump stderr:\n{dig_stderr}"
```
Jauler, line 907:
```suggestion
        ), f"dig stdout:\n{dig_stdout}\ndig stderr:\n{dig_stderr}"
```

nat-lab/tests/test_dns.py
Jauler: I agree with @tomaszklak as this sleep is pretty much guaranteed to be flaky. We need some better mechanism to wait for DNS server to become ready, which would mean to wait until WG session to internal virtual host is established.

In practical terms currently we do not report events for "internal" peers (like WG-STUN or internal virtual host acting as DNS). Therefore I see two option -> one is coupling to logs, although I am not 100% sure it is the best option, and the other is conditianlly reporting internal nodes for tests (e.g. enabled via feature config).

Jauler (approved):
+1.5

PR #359, opened 2024-02-13
[LLT-4916] Allow any case nickname

packgron:
### Problem

Nickname RFC allows nicknames to be passed
in any case. So there is no valid reason
to filter out nicknames that are only in
lowercase.

It was also agreed that nickname passed
with node event should be identical to one
passed via meshnet config.

### Solution

Remove validation that expects nicknames to be only in lowercase.
TrustDNS already handles these cases in an expected
manner. After passing a domain name, we can
query in any case.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 * LLT-4908: Lowered DNS TTL time for nicknames and NXDomain responses
 * LLT-4915: Update the moose tracker to v1.0.0-libtelioApp
 * LLT-4913: Skip qos analytics ping for unconnected peers
+* LLT-4916: Allow any case nicknames
 
 <br>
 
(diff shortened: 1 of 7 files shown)


LukasPukenis (dismissed):
+2.0

tomaszklak:
> After passing a domain name, we can query in any case.

Do we have a test that verifies that?

packgron:
> > After passing a domain name, we can query in any case.
> 
> Do we have a test that verifies that? It looks like we only have tests that verify that you can use the same string, but no test that checks that a different case version is also working.

Added a proper test

tomaszklak (dismissed):
+1

tomaszklak (approved):

PR #361, opened 2024-02-14
Do not update wg listen_port when not needed

jjanowsk:
Do not configure listen_port in wireguard adapter if it is already configured to the same value. At least in some implementations it is not a no-op and reconfiguring it causes the underlying socket to be closed and recreated

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
 * LLT-4406: Implement error handling for proxy egress
 * LLT-4948: Change missed tick behavior from burst to delay
 * LLT-4855: Add Aggregator struct for connectivity events gathering
+* LLT-4917: Fix the issue with wireguard-go failing to bind to a port on some devices
 
 <br>
 
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -620,6 +620,23 @@ impl State {
         Ok(())
     }
 
+    fn update_calculate_set_listen_port(
+        from_listen_port: Option<u16>,
+        to_listen_port: Option<u16>,
+    ) -> Option<u16> {
+        match (from_listen_port, to_listen_port) {
+            (Some(from), Some(to)) => {
+                if from != to {
+                    Some(to)
+                } else {
+                    None
+                }
+            }
+            (None, Some(to)) => Some(to),
+            _ => None,
+        }
+    }
+
     fn update_construct_set_device(
         &self,
         to: &uapi::Interface,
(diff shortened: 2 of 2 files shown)


matszczygiel (dismissed):
+1

LukasPukenis (commented):
The commit describes that setting the same listen port results in a socket being re-created, however I wonder why do we care? If a different port is being used we expect it to be recreated but if it's the same one and it's causing us issues it might mean that we have a funny control flow somewhere.

Could you shed some light on this one? Thanks!

jjanowsk:
> The commit describes that setting the same listen port results in a socket being re-created, however I wonder why do we care? If a different port is being used we expect it to be recreated but if it's the same one and it's causing us issues it might mean that we have a funny control flow somewhere.
> 
> Could you shed some light on this one? Thanks!

Under some circumstances binding immediately to the same protocol/address/port tuple can fail on Windows. At least this is the behavior we observe on one of the testing machines. Code as simple as this fails on this machine:
```go
package main

import (
	"net"
)

func main() {
	for i := 0; i < 100; i++ {
		addr := net.UDPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 12345}
		conn, err := net.ListenUDP("udp4", &addr)
		if err != nil {
			panic(err)
		}
		conn.Close()
	}
}
```
I've also confirmed the same behavior with the C++ example using basic winsock API: `getaddrinfo`, `socket`, `bind`, `closesocket` in a loop. So it's also not a bug in go's std library. That's the reason to not set listen port multiple times if it does not change. And when it comes to the logic that recreates the socket, that's just how the wireguard-go library works.

LukasPukenis:
Thanks for explaining. I still wonder what does it mean "recreates the underlying socket"? Looking at the example you provided I do not fully understand why it would cause any failures in the tunnel application itself somehow?

jjanowsk:
> Thanks for explaining. I still wonder what does it mean "recreates the underlying socket"? Looking at the example you provided I do not fully understand why it would cause any failures in the tunnel application itself somehow?

"recreates the underlying socket" means calling close on the old socket, creatin a new socket and calling bind on it :) And this call to `bind` can fail because of the reasons I've described above. Take a look at the https://www.wireguard.com/xplatform/ That's how you configure wireguard adapter. In case of wireguard-go if you include `listen_port` in a set operation it will close the old socket and open the new one.

LukasPukenis:
+1 thanks for explanation :)

LukasPukenis (approved):

LukasPukenis (approved):

PR #362, opened 2024-02-14
LLT-4948: Set missing tick behavior for tokio intervals

mathiaspeters:
### Problem
We were, for our purposes, incorrectly handling the cases where tokio misses one or more ticks when doing work based on an interval.

### Solution
We now set the missed tick behavior to delay, instead of the default burst. For more info, see:
https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/qos.rs
+++ b/crates/telio-nurse/src/qos.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use tokio::sync::mpsc;
-use tokio::time::{interval_at, Duration, Interval};
+use tokio::time::{interval_at, Duration, Interval, MissedTickBehavior};
 
 use telio_crypto::PublicKey;
 use telio_task::{io::mc_chan, Runtime, RuntimeExt, WaitResponse};
@@ -233,9 +233,11 @@ impl Analytics {
         } else {
             Arc::new(None)
         };
+        let mut rtt_interval = interval_at(tokio::time::Instant::now(), config.rtt_interval);
+        rtt_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
 
         Self {
-            rtt_interval: interval_at(tokio::time::Instant::now(), config.rtt_interval),
+            rtt_interval,
             io,
             nodes: HashMap::new(),
             ping_backend,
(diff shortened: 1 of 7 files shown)


mathiaspeters:
Personally, I don't see the need to cover this change with tests. We shouldn't be testing that tokio is working as advertised, in that case it would make more sense to upstream tests directly to tokio if there aren't already such tests, and I don't think we need to test in every place we use a task with an interval that it behaves as expected. I think the discussion in slack about introducing a lint for this would be much more useful

LukasPukenis:
The linter tackles this problem pretty nicely assuming intervals are the only thing we suspect and you're right that tokio should already be tested but I am thinking about the broader scope - the sleeping behaviour of the mobile device is an additional variable we need to reason about and be aware, I think it would be nice to have it covered. Maybe unit tests are too narrow for this but for integration tests I think it makes sense:
- do libtelio things
- fake sleep/advance time
- expect not a burst of events

@Jauler what is your take on this?

Jauler:
> The linter tackles this problem pretty nicely assuming intervals are the only thing we suspect and you're right that tokio should already be tested but I am thinking about the broader scope - the sleeping behaviour of the mobile device is an additional variable we need to reason about and be aware, I think it would be nice to have it covered. Maybe unit tests are too narrow for this but for integration tests I think it makes sense:
> 
>     * do libtelio things
> 
>     * fake sleep/advance time
> 
>     * expect not a burst of events
> 
> 
> @Jauler what is your take on this?

Well, IMHO there are a few points, which a clear to me:
* Covering sleep unit/integration tests is out of scope for this specific PR, as this should bigger/wider initiative than the bugfix
* Keeping in mind, that default `tokio::time::interval` `MissedTickBehavior` is almost always unsuitable for us, having a lint really helps us remember to think about it.

In terms of unit and/or integration tests, yes, I would agree that having that covered would be nice, but I have no good picture in my head how we could make this generic. Therefore maybe more discussion is needed?

tomaszklak:
> Personally, I don't see the need to cover this change with tests. We shouldn't be testing that tokio is working as advertised.

I don't think this is the case. We want to know with a test, that telio after sleep will not generate huge number of events/request proportional to the duration of sleep. It doesn't matter how we implement this - tokio interval and a way we configure it is just an implementation detail. Ideally, the tests will verify only **how** we behave, not how that behaviour is achieved.

LukasPukenis:
@Jauler thanks for elaborating. As discussed in the thread ther's no obvious way to do it how I want it but I will sleep on it(pun intended).

Regardless, it's out of scope for this PR but it's a good place to have such a discussion.

LukasPukenis:
+1

Jauler (approved):
+1.5

mathiaspeters:
I will create a ticket for adding a test for this, but I'm merging this without the test

PR #363, opened 2024-02-15
Backport LLT-4948 to 4.2

mathiaspeters:
### Problem
We were, for our purposes, incorrectly handling the cases where tokio misses one or more ticks when doing work based on an interval. 

### Solution
We now set the missed tick behavior to delay, instead of the default burst. For more info, see:
https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html

- [x] README.md is updated
- [x] changelog.md is updated


tomaszklak (approved):
+2

PR #370, opened 2024-02-20
Use null terminated strings in uniffi test for path length

tomaszklak:
### Problem
`String` was converted to a slice of bytes, from which a point was created. This doesn't guarantee that a resulting pointer points to an array ending with a nul byte.

Printing that string in lldb shows that it can end with 'random' bytes:

```
(lldb) set set target.max-string-summary-length 100000000
(lldb) p ptr
"aaa(...)aaaaa\xe8$N%\U00000002"
```

### Solution
Use `CString` to have a guaranteed null terminated c style string.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -1914,14 +1914,14 @@ mod tests {
             id: rand::thread_rng().gen::<usize>(),
         };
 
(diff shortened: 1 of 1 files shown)


lcruz99 (approved):
+2

gytsto (approved):

PR #371, opened 2024-02-21
Unit tests on mac

tomaszklak:
### Problem
Tests are not run at all on the mac, which causes accidental breakage. Additionally the test runner reruns the tests when they fail.

### Solution
Add a unit tests job on a mac runner. The tests that are currently not working will have separate tickets created to investigate them and fix them.
The test runner is modified so that it only runs the tests once.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/test_runner.sh
+++ b/test_runner.sh
@@ -1,2 +1,7 @@
 #!/usr/bin/env bash
-which sudo &> /dev/null && sudo $@ || $@
+
+if command -v sudo &> /dev/null; then
+    sudo $@
+else
+    $@
+fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,6 +24,15 @@ jobs:
         with:
           command: test
           args: --all --lib -- --nocapture
+          
+  test-mac:
+    runs-on: macos-13
+    steps:
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - uses: actions-rs/cargo@ae10961054e4aa8b4aa7dffede299aaf087aa33b # v1.0.3
+        with:
+          command: test
+          args: --all --lib -- --nocapture
 
   test-ffi:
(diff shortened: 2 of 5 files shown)


test_runner.sh
```diff
@@ -1,2 +1,7 @@
 #!/usr/bin/env bash
-which sudo &> /dev/null && sudo $@ || $@
+
+if command -v sudo &> /dev/null; then
+    sudo $@
+else
+    $@
+fi
```
LukasPukenis, line 7: What does this script do?

tomaszklak: It runs the tests executables under sudo if it exists.

LukasPukenis: Could you elaborate? Does it depend on different runners, like mac vs linux?

tomaszklak: I think the intention here is that when you run on a platform which has sudo, then use it but otherwise assume you are already a root user. To be honest I didn't investigate that much - it's in the MR 23 on the original libtelio repo and there are no information in that MR that would explain why it was done this way.

LukasPukenis: Thanks for explaining.

LukasPukenis (commented):
+1

jjanowsk (approved):
Have you disabled those tests in qos and stun because they are currently failing and you want to fix them in the future, or is there an inherent issue why they cannot be run? If the former then please create a JIRA if you do not plan to do it immediately after this is merged.

Btw you can close LLT-4454 with this PR.

tomaszklak:
> Have you disabled those tests in qos and stun because they are currently failing and you want to fix them in the future, or is there an inherent issue why they cannot be run? If the former then please create a JIRA if you do not plan to do it immediately after this is merged.

I don't know if there are any inherent issues - all I know is that those tests are not working right now on a mac. My plan is to create jira tickets to investigate each test in future, but I personally probably won't have the time to do this in the immediate future.

PR #372, opened 2024-02-21
Fix gitignore

jjanowsk:
Do not ignore all files and directories starting with . because we are actually tracking some of them.
Instead ignore only directories in the root folder starting with . except for a few that we track.

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (approved):
+2

PR #373, opened 2024-02-21
add string replacement in file script

gytsto:
### Problem
We need to update libtelio version in our builds w/o rebuilding them for future promotions of builds.

### Solution
Add script that allows to replace certain string in provided file.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/ci/replace_string.py
+++ b/ci/replace_string.py
@@ -12,6 +12,12 @@ def replace_string_in_file(file_path: str, original_str: str, new_str: str):
         if original_str.encode() not in content:
             raise ValueError("Original string not found in the file")
 
+        # Check if there are more than one matches of the original string
+        if content.count(original_str.encode()) > 1:
+            raise ValueError(
+                "More than one match of the original string found in the file"
+            )
+
         # Check if the length of the new string is greater than the original
         if len(new_str) > len(original_str):
             raise ValueError("New string is longer than the original string")
--- a/ci/replace_string.py
+++ b/ci/replace_string.py
@@ -1,0 +1,51 @@
+#!/usr/bin/python3
+
+import argparse
+
+
+def replace_string_in_file(file_path: str, original_str: str, new_str: str):
+    # Read the file
+    with open(file_path, "rb+") as file:
+        content = file.read()
+
+        # Check if the original string exists in the file
+        if original_str.encode() not in content:
+            raise ValueError("Original string not found in the file")
+
+        # Check if the length of the new string is greater than the original
+        if len(new_str) > len(original_str):
+            raise ValueError("New string is longer than the original string")
+
+        # If the new string is shorter than the original, fill with null terminators
+        null_terminators = ""
+        if len(new_str) < len(original_str):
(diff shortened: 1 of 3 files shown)


tomaszklak:
Please add a proper description to the PR and update the commit so that it's clear what is the purpose of this change.

ci/replace_string.py
```diff
@@ -0,0 +1,51 @@
...
+import argparse
+
+
+def replace_string_in_file(file_path: str, original_str: str, new_str: str):
+    # Read the file
+    with open(file_path, "rb+") as file:
+        content = file.read()
+
+        # Check if the original string exists in the file
+        if original_str.encode() not in content:
```
tomaszklak, line 12: Maybe we should also check that the string in question in present only in one place?

ci/replace_string.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/python3
+
+import argparse
+
+
+def replace_string_in_file(file_path: str, original_str: str, new_str: str):
```
tomaszklak, line 6: Please add tests

gytsto: added

ci/test_replace_string.py
```diff
@@ -0,0 +1,91 @@
...
+        original_str = "original"
+        new_str = "new"
+        replace_string_in_file(temp_file.name, original_str, new_str)
+
+        # Verify the contents of the file after replacement
+        with open(temp_file.name, "r", encoding="utf-8") as file:
+            content = file.read()
+            self.assertEqual(
+                content,
+                "new" + "\0" * (len(original_str) - len(new_str)),
```
tomaszklak, line 83: this is coping an expression from the implementation. In tests it would be better to be explicit, which in this case means to compare against a literal expected string.

ci/test_replace_string.py
```diff
@@ -0,0 +1,91 @@
...
+            )
+
+        # Clean up the temporary file
+        os.unlink(temp_file.name)
```
tomaszklak, line 87: Is this necessary? The documentation seems to imply that the file will be automatically deleted: 

> If delete is true (the default) and delete_on_close is true (the default), the file is deleted as soon as it is closed. If delete is true and delete_on_close is false, the file is deleted on context manager exit only, or else when the [file-like object](https://docs.python.org/3/glossary.html#term-file-like-object) is finalized. Deletion is not always guaranteed in this case (see [object.__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__)). If delete is false, the value of delete_on_close is ignored.

This also applies to other tests.

tomaszklak (approved):
+2, when two remaining comments in tests are addressed.

ci/test_replace_string.py
gytsto: done

ci/test_replace_string.py
gytsto: removed

tomaszklak (approved):

PR #375, opened 2024-02-22
Bump moose tracker to v2.0.0-libtelioApp

dfetti:
Update moose tracker version.

The previous version v1.0.0 broke the data model as it removed the current_state from the json.
The new version v2.0.0 readded it as it's visible in the diff.

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis:
It would be nice if the commit message would contain what was changed in the tracker instead of just a version it's being updated to. Judging from the diff it's hard to tell as it just contains renamed fields.

tomaszklak:
Is there some specific reason why we update tracker (do we need something which it brings)? Or is it a just a periodic update of a dependency? Whichever it is might be good idea to explain that in the commit/pr description.

dfetti:
I think the reason we want it in the next release is that the previous version of moose tracker broke the data model as it removed current_state field.
The new version readded it.
I think @packgron can elaborate more.

Jauler (approved):
+1.0

tomaszklak (approved):
+1

PR #376, opened 2024-02-22
Multicast packets forwarding POC

baltrus1:
This is a PoC for multicast communication.
It includes multicast <-> multicast communication (mdns like for eg) as well as support for multicast -> unicast response (SSDP like for eg).

P.S.  Since this is a multicast there are a number of architectural and code optimizations not done. There are also a few tests that might be missing from here.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4268,6 +4268,7 @@ dependencies = [
  "telio-proxy",
  "telio-relay",
  "telio-sockets",
+ "telio-starcast",
  "telio-task",
  "telio-test",
  "telio-traversal",
@@ -4567,6 +4568,24 @@ dependencies = [
  "version-compare",
  "winapi",
  "windows 0.34.0",
+]
+
+[[package]]
+name = "telio-starcast"
+version = "0.1.0"
+dependencies = [
+ "base64 0.13.1",
+ "boringtun",
+ "ipnetwork",
+ "pnet_packet",
+ "telio-crypto",
+ "telio-sockets",
+ "telio-task",
+ "telio-utils",
+ "telio-wg",
+ "tokio",
+ "tracing",
+ "x25519-dalek 2.0.1",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3476,9 +3476,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.21.9"
+version = "0.21.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "629648aced5775d558af50b2b4c7b02983a04b312126d45eeead26e7caa498b9"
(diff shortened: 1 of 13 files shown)


Cargo.lock
```diff
@@ -2907,6 +2922,15 @@ dependencies = [
  "no-std-net",
 ]
 
+[[package]]
```
packgron, line 2925: We introduced a second version of same lib, can you make sure we reuse version by adding to workspace deps?

crates/telio-starcast/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies]
+tokio = { workspace = true, features = ["rt", "net", "sync", "macros"] }
+tracing.workspace = true
+boringtun.workspace = true
+x25519-dalek.workspace = true
+base64.workspace = true
+ipnetwork.workspace = true
+pnet_packet.workspace = true
+pnet_transport = "0.34.0"
+etherparse = "0.14.2"
```
packgron: why is this and pnet_packet both needed ?

packgron: Theu are doing same things

crates/telio-starcast/src/multicast_peer.rs
```diff
@@ -0,0 +1,131 @@
...
+pub struct MulticasterIp {
+    pub ipv4: Ipv4Addr,
+    pub ipv6: Ipv6Addr,
+}
+
+impl MulticasterPeer {
+    pub async fn new(
+        telio_wg_port: u16,
+        public_key: &PublicKey,
+        wg: Arc<DynamicWg>,
```
packgron, line 32: Should probably be Arc<dyn Wg>, and why is port needed if we can fetch it from wg?

Cargo.lock
Hasan6979: Fixed!

crates/telio-starcast/src/multicast_peer.rs
Hasan6979: What should be Arc ? 
Port isn't needed. leftover.

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,295 @@
...
+        if let Some(handle) = &self.transport_to_local_task {
+            handle.abort();
+        }
+        if let Some(handle) = &self.tun_to_transport_task {
+            handle.abort();
+        }
+    }
+
+    pub async fn tun_to_transport(
+        read_chan: Arc<RwLock<Chan<Vec<u8>>>>,
```
packgron, line 107: This is a heavy anti patten, you should never have channel in RwLock, this causes often leads to deadlocks.

PR #389, opened 2024-02-26
LLT-4980: Missing sync in telio-proxy

mathiaspeters:
### Problem
We were getting the wrong endpoint map from the proxy when trying to configure the wg controller. There was recently some error handling added which made the problem visible, but the problem was there for longer. The root of the issue was that if there was a socket error, StateIngress would swallow the error while StateEgress would actually handle it, yet when we were fetching the endpoint map for the wg controller, we would get it from StateIngress. As a result, if there was a socket error, ingress and egress could be out of sync, and we would get the wrong values passed to the wg controller.

### Solution
There is now synchronization logic added for ingress and egress, and the endpoint map is fetched from egress

Also, with this PR, it's possible to pass the `-x` flag to `run_local.py` and have that be propagated to pytest

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-proxy/src/proxy.rs
+++ b/crates/telio-proxy/src/proxy.rs
@@ -20,7 +20,7 @@ use telio_task::{
 };
 use tokio::{net::UdpSocket, sync::mpsc::error::SendTimeoutError};
 
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, PinnedSleep};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn, PinnedSleep};
 
 type SocketMap = HashMap<PublicKey, Arc<UdpSocket>>;
 type SocketMuteMap = HashMap<PublicKey, (Arc<UdpSocket>, Option<PinnedSleep<PublicKey>>)>;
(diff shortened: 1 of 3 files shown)


crates/telio-proxy/src/proxy.rs
```diff
@@ -323,8 +334,18 @@ impl StateEgress {
...
                     }
                 };
+
                 if let Some(pk) = pk {
-                    self.sockets.insert(pk, (Arc::new(socket), None));
+                    let socket = Arc::new(socket);
+
+                    if let Err(e) = self.update_socket.send((pk, socket.clone())).await {
+                        // May happen when they are being closed, but is not critical
+                        telio_log_debug!(
```
Jauler, line 343: Would it be possible to somehow make this `telio_log` to have verbosity level info or above, but in case of many errors avoid spamming? It feels like issue syncing up sockets between two tasks should be something, that raises our attention in the logs if it occurs.

mathiaspeters: It's of course possible, but I wonder if the effort is really worth benefit compared to just doing the same level all the time. I can definitely change it to info or warn though

packgron (dismissed):
+1

packgron (approved):

PR #393, opened 2024-02-27
Fix spammy logs in telio-proxy

packgron:
### Problem
New telio-proxy error logging is to spamy

### Solution
Switch to warn, don't treat would block as a reportable error on ingress

- [x] README.md is updated
- [x] changelog.md is updated


mathiaspeters (approved):
+2

PR #395, opened 2024-02-27
Get rid of DERP dependencies in Nurse

stalowyjez:
### Problem
Nurse module currently depends on Relay module, and this dependency is unneeded

### Solution
Remove the unneeded dependency


crates/telio-nurse/src/heartbeat.rs
```diff
@@ -508,8 +491,21 @@ impl Analytics {
         }
     }
 
-    async fn handle_derp_event(&mut self, event: Server) {
-        self.derp_connection = event.conn_state == RelayState::Connected;
+    async fn handle_derp_event(&mut self, derp_event: Server) {
```
Hasan6979, line 494: Name is a bit misleading here. Earlier it was setting whether derp is connected or not, and now it is setting the current nat type.

Hasan6979 (commented):
Other than the above issue +1.5

Jauler (approved):
+1.0

PR #397, opened 2024-02-28
Add UpgradeDecision message and session field to Upgrade message

tomaszklak:
This will be used to decide if upgrade is accepted or rejected in LLT-4858. For now it is used to detect situations when the recipient of Upgrade message is unaware of the associated session id, most likely due to restart.

### Problem
Currently there is no way to distinguish initiators in a upgrade procedures when 2 procedures start at the same time. Furthermore when the other side is restarted after it sent CMM response, it will accept any update request as valid.

### Solution
Implement RFC LLT-0058 which prepares us for detecting initiator and solves the rejection of unknown update requests (e.g. after a restart).

`Upgrade` message is extended with a `session` field and a new message `UpgradeDecision` is added to be sent as a reply for the `Upgrade. The decision let's the other side reject unknown requests.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-proto/src/packet/relayed/upgrade.rs
+++ b/crates/telio-proto/src/packet/relayed/upgrade.rs
@@ -2,7 +2,7 @@ use std::net::SocketAddr;
 
 use crate::{
     messages::upgrade::*, Codec, CodecError, CodecResult, DowncastPacket, PacketRelayed,
-    PacketTypeRelayed, MAX_PACKET_SIZE,
+    PacketTypeRelayed, Session, MAX_PACKET_SIZE,
 };
 
 use bytes::BufMut;
@@ -13,6 +13,8 @@ use protobuf::Message;
 pub struct UpgradeMsg {
     /// Endpoint which message sender is requesting to upgrade to
     pub endpoint: SocketAddr,
+    /// Session id created by CrossPingCheck
+    pub session: Session,
 }
 
 impl Codec<PacketTypeRelayed> for UpgradeMsg {
@@ -36,7 +38,9 @@ impl Codec<PacketTypeRelayed> for UpgradeMsg {
                     .get_endpoint()
                     .parse()
                     .map_err(|_| CodecError::DecodeFailed)?;
-                Ok(Self { endpoint })
+                let session: Session = proto_upgrade.session;
+
+                Ok(Self { endpoint, session })
             }
             _ => Err(CodecError::DecodeFailed),
         }
@@ -46,6 +50,7 @@ impl Codec<PacketTypeRelayed> for UpgradeMsg {
         let mut bytes = Vec::with_capacity(MAX_PACKET_SIZE);
         let mut msg = Upgrade::new();
         msg.set_endpoint(self.endpoint.to_string());
+        msg.set_session(self.session);
 
         bytes.put_u8(PacketTypeRelayed::Upgrade as u8);
         msg.write_to_vec(&mut bytes)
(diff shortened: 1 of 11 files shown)


crates/telio-proto/src/packet/relayed/upgrade.rs
```diff
@@ -71,17 +76,125 @@ impl DowncastPacket<PacketRelayed> for UpgradeMsg {
...
+
+    fn decode(bytes: &[u8]) -> CodecResult<Self>
+    where
+        Self: Sized,
+    {
+        if bytes.is_empty() {
+            return Err(CodecError::InvalidLength);
+        }
+
+        match PacketTypeRelayed::from(*bytes.first().unwrap_or(&(PacketTypeRelayed::Invalid as u8)))
```
matszczygiel, line 136: This can be merged into one branch
```
let Some(first) = bytes.first() else {
    return Err(CodecError::InvalidLength);
};
```

crates/telio-proto/src/packet/relayed/upgrade.rs
```diff
@@ -71,17 +76,125 @@ impl DowncastPacket<PacketRelayed> for UpgradeMsg {
...
+        {
+            PacketTypeRelayed::UpgradeDecision => {
+                let proto_upgrade_decision = UpgradeDecision::parse_from_bytes(
+                    bytes.get(1..).ok_or(CodecError::DecodeFailed)?,
```
matszczygiel, line 140: You've already checked that bytes len is more than 1 twice. IMHO `bytes[1..]` is OK for this one

tomaszklak: I think I will combine both of your suggestions into `split_first`.

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -30,6 +30,7 @@ use tokio::time::{interval_at, Instant, Interval};
 use tokio::{sync::Mutex, time::MissedTickBehavior};
 
 const CPC_TIMEOUT: Duration = Duration::from_secs(10);
+const UPGRADE_TIMEOUT: Duration = Duration::from_secs(60);
```
Jauler, line 33: IMHO, maybe it is "safer" to use some bounded LRU container? 

E.g. at least theoretically storing values for fixed amount of time, allows for malicious meshnet node to spam so many session IDs, that "session_id_candidates" would explode in size.

IMHO, we can get some fixed capacity container, and store up to, for example, most recent 512 entries?

What are your thoughts about :arrow_up: ?

tomaszklak: I think that it's a good point and will implement it :)

matszczygiel (approved):
+1

src/device/wg_controller.rs
```diff
@@ -250,9 +250,19 @@ async fn consolidate_wg_peers<
                     requested_peer.peer.endpoint,
                     requested_peer.local_direct_endpoint,
                 ) {
-                    if let Some(us) = upgrade_sync {
-                        us.request_upgrade(&public_key, remote_endpoint, local_direct_endpoint)
-                            .await?;
+                    if let (Some(us), Some(cpc)) = (upgrade_sync, cross_ping_check) {
+                        match cpc.session_for_key(public_key).await? {
```
Jauler, line 254: I wonder if this is really correct.

As `session_for_key()` returns any session which matches provided `public_key`. In general CPC module creates this matrix of:
```
+----------------+------------+------------+------------+
| .              | Peer 1     | Peer 2     | Peer 3     |
+----------------+------------+------------+------------+
| EP Candidate 1 | session 11 | session 12 | session 13 |
+----------------+------------+------------+------------+
| EP Candidate 2 | session 21 | session 22 | session 23 |
+----------------+------------+------------+------------+
| EP Candidate 3 | session 31 | session 32 | session 33 |
+----------------+------------+------------+------------+
```

I think `session_for_key()` function should match not only public key, but also either `local_direct_endpoint`.

Jauler: I probably could have mentioned this without using ASCII tables :D but who can pass such an opportunity :hugs:

:laughing: :laughing: :laughing:

Jauler: Additionally there is possibility to add session ID to [WireGuardEndpointCandidateChangeEvent](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/connectivity_check/mod.rs#L35).

As this gets probed from CPC in wg_controller [here](https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L649), later passed to [select_endpoint_for_peer()](https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L701), which depending on conditions may select CPC [provided endpoint](https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L848), which uses previously mentioned [struct](https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L784). 

Of course, this would mean, that local_direct_endpoint, would have to carry "session ID" back up to the point where `update_request` is initiated. But it should be quite convenient to do via the [ReqestedPeer](https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L784) struct. This would allow us to avoid two separate lookups in CPC state. Which, perfromance-wise does not have any impact, but having single "pass" from CPC to WG of the map of [WireGuardEndpointCandidateChangeEvent](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/connectivity_check/mod.rs#L35) structs -> ensures consistency in a trivial manner :).

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -577,6 +602,61 @@ impl<E: Backoff> State<E> {
...
+
+    pub fn session_for_key(&self, pk: PublicKey) -> Option<Session> {
+        self.endpoint_connectivity_check_state
+            .iter()
+            .find(|(_, state)| state.public_key == pk)
+            .map(|(session, _)| *session)
+    }
+
+    pub fn check_if_upgrade_is_allowed(&mut self, session: Session, public_key: PublicKey) -> bool {
+        match self.session_id_candidates.remove(&session) {
```
Jauler, line 614: Generally the lifetime of the session in this case is tied to endpoint candidate and peer being in meshnet config. Where sessions are instantiated [here](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs#L373) and [here](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs#L434) respectively when peer list has changed or when endpoint (local) list has changed. 

Therefore it is valid to have multiple upgrade sessions for the same session, thus maybe we should not remove session id from candidates when upgrade is received?

tomaszklak: True, but I think that we other side will always get that session id in a call me maybe, so when we check if it allowed it will already be there even if we remove it while checking it.

src/device/wg_controller.rs
tomaszklak: Done

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
Jauler: Oh I was thinking about the case where single CPC cycle is used for multiple upgrades.

Jauler: Ok, I double-checked a bit, and now we are emitting "endpoint gone" event on every "downgrade" which would mean, that probably there are no cases where single CPC cycle is used for multiple upgrades. Just not sure if we want to couple these two things together.

nat-lab/tests/test_direct_connection.py
```diff
@@ -269,6 +269,122 @@ async def test_direct_working_paths(
...
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        env = await setup_mesh_nodes(exit_stack, setup_params)
+        alpha, beta = env.nodes
+        alpha_client, beta_client = env.clients
+        alpha_connection, _ = [conn.connection for conn in env.connections]
+
+        await testing.wait_defined(
+            asyncio.gather(
+                alpha_client.wait_for_state_peer(
```
gytsto, line 286: State check not needed here, if both of nodes have direct feature enabled, state is check in `setup_mesh_nodes`

gytsto (commented):
+1

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -193,6 +233,7 @@ impl<E: Backoff> CrossPingCheck<E> {
                 poll_timer,
                 ping_pong_handler,
                 exponential_backoff_helper_provider,
+                session_id_candidates: LruCache::new(UPGRADE_TIMEOUT, 512),
```
Jauler, line 236: Maybe it would be helpful to have constant for capacity as well? if we have one for timeout value :thinking:. Although this is nitpick TBH.

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
Jauler: You know, after realizing "endpoint gone" thing and thinking a bit more about this -> I have changed my mind, I think your suggestion is better :P

Jauler (dismissed):
+1.0

dfetti (approved):
+0.5

gytsto (approved):

PR #398, opened 2024-02-29
add version placeholder

gytsto:
### Problem

### Solution


ci/insert_libtelio_version.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import replace_string
+import sys
+from build_libtelio import LIBTELIO_CONFIG, PROJECT_CONFIG, PROJECT_ROOT
+from rust_build_utils.rust_utils_config import GLOBAL_CONFIG
+
+VERSION_PLACEHOLDER = "VERSION_PLACEHOLDER@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@"
+
+
+def insert_version_to_libtelio_binaries_in_dir(new_version: str, path: str):
```
Jauler, line 13: Quick thought, would you be able to ensure, that in case there is no replacement (e.g. zero placeholders found in file) -> this function throws an error, in the hopes of ensuring, that we do not miss somehow version replacement.

gytsto: `replace_string_in_file` throws an exception if original string, that we want to replace, is not found

.gitlab-ci.yml
```diff
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v0.4.17
+    branch: LLT-4902_add_tags_to_builds
```
matislovas, line 18: REMINDER: Remove this before merging.

.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v0.4.17
+          triggered-ref: LLT-4902_add_tags_to_builds
```
matislovas, line 34: REMINDER: Remove this before merging.

ci/replace_string.py
```diff
@@ -12,12 +13,6 @@ def replace_string_in_file(file_path: str, original_str: str, new_str: str):
         if original_str.encode() not in content:
             raise ValueError("Original string not found in the file")
 
-        # Check if there are more than one matches of the original string
-        if content.count(original_str.encode()) > 1:
-            raise ValueError(
-                "More than one match of the original string found in the file"
-            )
-
```
tomaszklak, line 20: Could you explain why do we now expect more than one occurrence of the placeholder?

gytsto: Somehow `macos aarch64` has two occurences of version string, therefore, I removed that exception and replaced 2 places in `macos aarch64`.

image

crates/telio-utils/src/git.rs
```diff
@@ -4,11 +4,8 @@ pub fn commit_sha() -> &'static str {
...
-/// Get the version tag that was used when building.
-/// If not present defaults to "dev".
+/// Get the version placeholder (half of the maximum length of git tag)
+/// Will be replaced during build promotions
 pub fn version_tag() -> &'static str {
-    option_env!("LIBTELIO_COMMIT_TAG")
-        .map(|tag| tag.trim())
-        .and_then(|tag| if tag.is_empty() { None } else { Some(tag) })
-        .unwrap_or("dev")
+    "VERSION_PLACEHOLDER@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@"
```
tomaszklak, line 10: Shouldn't this be some sort of `CStr`? Normal `&str` in not null terminated so will have a length independent of the position of first null byte: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a288ca452d106cf2604838c4610382df

ci/replace_string.py
tomaszklak: That is strange, I just built libtelio.a from this branch and it looks like there is only one match:
```
$ rg --binary --count VERSION_PLACEHOLDER target/debug/libtelio.a
1
$ rg --binary --count VERSION_PLACEHOLDER target/release/libtelio.a
1
```

gytsto: what if you were to download from gitlab artifacts?

gytsto: created issue at Jira LLT-5001

crates/telio-utils/src/git.rs
gytsto: fixed I guess :smile:

tomaszklak: I'm wondering if we shouldn't test this somehow. Maybe create an [example executable](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples) that all it does is output the result of `version_tag` to stdout, apply modification to it using the script and verify that the output has changed? This could be run on mac,win,linux on github CI to make sure that we aren't breaking the libraries somehow.

crates/telio-utils/src/git.rs
```diff
@@ -1,14 +1,23 @@
...
+    // Attempt to create a CStr from bytes with null termination
+    let cstr = match CStr::from_bytes_with_nul(b"VERSION_PLACEHOLDER@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\0") {
+        Ok(cstr) => cstr,
+        Err(_) => {
+            return "";
+        }
+    };
+    match cstr.to_str() {
+        Ok(str_ref) => str_ref,
+        Err(_) => "",
```
tomaszklak, line 21: nit: I think it would be better to return two different strings in both error cases so that if it ever happens, we know immediately which one of the errors happened.

matislovas:
+1

Jauler (approved):
+1.0

PR #400, opened 2024-03-04
LLT-4980: Improve proxy endpoint update logic

mathiaspeters:
### Problem
There was recently a commit that introduced synchronization for the endpoint maps of StateIngress and StateEgress, which fixed an issue where StateEgress would update it's endpoint map on socket error, but StateIngress would be used as the source of truth, leading to an incorrect endpoint map being used. 

### Solution
This commit improves that work in a few ways. It:
- Adds a test to check that the scenario works
- Makes wg_controller know about both the old endpoint and the new endpoint
- Resends the data to make libtelio adapt to endpoint changes quicker

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-proxy/src/proxy.rs
+++ b/crates/telio-proxy/src/proxy.rs
@@ -20,7 +20,10 @@ use telio_task::{
 };
 use tokio::{net::UdpSocket, sync::mpsc::error::SendTimeoutError};
 
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn, PinnedSleep};
+use telio_utils::{
+    exponential_backoff::{Backoff, ExponentialBackoff, ExponentialBackoffBounds},
+    telio_log_debug, telio_log_error, telio_log_info, telio_log_warn, PinnedSleep,
+};
 
 type SocketMap = HashMap<PublicKey, Arc<UdpSocket>>;
 type SocketMuteMap = HashMap<PublicKey, (Arc<UdpSocket>, Option<PinnedSleep<PublicKey>>)>;
@@ -134,6 +137,8 @@ struct StateEgress {
     wg_addr: Option<SocketAddr>,
     conn_state: Result<(), std::io::ErrorKind>,
     update_socket: Tx<(PublicKey, Arc<UdpSocket>)>,
+    peer_socket_backoffs: HashMap<PublicKey, (ExponentialBackoff, PinnedSleep<PublicKey>)>,
+    replaced_sockets: EndpointMap,
 }
 
 impl UdpProxy {
@@ -157,6 +162,8 @@ impl UdpProxy {
                 wg_addr: None,
                 conn_state: Err(ErrorKind::NotConnected),
                 update_socket,
+                peer_socket_backoffs: HashMap::new(),
+                replaced_sockets: HashMap::new(),
             }),
         }
     }
@@ -171,6 +178,7 @@ impl UdpProxy {
         .await??;
 
         task_exec!(&self.task_egress, async move |state| {
+            state.reset_socket_limit().await;
             Ok(state.configure(sockets, port).await)
         })
         .await?
@@ -180,6 +188,15 @@ impl UdpProxy {
     pub async fn stop(self) {
         let _ = self.task_egress.stop().await.resume_unwind();
         let _ = self.task_ingress.stop().await.resume_unwind();
+    }
+
+    /// Notify proxy about network change
+    pub async fn on_network_change(&self) {
+        let _ = task_exec!(&self.task_egress, async move |state| {
+            state.reset_socket_limit().await;
+            Ok(())
+        })
+        .await;
     }
 }
 
@@ -252,11 +269,18 @@ impl StateEgress {
         Ok(())
     }
 
-    async fn get_endpoints_map(&self) -> Result<EndpointMap, Error> {
+    async fn reset_socket_limit(&mut self) {
+        self.peer_socket_backoffs.clear();
+        self.replaced_sockets.clear();
+    }
+
+    async fn get_endpoints_map(&mut self) -> Result<EndpointMap, Error> {
         let mut map = EndpointMap::new();
 
         for (key, (sock, _)) in self.sockets.iter() {
-            map.insert(*key, sock.local_addr()?);
+            let mut peer_endpoints = self.replaced_sockets.remove(key).unwrap_or_default();
+            peer_endpoints.insert(0, sock.local_addr()?);
+            map.insert(*key, peer_endpoints);
         }
 
         Ok(map)
@@ -283,23 +307,28 @@ impl StateEgress {
     }
 
     async fn send_outbound_data(&mut self, pk: PublicKey, msg: DataMsg) {
-        // if let Some((pk, msg)) = self.input.recv().await {
-        if let (Some((socket, None)), Some(wg_addr)) = (self.sockets.get(&pk), self.wg_addr) {
-            match socket.send_to(msg.get_payload(), wg_addr).await {
-                Ok(_) => {
-                    if self.conn_state.is_err() {
-                        self.conn_state = Ok(());
-                        telio_log_info!("Outbound conn telio -> WG working!");
+        match (self.sockets.get(&pk), self.wg_addr) {
+            (Some((socket, None)), Some(wg_addr)) => {
+                match socket.send_to(msg.get_payload(), wg_addr).await {
+                    Ok(_) => {
+                        if self.conn_state.is_err() {
+                            self.conn_state = Ok(());
+                            telio_log_info!("Outbound conn telio -> WG working!");
+                        }
+                        self.peer_socket_backoffs.remove(&pk);
+                    }
+                    Err(err) => {
+                        self.handle_error(err, Some(pk)).await;
                     }
                 }
-                Err(err) => self.handle_error(err, Some(pk)).await,
             }
-        } else {
-            self.handle_error(
-                std::io::Error::new(ErrorKind::AddrNotAvailable, "WG Address not available"),
-                Some(pk),
-            )
-            .await;
+            _ => {
+                self.handle_error(
+                    std::io::Error::new(ErrorKind::AddrNotAvailable, "WG Address not available"),
+                    Some(pk),
+                )
+                .await;
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


crates/telio-proxy/src/proxy.rs
```diff
@@ -262,6 +286,12 @@ impl StateEgress {
         Ok(map)
     }
 
+    fn get_replaced_sockets(&mut self) -> Result<EndpointMap, Error> {
```
matszczygiel, line 289: Nitpick: since it clears `self.replaced_sockets` maybe let's call it `take_replaced_sockets()`?

crates/telio-proxy/src/proxy.rs
```diff
@@ -262,6 +286,12 @@ impl StateEgress {
...
+        let mut output = HashMap::new();
+        std::mem::swap(&mut output, &mut self.replaced_sockets);
+        Ok(output)
```
matszczygiel, line 292:
```suggestion
        Ok(std::mem::take(&mut self.replaced_sockets))
```

crates/telio-proxy/src/proxy.rs
```diff
@@ -283,27 +313,38 @@ impl StateEgress {
...
-        // if let Some((pk, msg)) = self.input.recv().await {
-        if let (Some((socket, None)), Some(wg_addr)) = (self.sockets.get(&pk), self.wg_addr) {
-            match socket.send_to(msg.get_payload(), wg_addr).await {
-                Ok(_) => {
-                    if self.conn_state.is_err() {
-                        self.conn_state = Ok(());
-                        telio_log_info!("Outbound conn telio -> WG working!");
+        if let Err(err) = self.send_data(pk, msg.clone()).await {
+            let mut should_resend = self.handle_error(err, Some(pk)).await;
+            while should_resend {
```
matszczygiel, line 318: Is there a possibility that this loop will spin forever? If yes, maybe let's limit the number of turns

crates/telio-proxy/src/proxy.rs
```diff
@@ -283,27 +313,38 @@ impl StateEgress {
...
-            .await;
+            Err(std::io::Error::new(
+                ErrorKind::AddrNotAvailable,
+                "WG Address not available",
+            ))
         }
     }
 
-    async fn handle_error(&mut self, err: std::io::Error, pk: Option<PublicKey>) {
+    async fn handle_error(&mut self, err: std::io::Error, pk: Option<PublicKey>) -> bool {
```
matszczygiel, line 347: Nitpick: bool as a return type is pretty ambitious. Would be nice to replace it with a custom enum or something like that

src/device/wg_controller.rs
```diff
@@ -905,6 +911,7 @@ fn peer_state(
     time_since_last_rx: Option<&Duration>,
     time_since_last_endpoint_change: Option<&Duration>,
     proxy_endpoint: Option<&SocketAddr>,
+    replaced_proxy_endpoint: Option<&SocketAddr>,
```
matszczygiel, line 914: I wouldn't be passing `SocketAddr` by reference but if you want to be consistent with the code around, then it's somewhat fine

crates/telio-proxy/src/proxy.rs
mathiaspeters: Good suggestion! I was thinking about having a name with `take` in it but couldn't think of something, yet it was so simple!

src/device/wg_controller.rs
mathiaspeters: I passed it by reference to be consistent, but I also tried just now to just deref it and my terminal overflowed with trait bound issues, so I will be keeping it as a ref

matszczygiel (commented):
+1

crates/telio-proxy/src/proxy.rs
```diff
@@ -181,10 +195,27 @@ impl UdpProxy {
         let _ = self.task_egress.stop().await.resume_unwind();
         let _ = self.task_ingress.stop().await.resume_unwind();
     }
+
+    ///
```
packgron, line 199: Add proper rsut doc

crates/telio-proxy/src/proxy.rs
```diff
@@ -283,27 +318,48 @@ impl StateEgress {
...
     async fn send_outbound_data(&mut self, pk: PublicKey, msg: DataMsg) {
-        // if let Some((pk, msg)) = self.input.recv().await {
-        if let (Some((socket, None)), Some(wg_addr)) = (self.sockets.get(&pk), self.wg_addr) {
-            match socket.send_to(msg.get_payload(), wg_addr).await {
-                Ok(_) => {
-                    if self.conn_state.is_err() {
-                        self.conn_state = Ok(());
-                        telio_log_info!("Outbound conn telio -> WG working!");
+        if let Err(err) = self.send_data(pk, msg.clone()).await {
+            let mut should_resend = matches!(
```
packgron, line 322: NIT: you could probably just add method to SocketResentBehaviour like ::should_resend(), reutining bool, stead of doing matches multiple times.

crates/telio-proxy/src/proxy.rs
```diff
@@ -345,10 +405,22 @@ impl StateEgress {
...
+                            e.insert(addr);
+                        }
+                    }
                     self.sockets.insert(pk, (socket, None));
+                    self.num_socket_changes += 1;
+                    return SocketResendBehavior::Send;
                 }
             }
         }
+        SocketResendBehavior::NoSend
```
packgron, line 423: NIT: Moving this into match statement would be nice, it's clearer witch path arrives there.

crates/telio-proxy/src/proxy.rs
```diff
@@ -181,10 +195,27 @@ impl UdpProxy {
         let _ = self.task_egress.stop().await.resume_unwind();
         let _ = self.task_ingress.stop().await.resume_unwind();
     }
+
+    ///
+    pub async fn on_network_change(&self) {
+        let _ = task_exec!(&self.task_egress, async move |state| {
+            state.num_socket_changes = 0;
```
packgron, line 202: Any reason not to reset this on reconfiguration?

Also if we have more then 3 peers, wouldn't a singular no net allow just two be recreated, leaving rest stuck? as these changes are global?

nat-lab/tests/utils/router/linux_router.py
```diff
@@ -429,6 +429,35 @@ async def break_udp_conn_to_host(self, address: str) -> AsyncIterator:
                 "icmp-host-unreachable",
             ]).execute()
 
+    @asynccontextmanager
+    async def block_udp_port(self, port: int) -> AsyncIterator:
```
packgron, line 433: Add documentation for this method, explaining that this does.
In future it will be very unclear, are we blocking incoming outgoing, or what kind of port, and why is this needed

nat-lab/tests/utils/router/mac_router.py
```diff
@@ -142,6 +142,12 @@ async def break_udp_conn_to_host(
     ) -> AsyncIterator:
         yield
 
+    @asynccontextmanager
```
packgron, line 145: Add documentation

nat-lab/tests/utils/router/router.py
```diff
@@ -113,6 +113,11 @@ async def break_tcp_conn_to_host(self, address: str) -> AsyncIterator:
     async def break_udp_conn_to_host(self, address: str) -> AsyncIterator:
         yield
 
+    @abstractmethod
+    @asynccontextmanager
+    async def block_udp_port(self, port: int) -> AsyncIterator:
```
packgron, line 118: add doc

src/device/mod.rs
```diff
@@ -687,7 +687,7 @@ impl Device {
         self.art()?.block_on(async {
             let upstream_servers = upstream_servers.to_vec();
             task_exec!(self.rt()?, async move |rt| {
-                Ok(rt.start_dns(&upstream_servers).await)
+                Ok(Box::pin(rt.start_dns(&upstream_servers)).await)
```
packgron, line 690: NIT: you can use .boxed() from FuturesExt, a bit more ergonomic

src/device/wg_controller.rs
```diff
@@ -151,8 +151,11 @@ async fn consolidate_wg_peers<
     post_quantum_vpn: &impl telio_pq::PostQuantum,
     features: &Features,
 ) -> Result {
-    let proxy_endpoints = if let Some(p) = proxy {
-        p.get_endpoint_map().await?
+    let (proxy_endpoints, replaced_proxy_endpoints) = if let Some(p) = proxy {
+        (
+            p.get_endpoint_map().await?,
+            p.take_replaced_sockets().await?,
```
packgron, line 157: Does the following code ensure, that after this consolidation we will have updated peer to have the new endpoint?

packgron: What happens if we recreate proxy endpoint multiple times before consolidation?

packgron (approved):
We are quite hevily realying on the network change, to make sure we don't get into error trap, let's go through bridge channels and ask if all teams are using network changed api

crates/telio-proxy/src/proxy.rs
mathiaspeters: No reason, just a hasty oversight. Fixed that, and updated the socket limit to be per public key, instead of a single global one

nat-lab/tests/utils/router/mac_router.py
mathiaspeters: For now, I think adding docs in `linux_router.py` is enough since that's the only place it's implemented

crates/telio-proxy/src/proxy.rs
```diff
@@ -283,27 +339,45 @@ impl StateEgress {
...
     async fn send_outbound_data(&mut self, pk: PublicKey, msg: DataMsg) {
-        // if let Some((pk, msg)) = self.input.recv().await {
-        if let (Some((socket, None)), Some(wg_addr)) = (self.sockets.get(&pk), self.wg_addr) {
-            match socket.send_to(msg.get_payload(), wg_addr).await {
-                Ok(_) => {
-                    if self.conn_state.is_err() {
-                        self.conn_state = Ok(());
-                        telio_log_info!("Outbound conn telio -> WG working!");
+        if let Err(err) = self.send_data(pk, msg.clone()).await {
+            let mut should_resend = self.handle_error(err, Some(pk)).await.should_resend();
```
packgron, line 343: So this is still pretty bad, if one socket errors out, assuming others
would be ok, you would be blocked in resend loop.

sleep should happen on wait loops, so update could interup it.

packgron: This would only be ok if each peer would have an individual task.
So the check should probably be more like "can I send to it now"
if not you skip send to this peer. and move out to upper select.

crates/telio-proxy/src/proxy.rs
```diff
@@ -319,36 +359,60 @@ impl StateEgress {
...
-                    Ok(skt) => skt,
-                    Err(err) => {
-                        telio_log_error!("Cannot open socket. {}", err);
-                        return;
+                if let Some(pk) = pk {
+                    if self.peer_socket_backoffs.get_mut(&pk).is_none() {
+                        #[allow(clippy::expect_used)]
+                        let backoff = ExponentialBackoff::new(ExponentialBackoffBounds {
+                            initial: Duration::from_secs(1),
+                            maximal: Some(Duration::from_secs(30)),
```
packgron, line 367: is it neccery to insert / remvoe backoffs ? it feels a bit wastefull.

It should be enough to create on configure, and clear on successful send.

src/device/wg_controller.rs
```diff
@@ -151,8 +151,11 @@ async fn consolidate_wg_peers<
     post_quantum_vpn: &impl telio_pq::PostQuantum,
     features: &Features,
 ) -> Result {
-    let proxy_endpoints = if let Some(p) = proxy {
-        p.get_endpoint_map().await?
+    let (proxy_endpoints, replaced_proxy_endpoints) = if let Some(p) = proxy {
```
packgron, line 154: replace proxy endpoints is not used in is_peer_proxying, 

Maybe we should just replace get_endpoints to return:
HashMap<PublicKey, Vec<SocketAddr>>, assuming first is the one that should be used, and additional are for doing x.contains(sock) for places during verification ?

packgron (approved):
+2.0

packgron (approved):

packgron (approved):

packgron (dismissed):

packgron (approved):

PR #401, opened 2024-03-04
LLT-4873: Remove SWIG FFI bindings

mathiaspeters:
### Problem
We are migrating our FFI bindings from SWIG to UniFFI. The UniFFI bindings are not yet being built and distributed by the CI, but the boilerplate is there and bindings can be generated. We have now chosen to make the breaking change and remove the old SWIG bindings.

### Solution
Remove all checked-in bindings, all SWIG code, all duplication introduced during UniFFI work that kept `main` building as normal

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (commented):
LGTM. Was UniFFI tested already and is fine by apps?

mathiaspeters:
> LGTM. Was UniFFI tested already and is fine by apps?

There is in-progress work to make natlab use UniFFI-generated bindings instead of tcli, and we made a conscious decision to not make `main` be distributable for a bit by removing SWIG now even though UniFFI bindings are not tested or distributed. The natlab migration will be done before anything is distributed to apps though, we got some feedback from apps back when this work was first being done back in June/July, and since it's a new major version, apps will have to deal with breaking changes :smile:

LukasPukenis (commented):
Ideally I would like to see natlab usinf UniFFI before deleting all legacy FFI code so we would have at least some sort of guard against this PR.

Basically if natlab migration to uniffy would become thorny I suspect it might be painful.

My personal take would be to leave this PR unmerged while working on natlab :) 

+1

mathiaspeters:
I understand the concern, but removing the old FFI code really only limits our ability to distribute builds from `main` which we aren't doing until 5.0 anyways. Until then we will do releases from 4.2 and 4.3 release branches. Everything else will work as before

.github/workflows/tests.yml
```diff
@@ -34,24 +34,6 @@ jobs:
           command: test
           args: --all --lib -- --nocapture
 
-  test-ffi:
```
Jauler, line 37: We probably will need to write a similar test, which induces panic and checks if we handle it in libtelio error events somewhere in UniFFI part of things.

@mathiaspeters Could you add this as one of the "things to do" into the UniFFI tests task, so we do not forget about it?

ci/build_libtelio.py
```diff
@@ -35,17 +35,19 @@
...
-        binding_destination = (
-            f"{PROJECT_CONFIG.root_dir}/{LIBTELIO_CONFIG[config.target_os]['binding_dest']}"
-            + telio_bindings.split("/")[-1]
-        )
-
-        if os.path.exists(binding_destination):
-            rutils.remove_tree_or_file(binding_destination)
-
-        rutils.copy_tree_or_file(telio_bindings, binding_destination)
+    # TODO(Mathias): reenable when packaging solution is done
```
Jauler, line 38: Do we have anything more for us to remind about these TODO's than just code comment? E.g. jira item under UniFFI epic or similar?

src/ffi/mod.rs
```diff
@@ -796,991 +746,6 @@ extern "C" {
     fn fortify_source();
 }
 
-#[no_mangle]
```
Jauler, line 799: Ater removing this part, I have not managed to understand how are we handling this FFI-level logic with UniFFI? Could you explain it a little bit?

Jauler:
We will need to build somewhat more generic error type than this enum:
* https://github.com/NordSecurity/libtelio/pull/401/files#diff-8b0fb5533b0a6b5ba21ffa572dd3f6ec1a0e7335eb9524207db1a268dd6d6d0bL20

Jauler:
Could we, instead of calling this `Explicitly` call it `Focribly` here?
https://github.com/NordSecurity/libtelio/pull/401/files#diff-8b0fb5533b0a6b5ba21ffa572dd3f6ec1a0e7335eb9524207db1a268dd6d6d0bR162

Jauler:
Is there a reason to use `[ByRef]` here? https://github.com/NordSecurity/libtelio/pull/401/files#diff-8b0fb5533b0a6b5ba21ffa572dd3f6ec1a0e7335eb9524207db1a268dd6d6d0bR205

.github/workflows/tests.yml
ThrasherLT: Done in [LLT-3475] ticket

src/ffi/mod.rs
ThrasherLT: Added to LLT-5022

PR #403, opened 2024-03-04
Llt 4859 fix aggregation for derp

stalowyjez:
### Problem
Current aggregator API for DERP events is a bit hard to use from Derp module

### Solution
Change the aggregator API for Derp events, so it uses `Server` instead of `AnalyticsEvent`

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/aggregator.rs
+++ b/crates/telio-nurse/src/aggregator.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Borrow, collections::hash_map::Entry, mem, sync::Arc, time::Duration};
+use std::{collections::hash_map::Entry, mem, net::SocketAddr, sync::Arc, time::Duration};
 
 use tokio::{sync::Mutex, time::Instant};
 
@@ -7,7 +7,7 @@ use serde::{ser::SerializeTuple, Serialize};
 use telio_crypto::PublicKey;
 use telio_model::{
     api_config::{EndpointProvider, FeatureNurse},
-    config::{RelayConnectionChangeReason, RelayState, Server},
+    config::{DerpAnalyticsEvent, RelayConnectionChangeReason, RelayState},
     HashMap,
 };
 use telio_utils::telio_log_warn;
@@ -35,9 +35,6 @@ impl From<RelayState> for RelayConnectionState {
         if let RelayState::Connected = value {
             RelayConnectionState::Connected
         } else {
-            if let RelayState::Disconnected = value {
-                telio_log_warn!("Got disconnected state for Relay server");
-            }
             RelayConnectionState::Connecting
         }
     }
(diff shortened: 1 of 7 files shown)


crates/telio-nurse/src/aggregator.rs
```diff
@@ -187,26 +174,32 @@ struct AggregatorData {
     relay_segments: Vec<ConnectionSegmentData>,
 }
 
-// A container to store the connectivity data for our meshnet peers
+/// A container to store the connectivity data for the Meshnet peers used by Nurse
 #[allow(dead_code)]
-pub struct ConnectivityDataAggregator<W: WireGuard> {
+pub struct ConnectivityDataAggregator {
```
mathiaspeters, line 179: Not arguing for or against either option (yet), just curious, why remove the generic argument in favor of `dyn`?

tomaszklak:
Can you clarify reasoning for this change? The `Server` seems to be much bigger struct with lots of data not needed for the task at hand - why do we want to expose this information to the aggregator?

stalowyjez:
@tomaszklak the main cause is that we have no `AnalyticsEvent` structs ready in Derp, but we have `Server` and it is already used to send data from `telio-relay` to Nurse (so we can even reuse it, or maybe move that NAT-discovering logic to the aggregator). Also `Server` is a structure from `telio-model`, which kinda makes it a part of common language between different modules, so I guess it makes some sense to use it. I can add some custom struct for that if you think that would be better, but the main point was to remove `AnalyticsEvent` from the API.

tomaszklak:
> @tomaszklak the main cause is that we have no `AnalyticsEvent` structs ready in Derp, but we have `Server` and it is already used to send data from `telio-relay` to Nurse (so we can even reuse it, or maybe move that NAT-discovering logic to the aggregator). Also `Server` is a structure from `telio-model`, which kinda makes it a part of common language between different modules, so I guess it makes some sense to use it. I can add some custom struct for that if you think that would be better

Yes, it that case I think a custom struct with a minimal set of attributes would be better.

crates/telio-nurse/src/aggregator.rs
matislovas: A very good thing :+1:  no need to propagate the generic argument up to the top structure.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -15,49 +16,37 @@ use telio_wg::{
...
-    fn from(value: PeerState) -> Self {
-        if let PeerState::Connected = value {
+impl From<RelayState> for RelayConnectionState {
+    fn from(value: RelayState) -> Self {
+        if let RelayState::Connected = value {
             RelayConnectionState::Connected
         } else {
-            if let PeerState::Disconnected = value {
+            if let RelayState::Disconnected = value {
                 telio_log_warn!("Got disconnected state for Relay server");
```
tomaszklak, line 39: I'm not sure a warning (or any log for that matter) whenever a conversion is done is a good idea.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -76,94 +65,107 @@ impl From<EndpointProvider> for EndpointType {
...
+        seq.end()
     }
 }
 
 // Ready to send connection data for some period of time
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
-pub struct ConnectionSegmentData {
-    node: PublicKey,
+pub(crate) struct RelayConnDataSegment {
+    node: SocketAddr,
```
tomaszklak, line 127: nit: to make it more consistent maybe use the same name here (`server_address`)?

crates/telio-nurse/src/aggregator.rs
```diff
@@ -222,9 +222,15 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
         }
     }
 
-    #[allow(dead_code)]
-    pub async fn change_peer_state_direct(
-        &mut self,
+    /// Inform Nurse about direct peer state change.
```
tomaszklak, line 225: I think this is not informing nurse, but recording a change (so that in future nurse can be notified about it).

crates/telio-nurse/src/aggregator.rs
```diff
@@ -239,8 +245,12 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
         .await
     }
 
-    #[allow(dead_code)]
-    pub async fn change_peer_state_relayed(&mut self, event: &AnalyticsEvent) {
+    /// Inform Nurse about relayed peer state change.
```
tomaszklak, line 248: Same, not really informing nurse just yet

crates/telio-nurse/src/aggregator.rs
```diff
@@ -251,49 +261,49 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
...
         data_guard.peer_segments.extend(new_segment);
     }
 
-    #[allow(dead_code)]
-    pub async fn change_relay_state(
-        &mut self,
-        event: &AnalyticsEvent,
-        reason: RelayConnectionChangeReason,
-    ) {
+    /// Inform Nurse about relay server state change.
```
tomaszklak, line 300: Same :)

crates/telio-nurse/src/aggregator.rs
stalowyjez: Well, I thought about the whole module, but maybe it is not the best wording ^^'

mathiaspeters:
+1

tomaszklak (approved):
+1

PR #404, opened 2024-03-04
LLT_4954 handle sleeping state

Hasan6979:
### Problem
NAT and DERP data aggregator needs to be notified when the system is going to sleep, and when it wakes up in order to not take those timings into consideration.

 - a small bug in the aggregator was also found and fixed.

### Solution
When the system goes to sleep, simply gather all the unacknowledged segments and clear the rest.
At wakeup, do an additional clearing of current segments to handle the race condition where a new segment might begin after clearing the segments when system goes to sleep. The beginning of new segment would be handled by wg notifications.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/aggregator.rs
+++ b/crates/telio-nurse/src/aggregator.rs
@@ -287,7 +287,7 @@ impl ConnectivityDataAggregator {
 
                 Some(new_segment)
             }
-            Entry::Vacant(entry) => {
+            Entry::Vacant(entry) if event.peer_state == PeerState::Connected => {
                 entry.insert((event.clone(), endpoints));
                 None
             }
@@ -327,11 +327,24 @@ impl ConnectivityDataAggregator {
         }
     }
 
-    #[allow(dead_code)]
-    pub(crate) async fn collect_unacknowledged_segments(
-        &self,
-        force_close: bool,
-    ) -> AggregatorCollectedSegments {
+    /// Clear current peer and relay monitoring segments
+    pub async fn clear_ongoinging_segments(&self) {
+        let mut data_guard = self.data.lock().await;
+        data_guard.current_peer_events.clear();
+        data_guard.current_relay_event.take();
+    }
+
+    /// Force saves unacknowledged segments
+    pub async fn force_save_unacknowledged_segments(&self) {
+        self.save_unacknowledged_segments(true).await;
+    }
+
+    /// Saves unacknowledged segments in a buffer
+    ///
+    /// # Arguments
+    ///
+    /// * `force_save` - Whether to force save current events
+    async fn save_unacknowledged_segments(&self, force_save: bool) {
         // Getting wg_interface is first, so it won't await while mutex is locked
         let wg_interface = self.wg_interface.get_interface().await;
         let mut data_guard = self.data.lock().await;
@@ -344,22 +357,12 @@ impl ConnectivityDataAggregator {
             .current_relay_event
             .as_mut()
             .filter(|relay_event| {
-                current_timestamp.duration_since(relay_event.timestamp) > DAY_DURATION
-                    || force_close
+                current_timestamp.duration_since(relay_event.timestamp) > DAY_DURATION || force_save
             })
         {
             new_relay_segments.push(RelayConnDataSegment::new(event, current_timestamp));
             event.timestamp = current_timestamp;
         }
-
-        if force_close {
-            data_guard.current_relay_event.take();
-        }
-
-        data_guard.relay_segments.extend(new_relay_segments);
-        data_guard
-            .relay_segments
-            .sort_by(|a, b| a.start.cmp(&b.start));
 
         // Handle long-lasting peer events
         let mut new_peer_segments = Vec::new();
@@ -367,7 +370,7 @@ impl ConnectivityDataAggregator {
             for (peer_event, peer_state) in data_guard.current_peer_events.values_mut() {
                 let since_last_event =
                     current_timestamp.duration_since(Instant::from_std(peer_event.timestamp));
-                if since_last_event > DAY_DURATION || force_close {
+                if since_last_event > DAY_DURATION || force_save {
                     wg_peers
                         .get(&peer_event.public_key)
                         .and_then(|wg_peer| wg_peer.rx_bytes.zip(wg_peer.tx_bytes))
@@ -394,17 +397,28 @@ impl ConnectivityDataAggregator {
             telio_log_warn!("Getting wireguard interfaces failed");
         }
 
-        if force_close {
+        if force_save {
+            data_guard.current_relay_event.take();
             data_guard.current_peer_events.clear();
         }
 
+        data_guard.relay_segments.extend(new_relay_segments);
+        data_guard
+            .relay_segments
+            .sort_by(|a, b| a.start.cmp(&b.start));
         data_guard
             .peer_segments
             .extend(new_peer_segments.into_iter());
         data_guard
             .peer_segments
             .sort_by(|a, b| a.start.cmp(&b.start));
+    }
 
+    /// Collects unacknowledged segments
+    #[allow(dead_code)]
+    pub(crate) async fn collect_unacknowledged_segments(&self) -> AggregatorCollectedSegments {
+        self.save_unacknowledged_segments(false).await;
+        let mut data_guard = self.data.lock().await;
         AggregatorCollectedSegments {
             relay: mem::take(&mut data_guard.relay_segments),
             peer: mem::take(&mut data_guard.peer_segments),
@@ -533,18 +547,14 @@ mod tests {
         current_relay_event.state = RelayState::Disconnected;
         aggregator.change_relay_state(current_relay_event).await;
 
-        let segments = aggregator
-            .collect_unacknowledged_segments(false)
-            .await
-            .relay;
-
+        let segments = aggregator.collect_unacknowledged_segments().await.relay;
         assert_eq!(segments.len(), 4);
 
         assert_eq!(
             segments[0],
             RelayConnDataSegment {
                 server_address: first_serv_addr,
-                start: segment_start.into(),
+                start: segment_start,
                 duration: first_segment_length1 + first_segment_length2,
                 connection_data: RelayConnectionData {
                     state: RelayConnectionState::Connected,
@@ -556,7 +566,7 @@ mod tests {
             segments[1],
             RelayConnDataSegment {
                 server_address: second_serv_addr,
-                start: second_segment_start.into(),
+                start: second_segment_start,
                 duration: second_segment_length,
                 connection_data: RelayConnectionData {
                     state: RelayConnectionState::Connected,
@@ -568,7 +578,7 @@ mod tests {
             segments[2],
             RelayConnDataSegment {
                 server_address: second_serv_addr,
-                start: third_segment_start.into(),
+                start: third_segment_start,
                 duration: third_segment_lenght,
                 connection_data: RelayConnectionData {
                     state: RelayConnectionState::Connecting,
@@ -580,7 +590,7 @@ mod tests {
             segments[3],
             RelayConnDataSegment {
                 server_address: first_serv_addr,
-                start: fourth_segment_start.into(),
+                start: fourth_segment_start,
                 duration: fourth_segment_length,
                 connection_data: RelayConnectionData {
                     state: RelayConnectionState::Connected,
@@ -589,10 +599,7 @@ mod tests {
             }
         );
 
-        let segments = aggregator
-            .collect_unacknowledged_segments(false)
-            .await
-            .relay;
+        let segments = aggregator.collect_unacknowledged_segments().await.relay;
         assert_eq!(segments.len(), 0);
     }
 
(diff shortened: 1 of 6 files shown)


src/ffi/mod.rs
```diff
@@ -1301,6 +1301,30 @@ pub extern "C" fn telio_notify_network_change(
     })
 }
 
+#[no_mangle]
```
Jauler, line 1304: Do not forget to Update `src/libtelio.udl` as well ;). 

Note that at least theoretically one should also remove it from swig bindings, but those are going to be removed soon, so probably no point in updating them anymore, as `main` is now on the track for v5.0.0 release, which will be UniFFI only.

crates/telio-nurse/src/nurse.rs
```diff
@@ -184,6 +202,24 @@ impl State {
         .await;
     }
 
+    pub async fn handle_sleep(&self) {
+        let _ = task_exec!(&self.heartbeat, async move |_state| {
+            // TODO(Hasan): Uncomment after segment timing is handled
+            // state.aggregator.collect_unacknowledged_segments(true);
```
Jauler, line 208: Adding a reminder github thread for TODO item ;)

crates/telio-nurse/src/aggregator.rs
```diff
@@ -322,11 +330,18 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
         }
     }
 
+    #[allow(dead_code)] //TODO(Hasan) : Remove this!
```
Jauler, line 333: Adding a reminder github thread for TODO item ;)

crates/telio-nurse/src/aggregator.rs
```diff
@@ -277,12 +279,18 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
...
         };
 
-        if event.peer_state != PeerState::Connected {
-            if let Entry::Occupied(entry) = event_entry {
-                entry.remove();
+        match event_entry {
+            Entry::Occupied(mut entry) => {
+                if event.peer_state != PeerState::Connected {
+                    entry.remove();
+                } else if update_entry {
```
Jauler, line 286: A quick question, while reading, I was not fully able to understand why "update_entry" flag is needed. Could you explain it? :pray:

Hasan6979: So we have 3 cases.
1) When there is already an entry, but now peer isn't connected so we remove it.
2) When there is already an entry, but now the connection state has changed so we need to overwrite the existing entry. The problem why `update_entry` was needed was because if there was already an entry, and peer state was still connected then the entry wasn't being updated.
3) When there is no entry then make one.

crates/telio-nurse/src/aggregator.rs
Hasan6979: Have already created a jira item and assigned myself to it. :)

src/ffi/mod.rs
Hasan6979: Done!

crates/telio-nurse/src/aggregator.rs
```diff
@@ -322,11 +330,18 @@ impl<W: WireGuard> ConnectivityDataAggregator<W> {
         }
     }
 
+    #[allow(dead_code)] //TODO(Hasan) : Remove this!
+    pub async fn clear_segments(&self) {
+        let mut data_guard = self.data.lock().await;
+        data_guard.current_peer_events.clear();
+        data_guard.current_relay_event.take();
+    }
+
```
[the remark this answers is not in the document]
Hasan6979: This is like a failsafe in case of race conditions where we get the notification to sleep, but before system actually goes to sleep there is a change in state and for the whole duration of the sleep that state is taken instead.

crates/telio-nurse/src/nurse.rs
```diff
@@ -184,6 +202,24 @@ impl State {
         .await;
     }
 
+    pub async fn handle_sleep(&self) {
+        let _ = task_exec!(&self.heartbeat, async move |_state| {
```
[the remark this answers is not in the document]
Hasan6979: I had a look at https://github.com/NordSecurity/libtelio/pull/396 and that is how it was done, so just kept it as it was.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -957,10 +1053,145 @@ mod tests {
...
+            )
+            .await;
+
+        time::advance(Duration::from_secs(800)).await;
+
+        // Sleep is called
+        aggregator.collect_unacknowledged_segments(true).await;
+
+        let segments = aggregator
+            .collect_unacknowledged_segments(false)
```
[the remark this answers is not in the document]
Hasan6979: Mm. I think refactoring `collect_unacknowledged_segments` makes sense because from a third person perspective calling `collect_unacknowledged_segments(true)` seems like you are collecting the segments, and well vice versa.

I feel that moving the `force_close` and `24hr cycle` should be moved out of this function and this function just calls that. When we go to sleep, we just `save` the segments. I mean, right now by just reading the line it isn't 100% what will happen.

src/device/mod.rs
```diff
@@ -1436,6 +1448,20 @@ impl Runtime {
         Ok(())
     }
 
+    async fn notify_sleep(&mut self) -> Result {
+        if let Some(nurse) = &self.entities.nurse {
+            nurse.handle_sleep().await;
+        }
```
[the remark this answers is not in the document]
Hasan6979: Events from WG should take care of that IMO.
@Jauler  what is your opinion ?

crates/telio-nurse/src/aggregator.rs
stalowyjez: @Hasan6979 - well, that was intended, when the state in the event is the same as the current state, there is no need to start another segment

Hasan6979: Yes. I agree. That is where the `update_entry` comes into play. When states are different, and there was already an entry in place.

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
stalowyjez: Ok, so if you want to just save them, I guess the best option is to move it out to some segment-saving function (Probably some common private one and separate public wrappers for segment collection and storing them? Maybe it could take a predicate which events should be used for segment creation?)

crates/telio-nurse/src/aggregator.rs
stalowyjez: Uhhh, that's right, that was a mistake - but I guess using an additional mutable boolean flag isn't that nice, could you just use `new_segment.is_come()` instead?

crates/telio-nurse/src/aggregator.rs
```diff
@@ -649,9 +670,46 @@ mod tests {
...
 
+        // Send one more event with the different state
+        let second_segment_start = current_peer_event.timestamp;
+        aggregator
+            .change_peer_state_direct(
+                &current_peer_event,
+                EndpointProvider::Upnp,
+                EndpointProvider::Upnp,
+            )
+            .await;
```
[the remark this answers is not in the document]
Hasan6979: Ah. I misunderstood the purpose. I thought this test checks multiple segments. I will create another test which checks multiple events being saved.

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Hasan6979: Yep. Thats what I am thinking

crates/telio-nurse/src/nurse.rs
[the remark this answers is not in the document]
stalowyjez: Ok, se we can leave it like that

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Hasan6979: New test added!

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Hasan6979: Abstracted saving segments from collecting segments

crates/telio-nurse/src/aggregator.rs
Hasan6979: Because there was a lot of code to be added that was kind of being duplicated (Being already added in https://github.com/NordSecurity/libtelio/pull/396)

Hasan6979: And well to avoid merge conflicts because there are some other changes as well made in https://github.com/NordSecurity/libtelio/pull/396 later on that I haven't made

src/ffi/mod.rs
```diff
@@ -546,6 +546,28 @@ impl telio {
         })
     }
 
+    /// Notify telio system is going to sleep.
+    pub fn notify_sleep(&self) -> FFIResult<()> {
+        #![allow(unused_variables)]
```
Jauler, line 551: Is this `allow` needed? As it looked like unintentional leftover :)

crates/telio-nurse/src/aggregator.rs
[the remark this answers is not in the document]
Jauler: > If you call the other function on sleep, this shouldn't be needed probably?


Yeah, to add a little bit more, we were discussing about situation like this:
```
  |-------------------------xxxxx| zzzzzzzzzzzz |xxxx------------------------------|
start                         ^                                        ^
                       sleep notification
```
The sleep and wake notification necessarily have to be called some time before actual sleep, and wakeup notification some time after actual sleep. Therefore there are these periods (I marked them with `x`), which is this "dangerous territory" where an event change would produce incorrect events.

While clearing at the wakeup notification is not ideal situation -> it is better than nothing.

crates/telio-nurse/src/nurse.rs
```diff
@@ -184,6 +202,24 @@ impl State {
...
+            // state.aggregator.collect_unacknowledged_segments(true);
+            Ok(())
+        })
+        .await;
+    }
+
+    pub async fn handle_wakeup(&self) {
+        let _ = task_exec!(&self.heartbeat, async move |_state| {
+            // TODO(Hasan): Uncomment after segment timing is handled
+            // state.aggregator.clear_segments();
```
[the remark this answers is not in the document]
Hasan6979: Uncommented!

src/device/mod.rs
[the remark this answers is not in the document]
Jauler: Generally all data is affected by sleeps, and we do have some incorrect reporting for older Quality Of Service metrics. But I believe it will be more efficient to address it with channel refactoring in LLT-5018 instead if using this PR.

crates/telio-nurse/src/aggregator.rs
Jauler: If I am not mistaken, this dead_code annotation is not needed anymore?

crates/telio-nurse/src/aggregator.rs
```diff
@@ -391,17 +394,33 @@ impl ConnectivityDataAggregator {
...
             .sort_by(|a, b| a.start.cmp(&b.start));
+    }
 
+    /// Record a relay server state change
+    ///
+    /// # Arguments
+    ///
+    /// * `server_state` - Current derp server state.
+    /// * `reason` - Reason of the state change.
+    #[allow(dead_code)]
```
Jauler, line 420: If I am not mistaken this dead_code annotation is not required anymore?

crates/telio-nurse/src/config.rs
```diff
@@ -146,3 +151,23 @@ impl Default for QoSConfig {
...
+
+impl AggregatorConfig {
+    /// Create a new Aggregator config
+    pub fn new(features: &FeatureNurse) -> Self {
+        Self {
+            relay_events: features.enable_relay_conn_data.unwrap_or(false),
+            nat_traversal_events: features.enable_nat_traversal_conn_data.unwrap_or(false),
+        }
+    }
+}
```
stalowyjez, line 173: Well, with the current design Nurse doesn't own Aggregator, so if you insist on adding a separate struct for aggregator config, I guess it would be better in the aggregator.rs

Hasan6979: I have added in this way so that there is less divergence from https://github.com/NordSecurity/libtelio/pull/396 where this is also being added.

Jauler (dismissed):
+1.0

crates/telio-nurse/src/aggregator.rs
```diff
@@ -327,11 +327,19 @@ impl ConnectivityDataAggregator {
...
+        data_guard.current_peer_events.clear();
+        data_guard.current_relay_event.take();
+    }
+
+    /// Saves unacknowledged segments in a buffer
+    ///
+    /// # Arguments
+    ///
+    /// * `force_save` - Whether to force save current events
+    pub async fn save_unacknowledged_segments(&self, force_save: bool) {
```
stalowyjez, line 342: Imho it would be better to make it private and create a public function without arguments

crates/telio-nurse/src/aggregator.rs
```diff
@@ -420,7 +429,7 @@ mod tests {
     };
 
     use csv::WriterBuilder;
-    use telio_model::config::Server;
+    use telio_model::{api_config::FeatureNurse, config::Server};
```
stalowyjez, line 432: Hmm, do you need to include this `FeatureNurse` struct again?

stalowyjez (commented):
Looks quite ok, +1

stalowyjez (approved):

PR #407, opened 2024-03-05
Fix .gitignore

jjanowsk:
Don't match any directory in repo's base directory 
Match directories starting with dot in repo's base directory


Jauler (approved):
+2.0

PR #409, opened 2024-03-05
LLT-5002: Preshared key comparison detect's non exsistant state change

matszczygiel:
### Problem
The zero-valued preshared key is reported for certain interfaces

### Solution
Validate the key

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -18,6 +18,7 @@
 * LLT-4980: Introduce synchronization for the endpoint map in telio-proxy
 * LLT-4970: PQ VPN is enabled by calling `telio_connect_to_exit_node_postquantum()`
 * LLT-4981: Add upgrade decision message
+* LLT-5002: Fix preshared key parsing logic
 
 <br>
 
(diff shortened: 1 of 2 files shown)


nat-lab/tests/test_node_state_flickering.py
```diff
@@ -3,6 +3,7 @@
 import telio
 from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_mesh_nodes
+from telio_features import TelioFeatures, Direct
```
tomaszklak, line 6: Do I understand correctly that the tests in this file would fail now (with direct enabled), without the changes in `uapi.rs`?

matszczygiel: Yeah, that's the premise. However, it seems to me that it might not be the case :shrug: and I'm still trying to confirm that.
However, the preshared keys parsing logic was flawed and needs to be fixed

tomaszklak (commented):
+1, seems fine, but if possible please add at least one test to verify this

packgron (approved):
+1.0

PR #410, opened 2024-03-06
Add NAT-Lab test for checking if there are any leftover WireGuard

ThrasherLT:
Check for leftover peers after setting mesh to off.

### Problem
In case of a bug, there may be leftover peers after turning mesh off and that can cause issues and even lead to IP leaks.

### Solution
Added a NAT-Lab test which checks if there are any leftover peers after turning mesh off.

Also note since this is an internal test commit, not adding any entry to the changelog as this information is not useful to app teams who are the consumers of the changelog.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_mesh_off.py
+++ b/nat-lab/tests/test_mesh_off.py
@@ -53,8 +53,8 @@ async def test_mesh_off(
             "tun10",
         ]).execute()
 
-        dig_stdout = process.get_stdout()
+        wg_show_stdout = process.get_stdout()
 
         assert (
-            "peer:" not in dig_stdout.strip().split()
-        ), f"There are leftover WireGuard peers after mesh is set to off: {dig_stdout}"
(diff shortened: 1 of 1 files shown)


Jauler (dismissed):
+1.0

nat-lab/tests/test_mesh_off.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Checking if no peer is left after turning mesh net off
+        # wg show outputs lines that start with the string "peer:" when any peer is present
+        process = await connection_alpha.create_process([
+            "wg",
+            "show",
+            "tun10",
+        ]).execute()
+
+        dig_stdout = process.get_stdout()
```
Jauler, line 56: nit:
`dig_stdout` is probably leftover copied name, we should name this somewhat differently, e.g. `wg_show_stdout`

PR #416, opened 2024-03-07
Fix wireguard go oscillation

packgron:
### Problem

### Solution


crates/telio-wg/src/wg.rs
```diff
@@ -292,7 +296,26 @@ impl WireGuard for DynamicWg {
...
+                    telio_log_debug!("Wireguard using port:{:?}", &port.to_string());
+                    break Ok(port);
+                }
+            }
+            let elapsed_time = start.elapsed()?;
+            if elapsed_time > d {
+                break telio_err_with_log!(Error::PortAssignmentTimeoutError);
+            }
+
+            sleep(Duration::from_millis(10)).await;
```
matszczygiel, line 308: Wow, that's quite hacky

matszczygiel (dismissed):
+1

crates/telio-wg/src/wg.rs
packgron: Yeh, sadly there is no signal from wireguard that listen port was set :cry:

gytsto:
test?

packgron:
> test?

Yeh will spend some time today, was expecting windows qa to pass it and merge it, but looks like that will not happen so can spend some time on cleantup

crates/telio-wg/src/wg.rs
```diff
@@ -292,7 +296,26 @@ impl WireGuard for DynamicWg {
...
+        }
+    }
+
+    async fn wait_for_proxy_listen_port(&self, d: Duration) -> Result<u16, Error> {
+        let start = std::time::SystemTime::now();
+        loop {
+            let ifa = self.get_interface().await?;
```
stalowyjez, line 315: What is ifa?  Maybe some smol comment on that? Just "Interface A", or maybe it has some smarter meaning?

stalowyjez (commented):
Looks ok, +1

crates/telio-wg/src/wg.rs
gytsto: I think it is shorter variant of `iface`, which is shorter variant of `interface` :kekw:

packgron: changed to interface, ifa is common shortening in kernel :sweat:

gytsto (approved):
+1

PR #419, opened 2024-03-08
[LLT-4859] Report DERP changes to aggregator

stalowyjez:
### Problem
We need to report data about Derp connectivity changes

### Solution
This commit adds aggregator pointer to the Derp module and uses it to report Derp connectivity events

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/aggregator.rs
+++ b/crates/telio-nurse/src/aggregator.rs
@@ -113,7 +113,7 @@ impl Serialize for RelayConnectionData {
         let RelayConnectionData { state, reason } = self;
         let mut seq = serializer.serialize_tuple(RELAY_TUPLE_LEN)?;
         seq.serialize_element(&(*state as u64))?;
-        seq.serialize_element(&(*reason as u64))?;
+        seq.serialize_element::<u64>(&(Into::into(*reason)))?;
         seq.end()
     }
 }
@@ -201,7 +201,7 @@ pub(crate) struct AggregatorCollectedSegments {
 impl ConnectivityDataAggregator {
     /// Create a new DataConnectivityAggregator instance
     pub fn new(
-        nurse_features: &FeatureNurse,
+        nurse_features: Option<FeatureNurse>,
         wg_interface: Arc<dyn WireGuard>,
     ) -> ConnectivityDataAggregator {
         ConnectivityDataAggregator {
@@ -211,9 +211,12 @@ impl ConnectivityDataAggregator {
                 peer_segments: Vec::new(),
                 relay_segments: Vec::new(),
             }),
-            aggregate_relay_events: nurse_features.enable_relay_conn_data.unwrap_or(false),
+            aggregate_relay_events: nurse_features
+                .as_ref()
+                .and_then(|features| features.enable_relay_conn_data)
+                .unwrap_or(false),
             aggregate_nat_traversal_events: nurse_features
-                .enable_nat_traversal_conn_data
+                .and_then(|features| features.enable_nat_traversal_conn_data)
                 .unwrap_or(false),
             wg_interface,
         }
@@ -411,7 +414,10 @@ impl ConnectivityDataAggregator {
 
 #[cfg(test)]
 mod tests {
-    use std::net::{IpAddr, Ipv4Addr};
+    use std::{
+        io::ErrorKind,
+        net::{IpAddr, Ipv4Addr},
+    };
 
     use csv::WriterBuilder;
     use telio_model::config::Server;
@@ -439,7 +445,7 @@ mod tests {
             enable_nat_traversal_conn_data: Some(enable_nat_traversal_conn_data),
         };
 
-        ConnectivityDataAggregator::new(&nurse_features, wg_interface)
+        ConnectivityDataAggregator::new(Some(nurse_features), wg_interface)
     }
 
     fn create_basic_peer_event(n: u8) -> AnalyticsEvent {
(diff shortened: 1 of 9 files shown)


dfetti (commented):
+1.0

crates/telio-relay/src/derp/mod.rs
```diff
@@ -250,10 +270,19 @@ impl State {
                     Ok(conn) => {
                         telio_log_info!("({}) Connected to {}", Self::NAME, server.get_address());
                         server.conn_state = RelayState::Connected;
+                        if let Some(aggregator) = aggregator.as_ref() {
```
matislovas, line 273: IMHO that's not a best place to place `aggregator`. I think we should sync this up with my [PR](https://github.com/NordSecurity/libtelio/pull/396). Because in this strategy there will be two aggregators (one for NAT traversal and one for DERP) and one instance can actually hold information for both cases. Maybe we need a channel, that goes down to `heartbeat` (Analytics) instance, and there the information gathering will take place ..?

stalowyjez: Nooo, aggregator shouldn't be private. This is just the right place to create it, just pass the shared pointer to Nurse.

matislovas: I think we need to sync up this part with @tomaszklak .

crates/telio-nurse/src/aggregator.rs
```diff
@@ -186,7 +192,7 @@ pub struct ConnectivityDataAggregator {
     aggregate_relay_events: bool,
     aggregate_nat_traversal_events: bool,
 
-    wg_interface: Arc<dyn WireGuard>,
+    wg_interface: Weak<dyn WireGuard>,
```
tomaszklak, line 195: Why is this change made? Up until today we avoided using weak pointers in favor of explicitly controlling the order of destruction of various things (eg. `MeshnetEntities::stop`).

crates/telio-relay/src/derp/mod.rs
```diff
@@ -115,6 +120,10 @@ struct State {
     derp_poll_session: Session,
     /// Cache the result of derp polling
     remote_peers_states: PeersStatesMap,
+    /// Connectivity data aggregator
+    aggregator: Option<Arc<ConnectivityDataAggregator>>,
```
tomaszklak, line 124: I think it would be better to extract an interface out of `ConnectivityDataAggregator` and provide a no-op implementation for it, so that we can avoid checking such optionals each time new derp/nat event is to be reported.

src/device/mod.rs
```diff
@@ -246,6 +246,8 @@ pub struct Entities {
     // Nurse
     nurse: Option<Arc<Nurse>>,
 
+    aggregator: Option<Arc<ConnectivityDataAggregator>>,
```
tomaszklak, line 249: The comment above about no-op implementation also applies here

crates/telio-relay/src/derp/mod.rs
tomaszklak: I'm confused - aggregator is not constructed here, but in the `device/mod.rs` as part of the `Entieties` 🤔

stalowyjez: Oh, yeah, I saw `mod.rs` and I thought @matislovas talks about the creation inside `device/mod.rs` ^^'

matislovas: I was talking about storing `aggregator` inside `heartbeat.rs` Analytics struct.

stalowyjez: @matislovas Yeah, I got that, we probably don't want to do that

src/device/mod.rs
stalowyjez: The thing is that Derp may not always need to use Aggregator - e.g. I think there is no point in adding aggregator to derp in tcli imho

crates/telio-relay/src/derp/mod.rs
stalowyjez: I guess I can just use the flags which are already present flag for that?

crates/telio-nurse/src/aggregator.rs
stalowyjez: Well, I thought that it makes sense that aggregator won't keep wg interface alive, but I can remove it explicitely

crates/telio-relay/src/derp/mod.rs
tomaszklak: Maybe, I'm not exactly sure what you have in mind. But overall I think it's probably better to do one conditional and either create a real aggregator or a no-op one and than use it unconditionally everywhere (derp, nat, nurse reporting) instead of using optional/flags in all those places each time you need to call a method on the aggregator.

stalowyjez: Hmm, I don't like that idea too much tbh, sounds like an overengineering to me

crates/telio-nurse/src/aggregator.rs
stalowyjez: While I think this particular dependency is kinda weak, idk whether that's the case in the other places, so you can add some task if you are convinced, but I'm not entirely convinced that it is a good idea to change it everywhere

src/device/mod.rs
stalowyjez: Well, the derp inside Telio is normally created and I think we want to test aggregator with it, I thought about the separate derp instance, which isn't really connected to Nurse or anything, so I don't think we could use it for aggregation tests even if we wanted

crates/telio-relay/src/derp/mod.rs
Jauler: In this case, I think we have only two call-sites call into aggregator from DERP module. So the "saving" on if's are not super big. But I I do agree, that lower number of conditionals makes things simpler. Additionally exported trait allows for easier mock implementations in unit tests.

Jauler: Oh, but aggregator is already mocked in unit tests. so nvm, about mocking.

Jauler (approved):
+1.0

PR #421, opened 2024-03-08
PMTU tests are flaky most probably due to the nature of docker networ…

LukasPukenis:
This needs more investigation but for now it blocks the CI

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_pmtu.py
+++ b/nat-lab/tests/test_pmtu.py
@@ -1,204 +1,204 @@
-import config
-import pytest
-from contextlib import AsyncExitStack
-from helpers import SetupParameters, setup_environment, setup_connections
-from telio import AdapterType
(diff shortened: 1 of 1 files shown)


matislovas:
I would vote to changing the filename to `_test_pmtu.py`. 

On the other hand, shouldn't we revert PR/commit, which introduces these tests instead ..?

LukasPukenis:
The test should have been merged alongside the feature so reverting would mean reverting the whole thing. The flakyness which I only now observed might be due to the docker networking intricacies.

Does renaming with underscore prefix skip it?

gytsto:
Add pytest.mark.skip("<jira ticket id for fixing test>"). It will disable test.

matszczygiel:
@LukasPukenis does this https://github.com/NordSecurity/libtelio/pull/422 solve the problem?

PR #422, opened 2024-03-11
LLT-5010: Fix PMTU detection by exhausting the error queue before reading the k…

matszczygiel:
The kernel's behavior is not consistent between its versions. This commit exhausts the ICMP socket error queue before reading the estimated PMTU value and handles the case it's reported as 0.

### Problem
The PMTU tests fail on runners with various Linux kernel version

### Solution
The problem seems to be not matching the error response with the proper outgoing packet, so matching it properly should solve the problem

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-pmtu/src/linux.rs
+++ b/crates/telio-pmtu/src/linux.rs
@@ -1,3 +1,20 @@
+//! This module implements PMTU probing for Linux.
+//!
+//! It does it in the following way:
+//! * The kernel has an estimate of the PMTU, we read it to get a guess of our starting value
+//! * We perform the binary search of PMTUs within the `super::PMTU_RANGE` interval
+//! * we check the specific PMTU value by sending an ICMP Echo Request of a length equal to PMTU and await the response
+//! * there are the following possibilities:
+//!   - the response arrives out of order,
+//!   - the response is randomly dropped,
+//!   - the response is dropped because the true PMTU is lower than our guess,
+//!   - or the error message `Fragmentation needed` arrives with or without `Next hop MTU`,
+//!   - the proper response arrives indicating that our guess is lower or equal to the true PMTU
+//!
+//! There are some caveats though, documented in the relevant places in the code:
+//! * the kernel does some things in an unpredicted way, for example `send()` and `recv()` calls spuriously report errors when sending packets larger than the kernel's PMTU guess
+//! * the kernel can keep multiple error message
+
 use std::{ffi::c_int, io, net::IpAddr, os::fd::AsRawFd, time::Duration};
 
 use pnet_packet::{icmp, icmpv6, Packet};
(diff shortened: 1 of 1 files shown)


crates/telio-pmtu/src/linux.rs
```diff
@@ -178,6 +198,7 @@ trait IcmpPkg: Sized {
     fn advance_seq(&mut self);
     fn seq(&self) -> u16;
     fn bytes(&self, pmtu: u32) -> &[u8];
+    fn extract_seq(pkg: &[u8]) -> io::Result<u16>;
```
LukasPukenis, line 201: why do we use word `sequence` for mtu calculations? Personally it confuses me a lot when viewing the PR. If it's u16 it's basically a mtu guess but sequence tells me it's something like an iterator :confused:

gytsto (commented):
+1

crates/telio-pmtu/src/linux.rs
LukasPukenis: As far as I understand the PMTU calculation should happen as follows:
- we have initial PMTU guess
- we try the PMTU guess and see if we receive a _fragmentation needed_ error
- if there was frag_needed we then reduce the guesstimate, else we increase it
- loop to the first point until we find out or exhaust

As an optimisation there's also reading from the kernel which is hinting the best guess for pmtu.

But when reading the code I am quite.. confused I must say, it's abstractions and functions all over the place and I feel like having a chase in a maze :).

My suggestion is to simplify(flatten and reduce abstractions) the code and make it easier to follow if that's possible.

LukasPukenis (commented):
Shouldn't this PR only include additional check against zero value to fix the issue?

crates/telio-pmtu/src/linux.rs
matszczygiel: So, it's not that simple because:
* you might not receive the fragmentation needed response
* fragmentation needed might not contain `next hop MTU`
* the kernel guess might not be accurate
* socket errors are a mess, sometimes they are just thrown out of a blue sky
* ICMP messages are unordered so you might receive a response out of order

The `seq` is the ICMP Echo Request/Response sequence number.

The abstraction is here just to share the code between IPv4 and IPv6 since we need to support both.

If you see this code complicated then I'd like to hear the suggestion on how to simplify it. But keep in mind that this PMTU discovery is quite a complicated problem by itself to be solved and the complexity is passed down the code.

matszczygiel: So this PR adds the following things:
* validate response `seq` number
* exhaust the error queue so that we care only about the last error
* guard against `next hop MTU` being 0 on error message

LukasPukenis: As we discussed about the problem it makes more sense. I would wish for it to be simpler however I can't offer a solution myself. It just made me stressed that I was unable to make sense in half an hour of going back and forth through the code.

I guess there's another problem is that it's very hard to test it properly since it depends on the kernel shenanigans. This makes a lot of logic appear to just be there for no reason.

I strongly suggest `telio-pmtu` to have these edge cases described in a comment or even maybe some readme.md about the approach and possible problems, basically so someone would not refactor it to be very simple in the future thinking it's very simple.

matszczygiel: @LukasPukenis I've added some comments to clarify a bit

LukasPukenis (commented):
+1 nice comment :)

crates/telio-pmtu/src/linux.rs
LukasPukenis: nice! I am happy :D

LukasPukenis (approved):

PR #425, opened 2024-03-13
Remove link_state flaky configurations

dfetti:
The link_state tests sometimes fails with timeout error due to a race condition in LinuxNative Wireguard implementation (described in LLT-5014).

This commit removes the flaky configs. We have to add them back when the issue is fixed.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_events_link_state.py
+++ b/nat-lab/tests/test_events_link_state.py
@@ -35,12 +35,14 @@ def _generate_setup_paramete_pair(
 
 
 FEATURE_ENABLED_PARAMS = [
-    pytest.param(
-        _generate_setup_paramete_pair([
-            (ConnectionTag.DOCKER_CONE_CLIENT_1, AdapterType.LinuxNativeWg),
-            (ConnectionTag.DOCKER_CONE_CLIENT_2, AdapterType.LinuxNativeWg),
-        ])
-    ),
+    # This scenario has been removed because it was causing flakyness due to LLT-5014.
+    # Add it back when the issue is fixed.
+    # pytest.param(
+    #     _generate_setup_paramete_pair([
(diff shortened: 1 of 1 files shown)


LukasPukenis (commented):
Can we add another case which would be Native -> BoringTun? That way we would have:
- Native <-> Boringtun :heavy_check_mark: 
- BoringTun <-> Boringtun :heavy_check_mark: 
- Native <-> Native :x: 
so we would cover more ground. Unless of course even one peer having Native makes it flaky.

EDIT
Sorry, i was just looking at the diff, didn't notice it's already like that.

+2

nat-lab/tests/test_events_link_state.py
```diff
@@ -35,12 +35,14 @@ def _generate_setup_paramete_pair(
 
 
 FEATURE_ENABLED_PARAMS = [
-    pytest.param(
-        _generate_setup_paramete_pair([
-            (ConnectionTag.DOCKER_CONE_CLIENT_1, AdapterType.LinuxNativeWg),
-            (ConnectionTag.DOCKER_CONE_CLIENT_2, AdapterType.LinuxNativeWg),
-        ])
-    ),
+    # This scenario has been removed because it was causing falkyness due to LLT-5014.
```
LukasPukenis, line 38:
```suggestion
    # This scenario has been removed because it was causing flakyness due to LLT-5014.
```

dfetti: Oh. Thanks!

dfetti:
We already have Boringtun <-> Native and Boringtun <-> Boringtun

LukasPukenis (approved):

PR #427, opened 2024-03-14
LLT-4983: expose telio panic method

lcruz99:
### Problem
libtelio has a method `__telio_generate_stack_panic` but currently it’s not being exposed to apps.

### Solution
expose this method and generate ffi bindings.

- [x] README.md is updated
- [x] changelog.md is updated


LukasPukenis (commented):
Could you slightly elaborate why apps would need such a call? It might good to overwrite commit message to includ ethat information as well.

The code itself doesn't raise any questions

lcruz99:
@LukasPukenis This is more of a convenience to apps team, so that they can properly test panic calls. Will add this info to the commit.

packgron:
> Could you slightly elaborate why apps would need such a call? It might good to overwrite commit message to includ ethat information as well.
> 
> The code itself doesn't raise any questions

We have mechanizim that allows apps to react to critical failures a.k.a panics, but they cannot actually test this and see if the code they implemented actually would work in real life scenario

packgron (approved):
+1 don't forget backports to release/4.2 release/4.3

PR #428, opened 2024-03-14
Make link state feature disabled test to not expect an exact number of events

dfetti:
Sometimes telio-proxy changes the relying port thus generating more than the 2 events the test expected previously.

- [x] README.md is updated
- [x] changelog.md is updated


matszczygiel (dismissed):
+2

nat-lab/tests/test_events_link_state.py
```diff
@@ -213,5 +213,7 @@ async def test_event_link_state_feature_disabled(
 
         # There should be 2 node events Connecting and Connected
```
LukasPukenis, line 214: This comment contradicts the code below a bit. It might be useful to paint the full picture - what we expect and why or remove the comment altogether in my opinion :/

dfetti: Removed it entirely

PR #430, opened 2024-03-15
LLT-4997: Add qos garbage collector

lcruz99:
### Problem
QoS tries to ping nodes that were already removed. A simple example can be seen with the wg-stun peers where we periodically rotate them and it can be seen on the logs that the nodes that were already removed are still being pinged by QoS.

### Solution
On each heartbeat event, remove nodes that no longer exist on the config.


matislovas:
How often, does the node cleanup takes place? Because it seems, that you're now cleaning up nodes, when the heartbeat aggregation happens, right ..?

lcruz99:
@matislovas correct, each heartbeat interval.

matislovas:
+1

Jauler (dismissed):
+1.0

Jauler (approved):
+1.0

PR #448, opened 2024-03-20
LLT-5009 - QoS enabled by default

jjanowsk:
Enable QoS feature whenever nurse is configured.
Remove unnecessary Options. Not optional field which has default is
still optional from deserialization point of view and can be ommited in
the feature string.
Introduce Duration instead of u32 as early as possible.
Move defaults related to QoS into one place.

Rename api_config.rs to features.rs - it's done in a separate commit to make review easier.

- [x] README.md is updated
- [x] changelog.md is updated


crates/telio-model/src/api_config.rs
```diff
@@ -77,65 +78,127 @@ impl FeatureWireguard {
...
-    fn get_default_initial_heartbeat_interval() -> Option<u32> {
-        // 300 seconds
-        Some(Self::DEFAULT_INITIAL_HEARTBEAT_INTERVAL.as_secs() as u32)
+    fn get_default_enable_relay_conn_data() -> bool {
+        true
+    }
+
+    fn get_default_enable_nat_traversal_conn_data() -> bool {
+        true
+    }
```
matszczygiel, line 188: Nitpick: I'm voting for stripping these `get_` prefixes

src/libtelio.udl
```diff
@@ -437,29 +443,29 @@ dictionary FeatureNurse {
...
+    duration rtt_interval;
     /// Number of tries for each node. Default value is 3.
-    u32? rtt_tries;
+    u32 rtt_tries;
     /// Types of rtt analytics. Default is Ping.
-    sequence<string>? rtt_types;
+    sequence<RttType> rtt_types;
     /// Number of buckets used for rtt and throughput. Default value is 5.
-    u32? buckets;
+    u32 buckets;
```
matszczygiel, line 468: These fields without `?` mean we do not provide defaults and force the clients to set up all files properly. While your approach with removing `Option` is fine for JSON features, it doesn't port to uniffy sadly

crates/telio-model/src/api_config.rs
packgron: +1

src/libtelio.udl
jjanowsk: That is true. I was thinking about it. But I feel like for uniffi we need to rethink/redo this part anyway. First of all the functionality is not there currently. We do not have machinery to apply defaults to the config passed as struct. We do it only via serde deserialization. Secondly even if we decide to implement it I see some problems. Few examples:
1. Let's take a look at QoS. It is really optional, meaning it can be disabled. It is now set by default after this PR. When configured via JSON it can be disabled by "qos = null". It cannot be disabled via normal config because if not set we will set it by default :) So it requires some extra work. Either an extra flag or some 3 state enum like: Feature{Some, Default, Disabled};
2. It is annoying for the fields in the config to be optional it they are really not. It causes problems when you use them in the code later. You can either unwrap() and hope that they are correctly set by the defaults. Or use unwrap_or which sucks because you then need to provide a default in a second place. Possible solutions: a) do not set defaults in the config at all. Let all the users of it set it accordingly. b) Have FeaturesInput with Options and then pass it via some validator which outputs Features with Options only in places which are really optional and defaults set in other places.

So that's the reason why I've decided to not care about this part, since it is not working anyway. Let me know what are your thoughts. IMO we should add some new task to uniffi epic to refactor this part.

crates/telio-model/src/api_config.rs
jjanowsk: 👍

src/libtelio.udl
matszczygiel: Yeah, I agree with you fully. Just please ensure people are aware of that and the ticket is there

matszczygiel (approved):
+1

nat-lab/tests/report.html
```diff
@@ -0,0 +1,482 @@
+<!DOCTYPE html>
```
gytsto, line 1: This file shouldn't be here. Can you also add it to `.gitignore`

nat-lab/tests/test_lana.py
```diff
@@ -1643,10 +1643,10 @@ async def test_lana_same_meshnet_id_is_reported_after_a_restart(
 @pytest.mark.moose
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    "initial_heartbeat_interval", [pytest.param(5), pytest.param(None)]
+    "initial_heartbeat_interval", [pytest.param(5), pytest.param(300)]
```
gytsto, line 1646: `pytest.param(300)`  is not needed here I think. Previously it was tested that no `initial_heartbeat` happens if it is not configured, so unless `initial_heartbeat_interval = 0` means disabled, other parameter is not needed besides `5`

gytsto: Or maybe nvm the upper comment, it is good to know if we don't create any heartbeats before the actual interval. But would still be nice to check if we get heartbeats if it is disabled (if possible to disable it, like before)

gytsto (dismissed):
Other than some test nitpicks and leftover file +1 :+1:

.gitignore
```diff
@@ -1,6 +1,6 @@
 Cargo.lock
 tcli.log
-./*/
+/.*/
```
gytsto, line 3: what does that actually exclude?

packgron: I think this was removed from main

packgron: As this was falsely added

jjanowsk: It excludes all directories starting with . that are in the root directory

jjanowsk: And I put it it here on purpose, because the previous `./*/` was a typo

packgron (approved):
+1.0

nat-lab/tests/test_lana.py
jjanowsk: It was not possible to disable it before. Even if null was passed the default 300 would be applied in the libtelio. That's why I put 300 here, because it exactly mimics the previous behavior.

src/libtelio.udl
packgron: I'm adding this convo as part of LLT-5022

packgron (approved):

nat-lab/tests/test_lana.py
gytsto: ah ok, good to know :smile:

PR #450, opened 2024-03-21
LLT-4746 DTCLI

ThrasherLT:
### Problem
TCLI is often inconvenient, because it requires manual user interactivity. Meaning that if a test scenario requires a lot of setup, it cannot be automated, except if NAT-Lab is being used, but that brings it's own layer of complexity and may be very hard to debug the running code.

### Solution
During the hackathon I implemented the skeleton for DTCLI (Daemonized TCLI) which runs libtelio in a background child process. This allows writing scripts (as shown in the example_vpn_enable.sh file). DTCLI essentially functions as a very simplified Linux app. It also supports almost (see readme.md) the same commands as TCLI. 

For easier analysis of the code I recommend building the documentation for dtcli with this command:
```shell
cargo doc -p tcli --bin dtcli --open --no-deps --document-private-items
```
And for easy testing you can run the example scripts:
```shell
sudo -E ./clis/tcli/distrib/example_vpn_enable_linux.sh
sudo -E ./clis/tcli/distrib/example_vpn_disable_linux.sh
```

This PR also makes TCLI incorrect command help messages more helpful. Besides that there are some nagger issues fixed in cli.rs file.

>NOTE: There may be some leftover TODOs which require input from the team, but most are issues that aren't breaking the usage of the tool or improvement ideas for the future which are simply out of scope for my current workload.

>NOTE: Leaving changelog empty, because we've agreed that changelog is consumed by the apps and this tool is for LLT use.

- [x] README.md is updated


clis/dtcli/src/doc/dtcli_documentation.md
```diff
@@ -0,0 +1,89 @@
...
+
+```shell
+cargo build -p dtcli
+```
+
+## How To Use DTCLI
+
+### Prerequisites
+
+DTCLI needs a user group called `nordvpn` to be present on the current machine to work. If you have nordvpn installed, the group probably already exists, but if it doesn't you can add it with `groupadd nordvpn`
```
packgron, line 17: you probably also what to add line saying, 'then add user to it  usermod -aG nordvpn $USER, and relogin'

packgron (commented):
If you have this common project maybe it would make more sence to strcture project like this:
```
tcli
├── Cargo.toml
└── src
    ├── bin
    │   ├── dtcli
    │   │   ├── comms.rs
    │   │   ├── daemon.rs
    │   │   ├── dtcli.rs
    │   │   └── main.rs
    │   └── tcli.rs
    ├── cli.rs
    ├── derp.rs
    ├── lib.rs
    └── nord.rs
```

clis/common/src/cli.rs
```diff
@@ -757,13 +774,13 @@ impl Cli {
                 stun_server.parse::<IpAddr>().expect("Invalid IPv4 address"),
                 stun_port,
             )) {
-                Ok(data) => {
-                    cli_res!(res; (i"Public Address: {:?}", data.public_ip));
-                    cli_res!(res; (i"Nat Type: {:?}", data.nat_type));
+                Ok(_data) => {
+                    // cli_res!(res; (i"Public Address: {:?}", data.public_ip));
```
packgron, line 778: why is this commented out?

clis/dtcli/src/coms.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// * `addr` - Address (more specifically path, but left address for clarity) of the IPC socket (same as `send_command`).
+    /// * `cmd` - Command to be sent to the daemon (same as `send_command`).
+    /// * `retry_delay_list` - a list of retry delays. The length of the list is the retry count.
+    /// # Returns
+    ///
+    /// A Result containing a response string (same as `send_command`).
+    pub fn send_command_with_retries(
+        addr: &Path,
+        cmd: &str,
+        retry_delay_list: Vec<tokio::time::Duration>,
```
packgron, line 138: why not &[Duration]

clis/dtcli/src/dtcli.rs
```diff
@@ -0,0 +1,113 @@
...
+use crate::get_ipc_socket_path;
+use crate::telio_cli_interface::TelioCliInterface;
+
+/// Function to setup the DTCLI daemon's dependencies (including the async runtime) and and run it.
+///
+/// # Arguments
+///
+/// * `features` - Optional string containing the features specification for Telio in JSON.
+pub fn run_dtcli(features: Option<String>) -> Result<()> {
+    let nord_token = std::env::var("NORD_TOKEN").ok();
```
packgron, line 18: Does this function? can we assume deamon that is ran with root, coppies env vars with fork?

clis/dtcli/src/dtcli.rs
```diff
@@ -0,0 +1,113 @@
...
+    let telio_cli = TelioCliInterface::new(nord_token, features)?;
+
+    let _ = tokio::runtime::Runtime::new()?.block_on(async {
```
packgron, line 21: We basicaly will have two tokio rutimes? one build for this and one for telio ?

clis/common/src/cli.rs
ThrasherLT: My bad, leftover code from debugging issues. The missing whitespace between i and the string literal was giving me weird compiler errors before. I'll bring it back.

clis/dtcli/src/dtcli.rs
ThrasherLT: It works and a forked process inherits the exported environment variables. I did update the runner to use sudo -E tho instead of just sudo. I used to pass the token through function arguments at an earlier stage of development, I can bring that back if you think it'll make the code more robust against issues with the environment.

clis/dtcli/src/dtcli.rs
ThrasherLT: Yes, added comment with our irl discussion points to make this clearer for developers in the future.

clis/dtcli/src/doc/dtcli_documentation.md
ThrasherLT: Done

ThrasherLT:
> If you have this common project maybe it would make more sence to strcture project like this:
> 
> ```
> tcli
> ├── Cargo.toml
> └── src
>     ├── bin
>     │   ├── dtcli
>     │   │   ├── comms.rs
>     │   │   ├── daemon.rs
>     │   │   ├── dtcli.rs
>     │   │   └── main.rs
>     │   └── tcli.rs
>     ├── cli.rs
>     ├── derp.rs
>     ├── lib.rs
>     └── nord.rs
> ```

@packgron Done

clis/dtcli/src/coms.rs
ThrasherLT: Done, somehow forgot that slices exist, when I was writing that :D

clis/tcli/src/bin/dtcli/daemon.rs
```diff
@@ -0,0 +1,118 @@
+/// Code for general daemon handling utilities and abstracting away related dependencies.
```
packgron, line 1:
```suggestion
//! Code for general daemon handling utilities and abstracting away related dependencies.
```

clis/tcli/src/bin/dtcli/dtcli.rs
```diff
@@ -0,0 +1,121 @@
+/// Code for the daemon which runs actual libtelio in the background.
```
packgron, line 1:
```suggestion
//! Code for the daemon which runs actual libtelio in the background.
```

clis/tcli/src/bin/dtcli/examples/example_vpn_disable.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
packgron, line 1: Maybe leet's move examples to the root of tcli for better visability

packgron: And probably nameing it contrib would be better not to confuse with rust examples

start_meshnet.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
packgron, line 1: comaybe should be in contrib dir

clis/tcli/src/bin/dtcli/coms.rs
```diff
@@ -0,0 +1,151 @@
+/// Code for handling inter process communication and abstracting away it's dependencies.
```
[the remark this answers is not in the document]
ThrasherLT: Done

clis/tcli/Cargo.toml
```diff
@@ -15,6 +16,15 @@ reqwest = { version = "0.11.16", default-features = false, features = [
...
 shellwords = "1.1.0"
+# Used only for checking if the daemon is running.
+sysinfo = "0.30.11"
+# Used as a lightweight and safe (because a TCP server has the risk of remote code execution)
+# way for the API and daemon to communicate.
+# Tokio support is needed, because the daemon runs on the async runtime.
+interprocess = { version = "1.2.1", features = [ "tokio_support" ] }
+# Needed for reading and writting to the local_socket of interprocess crate from the daemon side
+# which has an async runtime.
+futures = "0.3.30"
```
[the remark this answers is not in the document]
ThrasherLT: Done

clis/tcli/src/bin/dtcli/daemon.rs
ThrasherLT: Done

clis/tcli/src/bin/dtcli/dtcli.rs
ThrasherLT: Done

clis/tcli/src/bin/dtcli/examples/example_vpn_disable.sh
ThrasherLT: Done

start_meshnet.sh
ThrasherLT: Removed this one completely since it's not finished anyway. I'll find time to add it in the future.

clis/tcli/contrib/example_vpn_disable_linux.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# This is a simple example script on how to use DTCLI to turn off VPN and clean up routing.
+
+# NOTE: This is meant to be run from the libtelio directory.
+# And uses the debug build so that you could iterate quicker builds when debugging.
+
+DTCLI_PATH="target/debug/dtcli"
+
+# You may need to change these to your default values.
+TUNNEL_NAME="tun10"
```
Hasan6979, line 11: Can you add this as a parameter to the script ? If parameter not given, defaults to "tun10".

clis/tcli/contrib/example_vpn_disable_linux.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Check if the interface exists.
+if ! ip link show | grep -q "$TUNNEL_NAME"; then
+    echo "Tunnel $TUNNEL_NAME does not exists. Enable VPN before disabling it again."
+    exit 1
+fi
+
+# Ensure ip is installed.
+if ! command -v ip &> /dev/null; then
+    echo "iproute2 is required but not found. Please install it."
```
Hasan6979, line 38: Add command to install iproute2 to make this more one-stop solution

clis/tcli/contrib/example_vpn_enable_linux.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+
+# This is a simple example script on how to use DTCLI to set up VPN.
+
+# NOTE: This is meant to be run from the libtelio directory.
+# Also it assumes that your nord token is in the $NORD_TOKEN variable. So also make sure to run it with `sudo -E`.
+# And uses the debug build so that you could iterate quicker builds when debugging.
+
+# Keeping this in a separate variable so that a Mac OS compatible tunnel name could be added later.
+TUNNEL_NAME="tun10"
```
Hasan6979, line 10: Same for this

clis/tcli/src/bin/dtcli/doc/dtcli_documentation.md
```diff
@@ -0,0 +1,105 @@
...
+- MacOS is supported, but not tested.
+- Windows is not supported yet.
+
+## Compatibility
+
+All of the examples given here were tested with:
+
+- rustc 1.77.2 (25ef9e3d8 2024-04-09)
+- cargo 1.77.2 (e52e36006 2024-03-26)
+- rustdoc 1.77.2 (25ef9e3d8 2024-04-09)
```
Hasan6979, line 105: 👍 👍  for the documentation!

Hasan6979: You can add `#![deny(missing_docs)]` to maintain future update with relevant documentation

clis/tcli/src/bin/dtcli/coms.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// forwards them to the event loop.
+    /// When the event loop produces a response, sends it back to the API
+    /// over the same IPC socket.
+    /// Note that since a single socket is used, care should be taken to not
+    /// deadlock it.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - Address (more specifically path, but left address for clarity) of the IPC socket.
+    pub async fn listen(mut self, addr: &Path) -> Result<()> {
```
Hasan6979, line 58: nit: This could also be renamed to `IPC_sock_fd`

clis/tcli/src/bin/dtcli/main.rs
```diff
@@ -0,0 +1,187 @@
...
+    println!("Starting DTCLI daemon...");
+    match Daemon::start() {
+        Ok(ProcessType::Api) => {
+            // Verifying that the server had been started, but we have no way of synchronizing
+            // the verification with the daemon, so using retries instead.
+            DaemonIPC::send_command_with_retries(
+                &get_ipc_socket_path()?,
+                "status",
+                &[
+                    Duration::from_millis(500),
```
Hasan6979, line 163: How were these delay values chosen ?

ThrasherLT: Experimentally by testing how long it takes for the IPC listener and Telio. There wasn't really any numbers on which I could rely on nor was this a big risk of failure point.

clis/tcli/src/bin/dtcli/daemon.rs
```diff
@@ -0,0 +1,119 @@
...
+            // This crate only works on unix systems and not Windows.
+            use daemonize::Daemonize;
+
+            let dtcli_wd_path = get_wd_path()?;
+
+            std::fs::create_dir_all(&dtcli_wd_path)?;
+            let stdout = File::create(dtcli_wd_path.join("dtcli.out"))?;
+            let stderr = File::create(dtcli_wd_path.join("dtcli.err"))?;
+
+            // TODO Check if nordvpn group exists and inform the user that it must be created, if it doesn't.
```
Hasan6979, line 85: nit: Updatw TODO to create the nordvpn group if it doesn't exist, instead of just informing

clis/tcli/src/bin/dtcli/daemon.rs
```diff
@@ -0,0 +1,119 @@
...
+pub struct Daemon;
+
+impl Daemon {
+    /// Checks whether the daemon is currently running.
+    /// Won't work if pid file had been tampered with or it's location changed.
+    ///
+    /// # Returns
+    ///
+    /// Result containing a bool. True means that the daemon is running.
+    pub fn is_running() -> Result<bool> {
```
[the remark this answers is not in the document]
ThrasherLT: But doing it this way allows writing cleaner app code like `if is_running()? {/*do something*/} instead of matching everywhere to see if there's an issue with the system. Also a lot of the times the expected state is false so I wouldn't want to treat that as an error in code.

clis/tcli/src/bin/dtcli/main.rs
ThrasherLT: I'll add a comment about it

clis/tcli/contrib/example_vpn_disable_linux.sh
ThrasherLT: Done

clis/tcli/src/bin/dtcli/doc/dtcli_documentation.md
ThrasherLT: Done

clis/tcli/src/bin/dtcli/coms.rs
ThrasherLT: Done

clis/tcli/src/bin/dtcli/daemon.rs
ThrasherLT: Done

clis/tcli/src/bin/dtcli/main.rs
Hasan6979: Mm. Then these numbers could be incrementally increasing.

clis/tcli/src/bin/dtcli/daemon.rs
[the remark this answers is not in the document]
Hasan6979: Ah got it!

clis/tcli/src/bin/dtcli/main.rs
ThrasherLT: I'd rather keep them as they are, since the daemon never starts sooner than 500 ms and the largest possibility of it starting is between 500 and 1100 ms. I left the longer delay at the end in case there's some edge case where it takes longer for some reason.

clis/tcli/contrib/example_vpn_disable_linux.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+if ! [ -f "$TCLID_PATH" ]; then
+    echo "Cannot find TCLID in $TCLID_PATH directory. Please build TCLID and run it from libtelio root."
+    exit 127
+fi
+
+# Check if the interface exists.
+if ! ip link show | grep -q "$TUNNEL_NAME"; then
+    echo "Tunnel $TUNNEL_NAME does not exists. Enable VPN before disabling it again."
```
Jauler, line 37: Nit:
```suggestion
    echo "Tunnel $TUNNEL_NAME does not exist. Enable VPN before disabling it again."
```

clis/tcli/contrib/example_vpn_disable_linux.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Ensure ip is installed.
+if ! command -v ip &> /dev/null; then
+    echo "iproute2 is required but not found. Please install it."
+    exit 127
+fi
+
+# Save current external IP to verif that it had changed later.
+old_public_ip=$(curl -s ifconfig.me)
+echo "Externa IP before enabling VPN: ${old_public_ip}"
```
Jauler, line 49: Nit:
```suggestion
echo "External IP before enabling VPN: ${old_public_ip}"
```

clis/tcli/contrib/example_vpn_enable_linux.sh
```diff
@@ -0,0 +1,117 @@
...
+ip link set up dev ${TUNNEL_NAME}
+ip link set dev ${TUNNEL_NAME} mtu 1420
+
+# Connecting to VPN.
+eval ${TCLID_PATH} dev con ${recommended_server_public_key} ${recommended_server_ip}:51820
+
+# Creating VPN route.
+ip route add default dev ${TUNNEL_NAME} table 73110
+ip route add ${recommended_server_ip} dev ${TUNNEL_NAME} table 73110
+ip rule add priority 32111 not from all fwmark 11673110 lookup 73110
```
Jauler, line 108: I believe in this case, you may omit priority, as default behavior of `ip rule add` is desirable and more flexible. By default it adds the rule with just a bit (by one priority point) higher priority compared to any rule in the table (except the local table lookup, which has priority of `0`).

clis/tcli/contrib/example_vpn_disable_linux.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Stop WireGuard.
+eval ${TCLID_PATH} dev stop
+# Stop libtelio daemon.
+eval ${TCLID_PATH} quit
+
+# Delete VPN route.
+# Routing into tunnel and routing from tunnel to VPN server will be deleted automatically when stopping Wireguard.
+ip rule del priority 32111 not from all fwmark 11673110 lookup 73110
```
Jauler, line 58: Same as: https://github.com/NordSecurity/libtelio/pull/450/files#r1606432734
```suggestion
ip rule del not from all fwmark 11673110 lookup 73110
```

clis/tcli/contrib/example_vpn_enable_linux.sh
Jauler, line 108:
```suggestion
ip rule add not from all fwmark 11673110 lookup 73110
```

I believe in this case, you may omit priority, as default behavior of `ip rule add` is desirable and more flexible. By default it adds the rule with just a bit (by one priority point) higher priority compared to any rule in the table (except the local table lookup, which has priority of `0`).

clis/tcli/src/cli.rs
```diff
@@ -665,29 +689,25 @@ impl Cli {
...
-                                break;
-                            },
-                            () = &mut sleep => {
-                                cli_res!(res; (i "timeout"));
-                                break;
-                            },
-
-                        }
-                    }
+                            // Timeout for stunner
```
Jauler, line 692: Just to double-check, this is just reformatting, as previously the whole code block was off-by-one indented?

packgron (approved):
+0.5

clis/tcli/src/bin/tclid/coms.rs
```diff
@@ -0,0 +1,152 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - Address (more specifically path, but left address for clarity) of the IPC socket (same as `send_command`).
+    /// * `cmd` - Command to be sent to the daemon (same as `send_command`).
+    /// * `retry_delay_list` - a list of retry delays. The length of the list is the retry count.
+    /// # Returns
+    ///
+    /// A Result containing a response string (same as `send_command`).
+    pub fn send_command_with_retries(
```
Jauler, line 135: Is there really a retry here? 

I would imagine, that if socket failed to open, or any other `write()` returned an error, in this local setup we can probably consider this to be fatal error. But maybe there are some scenarios, which I am forgetting to consider?

clis/tcli/src/bin/tclid/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    let args: Args = Args::parse();
+
+    match run_api(args) {
+        Ok(()) => Ok(()),
+        Err(e) => {
+            #[cfg(unix)]
+            // For convenience printing stderr of the daemon to inform the user that something's wrong.
+            if let Ok(stderr) = Daemon::get_stderr() {
+                if !stderr.is_empty() {
+                    println!("Daemon stderr: {}", stderr);
```
Jauler, line 60: Maybe it would be nice to use some sort of logging, e.g. even https://docs.rs/simple_logger/latest/simple_logger/ instead of `println!`?  :thinking:

clis/tcli/src/bin/tclid/main.rs
```diff
@@ -0,0 +1,189 @@
...
+mod telio_cli_interface;
+
+use coms::DaemonIPC;
+use daemon::{Daemon, ProcessType};
+use telio_cli_interface::TelioCliInterface;
+
+/// Name for the local socket file which will be used for inter process communication.
+const DAEMON_IPC_SOCKET: &str = "tclid.sock";
+/// Path where all the logs, local socket and pid file will be stored, relative to the path of this binary.
+const TCLID_WD_PATH: &str = "/var/run/tclid_wd";
```
Jauler, line 25: :thinking: This may be not portable :/.

There is this (runtime_dirs)[https://docs.rs/dirs/latest/dirs/fn.runtime_dir.html] thing, which sounds like a perferct candidate for our purposes, but, unfortunately, it is only supported for Linux, so we would not gain any compatability. 

As a compromise, we _could_ use (dirs::data_dir())[https://docs.rs/dirs/latest/dirs/fn.data_dir.html] which is not ideal, but at least portable.

clis/tcli/src/bin/tclid/coms.rs
ThrasherLT: The daemon takes time to daemonize and start listening for commands and if we send commands before it's started, it returns "connection refused" errors. I've figured these retry times by experimenting with the startup times of the daemon. I might have missed something which was causing the startup delay, but when I was trying to debug it by measuring OS time at various points, everything seemed to make sense.

Jauler: OOoh, I see this is used at startup, as when the process is still being started. But I am still slightly confused why this is needed, as the socket file creation itself should be a good indicator for daemon process readyness.

clis/tcli/src/bin/tclid/daemon.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+/// Private helper function for getting the path to the daemon's working directory.
+///
+/// # Returns
+///
+/// Result containing the path of the daemons work directory.
+fn get_wd_path() -> Result<PathBuf> {
+    Ok(std::env::current_exe()?
```
Jauler, line 113: Maybe [current_dir](https://doc.rust-lang.org/std/env/fn.current_dir.html) would be better choise when `current_exe`?

clis/tcli/src/bin/tclid/daemon.rs
```diff
@@ -0,0 +1,119 @@
...
+            std::fs::create_dir_all(&tclid_wd_path)?;
+            let stdout = File::create(tclid_wd_path.join("tclid.out"))?;
+            let stderr = File::create(tclid_wd_path.join("tclid.err"))?;
+
+            // TODO Create nordvpn group if it doesn't exist.
+            let daemonize = Daemonize::new()
+                .pid_file("tclid.pid") // Every method except `new` and `start`
+                .chown_pid_file(true) // is optional, see `Daemonize` documentation
+                .working_directory(tclid_wd_path) // for default behavior.
+                .group("nordvpn") // Group name
```
Jauler, line 90: This group probably will not exist in a lot of context, e.g. like this:
```
Starting TCLID daemon...
Daemon stderr: Error: unable to resolve group name to group id
```

And IMHO, there is no need to actually setup group for the daemon ;). Unless, again, I am missing something.

clis/tcli/src/bin/tclid/coms.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// * `cmd` - Command to be sent to the daemon.
+    ///
+    /// # Returns
+    ///
+    /// A Result containing a response string.
+    pub fn send_command(addr: &Path, cmd: &str) -> Result<String> {
+        let mut response_buffer = String::new();
+        let mut stream = BlockingSocketStream::connect(addr)?;
+        writeln!(stream, "{}", cmd)?;
+        stream.read_to_string(&mut response_buffer)?;
```
Jauler, line 119: I am slightly confused about one thing. `stream.read_to_string(....)` is supposed to read all IPC contents until local Unix socket is closed (e.g. until EOF), `tclid` does seem to work, but I am unsure why this `read_to_string` call does not block until daemon is stopped/closes it's side of the socket.

clis/tcli/src/bin/tclid/daemon.rs
```diff
@@ -0,0 +1,119 @@
...
+                .stdout(stdout) // Redirect stdout to `/tmp/daemon.out`.
+                .stderr(stderr) // Redirect stderr to `/tmp/daemon.err`.
+                .privileged_action(|| "Executed before privileges are dropped");
```
Jauler, line 93:
```suggestion
                .stderr(stderr); // Redirect stderr to `/tmp/daemon.err`.
```

clis/tcli/src/cli.rs
Jauler: :+1:

clis/tcli/src/bin/tclid/daemon.rs
Jauler: Or one more idea: [data_dir](https://docs.rs/dirs/latest/dirs/fn.data_dir.html)

clis/tcli/src/bin/tclid/coms.rs
ThrasherLT: This was supposed to be blocking. I made the API blocking instead of using an async runtime for performance reasons. Anyway I've reworked everything so that TCLID doesn't use an async runtime at all which is actually solving a lot of issues.

clis/tcli/src/bin/tclid/daemon.rs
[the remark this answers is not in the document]
ThrasherLT: As much as I've read setting a user group is good practice for a daemon to limit it's permissions, but since I didn't really add any limiting, I'll remove it for now.

clis/tcli/contrib/example_vpn_enable_linux.sh
ThrasherLT: done

clis/tcli/contrib/example_vpn_disable_linux.sh
ThrasherLT: done

clis/tcli/contrib/example_vpn_disable_linux.sh
ThrasherLT: done

clis/tcli/contrib/example_vpn_disable_linux.sh
ThrasherLT: done

clis/tcli/src/bin/tclid/main.rs
ThrasherLT: Because Windows is not supported right now and adding support for it would require a lot of work, we've agreed to leave this as it is, because on Unix this is the standard location for pid and sock files.

clis/tcli/src/bin/tclid/daemon.rs
ThrasherLT: Changed to current_dir, as data_dir doesn't quite fit our use case

clis/tcli/src/bin/tclid/main.rs
ThrasherLT: I've moved the tracing subscriber from Cli to TCLI main and TCLID daemon, so now println is only used to respond to user commands

clis/tcli/src/bin/tclid/coms.rs
Jauler: Oh I was not thinking about async runtime ;), I meant, that `read_to_string()` should not return until daemon closes the unix domain socket, which, to my understanding it never does. Therefore I was confused why the client does not hang forever on this line.

Jauler (approved):
+1.0

Hasan6979:
+0.5

PR #452, opened 2024-03-22
LLT-5009 - QoS enabled by default - release/v4.3 backport

jjanowsk:
Enable QoS feature whenever nurse is configured.
Remove unnecessary Options. Not optional field which has default is
still optional from deserialization point of view and can be ommited in
the feature string.
Introduce Duration instead of u32 as early as possible.
Move defaults related to QoS into one place.

Rename api_config.rs to features.rs - it's done in a separate commit to make review easier.

- [x] README.md is updated
- [x] changelog.md is updated


matszczygiel (dismissed):
+1

packgron (approved):
+2.0

PR #453, opened 2024-03-22
[LLT-4802]Move downgrade handling to mod/device.rs

stalowyjez:
### Problem
Original task was already done in the solution for LLT-4981. We need to additionaly remove timeout from network_switch_direct test which should now be obsolete. But after removing it , so we move downgrade handling to mod/device.rs, to make it more robust.

### Solution
In this PR we remove the mentioned timeout, it also simplifies select_endpoint_for_peer function a bit

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -548,6 +548,7 @@ impl State {
         state: PeerState,
         link_state: Option<LinkState>,
         peer: Peer,
+        old_peer: Option<Peer>,
     ) -> Result<(), Error> {
         let link_state = link_state.and_then(|s| {
             if self.no_link_detection.is_disabled() {
@@ -562,6 +563,7 @@ impl State {
                 state,
                 link_state,
                 peer,
+                old_peer,
             }))
             .await
             .map_err(|_| Error::InternalError("Failed to send node event"))
@@ -585,8 +587,13 @@ impl State {
             // Remove all disconnected peers from no link detection mechanism
             self.no_link_detection.remove(key);
 
-            self.send_event(PeerState::Disconnected, Some(LinkState::Down), peer)
-                .await?;
+            self.send_event(
+                PeerState::Disconnected,
+                Some(LinkState::Down),
+                peer,
+                from.peers.get(key).cloned(),
+            )
+            .await?;
         }
 
         // Notify all new connections
@@ -600,8 +607,13 @@ impl State {
             self.no_link_detection
                 .insert(key, None, None, PeerState::Connecting);
 
-            self.send_event(PeerState::Connecting, Some(LinkState::Down), peer.clone())
-                .await?;
+            self.send_event(
+                PeerState::Connecting,
+                Some(LinkState::Down),
+                peer.clone(),
+                from.peers.get(key).cloned(),
+            )
+            .await?;
 
             if peer.is_connected() {
                 // If the new peer is connected, update last link state to Up.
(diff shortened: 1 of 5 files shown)


crates/telio-wg/src/wg.rs
```diff
@@ -612,11 +619,28 @@ impl State {
...
+                    PeerState::Connected,
+                    Some(LinkState::Up),
+                    peer.clone(),
+                    false,
+                )
+                .await?;
             }
         }
 
+        let is_direct = |peer: &Peer| {
```
Jauler, line 632: :thinking: I can help, but wonder, generally the endpont of `127.0.0.1` is only half-correct way of identifying direct vs relayed peers. E.g. it is, at least theoretically (although in practice it may be hard to achieve), to have another peer on the same computer reachable connected over localhost. 

For this reason, in most (hopefully all) places we are using the endpoints map acquired from `telio-proxy` and comparing wireguard endpoint with that one. 

Therefore I would like recommend a little refactoring. In all cases where `self.send_event(...)` is called, you have access  to variable called "from" which represents the state of WireGuard before the changes. Maybe lets just include `Peer` structure from this `from` wireguard state in the event being sent to the channel. And in the `wg_controller.rs` then we can compare "old" and "new" peer's endpoints against actual `telio-proxy` endpoint :) and determine in this way, whether it was downgraded/upgraded or stayed the same.

Jauler: oh, I meant `device/mod.rs` not `wg_controller.rs`. Sorry for mixup

nat-lab/tests/test_pinging.py
```diff
@@ -149,7 +149,12 @@ async def test_session_keeper(
         )
 
         async def wait_for_conntracker() -> None:
+            counter = 0
             while True:
+                if counter % 60 == 0:
+                    print(f"{alpha_conn_tracker.get_out_of_limits()}")
+                    print(f"{beta_conn_tracker.get_out_of_limits()}")
+                counter += 1
```
Jauler, line 157: Could you clarify this change a little bit? Why it is needed?

stalowyjez: Some debug left, removing it now

Jauler (dismissed):
+1.0

src/device/mod.rs
```diff
@@ -1997,6 +1998,32 @@ impl TaskRuntime for Runtime {
             },
 
             Some(mesh_event) = self.event_listeners.wg_event_subscriber.recv() => {
+                let public_key = mesh_event.peer.public_key;
+
+                if let Some(mesh_entities) = self.entities.meshnet.as_ref() {
+                    if let Some(proxy_endpoints) = mesh_entities.proxy.get_endpoint_map().await.ok().and_then(|proxy_map| proxy_map.get(&public_key).cloned()) {
```
matszczygiel, line 2004: This `.cloned()` could've been avoided with either storing `proxy_map` in a variable or simply using `remove()` instead of `get()` since the map itself is no longer used

src/device/mod.rs
```diff
@@ -1997,6 +1998,32 @@ impl TaskRuntime for Runtime {
...
+                        let is_proxying = mesh_event.peer.endpoint.map(|ep| proxy_endpoints.contains(&ep)).unwrap_or(false);
+                        let was_proxying = mesh_event.old_peer.and_then(|peer| peer.endpoint.map(|ep| proxy_endpoints.contains(&ep))).unwrap_or(false);
```
matszczygiel, line 2006: Nitpick: `map()` and then `unwrap_or()` can be shortened with single call `map_or()`

stalowyjez: In the first one it makes sense, in the second, as I have two layers of options, the current one looks better to me.

matszczygiel (approved):
+1

PR #458, opened 2024-03-27
Add in-repo documentation system skeleton

ThrasherLT:
### Problem
A need arose to have an in-repo documentation system for Libtelio

### Solution
Added mdbook documentation system skeleton to libtelio repo so that it can be populated by further PR's, but after discussions in this PR we've decided that it'd be simpler to use doc rs according to how [the Clap crate](https://docs.rs/clap/latest/clap/index.html) used it to document their tutorials instead.

The current documentation can be viewed by using `cargo doc --open` in the libtelio level of the repo. Proper generation by CI will be added with later PRs

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/doc/integrating_telio.md
+++ b/src/doc/integrating_telio.md
@@ -1,0 +1,3 @@
+# Documentation For Integrating Telio Into Other Languages
+
+<!-- TODO: This page is a placeholder and will be populated by upcoming PR's when the actual specification will be more clear -->
--- a/src/doc/introduction.md
+++ b/src/doc/introduction.md
@@ -1,0 +1,221 @@
+# Introduction
+
+Libtelio (pronounced 'lɪbtælɪɔ') is a client-side library for creating encrypted networks (called Meshnet) on the user's nodes.
+It supports a number of features, including:
+
+- adapting different Wireguard implementations to secure the connections,
+- exit nodes which might be either VPN servers or some nodes in the Meshnet,
+- "DNS" which allows name resolution in Meshnets for devices which does not support lookup tables,
+- adjustable firewall.
+
+## Quick Links
+
+- Telio integration [documentation][_telio_integration_documentation]
+- Telio Events [documentation][_telio_events_documentation]
+
+## Using the libtelio API
+
+See [this documentation module][_telio_integration_documentation] for notes on how to integrate Libtelio into an application using one of the different supported languages.
+
+### Initializing the telio device
+
+The main component of the Libtelio library is the `telio::Device` structure and its methods.
+Let's go through the most important ones.
+
+The first of them is the expected function `::new`.
+It takes three arguments, let's describe them briefly:
+ - [`features`](telio_model::features::Features) - tells Libtelio which of the additional features should be enabled - the
+ description of them is out of the scope of this README
+ - `event_cb` - event handler, takes a [`Box<Event>`][telio_model::event::Event], which will be called by Libtelio to handle
+ events of three types:
+   - `Error` - an error occurs, especially urgent are Critical error events, which means
+   that the library is unable to continue running and it requires a call to `hard_reset` method
+   - `Relay` - when the relay (i.e. Derp) server configuration is changed, contains
+   a JSON with a new one
+   - `Node` - appears when the Meshnet node's configuration is changed, it contains
+   a JSON with a new one
+ - `protect` - callback for excluding connections from VPN tunnel (currently used only for android).
+
+`telio::Device` implements `Drop` trait, so we don't need to worry
+about deinitialization in the end.
+Let's look at an example initialization of `telio::Device` with no additional
+features, handling only `Node` events and not using `protect` callback:
+
+```rust no_run
+use telio_model::features::Features;
+use telio_model::event::Event;
+
+let (sender, receiver) = std::sync::mpsc::channel::<Box<Event>>();
+
+let mut device = telio::device::Device::new(
+    Features::default(),
+    move |e| {
+        sender.send(e).unwrap();
+    },
+    None,
+).unwrap();
+
+// ...
+
+loop {
+    let event = receiver.recv().unwrap();
+    match *event {
+        Event::Node { body: Some(b) } => println!(
+            "event node: {:?}:{};  Path = {:?}",
+            b.state,
+            b.public_key,
+            b.path
+        ),
+        _ => {}
+    };
+}
+```
+
+### Starting the Telio device
+
+Once the device is initialized we can start it, so it will create a new network
+interface in your OS. This might be done using `start` method.
+We need to provide an instance of the `DeviceConfig` structure:
+
+```rust no_run
+use telio_wg::AdapterType;
+use telio_crypto::SecretKey;
+use telio_wg::Tun;
+
+pub struct DeviceConfig {
+    pub private_key: SecretKey,
+    pub adapter: AdapterType,
+    pub name: Option<String>,
+    pub tun: Option<Tun>,
+}
+```
+
+Let's discuss its fields shortly:
+ - `private_key` a `telio::crypto::SecretKey` instance containing a 256-bit key,
+ - `adapter` indicating which Wireguard implementation we want to use,
+ - `name` is the name of the network interface, when omitted, Telio uses the default one,
+ - `tun` a file descriptor of the already opened tunnel, if it's not provided Telio will open a new one.
+
+The API provides a default config which is almost sufficient for simple cases,
+the only need that needs to be done is the generation of a private key:
+
+```rust no_run
+use telio_model::{features::Features, event::Event};
+
+let (sender, receiver) = std::sync::mpsc::channel::<Box<Event>>();
+
+let mut device = telio::device::Device::new(
+    Features::default(),
+    move |e| {
+        sender.send(e).unwrap();
+    },
+    None,
+).unwrap();
+
+let config = telio::device::DeviceConfig {
+    private_key: telio_crypto::SecretKey::gen(),
+    ..Default::default()
+};
+
+device.start(&config).unwrap();
+```
+
+To stop the device we can simply call the argumentless `Device::stop` method.
+
+### Creating a Meshnet
+
+To turn on the Meshnet feature, you need to call the `Device::set_config` method
+with the proper config. After logging in and registering to the desired
+Derp server, the JSON with config may be downloaded from it.
+
+In the case of using the Derp server provided by Nord, you can use a token (like the one
+used in the tcli setup) passing it with a user HTTP header (in a form `token:<TOKEN>`),
+to receive the necessary credentials from
+`https://api.nord.com/v1/users/services/credentials`
+which will return you another token, with which you may register to
+`https://api.nord.com/v1/meshnet/machines`
+passing the token in the authorization data: `Bearer token:<TOKEN>`.
+You will be given an `ID` with which you can download the config from
+`https://api.nord.com/v1/meshnet/<ID>/`
+using the same bearer token.
+
+After the device is started and the JSON config downloaded, we can
+deserialize it and finally call `set_config`:
+
+```rust no_run
+use telio_model::{config::Config, features::Features, event::Event};
+# // Leaving this hidden, since there's no point in pasting the whole config here.
+# let serialized_config: &str = "";
+
+let config = telio::device::DeviceConfig {
+    private_key: telio_crypto::SecretKey::gen(),
+    ..Default::default()
+};
+
+let (sender, receiver) = std::sync::mpsc::channel::<Box<Event>>();
+
+let mut device = telio::device::Device::new(
+    Features::default(),
+    move |e| {
+        sender.send(e).unwrap();
+    },
+    None,
+).unwrap();
+
+let config: Config = serde_json::from_str(&serialized_config).unwrap();
+
+device.set_config(&Some(config)).unwrap();
+```
+
+To turn the Meshnet off, we need to just call `Device::set_config` with `None`.
+
+### Selecting an exit node
+
+The other thing you may do with the started device is to set up an exit node.
+It might be either one of the nodes in your internal network, or
+some VPN server compatible with Wireguard. To connect to them, we need
+their public key and endpoint.
+
+If we want to use the NordVPN server, we can obtain them from the JSON
+downloaded from
+`https://api.nordvpn.com/v1/servers/recommendations`.
+The received JSON is rather big, so to get the needed fields
+you can extract the `telio::tcli::nord::find_server` method and simplify
+it a bit to just return a pair of `public_key` and `endpoint`.
+When you have it, setting up the VPN connection is fairly simple:
+
+```rust no_run
+use telio_model::{event::Event, mesh::{IpNetwork, ExitNode}};
+use std::net::{SocketAddr, IpAddr, Ipv4Addr};
+use telio::crypto::PublicKey;
+
+// A very simplified function which returns example values for the public key, IP and port of the VPN server.
+fn find_server() -> (PublicKey, SocketAddr)
+{
+    (PublicKey::default(), SocketAddr::new(IpAddr::V4(Ipv4Addr::new(1, 2, 3, 4)), 51820))
+};
+
+let (sender, receiver) = std::sync::mpsc::channel::<Box<Event>>();
+
+let mut device = telio::device::Device::new(
+    telio_model::features::Features::default(),
+    move |e| {
+        sender.send(e).unwrap();
+    },
+    None,
+).unwrap();
+
+let (public_key, endpoint) = find_server();
+let exit_node = ExitNode {
+    identifier: "fa5bbe9b-338b-4bd2-8c97-166ceee65790".to_owned(),
+    public_key,
+    allowed_ips: Some(vec![IpNetwork::new(
+        IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
+        0,
+    )
+    .unwrap()]),
+    endpoint: Some(endpoint),
+};
+device.connect_exit_node(&exit_node);
+```
+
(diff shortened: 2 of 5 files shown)


stalowyjez (commented):
I have two things to consider:

1. I'm not convinced that we need to keep it inside the repo, it looks that it is designed to be a separate git project, so maybe we could follow this? I think we could even make the sources private as they are not that useful and publish only the book, as this is what user will use.
2. I think it would be nice to add to README some info about the book - if it will be hosted in the future, it would be nice to add some link, but also info that it exists and some instructions how to build it

telio-book/src/api.md
```diff
@@ -0,0 +1 @@
+# API
```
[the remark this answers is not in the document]
ThrasherLT: Renamed it to Telio Integration to prevent confusion, but it is a placeholder and it's contents are outside of the scope of this ticket.

telio-book/.gitignore
```diff
@@ -0,0 +1 @@
+book
```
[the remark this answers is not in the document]
ThrasherLT: Removed

ThrasherLT:
> I have two things to consider:
> 
>     1. I'm not convinced that we need to keep it inside the repo, it looks that it is designed to be a separate git project, so maybe we could follow this? I think we could even make the sources private as they are not that useful and publish only the book, as this is what user will use.
> 
>     2. I think it would be nice to add to README some info about the book - if it will be hosted in the future, it would be nice to add some link, but also info that it exists and some instructions how to build it

To my understanding the whole point of this ticket was to have an in-repo documentation system, which makes sense, since the documentation should be updated together with the code that it documents.

As for the readme, I don't see a point in adding it now since the very next ticket will actually implement it the building into github pages and until then there's no need for anyone to even concern themselves with the book as it is a work in progress.

ThrasherLT:
> I have two things to consider:
> 
>     1. I'm not convinced that we need to keep it inside the repo, it looks that it is designed to be a separate git project, so maybe we could follow this? I think we could even make the sources private as they are not that useful and publish only the book, as this is what user will use.
> 
>     2. I think it would be nice to add to README some info about the book - if it will be hosted in the future, it would be nice to add some link, but also info that it exists and some instructions how to build it

To my understanding the whole point of this ticket was to have an in-repo documentation system, which makes sense, since the documentation should be updated together with the code that it documents.

As for the readme, I don't see a point in adding it now since the very next ticket will actually implement it the building into github pages and until then there's no need for anyone to even concern themselves with the book as it is a work in progress. So out of scope.

packgron:
@stalowyjez Target audiance of this documentation is the consumer of libtelio as a library, main ones being our App developer. As such primary objective of docs is to explain how to use our library. For "how" exactly it works and "why" it works the way it does, we will continue to use confluence.

There few reasons I want to have source of documentation in same place as code (note there will probably be other repo for storing generated HTML):
- Ensuring we can update changes to docs at the same time we are updating the code. (we will append our PR requirements)
- We can generate part of the docs from code, add checks that changes to certain parts in codes are reflect changes in docs.

A good example for this is: https://github.com/helix-editor/helix.

So I'm strongly againt's having a separate repo, as typical human nature granties it will be easier to forget about.

packgron (approved):
+1.0

packgron:
> 2. I think it would be nice to add to README some info about the book - if it will be hosted in the future, it would be nice to add some link, but also info that it exists and some instructions how to build it

We currently don't have a link, then build is added in ci/cd and github pages is created, then yes.
For README content, build steps should be left in README for now. But all of usage docs for now should be removed, and links to docs added instead. But we can do that, then we add actual contents to docs.

ThrasherLT:
> @packgron Ehh, and there will be yet another thing to care about before merging? Maybe it would be better to update it on release? Also, in case of text editor it sounds like it makes sense to have a book about its usage, but imho when you have library like ours it sounds like good API documentation + some nice examples in README (and maybe in rustdocs as well?) is just what you need, like adding a whole book just to store a few usage examples seems a bit like an overkill to me.

1) I don't see how trying to figure out the documentation changes at release after everything is merged is simpler than just adding the documentation changes together with the code changes. If you change the way libtelio is used, then you update the documentation accordingly and have them both get past the PR review process. That's the point for keeping the documentation in the repo to keep the documentation in sync with the changes.

2) With all the routing, feature config, four different supported languages and other telio functionalities there is plenty of content for this book imo.

tomaszklak:
We have a wiki, jira tickets, `cargo doc` and some markdown docs scattered across this repo. Is this new thing will replace one of the old things, or will it be another place where one should search for information? 🙃

packgron:
> We have a wiki, jira tickets, `cargo doc` and some markdown docs scattered across this repo. Will this new thing replace one of the old things, or will it be another place where one should search for information? 🙃

Yes it is indeed another place, but it is different targets.
But yeh now i see how this will easily become aditional pain point :|

packgron:
So after a bit of thinking, @tomaszklak i don't think you picture is compeate.

## Currently we have:
- Markdowns in repo
- RFC/ADRs: for technical decision and documentation in conflunce, as there for time being internal they cannot be moved.
- Jira: for tracking work
- A rearly used internal cargo doc page in git hub
- Cargo docs, that you can generate localy
- A undefined amount of scattered docs in confluence:
   - Telio architectual documents
   - Telio integrational documents (integrating telio, feature config, events)
   - NatLab documentation
   - Random pages about random topics
   - Flow models
- A hidden amount of spread sheets, and technical drawing somewhere in google drives

## And we have two context of operations: Open & Internal

Open:
- markdown in repo
- cargo doc
Internal:
..rest

## What is plan now
Move telio integrational documents into Open context, and create a defined place. (meaning we deprecate what we had in conflunce).

So in this sense i did not see it as something major, but as you pointed we do introduce additional location.

But I think we need to have a better agreement how we interact with documentation in general.
From my perspective we currently trying to put almost evrtyhtin into RFC/ADR's, and more or less don't thinkg about anything else.
So with this i wanted to have a place there we think about consumer of a library, as a non rust developer.
In ideal world i would have everything related to usage of this library in cargo docs.

packgron:
So my question to @stalowyjez @tomaszklak how would you imagine we should deliver telio api documention to C#, Kotilin, Swift, Go, Python user's

If it would be just rust, i would not see any point in this, and just try to officially put everything into docs.rs, but we have a bit more complicated scenario.

packgron:
@Jauler Thoughts :arrow_up:

tomaszklak:
> So my question to @stalowyjez @tomaszklak how would you imagine we should deliver telio api documention to C#, Kotilin, Swift, Go, Python user's
> 
> If it would be just rust, i would not see any point in this, and just try to officially put everything into docs.rs, but we have a bit more complicated scenario.

I'm wondering what it is that `cargo doc` fails to provide that `mdbook` will (what problem are we solving) - it's not explained here nor in the jira ticket.

If all we need to have is an option to generate some nice html from markdown, I think `cargo doc` can handle that (for example see how clap is providing various forms of documentation using it: https://docs.rs/clap/latest/clap/index.html).

packgron:
> I'm wondering what it is that `cargo doc` fails to provide that `mdbook` will (what problem are we solving) - it's not explained here nor in the jira ticket.
> 
> If all we need to have is an option to generate some nice html from markdown, I think `cargo doc` can handle that (for example see how clap is providing various forms of documentation using it: https://docs.rs/clap/latest/clap/index.html).

Problem to solve now, show how to use uniffi genered bindings, so it would be easy for app devs to migrate to v5.0.0
To do that original plan was migrate that integrating telio document to internal docs.
Display code examples / usage in multiple languages (unclear how to do this in rustdocs).

There is one thing, that uniffi does pass generated docs string's to other languages, so something like feature config and event struct will likely will not need extra pages anymore. But this would not show usage examples.

stalowyjez:
@packgron Maybe instead of a book it would be nicer to just add some usage examples (I mean buildable) and point to them in readme? It would be easy to make sure that they are up-to-date as we can just add some CI jobs to build them on each commit and it won't be that simple in case of the book.

packgron:
> But `cargo doc` can display code snippets. If the only problem with this is missing code highlight, I think it's not a big enough issue to justify introduction of yet another documentation system.

So something like [claps cookbook](https://docs.rs/clap/latest/clap/_cookbook/index.html)  
you say would be better?

I would assume it would just confuse app devs:
I imagine structure would then look like:
docs/mod.rs -> general docs
docs/go.rs -> golang examples
docs/cs.rs -> cs examples
docs/kotlin.rs -> kotlin examples
docs/swift.rs -> swift code examples

Well maybe we coud probably just have .md and use #[doc(inlude_str!(./some.md)] pub mod some;

It's seems also possible to hack around and add aditional js, by adding header.html.

To me this feels like we are hacking around the tool, to not use a dedicated one. But maybe it's is nice, since all devs can go to a same place.

tomaszklak:
Yeah, something like what you describe is imo enough as a start. After all, the actual content will be in markdown so **if** we find that `rustdoc` is not enough for us or our users, we should be able to relatively easily migrate to `mdbook`.

packgron:
> @packgron Maybe instead of a book it would be nicer to just add some usage examples (I mean buildable) and point to them in readme? It would be easy to make sure that they are up-to-date as we can just add some CI jobs to build them on each commit and it won't be that simple in case of the book.

I do like this idea, just i feel that it's would take significant amount of time.
If we want to empower others to use this library, they would just need to take example and use it.
There is one unknown here I'm afraid, and is to show API we will need to get some conformations :|.
So for first version we would have to assume person has mesh config / server info's already and valid secret key already.

packgron:
> Yeah, something like what you describe is imo enough as a start. After all, the actual content will be in markdown so **if** we find that `rustdoc` is not enough for us or our users, we should be able to relatively easily migrate to `mdbook`.

Fair point

packgron:
So I think now I'm personally converted that we try to start off from rustdoc, (i guess since we haven't yet published libtelio to docs.rs we can use github pages as stop gap).
- we add in src/doc/mod.rs
- add integration guide docs for langs:  docs/using-<lang>.md
- expose them via mod.rs.
- build and publish them as github page

In future if we are migrating to telio as true SDK, we add example project in multiple languages?

Jauler:
> @Jauler Thoughts ⬆️

So the v5.0 integration itself, IMHO, is not such a big deal, that we would need to introduce extensive per-bindings-language documentation, and for sure, no need to introduce separate tools for handling it. The first iteration of libtelio public interface, generated by UniFFI, is on purpose - almost unchanged. Just some arguments/return values are converted to strong-typed ones, error reporting is improved significantly.

But I would not drop such documentation tools completely. What I mean is, that if the intention is that `mdbook` will be de-facto standard place for writing/publishing/managing documentation -> then it makes sense. But this would mean, that part of the `mdbook` introduction task must be to migrate/unify existing documentation into this place. Which should also include confluence pages (with the exception for RFC's/ADR's sections maybe :thinking:, as there are more LLT and external teams involved in those, and we cannot move them without aligning it).

ThrasherLT:
> > Display code examples / usage in multiple languages (unclear how to do this in rustdocs).
> 
> But `cargo doc` can display code snippets. If the only problem with this is missing code highlight, I think it's not a big enough issue to justify introduction of yet another documentation system.

@tomaszklak Doc rs can't create separate pages. So unless we modify the html manually we would either need to cram all documentation on integrating telio into one huge front page essentially burying the actual code documentation or create placeholder modules with documentation instead of code just to separate things out. And to me personally both these approaches look very hacky and unintuitive.

Besides that imported .md files look weird in doc (subheaders for example) unless we use the doc rs specific documentation format and write all the info directly into our code files which would be a pain to migrate later (and to rewrite now).

tomaszklak:
What can I say, both `rustfmt` and `mdbook` have pros and cons. At least for clap developers `rustfmt` seems to be enough so maybe the 'cleanliness' is subjective ;>

Whichever we choose, I think the more important part is the writing of the actual content, no the particular tool that will render markdown. Which is why I think is better to first write the the content (mvp) using our current tools, instead of commiting ahead of time to a specific new solution (especially, since as I pointed to in my first message, this wouldn't replace any of the old tools but instead would be yet another one).

This is obviously not the hill I'm willing to die on so if you strongly disagree, feel free to start using mdbook 😄

ThrasherLT:
> What can I say, both `rustfmt` and `mdbook` have pros and cons. At least for clap developers `rustfmt` seems to be enough so maybe the 'cleanliness' is subjective ;>
> 
> Whichever we choose, I think the more important part is the writing of the actual content, no the particular tool that will render markdown. Which is why I think it's better to first write the the content using our current tools (mvp), instead of commiting ahead of time to a specific new solution (especially, since as I pointed to in my first message, this wouldn't replace any of the old tools but instead would be yet another one).
> 
> This is obviously not the hill I'm willing to die on so if you strongly disagree, feel free to start using mdbook 😄

@tomaszklak I'm also not sure on the whole decision between the two, but I also wasn't aware of how Clap's documentation was structured and was instead relying more on Tokio style documentation where doc.rs contains the API and internal module docs while www.tokio.rs contains all the tutorials and instructions on getting started. I think the "quick links" and underscored documentation modules clean up the solution quite nicely, so I'll try and rework my doc rs example in the style of Clap's doc rs. I was actually really hoping that mdbook could replace Confluence wiki tho, but I guess that's out of scope right now :D

.gitignore
```diff
@@ -42,3 +42,4 @@ beta-events.db
 gamma-events.db
 nat-lab/**/report.html
 events-moose.log
+telio-book/book
```
packgron, line 45: I guess this is not needed anymore

src/lib.rs
```diff
@@ -1,5 +1,14 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+// Telio doc rs documentation pages:
+#![doc = include_str!["./doc/introduction.md"]]
```
packgron, line 4: Move this to the top, to not confuse the eye it's a doc for module

src/doc/introduction.md
```diff
@@ -0,0 +1,158 @@
...
+See [this documentation module][_telio_integration_documentation] for notes on how to integrate Libtelio into an application using one of the different supported languages.
+
+### Initializing the telio device
+
+The main component of the Libtelio library is the `telio::Device` structure and its methods.
+Let's go through the most important ones.
+
+The first of them is the expected function `::new`.
+It takes three arguments, let's describe them briefly:
+ - `features` - tells Libtelio which of the additional features should be enabled - the
```
packgron, line 27: here we can add link to features struct

src/doc/introduction.md
```diff
@@ -0,0 +1,158 @@
...
+ description of them is out of the scope of this README
+ - `event_cb` - event handler, takes a `Box<Event>`, which will be called by Libtelio to handle
```
packgron, line 29: link to event type ?

src/doc/integrating_telio.md
```diff
@@ -0,0 +1,262 @@
+# Documentation For Integrating Telio Into Other Languages
```
packgron, line 1: This weill need a bit more rework, so for now leave as empty, and let's introduce correct version from start

src/doc/telio_events.md
```diff
@@ -0,0 +1,149 @@
...
+
+- level:
+
+    - critical -  telio panicked, so it must be restarted calling telio_destroy_hard() and starting init procedure from scratch (invoking telio_new() or similar path). Note, that destroying and re-creating telio should not be pe3rformed while handling incoming event (not on telio thread’s context).
+
+## Telio Node Event
+
+Once registered using telio_new for events. One of the possible reported event is of type “node“, in format:
+
+```json
```
packgron, line 51: link to actual struct in code?

src/doc/telio_events.md
```diff
@@ -0,0 +1,150 @@
...
+
+    - noerror - default value, not an error.
+    - unknown - unknown error
+
+- msg - custom error string, can be empty
+
+
+### When error event is reported?
+
+One of the situations, when error event will be reported, is when panic happens on telio side.
```
stalowyjez, line 39: Hmm, what about other situations?

.gitignore
```diff
@@ -41,4 +41,4 @@ alpha-events.db
 beta-events.db
 gamma-events.db
 nat-lab/**/report.html
-events-moose.log
+events-moose.log
```
packgron, line 44: leave that end empty endline alone :D

ThrasherLT: Sowy, brought it back :/

packgron (approved):
+1.0

stalowyjez (commented):
Looks ok to me, +1

src/doc/telio_events.md
packgron: Currently it's the only case there error is reported via handler.

Other minnor ones are reported by ffi it self

src/doc/telio_events.md
```diff
@@ -0,0 +1,150 @@
...
+    "allowed_ips": [ "<CIDR>" ],
+    "endpoint":"<ip:port>",
+    "hostname":"<Node hostname>",
+    "path": "<direct|relay>",
+  }
+}
+```
+
+
+Fields:
```
tomaszklak, line 72: nit: Since we already document fields of the `Node` structure and link to it here, I'm not sure we need to duplicate that here.

src/doc/telio_events.md
```diff
@@ -0,0 +1,150 @@
...
+  "public_key": string,
+  "weight": int,
+  "use_plain_text": bool,
+  "conn_state": "<disconnected|connecting|connected>"
+}
+```
+
+### When node event is reported?
+
+A relay event is triggered on when it changes state or we start connecting to another relay server.
```
tomaszklak, line 150: nit: This sounds to me a little bit strange - something like 'we report an event, when the event changes'. I think it would be better to explicitly state that we report it when we change servers or when the configuration of the server changes.

tomaszklak:
The commit message needs to be updated.

tomaszklak (approved):
+1, I think it looks nice and I really like that we can link to the rust types for anyone that is curious :)

stalowyjez (approved):

packgron (approved):

PR #462, opened 2024-04-03
build: Set libtelio as cdylib crate type

lcruz99:
### Problem
This PR is part of the set of tasks to reduce the application binary sizes. For that, dynamic libraries are necessary to achieve isolation and the ability to use size optimization features that otherwise would be locked, due to possible c symbol collisions.

### Solution
It replaces static for dynamic libs (*.so, *.dylib) on linux and apple builds. Android, Windows builds are already compiling dynamic libraries.

It also removes bitcode from apple binaries and add linker flag `-s` for apple builds as this flag is now removed from `rust_build_utils` repo in order to allow future customizations.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -36,3 +36,6 @@ linker = "linkers/d-x86_64-linux-android21-clang"
 
 [target.i686-linux-android]
 linker = "linkers/d-i686-linux-android21-clang"
+
+[target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos", target_os = "tvos-sim"))']
+rustflags = ["-C", "link-arg=-s"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/NordSecurity/libtelio"
 normal = ["wasm-bindgen"]
 
 [lib]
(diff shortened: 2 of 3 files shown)


LukasPukenis (commented):
+1

jjanowsk:
> `It also removes bitcode from apple binaries and add RUST_FLAG opt-level = "s" for release builds as this flag is now removed from rust_build_utils repo in order to allow future customizations.`

This is incorrect. You've removed `-C link-arg=-s` from rust_build_utils which is a linker flags which stands for: "Omit all symbol information from the output file."
And here you've introduced size optimization level. I would not do that without first measuring performance impact of this flag, especially that rust documentation warn about possible performance degradation when used alone:

> These two optimization levels ("s" and "z") greatly reduce LLVM's inline threshold, a metric used to decide whether to inline a function or not. One of Rust principles are zero cost abstractions; these abstractions tend to use a lot of newtypes and small functions to hold invariants (e.g. functions that borrow an inner value like deref, as_ref) so a low inline threshold can make LLVM miss optimization opportunities (e.g. eliminate dead branches, inline calls to closures).

lcruz99:
@jjanowsk thanks! I've overlooked this, thought was the same flag.

PR #463, opened 2024-04-04
Draft: Resolve LLT-4644: Fix wintun adapter creation

gytsto:
### Problem
Sometimes when process that created TUN crashes, leaves adapter in windows registry, even tho it is not showed anywhere else.

### Solution
If TUN with same GUID exists, on first error, try to create TUN one more time, because after first error Wintun itself does some
cleanup of orphaned adapters and next try might be successful


wireguard-go-rust-wrapper/wireguard-go/platform_specific_windows.go
```diff
@@ -12,8 +14,39 @@ import (
 	"crypto/sha256"
 
 	"github.com/google/uuid"
+	"golang.org/x/sys/windows/registry"
 )
 
+func CheckIfAdapterExists(guid string) bool {
+	regKey, err := registry.OpenKey(registry.LOCAL_MACHINE, `SYSTEM\CurrentControlSet\Control\Class\{4d36e972-e325-11ce-bfc1-08002be10318}`, registry.ALL_ACCESS)
```
matislovas, line 21: Is that key (`d36e972-e325-11ce-bfc1-08002be10318`) universal on all Win versions/computers?

wireguard-go-rust-wrapper/wireguard-go/platform_specific_windows.go
```diff
@@ -24,10 +57,21 @@ func PlatformSpecific_CreateTun(ifname string) (*tun.NativeTun, error) {
 	networkProfileGuidString := "{" + networkProfileUuid.String() + "}"
 	networkProfileGuid, _ := windows.GUIDFromString(networkProfileGuidString)
 
+	exists := CheckIfAdapterExists(networkProfileGuidString)
+
 	infof("Creating wintun interface %s with GUID %s", ifname, networkProfileGuidString)
 	wintun, err := tun.CreateTUNWithRequestedGUID(ifname, &networkProfileGuid, 0)
 	if err != nil || wintun == nil {
-		return nil, err
+		// If TUN with same GUID exists, try to create TUN one more time,
```
LukasPukenis, line 65: It was mentioned that if you create and it fails, Windows does some cleanup and the second time it might succeed. Could it be a bit simpler to just try two times instead?

Other than that we have `exists` already, maybe we can destroy it beforehand?

wireguard-go-rust-wrapper/wireguard-go/platform_specific_windows.go
```diff
@@ -12,8 +14,39 @@ import (
...
+	"golang.org/x/sys/windows/registry"
 )
 
+func CheckIfAdapterExists(guid string) bool {
+	regKey, err := registry.OpenKey(registry.LOCAL_MACHINE, `SYSTEM\CurrentControlSet\Control\Class\{4d36e972-e325-11ce-bfc1-08002be10318}`, registry.ALL_ACCESS)
+	if err != nil {
+		return false
+	}
+
+	subkeyNames, err := regKey.ReadSubKeyNames(1000)
```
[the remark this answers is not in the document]
gytsto: just the amount of subkeys to read, but it reads until eof, so 1k is just to be "safe"

wireguard-go-rust-wrapper/wireguard-go/platform_specific_windows.go
gytsto: Wintun does some sort of orphaned interface cleanup, but since we do not use wintun directly, instead we use wireguard-go, which doesn't expose that cleanup function, I just let wintun itself to handle this thing.

LukasPukenis (commented):
+1 but natlab test would be really really good

matislovas:
+1 LGTM

matislovas (approved):

nat-lab/tests/test_wg_adapter.py
```diff
@@ -0,0 +1,80 @@
...
+else:
+    from python_wireguard import Key  # type: ignore
+
+
+async def test_wg_adapter_cleanup():
+    output_notifier = OutputNotifier()
+
+    async def output_checker(stdout: str) -> None:
+        for line in stdout.splitlines():
+            print(line)
```
LukasPukenis, line 21: This print could come with some extra context to make it easier to find if it's intentional to be here

LukasPukenis (commented):
The testcase might not look that great because of flakyness but now it can be improved upon in the future if needed. Looks good enough :)

nat-lab/tests/test_wg_adapter.py
gytsto: But it is not that hard to find that print, since this test doesn't come with all additional clutter from telio class, also only 1 node, and also not much is printed there, since we just start and stop tcli immediately

LukasPukenis: If it's easy to find in the logs then it's fine :)

LukasPukenis (approved):

PR #466, opened 2024-04-04
Use hash_one to simplify hashing in LruCache::entry

devzbysiu:
### Problem
Hi :wave:, just small change - there is no need to create hasher to hash only one value

### Solution
1. Use `hasher().hash_one(..)` which does the same thing as previous code

- [x] README.md is updated
- [x] changelog.md is updated


gytsto:
Seems like it was implemented by https://github.com/NordSecurity/libtelio/pull/475, therefore I am closing it.

PR #464, opened 2024-04-04
build: Update h2 dependency to patch RUSTSEC-2024-0332

lcruz99:
### Problem
`cargo-deny` job started to fail due to RUSTSEC-2024-0332 found in the `h2` crate.

### Solution
Upgrade crate: `h2 v0.3.24` -> `h2 v0.3.26`

### :ballot_box_with_check: Definition of Done checklist
- [x] Commit history is clean ([requirements](../blob/main/docs/git_commit_messages_requirements.md))
- [x] README.md is updated
- [x] changelog.md is updated
- [x] Functionality is covered by unit or integration tests

PR #467, opened 2024-04-05
LLT-5040 Fix nat-lab issues around dockers

jjanowsk:
Bug no. 1: Bug in docker compose: https://github.com/docker/compose/issues/11510
This bug randomly causes missing network interfaces in started containers. We just updated docker compose version to the version which is not affected.

Bug no. 2: Bug in docker daemon: https://github.com/moby/moby/issues/47619
Because of this bug eth0 is not created before sysctls are being set so it is not possible to set any sysctls related to the specific interfaces.
We can create a workaround by not setting eth0. Instead we can set default so eth0 will have this value assigned later when created.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.0.1
+          triggered-ref: v1.0.2
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v1.0.1
+    branch: v1.0.2
--- a/nat-lab/bin/configure_route.sh
+++ b/nat-lab/bin/configure_route.sh
@@ -6,42 +6,9 @@ print_help() {
     exit 1
 }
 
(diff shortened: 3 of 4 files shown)


arg0d (dismissed):
Looks good.

Cargo.lock
```diff
@@ -1320,18 +1320,18 @@ dependencies = [
 
 [[package]]
 name = "enum-map"
-version = "2.7.2"
+version = "2.7.3"
```
arg0d, line 1323: Is it necessary to update `Cargo.lock`?

matszczygiel (approved):
+1

Cargo.toml
```diff
@@ -21,7 +21,7 @@ cfg-if = "1.0.0"
 ffi_helpers = "0.3.0"
 num_cpus = "1.15.0"
 wasm-bindgen = "=0.2.83" # Only for compatability with moose wasm version
-h2 = "=0.3.24" # RUSTSEC-2024-0003
+h2 = "=0.3.26" # RUSTSEC-2024-0332
```
tomaszklak, line 24: that is unrelated change, already being reviewed in #464

tomaszklak (changes requested):
I think security related changes should have separate PRs

Cargo.toml
jjanowsk: So you would now recommend everyone to wait for #464 to be merged and block any other PRs? I don't see a point. For me it should be whoever merges it first. It's in a separate commit with a description. I don't see why it must also be in a separate PR.

tomaszklak: I don't think that this is urgent enough to make a mess.

packgron (approved):
+1.0

Cargo.lock
jjanowsk: It's outdated, but the answer is: yes

tomaszklak (approved):
+0.5

PR #470, opened 2024-04-08
build: Enable size reduction compiler flags

lcruz99:
# Problem
After switching to dynamic linked libraries, we're able to reduce the size of the binaries even further by enabling optimization flags on the compiler.

## Optimization experiments
For convenience, builds were made locally on a linux x86_64 machine, comparing with CI there's an error tolerance of +/- 1%.
Also, by default debug symbols are split into a separate file (329 MB) and stripped from the final binary (ie. `strip = "debug"`).

| linux-x86_64 optimizations | size (MB) | reduction gain (%) | Description                                                                                                      |
| -------------------------- | --------- | ------------------ | ---------------------------------------------------------------------------------------------------------------- |
| opt-level = 3 (default)    | 22.8      | -                  | Optimize for speed and remove debug symbols. |
| strip = "symbols"          | 17        | 25.4               | Remove all symbols.                                               |
| strip = "none"             | 345       | -1413              | Don’t remove any symbols.|
| opt-level = "z"            | 22.1      | 3.1                | Optimize for binary size, but also turn off loop vectorization.|
| opt-level = "s"            | 20.5      | 10                 | Optimize for binary size.                              |
| lto + codegen-units = 1    | 17.2      | 24.5               | Control’s LLVM link time optimizations (LTO). Produces better optimized code at the cost of longer linking time. |
| panic = "abort"            | 16.5      | 27.6               | Terminate the process upon panic. Disables unwinding/backtrace|

Combining the best results from the table above:

| linux-x86_64 optimizations                                                      | size (MB) | reduction gain (%) |
| ------------------------------------------------------------------------------- | --------- | ------------------ |
| opt-level = "s" + lto + codegen-units = 1                                       | 16.4      | 28                 |
| opt-level = "s" + lto + codegen-units = 1 + panic = "abort"                     | 13.3      | 41.6               |
| opt-level = "s" + lto + codegen-units = 1 + panic = "abort" + strip = "symbols" | 9.3       | 59                 |

### Nat-lab tests execution time

To benchmark the closest scenario with the real usage of the library, 3 different tests were chosen where direct connection, reconnections, network switches and moose database dumps are performed. These tests were repeated 20 times and the Mean, Median and Standard Deviation of the execution times were computed.

**test_direct_working_paths_are_reestablished**

| Configuration | Mean  (s)  | Median (s)  | Standard Deviation (s) |
|---------------|---------|---------|--------------------|
| opt-level = 3 (default) | 59.75  | 59.83  | 0.24               |
| opt-level = "s"         | 59.79  | 59.80  | 0.30             |
| opt-level = "s" + lto = true + codegen-units = 1 + panic = "abort" + strip = "symbols" | 59.67  | 59.64  | 0.23 |

**test_mesh_network_switch_direct**

| Configuration | Mean (s)   | Median (s) | Standard Deviation (s) |
|---------------|--------|--------|--------------------|
| opt-level = 3 (default) | 35.35 | 35.27 | 0.55 |
| opt-level = "s"         | 35.39 | 35.14 | 0.60 |
| opt-level = "s" + lto = true + codegen-units = 1 + panic = "abort" + strip = "symbols" | 35.30 | 35.17 | 0.61 |

**test_lana_with_disconnected_node**

| Configuration | Mean (s)   | Median (s)  | Standard Deviation (s) |
|---------------|---------|---------|--------------------|
| opt-level = 3 (default) | 42.03  | 41.98  | 0.61               |
| opt-level = "s"         | 41.40  | 41.37  | 0.36               |
| opt-level = "s" + lto = true + codegen-units = 1 + panic = "abort" + strip = "symbols" | 41.81  | 41.95  | 0.55 |

### Throughput
With two peers on a meshnet with a direct connection and using Boringtun adapter, Iperf3 tool was used to perform throughput tests for 60 seconds, using unlimited bandwitdh on UDP protocol:

| Configuration | Avg. Bitrate (Mb/s)   | Avg. Jitter (ms)  | Avg. Loss (%) |
|---------------|---------|---------|--------------------|
| opt-level = 3 (default) | 727 | 0.021  | 17               |
| opt-level = "s"         | 725  | 0.031  | 17               |
| opt-level = "s" + lto = true + codegen-units = 1 + panic = "abort" + strip = "symbols" | 733 | 0.022  | 17 |

## Conclusion
Tests reveal that not always the more extreme optimizations yield the best results, with rust versions this can change, however with the current rust version it’s shown that `opt-level = "s"` is outperforming `opt-level = "z"`.

The combination of best optimizations is `opt-level = "s" + lto + codegen-units = 1 + panic = "abort" + strip = "symbols"` and we see no relevant differences in terms of both throughput and performance metrics.

By default the debug symbols were already being stripped and split into a separate file. The optimization gains from removing all other symbols are not that significant against the information that would be lost, the team decided to keep as is and having only the debug symbols stripped.

`panic = "abort"` will also be kept out, as not only the stack unwinding is valuable in case of errors, the process is immediately terminated and it may results in desctructors not being called.

Therefore, the optimizations that best represent the chosen trade-offs are `opt-level = "s" + lto + codegen-units = 1`, while keeping the debug symbols on a separate file. This represents a reduction of 27% (22.3 MB → 16.4 MB) on the dynamic libraries.

# Solution
Given the above conclusion, this PR adds the following flags to `release` builds:
```
opt-level = "s"
lto = true
codegen-units = 1
```

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,3 +182,8 @@ telio-utils = { version = "0.1.0", path = "./crates/telio-utils" }
 telio-wg = { version = "0.1.0", path = "./crates/telio-wg" }
 telio-pq = { version = "0.1.0", path = "./crates/telio-pq" }
 telio-pmtu = { version = "0.1.0", path = "./crates/telio-pmtu" }
+
+[profile.release]
+opt-level = "s"
+lto = true
+codegen-units = 1
--- a/changelog.md
+++ b/changelog.md
@@ -33,6 +33,7 @@
 * LLT-4858: NAT state analytics reporting
 * LLT-5038: Add doc rs as in-repo documentation system skeleton
 * LLT-5034: Upgrade rust toolchain to v1.77.2
+* LLT-4924: Enable size reduction compiler flags
 
 <br>
 


tomaszklak:
Can you add here some numbers to back this up? 

1. What is the different in size?
2. What is the performance penalty we have to pay?

LukasPukenis:
Wow, great writeup +1

mathiaspeters:
I agree with Lukas! +1

lcruz99:
Thank you! However @tomaszklak asked me with which WG implementation I have ran the throughput tests, making me aware that I should have tested with `boringtun` instead of `linux_native` since we do packet processing in userspace.

I'll repeat the throughput tests and post the results.

jjanowsk:
Nice one! Thanks! 1 question and one comment:

**Question**: What is the compilation time difference without those 3 flags and with them?

**Comment**: 

> After switching to dynamic linked libraries, we're able to reduce the size of the binaries even further

and
> This represents a reduction of 85% (113 MB → 16.4 MB) against the statically linked binaries 

If anything using dynamically linked libraries will only increase resulting binary size. .so is smaller when you compare it to the .lib file but remember that you need to ship it. Symbols and other data that are needed for linking are removed from the final executable when it is linked, so most of the time final executable linked with static lib will be smaller then one linked with the .so + the size of the .so itself, since you need to ship it next to the exe.

So I understand that taking .so direction was driven by the desire to be able to have different versions of the rust in separate Nord libraries, not by the size reduction. Otherwise most likely it's counterproductive. Especially that when all libraries are of the dynamic type we are now shipping multiple instances of rust standard library with every one.

lcruz99:
> Question: What is the compilation time difference without those 3 flags and with them?

| Optimization flags | Duration (s) |
|--------|--------|
| `opt-level = 3` | 73 |
| `opt-level = "s" + lto + codegen-units = 1 ` | 65 | 
|opt-level = "s" + lto + codegen-units = 1 + panic = "abort" + strip = "symbols"| 60 |

@jjanowsk One motive to ship dynamic linked binaries is indeed binary reduction. Through dynamic linking we can unify the different binaries and ship only one instance of the same library (eg: libstd). Also now we can enable the optimization features that otherwise would be blocked due to c symbol collisions.

jjanowsk:
Thanks for the comilation times. It's nice that they decreased. I was expecting them to increase, but it seems that disabling optimizations during the compilation outweighs the cost of lto and codegen-units=1.

> Through dynamic linking we can unify the different binaries and ship only one instance of the same library (eg: libstd)

But it works in exactly the opposite way ;)
By building as dyn lib, you do not magically change how dependencies are linked. Let's stick to this rust std example. When we ship 2 libraries as static lib e.g. telio and drop, they will both contain parts of it, but when the final executable is linked, duplicated code will be thrown away. When you ship dynamic libraries, they are already linked (and rust std is linked statically into it). They contain all the code that they need, and the linker will never touch them again. (I mean, it will read them but never modify them). So both libraries can contain instances of the same code. 

Again I don't have numbers yet for our case, I don't know the impact it makes, but I just wanted to point out that firstly writing: "we switched to dyn libs from static to save size" is like saying we put him on 10kkcal diet so he can loose some weight, and secondly that comparing staticlib size to dynamic lib size is completely meaningless (assuming our main goal is to decrease the size of the final executable for the customer, not dependencies we ship to the app teams)

lcruz99:
Think I understood your POV now, and why do you say is meaningless to compare these sizes with the static libs. I also don't know what will this changes translate to in the final executable size, but will clarify this..

All in all, you can consider for this PR scope the comparison between the different size optimization flags, these are still valid.

jjanowsk (dismissed):
+1 for those changes anyways

jjanowsk (approved):

tomaszklak (approved):
+1, really nice discussion of results in PR description!

PR #471, opened 2024-04-08
Fix derp disconnect

gytsto:
Wait for derp connection join handle to finish when aborting. Not waiting sometimes prevents some other tasks from stopping. It is because of strong pointer being still held.

### Problem
Derp holds strong pointer to aggregator, which hold strong pointer to wireguard. When stopping Derp, strong pointer of aggregator not released on time, therefore wireguard pointer from aggregator is not released also, preventing wireguard task from stopping.

### Solution
Wait for derp connection join handle to finish when aborting.

- [x] README.md is updated
- [x] changelog.md is updated


packgron (approved):
+1.0

PR #474, opened 2024-04-09
Make connection state reporting unaffected by system time changes

tomaszklak:
### Problem
Changing time by few minutes, caused telio to report disconnect (even if the underlying wireguard connection was fine).

### Solution
Instead of using a system time based 'last handshake time', track the time (using monotonic clock) of the last received bytes. This way changes to the system time (by the user or ntp like service) will not affect our perception of how long ago did we last seen a packet from the other side.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -605,7 +605,7 @@ impl State {
 
             // Node is new and default LinkState is down. Save it before sending the event
             self.no_link_detection
-                .insert(key, None, None, PeerState::Connecting);
+                .insert(key, peer.rx_bytes, peer.tx_bytes, PeerState::Connecting);
 
             self.send_event(
                 PeerState::Connecting,
@@ -743,6 +743,9 @@ impl State {
 
     #[allow(mpsc_blocking_send)]
     async fn update(&mut self, mut to: uapi::Interface, push: bool) -> Result<bool, Error> {
+        for (pk, peer) in &mut to.peers {
+            peer.time_since_last_rx = self.time_since_last_rx(*pk);
+        }
         // Diff and report events
 
         // Adapter doesn't keep track of mesh addresses, therefore they are not retrieved from UAPI requests
@@ -759,7 +762,7 @@ impl State {
         // Because uapi_request() now needs a mut self, the mut borrow for self.uapi_request()
         // would collide with HashSet<&PublicKey> and with from=&self.interface,
         // which was used throughout this function for better readability.
-        if self.interface == to {
+        if Self::is_same_interface(&self.interface, &to) {
             return Err(Error::InternalError(
                 "Same interface received in update function",
             ));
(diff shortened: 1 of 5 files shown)


crates/telio-wg/src/wg.rs
```diff
@@ -817,6 +820,43 @@ impl State {
...
+    /// This is like the normal '==' but ignores the time_since_last_rx field
+    fn is_same_interface(lhs: &Interface, rhs: &Interface) -> bool {
+        fn is_same_peer(lhs: &Peer, rhs: &Peer) -> bool {
+            lhs.public_key == rhs.public_key
+                && lhs.endpoint == rhs.endpoint
+                && lhs.ip_addresses == rhs.ip_addresses
+                && lhs.persistent_keepalive_interval == rhs.persistent_keepalive_interval
+                && lhs.allowed_ips == rhs.allowed_ips
+                && lhs.rx_bytes == rhs.rx_bytes
+                // this is omitted by design: && lhs.time_since_last_rx == rhs.time_since_last_rx
```
gytsto, line 833: why is that?

tomaszklak: Since we use monotonic clock ([Instant](https://doc.rust-lang.org/std/time/struct.Instant.html#)) to compute the duration which is stored in this field, each time we try to get the current value will will **always** get a different bigger value. A playground [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=841f64816fd976e057bdd6db976087df).

gytsto: Ah, now I see. Before I tought it was set somewhere in lower lever and just stored in the variable, which would later be used in `update` function.

gytsto (commented):
other than failing test, lgtm +1

nat-lab/tests/test_proxy.py
```diff
@@ -31,7 +31,7 @@ async def test_proxy_endpoint_map_update() -> None:
         )
 
         async with Ping(alpha_connection, beta.ip_addresses[0]).run() as ping:
-            await testing.wait_long(ping.wait_for_next_ping())
+            await testing.wait_lengthy(ping.wait_for_next_ping())
```
gytsto, line 34: Just curious, why such big of a difference from main branch?
`this branch`: 19 seconds
`main branch`: 11 seconds

tomaszklak: Good question! I can get it back to ~10s with this little trick:
```
diff --git a/crates/telio-utils/src/exponential_backoff.rs b/crates/telio-utils/src/exponential_backoff.rs
index 275d80b..9e3aa6a 100644
--- a/crates/telio-utils/src/exponential_backoff.rs
+++ b/crates/telio-utils/src/exponential_backoff.rs
@@ -88,7 +88,7 @@ impl Backoff for ExponentialBackoff {
     }

     fn next_backoff(&mut self) {
-        self.current_backoff *= EXPONENTIAL_BACKOFF_MULTIPLIER;
+        self.current_backoff = self.current_backoff + (self.current_backoff / 3);

         if let Some(maximal) = self.bounds.maximal {
             if self.current_backoff > maximal {
```
And after some digging it seem to me that this PR introduces some small delay (expected) but it aligns in an unfortunate way some of the other timers in the `proxy.rs` that now jump up to the 4s range thanks to the exponential growth. Imo, this is not a real problem but let me know if you think I should do something here.

gytsto: It's all good, you have a valid explanation. I tought some bug might have been introduced

crates/telio-wg/src/wg.rs
```diff
@@ -605,7 +605,7 @@ impl State {
 
             // Node is new and default LinkState is down. Save it before sending the event
             self.no_link_detection
-                .insert(key, None, None, PeerState::Connecting);
+                .insert(key, peer.rx_bytes, peer.tx_bytes, PeerState::Connecting);
```
Hasan6979, line 608: Why are we saving tx and rx bytes here. I did not understand this.

crates/telio-wg/src/wg.rs
```diff
@@ -817,6 +820,43 @@ impl State {
 
         Ok(success)
     }
+
+    /// This is like the normal '==' but ignores the time_since_last_rx field
+    fn is_same_interface(lhs: &Interface, rhs: &Interface) -> bool {
+        fn is_same_peer(lhs: &Peer, rhs: &Peer) -> bool {
```
Hasan6979, line 826: Better if `is_same_peer` is on the outside, instead of embedded in `is_same_interface`. Makes function look cleaner and easier to read.

tomaszklak: There is no other user/caller of this function - I'm not sure if moving it outside to the wider scope is a good idea. I can do it if you are strongly convinced that it will be better - let me know.

crates/telio-wg/src/wg.rs
tomaszklak: This is a simplification - instead of assuming that the wireguard always starts with `None` for a fresh peer, we use the actual value as returned by the wireguard implementation (even if most of the time wireguard will indeed start from `None`).

crates/telio-wg/src/wg.rs
Hasan6979: I feel it would be a better idea to move it out.
1) Aesthetics
2) It isn't used by any other function now. It might be in the future.

What do you think ?

tomaszklak: 1. That is pretty subjective 😄 
2. If in future someone wants to use it, they can easily extract it. Until that time it's I think easier to reason about the code when the scope of visibility is smaller.

Despite what I wrote above, this is in the end imo very minor issue so I will move that function to the outside :)

gytsto: I'd leave it as it is :smile: if it will be needed in future, it could be extracted, like @tomaszklak said.

Hasan6979 (approved):
+1.0

PR #476, opened 2024-04-10
LLT-5078 proxy panics when being killed

matislovas:
### Problem
`tokio::select!` can panic, if branches are disabled. This was observed when killing `proxy` task. 

### Solution
Update all (exception - tests) `tokio::select!` invokations with `else`, to avoid panics.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+### v4.3.2
+### **Bravas**
+---
+* LLT-5078: Fix telio-proxy panic 
+
+<br>
+
 ### v4.3.1
 ### **Bravas**
 ---
--- a/crates/telio-proxy/src/proxy.rs
+++ b/crates/telio-proxy/src/proxy.rs
@@ -190,6 +190,12 @@ impl UdpProxy {
         let _ = self.task_ingress.stop().await.resume_unwind();
     }
 
+    #[cfg(test)]
+    pub async fn stop_reverse(self) {
+        let _ = self.task_egress.stop().await.resume_unwind();
+        let _ = self.task_ingress.stop().await.resume_unwind();
+    }
+
     /// Notify proxy about network change
     pub async fn on_network_change(&self) {
         let _ = task_exec!(&self.task_egress, async move |state| {
@@ -463,7 +469,9 @@ impl Runtime for StateIngress {
 
                 let _ = self.sockets.insert(pk, sock);
             }
-
+            else => {
+                telio_log_warn!("StateIngress: no events to wait on ...");
+            },
         }
 
         Self::next()
@@ -523,6 +531,9 @@ impl Runtime for StateEgress {
             (pk, _, _)= update => {
                 self.replace_socket(pk).await;
             },
+            else => {
+                telio_log_warn!("StateEgress: closed input channel");
+            },
         }
         Self::next()
     }
@@ -545,6 +556,38 @@ mod tests {
         .await;
 
         ts.stop().await;
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_graceful_shutdown() {
+        use telio_crypto::SecretKey;
+
+        for i in 0..1 {
+            let TestSystem { wg, proxy, relay } = TestSystem::start().await;
+
+            drop(relay);
+
+            let mut pks = Vec::new();
+            for _ in 0..2 {
+                let pk = SecretKey::gen().public();
+                pks.push(pk);
+            }
+
+            proxy
+                .configure(Config {
+                    wg_port: Some(wg.addr().port()),
+                    peers: pks.iter().copied().collect(),
+                })
+                .await
+                .expect("UdpProxy::configure() failed");
+
+            // Shouldn't panic in any order
+            if i % 2 == 0 {
+                proxy.stop().await;
+            } else {
+                proxy.stop_reverse().await;
+            }
+        }
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 3)]


crates/telio-proxy/src/proxy.rs
```diff
@@ -463,7 +463,7 @@ impl Runtime for StateIngress {
 
                 let _ = self.sockets.insert(pk, sock);
             }
-
+            else => pending().await,
```
tomaszklak, line 466: this will never finish - is that what we really want? I would have expected to see maybe an empty branch `else => {}` or maybe in addition `telio_log_warn` in it?

tomaszklak:
Please add tests.

crates/telio-proxy/src/proxy.rs
packgron: I'm debaiting this my self, in general this expected, as we do not lock the order of ingress and egress closing. So we would make QA thing that some error happaned? but warn is not that bad.

empty can cause looping, as closed channel continiously returnes None

crates/telio-relay/src/derp/proto.rs
```diff
@@ -244,6 +244,7 @@ pub async fn start_write<W: AsyncWrite + Unpin>(
                     break;
                 }
             }
+            else => break,
```
packgron, line 247: else not possible, there are no additional hands for else, as it receives option's and then matches on them

crates/telio-proxy/src/proxy.rs
packgron: As this is wait, update would cancel it out

packgron: I am thinking if we should use option = future => {
match option {}
}

instead of else, and then have more detailed warns.
I would be more allarmed on socket poll failure in comparison to socket_update

lcruz99:
lgtm +1

crates/telio-proxy/src/proxy.rs
matislovas: But will this not invoke the same error (closed branches) as before? I would more for the `else` case with some `telio_log_warn("Dirty exit")` and using `else` generally, this way we can lint out the "else-less" selects.

packgron: So panic is only invoked if there is some match state not covered, if you capture option, it is fully captured, why it failed now is because we had Some(x) = fut => {...}
But did not have None = fut => {..}, and that caused a panic

crates/telio-relay/src/derp/proto.rs
matislovas: Maybe we should leave it, for the sake of to have "else-less" selects ..?

stalowyjez: @matislovas it sounds like adding `_ => {}` to every `match`, imho weird idea. When it is exhaustive without that it sounds like a better design.

crates/telio-proxy/src/proxy.rs
stalowyjez: Yeah, I think that we don't want to lint else-less selects, but non-exhaustive ones. And also pending here looks strange to me, I would just add some log, or just nothing

matislovas:
> Please add tests.

Added unit test.

crates/telio-relay/src/derp/proto.rs
matislovas: @stalowyjez  but in that case, will you be able to lint out such specific cases, where it is prone to panic? It is similar to `match`, except that `match` does not panic :sweat_smile:

crates/telio-proxy/src/proxy.rs
matislovas: Ok, so will leave other `select!`s as it is.

crates/telio-proxy/src/proxy.rs
```diff
@@ -523,6 +531,9 @@ impl Runtime for StateEgress {
             (pk, _, _)= update => {
                 self.replace_socket(pk).await;
             },
+            else => {
+                telio_log_warn!("StateEgress: no events to wait on ...");
+            },
```
tomaszklak, line 536: I think this is not tested:

```
$ git diff
diff --git a/crates/telio-proxy/src/proxy.rs b/crates/telio-proxy/src/proxy.rs
index 83b84844..7b1e1693 100644
--- a/crates/telio-proxy/src/proxy.rs
+++ b/crates/telio-proxy/src/proxy.rs
@@ -531,9 +531,9 @@ impl Runtime for StateEgress {
             (pk, _, _)= update => {
                 self.replace_socket(pk).await;
             },
-            else => {
-                telio_log_warn!("StateEgress: no events to wait on ...");
-            },
+            // else => {
+            //     telio_log_warn!("StateEgress: no events to wait on ...");
+            // },
         }
         Self::next()
     }
$ cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.25s
     Running unittests src/lib.rs (/Users/tomaszklak/Development/libtelio/target/debug/deps/telio_proxy-7ef99710b91e3a47)

running 5 tests
test proxy::tests::test_single_endpoint ... ok
test proxy::tests::test_multiple_endpoints ... ok
test proxy::tests::test_graceful_shutdown ... ok
test proxy::tests::test_muting_unmuting ... ok
test proxy::tests::test_error_handling ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.02s

   Doc-tests telio-proxy

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

This is a pretty safe change, so if it's not possible to test this or it's very hard feel free to skip it, but if possible try to add a test that also verifies the egress case.

crates/telio-proxy/src/proxy.rs
```diff
@@ -523,6 +531,9 @@ impl Runtime for StateEgress {
             (pk, _, _)= update => {
                 self.replace_socket(pk).await;
             },
+            else => {
+                telio_log_warn!("StateEgress: no events to wait on ...");
```
stalowyjez, line 535: I think you could improve this message a bit, you don't know whether there are any events when you are waiting.

matislovas: What it should say?

crates/telio-proxy/src/proxy.rs
matislovas: I cannot think of any easy modifications on this, because of `Either<T>` future. It will always have sth to wait on :smile: . So maybe skipping this one ...

tomaszklak (dismissed):
+1

changelog.md
```diff
@@ -1,3 +1,10 @@
+### UNRELEASED
```
packgron, line 1: Since we are not takking anythin extra, i think bump to v4.3 at the same time, maybe just in a sperate commit, so proting would be easier

matislovas: Fixed

packgron (approved):
+1

PR #481, opened 2024-04-15
Convert dns failures with NoError to ServFail

tomaszklak:
### Problem
In debug mode, trust-dns asserts that `NoError` error type is never converted. But in practice seems that we sometimes receive such cases.

### Solution
Convert such errors to `ServFail` to avoid panic.

- [x] README.md is updated
- [x] changelog.md is updated


gytsto (commented):
+1

dfetti (approved):
+1.0

PR #487, opened 2024-04-16
Replace custom tvOS versions of crates, with updated upstream ones

tomaszklak:
### Problem
We are using forked versions of some crates that already contain tvOS support in latest upstream versions.

### Solution
Where possible removed the in house forks of the crates with added tvOS support and replaced with latest upstream versions that already contain tvOS support. Crates replaced with upstream versions:
 - cc
 - if-addrs
 - socket2
 - ring
 - parking_lot_core
 - nix

After this, there are still 2 crates not updated upstream:
 - block
 - objc

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/.cargo/config_apple_tvos
+++ b/.cargo/config_apple_tvos
@@ -1,10 +1,4 @@
 
 [patch.crates-io]
-cc = { git = 'https://github.com/nordsecurity/cc-rs.git', branch = 'add_apple_tvos_support' }
-if-addrs = { git = 'https://github.com/nordsecurity/if-addrs.git', branch = 'add_apple_tvos_support_ifaddrs' }
-socket2 = { git = 'https://github.com/nordsecurity/socket2.git', branch = 'add_apple_tvos_support_0_4_10' }
-ring = { git = 'https://github.com/nordsecurity/ring.git', branch = 'add_apple_tvos_support_0_16_20' }
-parking_lot_core = { git = 'https://github.com/nordsecurity/parking_lot.git', branch = 'add_apple_tvos_support_0.9.8' }
-nix = { git = 'https://github.com/nordsecurity/nix.git', branch = 'add_apple_tvos_support_0_26_4' }
 block = { git = 'https://github.com/nordsecurity/rust-block.git', branch = 'add_apple_tvos_support' }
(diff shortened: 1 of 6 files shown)


jjanowsk:
`objc` and `block` crates have not been maintained for 4 and 7 years respectively, so there is basically no chance that Luis' PRs will ever be merged. But there is an actively maintained project https://github.com/madsmtm/objc2 which is based on those and the owner has already added the support for tvos. So maybe it's a good idea to switch to it in this PR?

tomaszklak:
> `objc` and `block` crates have not been maintained for 4 and 7 years respectively, so there is basically no chance that Luis' PRs will ever be merged. But there is an actively maintained project https://github.com/madsmtm/objc2 which is based on those and the owner has already added the support for tvos. So maybe it's a good idea to switch to it in this PR?

Great idea, but I think it would be better to make it a separate PR and keep this one about just updating to latest and removing not needed forks.

tomaszklak:
> `objc` and `block` crates have not been maintained for 4 and 7 years respectively, so there is basically no chance that Luis' PRs will ever be merged. But there is an actively maintained project https://github.com/madsmtm/objc2 which is based on those and the owner has already added the support for tvos. So maybe it's a good idea to switch to it in this PR?

I don't think this is right now possible. I would have to update both objc and objc-foundation and the foundation crate is not ready yet: https://crates.io/crates/objc2-foundation/versions

tomaszklak:
> And what's wrong with version 0.1.1 of objc2-foundation? I would assume that it is the exact clone of objc-foundation v0.1.1 that we're using now, or not?

it still depends on both crates we want to get rid of:

```
$ cargo tree -p objc2-foundation@0.1.1 --target aarch64-apple-tvos
    Updating git repository `https://github.com/NordSecurity/uniffi-rs.git`
objc2-foundation v0.1.1
├── block v0.1.6
├── objc v0.2.7
│   └── malloc_buf v0.0.6
│       └── libc v0.2.153
└── objc_id v0.1.1
    └── objc v0.2.7 (*)
```

jjanowsk (approved):
+2 Thanks for the explanation

jjanowsk (approved):

PR #489, opened 2024-04-17
LLT-5033 Swift package shipped with dynamic libraries and with uniffi bindings

jjanowsk:
Previously we were building libtelio as a static library for Apple targets. Now we are switching to .dylibs.  This commit fixes xcframework creation. Additionally it also removes old Swift bindings and replaces them with the ones generated by uniffi.

- [x] README.md is updated


.github/workflows/build.yml
```diff
@@ -77,7 +80,10 @@ jobs:
         with:
           key: ${{ matrix.target_os }}-${{ matrix.arch }}
       - name: Build libtelio
-        run: python3 ci/build_libtelio.py build ${{ matrix.target_os }} ${{ matrix.arch }}
+        run: |
+          python3 -m pip install -r requirements.txt || python3 -m pip install --break-system-packages -r requirements.txt
```
packgron, line 84: Why is requirement are needed if you are using pipenv afterwards?

src/libtelio.udl
```diff
@@ -113,7 +113,7 @@ typedef u32 TtlValue;
 [Custom]
 typedef boolean FeatureBoringtunResetConns;
 
-namespace libtelio {
+namespace telio {
```
packgron, line 116: this is to avoid creating liblibtelio ?

.github/workflows/build.yml
packgron, line 84: what root os of this container, that you need a --breal-system-packages tag?

jjanowsk: Android image is based on Debian12 and requires this flag and Linux image is based on Debian11 and does no know it yet.

src/libtelio.udl
jjanowsk: Yes

Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-black = "*"
+black = "==24.1.1"
```
gytsto, line 7: We don't need to lock package here, since it is already locked in Pipfile.lock. Or was it specific reason of doing it?

requirements.txt
```diff
@@ -1 +1 @@
-pipenv == 2023.11.15 --hash=sha256:6eb537da5a27671f71dc596ea96a5b341aae9a1695413beeb95364467409bb3d
+pipenv == 2023.11.15
```
gytsto, line 1: this was intentionally locked to specific hash and it should be updated separately if needed

lcruz99:
could you build `uniffi-generators` docker image also for mac devices?

Pipfile
jjanowsk: I think we do need that :) Pipfile should contain the version hints if we care about the version of particular dependency. In case of black we do care and do not want to update it basically ever. I would argue that for some other we also care about not jumping to the next major versions but I fixed just black since it's the most obvious. I changed it in this PR because I was actually adding docker sdk. I later removed it because it sucked and used other way to run docker (See here: https://github.com/NordSecurity/rust_build_utils/pull/12) but this change is still beneficial so I left it in.

.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.0.3
+          triggered-ref: LLT-5033-dylib-uniffi-xcframework
```
matislovas, line 34: Reminder: change to tag.

requirements.txt
jjanowsk: `pipenv == 2023.11.15` locks to the exact version perfectly fine. The only thing it does not do is to validate the integrity of the version, but I don't think we need that here. Using `--hash` and not specifying hashes of all the dependencies in the `requrements.txt` forces you user to use `--no-deps`, which is wrong. If the user does not have all the dependencies preinstalled they should be allowed to simply use run `pip install -r requrements.txt`.

stalowyjez (commented):
+0.5

matislovas:
+1

packgron (changes requested):
I think we can remove linkers folder now. as it was needed to expose swift c functions.

ci/env.py
```diff
@@ -7,6 +7,7 @@
 LIBTELIO_ENV_ANDROID_BUILDER_TAG = "v0.0.1"
 LIBTELIO_ENV_LINUX_BUILDER_TAG = "v0.0.1"
 LIBTELIO_ENV_WINDOWS_BUILDER_TAG = "v0.0.1"
+LIBTELIO_ENV_UNIFFI_GENERATORS_TAG = "v0.25.0-8"
```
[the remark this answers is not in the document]
jjanowsk: No. Printed are only the ones still needed in build.sh in libtelio-build. This generation is done in python already. Once build.sh is rewritten to python those other prints will be also removed

jjanowsk:
> could you build `uniffi-generators` docker image also for mac devices?

I guess we should. It's a candidate for a separate task

.cargo/config
```diff
@@ -8,16 +8,16 @@ runner = './test_runner.sh'
 git-fetch-with-cli = true
 
 [target.x86_64-unknown-linux-gnu]
```
packgron, line 10: Are these even needed any more? I think now it's all default targets

jjanowsk: At least some of them are, because I saw the build failing without them. TBH I do not want to investigate it now, because I want to merge this one ASAP. I've created LLT-5117 for it

packgron (dismissed):
+1.0

packgron (approved):

PR #490, opened 2024-04-17
Enable nat traversal info validator

matislovas:
### Problem
Lana tests in `nat-lab` had disabled nat-traversal info validator.

### Solution
Enabled nat-traversal info validator and clear a bunch of TODOs. 
Changed nat traversal info validator. Now it standard `StringValidator` with added entries count validation. The reason for this, is that we currently do not have node numbering validator (the same as in connectivity matrix) so we cannot check if, which node is actually included in the nat traversal info entry. 

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/nat-lab/tests/test_lana.py
+++ b/nat-lab/tests/test_lana.py
@@ -123,6 +123,7 @@ async def clean_container(connection):
 
 def get_moose_db_file(container_tag, container_path, local_path):
     subprocess.run(["rm", "-f", local_path])
+    # sqlite3 -bail -batch moose.db "BEGIN EXCLUSIVE TRANSACTION; SELECT 1; ROLLBACK;" > /dev/null
     subprocess.run(
         ["docker", "cp", container_id(container_tag) + ":" + container_path, local_path]
     )
@@ -300,7 +301,7 @@ async def run_default_scenario(
         client_beta.wait_for_state_on_any_derp([telio.State.Connected]),
         client_gamma.wait_for_state_on_any_derp([telio.State.Connected]),
     )
-    # Note: GAMMA is symmetric, so it will not connect to ALPHA or BETA in diret mode
+    # Note: GAMMA is symmetric, so it will not connect to ALPHA or BETA in direct mode
     await asyncio.gather(
         client_alpha.wait_for_state_peer(
             beta.public_key,
@@ -408,6 +409,9 @@ async def run_default_scenario(
         alpha_expected_states,
         beta_expected_states,
         gamma_expected_states,
+        alpha.public_key,
+        beta.public_key,
+        gamma.public_key,
     ]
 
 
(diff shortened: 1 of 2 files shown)


LukasPukenis (commented):
+1 nice cleanup

nat-lab/tests/test_lana.py
```diff
@@ -1710,15 +1685,14 @@ async def test_lana_with_disconnected_node(
...
-            # TODO enable this when NAT traversal reporting is implemented (LLT-4858)
-            # .add_nat_traversal_conn_info_validator(
-            #     exists=False,
-            # )
             .add_derp_conn_info_validator(
                 exists=True,
                 servers=DERP_SERVERS_STRS,
             )
         )
+        add_nat_traversal_info_validator_to_non_symmetric_clients(
```
Hasan6979, line 1693: Keeping up with the above architecture, couldn't we move the condition checking which public key if lower directly into `add_nat_traversal_conn_info_validator` ?

nat-lab/tests/test_lana.py
```diff
@@ -482,6 +485,17 @@ def add_rtt_validators(
             )
 
 
+def add_nat_traversal_info_validator_to_non_symmetric_clients(
+    node_a_validator: EventValidator, node_a_pubkey: str, node_b_pubkey: str
+):
+    if base64.b64decode(node_a_pubkey) < base64.b64decode(node_b_pubkey):
```
Hasan6979, line 491: Just a confirmation, we only add the lower node validator because that will be initiator and will contain the NAT info ?

matislovas: That was a bit weird to me also, but as long as it makes decision between two pubkeys, I'm okay with that :smile:  [ref](https://github.com/NordSecurity/libtelio/blob/730cef24867b23ea84e3168290556f76abba775d/nat-lab/tests/test_direct_connection.py#L401)

nat-lab/tests/test_lana.py
matislovas: Ummm ... I guess I can move that, and it would be logical in some sense, but the question is about the maintenance. This one clearly states, that this for `non_symmetrical_clients`, and someone later can add `add_nat_traversal_conn_info_validator` to all clients (including symmetrical ones) and have a headache after that :smile: .

matislovas:
Added validators for symmetric clients

nat-lab/tests/test_lana.py
Hasan6979: To combine them both, we could add an additional field of `client_type` in `add_nat_traversal_conn_info_validator` which can be `symmetric` or `asymmetric` and further decisions could be based on that. I feel it would be better to deal with both types in a homogeneous way.

matislovas: Fixed

Hasan6979 (approved):
+1.0

PR #492, opened 2024-04-18
[LLT-5082] Use boringtun v0.6.0

stalowyjez:
### Problem
We use boringtun v0.5.2 when v0.6.0 is available

### Solution
Rebase Telio changes to boringtun on v0.6.0 version, use it in libtelio and add some fixes where necessary.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-dns/src/nameserver.rs
+++ b/crates/telio-dns/src/nameserver.rs
@@ -21,9 +21,9 @@ use std::{
     sync::Arc,
 };
 use telio_model::features::TtlValue;
-use tokio::net::UdpSocket;
 use tokio::sync::{RwLock, RwLockMappedWriteGuard, RwLockWriteGuard, Semaphore};
 use tokio::task::JoinHandle;
+use tokio::{net::UdpSocket, sync::Mutex};
 
 use telio_utils::{telio_log_debug, telio_log_error, telio_log_trace, telio_log_warn};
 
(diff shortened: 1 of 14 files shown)


gytsto (commented):
Libtelio changes are ok, on boringtun side I didn't see anything missing compared to `0.5.2` branch. Only thing I don't like, that boringtun version was incremented on patch instead of major/minor, I'd have chosen `v1.2.0` or smth like that.

Besides that, +1.

crates/telio-dns/src/nameserver.rs
```diff
@@ -113,26 +114,34 @@ impl LocalNameServer {
...
+                        let mut last_res =
+                            peer.lock()
+                                .await
+                                .decapsulate(None, &[], &mut temporary_buffer);
+                        while let TunnResult::WriteToNetwork(empty) = last_res {
                             if let Err(e) = socket.send_to(empty, dst_address).await {
                                 telio_log_warn!("[DNS] Failed to send handshake packet  : {:?}", e);
                                 return;
                             };
+                            last_res =
```
Hasan6979, line 141: Why are you decapsulating here twice ?

crates/telio-dns/src/nameserver.rs
```diff
@@ -113,26 +114,34 @@ impl LocalNameServer {
...
                     None,
                     receiving_buffer.get(..bytes_read).unwrap_or(&[]),
                     &mut sending_buffer,
-                ) {
+                );
+                match res {
                     // Handshake packets
                     TunnResult::WriteToNetwork(packet) => {
-                        if let Err(e) = socket.send_to(packet, dst_address).await {
+                        let res = socket.send_to(packet, dst_address).await;
```
Hasan6979, line 125: Why was if let changed here ?

crates/telio-dns/src/nameserver.rs
stalowyjez: When it was done in the while condition the compiler complained about that - we cannot have lock locked while we await and `decapsulate` now needs to be called with lock, so I splitted it to two calls - I guess I could do that with `loop` and `break` if you would like that more, or maybe you have some idea of writing it in a better way?

Hasan6979 (dismissed):
+1

gytsto:
Don't forget to apply `boringtun v1.1.12` changes to `v1.2`

gytsto (approved):

PR #493, opened 2024-04-18
Resolve LLT-5097: update wg go to throw error on uapi get

gytsto:
### Problem
Wireguard-Go implementation doesn't do any errors, when sending UAPI  Get request, when device is down. Not like UAPI Set requests. This is mixing up wg state by saying that device is down while setting something, but saying that device is up by getting something from UAPI.

### Solution
Patch wireguard-go by checking if device is down when doing UAPI Get request.

### Notes
Main changes are in this [commit](https://github.com/NordSecurity/libtelio/pull/493/commits/79134c404482145ce38aca7660504a664ba0503a). Other commits are just wireguard file fetch and update to compile with our `package main`.

Also if you know  better solution on how to patch `uapi.go`, would highly appreciate, since for 3 line changes, adding additional 5k feels so weird. :cry: 

- [x] README.md is updated
- [x] changelog.md is updated


gytsto:
Closing this, will make wireguard-go fork

PR #500, opened 2024-04-22
Add manual QoS trigger in tcli

matislovas:
### Problem
The `QoS` collection is not aligned by events to collecting heartbeat events and/or stopping the nodes, but by `rtt_interval`. Therefore it is very prone to race conditions during the tests.

### Solution
Make a manual `QoS` job trigger in `tcli` and use it in `test_lana_with_disconnected_node` test.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-nurse/src/nurse.rs
+++ b/crates/telio-nurse/src/nurse.rs
@@ -41,7 +41,9 @@ pub struct NurseIo<'a> {
     /// Event channel to gather meshnet config update
     pub config_update_channel: Option<mc_chan::Tx<Box<MeshConfigUpdateEvent>>>,
     /// Event channel to manual trigger a collection
-    pub collection_trigger_channel: Option<mc_chan::Tx<Box<()>>>,
+    pub collection_trigger_channel: Option<mc_chan::Tx<()>>,
+    /// Event channel to manual trigger a qos
+    pub qos_trigger_channel: Option<mc_chan::Tx<()>>,
 }
 
 /// Nurse entity
@@ -138,6 +140,9 @@ impl State {
         let collection_trigger_channel = io
             .collection_trigger_channel
             .unwrap_or_else(|| McChan::default().tx);
+        let qos_trigger_channel = io
+            .qos_trigger_channel
+            .unwrap_or_else(|| McChan::default().tx);
 
         // Heartbeat component
         let heartbeat_io = HeartbeatIo {
@@ -166,7 +171,10 @@ impl State {
 
             Some(Task::start(QoSAnalytics::new(
                 qos_config,
-                QoSIo { wg_channel },
+                QoSIo {
+                    wg_channel,
+                    manual_trigger_channel: qos_trigger_channel.subscribe(),
+                },
             )))
         } else {
             None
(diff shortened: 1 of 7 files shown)


matislovas:
Not sure about the changelog ... do we actually need this inner change documented ..?

crates/telio-nurse/src/qos.rs
```diff
@@ -189,6 +190,13 @@ impl Runtime for Analytics {
                 })
             },
 
+            // Manual trigger
+            Ok(_) = self.io.manual_trigger_channel.recv(), if self.ping_channel_tx.upgrade().is_none() => {
```
packgron, line 194: so it's more like manual ping trigger not, qos trigger.

matislovas: It still collects statistics, so it is kind of qos trigger.

crates/telio-nurse/src/nurse.rs
```diff
@@ -42,6 +42,8 @@ pub struct NurseIo<'a> {
     pub config_update_channel: Option<mc_chan::Tx<Box<MeshConfigUpdateEvent>>>,
     /// Event channel to manual trigger a collection
     pub collection_trigger_channel: Option<mc_chan::Tx<Box<()>>>,
```
packgron, line 44: Also let's switch to (), boxing does not really make sence here,

matislovas: Fixed

gytsto (dismissed):
+1

crates/telio-nurse/src/qos.rs
```diff
@@ -189,6 +190,13 @@ impl Runtime for Analytics {
...
+                self.rtt_interval.reset();
+                self.perform_ping();
+                Self::next()
+            },
+
             _ = self.rtt_interval.tick(), if self.ping_channel_tx.upgrade().is_none() => {
```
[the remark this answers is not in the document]
matislovas: Test is in the building!

crates/telio-nurse/src/qos.rs
```diff
@@ -544,15 +561,21 @@ fn is_from_ignored_peer(event: &AnalyticsEvent) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use core::time;
```
packgron, line 564: should probably be std::time

crates/telio-nurse/src/qos.rs
```diff
@@ -171,6 +172,8 @@ pub struct Analytics {
     ping_channel_rx: mpsc::Receiver<(PublicKey, DualPingResults)>,
     ping_channel_tx: mpsc::WeakSender<(PublicKey, DualPingResults)>,
     buckets: u32,
+    #[cfg(test)]
+    ping_cnt: u32,
```
packgron, line 176: Maybe it would make pore sence to create a TestPing ping_backend?

packgron (commented):
I would prefer that we would use some "test" or mocked ping_backend. But this works for now also, if you leave it add a commenent above ping backend, to remove it, if we ever add more then one ping backend.

crates/telio-nurse/src/qos.rs
matislovas: Shouldn't be here at all :smile:

crates/telio-nurse/src/qos.rs
matislovas: Yeah, it would make sense. But maybe I'll fill a separate task to do that, because this will touch all the tests (we'll need mocked latency to nodes).

packgron: +1

packgron (dismissed):
+1.0

gytsto (approved):

PR #505, opened 2024-04-23
When meshnet is off, use the last known state to determine path type

tomaszklak:
When meshnet is being turned off, save all the needed state so that it's possible to compute the correct (as it was just before turning off) path type of each peer.

### Problem
When reporting events for peers after turning off the meshnet, we always reported path type 'relayed', regardless of what was the actual path type just before turning off the meshnet.

### Solution
When meshnet is being turned off, save all the needed state so that it's possible to compute the correct (as it was just before turning off) path type of each peer.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -80,6 +80,7 @@ use telio_model::{
     },
     mesh::{ExitNode, LinkState, Node},
     validation::validate_nickname,
+    EndpointMap,
 };
 
 pub use wg::{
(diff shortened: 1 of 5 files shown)


src/device/mod.rs
```diff
@@ -863,7 +874,7 @@ impl RequestedState {
 }
 
 impl MeshnetEntites {
```
LukasPukenis, line 876: Can't help but notice a typo in this struct :)

src/device/mod.rs
```diff
@@ -238,7 +245,7 @@ pub struct Entities {
     firewall: Arc<StatefullFirewall>,
 
     // Entities for meshnet connections
-    meshnet: Option<MeshnetEntites>,
+    meshnet: Either<MeshnetEntites, MeshnetEntitiesLastState>,
```
LukasPukenis, line 248: `Either` introduces concepts of `left` and `right` which is not as readable as if it would be an enum IMHO, though the idea of `meshnet` having either entities or last state is very nice in itself even though unexpected(would except it to be saved separately and used as needed) :)

LukasPukenis (commented):
+1 looks pretty good, though I vote for native enum usage instead of `either`

src/device/mod.rs
tomaszklak: > `Either` introduces concepts of `left` and `right` which is not as readable as if it would be an enum IMHO, though the idea of `meshnet` having either entities or last state is very nice in itself even though unexpected(would except it to be saved separately and used as needed) :)

I wanted to avoid a case when it's ambiguous which one to use - entites or the saved state. This way it's impossible to have access to both at the same time.

I will replace either with custom enum with domain specific names.

src/device/mod.rs
tomaszklak: Done

src/device/mod.rs
tomaszklak: Done.

dfetti (approved):
+1.0 Looks good

PR #511, opened 2024-04-24
Move changelog check into dod

ThrasherLT:
### Problem
Changelog CI check is not working, it should fail if no file had been created in the .unreleased directory

### Solution
Fix

- [x] README.md is updated

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,25 +1,21 @@
-name: Check Unreleased
+name: Definition of Done
 on:
   pull_request:
-    types: [opened, edited]
-    branches-ignore:
-      - main
+    types: [opened, synchronize]
 permissions: {}
 
 jobs:
-  changelog:
+  check-changelog:
     runs-on: ubuntu-latest
     # Excluding Dependabot PRs from this check.
     if: github.actor != 'dependabot[bot]'
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
-          fetch-depth: 2
-
+          fetch-depth: 0
       - name: Check for new file in .unreleased
         run: |
-          FILE_COUNT=$(git diff --name-only main..HEAD -- .unreleased | wc -l)
+          FILE_COUNT=$(git diff --name-only ${{github.event.pull_request.base.sha}}..HEAD -- .unreleased | wc -l)
           if [ "$FILE_COUNT" -eq "0" ]; then
             echo "No new file in .unreleased directory. Please add a new file describing your changes."
             exit 1


.github/workflows/dod.yml
```diff
@@ -16,3 +16,19 @@ jobs:
         with:
           dod_yaml: 'dod.yml'
           github_token: ${{ secrets.GITHUB_TOKEN }}
+
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
packgron, line 24: needs a proper hash like upper one!

packgron (dismissed):
+2.0

jjanowsk (changes requested):
1. I would keep it in the separate file because the triggers for those 2 checks should be different. dod should run only on edits of the PRs which are modifications to the PR description, title or base branch, but changelog check should run on: `opened` and `synchronize` which is a change in head branch (e.g. push). In the current form if you forget to include changelog entry and then push a fix, the recheck will not be triggered
2. diff should not have `main` branch hardcoded. It should take `github.event.pull_request.base.ref`

.github/workflows/dod.yml
```diff
@@ -1,7 +1,7 @@
 name: Definition of Done
 on:
   pull_request:
-    types: [opened, edited]
+    types: [edited]
```
jjanowsk, line 4: opened is also fine here. Sorry, I was not clear in my comment.

.github/workflows/changelog.yml
```diff
@@ -1,25 +1,25 @@
...
-          fetch-depth: 2
-
+          fetch-depth: 0
       - name: Check for new file in .unreleased
+        env:
+            BASE_BRANCH: ${{github.event.pull_request.base.ref}}
         run: |
-          FILE_COUNT=$(git diff --name-only main..HEAD -- .unreleased | wc -l)
+          git fetch origin $BASE_BRANCH:$BASE_BRANCH
+          git branch
```
jjanowsk, line 21: is this a debug print?

.github/workflows/changelog.yml
```diff
@@ -1,25 +1,25 @@
...
         with:
-          fetch-depth: 2
-
+          fetch-depth: 0
       - name: Check for new file in .unreleased
+        env:
+            BASE_BRANCH: ${{github.event.pull_request.base.ref}}
         run: |
-          FILE_COUNT=$(git diff --name-only main..HEAD -- .unreleased | wc -l)
+          git fetch origin $BASE_BRANCH:$BASE_BRANCH
```
jjanowsk, line 20: I think with `fetch-depth: 0` all branches should be already fetched, shouldn't they?

.github/workflows/dod.yml
ThrasherLT: But theoretically DoD checks it the checkmarks had been placed in the PR description DoD section, so they shouldn't be present there when the PR had just been opened? Since the only way they can be added is through edits. I'm only not sure about reopened PRs.

.github/workflows/changelog.yml
ThrasherLT: Removed

.github/workflows/changelog.yml
ThrasherLT: Nope. It's still failing:

https://github.com/NordSecurity/libtelio/actions/runs/8829848747/job/24241755695?pr=511

.github/workflows/dod.yml
jjanowsk: But DoD action also places those checkmarks when the PR is opened

.github/workflows/changelog.yml
jjanowsk: That's interesting

.github/workflows/dod.yml
ThrasherLT: OK then, I left it there anyway, so should all be good now.

jjanowsk (approved):
+2

packgron (approved):

PR #513, opened 2024-04-24
Fix fuzzing jobs

jjanowsk:
Fuzzing jobs started randomly failing because of the dependencies not being locked.
1. Update nightly version of the compiler to the latest one to fix the compilation
2. Update cargo-fuzz to the latest version
3. Add Cargo.lock files to the fuzz targets in order to avoid this situation in the future
4. Add missing .gitignore to the telio-pq/fuzz


gytsto (approved):
+1, but I think you need to add changelog file tho

packgron:
> +1, but I think you need to add changelog file tho

Yeh, the ci check is still broken and being fixed :|

packgron (approved):
+1.0

PR #516, opened 2024-04-25
[LLT-5088] Update moose to v5.0.0

lcruz99:
### Problem
Moose v5.0.0 is now linking sqlite3 library dinamically, so we want to ship it together with the binaries and link where necessary.

### Solution
- Update moose tag to `v5.0.0-libtelioApp`.
- Change build scripts to fetch and ship the shared libsqlite3 library.
- Update `Cargo.lock`.
- Link libsqlite3 to the swift package.

- [x] README.md is updated


.gitlab-ci.yml
```diff
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v1.1.1
+    branch: LLT-5088_moose_5_0_0_upgrade
```
Jauler, line 18: Adding a reminder to replace this with the actual tag once the PR is finalized :)

ci/moose_utils.py
```diff
@@ -43,6 +43,8 @@ def _download_moose_file(opsys: str, arch: str, file_name: str):
         raise ValueError("LIBTELIO_ENV_SEC_NEXUS_URL not set")
 
     url = f"{nexus_url}/repository/ll-gitlab-release/{MOOSE_PROJECT_ID}/{LIBTELIO_ENV_MOOSE_RELEASE_TAG}/bin/common/{opsys}/{arch}/{file_name}"
+    if not url.startswith("https://"):
+        url = "https://" + url
```
Jauler, line 47: This may provide invalid URL if it is prefixed with `http://` or similar other protocol types. Not sure how relevant this is though :thinking:.

ci/build_libtelio.py
```diff
@@ -243,9 +294,14 @@ def main() -> None:
 
 def exec_build(args):
     if args.moose:
-        if args.os in ["windows", "android"]:
+        if args.os in ["linux", "windows", "android"]:
             sys.path.append(f"{PROJECT_ROOT}/ci")
-            moose_utils.fetch_moose_dependencies(args.os, MOOSE_MAP[args.arch])
+
+            if (args.os, args.arch) == ("linux", "armv7"):
+                moose_utils.fetch_moose_dependencies(args.os, MOOSE_MAP["armv7hf"])
```
Jauler, line 301: :thinking: maybe there were some discussions and I am not aware of its outcomes, but it is not safe to assume `armv7hf` architecture when `armv7` is requested. The reverse would be safe, but in this direction. This may produce segfaults on non-hardfload CPUs.

ci/moose_utils.py
lcruz99: This is not necessary, I needed to edit my local env vars, probably other users will also need to.

ci/build_libtelio.py
lcruz99: We call it `armv7` but underneath `armv7hf` is used.

lcruz99: Should I set this arch to the more explicit `armv7hf` on release versions? This might break things on other places

Jauler: IMHO it would be preferable to set the arch itself to `armv7hf` in order to avoid this silent "gotcha" where `armv7` in fact means `armv7hf`. 

But I wonder how many things will be broken by attempting to change it :thinking:. If it's major effort to update it, then probably we can live with `armv7`, otherwise, it would be cool to avoid this gotcha.

ci/moose_utils.py
Jauler: ok

ci/build_libtelio.py
lcruz99: Done.

Jauler (approved):
+1.0

ThrasherLT:
Not an expert on this part of the code, but lgtm +1.0

PR #517, opened 2024-04-26
Tcpdump traces collection from the vpn servers in nat-lab added

tomaszklak:
### Problem
When debugging issues related to VPN it's sometimes usefull to see the traffic from the point of view of the vpn server.

### Solution
Collect tcpdump traces from all vpn servers and store them in a similar way as the log files.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/mesh_api.py
+++ b/nat-lab/tests/mesh_api.py
@@ -8,6 +8,7 @@ from config import DERP_SERVERS, LIBTELIO_IPV6_WG_SUBNET, WG_SERVERS
 from ipaddress import ip_address
 from typing import Dict, Any, List, Tuple, Optional
 from utils.router import IPStack, IPProto, get_ip_address_type
+from utils.testing import test_name_safe_for_file_name
 
 if platform.machine() != "x86_64":
     import pure_wg as Key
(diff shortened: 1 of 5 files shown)


nat-lab/tests/mesh_api.py
```diff
@@ -402,3 +406,29 @@ def config_dynamic_nodes(
...
+        cmd = f"docker exec --privileged {server_name} rm {PCAP_FILE_PATH}"
+        os.system(cmd)
+
+        cmd = f"docker exec -d --privileged {server_name} tcpdump -i any -U -w {PCAP_FILE_PATH}"
+        os.system(cmd)
+
+
+def stop_tcpdump_on_vpns():
+    if os.environ.get("NATLAB_SAVE_LOGS") is None:
+        return
```
mathiaspeters, line 425: Nit: the can be made more consistent by either having both use an early return (like `stop_tcpdump_on_vpns`) or both do the logic inside the if-block (like `_start_tcpdum_on_vpn`)

mathiaspeters (commented):
+1

matislovas:
+1

matislovas (approved):

PR #518, opened 2024-04-29
LLT-5136 pq get keys/rekey mechanism refinements

matszczygiel:
### Problem
After the security review, it turned out that the current PQ protocol is flawed. The changes are required to the GET KEY and REKEY protocol requests and responses

### Solution
Adjust the protocol

- [x] README.md is updated


LukasPukenis (commented):
The fact this doesn't break any tests indicates we had them missing for the rekey :thinking:

matszczygiel:
@LukasPukenis I've added the rekey test

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +131,111 @@ async def test_pq_vpn_connection(
...
+        ip = await testing.wait_long(stun.get(client_conn, config.STUN_SERVER))
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        await _connect_vpn_pq(
+            client_conn,
+            client_alpha,
+        )
+
+        # Sleep for 6s to allow for the rekey to happen
+        await asyncio.sleep(6)
```
gytsto, line 238: isn't this racy condition, what is making sure that rekey happened in 6 seconds? maybe so event happens during that time?

matszczygiel: I'm setting the REKEY interval to 4s here https://github.com/NordSecurity/libtelio/pull/518/files/6ee7026e3ea09d2d3e1091dd6d71ee24931680ab#diff-d71adf294c60fb47c1fdaabf09dfb47c83a4566748a530105ca35e310f7f49fdR219

matszczygiel: There's no other way, the client is responsible for periodic rekey so whatever we feed the client it's fine as long as the VPN connection is not broken.
I believe that without rekeys at all the server has a hardcoded timeout on the connection but it is like 5 minutes or something

matszczygiel: Oh, do you mean something like inspecting the logs instead of sleeping? Good point actually

gytsto: Also, I think, you can easily ping PHOTO album even without vpn connection. So maybe do `stun.get` to check wether we still have VPN server IP as our public ip

matszczygiel: I've added the wait for log and it was not appearing. Good catch

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +130,110 @@ async def test_pq_vpn_connection(
...
+        ip = await testing.wait_long(stun.get(client_conn, config.STUN_SERVER))
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        await _connect_vpn_pq(
+            client_conn,
+            client_alpha,
+        )
+
+        await testing.wait_long(client_alpha.wait_for_log("Successful PQ REKEY"))
+
```
gytsto, line 237: can you do `stun.get` and check if we still on VPN ip after rekey

matszczygiel: Done

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +130,115 @@ async def test_pq_vpn_connection(
...
+        ip = await testing.wait_long(stun.get(client_conn, config.STUN_SERVER))
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        await _connect_vpn_pq(
+            client_conn,
+            client_alpha,
+        )
+
+        await testing.wait_long(client_alpha.wait_for_log("Successful PQ REKEY"))
```
LukasPukenis, line 236: I think the proper way to test it would be to make it actually fail after the server's timeout, so if the server should kill the session after like 5minutes, the testcase should wait for that amount of time at the very least to confirm it.

I see two testcases:
- Establish PQ and wait 5 minutes and check if we're still VPN connected( we shouldn't )
- Establish PQ and wait 5 minutes but at some time before we do a rekey and except the tunnel to be working after 5minutes

I think this is fine as that would show that functional correctness is achieved even though the testcase would be very long.

Maybe there's a way to configure the session length for the nordlynx?

LukasPukenis (commented):
+0.75

nat-lab/tests/test_pq.py
gytsto: if going for such long test, don't forget to mark it as long `pytesst.mark.long` ... Although it will be run only during releases, otherwise it would be painful to run every pipeline :smile:

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +130,110 @@ async def test_pq_vpn_connection(
...
+        #         adapter_type=AdapterType.BoringTun,
+        #         connection_tracker_config=generate_connection_tracker_config(
+        #             ConnectionTag.DOCKER_CONE_CLIENT_1,
+        #             stun_limits=ConnectionLimits(1, 1),
+        #             nlx_1_limits=ConnectionLimits(2, 2),
+        #         ),
+        #         is_meshnet=False,
+        #     ),
+        #     "10.0.254.1",
+        # ),
```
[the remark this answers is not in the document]
matszczygiel: Yeah, it doesn't. I created a separate PR since this is not as important as the one https://github.com/NordSecurity/libtelio/pull/547

nat-lab/tests/test_pq.py
matszczygiel: So I've confirmed that the timeout is 5 min. I don't think it's feasible to wait that long but maybe there's a way to inspect `nlx` output for the changed key?

nat-lab/tests/test_pq.py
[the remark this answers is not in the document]
matszczygiel: BTW I'll include it in this PR for convenience

nat-lab/tests/test_pq.py
matszczygiel: I did it by inspecting the preshared key with nlx

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +142,125 @@ async def test_pq_vpn_connection(
...
+
+    async with AsyncExitStack() as exit_stack:
+        env = await exit_stack.enter_async_context(
+            setup_environment(exit_stack, [alpha_setup_params])
+        )
+
+        client_conn, *_ = [conn.connection for conn in env.connections]
+        client_alpha, *_ = env.clients
+
+        ip = await testing.wait_long(stun.get(client_conn, config.STUN_SERVER))
```
gytsto, line 240: new changes appeared in main, could you remove `testing.wait_long` wrapper from `stun.get`, since `timeout` is now parameter of `stun.get`

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +142,125 @@ async def test_pq_vpn_connection(
...
+
+        preshared_before = inspect_preshared_key(nlx_conn)
+        await testing.wait_long(client_alpha.wait_for_log("Successful PQ REKEY"))
+
+        preshared_after = inspect_preshared_key(nlx_conn)
+        assert (
+            preshared_after != preshared_before
+        ), "Preshared key not changed on the nlx server"
+
+        ip = await testing.wait_long(stun.get(client_conn, config.STUN_SERVER))
```
gytsto, line 260: same here

nat-lab/tests/test_pq.py
```diff
@@ -130,3 +142,125 @@ async def test_pq_vpn_connection(
...
+        assert (
+            ip == config.NLX_SERVER["ipv4"]
+        ), f"wrong public IP when connected to VPN {ip}"
+
+        async with Ping(client_conn, config.PHOTO_ALBUM_IP).run() as ping:
+            await testing.wait_long(ping.wait_for_next_ping())
```
gytsto, line 266: same thing with `wait_for_next_ping`

nat-lab/tests/test_pq.py
matszczygiel: Done

nat-lab/tests/test_pq.py
matszczygiel: Done

nat-lab/tests/test_pq.py
matszczygiel: Done

nat-lab/bin/nordlynx
```diff
@@ -21,9 +21,26 @@ sleep 2
 ip addr s > /var/log/ipaddr.show
 nlx > /var/log/nlx.log
 /usr/bin/fakefm/fakefm -s testing123 2>&1 > /var/log/fakefm.log &
-/opt/nlx-radius/nlx-radius --log=stdout --verb=DEBUG 2>&1 > /var/log/nordlynx.log &
+/opt/nlx-radius/nlx-radius 2>&1 > /var/log/nordlynx.log &
```
gytsto, line 24: why debug log was removed?

gytsto: ah nvm, I see... :laughing:

matszczygiel: Those are no longer required with the new radius

gytsto (approved):
+1

tomaszklak (approved):

PR #523, opened 2024-05-06
LLT-5140: Replace duration with u64 in udl

mathiaspeters:
### Problem
One of the NordVPN apps cannot currently use the duration type

### Solution
For now, we will use u64 in the interface, and convert it in libtelio

- [x] README.md is updated


packgron (approved):
+2.0

crates/telio-model/src/features.rs
```diff
@@ -1,11 +1,10 @@
 //! Object descriptions of various
 //! telio configurable features via API
 
-use std::{collections::HashSet, fmt, time::Duration};
+use std::{collections::HashSet, fmt};
 
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use serde::{de::IntoDeserializer, Deserialize, Deserializer, Serialize};
-use serde_with::{serde_as, DurationSeconds};
```
jjanowsk, line 8: Please check if serde_with can be removed from the dependencies

jjanowsk (approved):
+1

crates/telio-model/src/features.rs
mathiaspeters: It still used in `telio-crypto`, but I removed it everywhere else

packgron (approved):

PR #527, opened 2024-05-06
Do not reenable management interface

Jauler:
### Problem
I have a suspicion that we still have some race conditions with DHCP on management interface for windows VMs, which causes non-deterministic test failures. I would like to observe how much and what tests are affected to be able to better decide on next actions, but for that, DHCP on management interface needs to not be reenabled. This PR makes that happen.

**Note:**

When management interface is disabled on windows VM:
```bash
vagrant ssh windows-1
```
will not work anymore, and one will have to use primary or secondary interface to access the VM, for example:
```bash
ssh vagrant@10.55.0.13
```
will be equivalent to the above command.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/network_switcher/network_switcher_windows.py
+++ b/nat-lab/tests/utils/network_switcher/network_switcher_windows.py
@@ -97,7 +97,12 @@ class NetworkSwitcherWindows(NetworkSwitcher):
         try:
             yield
         finally:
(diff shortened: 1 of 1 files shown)


gytsto (approved):
+2

mathiaspeters (approved):

gytsto (approved):

PR #530, opened 2024-05-08
Add pcap collection for docker based telio peer nodes

tomaszklak:
### Problem
Tcpdumps are collected only on vpn server docker containers. We would like to also collect them on the containers that run libtelio/tcli.

### Solution
Start tcpdump on docker containers where we have libtelio/tcli clients running.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -12,11 +12,11 @@ from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
 from dataclasses_json import DataClassJsonMixin, dataclass_json
 from enum import Enum
-from mesh_api import Meshmap, Node
+from mesh_api import Meshmap, Node, start_tcpdump, stop_tcpdump
 from telio_features import TelioFeatures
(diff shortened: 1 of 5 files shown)


nat-lab/tests/telio.py
```diff
@@ -522,6 +522,9 @@ def __init__(
     async def run(
         self, meshmap: Optional[Meshmap] = None, telio_v3: bool = False
     ) -> AsyncIterator["Client"]:
+        if isinstance(self._connection, DockerConnection):
```
gytsto, line 525: During `test_event_content_meshnet_node_upgrade_direct` where beta node is deleted and started again, only 2nd instance log will be saved and 1st log file will be overwritten, If I understand correctly.

Either create file iterations or maybe attach tcpdump to connection instead of telio instance? But not sure about that

tomaszklak: Thanks, that's a very good point and I totally missed that! I went with counter in the last part of the file name.

gytsto (approved):
+2

PR #531, opened 2024-05-08
Rebind sockets when network path changes on Apple

tomaszklak:
When NWPathMonitor call a callback with a new path, not only store it for future use but also notify all NativeProtector instances that the new path had been identified. This will now trigger rebind of protected sockets, if the new interface has changed.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -1,4 +1,5 @@
 use debug_panic::debug_panic;
+use once_cell::sync::Lazy;
 use parking_lot::{Mutex, RwLock};
 
 use std::{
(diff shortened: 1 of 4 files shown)


crates/telio-sockets/src/protector/apple.rs
```diff
@@ -466,6 +501,12 @@ pub fn setup_network_path_monitor() {
         };
 
         *INTERFACE_NAMES_IN_OS_PREFERENCE_ORDER.lock() = names.borrow().clone();
```
Jauler, line 503: Could we add a `telio_log_info!(...)` which would log what interfaces has been reported by the event and in what order. During in-branch testing, it would be nice to verify, that notifications are in-fact coming in a manner that we expect.

Jauler (approved):
+1.0, but lets add logging of what notifications has been issued by `NWPathMonitor`, and then lets request in-branch testing from apple team. Lets especially request scenarios where network migration occurs (e.g. WIFI->4G on iOS) and lets verify, that notifications that are coming from `NWPathMonitor` are indeed what we would expect :)

PR #532, opened 2024-05-09
Fix wg adapter gone test

gytsto:
### Problem
If tcli still exists, it is not possible to create or delete adapter with same GUID, since tcli is still holding a reference to it.

### Solution
Wait for tcli to be gone and then proceed with test

- [x] README.md is updated


packgron (approved):
+2.0

PR #535, opened 2024-05-10
Add no-link enhanced detection

dfetti:
Improve no-link detection by sending a ping when we have clues that the connection might be down or when the connection is silent for too long. Sending some data over the connection, other than keepalives, should be enough to see the state of the connection.


crates/telio-wg/src/link_detection/pinger.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    pub fn new(ping_channel: chan::Rx<IpAddr>) -> std::io::Result<Self> {
+        Ok(State {
+            ping_channel,
+            client_v4: Arc::new(Client::new(&Self::make_builder(ICMP::V4).build())?),
+            client_v6: Arc::new(Client::new(&Self::make_builder(ICMP::V6).build())?),
+        })
+    }
+
+    pub async fn perform(&self, target: IpAddr) {
```
LukasPukenis, line 55: nitpick: usually it's some of usual keywords: run, exec, execute

crates/telio-wg/src/link_detection/pinger.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        pinger.timeout(Self::PING_TIMEOUT);
+
+        match pinger.ping(PingSequence(0), &[0; 56]).await {
+            Ok(_) => telio_log_debug!("Ping to {} succeeded", target),
+            Err(e) => telio_log_debug!("Ping to {} failed: {}", target, e.to_string()),
+        }
+    }
+
+    fn make_builder(proto: ICMP) -> ConfigBuilder {
```
LukasPukenis, line 81: Can we reuse pingers from nurse or sessionkeeper maybe? The code is pretty straight forward so no harm in there but it would be nice to have less code to maintain

LukasPukenis (commented):
There are several questions I still have:

- Regarding this detection, it will be used together with increased direct keepalives in downgrade logic, however the downgrade logic is not in this MR, is it gonna be next task?

- Nat-Lab tests are missing to indicate any changes using the enhanced no-link detection. Using this implementation should yield some differences when having `PossibleDown` state already which might mean that we didn't have tests for that earlier as well.

stalowyjez (commented):
Lgtm, +1 - EDIT: sorry, wrong review, will get back to this one in a moment

stalowyjez (commented):
Soo, it sends some pings, but where does it actually use them to detect when the link is down? Or maybe the rest of the link detection mechanism will do the job?

crates/telio-wg/src/link_detection/pinger.rs
dfetti: I updated it to actually build it not to return the configbuilder. I did not reuse the other pingers, because the others are more specialized and I just needed a simple thing. Maybe the one from nurse can be moved into telio-utils and adapted to suite the needs of no-link detection too. I'll think more about it.

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -306,30 +363,36 @@ impl State {
...
                     self.variant = StateVariant::PossibleDown {
                         deadline: Instant::now()
                             .checked_add(Self::POSSIBLE_DOWN_DELAY)
                             .unwrap_or_else(Instant::now),
                     };
+                    Self::build_result(false, true, LinkState::Up)
+                } else if is_silent_connection {
+                    // The connection is silent. Might be down so let's ping.
+                    Self::build_result(false, true, LinkState::Up)
+                } else {
```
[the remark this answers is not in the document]
dfetti: PossibleDown is like a purgatory if you want to look at it like this. We never report PossibleDown to the integrator, The link state is still considered Up while we are in PossibleDown state.  And we try to check it/revive the connection in this state. We issue the ping when we transition from Up to PossibleDown and while in PossibleDown we just wait to see what's up with the connection. So in my opinion is ok as it is. Maybe we can increase the timeout for it to 5 seconds insted of 3 seconds to reduce more the posibility of oscilations.

crates/telio-wg/src/link_detection/pinger.rs
LukasPukenis: Extracting it and reusing would make sense, yes as this results now in a small cod eduplication, but to be perfectly honest, I am fine with it as it is

crates/telio-wg/src/link_detection/mod.rs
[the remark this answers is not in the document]
LukasPukenis: Thanks for explaining. It can be difficult to judge just looking at the code that not all of these states are gonna be emitted :)

crates/telio-wg/src/link_detection/pinger.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// Component used to check the link state when we think it's down.
+/// We can check it before we announce it as down.
+///
+/// Can be used with both IPv4 and IPv6 addresses.
+pub struct Pinger {
+    task: Task<State>,
+}
+
+impl Pinger {
```
matislovas, line 26: This file is a bit similar to `crates/telio-nurse/src/rtt/ping.rs`. Maybe it would be wise to export it to some helper object with API, which satisfies both use cases. But no hard opinion here though ....

matislovas:
Other than that +0.5

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -71,17 +87,37 @@ impl LinkDetection {
...
     rtt: Duration,
+    enhanced_detection: bool,
+    pinger: Option<pinger::Pinger>,
+    ping_channel: chan::Tx<IpAddr>,
     peers: HashMap<PublicKey, State>,
 }
 
 impl LinkDetectionEnabled {
-    fn new(rtt: Duration) -> Self {
+    fn new(rtt: Duration, enhanced_detection: bool) -> Self {
```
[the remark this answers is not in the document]
stalowyjez: @dfetti the response sounds reasonable, but it looks that it is to my different comment - what do you think about renaming this argument?

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -241,6 +294,28 @@ enum StateVariant {
...
+
+impl StateDecision {
+    fn should_notify(&self) -> bool {
+        matches!(self, StateDecision::Notify)
+    }
+
+    fn should_ping(&self) -> bool {
+        matches!(self, StateDecision::Ping)
+    }
+}
```
stalowyjez, line 317: Hmm, not sure whether it is worth it to wrap this matches when they are used exactly once?

.unreleased/LLT-5023
```diff
@@ -0,0 +1 @@
+Add no-link enhanced detection mod
```
LukasPukenis, line 1: Some time ago we talked about this feature being "Link detection" if you still agree to that, could you rename all `no-link` parts to `Link detection` so it would be less ambiguous? :)

dfetti: You're right, I forgot to update it.

crates/telio-wg/src/link_detection/pinger.rs
LukasPukenis: Doubling down on this comment - the code is very duplicate of another file and becasue it contains more stuff like some macos/ios/tvos conditionals I highly suggest reusing the existing pinger instead

LukasPukenis (commented):
Rethinking about the solution now with the lastest knowledge I think it would be best to provide more flexible configuration from the outside as a singular ping might not be enough to detect the link state.

Of course this is speaking about a small fraction of practical communications(probably) but I think it's better to have configurable amount of pings instead of a toggle. 

Configurable ping interval and cound would allow to flood the peer with pings and assuming he would respond with WG UDP packets, then we increase a chance of receiving at least one of them

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -111,7 +148,21 @@ impl LinkDetectionEnabled {
...
+                node_state,
+                self.rtt,
+                self.enhanced_detection,
+            );
+
+            if self.enhanced_detection && result.should_ping {
+                if let Err(e) = self.ping_channel.send(node_addresses).await {
+                    telio_log_warn!("Failed to trigger ping to {:?} {:?}", public_key, e);
+                }
+            }
```
LukasPukenis, line 163: Hmm.... What happens if pinging is in action but link detection noticed that the state is up because it received the packet? Does the state machine handle such a transition well? It's a bit hard to follow the code so I'm asking instead :)

src/libtelio.udl
```diff
@@ -556,10 +556,18 @@ dictionary FeatureFirewall {
...
+    /// Number of pings issued during the connection check
+    u32 no_of_pings;
+};
+
 /// Link detection mechanism
 dictionary FeatureLinkDetection {
     /// Configurable rtt in seconds
     u64 rtt_seconds;
+    /// Check the link state before reporting it as down
+    FeatureEnhancedDetection? enhanced_detection;
```
LukasPukenis, line 570: I think it might be worth just inlining the number of pings in here. That would reduce the code and should be a very simple modification, now we have a nested object one feature each and to be fair, nobody should use link detection without at least one ping(enhancement) :)

dfetti: Yes, but as you said we might add some interval config. 
And if I inline it as a number how would someone disable the feature? Making it 0?

LukasPukenis (approved):
+1.5

PR #537, opened 2024-05-10
Change the interval missed tick behaviour to Delay in RepeatedActions::add_action

tomaszklak:
### Problem
`RepeatedActions::add_action` used intervals with default missed tick behaviour. This caused many 'activations' after a long sleep, one for each missed tick. This is not desired - we want to ignore all the missed ticks.

### Solution
`add_action` is modified to use a missed tick behaviour that will skip all the missed ticks. Additionally, to make it a little bit less likely that this kind of a mistake will be once again made all the places where the intervals are created had been converted to use our internal wrapper that returns `Interval` with already correct missed tick behaviour.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5139
+++ b/.unreleased/LLT-5139
@@ -1,0 +1,1 @@
+Change the interval missed tick behaviour to Delay in RepeatedActions::add_action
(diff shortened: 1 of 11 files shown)


jjanowsk (approved):
+1 but please remove real time from the test

jjanowsk (approved):

stalowyjez (approved):

PR #543, opened 2024-05-14
LLT-4901 Fetch release binaries from nightly pipeline

matislovas:
### Problem
Final piece of the puzzle, to implement our build promotion system. When doing a release, we're rebuilding all binaries from scratch. 

### Solution
With these changes we will be fetching binaries (non-debug builds) and bindings from latest `nightly` pipeline.

- [x] README.md is updated


.github/workflows/build.yml
```diff
@@ -9,8 +9,10 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
           submodules: true
+      - name: install python3-requests
+        run: pip3 install requests
```
gytsto, line 13: I tought this was added to dockers itself, do we need it here?

.github/workflows/build.yml
```diff
@@ -46,6 +48,8 @@ jobs:
       - name: Setup MSVC
         uses: egor-tensin/vs-shell@9a932a62d05192eae18ca370155cf877eecc2202 # v2.1
         if: ${{ matrix.target_os == 'windows' }}
+      - name: install python3-requests
```
gytsto, line 51: I tought this was added to dockers itself, do we need it here?

.github/workflows/build.yml
```diff
@@ -89,6 +93,8 @@ jobs:
       - uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
         with:
           key: ${{ matrix.target_os }}-${{ matrix.arch }}
+      - name: install python3-requests
+        run: apt update && apt install -y python3-requests
```
gytsto, line 97: I tought this was added to dockers itself, do we need it here?

Also why `apt update` here, but not in other places?

.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.1.5
+          triggered-ref: LLT-4901_fetch_release_bins_from_nightly
```
gytsto, line 34: Don't forget to update :wink:

.gitlab-ci.yml
```diff
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v1.1.5
+    branch: LLT-4901_fetch_release_bins_from_nightly
```
gytsto, line 18: Don't forget to update :wink:

.github/workflows/build.yml
matislovas: This was added to `rust_build_utils` docker. These are the github's VMs, which is provisioned separately.

.github/workflows/build.yml
matislovas: Sames as https://github.com/NordSecurity/libtelio/pull/543#discussion_r1608039724

.github/workflows/build.yml
matislovas: Sames as https://github.com/NordSecurity/libtelio/pull/543#discussion_r1608039724

It is isolated environment, and it prohibits u to use `pip` for some reason ...

requirements.txt
```diff
@@ -1 +1,2 @@
 pipenv == 2023.11.15
+requests == 2.31.0
```
gytsto, line 2: we use Pipenv, so it should be added Pipfile and Pipfile.lock, instead of requirements.txt

matislovas: Fixed!

ci/fetch_binaries.py
```diff
@@ -0,0 +1,143 @@
...
+                if (
+                    target_os in job["name"] if target_os is not None else True
+                ) and target_arch in job["name"]:
+                    get_artifacts(path_to_save, job, unzip=True)
+                    return True
+
+    return False
+
+
+def fetch_build_artifacts(tag_prefix, target_os, target_arch, token, download_dir):
```
lcruz99, line 128: I think would be cleaner if we had some sort of config object where these parameters are stored and then pass it to `get_pipeline_build_artifacts(config)` where the data processing is done, globals would be avoided. Also wondering if we would be better to call `get_pipeline_build_artifacts()` directly from `build_libtelio.py`, why whould we need to have both `fetch_build_artifacts()` and `get_pipeline_build_artifacts()`, I guess this is more of a design question than a change request..

ci/build_libtelio.py
```diff
@@ -317,7 +329,49 @@ def main() -> None:
...
+            args.try_fetch_from_pipeline,
+            args.os,
+            args.arch,
+            token,
+            PROJECT_ROOT,
+        ):
+            if args.moose and args.os in ["linux", "windows", "android"]:
+                moose_utils.fetch_moose_dependencies(args.os, MOOSE_MAP[args.arch])
+
+            return 0
```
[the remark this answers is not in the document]
matislovas: My intention was to _try_ to fetch and if it fails (artifacts deleted, no nightly tag to date) than proceed with the ordinary build.

We need them because of the `finalize_win.py` script artifacts are not included, so need to re-download those leftovers.

ci/fetch_binaries.py
matislovas: This naming is a leftover, but yeah, I agree that it is confusing (the _api_ of that file should only include `fetch_build_artifacts()`).  I'll change it to an object based fetch.

ci/build_libtelio.py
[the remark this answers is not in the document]
lcruz99: So maybe we can warn the user and even ask if he wants to proceed with the build?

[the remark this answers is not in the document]
matislovas: It will be stuck on CI than, if we'll wait for something from `stdin`. For the error handling, I can wrap this into `try{}except{}` maybe ...

ci/fetch_binaries.py
matislovas: Fixed

ci/build_libtelio.py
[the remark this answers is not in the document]
matislovas: Fixed

ci/build_libtelio.py
```diff
@@ -317,7 +329,71 @@ def main() -> None:
...
+
 def exec_build(args):
+    # Try fetching from pipeline, if it fails, continue with the build
+    if args.try_fetch_from_pipeline:
+        if try_download_artifacts(
+            args.try_fetch_from_pipeline, PROJECT_ROOT, args.arch, args.os, args.moose
+        ):
+            return
+
+        print("Proceeding with build...")
```
lcruz99, line 395: I think this is too passive and will be lost in between all the logs.

```suggestion
        while True:
        response = input("Artifacts download failed, do you want to proceed with the build? (y/n): ").strip().lower()
        if response == 'y':
            print("Proceeding with build...")
            break
        elif response == 'n':
            return;
        else:
            print("Invalid input. Please enter 'y' or 'n'.")
```

ci/build_libtelio.py
```diff
@@ -317,7 +329,71 @@ def main() -> None:
...
+    if args.try_fetch_from_pipeline:
+        if try_download_artifacts(
+            args.try_fetch_from_pipeline,
+            PROJECT_ROOT,
+            target_arch=None,
+            target_os="uniffi",
+        ):
+            return
+
+    print("Proceeding with generating uniffi ...")
```
lcruz99, line 376: Same as https://github.com/NordSecurity/libtelio/pull/543/files#r1622533016.

ci/fetch_artifacts.py
```diff
@@ -0,0 +1,145 @@
...
+    def download(self) -> bool:
+        _, tag_msg = self._get_latest_tag(self.tag_prefix)
+
+        if tag_msg:
+            tag_json = json.loads(tag_msg)
+            return self._get_pipeline_build_artifacts(
+                tag_json["pipeline_id"],
+                self.download_dir,
+                self.target_arch,
+                self.target_os,
```
lcruz99, line 30: Redundant arguments, these are already attributes of the class so can be removed.

ci/build_libtelio.py
matislovas: Same as https://github.com/NordSecurity/libtelio/pull/543#discussion_r1616764373

ci/build_libtelio.py
matislovas: Same as https://github.com/NordSecurity/libtelio/pull/543#discussion_r1616764373

ci/fetch_artifacts.py
lcruz99, line 30: There are redundant since the object already have these attributes.

```suggestion
                self.download_dir,
                self.target_arch,
                self.target_os,
```

ci/build_libtelio.py
lcruz99: oh sry didn't see it.

ci/build_libtelio.py
lcruz99: oh sry didn't see it.

ci/fetch_artifacts.py
lcruz99: other than that +2

.github/workflows/gitlab.yml
matislovas: Fixed

.gitlab-ci.yml
matislovas: Fixed

tomaszklak (dismissed):

.github/workflows/build.yml
Jauler, line 13: Could you pin the hash and version?

.github/workflows/build.yml
```diff
@@ -46,6 +48,8 @@ jobs:
       - name: Setup MSVC
         uses: egor-tensin/vs-shell@9a932a62d05192eae18ca370155cf877eecc2202 # v2.1
         if: ${{ matrix.target_os == 'windows' }}
+      - name: install python3-requests
+        run: pip3 install requests
```
Jauler, line 52: Could you pin the hash and version?

.github/workflows/tests.yml
```diff
@@ -68,6 +68,12 @@ jobs:
       - name: Setup MSVC
         uses: egor-tensin/vs-shell@9a932a62d05192eae18ca370155cf877eecc2202 # v2.1
         if: ${{ matrix.target_os == 'windows' }}
+      - name: install python3-requests
+        run: pip3 install requests
```
Jauler, line 72: Could you pin the hash and version?

ci/build_libtelio.py
```diff
@@ -317,7 +329,71 @@ def main() -> None:
         assert False, f"command '{args.command}' not supported"
 
 
+def try_download_artifacts(
+    tag_prefix, path_to_save, target_arch, target_os, moose=False
+):
+    def get_token():
+        if "LLH_GROUP_TOKEN_FLAKY_TESTS" in os.environ:
```
Jauler, line 336: Maybe it would be better either to create dedicated token environment variable for fetching artifacts, or rename this one? 

It my be somewhat surprising, if `FLAKY_TESTS` token is changes/renewed/updated etc. and something happens to artifact downloading.

ci/build_libtelio.py
```diff
@@ -317,7 +329,71 @@ def main() -> None:
...
+        LIBTELIO_ENV_UNIFFI_GENERATORS_TAG,
+        ["python", "cs", "go", "swift", "kotlin"],
+        "src/libtelio.udl",
+        dockerized=args.dockerized,
+    )
+
+
 def exec_build(args):
+    # Try fetching from pipeline, if it fails, continue with the build
+    if args.try_fetch_from_pipeline:
```
Jauler, line 389: The release build must not continue if nightly tag fails to download (e.g. does not exist or similar).

matislovas: Fixed

ci/build_libtelio.py
matislovas: Fixed

.github/workflows/tests.yml
matislovas: Fixed, but without hashes. `--hash` is only available through `requirements.txt` and if I put it for one package, it is required for all deps and sub-deps, so only proceeding with version pining.

.github/workflows/build.yml
matislovas: Same as https://github.com/NordSecurity/libtelio/pull/543#discussion_r1624164533

.github/workflows/build.yml
matislovas: Same as https://github.com/NordSecurity/libtelio/pull/543#discussion_r1624164533

.github/workflows/build.yml
matislovas: Fixed

tomaszklak (approved):

Jauler (approved):
+1.0

PR #546, opened 2024-05-16
LLT-5186 improve natlab timeouts

gytsto:
### Problem
After decreasing resources in natlab VMs, some of the test processes became slower.

### Solution
- increase timeouts on natlab events and tests itself
- move asyncio timeout wrappers to under commonly used utils and add timeout parameter with default values (stun, ping, stding_ready)
- remove unnecessary timeout wrappers
- remove unnecessary asyncio function usages
- fix race condition in `test_derp_client_message_forward`
- update libtelio-build ref, which contains provision script fix

- [x] README.md is updated


Jauler (approved):
TBH I doubt that some of the test failures are caused by timeouts being sets to too low value, but, on the other hand this value is arbitrary number anyways.

But I really like this "reversion" of waiting for futures, as it makes tests muuuuch easier to read. Thus: +1.0.

gytsto:
> TBH I doubt that test failures are caused by timeouts being sets to too low value, but, on the other hand this value is arbitrary number anyways.
> 
> But I really like this "reversion" of waiting for futures, e.g. instead of doing that in the test code, it is now inside of the util code. This it makes tests code quite a bit cleaner. Thus: +1.0

Thanks for the +1, but alot of the test were flaky because of timeouts for sure, since the `vagrant-under-docker` improvements and resource reducement

nat-lab/tests/test_fire_connecting_event.py
```diff
@@ -3,15 +3,14 @@
...
-from utils import testing
 from utils.connection_tracker import ConnectionLimits
 from utils.connection_util import generate_connection_tracker_config, ConnectionTag
 from utils.ping import Ping
 
 
 @pytest.mark.asyncio
 @pytest.mark.long
-@pytest.mark.timeout(180 + 60)
+@pytest.mark.timeout(240)
```
Hasan6979, line 13: Can we parametrize all the different timeout values ? 
Changing the values later on would require only a couple of changes.

Hasan6979 (commented):
Other than 1 comment lgtm  +1.0

nat-lab/tests/test_fire_connecting_event.py
gytsto: Done!

nat-lab/tests/test_direct_connection.py
```diff
@@ -499,7 +490,7 @@ async def test_direct_short_connection_loss(
                     await asyncio.wait_for(task, 1)
 
         async with Ping(alpha_connection, beta.ip_addresses[0]).run() as ping:
-            await testing.wait_lengthy(ping.wait_for_next_ping())
+            await ping.wait_for_next_ping(60)
```
Hasan6979, line 493: These as well 🙈

nat-lab/tests/telio.py
```diff
@@ -450,7 +450,7 @@ async def wait_for_state_peer(
     ) -> None:
         await asyncio.wait_for(
             self._runtime.notify_peer_state(public_key, state, paths, is_exit, is_vpn),
-            timeout if timeout else 60 if PathType.Direct in paths else 30,
+            timeout if timeout else 90 if PathType.Direct in paths else 45,
```
Hasan6979, line 453: And these

Jauler (approved):
+1.0

nat-lab/tests/test_direct_connection.py
gytsto: You see, these can happen randomly everywhere, can't be putting all of these to CONSTANTS :smile:

nat-lab/tests/telio.py
gytsto: This I might put to constants tho

gytsto: done

Hasan6979 (approved):

PR #548, opened 2024-05-17
Fixes for apple interface selection and socket rebinding main

tomaszklak:
Merge of fixes from the 4.2 branch #531 and #524.

- [x] README.md is updated

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -1,4 +1,5 @@
 use debug_panic::debug_panic;
+use once_cell::sync::Lazy;
 use parking_lot::{Mutex, RwLock};
 
 use std::{
@@ -22,7 +23,11 @@ use system_configuration::{
     dynamic_store::{self, SCDynamicStore},
     sys::schema_definitions,
 };
-use tokio::{self, sync::Notify, task::JoinHandle};
+use tokio::{
+    self,
+    sync::{broadcast::Sender, Notify},
+    task::JoinHandle,
+};
 
 use crate::native::{interface_index_from_name, NativeSocket};
 use crate::Protector;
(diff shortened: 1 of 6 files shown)


crates/telio-sockets/src/protector/apple.rs
```diff
@@ -291,20 +328,44 @@ impl Sockets {
...
-            sockets.rebind_all(true);
+    let nw_path_monitor_monitor_handle = tokio::spawn({
+        let sockets = sockets.clone();
+        let mut notify = PATH_CHANGE_BROADCAST.subscribe();
+        async move {
+            loop {
+                match notify.recv().await {
+                    Ok(()) => {
+                        let mut sockets = sockets.lock();
+                        sockets.rebind_all(false);
```
LukasPukenis, line 343: What is the reason of `false` flag in comparison to `true` in the code below?

LukasPukenis (commented):
Judging from the looks of it looks like it should monitor network state changes and rebind sockets(though reason is unknown and I couldn't find it on the ticket description).

As an additional question - is it possible to exercise this fix in NatLab as we have MacOS runners? 

Code-wise it looks good :)

crates/telio-sockets/src/protector/apple.rs
tomaszklak: If the interface didn't change, there is no reason to rebind.

gytsto (approved):
+1, but natlab test would be really nice

crates/telio-sockets/src/protector/apple.rs
LukasPukenis: should we then even call the `rebind_all()`? I guess the function does more stuff than just rebinding then?

LukasPukenis:
I was under impression that we can run Mac tests via Vagrant? If so it would be really nice to explore if we can "natlabify" this thing :)

Despite the tests +1 but I really would like to push for it as you can see :D

crates/telio-sockets/src/protector/apple.rs
tomaszklak: > should we then even call the `rebind_all()`? I guess the function does more stuff than just rebinding then?

This will call `rebind_all` when some change was reported by the `NWPathMonitor`. In the rebind_all we 'compute' the actually interface that we would use for the current list of interfaces reported (in OS preference order) by `NWPathMonitor` and if the computed interface is different than the previous one, then we do the actual rebinding.

tomaszklak:
> I was under impression that we can run Mac tests via Vagrant? If so it would be really nice to explore if we can "natlabify" this thing :)
> 
> Despite the tests +1 but I really would like to push for it as you can see :D

We can, unless we use macbooks as work laptops ;)

I've created LLT-5196 to track writting nat-lab reproduction for this. If I get mac VMs running an macOS I will work on this myself (or if it's possible to run them on the proxmox).

PR #557, opened 2024-05-20
Use the latest version of nagger with interval lint

tomaszklak:
This also fixes few remaining places to use the telio-utils wrapper.


Jauler (approved):
+1.5

PR #564, opened 2024-05-21
increase vpn/exit conn event timeout

gytsto:
### Problem
some vpn connections are still flaky

### Solution
workaround by increasing timeout

- [x] README.md is updated


tomaszklak (commented):
+1

While I'm totally fine with this change I'm wondering if this is a good approach. If we are bumping timeout until we find stable values for them, maybe it would be better to completely remove such individual timeouts (and let individual operations/steps run as long as they need) and rely only on the overall test timeout?

gytsto:
> +1
> 
> While I'm totally fine with this change I'm wondering if this is a good approach. If we are bumping timeout until we find stable values for them, maybe it would be better to completely remove such individual timeouts (and let individual operations/steps run as long as they need) and rely only on the overall test timeout?

While I'd like to do that, since it would make life abit easier for sure, this might prolong natlab job. Instead of raising exception after individual timeout on event, we would have to wait for full test duration timeout. Which is compared to `vpn_event_timeout` is like 3x longer. And with 1 failure, it looks not much, but if 10 tests fails, it would add up.

Ofc, if we had resources to hog pipelines, I'd be 100% for the idea of using singular test timeout.

gytsto:
> > While I'd like to do that, since it would make life abit easier for sure, this might prolong natlab job. Instead of raising exception after individual timeout on event, we would have to wait for full test duration timeout. Which is compared to `vpn_event_timeout` is like 3x longer. And with 1 failure, it looks not much, but if 10 tests fails, it would add up.
> > Ofc, if we had resources to hog pipelines, I'd be 100% for the idea of using singular test timeout.
> 
> But we expect all tests to pass every time - so this is not going to prolong the passing tests. It can only prolong the failures, right?

Yes, that's true

gytsto:
@tomaszklak I think we can always test it, and see how it goes. I'll create issue for that.

PR #565, opened 2024-05-21
remove wait_for usage where timeout is not expected

gytsto:
### Problem
Waiting for individual event timeouts, where we are not expecting it to timeout, might cause flakiness, because it is kind of guessing game.

### Solution
Remove timeouts from places where it is not expected to timeout. Leave everything for general `pytest.timeout`.

### Note
Additional info in discussion https://github.com/NordSecurity/libtelio/pull/564

- [x] README.md is updated


tomaszklak (commented):
+1 hope it works and is stable - I really like the simplifications :)

nat-lab/tests/telio_test.py
```diff
@@ -100,39 +94,36 @@ async def test_set_peer_event(self) -> None:
...
+                runtime.notify_derp_state("1.1.1.1", [State.Disconnected]), 0.1
             )
 
         # it should fail (wrong IP)
         with pytest.raises(asyncio.TimeoutError):
-            await testing.wait_short(
-                runtime.notify_derp_state("1.1.1.2", [State.Connected])
+            await asyncio.wait_for(
+                runtime.notify_derp_state("1.1.1.2", [State.Connected]), 0.1
             )
```
stalowyjez, line 127: I think that `wait_short` is ok, is there any particular reason to replace it with `wait_for`?

stalowyjez (commented):
I'm a bit afraid of the scenario in which something will take like 3x longer, which would be a clear regression, but the test may allow it now, because it will still fit in the general test timeout

nat-lab/tests/telio_test.py
gytsto: I just wanted to use that testing.py wrapper

gytsto:
> I'm a bit afraid of the scenario in which something will take like 3x longer, which would be a clear regression, but the test may allow it now, because it will still fit in the general test timeout

As mentioned in the discussion from pr in notes, we will just have to experiment. If it wont fit us, we'll revert this change

stalowyjez:
>As mentioned in the discussion from pr in notes, we will just have to experiment. If it wont fit us, we'll revert this change

Yeah, I mean, I expect that when we relax the tests a bit the CI will be greener, so it probably will be nice for us, but I'm afraid that after a year someone from apps may tell us "hey, this handshake now takes 3x longer", which could be potentially catched by such per-item timeouts? IDK, on the other hand, simplification is always nice.

gytsto:
> >As mentioned in the discussion from pr in notes, we will just have to experiment. If it wont fit us, we'll revert this change
> 
> Yeah, I mean, I expect that when we relax the tests a bit the CI will be greener, so it probably will be nice for us, but I'm afraid that after a year someone from apps may tell us "hey, this handshake now takes 3x longer", which could be potentially catched by such per-item timeouts? IDK, on the other hand, simplification is always nice.

What you are saying would make perfect sense if natlab wouldn't be so flaky time wise. It doesn't always give exact performance tho. If we were to have perfect environment, we could return per-item timeouts, were we set default vexpected values and forget about them. :smile:

stalowyjez:
Ok, let's try it, +1

stalowyjez (approved):

PR #569, opened 2024-05-23
LLT-5198 Skip maven job on nightly

matislovas:
Bumping `libtelio-build` to v1.1.10, that skips maven job on nightly pipelines.


packgron (approved):
+2.0

PR #577, opened 2024-05-28
Verify that all non management interfaces in nat-lab have no 'Router' property

tomaszklak:
### Problem
No test was added to #548

### Solution
Turns out that we already have very recently set all non management interfaces to be static and unknowingly removed the 'Router' property. To keep it that way, I've added a check to make sure that only the vagrant management interface can have the 'Router' property.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5196
+++ b/.unreleased/LLT-5196
@@ -1,0 +1,1 @@
+Verify that all non management interfaces in nat-lab have no 'Router' property
--- a/nat-lab/bin/mac/list_interfaces_with_router_property.py
+++ b/nat-lab/bin/mac/list_interfaces_with_router_property.py
@@ -1,0 +1,46 @@
+# mypy: ignore-errors
+# autoflake: skip_file
+# pylint: skip-file
+
+from sys import exit
+from SystemConfiguration import SCDynamicStoreCreate, SCDynamicStoreCopyValue
+
+
+def main():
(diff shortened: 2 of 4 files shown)


.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.1.10
+          triggered-ref: LLT-5196-natlab-test-for-carplay-bug
```
tomaszklak, line 34: Reminder to update this before merge ❗

.gitlab-ci.yml
```diff
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v1.1.10
+    branch: LLT-5196-natlab-test-for-carplay-bug
```
tomaszklak, line 18: Reminder to update this before merge ❗

dfetti (commented):
+1.0

packgron (approved):
+1.0

packgron (approved):

PR #578, opened 2024-05-28
Attempt to address flakyness in VPN server configuration

Jauler:
### Problem
In some test runs it can be clearly seen that VPN servers are not configured correctly at startup
2024-05-28_15-31-33_grim

Quickly looking over setup code, we can clearly see a few issues immediately. One is that command execution is detached, and executed in background creating a race between VPN server setup and test startup. Additionally ephemeral port range is used to assign to VPN server, meaning that it is not guaranteed that the port will be free, This changes port to be outside of ephemeral ports range as well as makes the initiation of the server sync.


nat-lab/bin/nordlynx
```diff
@@ -9,7 +9,7 @@ ip addr add 100.64.0.1/10 dev nordlynx0  # meshnet
 ip link set mtu 1420 dev nordlynx0
 ip link set dev nordlynx0 up
 
-nlx set nordlynx0 listen-port 51820
```
LukasPukenis, line 12: I wanted to complain that this is standard WG port so it's more in-line with Wireguard even though it can be freely chosen, but actually I stand corrected - This port is only used in WG examples.

nat-lab/tests/mesh_api.py
```diff
@@ -357,24 +358,41 @@ def generate_peer_config(node: Node, allowed_ips: str) -> str:
...
+                    full_cmd = f"docker exec --privileged {server_config['container']} bash -c '{cmd}'"
+                    ret = os.system(full_cmd)
+                    print(
+                        datetime.now(),
+                        "Executing",
+                        full_cmd,
+                        "on",
+                        server_config["container"],
+                        "with result",
+                        ret,
```
LukasPukenis, line 374: Minor: might it be easy to use DockerProcess for that maybe? It already has this print inside of it.

LukasPukenis (commented):
+1.5 the removal of `-d` is a nice catch

PR #579, opened 2024-05-28
LLT-5220: Use UniFFI interface for TelioError

mathiaspeters:
### Problem
We want to be able to evolve our error definition without having to bump the major version of the lib

### Solution
Use UniFFI interface instead of enum for TelioError, and make necessary changes.

This change also allows us to use anyhow as the inner error of `TelioError::UnknownError`, which was previously blocked by UniFFI not being able to handle anyhow

- [x] README.md is updated

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -3,7 +3,7 @@ use telio_model::event::Event;
 use telio_utils::map_enum;
 use tracing::Level;
 
-use std::panic::RefUnwindSafe;
+use std::{panic::RefUnwindSafe, sync::Arc};
 
 use crate::device::{AdapterType, Error as DevError};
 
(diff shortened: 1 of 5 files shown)


LukasPukenis (commented):
+1 lgtm

src/ffi/types.rs
```diff
@@ -38,6 +38,27 @@ pub enum TelioError {
...
+            TelioError::LockError => "LockError".to_owned(),
+            TelioError::InvalidString => "InvalidString".to_owned(),
+            TelioError::AlreadyStarted => "AlreadyStarted".to_owned(),
+            TelioError::NotStarted => "NotStarted".to_owned(),
+        }
+    }
+
+    pub fn message(&self) -> String {
+        match self {
+            TelioError::UnknownError(inner) => inner.to_string(),
```
tomaszklak, line 56: nit: Is it possible that the `inner.to_string()` will result in the same string as the `self.kind()`? Maybe it would be safer to add some sort of prefix ("Unknown error: ") to distinguish it?

tomaszklak (approved):
+1 (with one small optional comment)

mathiaspeters:
I will reopen this PR when it is no longer blocked

PR #581, opened 2024-05-29
LLT-5103: Add feature flag for multicast

mathiaspeters:
### Problem
We need a feature flag do disable multicast both during development and potentially on the fly later

### Solution
Add such a feature flag. Multicast is disabled by default

- [x] README.md is updated


tomaszklak:
I don't see any tests changes or production code that would use this. Is there no implementation for this yet?

crates/telio-model/src/features.rs
```diff
@@ -535,6 +535,9 @@ pub struct Features {
     /// PMTU discovery configuration, enabled by default
     #[serde(default = "FeaturePmtuDiscovery::serde_default")]
     pub pmtu_discovery: Option<FeaturePmtuDiscovery>,
+    /// Multicast support
+    #[serde(default)]
+    pub multicast: bool,
```
tomaszklak, line 540: Since this is part of the boundary with external world, I think there should be 4 types of tests:
- what happens when the field is missing
- what happens when the field is set to true
- what happens when the field is set to false (afaics, this is already present)
- what happens when the field is set to incorrect value

mathiaspeters:
No there is not yet an implementation. We are adding the feature flag first so that the implementation can use it directly when the feature starts getting implemented, rather than implementing first and gating the feature afterwards

crates/telio-model/src/features.rs
stalowyjez: >what happens when the field is set to incorrect value
That would be nice, but do we have any tests of this kind?

mathiaspeters:
> Hmm, so it doesn't do anything for now - wouldn't that be better to make it one commit in somehow more holistic PR, so we can see how this flag is used? 

Multiple people are going to be working on different parts of the multicast implementation at the same time, all relying on the feature flag, so in this case it would be better to merge it first

> Also maybe it's worth considering to add this to telio_features.py and maybe to telio_features_test.py?

This is a good idea, and it's now added. It's a bit simpler than the unit tests added to parsing inside libtelio, but I think that's fine since it will be serialized to json and converted in libtelio anyways

crates/telio-model/src/features.rs
mathiaspeters: When I added it I was mostly looking at what tests were there already for the features struct and didn't think more of it, but you are right, we should be better about testing this. I added more tests to this section, and hopefully we can add similar tests for all the other fields in the `Features` struct in the near future

PR #582, opened 2024-05-29
[LLT-4979] Add test for set_secret_key

stalowyjez:
### Problem
Before rebase on Boringtun v0.6.0 calling `set_secret_key` while using Boringtun adapter caused panic. We need to make sure that setting private key will work in the future.

### Solution
Add a test for `set_secret_key` Telio method

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_vpn.py
+++ b/nat-lab/tests/test_vpn.py
@@ -4,7 +4,7 @@ import copy
 import pytest
 from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_environment, setup_connections
-from telio import AdapterType, Client
+from telio import AdapterType, Client, generate_public_key, generate_secret_key
 from telio_features import TelioFeatures, Firewall
 from typing import Optional
 from utils import testing, stun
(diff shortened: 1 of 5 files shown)


nat-lab/tests/test_vpn.py
```diff
@@ -564,3 +564,140 @@ async def on_stdout(stdout: str) -> None:
...
+                if cfg.get_key() == key:
+                    return cfg
+        return None
+
+    async with AsyncExitStack() as exit_stack:
+        alpha_setup_params = copy.deepcopy(alpha_setup_params)
+
+        cfg = find_cfg_by_key(alpha_setup_params, "vpn_1")
+        assert cfg is not None
+        cfg.limits = ConnectionLimits(1, 1)
```
packgron, line 660: what is the point of setting limits again, if they are already set in params?

nat-lab/tests/test_vpn.py
```diff
@@ -564,3 +564,140 @@ async def on_stdout(stdout: str) -> None:
...
+        params: SetupParameters, key: str
+    ) -> Optional[ConnectionTrackerConfig]:
+        if params.connection_tracker_config is not None:
+            for cfg in params.connection_tracker_config:
+                if cfg.get_key() == key:
+                    return cfg
+        return None
+
+    async with AsyncExitStack() as exit_stack:
+        alpha_setup_params = copy.deepcopy(alpha_setup_params)
```
packgron, line 656: Why do you need this deep copy?

nat-lab/tests/test_vpn.py
```diff
@@ -564,3 +564,140 @@ async def on_stdout(stdout: str) -> None:
...
+                    stun_limits=ConnectionLimits(1, 1),
+                ),
+                is_meshnet=False,
+            ),
+            "10.0.254.7",
+            marks=pytest.mark.mac,
+        ),
+    ],
+)
+async def test_vpn_connection_private_key_change(
```
packgron, line 642: We are now only focusing on vpn case?
Do we have assert to dissallow private key change while meshnet is on?

nat-lab/tests/test_vpn.py
```diff
@@ -564,3 +564,140 @@ async def on_stdout(stdout: str) -> None:
...
+            client_alpha,
+            alpha.ip_addresses[0],
+            config.WG_SERVER,
+        )
+
+        new_secret_key = generate_secret_key()
+        new_public_key = generate_public_key(new_secret_key)
+
+        # change public key on server
+        await vpn_connection.connection.create_process(
```
matislovas, line 689: Does this fails the test on the `exit 1` case from shell script?

nat-lab/tests/test_vpn.py
stalowyjez: This is disallowed in the code: https://github.com/NordSecurity/libtelio/blob/a3a4145c6b4329ff266f1b59472bc0eb87d6ac04/src/device/mod.rs#L1469

nat-lab/tests/test_vpn.py
stalowyjez: This are different connection_limits, movet them to params as well

nat-lab/tests/test_vpn.py
stalowyjez: When meshnet is on, this is forbidden

nat-lab/tests/test_vpn.py
stalowyjez: Yeah, it does

packgron (approved):
+1.0

matislovas (approved):
+1

PR #583, opened 2024-05-29
LLT-5223  ignore connection closed error on pyro daemon shutdown

Jauler:
### Problem
There seems to be a race when shutting down libtelio RPC services started for testing purposes. The race nature is that shutdown is requested via RPC command, which races between connection being closed and response being sent to the client. This PR allows for any order of events to be accepted.

- [x] README.md is updated


nat-lab/tests/telio.py
```diff
@@ -657,7 +657,7 @@ async def on_stderr(stdout: str) -> None:
                     if self._process.is_executing():
                         await self.stop_device()
                         self._quit = True
-                    self.get_proxy().shutdown()
+                    self.get_proxy().shutdown(self._connection.target_name())
```
mathiaspeters, line 660: It would have been enough to pass `self._node.name` but this works too :smile:

Jauler: oh, I missed it :D

Jauler: oh wait, but is'nt `self._node.name` just normally `alpha` or `beta` or similar? In that case, maybe it is a bit better to have specific container name?

mathiaspeters: I think the most important thing is to be able to see if the shutdown is happening multiple times on the same client, which the node name would tell us, but your method would make it easier to perhaps see patterns over time so let's go with it

ThrasherLT:
lgtm, +1.0

mathiaspeters (approved):
Looks good! +2

PR #584, opened 2024-05-29
Fix natlab after keys became lists instead of bytes

mathiaspeters:
### Problem
A recent commit changed the underlying type of keys in uniffi bindings from bytes to list[int] and a separate commit started working with the keys in a way that works for bytes but not list[int] making the changes incompatible, but there were no actual code overlap so the PRs could both be merged, causing natlab to break in main

### Solution
Convert the list to bytes

- [x] README.md is updated


Jauler (approved):
+2.0

PR #585, opened 2024-05-30
Remove wasm-bindgen as dependency

Jauler:
### Problem
Remove wasm-bindgen as dependency as it is not used anymore

- [x] README.md is updated


mathiaspeters (commented):
+1

Cargo.lock
```diff
@@ -4342,7 +4342,6 @@ dependencies = [
  "tracing",
  "uniffi",
  "uuid",
- "wasm-bindgen",
```
packgron, line 4345: Looks incorrect, should be etire dep

Jauler: Seems like `reqwest` has a dependency to `wasm-bindgen` which is not present in our selection of feature set for request. But it is still locked in `Cargo.lock` file.

packgron (approved):
+1.0

mathiaspeters (approved):

packgron (approved):

PR #592, opened 2024-06-04
[LLT-4928] Mask IP addresses in logs

stalowyjez:
Adds a filter to telio_log_* macros, which masks the IP addresses passed to it.

### Problem
We print unmasked IP addresses in logs

### Solution
This is not a solution to the whole problem - we still need to remember to use `Hidden` with IPs which are stored in structures.

This should work for the IPs and Endpoints which are directly passed to logs. It might be good to reduce a bit a possibility of the human error - it introduces some additional complexity to log macros, so I'm not sure if it's worth it in the end.

- [ ] README.md is updated
- [ ] Functionality is covered by unit or integration tests

### DIFF
--- a/crates/telio-utils/src/utils.rs
+++ b/crates/telio-utils/src/utils.rs
@@ -1,39 +1,61 @@
 /// Utilities
 ///
 
+#[macro_export]
+macro_rules! telio_log_generic {
+    ( $log_func: ident, $msg: expr) => {tracing::$log_func!($msg)};
+    ( $log_func: ident, $format: expr, $($arg:expr),+) => {{
+        use std::net::{IpAddr, SocketAddr};
+
+        #[allow(dead_code)]
+        trait HideSensitiveInfo {
+            fn hide_me_maybe(&self) -> &'static str {
+                "******"
+            }
+        }
+        impl HideSensitiveInfo for IpAddr {}
+        impl HideSensitiveInfo for SocketAddr {}
+
+        #[allow(dead_code)]
+        trait NotHideAtAll {
+            fn hide_me_maybe(&self) -> &Self {
+                self
+            }
+        }
+        impl<T> NotHideAtAll for &T {}
+
+        tracing::$log_func!($format,  $((&$arg).hide_me_maybe()),+)
+    }};
+}
+
 /// wrapping tracing::trace and adding more information
 #[macro_export]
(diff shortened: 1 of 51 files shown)


mathiaspeters:
Personally I'm not a fan of having that extra import everywhere for something that's only indirectly used. Would it be possible to fully qualify `telio_log_generic` inside of the other logging macros and that way get away from the import?

mathiaspeters:
I'm also wondering if it would be possible to create a new trait like `DebugSensitive` that we derive on structures that contain IP addresses, with an additional `#[sensitive]` attribute to put on the fields that need to be anonymized. Then in the `telio_log_generic` you can have a generic function for `T: DebugSensitive` to use the hide sensitive method and one generic function with the same name for just `T` that uses the normal `Debug` implementaiton. But this is all speculation.

crates/telio-utils/src/utils.rs
```diff
@@ -1,39 +1,61 @@
...
+    ( $log_func: ident, $msg: expr) => {tracing::$log_func!($msg)};
+    ( $log_func: ident, $format: expr, $($arg:expr),+) => {{
+        use std::net::{IpAddr, SocketAddr};
+
+        #[allow(dead_code)]
+        trait HideSensitiveInfo {
+            fn hide_me_maybe(&self) -> &'static str {
+                "******"
+            }
+        }
```
tomaszklak, line 15: This needs to distinguish debug from release builds - only in release mode should the values be hidden.

crates/telio-wg/src/adapter/boring.rs
```diff
@@ -7,7 +7,7 @@ use std::net::{IpAddr, Ipv4Addr};
 use std::sync::Arc;
 use std::{io, ops::Deref};
 use telio_crypto::PublicKey;
-use telio_utils::telio_log_debug;
+use telio_utils::{telio_log_debug, telio_log_generic};
```
packgron, line 10: I really don't like that now we have to have a magical depenency, it would be better if you use $crate::telio_log_generic and avoid this issue.

Same is allso a good way to rexpose all trace innteranls in same create, to avoid needing aditional tracing rexposes

stalowyjez:
Because we plan to get rid of our log wrappers this experiment makes no sense - closing this PR and moving to regex solution instead

PR #594, opened 2024-06-05
LLT-5125: Starcast transport component

mathiaspeters:
This is the transport component for telio starcast. It has two reponsibilities: forward packet from starcast transport socket to virtual starcast peer, and forward packet from virtual starcast peer to one of more peers over the starcast transport socket. Local multicast packets go to all peers, mapped unicast packets (responses to multicast from other peer) go to a single peer. 

Note: There are currently two commits, one for implementation and one for testing. These will be squashed before merging, but might be easier to review separately

- [x] README.md is updated

### DIFF
--- a/crates/telio-starcast/src/transport.rs
+++ b/crates/telio-starcast/src/transport.rs
@@ -1,0 +1,525 @@
+use std::{
+    net::{IpAddr, SocketAddr},
+    sync::Arc,
+};
+
+use async_trait::async_trait;
+use futures::{future::join_all, TryFutureExt};
+use ipnetwork::IpNetwork;
+use pnet_packet::{ipv4::MutableIpv4Packet, ipv6::MutableIpv6Packet};
+use telio_model::PublicKey;
+use tokio::net::UdpSocket;
+
+use telio_sockets::SocketPool;
+use telio_task::{
+    io::{wait_for_tx, Chan},
+    task_exec, Runtime, RuntimeExt, Task, WaitResponse,
+};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn};
+
+use crate::{
+    nat::{self, Nat, StarcastNat},
+    utils::{DualIpAddr, MutableIpPacket},
+};
+
+#[cfg(feature = "test-util")]
+const MULTICAST_TRANSPORT_PORT: u16 = 0;
+// Randomly chosen port for multicast transport
+// Will eventually be replaced with dynamic port selection
+#[cfg(not(feature = "test-util"))]
+const MULTICAST_TRANSPORT_PORT: u16 = 13569;
+const MAX_PACKET_SIZE: usize = u16::MAX as usize;
+
+/// Starcast transport-specific errors
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    // Could not bind to socket
+    #[error("Could not bind to socket")]
+    SocketBindError(String),
+    /// Invalid IP Packet
+    #[error("Invalid IP Packet")]
+    InvalidIpPacket,
+    /// Natted IP address does not belong to a peer
+    #[error("Natted IP address does not belong to a peer")]
+    NoIpForPeer,
+    /// Failed to fan out local multicast packet to some peers
+    #[error("Failed to fan out local multicast packet to peers {0:?}")]
+    FanoutFailed(Vec<PublicKey>),
+    /// Failed to send packet over socket
+    #[error("Failed to send packet over socket")]
+    SocketSendError,
+    /// Failed to send packet over channel
+    #[error("Failed to send packet over channel")]
+    ChannelSendError,
+    /// Failed to nat packet
+    #[error(transparent)]
+    NatError(#[from] nat::Error),
+    /// Task execution failed
+    #[error(transparent)]
+    TaskError(#[from] telio_task::ExecError),
+}
+
+/// A peer in the meshnet
+#[derive(Copy, Clone)]
+pub struct Peer {
+    /// The IP address of the peer
+    pub addr: SocketAddr,
+    /// The public key of the peer
+    pub public_key: PublicKey,
+}
+
+/// Config for transport component
+/// Contains fields that can change at runtime
+pub enum Config {
+    /// Simple transport config, has IP of peer but not port
+    Simple(Vec<(PublicKey, IpAddr)>),
+    /// Full transport config, has full socket address of peer
+    Full(Vec<(PublicKey, SocketAddr)>),
+}
+
+/// The starcast transport component
+/// Manages the task that does the actual transport
+pub struct Transport {
+    task: Task<State>,
+}
+
+impl Transport {
+    /// Starts the transport component
+    ///
+    /// Parameters:
+    /// * vpeer_ip - The IP address of the virtual starcast peer
+    /// * meshnet_ip - the meshnet IP of the node on which this component is currently running
+    /// * socket_pool - To create the transport socket
+    /// * packet_chan - A channel to send packets to and receive packets from the virtual peer component
+    /// * multicast_ips - IPs that native multicast packets can be sent to, e.g. 224.0.0.0/4
+    pub async fn start(
+        vpeer_ip: DualIpAddr,
+        meshnet_ip: IpAddr,
+        socket_pool: Arc<SocketPool>,
+        packet_chan: Chan<Vec<u8>>,
+        multicast_ips: Vec<IpNetwork>,
+    ) -> Result<Self, Error> {
+        let transport_socket = socket_pool
+            .new_internal_udp(SocketAddr::new(meshnet_ip, MULTICAST_TRANSPORT_PORT), None)
+            .await
+            .map_err(|e| {
+                telio_log_error!("Failed to bind multicast socket: {:?}", e);
+                Error::SocketBindError(e.to_string())
+            })?;
+        let transport_socket = Arc::new(transport_socket);
+
+        Ok(Self {
+            task: Task::start(State {
+                transport_socket,
+                packet_chan,
+                multicast_ips,
+                recv_buffer: vec![0; MAX_PACKET_SIZE],
+                nat: StarcastNat::new(vpeer_ip.ipv4, vpeer_ip.ipv6),
+                peers: Vec::new(),
+            }),
+        })
+    }
+
+    /// Update runtime values of the transport component
+    pub async fn configure(&self, config: Config) -> Result<(), Error> {
+        task_exec!(&self.task, async move |state| {
+            state.configure(config);
+            Ok(())
+        })
+        .await?;
+        Ok(())
+    }
+
+    /// Stop the transport component
+    pub async fn stop(self) {
+        let _ = self.task.stop().await.resume_unwind();
+    }
+}
+
+struct State {
+    transport_socket: Arc<UdpSocket>,
+    packet_chan: Chan<Vec<u8>>,
+    recv_buffer: Vec<u8>,
+    nat: StarcastNat,
+    peers: Vec<Peer>,
+    multicast_ips: Vec<IpNetwork>,
+}
+
+impl State {
+    fn configure(&mut self, config: Config) {
+        self.peers = match config {
+            Config::Simple(peers) => peers
+                .into_iter()
+                .map(|(public_key, addr)| Peer {
+                    public_key,
+                    addr: SocketAddr::new(addr, MULTICAST_TRANSPORT_PORT),
+                })
+                .collect(),
+            Config::Full(peers) => peers
+                .into_iter()
+                .map(|(public_key, addr)| Peer { public_key, addr })
+                .collect(),
+        };
+    }
+
+    async fn handle_local_multicast_packet(&mut self, packet: Vec<u8>) -> Result<(), Error> {
+        let failed_peers = join_all(self.peers.iter().map(|peer| {
+            self.transport_socket
+                .send_to(&packet, peer.addr)
+                .map_err(|_| peer.public_key)
+        }))
+        .await
+        .into_iter()
+        .filter_map(|res| match res {
+            Ok(_) => None,
+            Err(pk) => Some(pk),
+        })
+        .collect::<Vec<_>>();
+
+        if failed_peers.is_empty() {
+            Ok(())
+        } else {
+            Err(Error::FanoutFailed(failed_peers))
+        }
+    }
+
+    async fn handle_mapped_unicast_packet(&mut self, mut packet: Vec<u8>) -> Result<(), Error> {
+        let peer_ip = self
+            .nat
+            .translate_outgoing(&mut packet)
+            .map_err(Error::NatError)?;
+        let peer_ip = self
+            .peers
+            .iter()
+            .find(|peer| peer.addr.ip() == peer_ip)
+            .map(|p| p.addr)
+            .ok_or(Error::NoIpForPeer)?;
+        self.transport_socket
+            .send_to(&packet, peer_ip)
+            .await
+            .map(|_| ())
+            .map_err(|_| Error::SocketSendError)
+    }
+
+    fn has_multicast_dst(&self, packet: &mut [u8]) -> Result<bool, Error> {
+        let dst = match packet.first().ok_or(Error::InvalidIpPacket)? >> 4 {
+            4 => Self::get_packet_dst::<MutableIpv4Packet>(packet),
+            6 => Self::get_packet_dst::<MutableIpv6Packet>(packet),
+            _ => Err(Error::InvalidIpPacket),
+        }?;
+        Ok(self
+            .multicast_ips
+            .iter()
+            .any(|network| network.contains(dst)))
+    }
+
+    fn get_packet_dst<'a, P: MutableIpPacket<'a>>(packet: &'a mut [u8]) -> Result<IpAddr, Error> {
+        let packet = P::new(packet).ok_or(Error::InvalidIpPacket)?;
+        Ok(packet.get_destination().into())
+    }
+}
+
+#[async_trait]
+impl Runtime for State {
+    const NAME: &'static str = "StarcastTransport";
+
+    type Err = Error;
+
+    async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        let res = tokio::select! {
+            Some(mut packet) = self.packet_chan.rx.recv() => {
+                match self.has_multicast_dst(&mut packet) {
+                    Ok(val) if val => self.handle_local_multicast_packet(packet).await,
+                    Ok(_) => self.handle_mapped_unicast_packet(packet).await,
+                    Err(e) => Err(e),
+                }
+            }
+            Some((permit, Ok(bytes_read))) = wait_for_tx(&self.packet_chan.tx, self.transport_socket.recv(&mut self.recv_buffer)) => {
+                #[allow(clippy::expect_used)]
+                let mut packet = self.recv_buffer.get(..bytes_read).expect("We know bytes_read bytes should be in the buffer at this point").to_vec();
+                self.nat.translate_incoming(&mut packet)
+                    .map_err(Error::NatError)
+                    .map(|_| {
+                        let _ = permit.send(packet);
+                    })
+            }
+            else => {
+                telio_log_warn!("MutlicastListener: no events to wait on");
+                Ok(())
+            },
+        };
+        if let Err(e) = res {
+            telio_log_debug!("StarcastTransportError: {e}");
+        }
+        Self::next()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::net::SocketAddrV4;
+
+    use pnet_packet::{ipv4::Ipv4Packet, udp::UdpPacket};
+    use telio_crypto::SecretKey;
+
+    use super::*;
+    use crate::{assert_src_dst_v4, utils::test_utils::*};
+
+    const TEST_MAX_PACKET_SIZE: usize = 2048;
+
+    struct Scaffold {
+        task: Task<State>,
+        transport_socket: Arc<UdpSocket>,
+        channel: Chan<Vec<u8>>,
+        peers: Vec<(PublicKey, UdpSocket)>,
+    }
+
+    impl Scaffold {
+        async fn start() -> Self {
+            let transport_socket = Arc::new(Self::bind_local_socket().await);
+            let mut peers = Vec::with_capacity(3);
+            for _ in 0..3 {
+                peers.push((SecretKey::gen().public(), Self::bind_local_socket().await));
+            }
+            let (packet_chan, channel) = Chan::pipe();
+
+            let task_peers = peers
+                .iter()
+                .map(|(pk, s)| Peer {
+                    public_key: *pk,
+                    addr: s.local_addr().unwrap(),
+                })
+                .collect();
+            let multicast_ips = vec![IpNetwork::new("224.0.0.0".parse().unwrap(), 4).unwrap()];
+
+            let task = Task::start(State {
+                transport_socket: transport_socket.clone(),
+                packet_chan,
+                recv_buffer: vec![0; TEST_MAX_PACKET_SIZE],
+                nat: StarcastNat::new(IPV4_NAT_ADDR, IPV6_NAT_ADDR),
+                peers: task_peers,
+                multicast_ips,
+            });
+
+            Self {
+                task,
+                transport_socket,
+                channel,
+                peers,
+            }
+        }
+
+        async fn stop(self) {
+            let _ = self.task.stop().await;
+        }
+
+        async fn bind_local_socket() -> UdpSocket {
+            UdpSocket::bind(SocketAddr::new("127.0.0.1".parse().unwrap(), 0))
+                .await
+                .unwrap()
+        }
+    }
+
+    #[tokio::test]
+    async fn test_handle_local_multicast_packet() {
+        let scaffold = Scaffold::start().await;
+
+        let packet = make_udp_v4("127.0.0.1:12345", "224.0.0.251:5353");
+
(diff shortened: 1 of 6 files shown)


crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,252 @@
...
+const MULTICAST_TRANSPORT_PORT: u16 = 13569;
+const MAX_PACKET: usize = 2048;
+const IPV4_HEADER: usize = 20; // bytes
+
+/// Multicaster task that resends locally generated
+/// multicast traffic to other peers
+pub struct Multicaster {
+    wg: Arc<DynamicWg>,
+    multicast_socket: Arc<UdpSocket>,
+    local_socket: Arc<UdpSocket>,
```
packgron, line 33: you should only need one socket for transport

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,252 @@
...
+    fanout_task: Option<JoinHandle<()>>,
+    multicaster_ip: Arc<MulticasterIp>,
+}
+
+impl Multicaster {
+    pub async fn new(
+        local_socket: Arc<UdpSocket>,
+        wg: Arc<DynamicWg>,
+        socket_pool: Arc<SocketPool>,
+        packet_rx: Rx<Vec<u8>>,
```
packgron, line 45: so instead of receiving packet_rx and local socket, you should just get a singular `Chan<Vec<u8>>` that will be used for sending and receiving messages to and from @ThrasherLT MeshnetVirtualPeer

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,252 @@
...
+        multicaster_ip: Arc<MulticasterIp>,
+    ) -> Result<Self, String> {
+        let multicast_socket = socket_pool
+            .new_internal_udp(
+                SocketAddr::from(([0, 0, 0, 0], MULTICAST_TRANSPORT_PORT)),
```
packgron, line 50: Hmm I'm not 100% sure about this, we probably shoud do a bit of testing. But my initial assumtion is that you should bind this not to 0.0.0.0, but to private ip address. Basic question bing that if you bind it like this, then anyone from outside meshnet would be able to send packets to this sockets. As in general case we don't do anything that extra for a socket to be internal, it will likely be the case.

Also this should be a bit smarter in regards to what ip we are using, we can consult libdrop team a bit. but we should probably prefer using ipv6 if it's enabled.

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,252 @@
...
+        })
+    }
+
+    pub async fn start_multicaster(&mut self) {
+        let Some(packet_rx) = self.packet_rx.take() else {
+            telio_log_warn!("multicast can be started once for now");
+            return;
+        };
+
+        self.forward_to_vpeer_task = Some(tokio::spawn(Self::forward_to_vpeer_task(
```
packgron, line 76: I would preffer if you used impl Runtime and telio-task abstrcktion

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,281 @@
+use async_trait::async_trait;
```
packgron, line 1: NIT: in general i prefer to follow rust-lang code style for imports: https://doc.rust-lang.org/1.0.0/style/style/imports.html

Main thing beeing spliting std, 3rd-party creates, our crates, and mods into seperte groups

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,281 @@
...
+#[cfg(not(feature = "test-util"))]
+const MULTICAST_TRANSPORT_PORT: u16 = 13569;
+const MAX_PACKET_SIZE: usize = 2048;
+const IPV4_HEADER: usize = 20; // bytes
+const UDP_DEST_OFFSET: usize = 2;
+
+/// Multicaster task that resends locally generated
+/// multicast traffic to other peers
+pub struct Multicaster {
+    wg: Arc<Interface>,
```
packgron, line 29: So here we need to decide witch approach we take, theare are two:
- Having Arc<dyn Wg>, witch we query then we need to consult who is allive, this would meed we alwayas have an up to date state to who we need to send messages.
- Having some cached values of nodes, that is updated by wg_controller like vpeer does: https://github.com/NordSecurity/libtelio/blob/f1687494b60f8c0c6236bf607b89156ace20eb0d/crates/telio-starcast/src/multicast_peer.rs#L300

As it currently stands its broken on both fronts :D

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,281 @@
...
+
+#[async_trait]
+impl Runtime for MulticastListener {
+    const NAME: &'static str = "MulticastListener";
+
+    type Err = ();
+
+    async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        tokio::select! {
+            Some(packet) = self.packet_chan.rx.recv() => self.route_packet(packet, PacketOrigin::Channel).await,
```
packgron, line 270: We can probably allign terminology, with nat and say that this is route_incomming, or change names in Starcast Nat @jjanowsk your thoughts?

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,281 @@
...
+const MULTICAST_TRANSPORT_PORT: u16 = 0;
+#[cfg(not(feature = "test-util"))]
+const MULTICAST_TRANSPORT_PORT: u16 = 13569;
+const MAX_PACKET_SIZE: usize = 2048;
+const IPV4_HEADER: usize = 20; // bytes
+const UDP_DEST_OFFSET: usize = 2;
+
+/// Multicaster task that resends locally generated
+/// multicast traffic to other peers
+pub struct Multicaster {
```
packgron, line 28: NIT: Maybe just Transport, we could also do overkill and say StarcastTransport

crates/telio-starcast/src/multicaster.rs
packgron: Also i would prefer smaller 2 sammler functions over big function colltoled by an enum

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,281 @@
...
+impl Runtime for MulticastListener {
+    const NAME: &'static str = "MulticastListener";
+
+    type Err = ();
+
+    async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        tokio::select! {
+            Some(packet) = self.packet_chan.rx.recv() => self.route_packet(packet, PacketOrigin::Channel).await,
+            Ok(bytes_read) = self.multicast_socket.recv(&mut self.recv_buffer) => {
+                let packet = self.recv_buffer[..bytes_read].to_vec();
```
packgron, line 272: It would be smart to use the recv_from, and map ip to pk of other node, we will need it once extra filtering will start to happen.

crates/telio-starcast/src/multicaster.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    fn identify_packet_type(&self, dst_addr: IpAddr, origin: PacketOrigin) -> PacketType {
+        match (origin, dst_addr) {
+            (_, IpAddr::V6(_)) => {
+                telio_log_debug!("MulticastListener: IPv6 is not yet supported");
+                PacketType::Unsupported
+            }
+            (PacketOrigin::Channel, _) => PacketType::MappedUnicast,
+            (PacketOrigin::Socket, IpAddr::V4(addr)) if addr.octets()[0] == 244 => {
```
packgron, line 147: So multicast is 224.0.0.0/4, so this check is highly incaccurate, also I will need to hash this out a bit, but we will probably just use a subset of that also. So it would be could to make this an argument to entire compomentent, to make it easier i recommned using: Vec<https://docs.rs/ipnetwork/latest/ipnetwork/enum.IpNetwork.html>, we also will need to pass this to vpeer to change it's config.

packgron (approved):

packgron (changes requested):
-

crates/telio-starcast/src/multicaster.rs
mathiaspeters: `244` was a typo, it was supposed to be `224`, but using `IpNetwork` is a better option either way

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+#[cfg(not(feature = "test-util"))]
+const MULTICAST_TRANSPORT_PORT: u16 = 13569;
+const MAX_PACKET_SIZE: usize = 2048;
+const IPV4_HEADER: usize = 20; // bytes
+const UDP_DEST_OFFSET: usize = 2;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Could not bind to socket")]
+    SocketBindError(anyhow::Error),
```
packgron, line 35: anyhow should be avoided in library code

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+        self.peers = config.peers;
+        self.multicast_ips = config.multicast_ips;
+        if !self.features.ipv6 {
+            self.peers.retain(|p| p.is_ipv4());
+            self.multicast_ips.retain(|ip| ip.is_ipv4());
+        }
+    }
+
+    async fn handle_local_multicast_packet(&mut self, packet: Vec<u8>) {
+        let multicast_socket = self.multicast_socket.clone();
```
packgron, line 117: do you really need a clone here, I don't see a `move`?

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+use telio_model::features::Features;
+use telio_sockets::SocketPool;
+use telio_task::{io::Chan, task_exec, Runtime, RuntimeExt, Task, WaitResponse};
+use telio_utils::{telio_log_error, telio_log_warn};
+
+#[cfg(feature = "test-util")]
+const MULTICAST_TRANSPORT_PORT: u16 = 0;
+#[cfg(not(feature = "test-util"))]
+const MULTICAST_TRANSPORT_PORT: u16 = 13569;
+const MAX_PACKET_SIZE: usize = 2048;
```
packgron, line 28: To be fully correct it should be: 65535

packgron, line 28: To be fully correct it should be: 65535

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+#[async_trait]
+impl Runtime for State {
+    const NAME: &'static str = "StarcastTransport";
+
+    type Err = ();
+
+    async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        if !self.features.multicast {
```
packgron, line 210: We can refactor this out with proper integration, but we probably will want to not init this compnent at all if muticast is not enabled.

crates/telio-wg/src/uapi.rs
```diff
@@ -468,6 +468,26 @@ impl Peer {
         }
         dual_ip_addresses
     }
+
+    /// Returns whether the peer is a virtual peer for multicast
+    pub fn is_multicast_peer(&self) -> bool {
+        for ips in &self.allowed_ips {
+            if ips.network().eq(&Ipv4Addr::new(100, 64, 0, 5))
```
packgron, line 475: Add these magic ips to telio-model::constants

crates/telio-wg/src/uapi.rs
```diff
@@ -468,6 +468,26 @@ impl Peer {
...
+            }
+        }
+        false
+    }
+
+    /// Returns whether the peer is stun
+    pub fn is_stun_peer(&self) -> bool {
+        self.allowed_ips
+            .first()
+            .map(|ip| ip.network().eq(&Ipv4Addr::new(100, 64, 0, 4)))
```
packgron, line 488: Move to telio-model::constants

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+            return Self::sleep_forever();
+        }
+        tokio::select! {
+            Some(packet) = self.packet_chan.rx.recv() => self.handle_mapped_unicast_packet(packet).await,
```
packgron, line 214: So from the tunnel we can receive both multicast and unicast packets

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+            Ok((bytes_read, _sender)) = self.multicast_socket.recv_from(&mut self.recv_buffer) => {
+                let mut packet = self.recv_buffer[..bytes_read].to_vec();
+                if let Some(ip_packet) = MutableIpv4Packet::new(&mut packet) {
+                    let dst = IpAddr::V4(ip_packet.get_destination());
+                    if self.multicast_ips.iter().any(|network| network.ip() == dst) {
+                        self.handle_local_multicast_packet(packet).await
+                    } else {
+                        self.handle_misc_packet(packet).await;
```
packgron, line 222: So this is a unicast reply from other peer, or some trash

crates/telio-starcast/src/transport.rs
packgron: One is local multicast, so you localy generate a unicast message, and you need to send to other peers
Another then sombody replies to remotes multicast message, to a mapped port

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,232 @@
...
+            return Self::sleep_forever();
+        }
+        tokio::select! {
+            Some(packet) = self.packet_chan.rx.recv() => self.handle_mapped_unicast_packet(packet).await,
+            Ok((bytes_read, _sender)) = self.multicast_socket.recv_from(&mut self.recv_buffer) => {
+                let mut packet = self.recv_buffer[..bytes_read].to_vec();
+                if let Some(ip_packet) = MutableIpv4Packet::new(&mut packet) {
+                    let dst = IpAddr::V4(ip_packet.get_destination());
+                    if self.multicast_ips.iter().any(|network| network.ip() == dst) {
+                        self.handle_local_multicast_packet(packet).await
```
packgron, line 220: I would call this remote_multicast_packet

crates/telio-starcast/src/transport.rs
packgron: so remote_unicast (this should already have destination addres for you localy known address)

crates/telio-model/src/constants.rs
```diff
@@ -1,8 +1,17 @@
...
 pub const VPN_INTERNAL_IPV6: [u16; 8] = [0xfd74, 0x656c, 0x696f, 0, 0, 0, 0, 1];
 /// VPN IPv4 Non-Meshnet Address
 pub const VPN_EXTERNAL_IPV4: [u8; 4] = [10, 5, 0, 1];
+
+/// STUN IPv4 Meshnet Address
+pub const STUN_INTERNAL_IPV4: Ipv4Addr = Ipv4Addr::new(100, 64, 0, 4);
+/// Starcast Virtual Peer IPv4 Meshnet Address
+pub const STARCAST_VPEER_INTERNAL_IPV4: Ipv4Addr = Ipv4Addr::new(100, 64, 0, 5);
+/// Starcast Virtual Peer Multicast Address
+pub const STARCAST_VPEER_MULTICAST_IPV4: Ipv4Addr = Ipv4Addr::new(224, 0, 0, 4);
```
packgron, line 17: so it should be a ip network 224.0.0.0/4 not ip 224.0.0.4

packgron: Also, this multicast ip network will be more dynamic, so I think it's simplest to simply remote it and check just by a reserved ip.

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[error("Could not bind to socket")]
+    SocketBindError(String),
+    /// Task execution failed
+    #[error(transparent)]
+    Task(#[from] telio_task::ExecError),
+}
+
+pub struct Config {
+    peers: Vec<IpAddr>,
+    multicast_ips: Vec<IpNetwork>,
```
packgron, line 42: I think this can be moved, to  start as an arg, as it will not change after setup

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl State {
+    fn configure(&mut self, config: Config) {
+        self.peers = config.peers;
+        self.multicast_ips = config.multicast_ips;
+    }
+
+    async fn handle_local_multicast_packet(&mut self, packet: Vec<u8>) {
+        let fanout_res = join_all(self.peers.iter().map(|peer| {
```
packgron, line 109: This is one area, there it would be nice to have pk's also, and to have concrete peers we failed to send to.

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,243 @@
...
+            ip_packet.set_checksum(checksum(&ip_packet.to_immutable()));
+        } else {
+            return;
+        };
+        if let Some(mut udp_packet) = packet
+            .get_mut(IPV4_HEADER..)
+            .and_then(MutableUdpPacket::new)
+        {
+            udp_packet.set_destination(dst_port);
+            // Ignoring L4 checksum as it is already encapsulated in L3/L4 header
```
packgron, line 151: This is the fun part, that you actualy need to calculate both ip and udp :D. But we can wait for @jjanowsk nat

packgron (commented):
Look like you are mostly done from functional side, think how you could unit test this component

crates/telio-starcast/src/transport.rs
packgron: in trace bellow

crates/telio-starcast/src/transport.rs
mathiaspeters: That's why I haven't touched these packet rewrite sections since the PR opened, because it will be covered by the natting code anyways :smile:

crates/telio-model/src/constants.rs
mathiaspeters: This should only be used to check if a peer is a multicast peer, so checking within that network range should be enough. I'll make the change

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    type Err = Error;
+
+    async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        let res = tokio::select! {
+            Some(mut packet) = self.packet_chan.rx.recv() => {
+                match self.has_multicast_dst(&mut packet) {
+                    Ok(val) if val => self.handle_local_multicast_packet(packet).await,
+                    Ok(_) => self.handle_mapped_unicast_packet(packet).await,
+                    Err(e) => Err(e),
```
packgron, line 214: Do we want to stop entire transport task if error is encountered here?
I think we should just log it.

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,231 @@
...
+            .map(|_| ())
+            .map_err(|_| Error::SocketSendError)
+    }
+
+    async fn handle_remote_packet(&mut self, mut packet: Vec<u8>) -> Result<(), Error> {
+        self.nat
+            .translate_incoming(&mut packet)
+            .map_err(Error::NatError)?;
+        self.packet_chan
+            .tx
```
packgron, line 181: This should be via permit, recieved on select arm, as we are sending packets back and forth, we would run into constant error loop

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,231 @@
...
+            Err(Error::FanoutFailed(failed_peers))
+        }
+    }
+
+    async fn handle_mapped_unicast_packet(&mut self, mut packet: Vec<u8>) -> Result<(), Error> {
+        self.nat
+            .translate_outgoing(&mut packet)
+            .map_err(Error::NatError)?;
+        self.transport_socket
+            .send(&packet)
```
packgron, line 170: Where are you sending this packet?

transport socket is not connected to anything this will fail.
you need to use send to, and get dst ip as discussed

crates/telio-wg/src/uapi.rs
```diff
@@ -468,6 +471,26 @@ impl Peer {
         }
         dual_ip_addresses
     }
+
+    /// Returns whether the peer is a virtual peer for multicast
+    pub fn is_multicast_peer(&self) -> bool {
+        for ips in &self.allowed_ips {
+            if ips.network().eq(&STARCAST_VPEER_INTERNAL_IPV4)
```
packgron, line 478: Again, just check by internal ipv4.
Checking by multicast network does not really make sence

crates/telio-starcast/src/transport.rs
jjanowsk: I also forgot to return this address from translate ongoing. I wanted to do this during the integration. Those are the changes into `nat.rs`:
```
diff --git a/crates/telio-starcast/src/nat.rs b/crates/telio-starcast/src/nat.rs
index 9d5dc868c..21c0322d5 100644
--- a/crates/telio-starcast/src/nat.rs
+++ b/crates/telio-starcast/src/nat.rs
@@ -30,7 +30,7 @@ pub trait Nat {
     fn translate_incoming(&mut self, packet: &mut [u8]) -> Result<(), Error>;
     /// Translate outgoing packet (from the multicast peer to the transport socket)
     /// Change the destination to the peer's original ip and port
-    fn translate_outgoing(&mut self, packet: &mut [u8]) -> Result<(), Error>;
+    fn translate_outgoing(&mut self, packet: &mut [u8]) -> Result<IpAddr, Error>;
 }
 
 type Mapper = fn(src_addr: SocketAddr) -> u16;
@@ -140,7 +140,7 @@ impl StarcastNat {
     fn translate_outgoing_internal<'a, P: MutableIpPacket<'a>>(
         &mut self,
         packet: &'a mut [u8],
-    ) -> Result<(), Error> {
+    ) -> Result<IpAddr, Error> {
         let mut ip_packet = P::new(packet).ok_or(Error::PacketTooShort)?;
         if ip_packet.get_next_level_protocol() != IpNextHeaderProtocols::Udp {
             return Err(Error::UnexpectedTransportProtocol);
@@ -171,7 +171,7 @@ impl StarcastNat {
         ip_packet.set_destination(new_dst_ip);
         ip_packet.fix_ip_header_checksum(old_dst_ip, new_dst_ip);
 
-        Ok(())
+        Ok(new_dst_ip.into())
     }
 }
 
@@ -187,7 +187,7 @@ impl Nat for StarcastNat {
         }
     }
 
-    fn translate_outgoing(&mut self, packet: &mut [u8]) -> Result<(), Error> {
+    fn translate_outgoing(&mut self, packet: &mut [u8]) -> Result<IpAddr, Error> {
         match packet.first().ok_or(Error::PacketTooShort)? >> 4 {
             4 => self.translate_outgoing_internal::<MutableIpv4Packet>(packet),
             6 => self.translate_outgoing_internal::<MutableIpv6Packet>(packet),
```
As @packgron said you need to take this IP address and combine it with transport port to get SocketAddr to use in `sendto`

crates/telio-starcast/src/transport.rs
packgron: Ok you are not using try_send, you are using a send. This has a change to casue a deadlock then to task are both sending and reading from each other, if not done via permits.

These test's is an artificial example how this can happen: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-task/src/io.rs#L126

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,234 @@
...
+        } else {
+            Err(Error::FanoutFailed(failed_peers))
+        }
+    }
+
+    async fn handle_mapped_unicast_packet(&mut self, mut packet: Vec<u8>) -> Result<(), Error> {
+        let peer_ip = self
+            .nat
+            .translate_outgoing(&mut packet)
+            .map_err(Error::NatError)?;
```
packgron, line 172: NIT: would be nice to encapsutale inner error, you could simply add transparent

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,234 @@
...
+    SocketBindError(String),
+    /// Invalid IP Packet
+    #[error("Invalid IP Packet")]
+    InvalidIpPacket,
+    /// Failed to fan out local multicast packet to some peers
+    #[error("Failed to fan out local multicast packet to peers {0:?}")]
+    FanoutFailed(Vec<PublicKey>),
+    /// Failed to nat packet
+    #[error("Failed to nat packet: {0:?}")]
+    NatError(nat::Error),
```
packgron, line 45: Would it not make sence to also make it transparemnt, or even just add #[from] to avoid dational mappings

packgron (commented):
+1.0, once small suggestion, and you need to decide what you will do for unit tests.

tomasz-grz:
+0.25

ThrasherLT:
As far as I can see, the starcast virtual peer will integrate with the starcast transport through just one channel and everything seems in order in that regard.

jjanowsk (approved):
I'm leaving those two comments for later improvements

crates/telio-starcast/src/transport.rs
```diff
@@ -0,0 +1,533 @@
...
+    #[error(transparent)]
+    NatError(#[from] nat::Error),
+    /// Task execution failed
+    #[error(transparent)]
+    TaskError(#[from] telio_task::ExecError),
+}
+
+/// A peer in the meshnet
+#[derive(Copy, Clone)]
+pub struct Peer {
```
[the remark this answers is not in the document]
jjanowsk: I think it's fine

PR #597, opened 2024-06-05
LLT-5246 Add netsh checks on config changes in win VM

matislovas:
### Problem
Some commands in VM do not generate results instantly. This was observed primarily on Windows, when setting IP addresses on interface using `netsh` command.

### Solution
After command execution, launch a "check-and-grep-like" commands for some predetermined amount of time, to search for a command execution result.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/command_grepper.py
+++ b/nat-lab/tests/utils/command_grepper.py
@@ -1,0 +1,116 @@
+import asyncio
+from typing import List, Optional
+from utils.connection import Connection
+from utils.process import ProcessExecError
+
+
+class CommandGrepper:
+    _connection: Connection
+    _check_cmd: List[str]
+    _timeout: Optional[float]
+
+    def __init__(
+        self,
+        connection: Connection,
+        check_cmd: List[str],
+        timeout: Optional[float] = None,
+    ):
+        """
+        A class that runs a command and checks if the expected strings are present or not present in the output.
+
+        Args:
+            connection: Connection object to container to run the command on.
+            check_cmd: Command to run.
+            timeout: Timeout for the check. If the check does not have expected results within the timeout, it will return False.
+        """
+        self._connection = connection
+        self._check_cmd = check_cmd
+        self._timeout = timeout
+
+    async def check_exists(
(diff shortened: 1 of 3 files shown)


nat-lab/tests/utils/command_grepper.py
```diff
@@ -0,0 +1,104 @@
...
+                self._timeout,
+            )
+        except ProcessExecError:
+            return False
+
+    # Check if the expected strings are not present in the output of the command.
+    # If only the primary string is provided, check for its absence in whole stdout.
+    # If secondary strings are also provided, checks, if primary and secondary are NOT in one line
+    async def check_not_exists(
+        self, exp: str, exp_secondary: Optional[List[str]] = None
```
mathiaspeters, line 40: Nit: `exp` should be renamed to `exp_primary` for consistency reasons and to make the "primary" reference in the doc comment clearer

nat-lab/tests/utils/command_grepper.py
```diff
@@ -0,0 +1,104 @@
...
+                self._timeout,
+            )
+        except ProcessExecError:
+            return False
+
+    async def _run_check(
+        self, exp_primary: str, exp_secondary: Optional[List[str]], exists: bool
+    ) -> bool:
+        while True:
+            process = await self._connection.create_process(self._check_cmd).execute()
```
mathiaspeters, line 54: Is it expected to rerun the command on each iteration?

matislovas: Yup, that's the idea. Run _check_ command every second, and when it succeeds - `return True`. Or timeout and die.

nat-lab/tests/utils/command_grepper.py
matislovas: Fixed

mathiaspeters:
+1

ThrasherLT:
lgtm +10 maybe would like it if there were less duplicated code between the exists and not_exists functions, but their usage seems convenient, so won't block the merge on that

ThrasherLT (approved):

PR #599, opened 2024-06-05
Make failed SessionKeeper startup non-fatal for the meshnet

tomaszklak:
### Problem
Failed `SessionKeeper` startup is "fatal" - whole meshnet startup fails. Yet it only prevents the direct part of the meshnet from working.

### Solution
When the SessionKeeper fails to start (for whatever reason), make it as if the config didn't have the 'direct' section and continue the startup. This will result in meshnet still working, but only in the relayed mode.

- [x] README.md is updated

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -245,6 +245,9 @@ impl Protector for NativeProtector {
             socks.notify.notify_waiters();
         }
     }
+
+    /// This is a no-op on apple platforms.
+    fn set_fwmark(&self, _fwmark: u32) {}
 }
 
 #[derive(Debug)]
--- a/.unreleased/LLT-4641
+++ b/.unreleased/LLT-4641
@@ -1,0 +1,1 @@
+Make SessionKeeper startup failures non-fatal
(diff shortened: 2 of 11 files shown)


mathiaspeters:
+1

crates/telio-sockets/src/protector/apple.rs
```diff
@@ -245,6 +245,8 @@ impl Protector for NativeProtector {
             socks.notify.notify_waiters();
         }
     }
+
+    fn set_fwmark(&self, _fwmark: u32) {}
```
jjanowsk, line 249: What was the reason behind unifying the trait between the platforms? Just to get rid of `cfg`s in multiple places? Maybe we can use `unimplemented!();` to make it clear that those should not be used on particular platforms?

tomaszklak: Yes, I decided that we don't really gain anything by not having one or two methods on some systems. It's not that they shouldn't be used - the just have an empty implementation. I think that's pretty normal for an interface to have a no-op as an implementation for some of its methods and I would prefer to avoid something that can theoretically panic at runtime when called. Can we agree on some doc comments for those 'new' methods that will explain that currently they have no effect?

jjanowsk: Yes, I thought  mostly about documentation value when proposing `unimplemented!();` and if you prefer comments to be on the safe side I'm fine with those

tomaszklak: Comments added :)

jjanowsk (approved):
+1

PR #601, opened 2024-06-06
LLT 5183 list local if ip

Hasan6979:
### Problem
https://github.com/NordSecurity/libtelio/pull/529 introduces filtering in firewall on incoming packets on local IP. Calling if-addr on every packet is an expensive process.

### Solution
Cache local IPs, and update it on notification from the OS.

P..S.
This PR is currently divided into 5 commits. I have tried to separate them as logically as possible. There might be some overlap. Each commit can be looked at separately, but some things (very minor things) might have changed in future commits.

1 - Separates out fetching local interfaces into its own module
2 - Creates a network monitor module and adds network monitoring for Apple. Apple has a callback for network and link changes. As far as I could see, there is no need to deregister or cleanup.
3 - Creates a network monitor for linux and windows. Uses netlink sockets for monitoring in linux, and network callback in windows. The callback in windows has to be deregistered after use, but a caveat that it has to be done on a separate thread to avoid deadlock.
4 - Adds a trigger for Android. Android unfortunately does not support netlink sockets or any other callback, and its trigger would come from App side
5 - Adds an integration test for network monitor. Integration test adds an interface and then removes it to trigger network monitor. Network monitor gets triggered on both occasions though.

- [x] README.md is updated

### DIFF
--- a/crates/telio-network-monitors/Cargo.toml
+++ b/crates/telio-network-monitors/Cargo.toml
@@ -1,0 +1,38 @@
+[package]
+name = "telio-network-monitors"
+version = "0.1.0"
+edition = "2018"
+license = "GPL-3.0-only"
+repository = "https://github.com/NordSecurity/libtelio"
+publish = false
+
+[dependencies]
+if-addrs.workspace = true
+ipnet.workspace = true
+telio-utils.workspace = true
+tracing.workspace = true
+parking_lot.workspace = true
+tokio = { workspace = true, features = ["sync", "test-util", "time", "rt", "macros"] }
+once_cell.workspace = true
+mockall.workspace = true
+
+[target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))'.dependencies]
+dispatch = { git = "https://github.com/NordSecurity/rust-dispatch.git", rev = "13447cd7221a74ebcce1277ae0cfc9a421a28ec5" }
+network-framework-sys = "0.1"
+block = "0.1"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+neli = { version = "0.6.4", features = ["async"] }
+
+[target.'cfg(windows)'.dependencies]
+winapi = { workspace = true, features = ["iphlpapi", "netioapi", "winnt", "ws2def"] }
+windows = { version = "0.34.0", features = [
+    "alloc",
+    "Win32_Networking_WinSock",
+    "Win32_NetworkManagement_IpHelper",
+] }
+
+[dev-dependencies]
+assert_matches = "1.5.0"
+lazy_static.workspace = true
+serial_test = "3.1.1"
--- a/crates/telio-network-monitors/src/apple.rs
+++ b/crates/telio-network-monitors/src/apple.rs
@@ -1,0 +1,80 @@
+//! Module to monitor network changes in Apple
+
+use network_framework_sys::{
+    nw_path_monitor_create, nw_path_monitor_set_queue, nw_path_monitor_start, nw_path_monitor_t,
+    nw_path_t,
+};
+use std::ffi::{c_long, c_void};
+use telio_utils::{telio_log_info, telio_log_warn};
+
+use crate::monitor::PATH_CHANGE_BROADCAST;
+
+/// Dispatch queue priority as high priority queue
+pub const DISPATCH_QUEUE_PRIORITY_HIGH: c_long = 2;
+/// Dispatch queue priority as default queue
+pub const DISPATCH_QUEUE_PRIORITY_DEFAULT: c_long = 0;
+/// Dispatch queue priority as low priority queue
+pub const DISPATCH_QUEUE_PRIORITY_LOW: c_long = -2;
+/// Dispatch queue priority as background queue
+pub const DISPATCH_QUEUE_PRIORITY_BACKGROUND: c_long = -1 << 15;
+
+extern "C" {
+    /// Obj-c signature:
+    /// void nw_path_monitor_set_update_handler(nw_path_monitor_t monitor, nw_path_monitor_update_handler_t update_handler);
+    /// typedef void (^nw_path_monitor_update_handler_t)(nw_path_t path);
+    pub fn nw_path_monitor_set_update_handler(
+        monitor: nw_path_monitor_t,
+        update_handler: *const c_void,
+    );
+    /// Obj-c signature:
+    /// void nw_path_enumerate_interfaces(nw_path_t path, nw_path_enumerate_interfaces_block_t enumerate_block);
+    /// typedef bool (^nw_path_enumerate_interfaces_block_t)(nw_interface_t interface);
+    pub fn nw_path_enumerate_interfaces(path: nw_path_t, enumerate_block: *const c_void);
+}
+
+/// This function configuresa a network path monitor using Apple's networking framework that tracks
+/// changes in the network path and updates the local interface cache.
+///
+/// # How it Works
+///
+/// 1. **Objective-C Blocks**: The Apple `Network.framework` uses blocks as callbacks
+///    for handling asynchronous events.
+///
+/// 3. **Global Queue Setup**: The monitor runs on a background queue using Grand Central Dispatch (GCD).
+///    This allows it to handle updates without blocking the main thread.
+///
+/// 4. **Broadcast Notification**: After the interfaces are updated, a notification is sent via a broadcast channel
+///    (`PATH_CHANGE_BROADCAST`) to inform other parts of the system that the path has changed.
+///
+/// For more details on Apple's `Network.framework` see:
+/// - [Apple Network.framework Documentation](https://developer.apple.com/documentation/network)
+pub fn setup_network_monitor() {
+    let update_handler = block::ConcreteBlock::new(|_path: nw_path_t| {
+        if let Err(e) = PATH_CHANGE_BROADCAST.send(()) {
+            telio_log_warn!("Failed to notify about changed path, error: {e}");
+        }
+
+        telio_log_info!("Path change notification sent");
+    })
+    .copy();
+    unsafe {
+        let monitor = nw_path_monitor_create();
+        if monitor.is_null() {
+            telio_log_warn!("Failed to start network path monitor");
+            return;
+        }
+
+        let queue = dispatch::ffi::dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0);
+        if queue.is_null() {
+            telio_log_warn!("Failed to get global background queue");
+            return;
+        }
+        nw_path_monitor_set_queue(monitor, std::mem::transmute(queue));
+
+        nw_path_monitor_set_update_handler(
+            monitor,
+            &*update_handler as *const block::Block<_, _> as *const c_void,
+        );
+        nw_path_monitor_start(monitor);
+    }
+}
--- a/crates/telio-network-monitors/src/lib.rs
+++ b/crates/telio-network-monitors/src/lib.rs
@@ -1,0 +1,19 @@
+#![deny(missing_docs)]
+
+//! Network monitor module
+pub mod monitor;
+
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+/// Utility to get monitor network on apple
+pub mod apple;
+
+#[cfg(target_os = "linux")]
+/// Utility to get monitor network on linux
+pub mod linux;
+
+#[cfg(target_os = "windows")]
+/// Utility to get monitor network on windows
+pub mod windows;
+
+/// Get local interfaces
+pub mod local_interfaces;
--- a/crates/telio-network-monitors/src/linux.rs
+++ b/crates/telio-network-monitors/src/linux.rs
@@ -1,0 +1,72 @@
+use crate::monitor::PATH_CHANGE_BROADCAST;
+use neli::{
+    consts::socket::NlFamily::Route,
+    socket::{tokio::NlSocket as TokioSocket, NlSocket},
+    FromBytesWithInput,
+};
+use std::io;
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn};
+use tokio::task::JoinHandle;
+
+const RTMGRP_LINK: u32 = 0x0001;
+const RTMGRP_IPV4_IFADDR: u32 = 0x0010;
+
+/// Sets up linux network monitoring task
+pub async fn setup_network_monitor() -> Option<JoinHandle<io::Result<()>>> {
+    match open_netlink_socket().await {
+        Ok(sock) => read_event(sock).await,
+        Err(e) => {
+            telio_log_error!("Unable to open network monitor socket: {e:?}");
+            None
+        }
+    }
+}
+
+async fn open_netlink_socket() -> Result<TokioSocket, io::Error> {
+    let s = NlSocket::connect(Route, None, &[RTMGRP_LINK, RTMGRP_IPV4_IFADDR])?;
+    let sock = TokioSocket::new(s)?;
+    telio_log_debug!("Network monitor socket opened");
+    Ok(sock)
+}
+
+async fn read_event(mut sock: TokioSocket) -> Option<JoinHandle<io::Result<()>>> {
+    Some(tokio::spawn({
+        let mut buffer: Vec<u8> = Vec::new();
+
+        // Uninitialized object to satisfy recv constraints.
+        // Since data isn't required from recv, hence initialzing
+        // it properly isn't necessary as well.
+        #[derive(Debug, FromBytesWithInput)]
+        struct Payload;
+
+        async move {
+            loop {
+                let _ = sock
+                    .recv::<neli::consts::rtnl::Rtm, Payload>(&mut buffer)
+                    .await;
+                {
+                    telio_log_debug!("Network path updated");
+                    if let Err(e) = PATH_CHANGE_BROADCAST.send(()) {
+                        telio_log_warn!("Failed to notify about changed path: {e}");
+                    }
+                }
+            }
+        }
+    }))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_unregister_callback() {
+        let monitor_handle = setup_network_monitor().await;
+
+        if let Some(handle) = monitor_handle {
+            handle.abort();
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+            assert!(handle.is_finished());
+        }
+    }
+}
--- a/crates/telio-network-monitors/src/local_interfaces.rs
+++ b/crates/telio-network-monitors/src/local_interfaces.rs
@@ -1,0 +1,107 @@
+use ipnet::Ipv4Net;
+use mockall::automock;
+use std::net::{IpAddr, Ipv4Addr};
+
+/// A trait that provides an interface for retrieving network interface addresses.
+#[automock]
+pub trait GetIfAddrs: Send + Sync + Default + 'static {
+    /// Fetches the list of network interfaces available on the system.
+    /// Returns a `Result` which is contains a `Vec` of `if_addrs::Interface`
+    /// representing the available network interfaces.
+    fn get(&self) -> std::io::Result<Vec<if_addrs::Interface>>;
+}
+
+/// A concrete implementation of the `GetIfAddrs` trait that retrieves network interface
+/// addresses from the operating system using the `if_addrs` crate.
+#[derive(Default)]
+pub struct SystemGetIfAddrs;
+impl GetIfAddrs for SystemGetIfAddrs {
+    fn get(&self) -> std::io::Result<Vec<if_addrs::Interface>> {
+        if_addrs::get_if_addrs()
+    }
+}
+
+/// Method that returns vector of Interfaces on the system
+/// Filtering out meshnet and loopback IP
+pub fn gather_local_interfaces<G: GetIfAddrs>(
+    get_if_addr: &G,
+) -> std::io::Result<Vec<if_addrs::Interface>> {
+    let shared_range: Ipv4Net = Ipv4Net::new(Ipv4Addr::new(100, 64, 0, 0), 10).unwrap_or_default();
+    Ok((*get_if_addr)
+        .get()?
+        .into_iter()
+        .filter(|x| {
+            !x.addr.is_loopback() && {
+                match x.addr.ip() {
+                    // Filter 100.64/10 libtelio's meshnet network.
+                    IpAddr::V4(v4) => !shared_range.contains(&v4),
+                    // Filter IPv6
+                    _ => false,
+                }
+            }
+        })
+        .collect())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::net::{Ipv4Addr, Ipv6Addr};
+
+    #[tokio::test]
+    async fn gather_local_interfaces_filtering() {
+        let mut get_if_addrs_mock = MockGetIfAddrs::new();
+        get_if_addrs_mock.expect_get().return_once(|| {
+            Ok(vec![
+                if_addrs::Interface {
+                    name: "localhost".to_owned(),
+                    addr: if_addrs::IfAddr::V4(if_addrs::Ifv4Addr {
+                        ip: Ipv4Addr::new(127, 0, 0, 1),
+                        netmask: Ipv4Addr::new(255, 0, 0, 0),
+                        broadcast: None,
+                    }),
+                    index: None,
+                    #[cfg(windows)]
+                    adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+                },
+                if_addrs::Interface {
+                    name: "correct".to_owned(),
+                    addr: if_addrs::IfAddr::V4(if_addrs::Ifv4Addr {
+                        ip: Ipv4Addr::new(10, 0, 0, 1),
+                        netmask: Ipv4Addr::new(255, 255, 255, 0),
+                        broadcast: None,
+                    }),
+                    index: None,
+                    #[cfg(windows)]
+                    adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+                },
+                if_addrs::Interface {
+                    name: "internal".to_owned(),
+                    addr: if_addrs::IfAddr::V4(if_addrs::Ifv4Addr {
+                        ip: Ipv4Addr::new(100, 64, 0, 1),
+                        netmask: Ipv4Addr::new(255, 192, 0, 0),
+                        broadcast: None,
+                    }),
+                    index: None,
+                    #[cfg(windows)]
+                    adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+                },
+                if_addrs::Interface {
+                    name: "ipv6".to_owned(),
+                    addr: if_addrs::IfAddr::V6(if_addrs::Ifv6Addr {
+                        ip: Ipv6Addr::new(0xfd74, 0x656c, 0x696f, 0, 0x12, 0x34, 0x56, 0),
+                        netmask: Ipv6Addr::new(255, 255, 255, 255, 0, 0, 0, 0),
+                        broadcast: None,
+                    }),
+                    index: None,
+                    #[cfg(windows)]
+                    adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+                },
+            ])
+        });
+
+        let interfaces = gather_local_interfaces(&get_if_addrs_mock).unwrap();
+        assert!(interfaces.len() == 1);
+        assert!(interfaces[0].name == "correct");
+    }
+}
--- a/crates/telio-network-monitors/src/monitor.rs
+++ b/crates/telio-network-monitors/src/monitor.rs
@@ -1,0 +1,185 @@
+//! Module to monitor changes in network.
+//! Get notified when network paths change
+//! and update local IP address cache
+use crate::{local_interfaces::gather_local_interfaces, local_interfaces::GetIfAddrs};
+use once_cell::sync::Lazy;
+use parking_lot::Mutex;
+use std::io;
+use telio_utils::{telio_log_debug, telio_log_warn};
+use tokio::{sync::broadcast::Sender, task::JoinHandle};
+/// Sender to notify if there is a change in OS interface order
+pub static PATH_CHANGE_BROADCAST: Lazy<Sender<()>> = Lazy::new(|| Sender::new(2));
+/// Vector containing all local interfaces
+pub static LOCAL_ADDRS_CACHE: Mutex<Vec<if_addrs::Interface>> = Mutex::new(Vec::new());
+#[cfg(all(
+    not(test),
+    any(target_os = "macos", target_os = "ios", target_os = "tvos")
+))]
+static NW_PATH_MONITOR_START: std::sync::Once = std::sync::Once::new();
+
+#[derive(Debug)]
+/// Struct to monitor network
+pub struct NetworkMonitor {
+    if_cache_updater_handle: Option<JoinHandle<io::Result<()>>>,
+    // Mac and Windows use callbacks, whereas in Linux a socket is registered
+    // with netlink multicast group messages
+    #[cfg(target_os = "linux")]
+    monitor_handle: Option<JoinHandle<io::Result<()>>>,
+    #[cfg(target_os = "windows")]
+    monitor_handle: crate::windows::SafeHandle,
+}
+
+fn save_local_interfaces<G: GetIfAddrs>(get_if_addr: &G) {
+    match gather_local_interfaces(get_if_addr) {
+        Ok(v) => {
+            telio_log_debug!("Updating local addr cache");
+            *(LOCAL_ADDRS_CACHE.lock()) = v;
+        }
+        Err(e) => telio_log_warn!("Unable to get local interfaces {e}"),
+    }
+}
+
+impl NetworkMonitor {
+    /// Sets up and spawns network monitor
+    pub async fn new<G: GetIfAddrs>(get_if_addr: G) -> std::io::Result<NetworkMonitor> {
+        save_local_interfaces(&get_if_addr);
+
+        let if_cache_updater_handle = Some(tokio::spawn({
+            let mut notify = PATH_CHANGE_BROADCAST.subscribe();
+            async move {
+                loop {
+                    match notify.recv().await {
+                        Ok(()) => save_local_interfaces(&get_if_addr),
+                        Err(e) => {
+                            telio_log_warn!("Failed to receive new path for sockets ({e}): ");
+                        }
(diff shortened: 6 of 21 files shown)


crates/telio-network-monitors/src/linux.rs
```diff
@@ -0,0 +1,66 @@
...
+            msg_name: &mut sa as *mut _ as *mut _,
+            msg_namelen: mem::size_of_val(&sa) as u32,
+            msg_iov: &mut iov,
+            msg_iovlen: 1,
+            msg_control: std::ptr::null_mut(),
+            msg_controllen: 0,
+            msg_flags: 0,
+        };
+        loop {
+            let _ = unsafe { recvmsg(socket, &mut msg, 0) };
```
Jauler, line 59: Note that it is note super correct to have blocking IO inside of tokio task. But there is [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) to resolve this.

crates/telio-network-monitors/src/mac.rs
```diff
@@ -0,0 +1,103 @@
...
+                names.borrow()
+            );
+        }
+
+        telio_log_info!(
+            "Path change notification sent, current network path in os preference order: {:?}",
+            names.borrow()
+        );
+    })
+    .copy();
```
tomasz-grz, line 82: Why are we making a copy already when declaring it?

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,72 @@
+#[cfg(feature = "test-util")]
+use if_addrs::{IfAddr, Ifv4Addr, Interface};
+
+use ipnet::{Ipv4Net, PrefixLenError};
+use std::net::{IpAddr, Ipv4Addr};
+use thiserror::Error as TError;
+
+#[cfg_attr(any(test, feature = "mockall"), mockall::automock)]
+/// Trait to get IF Address
```
lcruz99, line 9: The code is in such a good readable state that several comments are trivial. Either modify them so that they explicitly add extra information or just remove them.

lcruz99: What I mean by "explicitly add extra information" is:
- how to use the code on a higher level
- explain its behaviour if the functionality is complex or isn't that obvious
- justifying the changes (eg: "This was changed due to X, see Y"
- some other info that's not implicit when reading the code

crates/telio-network-monitors/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![deny(missing_docs)]
+
+//! Network monitor module
+pub mod monitor;
+
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+/// Utility to get monitor network on apple
+pub mod mac;
```
lcruz99, line 8: IMHO, the name for this mod is inaccurate, "mac" doesn't describe this group of OSes.. eg: "darwin" would be more suitable.

crates/telio-network-monitors/src/mac.rs
```diff
@@ -0,0 +1,103 @@
...
+        monitor: nw_path_monitor_t,
+        update_handler: *const c_void,
+    );
+    /// Obj-c signature:
+    /// void nw_path_enumerate_interfaces(nw_path_t path, nw_path_enumerate_interfaces_block_t enumerate_block);
+    /// typedef bool (^nw_path_enumerate_interfaces_block_t)(nw_interface_t interface);
+    pub fn nw_path_enumerate_interfaces(path: nw_path_t, enumerate_block: *const c_void);
+}
+
+/// Setup network path monitor in Apple framework
```
lcruz99, line 43: Again the fn comments are not the useful. 

It's clear for the dev that this function does that, but for instance "why do we need an objective C block?" and "how does monitor is setup?". If useful you can link some website for references.

crates/telio-network-monitors/src/mac.rs
tomasz-grz: Ok to answer to myself, this is copying the stack allocated `ConcreteBlock` to the heap as an `RcBlock` before
passing it to Objective-C, in order to prevent a double free.

crates/telio-network-monitors/src/mac.rs
lcruz99: also, shouldn't we state/mark it as unsafe?

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub async fn new<G>(if_addr: G) -> Result<NetworkMonitor, GetIFError>
+    where
+        G: GetIfAddrs + Clone,
+    {
+        if let Ok(mut guard) = LOCAL_ADDRS_CACHE.lock() {
+            *guard = local_interfaces::gather_local_interfaces(&if_addr)?;
+            telio_log_debug!("local cache {:?}", *guard);
+        }
+
+        let nw_path_monitor_monitor_handle = Some(tokio::spawn({
```
tomasz-grz, line 40: Looking at this again if I understand correctly this is platform independent code, and not setting up `NWPathMonitor`? If that is the case what do you think about calling it something other than `nw_path_monitor`?

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+                    ip: Ipv4Addr::new(10, 0, 0, 1),
+                    netmask: Ipv4Addr::new(255, 255, 255, 0),
+                    broadcast: None,
+                }),
+                index: None,
+                #[cfg(windows)]
+                adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+            }])
+        });
+        let network_monitor = NetworkMonitor::new(get_if_addrs_mock).await.unwrap();
```
lcruz99, line 120: unused var

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+                        Err(e) => {
+                            telio_log_warn!("Failed to receive new path for sockets ({e}): ");
+                        }
+                    }
+                }
+            }
+        }));
+
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+        NETWORK_PATH_MONITOR_START.call_once(crate::mac::setup_network_monitor);
```
tomasz-grz, line 63: `NWPathMonitor` is actually created in here?

tomasz-grz (changes requested):

crates/telio-network-monitors/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+license = "GPL-3.0-only"
+repository = "https://github.com/NordSecurity/libtelio"
+publish = false
+
+[dependencies]
+if-addrs.workspace = true
+telio-utils.workspace = true
+tracing.workspace = true
+parking_lot.workspace = true
+tokio = { workspace = true, features = ["full"] }
```
lcruz99, line 14: Do we need the full featured tokio?

I was wondering if the `unused-features` linter job could be bugged 😅

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+        #[cfg(target_os = "windows")]
+        let windows_monitor_handle = crate::windows::setup_network_monitor();
+
+        Ok(Self {
+            nw_path_monitor_monitor_handle,
+            #[cfg(target_os = "linux")]
+            linux_monitor_handle,
+            #[cfg(target_os = "windows")]
+            windows_monitor_handle,
+        })
```
lcruz99, line 77: I think we don't need to have different names for the `monitor_handle`
```suggestion
        #[cfg(target_os = "linux")]
        let monitor_handle = crate::linux::setup_network_monitor().await;

        #[cfg(target_os = "windows")]
        let monitor_handle = crate::windows::setup_network_monitor();

        Ok(Self {
            nw_path_monitor_monitor_handle,
            #[cfg(any(target_os = "windows", target_os = "linux"))]
            monitor_handle,
        })
```

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+    use telio_utils::local_interfaces::MockGetIfAddrs;
+
+    use super::*;
+    use std::net::Ipv4Addr;
+
+    #[tokio::test]
+    async fn test_path_change_notification() {
+        let mut get_if_addrs_mock = MockGetIfAddrs::new();
+        get_if_addrs_mock.expect_get().times(2).returning(|| {
+            Ok(vec![if_addrs::Interface {
```
lcruz99, line 108: something's off, this test is passing even though this expectation is not met. I was playing with it locally and changed time(2) -> times(1) and although the logs print:
`MockGetIfAddrs::get: Expectation(<anything>) called 2 times which is more than the expected 1`
the test still passes, which shouldn't right?

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[derive(Default, Clone)]
+/// Defination of struct to get system interfaces
+pub struct SystemGetIfAddrs;
+impl GetIfAddrs for SystemGetIfAddrs {
+    fn get(&self) -> std::io::Result<Vec<if_addrs::Interface>> {
+        #[cfg(not(feature = "test-util"))]
+        return if_addrs::get_if_addrs();
+        #[cfg(feature = "test-util")]
```
lcruz99, line 40: I think you you can use async traits on test mods instead.

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+                adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+            }])
+        });
+        let network_monitor = NetworkMonitor::new(get_if_addrs_mock).await.unwrap();
+
+        if let Err(e) = PATH_CHANGE_BROADCAST.send(()) {
+            println!("Failed to notify about changed path, error: {e}");
+        }
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+    }
```
lcruz99, line 126: Shouldn't we test the content of the returned interfaces?

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,72 @@
+#[cfg(feature = "test-util")]
```
lcruz99, line 1: Try to avoid having this conditional compilation and use proper dev build dependencies

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,72 @@
...
+        .into_iter()
+        .filter(|x| !x.addr.is_loopback())
+        .filter(|x| match x.addr.ip() {
+            // Filter 100.64/10 libtelio's meshnet network.
+            IpAddr::V4(v4) => !shared_range.contains(&v4),
+            // Filter IPv6
+            _ => false,
+        })
+        .collect())
+}
```
lcruz99, line 72: Please add tests.

nat-lab/tests/test_network_monitor.py
```diff
@@ -0,0 +1,50 @@
...
+    alpha_setup_params: SetupParameters,
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        env = await setup_mesh_nodes(exit_stack, [alpha_setup_params])
+        [client_alpha] = env.clients
+
+        time.sleep(DEFAULT_WAITING_TIME)
+        await client_alpha.restart_interface()
+
+        time.sleep(DEFAULT_WAITING_TIME)
```
lcruz99, line 49: This is blocking sleep on an async function, is intentional?
Despite of that, you can instead do an async `wait_for()` call to `wait_for_log()` with a timeout.

crates/telio-network-monitors/src/monitor.rs
lcruz99: you're probably missing the checkpoint() at the end of test.

lcruz99: .. or mock is moving out of scope and thus cannot assert the expectations.

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+                #[cfg(windows)]
+                adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+            }])
+        });
+        let network_monitor = NetworkMonitor::new(get_if_addrs_mock).await.unwrap();
+
+        if let Err(e) = PATH_CHANGE_BROADCAST.send(()) {
+            println!("Failed to notify about changed path, error: {e}");
+        }
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
```
lcruz99, line 125: why do we need to wait at the end of test?

Hasan6979: We need as it takes a bit of time for the callback/interrupt to occur from the OS and then monitor notifying it since the purpose of this test is to mainly see whether send() triggers the if.get().

crates/telio-utils/src/local_interfaces.rs
Hasan6979: There is already a test for it  https://github.com/NordSecurity/libtelio/blob/LLT-5183_list_local_IF_IP/crates/telio-traversal/src/endpoint_providers/local.rs#L396. 
I will move that test to this module.

crates/telio-network-monitors/src/monitor.rs
Hasan6979: This is intentional otherwise compiler was omitting this completely.

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Vector containing all local interfaces
+pub static LOCAL_ADDRS_CACHE: Lazy<Arc<StdMutex<Vec<if_addrs::Interface>>>> =
+    Lazy::new(|| Arc::new(StdMutex::new(Vec::new())));
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
+static NETWORK_PATH_MONITOR_START: std::sync::Once = std::sync::Once::new();
+
+#[derive(Debug)]
+/// Struct to monitor network
+pub struct NetworkMonitor {
+    nw_path_monitor_monitor_handle: Option<JoinHandle<io::Result<()>>>,
```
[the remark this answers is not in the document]
Hasan6979: This ins't for macos hence no directive but it is for the thread that waits for notifications from network monitors to update the local IF cache. Naming can be better.

nat-lab/tests/test_network_monitor.py
Hasan6979: Intentional yes for triggers from OS to occur. `wait_for_log()` with a timeout can also be done.

crates/telio-network-monitors/src/monitor.rs
lcruz99: then you should replace or prefix it with '_'

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Method that returns vector of Interfaces on the system
+/// Filtering out meshnet and loopback IP
+pub fn gather_local_interfaces(
+    get_if_addr: GetIfAddrs,
+) -> std::io::Result<Vec<if_addrs::Interface>> {
+    let shared_range: Ipv4Net = Ipv4Net::new(Ipv4Addr::new(100, 64, 0, 0), 10).unwrap_or_default();
+    Ok(get_if_addr()?
+        // .get()?
```
tomasz-grz, line 19: Commented out code

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,91 @@
...
+        .into_iter()
+        .filter(|x| !x.addr.is_loopback())
+        .filter(|x| match x.addr.ip() {
```
tomasz-grz, line 22: What is the benefit of having two separate `filter()`?

crates/telio-network-monitors/src/apple.rs
```diff
@@ -0,0 +1,118 @@
...
+///    for handling asynchronous events.
+///
+/// 3. **Global Queue Setup**: The monitor runs on a background queue using Grand Central Dispatch (GCD).
+///    This allows it to handle updates without blocking the main thread.
+///
+/// 4. **Broadcast Notification**: After the interfaces are updated, a notification is sent via a broadcast channel
+///    (`PATH_CHANGE_BROADCAST`) to inform other parts of the system that the path has changed.
+///
+/// For more details on Apple's `Network.framework` see:
+/// - [Apple Network.framework Documentation](https://developer.apple.com/documentation/network)
```
tomasz-grz, line 58: Nice! 🙌

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }));
+
+        // Only add in production, otherwise network triggers from OS can occur
+        // which can make the test (test_path_change_notification) flaky sometimes
+        #[cfg(all(
+            not(test),
+            any(target_os = "macos", target_os = "ios", target_os = "tvos")
+        ))]
+        NETWORK_PATH_MONITOR_START.call_once(crate::apple::setup_network_monitor);
```
tomasz-grz, line 68: nit: since this is Apple specific, I would personally call it `NW_PATH_MONITOR_START`

crates/telio-network-monitors/src/apple.rs
```diff
@@ -0,0 +1,118 @@
...
+pub const DISPATCH_QUEUE_PRIORITY_HIGH: c_long = 2;
+/// Dispatch queue priority as default queue
+pub const DISPATCH_QUEUE_PRIORITY_DEFAULT: c_long = 0;
+/// Dispatch queue priority as low priority queue
+pub const DISPATCH_QUEUE_PRIORITY_LOW: c_long = -2;
+/// Dispatch queue priority as background queue
+pub const DISPATCH_QUEUE_PRIORITY_BACKGROUND: c_long = -1 << 15;
+
+/// Vector containing interface names in OS order
+pub static INTERFACE_NAMES_IN_OS_PREFERENCE_ORDER: Mutex<Vec<String>> = Mutex::new(Vec::new());
```
tomasz-grz, line 27: Is this similar to `pub static LOCAL_ADDRS_CACHE: Lazy<Arc<StdMutex<Vec<if_addrs::Interface>>>>` from `crates/telio-network-monitors/src/monitor.rs`, if so why do we want to keep two copies of the same information in memory?

crates/telio-utils/src/local_interfaces.rs
Hasan6979: Leftover commented out code 😅

crates/telio-utils/src/local_interfaces.rs
Hasan6979: None 😛

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        #[cfg(all(not(test), target_os = "windows"))]
+        let monitor_handle = crate::windows::setup_network_monitor();
+
+        Ok(Self {
+            if_cache_updater_handle,
+            #[cfg(all(not(test), target_os = "linux"))]
+            monitor_handle,
+            #[cfg(all(not(test), target_os = "windows"))]
+            monitor_handle,
```
lcruz99, line 76:
```suggestion
            #[cfg(all(not(test), any(target_os = "linux", target_os = "windows")))]
```

lcruz99, line 76:
```suggestion
            #[cfg(all(not(test), any(target_os = "linux", target_os = "windows")))]
            monitor_handle,
```

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }));
+
+        // Only add in production, otherwise network triggers from OS can occur
+        // which can make the test (test_path_change_notification) flaky sometimes
+        #[cfg(all(
+            not(test),
+            any(target_os = "macos", target_os = "ios", target_os = "tvos")
+        ))]
+        NW_PATH_MONITOR_START.call_once(crate::apple::setup_network_monitor);
```
lcruz99, line 64: Out of curiosity, why not use `Once` also for Linux/Windows?

Hasan6979: It is not needed. I am not sure about Apple as well. I had just moved the Apple monitor. Earlier it was initiated from  `start_device`. Maybe it is just safer to add for Linux/Windows as well.

tomasz-grz (approved):
+1.0

crates/telio-utils/src/local_interfaces.rs
```diff
@@ -0,0 +1,93 @@
...
+pub type GetIfAddrs = fn() -> std::io::Result<Vec<if_addrs::Interface>>;
+
+/// Function returning interface addresses from the system
+pub fn system_get_if_addr() -> std::io::Result<Vec<if_addrs::Interface>> {
+    if_addrs::get_if_addrs()
+}
+
+/// Method that returns vector of Interfaces on the system
+/// Filtering out meshnet and loopback IP
+pub fn gather_local_interfaces(
```
lcruz99, line 14: IMO this should be part of the `telio-network-monitor` crate, no need to have it isolated here in `telio-utils`

lcruz99: eg: `NetworkMonitor::gather_local_interfaces()`

Hasan6979: Yep. That makes better sense!

crates/telio-network-monitors/src/monitor.rs
Hasan6979: Problem with adding `Once` for Linux/Windows is that it requires a return parameter. It is possible to return the monitor handle through some magic, but nothing straight forward so letting this be as is.

crates/telio-network-monitors/src/local_interfaces.rs
```diff
@@ -0,0 +1,93 @@
...
+pub type GetIfAddrs = fn() -> std::io::Result<Vec<if_addrs::Interface>>;
+
+/// Function returning interface addresses from the system
+pub fn system_get_if_addr() -> std::io::Result<Vec<if_addrs::Interface>> {
+    if_addrs::get_if_addrs()
+}
+
+/// Method that returns vector of Interfaces on the system
+/// Filtering out meshnet and loopback IP
+pub fn gather_local_interfaces(
```
lcruz99, line 14: This fn can be part of the `monitor` module. Then you can use traits and define a default behaviour where `if_addrs::get_if_addrs()` is used, so that this crate users don't have to worry about which get_if_addrs fn implementation will be used. And then on tests you can mock and provide your custom implementation with the return values you need..

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+        #[cfg(all(not(test), target_os = "windows"))]
+        if !self.monitor_handle.0.is_null() {
+            crate::windows::deregister_network_monitor(self.monitor_handle.clone());
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lcruz99, line 95: Please add more tests, for the init/deinit as we're providing custom constructor/destructor and specially for the `NetworkMonitor::if_cache_updated_handle` task:

```
loop {
    match notify.recv().await {
        Ok(()) => match local_interfaces::gather_local_interfaces(if_addr) {
            Ok(v) => {
                telio_log_debug!("Updating local addr cache");
                *(LOCAL_ADDRS_CACHE.lock()) = v;
            }
            Err(e) => telio_log_warn!("Unable to get local interfaces {e}"),
        },
        Err(e) => {
            telio_log_warn!("Failed to receive new path for sockets ({e}): ");
        }
    }
}
```

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,174 @@
...
+            #[cfg(windows)]
+            adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+        }])
+    }
+
+    #[tokio::test]
+    async fn test_path_change_notification() {
+        let _network_monitor = NetworkMonitor::new().await.unwrap();
+        if let Err(e) = PATH_CHANGE_BROADCAST.send(()) {
+            println!("Failed to notify about changed path, error: {e}");
```
lcruz99, line 127: shouldn't the test fail here?

crates/telio-network-monitors/src/local_interfaces.rs
```diff
@@ -0,0 +1,93 @@
...
+    get_if_addr: GetIfAddrs,
```
lcruz99, line 15: I think this method doesn't need to receive this fn pointer as parameter anymore right?

lcruz99: and you can use `if_addrs::get_if_addrs()` directly

Hasan6979: No we need that still for the `gather_local_interface_filtering` otherwise we can't mock it.

crates/telio-network-monitors/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Utility to get monitor network on linux
+pub mod linux;
+
+#[cfg(target_os = "windows")]
+/// Utility to get monitor network on windows
+pub mod windows;
+
+/// Get local interfaces
+pub mod local_interfaces;
+pub use local_interfaces::*;
```
lcruz99, line 20: is this `use` needed?

crates/telio-network-monitors/src/local_interfaces.rs
lcruz99: you could by simply doing:
```
#[cfg(not(test))]
use if_addrs::get_if_addrs as GetIfAddrs;
#[cfg(test)]
use tests::get_if_addrs as GetIfAddrs;
```

lcruz99: so you wouldn't need `GetIfAddrs` type or `system_get_if_addr` function.

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,262 @@
...
+            // Release the lock before spinning
+            std::thread::yield_now();
+        }
+    }
+
+    fn increment_order() {
+        let mut lock = TEST_MUTEX.lock().unwrap();
+        *lock = (*lock).next();
+        CALL_COUNT.store(0, Ordering::Relaxed);
+    }
```
lcruz99, line 174: IMHO this is overengineered..

Locking a single mutex at the beginning of each test is enough:
```
static TESTS_SYNC_MTX: Mutex<()> = Mutex::new(());
    #[tokio::test()]
    async fn test_path_change_notification() {
        let _m = TESTS_SYNC_MTX.lock();
        CALL_COUNT.store(0, Ordering::Relaxed);
       // ...
    }
    #[tokio::test()]
    async fn test_multiple_monitor_instances() {
       let _m = TESTS_SYNC_MTX.lock();
       CALL_COUNT.store(0, Ordering::Relaxed);
       // ...
    }
    ....
```

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,262 @@
...
+                            telio_log_warn!("Failed to receive new path for sockets ({e}): ");
+                        }
+                    }
+                }
+            }
+        }));
+
+        // Only add in production, otherwise network triggers from OS can occur
+        // which can make the test (test_path_change_notification) flaky sometimes
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
```
lcruz99, line 64: #[cfg(all(test, ...)] missing ?

crates/telio-network-monitors/src/monitor.rs
Hasan6979: The main problem comes CALL_ONCE on apple monitor, and that BROADCAST_SEND() callback is issued as soon as handler is subscribed which leads to 1 more call than needed. I would like to keep the CALL_ONCE on Apple Monitor. Cancel call requires for me to get the monitor handle, which again complicates things alot because of CALL_ONCE. Looking at it again, it can be simplified, but a simple lock mechanism won't work.

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -107,105 +111,54 @@ mod tests {
...
+                netmask: Ipv4Addr::new(255, 255, 255, 0),
+                broadcast: None,
+            }),
+            index: None,
+            #[cfg(windows)]
+            adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+        }])
+        // } else {
+        //     Err(std::io::ErrorKind::NotFound.into())
+        // }
```
lcruz99, line 136: leftover?

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -107,105 +111,54 @@ mod tests {
...
         #[cfg(not(target_os = "linux"))]
-        assert_eq!(CALL_COUNT.load(Ordering::SeqCst), 3);
+        assert_eq!(CALL_COUNT.load(Ordering::SeqCst), 2);
 
         #[cfg(target_os = "linux")]
         assert_eq!(CALL_COUNT.load(Ordering::SeqCst), 2);
 
-        // Increment test order
-        increment_order();
+        CALL_COUNT.store(0, Ordering::Relaxed);
```
lcruz99, line 155: only needed at the beginning of each test?

crates/telio-network-monitors/src/apple.rs
```diff
@@ -0,0 +1,80 @@
...
+///
+/// 4. **Broadcast Notification**: After the interfaces are updated, a notification is sent via a broadcast channel
+///    (`PATH_CHANGE_BROADCAST`) to inform other parts of the system that the path has changed.
+///
+/// For more details on Apple's `Network.framework` see:
+/// - [Apple Network.framework Documentation](https://developer.apple.com/documentation/network)
+pub fn setup_network_monitor() {
+    let update_handler = block::ConcreteBlock::new(|_path: nw_path_t| {
+        if let Err(e) = PATH_CHANGE_BROADCAST.send(()) {
+            telio_log_warn!("Failed to notify about changed path, error: {e}");
```
lcruz99, line 54: the test should fail, right?

crates/telio-network-monitors/src/lib.rs
Hasan6979: Yes as local_interfaces is exported out to be used by `local.rs`. This module is as a monitor util.

crates/telio-network-monitors/src/local_interfaces.rs
```diff
@@ -0,0 +1,104 @@
+use ipnet::Ipv4Net;
+use mockall::automock;
+use std::net::{IpAddr, Ipv4Addr};
+
+/// #[cfg_attr(test, automock)]
```
lcruz99, line 5: leftover

crates/telio-network-monitors/src/local_interfaces.rs
```diff
@@ -0,0 +1,104 @@
...
+#[automock]
+pub trait GetIfAddrs: Send + Sync + Default + 'static {
+    /// My old friend
+    fn get(&self) -> std::io::Result<Vec<if_addrs::Interface>>;
+}
+
+/// Hello darkness
```
lcruz99, line 12: leftover

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,187 @@
...
+    not(test),
+    any(target_os = "macos", target_os = "ios", target_os = "tvos")
+))]
+static NW_PATH_MONITOR_START: std::sync::Once = std::sync::Once::new();
+
+#[derive(Debug)]
+/// Struct to monitor network
+pub struct NetworkMonitor {
+    if_cache_updater_handle: Option<JoinHandle<io::Result<()>>>,
+    // get_if_addrs: G,
```
lcruz99, line 24: leftover

nat-lab/tests/test_network_monitor.py
```diff
@@ -0,0 +1,49 @@
...
+    ],
+)
+async def test_network_monitor(
+    alpha_setup_params: SetupParameters,
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        env = await setup_mesh_nodes(exit_stack, [alpha_setup_params])
+        [client_alpha] = env.clients
+
+        time.sleep(DEFAULT_WAITING_TIME)
```
lcruz99, line 46: async sleep perhaps?

lcruz99: why do we have to wait? try to avoid time dependency on tests, it increases flakyness probability

crates/telio-network-monitors/src/windows.rs
```diff
@@ -0,0 +1,94 @@
...
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_unregister_callback() {
+        let handle = setup_network_monitor();
+
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+
+        tokio::spawn({
```
lcruz99, line 86: You're dropping this JoinHandle without waiting

crates/telio-network-monitors/src/windows.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        SafeHandle(handle)
+    }
+}
+
+/// Clean up the network notification
+pub fn deregister_network_monitor(safe_handle: SafeHandle) {
+    // Denotification has to be done in a separate thread
+    // otherwise would lead to deadlock
+    tokio::spawn({
```
lcruz99, line 59: You're dropping this JoinHandle without waiting

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -0,0 +1,185 @@
...
+                index: None,
+                #[cfg(windows)]
+                adapter_name: "{78f73923-a518-4936-ba87-2a30427b1f63}".to_string(),
+            }])
+        });
+        NetworkMonitor::new(get_if_addrs_mock).await.unwrap()
+    }
+
+    #[tokio::test]
+    #[serial]
```
lcruz99, line 132: you can discard the serial crate and make them run serially with a mutex:
```suggestion
static TESTS_SYNC_MTX: Mutex<()> = Mutex::new(());

    #[tokio::test]
    async fn test_path_change_notification() {
        let _mut = TESTS_SYNC_MTX.lock();
```

crates/telio-network-monitors/src/lib.rs
lcruz99: why don't you import what you need from `local.rs`?

crates/telio-network-monitors/src/monitor.rs
lcruz99: (optionally)

PR #602, opened 2024-06-06
Don't compile tclid by default when building tcli package

tomaszklak:
### Problem
When building `tcli` package, cargo will build 2 executables - `tcli` and `tclid`. This has 2 downsides when building for nat-lab. First, it slows down the whole build process. Secnod, on a macOS host where building happens in a docker with a limited memory, concurrent linking processes consume whole memory and one of them is being killed by oom-killer.

### Solution
Make the `tclid` binary optional. To build it the required command will be:
```
cargo build --package tcli --features tclid
```

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5247
+++ b/.unreleased/LLT-5247
@@ -1,0 +1,1 @@
+Don't compile tclid by default when building tcli package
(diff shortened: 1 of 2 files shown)


LukasPukenis (approved):
+1

PR #603, opened 2024-06-06
LLT-5252: Clean unneeded words

lcruz99:
Gitlab URL's updated so these variables have to be updated.

- [x] README.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.1.12
+          triggered-ref: LLT-5252_update_gitlab_url
--- a/.unreleased/LLT-5252
+++ b/.unreleased/LLT-5252
@@ -1,0 +1,1 @@
+ci: Update gitlab urls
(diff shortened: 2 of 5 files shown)


.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.1.12
+          triggered-ref: LLT-5252_update_gitlab_url
```
stalowyjez, line 34: I guess you will update it to `v1.1.13` after merging?

.unreleased/LLT-5252
```diff
@@ -0,0 +1 @@
+ci: Update gitlab urls
```
stalowyjez, line 1: I guess when it is not a PR introducing sth used by apps this file should be empty

stalowyjez (commented):
I find the name of this PR a bit misleading - I think sth like "fix moose path" would be nicer

.github/workflows/gitlab.yml
lcruz99: correct, thanks for leaving a reminder :)

.unreleased/LLT-5252
lcruz99: How can I tell that to `check-changelog` job? 👀

PR #607, opened 2024-06-11
Add firewall feature config

Hasan6979:
This PR creates a new feature for firewall. This answers the PR review comment https://github.com/NordSecurity/libtelio/pull/529/files#r1626552500

Above mentioned PR is already too large, so creating a separate PR to have `boringtun_reset_connection` as part of `FeatureFirewall` config. 

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5278
+++ b/.unreleased/LLT-5278
@@ -1,0 +1,1 @@
+Make separate feature config for firewall
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -369,10 +369,13 @@ impl FeatureEndpointProvidersOptimization {
     }
 }
 
-/// Turns on connection resets upon VPN server change
+/// Feature config for firewall
 #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, Deserialize)]
-#[serde(transparent)]
-pub struct FeatureBoringtunResetConns(pub bool);
+pub struct FeatureFirewall {
+    /// Turns on connection resets upon VPN server change
+    #[serde(default)]
+    pub boringtun_reset_conns: bool,
+}
 
 /// Turns on post quantum VPN tunnel
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize)]
@@ -520,7 +523,7 @@ pub struct Features {
     pub nicknames: bool,
     /// Flag to turn on connection reset upon VPN server change for boringtun adapter
     #[serde(default)]
-    pub boringtun_reset_connections: FeatureBoringtunResetConns,
+    pub firewall: FeatureFirewall,
     /// If and for how long to flush events when stopping telio. Setting to Some(0) means waiting until all events have been flushed, regardless of how long it takes
     pub flush_events_on_stop_timeout_seconds: Option<u64>,
     /// Post quantum VPN tunnel configuration
@@ -639,7 +642,7 @@ mod tests {
             "validate_keys": false,
             "ipv6": true,
             "nicknames": true,
-            "boringtun_reset_connections": true,
+            "firewall": {"boringtun_reset_conns": true},
             "post_quantum_vpn":
             {
                 "handshake_retry_interval_s": 16,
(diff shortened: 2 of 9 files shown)


crates/telio-model/src/features.rs
```diff
@@ -369,10 +369,13 @@ impl FeatureEndpointProvidersOptimization {
...
-/// Turns on connection resets upon VPN server change
+/// Feature config for firewall
 #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, Deserialize)]
-#[serde(transparent)]
-pub struct FeatureBoringtunResetConns(pub bool);
+pub struct FeatureFirewall {
+    /// Turns on connection resets upon VPN server change
+    #[serde(default)]
+    pub boringtun_reset_conns: bool,
+}
```
tomaszklak, line 378: This is a breaking change - is this indented? Maybe it would be better to avoid it, and use something like `[serde(flatten)]` to keep it compatible.

LukasPukenis (commented):
The code changes look good. However I have additional thoughts regarding moving just one flag to a separate config option and breaking the API as a result. 

I've reviewed te linked comment so it seems it's gonna have more than one flag, in that case it's okay. 

+1

crates/telio-model/src/features.rs
mathiaspeters: I think the breaking change is okay since it would go out with a new major version. If it makes more sense to avoid the breakage then let's go with that, but since the next release is 5.0, we don't have to worry much about breaking changes, especially not in feature and meshnet configs since apps will most likely use the `deserialize_*_config' functions. So as long as we can parse both the new and the old way of doing this, it should be fine

tomaszklak: Good point, although I still think that breaking compatibility when a fix is easy should never be done. So feel free to treat this as a nitpick and ignore it if you see it differently :)

tomaszklak (approved):
+1

crates/telio-model/src/features.rs
mathiaspeters: I highly agree that we shouldn't break compat if it's not necessary and I'm not really weighing in on this specific case, just putting it out there that if we need to break compat, now is a great time :smile:

LukasPukenis (approved):

crates/telio-model/src/features.rs
Jauler: In this case, the breaking change is intentional, as this PR has been specifically created to include this change in major libtelio release. 

More firewall configuration options in the FeatureConfig are coming soon :tm:, therefore we wanted to group them into single group. 

But as you correctly identified - this is a breaking change, and this is the reason for separating this PR from actual firewall changes, as we would like to introduce changes with libtelio v5.0 release, which is already breaking change.

Jauler:
> The code changes look good. However I have additional thoughts regarding moving just one flag to a separate config option and breaking the API as a result.
> 
> I've reviewed te linked comment so it seems it's gonna have more than one flag, in that case it's okay.
> 
> +1

For more info, see comment here: https://github.com/NordSecurity/libtelio/pull/607#discussion_r1637813710

Jauler (dismissed):
+1.0

.unreleased/LLT-5278
```diff
@@ -0,0 +1 @@
+Make separate config for firewall
```
mathiaspeters, line 1: Nit: config is ambiguous, better to specify feature config

crates/telio-model/src/features.rs
```diff
@@ -369,10 +369,13 @@ impl FeatureEndpointProvidersOptimization {
     }
 }
 
-/// Turns on connection resets upon VPN server change
+/// Feature config for firewall
 #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, Deserialize)]
-#[serde(transparent)]
-pub struct FeatureBoringtunResetConns(pub bool);
+pub struct FeatureFirewall {
```
mathiaspeters, line 374: I started adding tests in a different way in here for multicast. What do you think of creating similar tests for this new firewall feature config block?

mathiaspeters:
Could you add a firewall config check to `test_telio_features.py` as well?

crates/telio-model/src/features.rs
Hasan6979: Can you elaborate more on different way ?

mathiaspeters: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-model/src/features.rs#L1255 basically a module for `FeatureFirewall` that checks that it is parsed correctly

mathiaspeters (approved):
+1

mathiaspeters (approved):

PR #610, opened 2024-06-11
[LLT-4928] Mask IPs with regex

stalowyjez:
### Problem
External IP addresses are visible in the logs

### Solution
We replace the IP addresses in the logs with some hashes, so they are not visible anymore

- [x] README.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ crate-type = ["cdylib", "lib"]
 pretend_to_be_macos = ["telio-model/pretend_to_be_macos"]
 
 [dependencies]
+blake3 = "1.5.1"
 cfg-if = "1.0.0"
 ffi_helpers = "0.3.0"
 h2 = "=0.3.26" # RUSTSEC-2024-0332
@@ -28,6 +29,7 @@ libc.workspace = true
 modifier.workspace = true
 num_cpus.workspace = true
 parking_lot.workspace = true
+regex.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 rand.workspace = true
@@ -65,6 +67,7 @@ rustls-platform-verifier = { git = "https://github.com/tomaszklak/rustls-platfor
 
 [target.'cfg(windows)'.dependencies]
 winapi = { workspace = true, features = ["ntdef", "winerror"] }
+
 
 [dev-dependencies]
 maplit.workspace = true
@@ -136,6 +139,7 @@ proptest = "1.2.0"
 proptest-derive = "0.3.0"
 protobuf-codegen-pure = "2"
 rand = "0.8"
+regex = "1.5.5"
 rstest = "0.11.0"
 rupnp = { version = "1.1.0", default-features = false }
 rustc-hash = "1"
--- a/.unreleased/LLT-4928
+++ b/.unreleased/LLT-4928
@@ -1,0 +1,1 @@
+Hide IP addresses in release version logs by replacing them with some hashes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,6 +88,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
 
 [[package]]
+name = "arrayref"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
 name = "askama"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -418,6 +430,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "46502ad458c9a52b69d4d4d32775c788b7a1b85e8bc9d482d92250fc0e3f8efe"
 dependencies = [
  "digest",
+]
+
+[[package]]
+name = "blake3"
+version = "1.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30cca6d3674597c30ddf2c587bf8d9d65c9a84d2326d941cc79c9842dfe0ef52"
+dependencies = [
+ "arrayref",
+ "arrayvec",
+ "cc",
+ "cfg-if",
+ "constant_time_eq",
 ]
 
 [[package]]
@@ -771,6 +796,12 @@ checksum = "4ca0197aee26d1ae37445ee532fefce43251d24cc7c166799f4d46817f1d3973"
 dependencies = [
  "crossbeam-utils",
 ]
+
+[[package]]
+name = "constant_time_eq"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7144d30dcf0fafbce74250a3963025d8d52177934239851c917d29f1df280c2"
 
 [[package]]
 name = "core-foundation"
@@ -4223,6 +4254,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "base64 0.13.1",
+ "blake3",
  "cc",
  "cfg-if",
  "crypto_box",
@@ -4243,6 +4275,7 @@ dependencies = [
  "parking_lot",
  "pretty_assertions",
  "rand",
+ "regex",
  "rstest",
  "rustls-platform-verifier",
  "serde",
--- a/clis/tcli/Cargo.toml
+++ b/clis/tcli/Cargo.toml
@@ -11,7 +11,6 @@ tclid = ["sysinfo", "interprocess", "daemonize"]
 
 [dependencies]
 dirs = "4.0.0"
-regex = "1.5.5"
 reqwest = { version = "0.11.16", default-features = false, features = [
     "json",
     "blocking",
@@ -38,6 +37,7 @@ tracing-subscriber.workspace = true
 tracing-appender.workspace = true
 parking_lot.workspace = true
 rand = { workspace = true, features = ["std", "std_rng"] }
+regex.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
--- a/src/ffi/logging.rs
+++ b/src/ffi/logging.rs
@@ -1,8 +1,12 @@
 use std::{
     io::{self, ErrorKind},
-    str::from_utf8,
+    net::IpAddr,
+    str::{from_utf8, FromStr},
     sync::{Arc, Mutex},
 };
+
+use rand::Rng;
+use regex::{Captures, Error as RegexError, Regex, RegexBuilder};
 
 use tracing::{level_filters::LevelFilter, Subscriber};
 use tracing_subscriber::{
@@ -57,11 +61,15 @@ where
 
 pub struct FfiCallback {
     callback: Arc<dyn TelioLoggerCb>,
+    #[cfg(not(debug_assertions))]
+    censor: Option<LogCensor>,
 }
 impl FfiCallback {
     fn new(logger: Box<dyn TelioLoggerCb>) -> Self {
         Self {
             callback: logger.into(),
+            #[cfg(not(debug_assertions))]
+            censor: LogCensor::new().ok(),
         }
     }
 }
@@ -77,6 +85,8 @@ impl<'a> MakeWriter<'a> for FfiCallback {
         FfiCallbackWriter {
             cb: self.callback.clone(),
             level: (*meta.level()).into(),
+            #[cfg(not(debug_assertions))]
+            censor: self.censor.clone(),
         }
     }
 }
@@ -84,6 +94,8 @@ impl<'a> MakeWriter<'a> for FfiCallback {
 pub struct FfiCallbackWriter {
     cb: Arc<dyn TelioLoggerCb>,
     level: TelioLogLevel,
+    #[cfg(not(debug_assertions))]
+    censor: Option<LogCensor>,
 }
 
 impl io::Write for FfiCallbackWriter {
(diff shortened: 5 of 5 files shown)


src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+        )
+        .ok()
+        .map(|re| LogCensor {
+            ip_masks: parking_lot::Mutex::new(HashMap::new()),
+            ip_mask_seed: rand::thread_rng().gen::<[u8; 32]>(),
+            ip_regex: re,
+        })
+    }
+
+    fn censor_logs(&self, log: String) -> String {
```
tomaszklak, line 884: This method should do nothing in debug mode (or there should be no instance of `LogCensor`).

src/ffi/mod.rs
```diff
@@ -929,7 +1012,17 @@ impl Subscriber for TelioTracingSubscriber {
...
+            censor.censor_logs(visitor.message)
+        } else {
+            visitor.message
+        };
+
+        #[cfg(not(feature = "hide_ips"))]
+        let message = visitor.message;
+
+        if let Some(filtered_msg) = filter_log_message(message) {
             let _ = self.callback.log(level.into(), filtered_msg);
```
tomaszklak, line 1026: I think that modification of logs would be best done here - to avoid needlessly modifying logs that will not be printed.

Cargo.toml
```diff
@@ -28,8 +30,10 @@ lazy_static.workspace = true
 libc.workspace = true
 modifier.workspace = true
 parking_lot.workspace = true
+regex = { workspace = true, optional = true }
 serde.workspace = true
 serde_json.workspace = true
+sha2.workspace = true
```
tomaszklak, line 36: can `sha2` also become optional and only enabled when `hide_ips` is turned on?

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+    ip_mask_seed: [u8; 32],
+    ip_masks: parking_lot::Mutex<HashMap<Ipv4Addr, String>>,
+    ip_regex: Regex,
+}
+
+#[cfg(feature = "hide_ips")]
+impl LogCensor {
+    fn new() -> Option<Self> {
+        Regex::new(
+            r"(?x)\b
```
tomaszklak, line 858: nit: I think that using `RegexBuild` + `ignore_whitespace(true)` is more clear.

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+                #[allow(index_access_check)]
+                let ip_match = &captures[0];
+                Ipv4Addr::from_str(ip_match)
+                    .map(|ip_addr| {
+                        self.ip_masks
+                            .lock()
+                            .entry(ip_addr)
+                            .or_insert_with(|| {
+                                // Probably good enough protection for this usecase
+                                let mut hasher = Sha256::new();
```
tomaszklak, line 898: **IF** we are using cache (which I don't think is a good idea) and compute a replacement string only once for a given sequence, than we don't need to use any deterministic function and we could just generated random replacement string.

stalowyjez: I'll get rid of the hashmap

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+                    )
+                    \.
+                ){3}
+                (?:
+                    25[0-5]
+                    |  2[0-4][0-9]
+                    |  1[0-9]{2}
+                    |  [1-9]?[0-9]
+                )
+                \b",
```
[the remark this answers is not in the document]
stalowyjez: Aren't we using IPv6 only internally, so there would be no need to replace them?

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
     fn fortify_source();
 }
 
+#[cfg(feature = "hide_ips")]
+struct LogCensor {
+    ip_mask_seed: [u8; 32],
+    ip_masks: parking_lot::Mutex<HashMap<Ipv4Addr, String>>,
```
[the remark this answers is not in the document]
stalowyjez: I thought that we are going to have a lot of repetitions, so keeping these IP replacements in some hashmap is only natural, but yeah, when I think about it now, keeping hashes in hashmap is riddiculous

src/ffi/mod.rs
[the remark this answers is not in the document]
tomaszklak: Hm, you might be right, but the task doesn't clarify this. I think it would be safer (and future proof) to also handle ipv6 but that is indeed more complexity. @Jauler - could you clarify this? :)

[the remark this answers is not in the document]
packgron: I think it's the same precedence, also we might support ipv6 connections at some point, or if something is comming into tunnel, something like firewall, nat or any other component could capture ipv6 that are not controlled by us.

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+                            .lock()
+                            .entry(ip_addr)
+                            .or_insert_with(|| {
+                                // Probably good enough protection for this usecase
+                                let mut hasher = Sha256::new();
+                                hasher.update(ip_addr.octets());
+                                hasher.update(self.ip_mask_seed);
+                                let mut hashed_ip = format!("{:x}", hasher.finalize());
+                                hashed_ip.truncate(16);
+                                hashed_ip
```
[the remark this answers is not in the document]
stalowyjez: Maybe? I'm not sure about this - `ahash` is explicitely described as not secure, while SHA256 is explicitly designed for security and it is also pretty fast.

src/ffi/mod.rs
```diff
@@ -11,6 +11,17 @@ use tracing::{error, trace, Subscriber};
 use telio_sockets::protector::make_external_protector;
 use uuid::Uuid;
 
+#[cfg(feature = "hide_ips")]
```
packgron, line 14: [cfg_if](https://docs.rs/cfg-if/latest/cfg_if/) is a nice way to make this a bit more clean

src/ffi/mod.rs
```diff
@@ -80,6 +91,8 @@ pub fn set_global_logger(log_level: TelioLogLevel, logger: Box<dyn TelioLoggerCb
     let tracing_subscriber = TelioTracingSubscriber {
         callback: logger,
         max_level: log_level.into(),
+        #[cfg(feature = "hide_ips")]
+        censor: LogCensor::new(),
```
packgron, line 95: To make it more nicely composable, recomended way would be to make cencor a layer you optionaly add https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/index.html.

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+    ip_masks: parking_lot::Mutex<HashMap<Ipv4Addr, String>>,
+    ip_regex: Regex,
+}
+
+#[cfg(feature = "hide_ips")]
+impl LogCensor {
+    fn new() -> Option<Self> {
+        Regex::new(
+            r"(?x)\b
+                (?:
```
packgron, line 859: This additional complexity exsists, to avoid capturing local ips ?

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
...
+                // First capture is guaranteed to be present by regex crate
+                #[allow(index_access_check)]
+                let ip_match = &captures[0];
+                Ipv4Addr::from_str(ip_match)
+                    .map(|ip_addr| {
+                        self.ip_masks
+                            .lock()
+                            .entry(ip_addr)
+                            .or_insert_with(|| {
+                                // Probably good enough protection for this usecase
```
packgron, line 897: So we have already had a place there in anlytics we are hasing server ip's with md5, i feel it would be nice to keep this consistency.

packgron: https://github.com/NordSecurity/libtelio/blob/main/crates/telio-nurse/src/heartbeat.rs#L518

Cargo.toml
stalowyjez: Hmm, `hide_ips` was a replacement for having it enabled in debug mode only - because having it enabled in debug only seems to be better solution than that, and it seems that there is no way (i.e. I couldn't find one) to use a dependency only in release mode, so I guess both of them need to be compiled in all of the cases (As debug builds are not that optimized, it might not be a big issue)

src/ffi/mod.rs
stalowyjez: No, to capture just IPs ;)

src/ffi/mod.rs
stalowyjez: I saw it, but don't really see the benefits - what are the gains from that?

src/ffi/mod.rs
stalowyjez: `LogCensor` doesn't exist in debug mode now

src/ffi/mod.rs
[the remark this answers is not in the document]
stalowyjez: Hmm, I'm not sure about this TBH - should I also replace all of the `::` with some hashes?

src/ffi/mod.rs
```diff
@@ -831,6 +844,72 @@ extern "C" {
     fn fortify_source();
 }
 
+#[cfg(feature = "hide_ips")]
+struct LogCensor {
```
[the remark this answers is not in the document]
stalowyjez: Added some test - do you think is it enough to test it in unit tests, or should it have some nat-lab tests as well?

src/ffi/mod.rs
stalowyjez: Now it exists, but there is no instance of it

Cargo.toml
stalowyjez: Got rid of this feature

src/ffi/mod.rs
stalowyjez: Changed the hash algorithm to md5 to be more compliant, but left some additional noise to make it harder to reverse - is it ok @packgron

src/ffi/mod.rs
packgron: So primary benefit that it would make it more composable, but i guess we can leave this for then we want to reuse this at some point.

packgron: For examples if we would want to test it out with tcli, but leaving for you to decide, if you considered it and did not find it usefull, leaving it to your decision.

src/ffi/mod.rs
packgron: I feel that for ipv4 it's never to to simply brute force all ip's. I think what you did is more than enough.

src/ffi/mod.rs
```diff
@@ -838,6 +843,83 @@ extern "C" {
...
+                            data_to_hash.extend_from_slice(ipv4.octets().as_slice())
+                        }
+                        IpAddr::V6(ipv6) => {
+                            data_to_hash.extend_from_slice(ipv6.octets().as_slice())
+                        }
+                    };
+                    data_to_hash.extend_from_slice(&self.ip_mask_seed);
+                    // MD5 hash is much bigger than 8 bytes
+                    #[allow(index_access_check)]
+                    let prefix = &md5::compute(data_to_hash).0[..8];
```
tomaszklak, line 915: Switching from sha256 to md5 is not speeding this up by much. When you compare sha256 and md5 on ip like inputs md5 is *only* 60% faster. Ahash is more than 40 **times* as fast as md5 and 70 **times* as fast as sha256:

```
ahash                   time:   [2.3799 µs 2.3920 µs 2.4061 µs]
md5                     time:   [106.22 µs 106.90 µs 107.69 µs]
sha256                  time:   [170.74 µs 171.77 µs 172.99 µs]

str_ahash               time:   [3.0648 µs 3.0818 µs 3.1020 µs]
str_md5                 time:   [108.12 µs 108.67 µs 109.25 µs]
str_sha256              time:   [175.05 µs 176.11 µs 177.31 µs]
```

See [here](https://github.com/tomaszklak/hash-perf) for details.

stalowyjez: @tomaszklak yeah, I did it with md5 because @packgron wanted that to be more consistent with hashing it lana, not to make it faster

tomaszklak: I see, in that case **if** we want to keep it as it is I think a comment explaining why md5 was chosen here would be a good idea.

On the other hand I'm not convinced that the lana precedent should be a decisive here. Since it's not using any salting, it's pretty useless (in terms of security/hiding them) for ipv4 addresses.

stalowyjez: Ok, so we are not using md5 - so I guess I'll make a SHA version for now. `ahash` is quick, but it feels a bit wrong to me to use for kinda security purposes hash which is meant to be used for hashmaps _exclusively_ (https://github.com/tkaitchuck/aHash?tab=readme-ov-file#ahash-------).

tomaszklak: I still think that using cryptographically secure hash function here is an overkill. If we have salt and truncate the output I don't see how the result could be reversed.

src/ffi/mod.rs
```diff
@@ -838,6 +843,98 @@ extern "C" {
...
+                                IpAddr::V4(ipv4) => {
+                                    data_to_hash.extend_from_slice(ipv4.octets().as_slice())
+                                }
+                                IpAddr::V6(ipv6) => {
+                                    data_to_hash.extend_from_slice(ipv6.octets().as_slice())
+                                }
+                            };
+                            data_to_hash.extend_from_slice(&self.ip_mask_seed);
+
+                            // SHA256 hash is much bigger than 8 bytes
```
tomaszklak, line 922: This comment is out of date

tomaszklak (commented):
+1

src/ffi/mod.rs
```diff
@@ -838,6 +843,98 @@ extern "C" {
...
+                            .flat_map(|pos| log.chars().nth(*pos))
+                            .any(|c| c.is_alphanumeric() || c == '_' || c == '.');
+                    if incorret_chars_on_bounds {
+                        return ip_match.as_str().to_string();
+                    }
+
+                    IpAddr::from_str(ip_match.as_str())
+                        .map(|ip_addr| {
+                            // Probably good enough protection for this usecase
+                            let mut data_to_hash = vec![];
```
tomaszklak, line 911: nit: this probably doesn't have to be an allocation. We know that array of size 16+32 is big enough.

packgron (changes requested):
Needs a rebase

packgron (approved):
+1.0

PR #611, opened 2024-06-11
LLT-5126 Starcast nat

jjanowsk:
Starcast needs to nat messages comming from the transport socket in
order to know to which peer the response to the multicast message
should be sent to.

There are multiple ways to implement the mapping to the port.
The full-cone approach was chosen because in case of the multicast
the source ip of the response packet will not match the destination
address of the multicast that has opened the mapping so the full-cone is
actually the only option.

In case of the mapping collision the next port number is chosen.

- [x] README.md is updated


crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,635 @@
...
+    MutablePacket,
+};
+use std::{
+    hash::{DefaultHasher, Hash, Hasher},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    time::Duration,
+};
+use telio_utils::{telio_log_warn, LruCache};
+
+trait MutableIpPacket<'a>: Sized + MutablePacket {
```
mathiaspeters, line 15: Let's move this trait and the impls for v4 and v6 to make it more general, at least to it's own file in the same crate

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,734 @@
...
+            Error::UnexpectedIpVersion
+        );
+        assert_err!(
+            nat.translate_outgoing(&mut packet),
+            Error::UnexpectedIpVersion
+        );
+    }
+
+    #[test]
+    fn shorter_then_min_ip() {
```
mathiaspeters, line 590: The smallest of nits: then -> than

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,734 @@
...
+            Error::UnexpectedTransportProtocol
+        );
+        assert_err!(
+            nat.translate_outgoing(&mut udp_v6),
+            Error::UnexpectedTransportProtocol
+        );
+    }
+
+    #[test]
+    fn shorter_then_min_udp() {
```
mathiaspeters, line 641: Same nit: then -> than

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,734 @@
...
+        let src_ip = ip_packet.get_source();
+        let dst_ip = ip_packet.get_destination();
+
+        match src_ip {
+            IpAddr::V4(_) => {
+                ip_packet.set_source(IpAddr::V4(self.vpeer_ipv4_addr));
+                ip_packet.set_checksum(ipv4::checksum(&ip_packet.to_immutable_ipv4()));
+            }
+            IpAddr::V6(_) => {
+                ip_packet.set_source(IpAddr::V6(self.vpeer_ipv6_addr));
```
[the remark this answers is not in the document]
jjanowsk: Nope, it was removed from IPv6 header

crates/telio-starcast/src/nat.rs
jjanowsk: 👍

crates/telio-starcast/src/nat.rs
jjanowsk: :+1:

crates/telio-starcast/src/nat.rs
jjanowsk: :+1:

crates/telio-starcast/fuzz/fuzz_targets/translate_outgoing.rs
```diff
@@ -0,0 +1,13 @@
...
+use libfuzzer_sys::fuzz_target;
+use telio_starcast::nat::{Nat, StarcastNat};
+
+fuzz_target!(|data: &[u8]| {
+    let mut nat = StarcastNat::new(
+        "1.2.3.4".parse().unwrap(),
+        "1:2::3:4".parse().unwrap(),
+    );
+    let mut packet = data.to_vec();
+    let _ = nat.translate_incomming(&mut packet);
```
jjanowsk, line 12: tanslate_outgoing

stalowyjez (commented):
Hmm, as I understand this is a part of multicast protocol - why is it called starcast? It means sth (https://en.wiktionary.org/wiki/starcast) but I cannot see how it is connected.

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+            return Err(Error::UnexpectedTransportProtocol);
+        }
+
+        let src_ip = ip_packet.get_source();
+        let dst_ip = ip_packet.get_destination();
+
+        match src_ip {
+            IpAddr::V4(_) => {
+                ip_packet.set_source(IpAddr::V4(self.vpeer_ipv4_addr));
+                ip_packet.set_checksum(ipv4::checksum(&ip_packet.to_immutable_ipv4()));
```
packgron, line 101: So now we do a full recalculation of the checksum, it's the easiest way with api provided, but not the most unoptimal.
It's smarter to just ajust checksum, see https://datatracker.ietf.org/doc/html/rfc1141 and https://datatracker.ietf.org/doc/html/rfc1624

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+            if current_mapping != src_addr {
+                nat_port = nat_port.wrapping_add(1);
+            } else {
+                break;
+            }
+        }
+
+        self.mappings.insert(nat_port, src_addr);
+
+        udp_packet.set_source(nat_port);
```
packgron, line 124: upd packet also has it's own checksum, so it's would be better to move checksum ajusting after all of updates, and ajust them both at once

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+            }
+        };
+
+        let mut udp_packet =
+            MutableUdpPacket::new(ip_packet.payload_mut()).ok_or(Error::PacketTooShort)?;
+        let src_port = udp_packet.get_source();
+        let src_addr = SocketAddr::new(src_ip, src_port);
+        let mut nat_port = (self.mapper)(src_addr);
+
+        while let Some(&current_mapping) = self.mappings.peek(&nat_port) {
```
packgron, line 114: This introduces a posibility of infinite loop by a malicious user, somebody could span a a lot of quick packets of multicast, and make this part loop for longer and longer periods.
Would be nice to add some kind of protection.

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+    PacketTooShort,
+    /// The packet has an unexpected IP version
+    UnexpectedIpVersion,
+    /// The packet has an unexpected transport protocol
+    UnexpectedTransportProtocol,
+    /// There is no mapping for the given port
+    MappingNotFound,
+}
+
+pub trait Nat {
```
packgron, line 27: Lets add a rustdoc comment explaining that outgoing direction is transport socket to other peer.
And packet's are incomming from tunnel / virtual peer

jjanowsk:
> Hmm, as I understand this is a part of multicast protocol - why is it called starcast? It means sth (https://en.wiktionary.org/wiki/starcast) but I cannot see how it is connected.

The name comes from the RFC. It is the name for this particular implementation of multicast inside of a meshnet (so it's not just replacement word for multicast itself)

crates/telio-starcast/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod nat;
+pub(crate) mod utils;
```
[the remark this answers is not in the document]
jjanowsk: :+1:

crates/telio-starcast/src/utils.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    fn set_checksum(&mut self, _: u16) {
+        unimplemented!()
+    }
+
+    fn to_immutable_ipv4<'p>(&'p self) -> Ipv4Packet<'p> {
+        unimplemented!()
+    }
```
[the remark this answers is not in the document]
jjanowsk: Nope, the reason is that this code is directly handling random bytes from the wire (although to be precise coming only from the hosts in the meshnet)

crates/telio-starcast/src/utils.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    fn set_destination(&mut self, addr: IpAddr) {
+        if let IpAddr::V6(ipv6addr) = addr {
+            self.set_destination(ipv6addr);
+        }
+    }
+
+    fn set_source(&mut self, addr: IpAddr) {
+        if let IpAddr::V6(ipv4addr) = addr {
```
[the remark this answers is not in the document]
jjanowsk: typo

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+    /// Create new starcast nat with default capcity of 2048 entries and 2 minute timeout
+    /// 2min default is taken from (https://datatracker.ietf.org/doc/html/rfc4787#section-4.3)
+    pub fn new(vpeer_ipv4: Ipv4Addr, vpeer_ipv6: Ipv6Addr) -> Self {
+        const LRU_CAPACITY: usize = 2048;
+        const LRU_TIMEOUT: Duration = Duration::from_secs(120);
+        Self::new_custom(
+            vpeer_ipv4,
+            vpeer_ipv6,
+            LRU_CAPACITY,
+            LRU_TIMEOUT,
```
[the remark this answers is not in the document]
jjanowsk: I disagree. Because I don't want to expose what type of cache I'm using internally. Imagine now you have 100 tests using this new_custom method. If you want to then change the implementation and use different cache internally you need to modify all of those tests.

Cargo.lock
```diff
@@ -4640,6 +4640,16 @@ dependencies = [
  "windows 0.34.0",
 ]
 
+[[package]]
+name = "telio-starcast"
+version = "0.1.0"
+dependencies = [
+ "pnet_packet 0.33.0",
```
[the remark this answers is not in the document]
jjanowsk: Will update the version in our workspace to get rid of this duplication

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+                ip_packet.set_checksum(ipv4::checksum(&ip_packet.to_immutable_ipv4()));
+            }
+            (IpAddr::V6(src_ip), IpAddr::V6(dst_ip)) => {
+                udp_packet.set_checksum(udp::ipv6_checksum(
+                    &udp_packet.to_immutable(),
+                    &src_ip,
+                    &dst_ip,
+                ));
+                ip_packet.set_destination(IpAddr::V6(dst_ip));
+            }
```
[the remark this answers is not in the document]
jjanowsk: You cannot get rid of those completely, because the dst_ip comes from cache. It's not known at compile time. I don't know if you've missed this part or you meant something else. Please let me know

crates/telio-starcast/fuzz/fuzz_targets/translate_outgoing.rs
```diff
@@ -0,0 +1,13 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+use telio_starcast::nat::{Nat, StarcastNat};
+
+fuzz_target!(|data: &[u8]| {
```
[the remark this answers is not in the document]
jjanowsk: Could e.g. find some invalid access to the slice via index operator. You can still panic in rust without unwraps.

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+            let dst_ip = ip_packet.get_destination().to_string();
+
+            let udp_packet = UdpPacket::new(&$buffer[IPV6_HEADER_LENGTH..]).unwrap();
+            let src_port = udp_packet.get_source().to_string();
+            let dst_port = udp_packet.get_destination().to_string();
+
+            assert_eq!(
+                format!("[{src_ip}]:{src_port}"),
+                $src_addr,
+                "Packet's src_addr does not match"
```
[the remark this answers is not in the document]
jjanowsk: 👍

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,648 @@
...
+            );
+        };
+    }
+
+    macro_rules! assert_err {
+        ($x:expr, $err:expr) => {
+            match $x {
+                Ok(_) => panic!(),
+                Err(e) => assert_eq!(e, $err),
+            }
```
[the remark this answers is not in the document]
jjanowsk: Would be, but only if you know about the existence of `matches!` ;)
:+1:

crates/telio-starcast/src/nat.rs
jjanowsk: I would propose to create an improvement task for this purpose and leave it as is for now

crates/telio-starcast/src/nat.rs
jjanowsk: I thought about it but I don't know if this is really a big deal. And what could be a possible protection? Remember that those are the packets coming from the other peer via transport socket. You could in theory limit this peer to some number of concurrent entries so that the other peers are not affected by some malicious software running on that peer, but other than that I don't know. Any ideas?

crates/telio-starcast/src/nat.rs
jjanowsk: It's actually the other way around ;) But I'll comment it to make it clear

crates/telio-starcast/src/lib.rs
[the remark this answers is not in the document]
jjanowsk: Please take a look if that's enough for you

packgron:
> > Hmm, as I understand this is a part of multicast protocol - why is it called starcast? It means sth (https://en.wiktionary.org/wiki/starcast) but I cannot see how it is connected.
> 
> The name comes from the RFC. It is the name for this particular implementation of multicast inside of a meshnet (so it's not just replacement word for multicast itself)

As we there targeting multicast and broadcast, (and in future we should be able to reuse these component for broadcast on some of the plaforms) i thought that it's '*cast' -- spelled out --> starcast :D

crates/telio-starcast/src/nat.rs
packgron: Oh i think i missed that you do udp checksum under this :upside_down_face:

crates/telio-starcast/src/nat.rs
packgron: I would be more for adding it now, it should not be that hard, expecialy if you have unit tests for it

crates/telio-starcast/src/nat.rs
packgron: So in reality my main fear would be a compleate fill up of allowed ports, as now its possible (probably hard, and lru should mitigate greatly), but a simple check that we dont execeed u16::MAX would suffice for me.
or we could even hallf it so in avergage case you would need only two tries to find empty place.

Further improving we can leave for later on

packgron: As now, because you are using peek, if at any point somebody manages to spam all 2^16 port's, you fall into loop that would never exit.

crates/telio-starcast/src/lib.rs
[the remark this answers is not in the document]
stalowyjez: Looks good!

crates/telio-starcast/src/nat.rs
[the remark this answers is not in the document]
stalowyjez: Hmm, maybe? Maybe `DEFAULT_*` would be better for the names as every capacity is LRU capacity here, so there is nothing specially LRU about this one? idk

crates/telio-starcast/src/nat.rs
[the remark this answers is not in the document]
stalowyjez: There could be parameterized proxy for this one as well, I guess? ;)
I'm not insisting though.

crates/telio-starcast/src/nat.rs
jjanowsk: Added. Please take a look

crates/telio-starcast/src/nat.rs
jjanowsk: I've limited the attempts to some hardcoded value to fix the infinite loop issue. That's the thing Linux kernel does in their nat implementation (as a last resort).

jjanowsk:
@packgron @stalowyjez Here's the diff since your last review: [diff](https://github.com/NordSecurity/libtelio/compare/02138cde9ce6de98f54da84d3619396c5f2ce44f..34ec793e3abaa4b5288734d295f514136122700b)

crates/telio-starcast/src/utils.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    fn set_source(&mut self, addr: IpAddr) {
+        if let IpAddr::V6(ipv6addr) = addr {
+            self.set_source(ipv6addr);
+        }
+    }
+
+    fn set_checksum(&mut self, _: u16) {
+        unimplemented!()
```
packgron, line 91: leftover ?

crates/telio-starcast/src/utils.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    fn to_immutable_ipv4<'p>(&'p self) -> Ipv4Packet<'p> {
+        unimplemented!()
```
packgron, line 95: leftover?

packgron (commented):
Awesome, few nits

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,794 @@
...
+    /// Cannot map the packet
+    CannotMap,
+    /// There is no mapping for the given port
+    MappingNotFound,
+}
+
+pub trait Nat {
+    /// Translate incomming packet (from the transport socket to the multicast peer)
+    /// Change the source to the multicast peer's ip and natted port
+    fn translate_incomming(&mut self, packet: &mut [u8]) -> Result<(), Error>;
```
mathiaspeters, line 30: Nit: `translate_incoming`

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,794 @@
...
+    vpeer_ipv4_addr: Ipv4Addr,
+    vpeer_ipv6_addr: Ipv6Addr,
+    mapper: Mapper,
+}
+
+impl StarcastNat {
+    /// Create new starcast nat with default capcity of 2048 entries and 2 minute timeout
+    /// 2min default is taken from (https://datatracker.ietf.org/doc/html/rfc4787#section-4.3)
+    pub fn new(vpeer_ipv4: Ipv4Addr, vpeer_ipv6: Ipv6Addr) -> Self {
+        const DEFAULT_CACHE_CAPACITY: usize = 2048;
```
packgron, line 59: NIT: move contants to the top of the file

packgron (dismissed):
+1.0 rest are nit's

crates/telio-starcast/src/utils.rs
```diff
@@ -0,0 +1,317 @@
...
+        sum + data as u32
+    }
+
+    fn sub_2_bytes(sum: u32, data: u16) -> u32 {
+        sum + !data as u32
+    }
+
+    #[allow(unwrap_check)]
+    #[allow(index_access_check)]
+    fn add_4_bytes(mut sum: u32, data: [u8; 4]) -> u32 {
```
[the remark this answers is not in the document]
jjanowsk: Done

crates/telio-starcast/src/utils.rs
jjanowsk: Nope, they should not be used. There is no checksum to set in ipv6 packet and to_immutable_ipv4 also does not make sense there. With the proper ut coverage I feel safe leaving them as panicking, but let me know if you want me to change them to no-op + comment

crates/telio-starcast/src/nat.rs
jjanowsk: :+1:

crates/telio-starcast/src/utils.rs
packgron: So add a comment at least

jjanowsk: This comment is obsolete in the new implementation. Resolving

crates/telio-starcast/src/nat.rs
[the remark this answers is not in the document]
jjanowsk: Fixed according to your suggestions. Thanks again. Looks neat

crates/telio-starcast/src/nat.rs
jjanowsk: @packgron I'm not a fan TBH and I wouldn't do it

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,794 @@
...
+            }
+        };
+
+        let mut udp_packet =
+            MutableUdpPacket::new(ip_packet.payload_mut()).ok_or(Error::PacketTooShort)?;
+        let src_port = udp_packet.get_source();
+        let src_addr = SocketAddr::new(src_ip, src_port);
+        let mut nat_port = (self.mapper)(src_addr);
+
+        const MAX_ATTEMPTS: u32 = 128;
```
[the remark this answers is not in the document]
jjanowsk: @packgron I'm not a fan TBH and I wouldn't do it

jjanowsk:
@packgron If you want you can take a look at the latest [diff](https://github.com/NordSecurity/libtelio/compare/50e512651e0212fc2c694bcdc1fb5a925f8ea527..b0229089dd1e10ebda86f23f64cea875d382f349)

This is base on @stalowyjez' suggestions. It does not contain any functional changes. Just a refactor to get rid of some unnecessary code duplication in the match branches. Also as a nice bonus it get rids of `unimplemented!()`.

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,744 @@
...
+    /// The packet has an unexpected transport protocol
+    UnexpectedTransportProtocol,
+    /// Cannot map the packet
+    CannotMap,
+    /// There is no mapping for the given port
+    MappingNotFound,
+}
+
+pub trait Nat {
+    /// Translate incomming packet (from the transport socket to the multicast peer)
```
mathiaspeters, line 28: Nit again: incoming :smile: There's plenty of the same typo in the tests in this file as well

crates/telio-starcast/src/nat.rs
```diff
@@ -0,0 +1,744 @@
...
+type Mapper = fn(src_addr: SocketAddr) -> u16;
+
+fn default_mapper(src_addr: SocketAddr) -> u16 {
+    let mut hasher = DefaultHasher::new();
+    match src_addr.ip() {
+        IpAddr::V4(src_ip) => Hash::hash_slice(&src_ip.octets(), &mut hasher),
+        IpAddr::V6(src_ip) => Hash::hash_slice(&src_ip.octets(), &mut hasher),
+    };
+    src_addr.port().hash(&mut hasher);
+    hasher.finish() as u16 // TODO: think if we want to use the full 16 bits or maybe some range e.g. >=1024
```
packgron, line 45: So we are full controll of all port's so it should be fine to use a full u16 range

packgron (approved):

stalowyjez (approved):
lgtm, +1

mathiaspeters (approved):
+1

PR #613, opened 2024-06-12
LLT-5277 port telio sockets crate from libtelio to external crate

matszczygiel:
### Problem
The VPN socket protection logic must be reusable by other libs, introducing as little dependencies as possible.

### Solution
Extract socket protection into a separate crate without any telio-specific stuff.

- [x] README.md is updated


matszczygiel:
My initial idea was to ditch the tokio dependency in favor of `std::thread` because it's only used for spawning long-running tasks. I decided not to do it for now and wait for some feedback before implementing that.
For logging I used the `log` crate for simplicity, in general, I've tried to keep the dependencies at bay.

.unreleased/LLT-5277
```diff
@@ -0,0 +1 @@
+Extract socket protection mechanism from `telio-sockets` into a separate `sock-prot` crate for reuse by other libs
```
mathiaspeters, line 1: The changelog is mostly there for apps integrating libtelio, and this is a very under-the-hood change, so I might be wrong but I don't think it needs a changelog entry

crates/sock-prot/src/protector/apple.rs
```diff
@@ -67,15 +66,15 @@ use objc_foundation::{INSString, NSString};
 #[cfg(any(target_os = "ios", target_os = "tvos"))]
 use version_compare::{compare_to, Cmp};
 
-pub(crate) fn bind_to_tun(sock: NativeSocket, tunnel_interface: u64) -> io::Result<()> {
-    telio_log_debug!(
+fn bind_to_tun(sock: NativeSocket, tunnel_interface: u64) -> io::Result<()> {
+    log::debug!(
```
tomaszklak, line 70: I'm not sure if this will hook into the telio logging facilities and send the log messages to apps callback. I think you will have to switch this to analogous macros from the `tracing` crate.

src/device/mod.rs
```diff
@@ -435,10 +426,6 @@ impl Device {
 
         telio_log_info!("libtelio is starting up with features : {:?}", features);
 
-        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
-        NETWORK_PATH_MONITOR_START
-            .call_once(telio_sockets::protector::platform::setup_network_path_monitor);
```
tomaszklak, line 440: Removing this and moving it into the `NativeProtector::new` might significantly reduce the time the OS has to generate the notification. I think this call should remain here in some form.

crates/sock-prot/src/protector/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+use std::io;
+
+pub use platform::NativeProtector;
+
+use crate::NativeSocket;
+
+pub trait Protector: Send + Sync {
+    /// # Safety
+    /// The socket descriptor must be valid
+    unsafe fn make_external(&self, _: NativeSocket) -> io::Result<()> {
```
tomaszklak, line 26: why has this become unsafe?

crates/sock-prot/src/protector/windows.rs
```diff
@@ -1,25 +1,21 @@
 use libc::setsockopt;
-use parking_lot::Mutex;
```
tomaszklak, line 2: nit: I would keep using the parking_lot

.unreleased/LLT-5277
matszczygiel: Ok, I'll remove it

crates/sock-prot/src/protector/apple.rs
matszczygiel: I tried to avoid that, `tracing` is a larger thing and quite tied to tokio, which ideally shouldn't be the dependency in the first place (tokio I mean). Tracing is used by telio but not by the other lib.
There's a way for `log` macros to work within telio, by the compatibility layer https://docs.rs/tracing-log/latest/tracing_log/#convert-log-records-to-tracing-events. Do you think we should use that?

crates/sock-prot/src/protector/mod.rs
matszczygiel: It takes in raw file descriptors, the caller needs to guarantee that this is a proper thing

crates/sock-prot/src/protector/windows.rs
matszczygiel: I don't see a reason why. It adds to a dependency that I deliberately tried to keep to a minimum

src/device/mod.rs
matszczygiel: We can create the protector in the `Device::new()`. Do you think that makes sense?

crates/sock-prot/src/protector/apple.rs
tomaszklak: `tracing_log` already [depends](https://github.com/tokio-rs/tracing/blob/master/tracing-log/Cargo.toml#L26) on `tracing_core` so imo you can just as well depend on `tracing`. Unless we know about some very likely future consumer of this crate that would find this to be a big issue.

crates/sock-prot/src/protector/mod.rs
tomaszklak: I'm not sure I understand. If someone passes calls `make_external` with a instance of `NativeSocket` - are there any additional requirements that need to be true for the `make_external` to work correctly that the compiler is not verifying?

crates/sock-prot/src/protector/apple.rs
matszczygiel: I mean using `tracing_log` from telio, for capturing the logs. This will save `sock-prot` from using tracing.
As I said we already have a lib that needs to use it and is not using tracing

crates/sock-prot/src/protector/mod.rs
matszczygiel: Yeah, since `NativeSocket` is `i32` (of Linux at least) then the caller needs to ensure this `i32` is a valid socket FD and not for example a file on disk or stdin (0)

crates/sock-prot/src/protector/windows.rs
matszczygiel: The thing is that this is inherently fallible and `parking_lot` simply ignores the error.
If the poisoning happens, then there's no way to handle this IMHO https://doc.rust-lang.org/nomicon/poisoning.html, and ignoring the problem sometimes makes sense (what `parking_lot` is doing) in the case of primitives but not in the case of a complex structure being protected like `Vec`.
And the additional thing is that adds to the dependencies bag.

matszczygiel: In reality, in our case, the mutexes can become poisoned in case of panic and that's already bad enough. It probably can happen only in the case of OOM. I don't think there's a need to optimize for such a case

.unreleased/LLT-5277
matszczygiel: Done

src/device/mod.rs
matszczygiel: I've moved the socket pool creation to `Device::new()`

matszczygiel:
Ok, now it seems clear that it's hard to make everyone happy with this change, meaning the libtelio team and the other lib need this component. Thus I will copy the component over into the other lib and leave libtelio unchanged.
Closing PR.

PR #614, opened 2024-06-12
When matching virtual peers specify all the common bits

tomaszklak:
### Problem
`AnalyticsEvent::is_from_virtual_peer` could mistakenly return true for more ipv6 addresses than the 7 specified for virtual peers.

### Solution
Match statement is corrected to specify all the common bits of the addresses.


mathiaspeters (approved):
+1

stalowyjez:
+1

stalowyjez (approved):

PR #617, opened 2024-06-13
LLT-5292: Make string the underlying UniFFI type of keys

mathiaspeters:
### Problem
All official apps use strings to represent keys internally. Using byte arrays for keys in the ffi layer meant that they had to do the conversion on their side before interacting with libtelio. By using string as the underlying type in UniFFI, the conversion happens transparently

### Solution
Use string as the underlying type for keys in UniFFI

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5292
+++ b/.unreleased/LLT-5292
@@ -1,0 +1,1 @@
+Change underlying FFI type for keys to be string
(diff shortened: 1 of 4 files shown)


gytsto (dismissed):
+1, but maybe commit for `making the body of events non-optional` is not needed here?

mathiaspeters:
> +1, but maybe commit for `making the body of events non-optional` is not needed here?

You're right, it was a mistake, but it's fixed now. Thanks!

PR #619, opened 2024-06-13
LLT-5291: Make event body non-optional

mathiaspeters:
### Problem
So far, the body of the event we send has been an Option, event though it should never be. The reason was to use a sort of builder pattern directly on the events, but this introduces some extra steps for apps where they have to check that the event, that should never be None, is none or not before they proceed. 

### Solution
To make this easier for them, we can make a proper builder pattern implementation for events instead

### DIFF
--- a/.unreleased/LLT-5291
+++ b/.unreleased/LLT-5291
@@ -1,0 +1,1 @@
+Make the body of events non-optional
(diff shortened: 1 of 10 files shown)


.unreleased/LLT-5291
```diff
@@ -0,0 +1 @@
+Make the body of events non-optional
```
gytsto, line 1: idk if this affects library users, so not sure if this is actually needed :shrug:

gytsto (approved):
+1

.unreleased/LLT-5291
packgron: It dies affect, we add do it to also improve udl

packgron (approved):
+1.0

PR #624, opened 2024-06-14
LLT-5325 Add OpenWRT linker

matislovas:
### Problem
Libtelio does not have openwrt targeted linkers defined in linkers config

### Solution
Add those linkers

- [x] README.md is updated

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -42,5 +42,11 @@ linker = "linkers/d-x86_64-linux-android21-clang"
 [target.i686-linux-android]
 linker = "linkers/d-i686-linux-android21-clang"
 
+[target.x86_64-openwrt-linux-musl]
+linker = "linkers/d-x86_64-openwrt-linux-musl"
+
+[target.aarch64-openwrt-linux-musl]
+linker = "linkers/d-aarch64-openwrt-linux-musl"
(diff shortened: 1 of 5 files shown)


gytsto (approved):
+1

Jauler:
I do not think this is required for OpenWRT, or by now, for any other platform. 

The reason we have these linker "interceptor" scripts is because, previously, when using swig cargo would configure linkers not to export any of the swig generated C symbols when building dynamic libs, resulting in linking errors on integrators side. Swig has been removed since that time and thus - we should not require to intercept linkers anymore.

matislovas:
> I do not think this is required for OpenWRT, or by now, for any other platform.
> 
> The reason we have these linker "interceptor" scripts is because, previously, when using swig cargo would configure linkers not to export any of the swig generated C symbols when building dynamic libs, resulting in linking errors on integrators side. Swig has been removed since that time and thus - we should not require to intercept linkers anymore.

Could be, I just added those, because I had errors while compiling `nordvpn-linux` app with `libtelio v4.3.3`. I will this _hanging_ there, until the `nordvpn-linux` will take the `libtelio 5.0.x` release and will test it than.

PR #627, opened 2024-06-17
Enforce conntrack limits only to succesfull tests

Jauler:
### Problem
Conntrack limits may be enforced for tests which has failed, which, in-fact is likely to go out of limits. This PR attempts to enforce the conntrack limits only for succesfull tests.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/helpers.py
+++ b/nat-lab/tests/helpers.py
@@ -3,6 +3,7 @@ import pytest
 from config import WG_SERVERS
 from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass, field
+from datetime import datetime
 from itertools import product, zip_longest
 from mesh_api import Node, Meshmap, API, stop_tcpdump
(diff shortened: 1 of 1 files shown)


nat-lab/tests/helpers.py
```diff
@@ -286,11 +287,13 @@ async def setup_environment(
 
     try:
         yield Environment(api, nodes, connection_managers, clients)
-    finally:
+
+        print(datetime.now(), "Checking connection limits")
```
Hasan6979, line 291: Is this print needed here ?

Hasan6979 (commented):
+1

nat-lab/tests/helpers.py
Jauler: IMHO, it makes it explicit when the connections limits are being checked in tests logs. Which helps a bit when analyzing failures.

ThrasherLT:
+1.0

ThrasherLT (approved):

PR #634, opened 2024-06-19
LLT-5311 Add event-log capture on windows VM in nat-lab

matislovas:
### Problem
We do sometimes get failures from wireguard adapters in windows, which do not produce any errors.

### Solution
Get event-log from windows VMs when test ends and append it to test logs.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -603,6 +603,8 @@ class Client:
                 f"-f {self._telio_features.to_json()}",
             ])
 
+        await self.clear_system_log()
+
         async with self._process.run(
             stdout_callback=on_stdout, stderr_callback=on_stderr
         ):
@@ -1048,6 +1050,76 @@ class Client:
         await process.execute()
         return process.get_stdout()
 
+    async def get_system_log(self) -> Optional[str]:
+        """
+        Get the system log on the target machine
+        Windows only for now
+        """
+        if self._connection.target_os == TargetOS.Windows:
+            output_notifier = OutputNotifier()
+            done_event = asyncio.Event()
+            done_event_str = "Get-EventLog - Done!"
+
+            async def on_stdout(stdout: str) -> None:
+                for line in stdout.splitlines():
+                    output_notifier.handle_output(line)
+
+            output_notifier.notify_output(
+                f"{done_event_str}",
+                done_event,
+            )
+
+            process = self._connection.create_process([
+                "powershell",
+            ])
+            async with process.run(
+                stdout_callback=on_stdout, stderr_callback=on_stdout
+            ):
+                await process.wait_stdin_ready()
+                await asyncio.sleep(1)
+                await process.write_stdin(
+                    "Get-EventLog -LogName Application -Newest 100 | format-table -wrap\n"
+                )
+                await process.write_stdin(
+                    "Get-EventLog -LogName System -Newest 100 | format-table -wrap\n"
+                )
+                await process.write_stdin(f"echo '{done_event_str}'\n")
+                await done_event.wait()
+                return process.get_stdout()
+        return None
+
+    async def clear_system_log(self) -> None:
+        """
+        Clear the system log on the target machine
+        Windows only for now
+        """
+        if self._connection.target_os == TargetOS.Windows:
+            output_notifier = OutputNotifier()
+            done_event = asyncio.Event()
+            done_event_str = "Clear-EventLog - Done!"
+
+            async def on_stdout(stdout: str) -> None:
+                for line in stdout.splitlines():
+                    output_notifier.handle_output(line)
+
+            output_notifier.notify_output(
+                f"{done_event_str}",
+                done_event,
+            )
+
+            process = self._connection.create_process([
+                "powershell",
+            ])
+            async with process.run(
+                stdout_callback=on_stdout, stderr_callback=on_stdout
+            ):
+                await process.wait_stdin_ready()
+                await asyncio.sleep(1)
+                await process.write_stdin("Clear-EventLog -LogName Application\n")
+                await process.write_stdin("Clear-EventLog -LogName System\n")
+                await process.write_stdin(f"echo '{done_event_str}'\n")
+                await done_event.wait()
+
     async def get_log_lines(self, regex: Optional[str] = None) -> List[str]:
         """
         Get the tcli log as a list of strings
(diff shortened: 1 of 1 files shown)


nat-lab/tests/telio.py
```diff
@@ -1049,6 +1051,44 @@ async def get_log(self) -> str:
...
+        Get the system log on the target machine
+        Windows only for now
+        """
+        if self._connection.target_os == TargetOS.Windows:
+            process = self._connection.create_process([
+                "powershell",
+            ])
+            async with process.run():
+                await process.wait_stdin_ready()
+                await asyncio.sleep(1)
```
gytsto, line 1065: Those sleeps ain't good, can we get some feedback between commands and outputs? Applies for all sleeps in this PR

matislovas: I took this idea from `libtelio/nat-lab/tests/test_fullcone_nat.py`. There are sleeps after `wait_stdin_ready()`. If I understood correctly, this does not mean _really really_ ready, that's why I added those.

gytsto: this maybe, but the following sleeps might be prevented

matislovas: I was worried about that, between issuing a command and fetching stdout, if we'll won't wait for actual output, the log can be corrupted or not retrieved fully.

matislovas: But I'll try without those sleeps ...

matislovas: Tried without additional sleeps so it fails to collect event-logs.

gytsto: you could do something like this:
`"Get-EventLog -LogName Application -Newest 100; echo 'done'\n"`
and then tell output notifier to wait for that `done` output, then you wouldn't need those sleeps

gytsto: same applies for other commands

matislovas: Noice! Changed it!

nat-lab/tests/telio.py
```diff
@@ -1048,6 +1050,76 @@ async def get_log(self) -> str:
...
+        Windows only for now
+        """
+        if self._connection.target_os == TargetOS.Windows:
+            output_notifier = OutputNotifier()
+            done_event = asyncio.Event()
+            done_event_str = "Clear-EventLog - Done!"
+
+            async def on_stdout(stdout: str) -> None:
+                for line in stdout.splitlines():
+                    output_notifier.handle_output(line)
```
jjanowsk, line 1103: Is this really needed here? Can't `output_notifier.handle_output` just be passed as a callback?

nat-lab/tests/telio.py
```diff
@@ -1048,6 +1050,76 @@ async def get_log(self) -> str:
...
+            )
+
+            process = self._connection.create_process([
+                "powershell",
+            ])
+            async with process.run(
+                stdout_callback=on_stdout, stderr_callback=on_stdout
+            ):
+                await process.wait_stdin_ready()
+                await asyncio.sleep(1)
```
jjanowsk, line 1117: Left over or is it needed? If it is needed then why?

jjanowsk (commented):
Those two comments also apply to the `get_system_log` function

nat-lab/tests/telio.py
gytsto: `output_notifier.handle_output` kinda expects line instead of multiple lines and while `ssh_process` executes callbacks on line, `docker_process` doesn't. I think it is better this way, to be future proof.

nat-lab/tests/telio.py
gytsto: `wait_stdin_ready` is more of a internal event, stating that `process` has been started and `stdin` should be ready. But that's not always the case. Like in this case, `powershell` process still needs some time to handle input unfortunately.

nat-lab/tests/telio.py
jjanowsk: But why does `output_notifier.handle_output` expect line? I looked at the code and it seems that it should handle multiline input as well. We have multiple places where this thin `on_stdout` wrapper is used and most likely those could be removed. It may be done separately as a small cleanup task. But maybe I'm missing sth

nat-lab/tests/telio.py
jjanowsk: Do we know exactly what is happening? Because it seems like some strange bug. If the process is not reading from stdin because it is busy with initialization the input data should just be waiting in a buffer and be consumed later. Where is it disappearing? Unless the powershell process is reading it and discarding during initialization but that would be very surprising behavior

nat-lab/tests/telio.py
gytsto: We could easily provide multiple lines to output notifier, just variable naming in in it's code suggest, that we expect line, no other constrains :smile: Probably we could easily get rid of those wrappers, will create an issue.

nat-lab/tests/telio.py
gytsto: Now that I've tested it myself, sleep might be not needed. Something is wrong with following commands, when there is no event to be displayed. Will dig more into it.

nat-lab/tests/telio.py
```diff
@@ -1048,6 +1050,76 @@ async def get_log(self) -> str:
...
+                for line in stdout.splitlines():
+                    output_notifier.handle_output(line)
+
+            output_notifier.notify_output(
+                f"{done_event_str}",
+                done_event,
+            )
+
+            process = self._connection.create_process([
+                "powershell",
```
Jauler, line 1073: Have you tried `powershell -Command Get-EventLog....`? or if that does not work `powershell -File <some script with Get-EventLog command>`? It sounds like it should simplify things.

gytsto: @matislovas tried and couldn't manage to make it work, myself I haven't tried

nat-lab/tests/telio.py
gytsto: LLT-5348

nat-lab/tests/telio.py
gytsto: Rewrote it now, no sleeps are needed

nat-lab/tests/telio.py
gytsto: Managed to make it work with `powershell -Command Get-EventLog`, so should be more clean now

Jauler (approved):
+2.0

PR #636, opened 2024-06-19
Wrap libtelio_remote runtime into try-except block

matislovas:
### Problem
During nightly tests, we're sometimes getting the `FATAL: exception not rethrown` error, when executing `libtelio_remote.py` file.

### Solution
Try to catch and print, what does this not-rethrown exception has in it.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/uniffi/libtelio_remote.py
+++ b/nat-lab/tests/uniffi/libtelio_remote.py
@@ -188,12 +188,16 @@ def main():
     container_ip = sys.argv[2]
     port = int(sys.argv[3])
 
-    daemon = Pyro5.server.Daemon(host=container_ip, port=port)
-    wrapper = LibtelioWrapper(daemon)
-    daemon.register(wrapper, objectId=object_name)
+    try:
+        daemon = Pyro5.server.Daemon(host=container_ip, port=port)
+        wrapper = LibtelioWrapper(daemon)
+        daemon.register(wrapper, objectId=object_name)
 
-    daemon.requestLoop()
-    daemon.close()
+        daemon.requestLoop()
+        daemon.close()
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        print(f"libtelio_remote error: {e}")
+        raise e
 
 
 if __name__ == "__main__":


nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -188,12 +188,16 @@ def main():
...
-    daemon.requestLoop()
-    daemon.close()
+    try:
+        daemon = Pyro5.server.Daemon(host=container_ip, port=port)
+        wrapper = LibtelioWrapper(daemon)
+        daemon.register(wrapper, objectId=object_name)
+
+        daemon.requestLoop()
+        daemon.close()
+    except (Exception,) as e:
```
gytsto, line 198: Why not just `Exception` instead of using tuple?
Or is it to satisfy linter, because of too general exception? If yes, just disable `pylint` on that line or something :smile:

matislovas: Yeah, that was an attempt to satisfy linter, but it seems, that it didn't work :smile:  will disable that ...

gytsto: Also, now that I think. With sys.exit we kinda break flow of exception. Instead of it, just do `raise` after the print

matislovas: Fixed

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -188,12 +188,16 @@ def main():
...
-    daemon.close()
+    try:
+        daemon = Pyro5.server.Daemon(host=container_ip, port=port)
+        wrapper = LibtelioWrapper(daemon)
+        daemon.register(wrapper, objectId=object_name)
+
+        daemon.requestLoop()
+        daemon.close()
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        print(f"libtelio_remote error: {e}")
```
mathiaspeters, line 199: I don't think this print is going to show up anywhere. It might better to write it to `tcli.log`

PR #638, opened 2024-06-20
Ignore RUSTSEC-2024-0344 in cargo-deny lint

jjanowsk:
This is a temporary ignore until we are able to get rid of the affected dependency.


Jauler (approved):
+2.0

packgron (approved):
+2.0

PR #639, opened 2024-06-20
Resolve LLT-5293: Fix mesh state after disconnecting

gytsto:
### Problem
After mesh peer is gone, `Connected` event is being generated, because last rx timestamp is being resetted.

### Solution
Skip resetting rx timestamp when peer is actually in `Disconnected` state

- [x] README.md is updated


dfetti (commented):
+1.0

crates/telio-wg/src/wg.rs
```diff
@@ -115,6 +121,81 @@ pub struct Io {
...
+}
+
+impl BytesAndTimestamps {
+    /// docs
+    pub fn new(rx_bytes: Option<u64>, tx_bytes: Option<u64>) -> Self {
+        let now = Instant::now();
+        BytesAndTimestamps {
+            rx_bytes: rx_bytes.unwrap_or_default(),
+            tx_bytes: tx_bytes.unwrap_or_default(),
+            rx_ts: now,
```
Jauler, line 149: IMHO, we have a slightly dangerous "condition" here. As `BytesAndTimestamps` is being inserted when new node is added -> no rx or no tx has happened, ever. In current variant this is indicated using `rx_bytes` or `tx_bytes` of zero indicates that. In addition, one can only know that `rx_ts/tx_ts` is "valid" only after looking at `rx_bytes` and `tx_bytes`.

But maybe it would be clearer/easier to understand if we would use `Option<u64>` and `Option<Instant>` and simply set `None` under such conditions.

@gytsto WDYT?

gytsto: Yup, you are totally right. Will do that

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -315,48 +256,26 @@ impl StateDecision {
...
-            State {
-                stats,
-                variant: StateVariant::Up,
-            }
-        } else {
-            State {
-                stats,
-                variant: StateVariant::Down,
-            }
+    fn new(stats: Arc<Mutex<BytesAndTimestamps>>, rtt: Duration) -> Self {
```
Jauler, line 266:
```suggestion
    fn new(stats: Arc<Mutex<BytesAndTimestamps>>, cfg_max_allowed_rtt: Duration) -> Self {
```

nit: as `rtt` is representing configured maximum allowed RTT value which is used to declare link down -> would be cool to update the name to make it slightly more clear.

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -315,48 +256,26 @@ impl StateDecision {
...
-            *self = State::new(rx_bytes, tx_bytes, node_state);
-            return Self::build_result(StateDecision::NoAction, LinkState::Down);
-        }
-
-        let new_rx = rx_bytes.unwrap_or_default();
-        let new_tx = tx_bytes.unwrap_or_default();
-        self.stats.update(new_rx, new_tx);
-
-        let is_link_up = self.stats.is_link_up(rtt);
+    fn update(&mut self, rtt: Duration, ping_enabled: bool) -> StateUpdateResult {
```
Jauler, line 277:
```suggestion
    fn update(&mut self, cfg_max_allowed_rtt: Duration, ping_enabled: bool) -> StateUpdateResult {
```
same as https://github.com/NordSecurity/libtelio/pull/639/files#r1665336881

crates/telio-wg/src/link_detection/mod.rs
```diff
@@ -33,47 +36,24 @@ impl LinkDetection {
...
-        &mut self,
-        public_key: &PublicKey,
-        rx_bytes: Option<u64>,
-        tx_bytes: Option<u64>,
-        node_state: NodeState,
-    ) {
-        match self {
-            LinkDetection::Enabled(l) => l.insert(public_key, rx_bytes, tx_bytes, node_state),
-            LinkDetection::Disabled(l) => l.insert(public_key, rx_bytes),
+            LinkDetection::Enabled(l) => l.insert(public_key, stats),
```
Jauler, line 41: I believe that in case link detection is disabled -> we do not need to do anything here anymore, because `time_since_last_rx()` is not being retreived via `link_state` module anymore, but rather directly from `wg.rs`.

Which would mean, that we can delete `struct LinkDetectionDisabled` and simplify the code a bit more :)

crates/telio-wg/src/wg.rs
gytsto: updated

crates/telio-wg/src/link_detection/mod.rs
gytsto: renamed

crates/telio-wg/src/link_detection/mod.rs
gytsto: same here

crates/telio-wg/src/link_detection/mod.rs
gytsto: removed entirely

crates/telio-wg/src/wg.rs
```diff
@@ -115,6 +122,86 @@ pub struct Io {
...
+}
+
+impl BytesAndTimestamps {
+    /// docs
+    pub fn new(rx_bytes: Option<u64>, tx_bytes: Option<u64>) -> Self {
+        let now = Instant::now();
+        BytesAndTimestamps {
+            rx_bytes: rx_bytes.unwrap_or_default(),
+            tx_bytes: tx_bytes.unwrap_or_default(),
+            rx_ts: if rx_bytes.is_none() { None } else { Some(now) },
```
Jauler, line 150: I believe, we should check both, `rx_bytes.is_none()` as well as `rx_bytes == 0`. And of course, same for `tx`.

gytsto: fixed

Jauler (dismissed):
+1.0

matislovas (approved):

matislovas (dismissed):

dfetti (approved):
Looks good to me

crates/telio-wg/src/wg.rs
```diff
@@ -419,9 +516,21 @@ impl WireGuard for DynamicWg {
     }
 
     async fn time_since_last_rx(&self, public_key: PublicKey) -> Result<Option<Duration>, Error> {
-        Ok(task_exec!(&self.task, async move |s| Ok(
-            s.time_since_last_rx(public_key)
-        ))
+        Ok(task_exec!(&self.task, async move |s| Ok(s
```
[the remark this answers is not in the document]
gytsto: you are right, changed it

PR #640, opened 2024-06-21
Fix wrong key type error

mathiaspeters:
### Problem
A recent set of commits again broke natlab. One commit was changing the underlying type for keys to strings, while another commit added more functions for dealing with keys. There was no code overlap so there were no conflicts, yet the change of underlying type for keys is made the added key handling break, causing type check errors and broke a test

### Solution
Make the new key-handling code use strings directly instead of trying to convert to and from bytes

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -1,7 +1,6 @@
 # pylint: disable=too-many-lines
 
 import asyncio
-import base64
 import datetime
 import json
 import os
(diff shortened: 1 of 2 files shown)


LukasPukenis (commented):
Regarding the fix - it looks like it replaces few places how the key is generated however the changes don't look like they should break anything. Is it because the keys are generated somewhere else or probably expected to be in a public-key form?

mathiaspeters:
Previously, keys would be `bytes` since that is what we use in libtelio, but the different apps all use base64-encoded strings so we changed the underlying type to be strings instead, which means that the uniffi bindings expect strings wherever we talk about keys, not bytes as it was before. There was a PR recently that added a test around changing the private key, and with that introduced wrapper functions in `telio.py` for generating keys (which calls the bindings, to do the actual work). When that PR was merged, keys were still bytes, and then after that, the PR that changed keys to be strings was merged. The newly introduced code that asked libtelio for keys, as bytes, and then parsed them to strings, no longer worked. This PR fixes those places to expect and use the strings directly, since that's what were doing now

LukasPukenis (commented):
+1.5

dfetti (approved):
+1.0

PR #642, opened 2024-06-25
Add downgrade logic using link detection

dfetti:
The old downgrade logic was triggered if we did not receive anything for three direct keepalive period.

For battery improvement purpouses we want to increase these keepalive periods from 5 seconds to somewhere around 120 +- 5 seconds, therefore the old mechanism for downgrade won't work anymore.

The new solution uses the link detection's reports to trigger the downgrades. If the link_state of a peer changes to Down and the peer has a direct connection, we will trigger the downgrade.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -447,9 +447,13 @@ pub struct FeatureLinkDetection {
     #[serde(default = "FeatureLinkDetection::default_configurable_rtt")]
     pub rtt_seconds: u64,
 
-    /// Check the link state before reporting it as down.
+    /// Check the link state before reporting it as down
     #[serde(default = "FeatureLinkDetection::default_no_of_pings")]
     pub no_of_pings: u32,
+
+    /// Use link detection for downgrade logic
+    #[serde(default = "FeatureLinkDetection::default_use_for_downgrade")]
+    pub use_for_downgrade: bool,
 }
 
 impl FeatureLinkDetection {
@@ -459,6 +463,10 @@ impl FeatureLinkDetection {
 
     const fn default_no_of_pings() -> u32 {
         0
+    }
+
+    const fn default_use_for_downgrade() -> bool {
+        false
     }
 }
 
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -60,6 +60,8 @@ pub trait WireGuard: Send + Sync + 'static {
     async fn wait_for_proxy_listen_port(&self, d: Duration) -> Result<u16, Error>;
     /// Get adapter file descriptor if supported by platform, `None` otherwise
     async fn get_wg_socket(&self, ipv6: bool) -> Result<Option<i32>, Error>;
+    /// Get current link state for a peer
+    async fn get_link_state(&self, key: PublicKey) -> Result<Option<LinkState>, Error>;
     /// Set secret key for adapter
     async fn set_secret_key(&self, key: SecretKey) -> Result<(), Error>;
     /// Set adapter fwmark, unix only
@@ -339,7 +341,6 @@ impl DynamicWg {
         #[cfg(unix)] cfg: Config,
     ) -> Self {
         let interval = interval(Duration::from_millis(POLL_MILLIS));
-
         Self {
             task: Task::start(State {
                 #[cfg(unix)]
@@ -434,6 +435,15 @@ impl WireGuard for DynamicWg {
         task_exec!(&self.task, async move |s| Ok(s.adapter.get_wg_socket(ipv6)))
             .await
             .unwrap_or(Ok(None))
+    }
+
+    async fn get_link_state(&self, key: PublicKey) -> Result<Option<LinkState>, Error> {
+        Ok(task_exec!(&self.task, async move |s| {
+            Ok(s.link_detection
+                .as_ref()
+                .and_then(|ld| ld.get_link_state_for_downgrade(&key)))
+        })
+        .await?)
     }
 
     async fn set_secret_key(&self, key: SecretKey) -> Result<(), Error> {
@@ -1182,7 +1192,6 @@ pub mod tests {
     pub async fn setup(#[cfg(all(not(test), feature = "test-adapter"))] cfg: Config) -> Env {
         let events_ch = Chan::default();
         let analytics_ch = Some(McChan::default().tx);
-
         let adapter = Arc::new(Mutex::new(MockAdapter::new()));
 
         adapter
--- a/nat-lab/tests/telio_features.py
+++ b/nat-lab/tests/telio_features.py
@@ -81,6 +81,7 @@ class Nurse(DataClassJsonMixin):
 class LinkDetection(DataClassJsonMixin):
     rtt_seconds: Optional[int] = None
     no_of_pings: Optional[int] = 0
+    use_for_downgrade: Optional[bool] = False
 
 
 @dataclass_json(undefined=Undefined.EXCLUDE)
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -10,6 +10,7 @@ use telio_dns::DnsResolver;
 use telio_firewall::firewall::{Firewall, FILE_SEND_PORT};
 use telio_model::constants::{VPN_EXTERNAL_IPV4, VPN_INTERNAL_IPV4, VPN_INTERNAL_IPV6};
 use telio_model::features::Features;
+use telio_model::mesh::LinkState;
 use telio_model::mesh::NodeState::Connected;
 use telio_model::EndpointMap;
 use telio_model::SocketAddr;
@@ -685,11 +686,14 @@ async fn build_requested_meshnet_peers_list<
             _ => None,
         };
 
+        let link_state = wireguard_interface.get_link_state(*public_key).await?;
+
         // Compute the current endpoint state
         let peer_state = peer_state(
             actual_peer,
             time_since_last_rx_or_handshake.as_ref(),
             time_since_last_endpoint_change.as_ref(),
+            link_state,
             match proxy_endpoint {
                 Some(eps) => eps,
                 None => &[],
@@ -922,6 +926,7 @@ fn peer_state(
     peer: Option<&telio_wg::uapi::Peer>,
     time_since_last_rx: Option<&Duration>,
     time_since_last_endpoint_change: Option<&Duration>,
+    link_state: Option<LinkState>,
     proxy_endpoints: &[SocketAddr],
     requested_state: &RequestedState,
 ) -> PeerState {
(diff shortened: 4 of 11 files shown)


crates/telio-wg/src/wg.rs
```diff
@@ -677,9 +688,23 @@ impl State {
                     )
                     .await?;
                 }
+
+                if self.should_emit_downgrade_events
+                    && link_detection_update_result.should_notify
```
LukasPukenis, line 693: Both `should_emit_downgrade_events` and `should_notify` look like they mean the same thing, are they both needed?

src/device/wg_controller.rs
```diff
@@ -943,13 +943,27 @@ fn peer_state(
...
+            // If link detection reported link state down and path is direct
+            false
+        } else {
+            true
+        }
+    } else {
+        // Use old keepalive method
+        time_since_last_rx < &peer_connectivity_timeout
+    };
+
```
LukasPukenis, line 966: This looks pretty muh how I imagine it should be, very minimal, which is great IMO :)

crates/telio-wg/src/wg.rs
```diff
@@ -677,9 +688,23 @@ impl State {
...
+                    && link_detection_update_result.link_state == Some(LinkState::Down)
+                {
+                    downgrade_event.to_be_downgraded.push(*key);
```
LukasPukenis, line 696: I am a bit anxious about this list. The problem with it is that it lives a separate live where it's being filled with keys but I never see it being `poped` or `cleared`. This might introduce problems or fluctuations down the road where this list is consulted when judging the downgrade possibility. However the list needs to be carefully managed as it's a liability.

I have an idea - would it be possible to spin a tokio task in the wg_controller basically listening for the events additionally to the current logic? That way instead of saving the the list, the value would be pused down the channel and wg_controller would catch it and break the direct connection immediately. This way there would be no need for bookkeeping. How does that sound?

crates/telio-model/src/features.rs
```diff
@@ -438,9 +438,13 @@ pub struct FeatureLinkDetection {
...
     pub rtt_seconds: u64,
 
-    /// Check the link state before reporting it as down.
+    /// Check the link state before reporting it as down
     #[serde(default = "FeatureLinkDetection::default_no_of_pings")]
     pub no_of_pings: u32,
+
+    /// Use link detection for downgrade logic
+    #[serde(default = "FeatureLinkDetection::default_use_for_downgrade")]
+    pub use_for_downgrade: bool,
```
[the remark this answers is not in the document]
dfetti: For me it made more sense to be in the link_detection config. This way you can't missconfigure it in a way that downgrade trigger is link detection and link detection is off.

crates/telio-wg/src/lib.rs
```diff
@@ -12,7 +12,17 @@ pub mod uapi;
...
+
 pub use crate::{
     adapter::{Adapter, AdapterType, Error, FirewallCb, Tun},
     wg::*,
 };
+
+/// Event used for triggering connections downgrade
+#[derive(Clone, Default)]
+pub struct DowngradeEvent {
+    // TODO: Make it a hashset
```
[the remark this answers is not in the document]
dfetti: Not the duplicates are the reason. This is used only for push and contains checks so a hashset would be more efficient.

crates/telio-wg/src/wg.rs
dfetti: self.should_emit_downgrade_events - the config says we use link detection for downgrade
link_detection_update_result.should_notify - the link state of the peer has been changed

crates/telio-wg/src/wg.rs
dfetti: I don't understand what you mean. This is exactly what I am doing. I build the list here, is sent through the channel, received in src/device/mod.rs, saved into requested_state, then I call wg_consolidate and then I clear the list. So there are no leftovers after the operation.

crates/telio-model/src/features.rs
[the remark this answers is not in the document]
LukasPukenis: I think features should live neraby where they are used, it makes for an easier discovery. The way I look - link detection shouldn't know where it's gonna be used, it is it's own separate thing living a separate live, meanwhile direct connections config is all about direct connections, how they behave, so downgrade living there makes more sense to me :)

crates/telio-wg/src/wg.rs
LukasPukenis: Sorry, I don't understand :/

dfetti: So, the first one, self.should_emit_downgrade_events, indicates that libtelio is configured to use link detection for downgrade. The feature flag use_for_downgrade is set.
The second one is the result of link_detection.update() which indicates that the link_state has changed at this update call.

crates/telio-wg/src/wg.rs
```diff
@@ -113,6 +113,9 @@ pub struct Io {
     pub analytics_tx: Option<mc_chan::Tx<Box<AnalyticsEvent>>>,
     /// Channel to transmit event to registered event callback
     pub libtelio_wide_event_publisher: Option<mc_chan::Tx<Box<LibtelioEvent>>>,
+    /// Channel to transmit the list of peers that are detected with LinkState::Down
+    /// by the LinkDetection mechanism
+    pub downgrade_event_publisher: Tx<Box<DowngradeEvent>>,
```
matszczygiel, line 118: Why `Box<>`?

src/device/mod.rs
```diff
@@ -1136,6 +1147,10 @@ impl Runtime {
         let requested_state = RequestedState {
             device_config: config.clone(),
             keepalive_periods: features.wireguard.persistent_keepalive.clone(),
+            use_link_detection_for_downgrade: features
+                .link_detection
+                .map(|f| f.use_for_downgrade)
+                .unwrap_or(false),
```
matszczygiel, line 1153:
```suggestion
                .map_or(false, |f| f.use_for_downgrade),
```

src/device/mod.rs
```diff
@@ -2242,6 +2259,27 @@ impl TaskRuntime for Runtime {
                 Ok(())
             },
 
+            Some(downgrade_event) = self.event_listeners.downgrade_event_subscriber.recv() => {
+                telio_log_debug!("WG consolidation triggered by downgrade event");
+
+                // Update this in case config has been changed
+                self.requested_state.use_link_detection_for_downgrade = self.features.link_detection.as_ref().map(|f| f.use_for_downgrade).unwrap_or(false);
```
matszczygiel, line 2266: Same `map_or()`

src/device/wg_controller.rs
```diff
@@ -943,13 +943,27 @@ fn peer_state(
...
         .endpoint
         .map(|ep| proxy_endpoints.contains(&ep))
         .unwrap_or_default();
     let is_in_upgrade_window = time_since_last_endpoint_change < &peer_upgrade_window;
 
+    let has_contact = if requested_state.use_link_detection_for_downgrade {
+        if time_since_last_rx > &Duration::from_secs(180) {
+            // Safety duration
+            false
+        } else if requested_state.down_connections.contains(&peer.public_key) && !is_proxying {
```
matszczygiel, line 956: Have you considered using `HashSet` for storing down connections?

tomasz-grz (commented):
+0.5

crates/telio-model/src/features.rs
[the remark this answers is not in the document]
matszczygiel: My opinion is not strong but I agree with @dfetti. The current approach makes it difficult to misconfigure if link detection is off. Logically it makes sense that these kinds of optional features affect the connection logic but for the connection logic to know about every optional feature is a spaghetti dependency

matszczygiel:
+1

LukasPukenis (commented):
I think it looks okay, it should work as intended, if the flag is toggled, then link detection is used for the downgrade in `wg_controller`.

We have this blind spot though where if there's no traffic:
1) A -> B
2) B <- A
3) [silence, link detection thinks all is good]
4) [break the connection, link detection has no idea]

Link detection is oblivious to it, meanwhile the old method would catch it. As we discussed, it's fine, as it would simply cause a small delay to downgrade to relayed, however I think it might be nice to actually have a testcase for it where both A and B would send data to each other and then break the direct link, idle for a bit, like 30seconds, and then would try sending data.

In that case we should observe that it still works, through relayed, properly(we could also observe and enforce how long it must take. It must not take more than 10seconds + rtt(configurable) for link detection to kick in).

I think such a testcase would be nice to have as it would cleraly show our blind spot, but also would show that it still works.

+0.75 otherwise

nat-lab/tests/telio_features.py
```diff
@@ -80,6 +80,7 @@ class Nurse(DataClassJsonMixin):
 class LinkDetection(DataClassJsonMixin):
     rtt_seconds: Optional[int] = None
     no_of_pings: Optional[int] = 0
+    use_for_downgrade: Optional[bool] = False
```
mathiaspeters, line 83: There's some testing for these `TelioFeatures`, let's add this field to that test as well

dfetti: Ok, on it

mathiaspeters (dismissed):
+1

src/device/mod.rs
```diff
@@ -214,6 +214,12 @@ pub struct RequestedState {
 
     // Requested keepalive periods
     pub(crate) keepalive_periods: FeaturePersistentKeepalive,
+
+    // Use link detection for downgrades
+    pub use_link_detection_for_downgrade: bool,
+
+    // Down connections
+    pub down_connections: HashSet<PublicKey>,
```
Jauler, line 222: Maybe it would be possible to make connection management based on link status stateless? I mean why do we need a copy of link states for peers when WG already has this info?

src/device/mod.rs
```diff
@@ -346,6 +352,7 @@ pub struct EventListeners {
     endpoint_upgrade_event_subscriber: chan::Rx<UpgradeRequestChangeEvent>,
     stun_server_subscriber: chan::Rx<Option<StunServer>>,
     post_quantum_subscriber: chan::Rx<telio_pq::Event>,
+    downgrade_event_subscriber: chan::Rx<telio_wg::DowngradeEvent>,
```
Jauler, line 355: Would it be possible to avoid events? 

In this case, `wg_controller` which is controlling the peers communication path has direct access to wg module. Maybe we can use simple method calls instead of events, as events+statefull things are usually very error-prone.

crates/telio-model/src/features.rs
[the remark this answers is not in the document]
Jauler: I would agree with @matszczygiel and @dfetti that having a structure which "naturally" forbids invalid configs - is preferable :).

Jauler (approved):
+1.0

PR #643, opened 2024-06-26
Llt 5319 move to our chachabox wrapper

stalowyjez:
### Problem
There is a vulnerability in the `curve25519-dalek` crate in the currently used version - unfortunately we cannot update it simply, because it requires changing the version of the `crypto_box` crate, which changed the `ChaChaBox` implementation and thus it would make our client incompatible with the older versions.

### Solution
This commit bumps the `crypto_box` and `curve25519-dalek` crates while replacing the `crypto_box::ChaChaBox` with our wrapper around `chacha20poly1305::XChaCha20Poly1305` crypto box.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5319
+++ b/.unreleased/LLT-5319
@@ -1,0 +1,1 @@
+Bump crypto_box and thus curve25519-dalek to get rid of RUSTSEC-2024-0344
--- a/crates/telio-crypto/src/lib.rs
+++ b/crates/telio-crypto/src/lib.rs
@@ -160,13 +160,13 @@ impl PresharedKey {
 
 impl From<crypto_box::SecretKey> for SecretKey {
     fn from(sk: crypto_box::SecretKey) -> Self {
-        Self(*sk.as_bytes())
+        Self(sk.to_bytes())
     }
 }
 
 impl From<&crypto_box::SecretKey> for SecretKey {
     fn from(sk: &crypto_box::SecretKey) -> Self {
-        Self(*sk.as_bytes())
+        Self(sk.to_bytes())
     }
 }
 
--- a/crates/telio-crypto/src/lib.rs
+++ b/crates/telio-crypto/src/lib.rs
@@ -20,6 +20,7 @@
 //! # }
 //! ```
 
+pub mod chachabox;
 pub mod encryption;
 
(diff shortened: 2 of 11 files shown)


.unreleased/LLT-5319
```diff
@@ -0,0 +1 @@
+Bump crypto_box version and move to our ChaChaBox implementation
```
mathiaspeters, line 1: I don't think this needs a changelog entry since apps don't need to care about this change

crates/telio-crypto/src/lib.rs
```diff
@@ -130,7 +132,7 @@ impl SecretKey {
     /// assert_ne!(pub_key_a, pub_key_b);
     /// ```
     pub fn public(&self) -> PublicKey {
-        crypto_box::SecretKey::from(self.0).public_key().into()
+        PublicKey(MontgomeryPoint::mul_base(&Scalar::from_bytes_mod_order(clamp_integer(self.0))).0)
```
mathiaspeters, line 135: This is a very dense line of code. Can we make it more understandable? Perhaps with a comment or split it into stages with descriptive variable names?

mathiaspeters (commented):
I'm not too familiar with these things but code looks good. +0.75

gytsto (commented):
+1

crates/telio-crypto/src/lib.rs
stalowyjez: @mathiaspeters well, I changed it while thinking about getting rid of this crate completely, but I guess in the current state it will be better to just leave the old version here

.unreleased/LLT-5319
stalowyjez: Hmm maybe? There was a vulnerability found in one of the crates used by the old `crypto_box`, co I thought it is worth mentioning.

mathiaspeters: That's a good point, but in that case I think the CVE that's fixed should be mentioned to make that more clear

mathiaspeters (approved):
Looks good! +1

Jauler (approved):
+1.0

Jauler (approved):

PR #644, opened 2024-06-27
LLT-5349: Make docker port mapping less flaky

mathiaspeters:
### Problem
The port mapping is necessary for natlab to be run with macos or windows as a host, but is not necessary on linux and sometimes breaks our CI pipeline due to the port being busy on the host.

### Solution
Natlab already discovers the port mapping fully from the docker information so we  don't need to have the same port on both container and host. So a better option is to just bind a port in the container and let docker choose an available port on the host. We can also use non-ephemeral ports to make chance of collision in the container even less likely

- [x] README.md is updated

### DIFF
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -72,7 +72,7 @@ services:
       cone-net-01:
         ipv4_address: 192.168.101.104
     ports:
-      - "58001:58001"
+      - "8081"
     environment:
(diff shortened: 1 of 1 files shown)


ThrasherLT (commented):
+1.0

dfetti (dismissed):
+1.0

tomasz-grz (approved):
+1.0

PR #645, opened 2024-06-28
[LLT-5359] add logging to catch dns errors on android

Jauler:
### Problem
In same rare cases we observe unexplained burst of failures in DNS resolution on Android. This PR attempts to add some additional information which should help catch the issue.

- [x] README.md is updated
- [x] changelog.md is updated

### DIFF
--- a/crates/telio-dns/src/forward.rs
+++ b/crates/telio-dns/src/forward.rs
@@ -218,6 +218,10 @@ impl Authority for ForwardAuthority {
         let name: LowerName = name.clone();
         let resolve = self.resolver.lookup(name, rtype).await;
 
+        if resolve.is_err() {
+            telio_log_warn!("DNS name resolution failed with {:?}", resolve);
+        }
+
         resolve
             .map(ForwardLookup)
             .map_err(|code| match code.kind() {
--- a/crates/telio-dns/src/nameserver.rs
+++ b/crates/telio-dns/src/nameserver.rs
@@ -157,7 +157,11 @@ impl LocalNameServer {
                         {
                             Ok(length) => length,
                             Err(e) => {
-                                telio_log_error!("[DNS] {}", e);
+                                telio_log_error!(
+                                    "[DNS] {}. Offending request packet: {:?}",
+                                    e,
+                                    packet
+                                );
                                 return;
                             }
                         };
--- a/deny.toml
+++ b/deny.toml
@@ -9,7 +9,9 @@ vulnerability = "deny"
 unmaintained = "warn"
 yanked = "warn"
 notice = "warn"
-ignore = []
+ignore = [
+    { id = "RUSTSEC-2024-0344", reason = "This is temporary ignore untill we get rid of this dependency" },
+]
 
 
 [licenses]
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -391,6 +391,15 @@ pub extern "C" fn telio_start_named(
     adapter: telio_adapter_type,
     name: *const c_char,
 ) -> telio_result {
+    let sk = ffi_try!(char_ptr_to_type::<SecretKey>(private_key));
+    telio_log_info!(
+        "telio_start_named entry with instance id: {}. Public key: {:?}. Adapter: {:?}. Name: {:?}",
+        dev.id,
+        sk.public(),
+        &adapter,
+        char_ptr_to_type::<String>(name)
+    );
+
     ffi_catch_panic!({
         let mut dev = ffi_try!(dev.inner.lock().map_err(|_| TELIO_RES_LOCK_ERROR));
 
@@ -424,6 +433,15 @@ pub extern "C" fn telio_start_with_tun(
     adapter: telio_adapter_type,
     tun: c_int,
 ) -> telio_result {
+    let sk = ffi_try!(char_ptr_to_type::<SecretKey>(private_key));
+    telio_log_info!(
+        "telio_start_with_tun entry with instance id: {}. Public key: {:?}. Adapter: {:?}. Tunnel FD: {:?}",
+        dev.id,
+        sk.public(),
+        &adapter,
+        tun
+    );
+
     ffi_catch_panic!({
         let mut dev = ffi_try!(dev.inner.lock().map_err(|_| TELIO_RES_LOCK_ERROR));
         let private_key = ffi_try!(char_ptr_to_type::<SecretKey>(private_key));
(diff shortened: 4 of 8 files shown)


deny.toml
```diff
@@ -9,7 +9,9 @@ vulnerability = "deny"
 unmaintained = "warn"
 yanked = "warn"
 notice = "warn"
-ignore = []
+ignore = [
+    { id = "RUSTSEC-2024-0344", reason = "This is temporary ignore untill we get rid of this dependency" },
+]
```
LukasPukenis, line 14: Sneaky

gytsto (approved):
since these are just log, easy +2 imo

LukasPukenis (commented):
+1 Though I don't like that it also touches linter settings and modifies ignore list

LukasPukenis (approved):

gytsto (approved):

LukasPukenis (approved):

src/ffi/mod.rs
```diff
@@ -719,6 +739,11 @@ pub extern "C" fn telio_connect_to_exit_node_postquantum(
     allowed_ips: *const c_char,
     endpoint: *const c_char,
 ) -> telio_result {
+    telio_log_info!(
```
gindoslaw, line 742: Do we need this?

src/ffi/mod.rs
```diff
@@ -973,7 +998,7 @@ pub extern "C" fn telio_get_commit_sha() -> *mut c_char {
 
 #[no_mangle]
 pub extern "C" fn telio_get_status_map(dev: &telio) -> *mut c_char {
-    trace!("acquiring dev lock");
+    telio_log_info!("telio_get_status_map entry with instance id: {}.", dev.id);
```
gindoslaw, line 1001: how often it is going to be executed?

src/ffi/mod.rs
```diff
@@ -567,10 +587,6 @@ pub extern "C" fn telio_connect_to_exit_node(
     allowed_ips: *const c_char,
     endpoint: *const c_char,
 ) -> telio_result {
-    telio_log_info!(
```
gindoslaw, line 570: Is there a reason to move it?

src/ffi/mod.rs
```diff
@@ -499,6 +517,8 @@ pub extern "C" fn telio_set_private_key(dev: &telio, private_key: *const c_char)
 
 #[no_mangle]
 pub extern "C" fn telio_get_private_key(dev: &telio) -> *mut c_char {
+    telio_log_info!("telio_get_private_key entry with instance id: {}", dev.id);
```
gindoslaw, line 520: There is not __func__ or something like that in rust?

crates/telio-dns/src/nameserver.rs
```diff
@@ -157,7 +157,11 @@ impl LocalNameServer {
                         {
                             Ok(length) => length,
                             Err(e) => {
-                                telio_log_error!("[DNS] {}", e);
+                                telio_log_error!(
+                                    "[DNS] {}. Offending request packet: {:?}",
+                                    e,
+                                    packet
```
gindoslaw, line 163: what is most pessimistic scenario in the context of packet size (length) and how often it can be called?

crates/telio-dns/src/forward.rs
```diff
@@ -218,6 +218,10 @@ impl Authority for ForwardAuthority {
         let name: LowerName = name.clone();
         let resolve = self.resolver.lookup(name, rtype).await;
 
+        if resolve.is_err() {
+            telio_log_warn!("DNS name resolution failed with {:?}", resolve);
```
gindoslaw, line 222: will this contain server + error message?

src/ffi/mod.rs
Jauler: We would like to log all FFI interactions in user logs, especially state-changing interactions, because this is first line of "debugging", when receiving logs. When receiving logs, it is providing the scenario on what kind of "scenario" was running. Therefore yes, PQ is needed here as well.

src/ffi/mod.rs
Jauler: Every time when `get_status_map()` is called. But this method is in fact used only on linux when executing `nordvpn mesh peer list` command.

src/ffi/mod.rs
Jauler: `telio_connect_to_exit_node()` is only a wrapper around `telio_connect_to_exit_node_with_id()`. Previously only `telio_connect_to_exit_node()` was logged, and we were unable to see any calls to `telio_connect_to_exit_node_with_id()`. That method is being used on android, meaning, that we were unable to see the fact that android app is initiating VPN connection. First debugging attempt lead us to the fact, that DNS is enabled, but VPN is not. Later it turned out that only log line is missing.

src/ffi/mod.rs
Jauler: there are smarter methods to log this (and in fact it is done with UniFFI), but on v4.3.3 "manual" method has been used, and I do not feel like now is the time to migrate to another method of logging FFI in v4.3 line.

crates/telio-dns/src/nameserver.rs
Jauler: worst case scenario - 1420 bytes, called on every DNS request. It is rather verbose, but we are specifically trying to why this error is occurring.

crates/telio-dns/src/forward.rs
Jauler: This is an example of the error message, when routing table does not have route towards the DNS server:
```
2024-06-28T08:49:33.384834Z  WARN telio_dns::forward: 222: DNS name resolution failed with Err(ResolveError { kind: Proto(ProtoError { kind: Io(Os { code: 113, kind: HostUnreachable, message: "No route to host" }) }) })
```

src/ffi/mod.rs
```diff
@@ -391,6 +391,15 @@ pub extern "C" fn telio_start_named(
     adapter: telio_adapter_type,
     name: *const c_char,
 ) -> telio_result {
+    let sk = ffi_try!(char_ptr_to_type::<SecretKey>(private_key));
```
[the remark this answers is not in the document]
Jauler: >  How log this variable can survive in the memory?

Process running libtelio code, as long as it is running, has a copy of private key in memory. There is no other way to process WireGuard handshakes.

> After release - will it be auto-magically obfuscated? (so can this be part of the memory returned to operating system and used by different process and in fact read by a different process - so in fact reviled with using different process)

No operating system which we are supporting allows for such sharing of memory. Any new pages allocated to the process are always zeroed at first allocation. Note that after freeing and allocating same pages again in the same process - it is not required to be zeroed. Having ability to read different process memory (for non-privileged processes) is considered to be a big security issue.

crates/telio-dns/src/forward.rs
Jauler: Upstream servers are configured over FFI layer, where it may be seen.

src/ffi/mod.rs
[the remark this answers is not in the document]
gindoslaw: It is not about sharing memory. It is about use after free. But once it is always zeroed - it is fine.

tomaszklak (approved):

gindoslaw (approved):

PR #646, opened 2024-06-28
LLT-253: Enable connected socket for Apple platforms

tomasz-grz:
### Problem
Using connected sockets can boost performance in direct connections. But for unknown reasons in the past it was causing problems on Apple platforms (iOS/MacOS) and packets were not being transmitted.

After testing more recent versions of libtelio and BoringTun this no longer appears to be the case. Moreover, with the tested libtelio `v4.3.3`, enabling connected sockets almost doubles the download throughput.

### Solution
- Enable connected sockets by default on all platforms.
- Add a feature/remote config to disable connected sockets, in case needed.

- [x] README.md is updated


nat-lab/tests/test_mesh_connected_socket.py
```diff
@@ -0,0 +1,103 @@
...
+                SetupParameters(
+                    connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_1,
+                    adapter_type=telio.AdapterType.BoringTun,
+                    features=features_alpha,
+                ),
+                SetupParameters(
+                    connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_2,
+                    adapter_type=telio.AdapterType.BoringTun,
+                    features=features_beta,
+                ),
```
tomaszklak, line 47: The PR enables connected sockets on apple platforms - in that case it would be more appropriate to use mac vm instead of linux docker as node(s).

nat-lab/tests/test_mesh_connected_socket.py
```diff
@@ -0,0 +1,103 @@
...
+            ),
+        )
+
+        async with Ping(connection_alpha, beta.ip_addresses[0]).run() as ping:
+            await ping.wait_for_next_ping()
+        async with Ping(connection_beta, alpha.ip_addresses[0]).run() as ping:
+            await ping.wait_for_next_ping()
+
+        async with IperfServer(connection_alpha, "server").run():
+            # await asyncio.sleep(1)
```
tomaszklak, line 72: remove :)

nat-lab/tests/test_mesh_connected_socket.py
tomasz-grz: Good point, we only have one VM though, but still would be good to test the combination.

nat-lab/tests/test_mesh_connected_socket.py
```diff
@@ -0,0 +1,105 @@
...
+        ]
+
+        asyncio.gather(
+            client_alpha.wait_for_state_peer(
+                beta.public_key, [State.Connected], [PathType.Direct]
+            ),
+            client_beta.wait_for_state_peer(
+                alpha.public_key, [State.Connected], [PathType.Direct]
+            ),
+        )
```
mathiaspeters, line 68: This is already done by `setup_mesh_nodes` since direct connections are enabled

mathiaspeters:
+1

nat-lab/tests/test_mesh_connected_socket.py
tomasz-grz: Good point, I'll simplify it, thanks!

.unreleased/LLT-253
Jauler, line 1: IMHO, these changes are relevant to the app devs, meaning that we should add at least some sort of message in changelog :).

nat-lab/tests/test_mesh_connected_socket.py
```diff
@@ -0,0 +1,95 @@
...
+@pytest.mark.parametrize(
+    "beta_tag",
+    [ConnectionTag.MAC_VM, ConnectionTag.DOCKER_OPEN_INTERNET_CLIENT_2],
+)
+async def test_mesh_connected_socket(
+    disable_connected_socket_a, disable_connected_socket_b, beta_tag
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        features_alpha = TelioFeatures(
+            direct=Direct(providers=["stun", "local"]),
```
Jauler, line 32: Here it probably make sense to leave `Direct()` and avoid overriding defaults for `providers`. As the test does not depend on specific providers, it just needs direct connection. Which will be ensured if `direct is not None`.

tomaszklak (approved):
+1

tomasz-grz:
Unfortunately due to the changes in BoringTun v1.2.x it is not advisable to enable connected sockets on Apple platforms, therefore this task is cancelled.

PR #647, opened 2024-06-28
LLT-5358: Let Pyro5 select port automatically

mathiaspeters:
### Problem
We occasionally have problems in CI that the socket we ask Pyro5 to bind is busy, causing the pipeline to fail even though nothing is wrong with the code. This problem has only happened with docker containers in CI so far, so it's only fixed for that scenario (not just for CI, but any system running linux as the host OS) but can be expanded to other platforms if needed

### Solution
Pyro5 can automatically select an available port and we can find out about the port afterwards

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -4,7 +4,6 @@ import asyncio
 import json
 import os
 import platform
-import random
 import re
 import shlex
 import uniffi.telio_bindings as libtelio  # type: ignore
@@ -521,6 +520,7 @@ class Client:
         self._quit = False
         self._start_time = datetime.now()
         self._libtelio_proxy: Optional[LibtelioProxy] = None
+        self._proxy_port = ""
         # Automatically enables IPv6 feature when the IPv6 stack is enabled
         if (
             self._node.ip_stack in (IPStack.IPv4v6, IPStack.IPv6)
@@ -546,6 +546,8 @@ class Client:
                 "task started - ",
             ]
             for line in stdout.splitlines():
+                if line.startswith("libtelio-port:"):
+                    self._proxy_port = line[len("libtelio-port:") :]
                 if not any(string in line for string in supress_print_list):
                     print(f"[{self._node.name}]: stdout: {line}")
                 if self._runtime:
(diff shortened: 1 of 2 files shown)


Hasan6979:
+1

nat-lab/tests/telio.py
```diff
@@ -625,6 +626,37 @@ async def on_stderr(stdout: str) -> None:
...
+                            r"([0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}:[0-9]{1,5}).*\s"
+                            + pid
+                            + "/python"
+                        )
+                        sockets = re.findall(port_regex, fds)
+                        if len(sockets) == 0:
+                            print(
+                                f"Pyro5 on node {self._node.name} has not bound any sockets"
+                            )
+                        object_uri = f"PYRO:{object_name}@{sockets[0]}"
```
Jauler, line 656: This method of figuring out listening port very heavily relies that the process creates only single socket, at least at the time of querying. While it is probably true at the time, it introduces this "hidden" requirement for `libtelio_remote.py` which is very easy to break in non-trivial ways :/.

IMHO it would be much nicer if `libtelio_remote.py` would report listening port chosen by itself, or would use a range which is safe(r) to use (e.g. 0-1024), as OS (in nat-lab configuration) is guaranteed not to select those ports inside of containers.

Jauler:
In addition, there seems to be an issue due to a fact port usage is separated inside of docker containers and host, meaning, that port selected for pyro5 server by the OS is guaranteed to be free inside of container, but **not** guaranteed to be free on docker host, thus exposing it will cause the same errors we have been observing.

Jauler (changes requested):
In addition, there seems to be an issue due to a fact port usage is separated inside of docker containers and host, meaning, that port selected for pyro5 server by the OS is guaranteed to be free inside of container, but not guaranteed to be free on docker host, thus exposing it will cause the same errors we have been observing.

tomasz-grz (commented):
+0.5

olekoliinyk (approved):
+0.5

olekoliinyk (approved):
+1.0

tomasz-grz (dismissed):
+1.0 🚀

Jauler (approved):
+1.0

PR #648, opened 2024-07-01
[LLT-5366] Update requests dependency due to BDSA-2024-2985

Jauler:
### Problem
BDSA-2024-2985 was reported for psf-request, Update should be easy, so doing it in this PR.

- [x] README.md is updated

### DIFF
--- a/nat-lab/Pipfile
+++ b/nat-lab/Pipfile
@@ -44,6 +44,7 @@ types-protobuf = "==3.19.21"
 typing-extensions = "*"
 typing-inspect = "*"
 yarl = "*"
+requests = "*"
 
 [dev-packages]
 
--- a/nat-lab/Pipfile.lock
+++ b/nat-lab/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "4a95b7874669bcd26a8cf9f906bc6f38bae84cc4e80798f04c42d1b9f15dcfd2"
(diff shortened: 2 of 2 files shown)


Jauler:
This PR will also have to add a reference to new tag of libtelio-build.

Jauler:
So after fixing ports to some non-ephemeral range, I realised that there is another problem, as seems like Pyro5 does not se SO_REUSEADDR or SO_REUSEPORT (well why should it without any additional configs), and in server shutdown may happen "in any direction" leaving TCP connection in those `TIME_WAIT` or state, which causes next tests to fail to open sockets. 

Will have to think about the approach, on solving this. Can we start using SO_REUSEADDR? :thinking: As it would be appropriate in this context.

tomaszklak (dismissed):
+2

stalowyjez:
+1

stalowyjez (approved):

.gitlab-ci.yml
```diff
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v1.2.5
+    branch: v1.2.5
```
stalowyjez, line 18: Why these newlines are removed?

Jauler: Oh it is a leftover after conflict resolution. I will fix it.

Jauler: fixed

stalowyjez (approved):
+2

PR #652, opened 2024-07-02
Fix MacOS python-requests install as brew does not support it anymore

Jauler:
### Problem
`brew install` stopped working for python-requests. Use pip instead.


olekoliinyk (approved):
+2.0

LukasPukenis (commented):
+1.5

tomaszklak (approved):

Jauler:
My first attempt: https://github.com/NordSecurity/libtelio/commit/399b49cf0adf6546f7f6a0146eaddd21b8dc390f, was by simply using `--break-system-packages`. But I do not really like this solution, therefore I moved to virtual envs, even if it is slightly more complex. The reason is that pip considers `requests` to be a package which clashes with externally-managed version of it, and does not allow to mix pip-managed and externally-managed packages  by default. One can of course override this check with `--break-system-packages`, but I have had experience of weird behavior happening as a result.

Therefore using virtual environment seems safer to me.

PR #653, opened 2024-07-02
[LLT-5358] Avoid using ephemeral ports for libtelio_remote.py

Jauler:
### Problem
Currently listening port for pyro5 server is randomly selected by `pytest` process without any regard for ports already in use. This kind of selection is inherently flaky, and will fail to startup if the port is already in use on either docker host or inside of container.

There has been an attempt to fix this issue: https://github.com/NordSecurity/libtelio/pull/647, but IMHO, it still leaves racy port selection, and adds some non-obvious requirements on `libtelio_remote.py`. 

I believe there is a simpler approach, by simply using non-ephemeral port ranges. 

- [x] README.md is updated

### DIFF
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -72,7 +72,7 @@ services:
       cone-net-01:
         ipv4_address: 192.168.101.104
     ports:
-      - "58001:58001"
+      - "8080:8080"
     environment:
       CLIENT_GATEWAY_PRIMARY: 192.168.101.254
     extra_hosts:
@@ -92,7 +92,7 @@ services:
       cone-net-02:
         ipv4_address: 192.168.102.54
     ports:
-      - "58002:58002"
+      - "8081:8081"
     environment:
       CLIENT_GATEWAY_PRIMARY: 192.168.102.254
   # Client that shares two cone networks on different interfaces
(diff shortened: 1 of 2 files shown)


gytsto (commented):
looks good +1

Jauler:
So after fixing ports to some non-ephemeral range, I realised that there is another problem, as seems like Pyro5 does not se SO_REUSEADDR or SO_REUSEPORT (well why should it without any additional configs), and in server shutdown may happen "in any direction" leaving TCP connection in those TIME_WAIT or state, which causes next tests to fail to open sockets.

Will have to think about the approach, on solving this. Can we start using SO_REUSEADDR? 🤔 As it would be appropriate in this context.

dfetti (commented):
+1.0

mathiaspeters:
> Will have to think about the approach, on solving this. Can we start using SO_REUSEADDR? 🤔 As it would be appropriate in this context.

Me and Povilas was looking into this in the past and I think we came to the conclusion that Pyro5 is using `SO_REUSEADDR` by default, but not `SO_REUSEPORT`

mathiaspeters:
And I would say, this PR is also quote closely related to #644

Jauler:
Closing this one due to https://github.com/NordSecurity/libtelio/pull/644

PR #654, opened 2024-07-02
LLT-5211 Wait for drivers loaded in Windows

matislovas:
### Problem
In Windows, we're sometimes seeing the errors: `Telio::start_named: Err(Adapter(WireguardGo(FailedToStart(-1))))` (the same with WireguardNT adapter). Primary suspect is the `*.dll` loader mechanism, which returns, but still haven't fully initialized the library (error happens only on first library load).

### Solution
Add windows service checker for `Wintun` and `WireGuard` services.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/windows/service.rs
+++ b/crates/telio-wg/src/windows/service.rs
@@ -1,0 +1,93 @@
+#![cfg(windows)]
+
+use std::convert::{TryFrom, TryInto};
+use std::time::{Duration, Instant};
+use telio_utils::{telio_log_debug, telio_log_warn};
+use windows::{
+    core::{w, Error, Result as WindowsResult, HRESULT, PCWSTR},
+    Win32::{
+        Security::SC_HANDLE,
+        System::Services::{
+            OpenSCManagerW, OpenServiceW, QueryServiceStatus, SC_MANAGER_CONNECT,
+            SERVICE_QUERY_STATUS, SERVICE_STATUS,
+        },
+    },
+};
+
+pub const DEFAULT_SERVICE_WAIT_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[derive(Debug)]
+enum WinServiceStatus {
+    Stopped = 1,
+    Starting = 2,
+    Stopping = 3,
+    Running = 4,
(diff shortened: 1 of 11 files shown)


crates/telio-wg/src/windows/service.rs
```diff
@@ -0,0 +1,93 @@
...
+            SERVICE_QUERY_STATUS, SERVICE_STATUS,
+        },
+    },
+};
+
+pub const DEFAULT_SERVICE_WAIT_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[derive(Debug)]
+enum WinServiceStatus {
+    Stopped = 1,
```
gytsto, line 21: would be nice to get a comment where the values are coming from :smile:

matislovas: They're taken from [here](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_status). I did not find an enum in `windows-rs` crate, so made one from there.

gytsto: isn't this the one -> https://docs.rs/windows-service/latest/windows_service/service/enum.ServiceState.html

gytsto: ah but not in `windows-rs`

matislovas: Yeah the same, but that's a different crate, and I though siding with our plain implementation

gytsto: anyway, good enough :smile:

LukasPukenis (commented):
+1. It makes sense what it does, but it 's hard to validate that it does so correctly without a testcase :)

matislovas:
> +1. It makes sense what it does, but it 's hard to validate that it does so correctly without a testcase :)

Yeah, will add it, but as I imagine, it will still be flaky. I can un-load the driver in windows, and wait for successful `tcli` startup. But even if my service monitor will not work, the test won't fail in all cases (will be flaky).

matislovas:
P.S. test is added

gytsto (approved):
+1

PR #655, opened 2024-07-02
Improve error handling for case when PingerClient creation fails

tomaszklak:
### Problem
It's not 100% clear why surge ping fails to create a socket on windows for pinging.

### Solution
Surge returns an `io::Result` so not much we can do about it but maybe this only happens for one kind of IP (4/6) and this change will show this.

- [x] README.md is updated


dfetti (approved):
+2.0

PR #661, opened 2024-07-05
Llt 5309 integrate multicast components

ThrasherLT:
### Problem
All the separate parts of starcast (see RFC LLT-0060) are ready to be integrated into Telio.

### Solution
This PR integrates all the separate starcast parts (virtual peer, transport and nat) into libtelio.
You can see the doc rs in code of every separate component to understand how they're supposed to be integrated into telio. 

The Nat-Lab test for multicast is being implemented with a separate PR, so this one is untested yet.

- [x] README.md is updated

### DIFF
--- a/crates/telio-starcast/src/transport.rs
+++ b/crates/telio-starcast/src/transport.rs
@@ -17,9 +17,13 @@ use telio_task::{
 };
 use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn};
 
+use telio_model::constants::{
+    IPV4_MULTICAST_NETWORK, IPV4_STARCAST_ADDRESS, IPV6_MULTICAST_NETWORK, IPV6_STARCAST_ADDRESS,
+};
+
 use crate::{
     nat::{self, Nat, StarcastNat},
-    utils::{DualIpAddr, MutableIpPacket},
+    utils::MutableIpPacket,
 };
 
 #[cfg(feature = "test-util")]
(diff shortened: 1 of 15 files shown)


LukasPukenis:
As a reviewer I can't effectively review because while the PR description tells what this PR does(I understand words), it doesn't add any context, thus `starcast` for me is just a buzzword :/ It would really help to have more information what it is

mathiaspeters:
>  thus `starcast` for me is just a buzzword :/

In a nutshell, starcast (*cast was the working name to cover multicast and broadcast as one word, and it could be pronounced as starcast if you say it as an actual word) is the meshnet implementation of multicast. I guess this is a general problem with "integration PRs" where you need to understand the components that are being integrated to properly understand and review the integration work. For more background you can check RFCs LLT-0060 and LLT-0047, or if it makes more sense, this could be a candidate for a review meeting or something like that to make it easier for you and to share some knowledge

crates/telio-starcast/src/transport.rs
packgron, line 1: We recently switched to ipnet, so we should do that here also

crates/telio-starcast/src/lib.rs
```diff
@@ -5,6 +5,15 @@
 //! 2. Starcast transport component responsible for multicasting intercepted traffic to the meshnet peers and handling response messages.
 //! 3. Nat is transport's utility used to SNAT the packets comming from multiple peers into one multicast peer address.
 
+use const_format::formatcp;
+
+pub(crate) const IPV4_MULTICAST_NETWORK: &str = "224.0.0.0/4";
```
packgron, line 10: We had these constant in telio-model::constants, i would prefer to keep than grouped there so it would be clear what global ip's are reserved

crates/telio-starcast/src/lib.rs
```diff
@@ -5,6 +5,15 @@
...
+pub(crate) const IPV6_MULTICAST_NETWORK: &str = "ff02::fb/128";
```
packgron, line 11: NIT: as IpNet and IpAddr have constatnts maybe we should construct them in place and avoid parsing in runtime

packgron (commented):
Few nit's

crates/telio-starcast/src/lib.rs
jjanowsk: That would be great and I wanted to do this, unfortunately even though those APIs are const they return Result and unwrap is not const. So yeah, we can in theory keep those as `Result<Ipv4Net, PrefixLenError>`, but still we need unwraps in our code so I've decided that strs are cleaner and parsing is not gonna cost us a lot

jjanowsk: I've create an issue in ipnet repo: https://github.com/krisprice/ipnet/issues/55 and two PRs: https://github.com/krisprice/ipnet/pull/56 https://github.com/krisprice/ipnet/pull/57 that could allow switch to const IpvXNet items in the future.

mathiaspeters:
+1

crates/telio-starcast/src/lib.rs
jjanowsk: Also for now I've created the ConstIpvXNet wrappers in our utils

packgron (approved):
+1.0

packgron (dismissed):

packgron (approved):

packgron (approved):

PR #666, opened 2024-07-08
Do not copy binaries on every test to Mac and Win VMs

matislovas:
### Problem
During `nat-lab` test runs, binaries are copied to VMs on every new `connection` object creation (meaning - on every test run, involving MAC or WIN VM).

### Solution
Copy the binaries to VM only once per `pytest` session.


mathiaspeters (dismissed):
Seems devilishly reasonable, +1 :smiling_imp:

nat-lab/tests/utils/vm/mac_vm_util.py
```diff
@@ -28,21 +30,22 @@ async def new_connection(ip: str = config.MAC_VM_IP) -> AsyncIterator[Connection
...
+            ip,
+            username="root",
+            password="vagrant",  # NOTE: this is hardcoded password for transient vm existing only during the tests
+            known_hosts=None,
+            options=ssh_options,
+        ) as ssh_connection:
+            connection = SshConnection(ssh_connection, TargetOS.Mac)
 
-        await _copy_binaries(ssh_connection, connection)
+            await _copy_binaries(ssh_connection, connection)
```
tomasz-grz, line 43: What happens if `_copy_binaries()` fails the first time (like with those time outs)? Do we need to check for that, or would the whole test fail?

nat-lab/tests/utils/vm/windows_vm_util.py
```diff
@@ -34,21 +36,22 @@ async def new_connection(ip: str = WINDOWS_1_VM_IP) -> AsyncIterator[Connection]
...
-    ) as ssh_connection:
-        connection = SshConnection(ssh_connection, TargetOS.Windows)
-
-        await _copy_binaries(ssh_connection, connection)
-
-        try:
-            yield connection
-        finally:
-            pass
+    if copy_binaries:
```
tomasz-grz, line 39: What happens if `copy_binaries` is False? would this still return a `Connection`?

tomasz-grz (commented):
+0.5

nat-lab/tests/utils/vm/mac_vm_util.py
matislovas: I've changed a bit whole concept, but yeah, if copying binaries fails to one of the VMs - the pytest session fails.

nat-lab/tests/utils/vm/windows_vm_util.py
matislovas: Changed to pytest hook function, `new_connection` always yields a `Connection`

gytsto (approved):
+2 this is what I had in mind. Perfect execution!

PR #670, opened 2024-07-09
LLT-5133 Local ICMP Request Packet Batching in Direct Connections

LukasPukenis:
### Problem
According to empirical testing and some research it appears that sending more data at once is beneficial to battery instead of sending sporadically.

Additionally, WireGuard's persistent-keepalives were used for direct connections when SessionKeeper was already used, one being redundant.

### Solution
Introduce Batcher entity that registers intervals and thresholds and then sends signals when batching should be done. To observe if it works, NatLab tests check for time deltas after capturing the traffic with `tcpdump` in specified nodes.

Additionally, persistent-keepalive field is set to 0 for direct connections

- [ ] README.md is updated


crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,526 @@
...
+                let mut batched_count = 0;
+                let mut keys_to_delete = vec![];
+
+                {
+                    for (key, &mut ref mut entry) in self.notifiers.iter_mut() {
+                        let deadline_with_threshold = entry.deadline - entry.threshold;
+                        if now >= deadline_with_threshold {
+                            batched_count +=1;
+                            telio_log_debug!("batcher inside tokio::select! Batch!, will add new sleep with interval {:?} @ now {:?}", entry.interval, now);
+                            entry.deadline = now + entry.interval;
```
LukasPukenis, line 99: do i even need a key?

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,172 @@
...
+                    20, 200, ObservationTarget(Direction.Outgoing, mesh_ip)
+                )
+                
+                print(">>>>>>> batching results vvv: ", gw._name)
+                print("hs_inc:", hs_inc)
+                print("hs_out", hs_out)
+                print(">>>>>>> batching results ^^^")
+
+        env = await exit_stack.enter_async_context(
+            setup_environment(exit_stack, setup_params)
```
LukasPukenis, line 126: Better to misalign here

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,526 @@
...
+
```
LukasPukenis, line 100: It's apparent that maybe SessionKeeper could be responsible for all of the ICMP ECHO requests/pings that are being done by persistent_keepalive feature of WireGuard. Given that it looks like the case, it might be a much cleaner approach to:
- move batcher into session keeper
- restore repeated_actions, rename to batcher and modify to be able to work with thresholds.

This way modifications to the code would be kept at a minimum and no future-proofing would be there as it might be unecessary.

crates/telio-batcher/src/batcher.rs
LukasPukenis: The key is not necessary, except for being able to delete the entry in the Batcher. A slightly better mechanism could be to return a cancellation token that upon calling would delete the entry, as the knowledge of key serves no other purpose at the moment.

nat-lab/tests/test_batching.py
LukasPukenis: Misalignment now is done on initialization by creating 2 meshnet peers and then adding another one after sleeping for a small amount of time.

nat-lab/docker-compose.yml
```diff
@@ -51,6 +51,8 @@ services:
     cap_drop:
       - ALL
     cap_add:
+      - CAP_DAC_OVERRIDE
+      - CAP_CHOWN
```
LukasPukenis, line 55: These two are used in order to make tcpdump work, as else it will fail to create packet capture files

PR #671, opened 2024-07-09
[LLT-5393] Collect coredumps

tomasz-grz:
### Problem

Sometimes the tests crash with a segmentation fault, but we do not have core-dumps to analyze the root cause.

### Solution

Enable, collect and attach core dumps to the CI runner artifacts.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -534,6 +534,7 @@ class Client:
     ) -> AsyncIterator["Client"]:
         if isinstance(self._connection, DockerConnection):
             start_tcpdump(self._connection.container_name())
+            await self.clear_core_dumps()
 
         async def on_stdout(stdout: str) -> None:
             supress_print_list = [
@@ -658,9 +659,13 @@ class Client:
                     print(datetime.now(), "Test cleanup stage 1. Saving logs")
                     await self.save_logs()
 
-                    print(datetime.now(), "Test cleanup stage 2. Stopping tcpdump")
+                    print(
+                        datetime.now(),
+                        "Test cleanup stage 2. Stopping tcpdump and collecting core dumps",
+                    )
                     if isinstance(self._connection, DockerConnection):
                         stop_tcpdump([self._connection.container_name()])
+                        await self.collect_core_dumps()
 
                     print(
                         datetime.now(),
@@ -1226,6 +1231,64 @@ class Client:
         ev = self.wait_for_output(f"PMTU -> {host}: {expected}")
         await self._write_command(["pmtu", host])
         await ev.wait()
+
+    # This is where natlab expects coredumps to be placed
+    # For CI and our internal linux VM, this path is set in our provisioning scripts
+    # If you're running locally without the aforementioned linux VM, you are expected to configure this yourself
+    # However, this only needs to be set iff you're:
+    #  - running a test targeting a docker image
+    #  - have set the NATLAB_SAVE_LOGS environment variable
+    #  - want to have natlab automatically collect core dumps for you
+    def get_coredump_folder(self) -> tuple[str, str]:
+        return "/var/crash", "core-"
+
+    def should_skip_core_dump_collection(self) -> bool:
+        return (
+            os.environ.get("NATLAB_SAVE_LOGS") is None
+            or self._connection.target_os != TargetOS.Linux
+        )
+
+    async def clear_core_dumps(self):
+        if self.should_skip_core_dump_collection():
+            return
+
+        coredump_folder, _ = self.get_coredump_folder()
+
+        # clear the existing system core dumps
+        await self._connection.create_process(["rm", "-rf", coredump_folder]).execute()
+        # make sure we have the path where the new cores will be dumped
+        await self._connection.create_process(
+            ["mkdir", "-p", coredump_folder]
+        ).execute()
+
+    async def collect_core_dumps(self):
+        if self.should_skip_core_dump_collection():
+            return
+
+        coredump_folder, file_prefix = self.get_coredump_folder()
+
+        # find all core dump files
+        process = await self._connection.create_process(
+            ["find", coredump_folder, "-maxdepth", "1", "-name", f"{file_prefix}*"]
+        ).execute()
+        dump_files = process.get_stdout().strip().split()
+
+        coredump_dir = "coredumps"
+        os.makedirs(coredump_dir, exist_ok=True)
+
+        should_copy_coredumps = len(dump_files) > 0
+
+        # if we collected some core dumps, copy them
+        if isinstance(self._connection, DockerConnection) and should_copy_coredumps:
+            container_name = self._connection.container_name()
+            test_name = test_name_safe_for_file_name()
+            for i, file_path in enumerate(dump_files):
+                file_name = file_path.rsplit("/", 1)[-1]
+                core_dump_destination = (
+                    f"{coredump_dir}/{test_name}_{file_name}_{i}.core"
+                )
+                cmd = f"docker container cp {container_name}:{file_path} {core_dump_destination}"
+                os.system(cmd)
 
 
 def generate_secret_key() -> str:
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v1.2.5
+          triggered-ref: v1.2.7
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -73,6 +73,10 @@ services:
         ipv4_address: 192.168.101.104
     ports:
       - "8081"
+    ulimits:
+      core:
+        soft: -1
+        hard: -1
     environment:
       CLIENT_GATEWAY_PRIMARY: 192.168.101.254
     extra_hosts:


nat-lab/run_local.py
```diff
@@ -156,6 +156,7 @@ def get_pytest_arguments(options) -> List[str]:
         if options.moose:
             marks = marks.replace("and not moose", "")
         args.extend(["-m", marks])
+    args.extend(["-k", "test_coredump"])
```
tomasz-grz, line 159: Ignore this for now, I'll remove later once all review change requests are done and tested.

nat-lab/tests/telio.py
```diff
@@ -1211,6 +1213,82 @@ async def probe_pmtu(self, host: str, expected: int):
...
+        # if there is only one element it means that there is no path present in pattern
+        if len(core_pattern) > 1:
+            path = core_pattern[0]
+        # split common patterns like 'core-%h-%e-%s-%p'
+        file_name = core_pattern[-1].split("%", 1)[0]
+
+        return path, file_name
+
+    async def clear_core_dumps(self):
+        if self._connection.target_os == TargetOS.Linux:
```
mathiaspeters, line 1232: To be consistent with `collect_core_dumps`, this should be an early return

nat-lab/tests/telio.py
```diff
@@ -534,6 +534,7 @@ async def run(
     ) -> AsyncIterator["Client"]:
         if isinstance(self._connection, DockerConnection):
             start_tcpdump(self._connection.container_name())
+        await self.clear_core_dumps()
```
LukasPukenis, line 537: Maybe it should go to the call site where `start_tcpdump` is being called? This implies that `start_tcpdump` is the starting point of the container which doesn't _feel_ like it(might be wrong) :)

LukasPukenis (commented):
I have several questions regarding the solution. It appears as if `start_tcpdump()` initiates coredump collection, however coredumps will be only present _after_ the crash and not before(implied because of `start_tcpdump()`). After docker containers are stopped, changes are discarded, so coredumps should also be discarded unless the directory is mounted on the host.

TLDR: we should collect coredumps and copy them at the end of natlab, while the containers are still present. :thinking:

nat-lab/tests/telio.py
tomasz-grz: You mean inside the if statement like this? 
```python
        if isinstance(self._connection, DockerConnection):
            start_tcpdump(self._connection.container_name())
            await self.clear_core_dumps()
```

tomasz-grz: Hi, thanks for taking a look, I'm not sure I understand correctly what you mean, I tried to follow the pattern of `tcpdump()` or collecting logs.

Where each test case would have a core dump, if one was created during the duration of each individual test. Of course if the container itself crashes or maybe `libtelio_remote.py`, we won't be able to collect core dumps in that case, but the same goes for logs and pcaps. 

`clear_core_dumps()` only deletes the old core dumps at the beginning of the test, so that we know if any present, were created during the duration of the test case.

the core dumps themselves are collected in `collect_core_dumps()` here

```python
finally:
                    await self.save_logs()
                    await self.collect_core_dumps()
```

nat-lab/tests/telio.py
```diff
@@ -534,6 +534,7 @@ async def run(
     ) -> AsyncIterator["Client"]:
         if isinstance(self._connection, DockerConnection):
             start_tcpdump(self._connection.container_name())
+            await self.clear_core_dumps()
```
tomasz-grz, line 537: Actually I'm wondering if it's better to clear core dumps only after collecting them,
not before the test run, in case if something crashed in the previous run and we failed to collect them..

Jauler: It is common to either do both, or just at the start. The idea being, that in case things are crashing -> we cannot be sure that cleanup will be executed reliably.

Of course in this case, the crashing process and the one collecting dumps are two different ones, so the idea above is more rule of thumb rather then something strong.

nat-lab/tests/telio.py
```diff
@@ -1211,6 +1213,84 @@ async def probe_pmtu(self, host: str, expected: int):
         await self._write_command(["pmtu", host])
         await ev.wait()
 
+    async def get_coredump_pattern(self) -> tuple[str, str]:
```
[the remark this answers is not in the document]
Jauler: Inisde of CI we are running in our own VM, therefore, maybe it would make sense to simply change the `core_pattern` to something we control?

This way we can control exactly the pattern, and there will be no need to parse it. IMHO it is simpler and more reliable as well.

And when running the tests outside of CI -> we probably can skip coredump collection, as this is something, the one who is running tests can control and take directly from containers.

nat-lab/tests/telio.py
tomasz-grz: Good point I was thinking along those lines too. I think I will keep it at the beginning for now, In case we see that something is crashing before and we need those logs, we can change it after.

Jauler: Oh, I just realised that, if we do cleanup non-conditionally (whether it is being run in CI or not) it helps to avoid cleanup after test failure when running outside of CI, as it allows for devs to gather the coredumps manually from containers.

nat-lab/tests/telio.py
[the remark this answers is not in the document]
tomasz-grz: Good point, I've seen that on my other system, unfortunately it's not very helpful if it's not creating a file though. How would you handle that case?

nat-lab/tests/telio.py
```diff
@@ -1211,6 +1213,84 @@ async def probe_pmtu(self, host: str, expected: int):
...
+                ["rm", "-f", f"{file_name}*"]
+            ).execute()
+
+    async def collect_core_dumps(self):
+        if (os.environ.get("NATLAB_SAVE_LOGS") is None) or (
+            self._connection.target_os != TargetOS.Linux
+        ):
+            return
+
+        print("Collecting core dumps")
```
Jauler, line 1258: Could you add timestamp to the log? It helps when analyzing logs later.

nat-lab/tests/telio.py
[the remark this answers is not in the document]
Jauler: > Good point, I've seen that on my other system, unfortunately it's not very helpful if it's not creating a file though. How would you handle that case?

IMHO, we can control `core_pattern` inside of CI and set it to something convenient for us to use. And leave it up to devs outside of CI.

nat-lab/tests/telio.py
tomasz-grz: Good point, I'll add `if (os.environ.get("NATLAB_SAVE_LOGS") is None)` here too

nat-lab/tests/telio.py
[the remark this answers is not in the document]
tomasz-grz: TLDR: we can not change the `core_pattern` from within the tests or docker images because of permission issues `sysctl: permission denied on key "kernel.core_pattern"`. Will have another PR for the CI runners.

nat-lab/tests/telio.py
```diff
@@ -1211,6 +1213,83 @@ async def probe_pmtu(self, host: str, expected: int):
...
+            return
+        path, file_prefix = await self.get_coredump_pattern()
+
+        process = await self._connection.create_process(
+            ["find", path, "-maxdepth", "1", "-name", f"{file_prefix}*"]
+        ).execute()
+        files = process.get_stdout().strip().split()
+
+        process = self._connection.create_process(["cat", "/etc/hostname"])
+        await process.execute()
```
mathiaspeters, line 1267: Nit: another consistency thing, other lines create and execute the process on the same line but here you have it on separate lines

nat-lab/tests/telio.py
```diff
@@ -1211,6 +1213,83 @@ async def probe_pmtu(self, host: str, expected: int):
...
+            path = core_pattern[0]
+        # remove the filename flags 'core-%h-%e-%s-%p'
+        file_name = core_pattern[-1].split("%", 1)[0]
+
+        return path, file_name
+
+    async def clear_core_dumps(self):
+        if (os.environ.get("NATLAB_SAVE_LOGS") is None) or (
+            self._connection.target_os != TargetOS.Linux
+        ):
```
mathiaspeters, line 1236: Nit: I'm kinda surprised black doesn't complain about it but I think the unnecessary parentheses here makes the conditional more complicated to read

mathiaspeters:
+1

nat-lab/tests/telio.py
tomasz-grz: The alternative that `black` suggests is 
```python
        if (
            os.environ.get("NATLAB_SAVE_LOGS") is None
            or self._connection.target_os != TargetOS.Linux
        ):
```

Which is slightly more readable

nat-lab/run_local.py
```diff
@@ -157,6 +157,8 @@ def get_pytest_arguments(options) -> List[str]:
             marks = marks.replace("and not moose", "")
         args.extend(["-m", marks])
 
+    args.extend(["-k", "test_coredump"])
```
tomasz-grz, line 160: REVERT THIS!!!

nat-lab/tests/telio.py
```diff
@@ -1227,6 +1229,57 @@ async def probe_pmtu(self, host: str, expected: int):
...
+    # For CI and our internal linux VM, this path is set in our provisioning scripts
+    # If you're running locally without the aforementioned linux VM, you are expected to configure this yourself
+    # However, this only needs to be set iff you're: 
+    #  - running a test targeting a docker image
+    #  - have set the NATLAB_SAVE_LOGS environment variable
+    #  - want to have natlab automatically collect core dumps for you
+    def get_coredump_folder(self) -> tuple[str, str]: 
+        return "/var/crash", "core-"
+
+    def should_collect_core_dumps(self) -> bool:
```
Jauler, line 1242: The function name seems to be reverse of what it actually returns, e.g.: `should_collect_core_dumps() == true` means that we should _skip_ collecting dumps

Jauler (approved):
+2.0

nat-lab/tests/telio.py
mathiaspeters: fixed

Jauler (approved):

PR #672, opened 2024-07-10
Additional error logging on wg nt failure

matislovas:
### Problem
When starting wg-nt adapter, there can still be failures, but we do not have any error description from OS (only bool on return).

### Solution
Add `GetLastError` method to those calls, to retrieve the error code and message.

- [x] README.md is updated


gytsto (approved):
looks good +2 :smile:

PR #675, opened 2024-07-10
Fix telio shutdown

jjanowsk:
Change shutdown_art and consequently Telio.shutdown implementation to blocking. The old implementation was using shutdown_background method which just left worker threads running without joining them. That could result in a crash e.g. in a scenario when the telio user called shutdown and then unloaded libtelio library. The new implemntation uses Runtime's Drop implementation which is blocking and joins all threads.

There are still issues I see when it comes to Telio's shutdown API but I think it can be done in the future PR:
1. In the current state both `shutdown` and `shutdown_hard` close runtime but only `shutdown_hard's` documentation mentions it.
2. `shutdown` closes runtime and after that there is no possibility to call anything else on it (Previous name destroy was better I guess). Given that why is it not destroying the Device? `shutdown_hard` does it correctly.
3. If we want to keep both I think both should destroy the Device with the only difference that `shutdown_hard` should do it via `shutdown_timeout` so it can try to recover even if something bad happens which could lead to a hang, although I'm not even convinced that this is a correct way to go (I would like to hear why it was introduced in the first place). 

- [x] README.md is updated

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -54,7 +54,7 @@ use wg::uapi::{self, AnalyticsEvent, PeerState};
 use std::{
     collections::{hash_map::Entry, HashMap, HashSet},
     future::Future,
-    io::{self, Error as IoError, ErrorKind},
+    io::{self, Error as IoError},
     net::{IpAddr, Ipv4Addr, SocketAddr},
     sync::Arc,
     time::{Duration, Instant},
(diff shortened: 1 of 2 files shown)


mathiaspeters:
+1

gytsto (approved):
Looks good +1

PR #676, opened 2024-07-10
LLT-5310: Make nurse state duration cap configurable

mathiaspeters:
### Problem
Analytics may be skewed by sessions being forcibly reported after 24h, which is the default time limit, and POs want to make the time limit configurable

### Solution
Add `state_duration_cap` to nurse config and use it in nurse

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -154,6 +154,9 @@ pub struct FeatureNurse {
     /// Enable/disable NAT-traversal connections data. Enabled by default.
     #[serde(default = "FeatureNurse::default_enable_nat_traversal_conn_data")]
     pub enable_nat_traversal_conn_data: bool,
+    /// How long a session can be before it is forcibly reported, in seconds. Default value is 24h.
+    #[serde(default = "FeatureNurse::default_state_duration_cap_in_seconds")]
+    pub state_duration_cap: u64,
 }
 
 impl FeatureNurse {
@@ -180,6 +183,11 @@ impl FeatureNurse {
     fn default_enable_nat_traversal_conn_data() -> bool {
         true
     }
+
+    // The state duration cap was originally hardcoded at 24h, which is now the default value
+    fn default_state_duration_cap_in_seconds() -> u64 {
+        60 * 60 * 24
+    }
 }
 
 impl Default for FeatureNurse {
@@ -192,6 +200,7 @@ impl Default for FeatureNurse {
             enable_nat_type_collection: Self::default_enable_nat_type_collection(),
             enable_relay_conn_data: Self::default_enable_relay_conn_data(),
             enable_nat_traversal_conn_data: Self::default_enable_nat_traversal_conn_data(),
+            state_duration_cap: Self::default_state_duration_cap_in_seconds(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Jauler (dismissed):
+1.0

crates/telio-model/src/features.rs
```diff
@@ -690,6 +698,7 @@ mod tests {
             enable_nat_type_collection: false,
             enable_relay_conn_data: true,
             enable_nat_traversal_conn_data: true,
+            state_duration_cap: 60 * 60 * 24,
```
ThrasherLT, line 701: Why not also use default_state_duration_cap() here? Less manual work if we needed to change the value in the future.

crates/telio-model/src/features.rs
```diff
@@ -180,6 +183,10 @@ impl FeatureNurse {
     fn default_enable_nat_traversal_conn_data() -> bool {
         true
     }
+
+    fn default_state_duration_cap() -> u64 {
+        60 * 60 * 24
```
ThrasherLT, line 188: Nit: It's always nice to have some comment next to constant like these to understand why the specific value was chosen.

crates/telio-model/src/features.rs
```diff
@@ -180,6 +183,10 @@ impl FeatureNurse {
     fn default_enable_nat_traversal_conn_data() -> bool {
         true
     }
+
+    fn default_state_duration_cap() -> u64 {
```
ThrasherLT, line 187: If it's not possible to use Duration type to describe this duration, it would help a lot if the measurement units were appended to this function name or at least in a comment. e.g. default_state_duration_cap_s(). Especially if the value changed in the future and the measurement units would become less obvious for this value.

crates/telio-nurse/src/config.rs
```diff
@@ -96,6 +99,17 @@ impl AggregatorConfig {
...
             nat_traversal_events: features.enable_nat_traversal_conn_data,
+            state_duration_cap: Duration::from_secs(features.state_duration_cap),
+        }
+    }
+}
+
+impl Default for AggregatorConfig {
+    fn default() -> Self {
+        Self {
+            relay_events: Default::default(),
```
ThrasherLT, line 110: Nit: I think you can use ..Default:default() here for all the remaining members, but then again if we had relay_events: false, we'd know that the default value for relay::events is deliberately chosen to be false and not just forgotten, but this is a total nit.

crates/telio-model/src/features.rs
```diff
@@ -154,6 +154,9 @@ pub struct FeatureNurse {
     /// Enable/disable NAT-traversal connections data. Enabled by default.
     #[serde(default = "FeatureNurse::default_enable_nat_traversal_conn_data")]
     pub enable_nat_traversal_conn_data: bool,
+    /// How long a session can be before it is forcibly reported, in seconds. Default value is 24h.
+    #[serde(default = "FeatureNurse::default_state_duration_cap")]
+    pub state_duration_cap: u64,
```
matislovas, line 159: If the Duration is not possible, I would make a typedef of `DurationCapSecs` or similar ...

matislovas:
+0.5

crates/telio-nurse/src/config.rs
mathiaspeters: I used `..Default::default()` at first but cargo was saying that it would always recurse so I changed it to setting them manually

crates/telio-model/src/features.rs
mathiaspeters: I see where you're coming from but the `duration` in the name doesn't refer to the type duration but rather the non-programming concept of duration, so I don't think this needs a typedef, especially not with the doc comment being there

ThrasherLT (approved):
+0.5

PR #679, opened 2024-07-13
LLT-5339: Session Keeper ping packet batching

LukasPukenis:
### Problem
According to empirical testing and some research it appears that sending more data at once is beneficial to battery instead of sending sporadically.

### Solution
- SessionKeeper is modified to hold both `RepeatedActions` and `Batcher` instances and uses each one depending on the passed enum parameter. This is done for safety so the chances of problems introduced.

Having no batcher option should not ever trigger it.

- [x] README.md is updated

### DIFF
--- a/crates/telio-batcher/Cargo.toml
+++ b/crates/telio-batcher/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "telio-batcher"
+version = "0.1.0"
+edition = "2021"
+license = "GPL-3.0-only"
+repository = "https://github.com/NordSecurity/libtelio"
+publish = false
+
+[dependencies]
+tokio = { workspace = true, features = ["net", "sync", "test-util"] }
+telio-utils.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
+futures.workspace = true
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["time", "rt", "macros", "test-util"] }
--- a/crates/telio-batcher/src/batcher.rs
+++ b/crates/telio-batcher/src/batcher.rs
@@ -1,0 +1,416 @@
+use futures::future::BoxFuture;
+use std::fmt::Debug;
+use std::hash::Hash;
+use std::{collections::HashMap, sync::Arc};
+
+use telio_utils::telio_log_debug;
+use tokio::time::sleep_until;
+
+type Action<V, R = std::result::Result<(), ()>> =
+    Arc<dyn for<'a> Fn(&'a mut V) -> BoxFuture<'a, R> + Sync + Send>;
+
+pub struct Batcher<K, V> {
+    actions: HashMap<K, (BatchEntry, Action<V>)>,
+    notify_add: tokio::sync::Notify,
+}
+
+struct BatchEntry {
+    deadline: tokio::time::Instant,
+    interval: std::time::Duration,
+    threshold: std::time::Duration,
+}
+
+impl<K, V> Default for Batcher<K, V>
+where
+    K: Eq + Hash + Send + Sync + Debug + Clone,
+{
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<K, V> Batcher<K, V>
+where
+    K: Eq + Hash + Send + Sync + Debug + Clone,
+{
+    pub fn new() -> Self {
+        Self {
+            actions: HashMap::new(),
+            notify_add: tokio::sync::Notify::new(),
+        }
+    }
+
+    /// Batching works by sleeping until the nearest future and then trying to batch more actions
+    /// based on the threshold value. Higher delay before calling the function will increase the chances of batching
+    /// because the deadlines will _probably_ be in the past already.
+    /// Adding a new action wakes up the batcher due to immediate trigger of the action.
+    pub async fn get_actions(&mut self) -> Vec<(K, Action<V>)> {
+        let mut batched_actions: Vec<(K, Action<V>)> = vec![];
+
+        loop {
+            if !self.actions.is_empty() {
+                let actions = &mut self.actions;
+
+                // TODO: This can be optimized by early breaking and precollecting items beforehand
+                if let Some(closest_entry) = actions.values().min_by_key(|entry| entry.0.deadline) {
+                    tokio::select! {
+                        _ = self.notify_add.notified() => {
+                            // Item was added, we need to immediately emit it
+                        }
+                        _ = sleep_until(closest_entry.0.deadline) => {
+                            // Closest action should now be emitted
+                        }
+                    }
+
+                    let now = tokio::time::Instant::now();
+                    // at this point in time we know we're at the earliest spot for batching, thus we can check if we have more actions to add
+                    for (key, action) in actions.iter_mut() {
+                        let adjusted_action_deadline = now + action.0.threshold;
+
+                        if action.0.deadline <= adjusted_action_deadline {
+                            action.0.deadline = now + action.0.interval;
+                            batched_actions.push((key.clone(), action.1.clone()));
+                        }
+                    }
+                }
+
+                return batched_actions;
+            } else {
+                let _ = self.notify_add.notified().await;
+            }
+        }
+    }
+
+    /// Remove batcher action. Action is no longer eligible for batching
+    pub fn remove(&mut self, key: &K) {
+        telio_log_debug!("removing item from batcher with key({:?})", key);
+        self.actions.remove(key);
+    }
+
+    /// Add batcher action. Batcher itself doesn't run the tasks and depends
+    /// on actions being manually invoked. Adding an action immediately triggers it
+    /// thus if the call site awaits for the future then it will resolve immediately after this
+    /// function call.
+    pub fn add(
+        &mut self,
+        key: K,
+        interval: std::time::Duration,
+        threshold: std::time::Duration,
+        action: Action<V>,
+    ) {
+        telio_log_debug!(
+            "adding item to batcher with key({:?}), interval({:?}), threshold({:?})",
+            key,
+            interval,
+            threshold,
+        );
+        let entry = BatchEntry {
+            deadline: tokio::time::Instant::now(),
+            interval,
+            threshold,
+        };
+
+        self.actions.insert(key, (entry, action));
+        self.notify_add.notify_waiters();
+    }
+}
+
+/// Tests provide near-perfect immediate sleeps due to how tokio runtime acts when it's paused(basically sleeps are resolved immediately)
+#[cfg(test)]
+mod tests {
+
+    pub struct TestChecker {
+        values: Vec<(String, tokio::time::Instant)>,
+    }
+
+    use std::{sync::Arc, time::Duration};
+
+    use crate::batcher::Batcher;
+
+    #[tokio::test(start_paused = true)]
+    async fn batch_one() {
+        let start_time = tokio::time::Instant::now();
+        let mut batcher = Batcher::<String, TestChecker>::new();
+        batcher.add(
+            "key".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(0),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        let mut test_checker = TestChecker { values: Vec::new() };
+
+        // pick up the immediate fire
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+        assert!(test_checker.values.len() == 1);
+
+        // pick up the second event
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+
+        assert!(test_checker.values.len() == 2);
+
+        assert!(test_checker.values[0].1.duration_since(start_time) == Duration::from_secs(0));
+        assert!(test_checker.values[1].1.duration_since(start_time) == Duration::from_secs(100));
+
+        // after a pause of retrieving actions there will be a delay in signals as well as they are not active on their own
+        tokio::time::advance(Duration::from_secs(550)).await;
+        assert!(test_checker.values.len() == 2);
+
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+
+        assert!(test_checker.values.len() == 3);
+        assert!(test_checker.values[2].1.duration_since(start_time) == Duration::from_secs(650));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn batch_two_with_threshold_and_delay() {
+        let start_time = tokio::time::Instant::now();
+        let mut batcher = Batcher::<String, TestChecker>::new();
+        batcher.add(
+            "key0".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(50),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key0".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        tokio::time::advance(Duration::from_secs(30)).await;
+
+        batcher.add(
+            "key1".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(50),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key1".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        // At this point there are two actions added, one at t(0) and another at t(30).
+        let mut test_checker = TestChecker { values: Vec::new() };
+
+        // pick up the immediate fires
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+        assert!(test_checker.values.len() == 2);
+
+        // Do again and batching should be in action since the threshold of the second signal is bigger(50) than the delay between the packets(30)
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+
+        assert!(test_checker.values.len() == 4);
+
+        let key0_entries: Vec<tokio::time::Duration> = test_checker
+            .values
+            .iter()
+            .filter(|e| e.0 == "key0")
+            .map(|e| e.1.duration_since(start_time))
+            .collect();
+        let key1_entries: Vec<tokio::time::Duration> = test_checker
+            .values
+            .iter()
+            .filter(|e| e.0 == "key1")
+            .map(|e| e.1.duration_since(start_time))
+            .collect();
+
+        // Immediate fires were supressed because we need to poll them and we did so only after 30seconds
+        // Thus everything will be aligned at 30seconds
+
+        assert!(key0_entries[0] == Duration::from_secs(30));
+        assert!(key1_entries[0] == Duration::from_secs(30));
+
+        assert!(key0_entries[1] == Duration::from_secs(130));
+        assert!(key1_entries[1] == Duration::from_secs(130));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn batch_two_no_threshold_delayed_check() {
+        let _start_time = tokio::time::Instant::now();
+        let mut batcher = Batcher::<String, TestChecker>::new();
+        batcher.add(
+            "key0".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(0),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key0".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        tokio::time::advance(Duration::from_secs(30)).await;
+
+        batcher.add(
+            "key1".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(0),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key1".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        // At this point there are two actions added, one at t(0) and another at t(30).
+        let mut test_checker = TestChecker { values: Vec::new() };
+
+        // pick up the immediate fire
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+        assert!(
+            test_checker.values.len() == 2,
+            "Both actions should be emitted immediately"
+        );
+
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+
+        assert!(test_checker.values.len() == 4, "Even though there was no threshold and actions were added at different times, we query only after adding the second one which resets both deadlines and aligns them");
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn batch_two_no_threshold_nodelay_check() {
+        let start_time = tokio::time::Instant::now();
+        let mut batcher = Batcher::<String, TestChecker>::new();
+        batcher.add(
+            "key0".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(0),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key0".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        let mut test_checker = TestChecker { values: Vec::new() };
+        // pick up the immediate fire
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+        assert!(test_checker.values.len() == 1);
+
+        tokio::time::advance(Duration::from_secs(30)).await;
+
+        batcher.add(
+            "key1".to_owned(),
+            Duration::from_secs(100),
+            Duration::from_secs(0),
+            Arc::new(|s: _| {
+                Box::pin(async move {
+                    s.values
+                        .push(("key1".to_owned(), tokio::time::Instant::now()));
+                    Ok(())
+                })
+            }),
+        );
+
+        // pick up the immediate fire from the second action
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+        assert!(test_checker.values.len() == 2);
+
+        for ac in batcher.get_actions().await {
+            ac.1(&mut test_checker).await.unwrap();
+        }
+
+        assert!(
+            test_checker.values.len() == 3,
+            "No threshold but a delay was given, thus one signal should have resolved"
+        );
+
+        test_checker.values = vec![];
+        for _ in 0..12 {
+            for ac in batcher.get_actions().await {
+                ac.1(&mut test_checker).await.unwrap();
+            }
+        }
+
+        let key0sum: tokio::time::Duration = test_checker
+            .values
+            .iter()
+            .filter(|e| e.0 == "key0")
+            .map(|e| e.1.duration_since(start_time))
+            .collect::<Vec<Duration>>()
+            .windows(2)
+            .map(|w| w[1] - w[0])
+            .sum();
+
+        let key1sum: tokio::time::Duration = test_checker
+            .values
+            .iter()
+            .filter(|e| e.0 == "key1")
+            .map(|e| e.1.duration_since(start_time))
+            .collect::<Vec<Duration>>()
+            .windows(2)
+            .map(|w| w[1] - w[0])
+            .sum();
+
+        // Because of misalignment, each call produces only one action instead of both
+        assert!(key0sum == Duration::from_secs(500));
+        assert!(key1sum == Duration::from_secs(500));
+
+        // Now let's wait a bit so upon iterating again signals would be aligned
+        test_checker.values = vec![];
+        tokio::time::advance(tokio::time::Duration::from_secs(500)).await;
+        for _i in 0..11 {
+            for ac in batcher.get_actions().await {
+                ac.1(&mut test_checker).await.unwrap();
+            }
+        }
+
+        let key0sum: tokio::time::Duration = test_checker
+            .values
+            .iter()
+            .filter(|e| e.0 == "key0")
+            .map(|e| e.1.duration_since(start_time))
+            .collect::<Vec<Duration>>()
+            .windows(2)
+            .map(|w| w[1] - w[0])
+            .sum();
+
+        let key1sum: tokio::time::Duration = test_checker
+            .values
+            .iter()
+            .filter(|e| e.0 == "key1")
+            .map(|e| e.1.duration_since(start_time))
+            .collect::<Vec<Duration>>()
+            .windows(2)
+            .map(|w| w[1] - w[0])
+            .sum();
+
+        assert!(key0sum == Duration::from_secs(1000));
+        assert!(key1sum == Duration::from_secs(1000));
+    }
+}
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -54,6 +54,17 @@ pub struct Features {
     pub pmtu_discovery: Option<FeaturePmtuDiscovery>,
     /// Multicast support
     pub multicast: bool,
+    /// Batching feature configuration, disabled by default, used for batching keep-alives
+    pub batching: Option<FeatureBatching>,
+}
+
+/// Configure keepalive batching
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, SmartDefault)]
+#[serde(default)]
+pub struct FeatureBatching {
+    /// Direct connection threshold when batching (in seconds) [default 0s]
+    #[default(0)]
+    pub direct_connection_threshold: u32,
 }
 
 /// Configurable features for Wireguard peers
@@ -487,7 +498,10 @@ mod tests {
             "pmtu_discovery": {
                 "response_wait_timeout_s": 20
             },
-            "multicast": true
+            "multicast": true,
+            "batching": {
+                "direct_connection_threshold": 60
+            }
         }
         "#;
             let actual = from_str::<Features>(json).unwrap();
@@ -569,6 +583,9 @@ mod tests {
                     response_wait_timeout_s: 20,
                 }),
                 multicast: true,
+                batching: Some(FeatureBatching {
+                    direct_connection_threshold: 60,
+                }),
             };
 
             assert_eq!(actual, expected);
--- a/crates/telio-traversal/src/session_keeper.rs
+++ b/crates/telio-traversal/src/session_keeper.rs
@@ -1,6 +1,6 @@
 use async_trait::async_trait;
-use futures::Future;
 use socket2::Type;
+use std::future::Future;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::sync::Arc;
 use std::time::Duration;
@@ -8,12 +8,12 @@ use surge_ping::{
     Client as PingerClient, Config as PingerConfig, ConfigBuilder, PingIdentifier, PingSequence,
     SurgeError, ICMP,
 };
+use telio_batcher::batcher::Batcher;
 use telio_crypto::PublicKey;
 use telio_sockets::SocketPool;
 use telio_task::{task_exec, BoxAction, Runtime, Task};
 use telio_utils::{
-    dual_target, repeated_actions, telio_log_debug, telio_log_trace, telio_log_warn, DualTarget,
-    RepeatedActions,
+    dual_target, repeated_actions, telio_log_debug, telio_log_warn, DualTarget, RepeatedActions,
 };
 
 const PING_PAYLOAD_SIZE: usize = 56;
@@ -48,12 +48,12 @@ type Result<T> = std::result::Result<T, Error>;
 pub trait SessionKeeperTrait {
     async fn add_node(
         &self,
-        public_key: &PublicKey,
+        public_key: PublicKey,
         target: dual_target::Target,
         interval: Duration,
+        threshold: Option<Duration>,
     ) -> Result<()>;
-    async fn update_interval(&self, public_key: &PublicKey, interval: Duration) -> Result<()>;
-    async fn remove_node(&self, public_key: &PublicKey) -> Result<()>;
+    async fn remove_node(&self, key: &PublicKey) -> Result<()>;
 }
 
 pub struct SessionKeeper {
@@ -74,9 +74,12 @@ impl SessionKeeper {
 
         Ok(Self {
             task: Task::start(State {
-                pinger_client_v4: client_v4,
-                pinger_client_v6: client_v6,
-                keepalive_actions: RepeatedActions::default(),
+                pingers: Pingers {
+                    pinger_client_v4: client_v4,
+                    pinger_client_v6: client_v6,
+                },
+                batched_actions: Batcher::new(),
+                nonbatched_actions: RepeatedActions::default(),
             }),
         })
     }
@@ -111,8 +114,8 @@ impl SessionKeeper {
     async fn get_pinger_client(&self, proto: ICMP) -> Result<PingerClient> {
         task_exec!(&self.task, async move |s| {
             Ok(match proto {
-                ICMP::V4 => s.pinger_client_v4.clone(),
-                ICMP::V6 => s.pinger_client_v6.clone(),
+                ICMP::V4 => s.pingers.pinger_client_v4.clone(),
+                ICMP::V6 => s.pingers.pinger_client_v6.clone(),
             })
         })
         .await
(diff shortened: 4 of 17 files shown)


nat-lab/docker-compose.yml
```diff
@@ -51,6 +51,8 @@ services:
     cap_drop:
       - ALL
     cap_add:
+      - CAP_DAC_OVERRIDE
+      - CAP_CHOWN
```
LukasPukenis, line 55: Without both these, tcpdump fails in Docker container to create the capture file

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,533 @@
...
+
+pub struct Batcher {
+    task: Task<State>,
+}
+
+struct BatchEntry {
+    deadline: tokio::time::Instant,
+    interval: std::time::Duration,
+    threshold: std::time::Duration,
+    tx: tokio::sync::watch::Sender<()>,
```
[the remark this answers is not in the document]
LukasPukenis: Could you elaborate a bit more how those could be missed? There's supposed to be 1:1 sender:receiver mapping by the current design

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,533 @@
...
+                            entry.deadline = now + entry.interval;
+
+                            if entry.tx.send(()).is_err() {
+                                keys_to_delete.push(key.clone());
+                            }
+                        }
+                    }
+                }
+
+                // Safety guard. Should never be logged
```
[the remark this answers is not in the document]
LukasPukenis: The sending of the data to the channel would result in an error as the receiver was dropped I think? Basically channels return an error if the corresponding end is not present, given that SessionKeeper is the owner of the Receiver it should work out naturally

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
...
+            ),
+            histogram_check_fn,
+        )
+        for conn_tag, endpoint_providers, batching, histogram_check_fn in cfg
+    ]
+
+
+def NO_BATCH_DOCKER_CONE_CLIENT_1_histogram_check(hs: HistogramResult):
+    # should have no naturally aligned keepalives due to sleep
+    assert hs.outgoing[8] == 0
```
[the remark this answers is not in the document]
LukasPukenis: So the numbers are hand adjusted and are very much tied to the batching and misalignment intervals. I should probably express the idea that numbers are meant to be hardcoded and hand crated, basically - 4seconds interval with 500ms buckets in histogram means the 8th bucket should be either fill(in batched case) or empty in a non-batched case. I will try to put in more commentary

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
...
+@pytest.mark.parametrize("client_params", [CLIENTS_NO_BATCHING, CLIENTS_BATCHING])
+async def test_direct_batching(client_params: Any) -> None:
+    async with AsyncExitStack() as exit_stack:
+        for setup_params in client_params:
+            # This testcase initiates 2 meshnet peers, and after a small delay initiates the third peer.
+            # The reason for this is to misalign their keepalive echo requests and observe batching in action
+            # The third peer has naturally aligned keepalives even if with no batching as it received meshmap
+            # with both peers at once and added the keys
+
+            setup_params = setup_params[0][0]  # TODO: Pytest, wtf
```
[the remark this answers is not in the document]
LukasPukenis: So I was a bit frustrated because Pytest has no worries iteracting over a singular SetupParameters by symply accessing `setup_parameters` however once I convert that to an array suddenly I need double the indirection which does'nt really make any sense to me, but of course, I will explain it further, good point :)

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -75,6 +75,7 @@ def next_event(self):
 
 class TelioLoggerCbImpl(libtelio.TelioLoggerCb):
     def log(self, log_level, payload):
+        # print(f"{datetime.datetime.now()} {log_level} {payload}", flush=True)
```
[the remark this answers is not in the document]
LukasPukenis: To be honest I left it here deliberately as I found this super useful when coding - this way I could see all `telio_*_log!()` macro statements in the natlab run. I just commented it out so it would be there. Any other way to achieve this?

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
tomaszklak: You are right, when writing that I somehow thought that the `batched_count` is incremented only when send succeeded.

nat-lab/tests/utils/batch_observer.py
```diff
@@ -0,0 +1,155 @@
...
+        filepath = f"./{self._target.tag}.pcap"
+        print("Looking for a pcap at", filepath)
+
+        total_time_ms = 0
+        hist = Histogram()
+
+        def is_icmp_echo_request(pkt):
+            return (
+                pkt.haslayer(ICMP) and pkt[ICMP].type == 8
+            )  # ICMP type 8 is echo request
```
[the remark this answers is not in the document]
LukasPukenis: I wasn't aware of this syntax, I guess it's a nice alternative, though I like the idea of filtering in the code more as it's more flexible and easier to debug.

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,533 @@
...
+                "signals should be aligned"
+            );
+            assert!(
+                y.duration_since(*x).as_secs_f64() < 0.01,
+                "signals should be aligned"
+            );
+        });
+    }
+
+    #[tokio::test(start_paused = true)]
```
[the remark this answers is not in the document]
LukasPukenis: Could you elaborate more on that? I though that pausing tokio time is already sort of _simulation_ as we can manually advance it(and I have a testcase for that), however I would be very interested in an alternative

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
tomaszklak: Under heavy load and short interval, the `select!` in the SessionKeepers' Task might not be called often enough to notice all notifications. But thinking about it now, this might actually be a desired behaviour... so I think you can ignore this comment.

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
+# Batching works by checking if certain packets can be sent together. This means that
+# in order to test it's effects we can filter outgoing data and then check the time deltas.
+# When batching is in action time deltas should fall into the left side of the histogram because
+# deltas should be very small as packets leave in succession. There should be no action in
+# the middle and then histogram should have data on the interval boundary. Basically letter "U" shaped.
+# When using no batching, histogram might look very differently depending on the signal misalignment. Testing
+# depends highly on misalignment and must be taken into an account.
```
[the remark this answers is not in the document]
LukasPukenis: Could you share more of your concerns? I sit hard on this problem and thought that actually using histograms for testing is a good way to test. Deadlines should be adhered to stricly in SessionKeeper, thus we can always expect a U shape with proper filtering.

In the current setup I don't see much possibilities for flakyness, unless of course, the testcae will do a `Ping()` then it will ruin it, but that's expected :)

nat-lab/tests/uniffi/libtelio_remote.py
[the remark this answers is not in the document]
tomaszklak: Could this be a optional behaviour controlled by a flag in `run_local.py` ?

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
...
+            # with both peers at once and added the keys
+
+            setup_params = setup_params[0][0]  # TODO: Pytest, wtf
+
+            mesh_params = [p for p, _ in setup_params]
+            env = await setup_mesh_nodes(exit_stack, mesh_params)
+            alpha, beta, gamma = env.nodes
+            _, _, gamma_client = env.clients
+
+            await asyncio.sleep(2)
```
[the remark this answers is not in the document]
tomaszklak: Yeah, I a comment before that sleep stating that this is here to force misalignment of timers would be great :)

nat-lab/tests/utils/batch_observer.py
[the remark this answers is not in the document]
tomaszklak: Well, I should have marked it as nitpick but forgot. If you find it better this way I'm ok with that (even though I would prefer to only collect packet of interest so that if that file becomes an artifact in future it's easier to analyze by hand).

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
tomaszklak: I'm not sure how to elaborate. I was just wandering if you tried for example to write a test like for example `batch_no_threshold` but with `#[tokio::test(start_paused = true)]`.

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+async def get_batch_histogram(
+    target: TargetMachine, mesh_node: Node
+) -> HistogramResult:
+    print(f"Running batch observer for target: {target.tag}")
+
+    await asyncio.sleep(2)
+    async with BatchObserver(target).run() as bo:
+        await asyncio.sleep(20)
```
tomaszklak, line 207: Can this sleep be avoided. Maybe we know the expected count of ping packets and could convert this to a loop that keeps getting the pcap and only breaks when the expected packet count have been found?

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
...
+    ),
+]
+
+
+async def get_batch_histogram(
+    target: TargetMachine, mesh_node: Node
+) -> HistogramResult:
+    print(f"Running batch observer for target: {target.tag}")
+
+    await asyncio.sleep(2)
```
tomaszklak, line 205: it's not clear why we need both the 2s and 20s sleeps here

nat-lab/tests/test_batching.py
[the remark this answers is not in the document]
tomaszklak: For example adding a sleep of 2s doesn't guarantee that the timers will be misaligned. They can just as well align at the 4s if the startup of postponed peer takes long enough. Unless I'm misunderstanding the test design :)

nat-lab/tests/test_batching.py
LukasPukenis: 20seconds - how much to capture
2seconds - how much to sleep before a third meshnet peer(gamma) is registered to the meshnet :)

LukasPukenis: Sorry, wrong explanation and this sleep I think is not needed at all. Good catch!

nat-lab/tests/test_batching.py
LukasPukenis: Hmm. Didn't think about the loop idea to be honest, it sounds pretty nice, I will consider it and see how it goes :thinking:

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
LukasPukenis: Oh I think I understand what you mean. By doing simulated time I could make the unit tests shorter as there would be no need to actually do the waiting itself. That's a fair point :thinking:

nat-lab/tests/test_batching.py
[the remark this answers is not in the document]
LukasPukenis: You're right that if the meshnet peer registration would take too long, it has a possibility to align with the 4seconds, introducing a flake, however should that happen in NatLab? We control all the necessary infrastructure so that should be pretty fast I suppose?

I initially played with low numbers such as 1seconds for intervals, thus 500ms for sleeping and those were flaking a bit thus I increased by 4x. Since then locally, at least, it didn't flake through a number of times.

---

I think the numbers are pretty safe to go and we can just adjust them as the time goes if they flake.

I have an idea how to stremaline the flakyness a bit - I could introduce some constants instead of hardcoded numbers so in a case of a flake, it would be very easy to update, how does that sound?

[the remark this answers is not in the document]
tomaszklak: I do know that machines we use for CI can get veeeery slow, especially when disk I/O is taking place. I think that last year in December we observed lana tests failures that were the result of 2 orders of magnitude (if not more, the slack thread is obviously long gone so I can't verify this) slowdown of disk access compared to healthy runner. That runner got replaced but the issue can happen is less pronounced way but still severe enough to add few seconds of slowdown.
But as I said I don't really have a better idea, so maybe this is just something to keep in mind if those tests ever become flaky.

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,304 @@
...
+    for _ in range(10):
+        bo.add_value(2)
+        bo.add_value(3)
+
+    for _ in range(50):
+        bo.add_value(4)
+
+    bo.add_value(9)
+
+    assert bo.get(10, 1) == [0, 0, 10, 10, 50, 0, 0, 0, 0, 1]
```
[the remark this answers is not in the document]
LukasPukenis: Moving to the same file might not work as `pytest` probably expects the `test_` prefix in the filename, but it's possible to move it out into a `test_histogram.py` inside of utils which I am not sure if is very useful

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,533 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Generic error")]
+    Generic(#[from] ExecError),
+}
+
+#[cfg_attr(any(test, feature = "mockall"), mockall::automock)]
+#[async_trait]
+pub trait BatcherTrait {
```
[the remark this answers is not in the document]
LukasPukenis: To be honest I lack the skill to do so, Trying few iterations yields me with various Rust compilation errors, until I end up with quite a messy implementation. I still need to later on pass the thing into the async task, so I need to clone it or the lifetimes become messy.. Anyway I would still love to do it so if you would be down to help me I would be very glad :)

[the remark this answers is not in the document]
tomaszklak: Heh, I tried to do it myself today and it's like you said. Given the fact that for `task_exec` would would either way had to `clone`/`to_owned` (even if other problem would be solved) I think that there is no point in making this change.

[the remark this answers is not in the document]
LukasPukenis: Thanks for explaining. I went with `clone()` solution and felt that there's really no way of doing else wise due to usage of `task` that requires `'static` lifetime as you mentioned :)

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -111,113 +138,202 @@ impl SessionKeeper {
...
-        .await
-        .map_err(Error::Task)
+    async fn remove_node<'a>(&self, params: RemoveNodeParameters<'a>) -> Result<()> {
+        match params {
+            RemoveNodeParameters::NonBatching { public_key } => {
+                let public_key = *public_key;
+                let _ = self.batcher.remove(&public_key.to_string()).await;
+
+                task_exec!(&self.task, async move |s| {
+                    if s.nonbatched_actions.contains_action(&public_key) {
```
tomaszklak, line 299: nit: this check is probably not needed

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -74,10 +97,14 @@ impl SessionKeeper {
...
             task: Task::start(State {
-                pinger_client_v4: client_v4,
-                pinger_client_v6: client_v6,
-                keepalive_actions: RepeatedActions::default(),
+                pingers: Pingers {
+                    pinger_client_v4: client_v4,
+                    pinger_client_v6: client_v6,
+                },
+                batched_actions: HashMap::new(),
+                nonbatched_actions: RepeatedActions::default(),
```
tomaszklak, line 105: where do we run non batched actions 🤔

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -43,25 +50,41 @@ pub enum Error {
...
+        target: dual_target::Target,
+        interval: Duration,
+    },
+    Batching {
+        public_key: PublicKey,
+        target: dual_target::Target,
+        interval: Duration,
+        threshold: Duration,
+    },
+}
```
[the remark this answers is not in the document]
LukasPukenis: My way of choosing an explicit enum was fueled by the fact it should be very explicit what is being passed and where because of the friction batcher introduces - it should not harm anything, thus being very explicit about what's being passed makes sense to me.

Though I can understand that this duplication of code here might not be the nicest thing

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,344 @@
...
+                .collect();
+
+            futures
+        };
+
+        tokio::select! {
+            update = update => {
+                return update(self).await;
+            }
+            Some(_) = sleep_futures.next() => {
```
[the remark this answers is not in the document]
LukasPukenis: Oh I think you're right

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,344 @@
...
+    ) -> std::result::Result<tokio::sync::watch::Receiver<()>, Error> {
+        let key = String::from(key);
+        task_exec!(&self.task, async move |s| {
+            telio_log_debug!(
+                "adding item to batcher with key({:?}), interval({:?}), threshold({:?})",
+                key,
+                interval,
+                threshold
+            );
+            let (tx, rx) = tokio::sync::watch::channel(());
```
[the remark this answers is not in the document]
LukasPukenis: I think it looks nicer as I don't need any value to send. However it looks much more complex reading it's documentation:

- no error when the sender is dropped, thus I can't collect tasks that don't have corresponding notifier in the batcher. It's not a big deal, the proper API should allow for removal of the key from the batcher instead and should not depend on this behaviour.

- there's no notion of sender and receiver, returned instance can be used to notify as well, so the ergonomis are not as nice.

These are my personal two cents from quickly looking at the docs :)

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: So that was a bug actually, it was caught by natlab, fix this one. SessionKeeper's `tokio::select!` now awaits for bothh: bathed and nonbatched futures

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: You're right, it doesn't check for any errors thus removing this is the way to go, fixed it

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
tomaszklak: Ok, if you feel that the downsides are too big feel free to ignore this comment :)

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,344 @@
...
+            "Should be not signalled just before hitting the deadline"
+        );
+        time::advance(std::time::Duration::from_millis(1)).await;
+        tokio::task::yield_now().await;
+        assert!(
+            rx.has_changed().unwrap(),
+            "should have signalled right after hitting the deadline"
+        );
+
+        // immitate a ton of signals and that they all trigger
```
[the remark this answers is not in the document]
LukasPukenis: It's just a weird wording from my side, the comment was basically stating the obvious - immitate a ton of signals and that each of them arrives as expected. I changed the wording on both instances of this copy-pasted comment :)

crates/telio-model/src/features.rs
```diff
@@ -18,6 +18,20 @@ pub const DEFAULT_DIRECT_PERSISTENT_KEEPALIVE_PERIOD: u32 = 5;
 /// Type alias for UniFFI
 pub type EndpointProviders = HashSet<EndpointProvider>;
 
+/// Batching configuraion
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Default)]
+pub struct FeatureBatching {
+    /// Direct connection threshold
```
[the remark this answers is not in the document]
LukasPukenis: Sure thing. I copy pasted the same approach as used elsewhere, indicating seconds only in the comment so it would be consistent

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -111,113 +138,202 @@ impl SessionKeeper {
...
+                                        };
+
+                                        let _ = secondary_client
+                                            .pinger(second, PingIdentifier(rand::random()))
+                                            .await
+                                            .send_ping(PingSequence(0), &[0; PING_PAYLOAD_SIZE])
+                                            .await?;
+                                    }
+                                }
+                                Ok(())
```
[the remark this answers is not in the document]
LukasPukenis: Let me see if I can do it(probably yes)

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,344 @@
...
+            )
+            .await
+            .unwrap();
+
+        assert!(
+            rx.has_changed().unwrap(),
+            "After adding a signal it should be immediately dispatched"
+        );
+        rx.mark_unchanged();
+        assert!(!rx.has_changed().unwrap());
```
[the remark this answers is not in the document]
tomaszklak: For me this test is passing:
```rust
    #[tokio::test(start_paused = true)]
    async fn dummy_test() {
        let (tx, mut rx) = tokio::sync::watch::channel(0);

        assert!(!rx.has_changed().unwrap());
        assert!(timeout(Duration::from_secs(1), rx.changed()).await.is_err());
        dbg!(rx.borrow().clone());

        assert!(!rx.has_changed().unwrap());
        tx.send(1).unwrap();
        assert!(rx.has_changed().unwrap());

        assert!(timeout(Duration::from_secs(0), rx.changed()).await.is_ok());
        assert!(!rx.has_changed().unwrap());
        dbg!(rx.borrow().clone());

        assert!(timeout(Duration::from_secs(1), rx.changed()).await.is_err());
    }
```
and prints:
```
running 1 test
[crates/telio-batcher/src/batcher.rs:179:9] rx.borrow().clone() = 0
[crates/telio-batcher/src/batcher.rs:187:9] rx.borrow().clone() = 1
test batcher::tests::dummy_test ... ok
```
I think this shows that `changed()` doesn't always evaluate to the last value and that it correctly marks the last received value.

crates/telio-traversal/src/session_keeper.rs
tomaszklak: I think this means that we are missing some tests that should have failed.

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
LukasPukenis: The comment is relevant I think, might take a look into it at the very end once this PR reaches the final state :) Leaving this as unresolved for now.

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,533 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Generic error")]
+    Generic(#[from] ExecError),
+}
+
+#[cfg_attr(any(test, feature = "mockall"), mockall::automock)]
+#[async_trait]
+pub trait BatcherTrait {
+    async fn remove(&self, key: String);
```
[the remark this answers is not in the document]
LukasPukenis: done, even though the string is still cloned in the implementation but the API is nicer with `&str` :)

src/device/wg_controller.rs
```diff
@@ -1808,6 +1831,64 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn when_upgrade_requested_by_peer_then_upgrade_and_batching() {
+        let mut f = Fixture::new();
+        f.features.ipv6 = true;
+        f.features.batching = Some(Default::default());
+
+        let pub_key = SecretKey::gen().public();
+        let ip1 = IpAddr::from([1, 2, 3, 4]);
```
LukasPukenis, line 1841: This is just a copy-paste on an existing testcase just with batching added

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+}
+
+impl<K, V> Batcher<K, V>
+where
+    K: Eq + Hash + Send + Sync + Debug + Clone,
+{
+    pub fn new() -> Self {
+        Self {
+            actions: HashMap::new(),
```
tomaszklak, line 37: nit: This probably has no real world significance because I assume there will be max 20-30 events stored in real world. But if you think about it, the more common operation will be triggering futures, not adding/removing them. Which means that the linear time it takes to find a closest entry could be improved (at the cost of adding/removing) if we would use different structure for storing actions, like `std::collections::BinaryHeap`.

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
tomaszklak: If you plan to keep this behaviour, I think it would be a good idea to document it in the `fn add` docstring.

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -228,18 +261,28 @@ impl Runtime for State {
...
-
                 action(self)
                     .await
                     .map_or_else(|e| {
                         telio_log_warn!("({}) Error sending keepalive to {} node: {}", Self::NAME, pk, e.to_string());
                         Ok(())
                     }, |_| Ok(()))?;
             }
+            batched_actions = self.batched_actions.get_future() => {
+                let batched_actions = batched_actions;
```
tomaszklak, line 274: nit: Is this needed? If so why?

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -228,18 +261,28 @@ impl Runtime for State {
...
+
+                for action in batched_actions {
+                    let pk = action.0;
+                    let action = action.1;
```
tomaszklak, line 278: nit: maybe `for (pk, action) in batched_actions {` could be used here?

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -111,113 +114,143 @@ impl SessionKeeper {
...
 
-    async fn remove_node(&self, public_key: &PublicKey) -> Result<()> {
-        // FIXME: error handling with task_exec! seems to suck a lot. Need to fix that.
-        let public_key = *public_key;
+    // TODO: this only removes the batched key. There's probably a leak in the current implementation where the entries are not cleaned-up, only added.
+    async fn remove_node(&self, key: &PublicKey) -> Result<()> {
+        let key = *key;
         task_exec!(&self.task, async move |s| {
-            let _ = s.keepalive_actions.remove_action(&public_key);
+            let _ = s.batched_actions.remove(&key);
```
tomaszklak, line 236: why don't we do here `s.nonbatched_actions.remove_action(&key)` also?

src/device.rs
```diff
@@ -2285,7 +2290,10 @@ impl TaskRuntime for Runtime {
                                 direct_entities.cross_ping_check.notify_failed_wg_connection(public_key)
                                     .await?;
 
-                                direct_entities.session_keeper.remove_node(&public_key).await?;
+
+                                if self.features.batching.is_some() {
+                                    direct_entities.session_keeper.remove_node(&public_key).await?;
+                                };
```
tomaszklak, line 2296: I'm not sure why this is done conditionally. If that is the correct behaviour here it would be good to explain that in comment.

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,401 @@
...
+    // Batching works by sleeping until the nearest future and then trying to batch more actions
+    // based on the threshold value. Higher delay before calling the function will increase the chances of batching
+    // because the deadlines will _probably_ be in the past already.
+    pub async fn get_future(&mut self) -> Vec<(K, Action<V>)> {
+        let mut batched_actions: Vec<(K, Action<V>)> = vec![];
+
+        match self.actions.len() {
+            0 => {
+                // block if there's nothing to do. This enables the user of this function to use tokio::select properly.
+                std::future::pending::<Vec<(K, Action<V>)>>().await;
```
tomaszklak, line 50: With the current implementation, this code will never finish evaluating `get_future` in the assertion:
```rust
        let batcher = Arc::new(Mutex::new(Batcher::<String, TestChecker>::new()));

        let b = batcher.clone();
        let _ = tokio::spawn(async move {
            sleep(Duration::from_secs(1)).await;
            b.lock().await.add(
                "key".to_owned(),
                Duration::from_secs(1),
                Duration::from_secs(0),
                Arc::new(|s: _| {
                    Box::pin(async move {
                        s.values
                            .push(("key".to_owned(), tokio::time::Instant::now()));
                        Ok(())
                    })
                }),
            );
        });

        assert_eq!(1, batcher.lock().await.get_future().await.len());

```

crates/telio-traversal/src/session_keeper.rs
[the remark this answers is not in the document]
LukasPukenis: Threshold became an `Option<>`

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,401 @@
...
+{
+    pub fn new() -> Self {
+        Self {
+            actions: HashMap::new(),
+        }
+    }
+
+    // Batching works by sleeping until the nearest future and then trying to batch more actions
+    // based on the threshold value. Higher delay before calling the function will increase the chances of batching
+    // because the deadlines will _probably_ be in the past already.
```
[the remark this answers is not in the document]
LukasPukenis: done

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: Not needed, a leftover from the refactoring :))

LukasPukenis: removed it

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: Thanks for the suggestion, I've added it. Not yet familiar with the iterator syntax :)

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: Hmmm, my bad. I deliberately added `TODO` on top to address this while I left a bug. Thanks for picking up

LukasPukenis: fixed it

src/device.rs
LukasPukenis: My bad. I was under impression that original implementation never did delete the entries so wanted to leave the behaviour the same. Now it's apparent that it's wrong. Thanks for picking it up, I've patched it

crates/telio-batcher/src/batcher.rs
LukasPukenis: This one is deliberate. I think it makes for a nice usage in `tokio::select!` blocks. Do you have any suggestions regarding that?

I guess I could _deoptimize_ and not return a `pending()` future and just leave the call site handle everything. _should_ be safer to use. What's your opinion on this?

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
LukasPukenis: **TLDR:** I followed your suggestion and documented at the moment on `add()`.

I think the solution could be in the main batcher loop by adding an internal channel to communicate that "hey a thing was added". But I have two major points against it:
- it's simpler to be this way now. We also plan on moving all possible keepalives into there so the intervals are gonna be _almost_ the same. Still possibly would be a nice thing to have.
- I don't want to invest too much time in it being perfect at the moment but rather would prefer a gradual approach.

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
LukasPukenis: Current code eliminated channel use entirely thus marking it as resolved :)

crates/telio-batcher/src/batcher.rs
tomaszklak: What I would expect here is something like conditional variable. Sadly there is no such thing for tokio runtime. I think this can be easily simulated by:

1. wrapping match in a `loop { ... }`
2. explicit returning from the non-empty case
3. waiting for notification in empty case (eg using `tokio::sync::Notify::notified`) and letting the loop run for another iteration
4. generating a notification when new entry is added (eg using `tokio::sync::Notify::notify_waiters`)

I haven't tested this so take it with a grain of salt

crates/telio-batcher/src/batcher.rs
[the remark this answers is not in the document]
tomaszklak: I agree that this behaviour can be kept as long as it's documented.

crates/telio-batcher/src/batcher.rs
LukasPukenis: I've taken a quick-look into `BinaryHeap` but I don't see guarantees about the query time except for adding/removing and peeking the top-most item. Implying it's a binary tree under the hood(based on name) then it should be faster for lookups, in theory at least but I wonder what's the reality.

Could you share more as I failed to find it in the recommendations on google regarding `HashMap` replacements for read-heavy scenarios?

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,344 @@
...
+            "should have signalled right after hitting the deadline"
+        );
+
+        // immitate a ton of signals and that they all trigger
+        for _ in 0..1000 {
+            time::advance(std::time::Duration::from_secs(1)).await;
+            tokio::task::yield_now().await;
+            assert!(rx.has_changed().unwrap());
+            rx.mark_unchanged();
+            assert!(!rx.has_changed().unwrap());
```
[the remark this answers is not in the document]
LukasPukenis: New code doesn't use channels anymore, thus resolving

crates/telio-batcher/src/batcher.rs
LukasPukenis: I may have misunderstood then. The comment implies that `Batcher` will never resolve, is this due to Mutex usage..?

My response about being deliberate was for returning `pending()` future when there were no items :)

crates/telio-batcher/src/batcher.rs
tomaszklak: It's a kind of binary tree, but implemented using an array (wikipedia has a nice visualisation: https://en.wikipedia.org/wiki/Binary_heap#/media/File:Max-Heap.svg). Basically finding smallest (or biggest - depends how you set it up) value is constant (removing it is logarithmic).

But like I mentioned in the beginning - this point is relevant only if we expect to store probably hundereds at least events. Otherwise using less known struct for 20-30 entries doesn't make sense.

crates/telio-batcher/src/batcher.rs
tomaszklak: It will never resolve, because [pending](https://doc.rust-lang.org/std/future/fn.pending.html) never resolves.

LukasPukenis: Oh okay, now I see your point. Because the task spawned will take a short-while until it will be executed and it's locked.

I've address this by returning `Future<Option<Vec<T>>>` and just using `Some(futures)` inside of the call-site. This should now work properly by not blocking unecessarily and actually looks very nice in the call site as well

crates/telio-batcher/src/batcher.rs
LukasPukenis: Thanks for at least highlighting me an option. Will check it out some time later :)

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,404 @@
...
+    pub fn new() -> Self {
+        Self {
+            actions: HashMap::new(),
+        }
+    }
+
+    /// Batching works by sleeping until the nearest future and then trying to batch more actions
+    /// based on the threshold value. Higher delay before calling the function will increase the chances of batching
+    /// because the deadlines will _probably_ be in the past already.
+    pub async fn get_actions(&mut self) -> Option<Vec<(K, Action<V>)>> {
```
packgron, line 44: What is the benefit for having both None and vec![], here?
Maybe just return empty vector, if there are no actions to do

crates/telio-batcher/src/batcher.rs
```diff
@@ -0,0 +1,404 @@
...
+        let mut batched_actions: Vec<(K, Action<V>)> = vec![];
+
+        match self.actions.len() {
+            0 => {
+                return None;
+            }
+            _ => {
+                let actions = &mut self.actions;
+
+                if let Some(closest_entry) = actions.values().min_by_key(|entry| entry.0.deadline) {
```
packgron, line 54: NIT: maybe it makes sence to create (deadline, key) sorted vector and reuse that, you would avoid double iteration, and could also do an early return in batching step

crates/telio-model/src/features.rs
```diff
@@ -487,7 +498,10 @@ mod tests {
             "pmtu_discovery": {
                 "response_wait_timeout_s": 20
             },
-            "multicast": true
+            "multicast": true,
+            "batching": {
```
packgron, line 502: As we now have a feature builder, it would be nice to have a `enable_batching()` option, see src/ffi/defaults_builder

src/ffi/defaults_builder.rs
```diff
@@ -33,6 +33,7 @@ impl FeaturesDefaultsBuilder {
             multicast: false,
             ipv6: false,
             nicknames: false,
+            batching: default(),
```
packgron, line 36: NIT: Please explicitly use None, so it would be clear it's not some unknown defaults

crates/telio-batcher/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "telio-batcher"
+version = "0.1.0"
+edition = "2018"
```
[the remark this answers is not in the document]
LukasPukenis: Absolutely no reason, I think this is what my `cargo init` initialized with or I just copied from another crate. I've updated it to 2021

src/ffi/defaults_builder.rs
LukasPukenis: Good point. I like that too. Updated :)

crates/telio-batcher/src/batcher.rs
LukasPukenis: You're totally right - an empty vector and None are basically the same thing. However from my experiments SessionKeeper tried really hard calling it's `wait_with_update()` issued by the `Task` when I resolved immediately that I decided to do something about it.

To be honest I don't feel like this is the best solution except for the fact I can do `Some(_) = ... => ` inside of tokio::select!  :)

crates/telio-batcher/src/batcher.rs
LukasPukenis: It's a bit more difficult since each entry has a `threshold` as well. Sure we can calculate entries as `deadline - threshold` and then sort. That _could_ work as expected but I can't prove it.

Reasoning that we hit the closest deadline and then aggregate based on threshold is very easy to reason about and to prove that it works.

However I think it's a good point to leave a comment on to optimize once we maybe have more tests regarding the subject as now basically there isn't a test that shows it works actually :D

LukasPukenis: I left the comment for the future-me regarding your comment if you're ok with that

crates/telio-model/src/features.rs
LukasPukenis: Sure thing. Did that :)

crates/telio-batcher/src/batcher.rs
packgron: Then it should mostly be hidden area, there you are entering else pending or somthing of the like, you can always add pending on 0

crates/telio-batcher/src/batcher.rs
packgron: It's a nit for a reason :D

crates/telio-batcher/src/batcher.rs
LukasPukenis: @tomaszklak @packgron I've added a notifier into the `get_actions()`.

It also should handle the case where adding an action to batcher would delay if already awaiting, this should be handled now as well.

LukasPukenis: There's one caveat though. Given that Batcher is under a `Mutex` it will block for however long it is needed, naturally, so for example `get_actions()` will block until there's one item which means it will hold the mutex and will not allow it to be released for adding the item, in order for this to break, the call site needs to have a  tokio::select or tokio::timeout or some other form of _retry_.

Currently SessionKeeper and RepeatedTask doesn't have a mutex and RepeatedTask blocks until one of it's features resolves, meaning it would also block. However IMHO from my research the SessionKeeper's `wait_with_update()` will have another tokio::select! branch triggered which basically _fixes_ the blocking.

Not sure yet how relevant this is for Batcher though.

Here's a simple unit-test that fails at the moment:
```
#[tokio::test(start_paused = true)]
    async fn batch_blocks_until_theres_an_item() {
        let start_time = tokio::time::Instant::now();
        let mut batcher = Arc::new(Mutex::new(Batcher::<String, TestChecker>::new()));

        let batcher2 = batcher.clone();
        let handle = tokio::spawn(async move {
            tokio::time::advance(time::Duration::from_secs(50)).await;

            batcher2.lock().await.add(
                "key0".to_owned(),
                Duration::from_secs(100),
                Duration::from_secs(0),
                Arc::new(|s: _| {
                    Box::pin(async move {
                        s.values
                            .push(("key0".to_owned(), tokio::time::Instant::now()));
                        Ok(())
                    })
                }),
            );
        });

        
        let mut batcher = batcher.lock().await;
        let actions_fut = batcher.get_actions();
        let actions = tokio::time::timeout(time::Duration::from_secs(100), actions_fut).await;
        
        assert!(actions.is_ok());
        assert!(actions.unwrap().len() == 1);
    }
```


However this problem is basically not related to Batcher but in general is a problem in such situations. @tomaszklak @packgron opinions on this?

packgron: So there is 3 ways two think about this IMO:
- Add's are done in built faze, and repeats in wait step, so maybe we could just make api reflect that (I don't know if this is 100% true though). If we are sure those are two idepenedent actions we don't need to wait
- We need runtime additions / dealetions, in this case we could use tokio::sync::watch, instead of Mutex, and react to change in configuration.
- Leave it as small pitfall, and rely it's aways used in select arm

nat-lab/tests/test_features_builder.py
```diff
@@ -87,7 +88,10 @@ def test_telio_features_builder_all_defaults():
         "flush_events_on_stop_timeout_seconds": 0,
         "multicast": true,
         "ipv6": true,
-        "nicknames": true
+        "nicknames": true,
+        "batching": {
+            "direct_connection_threshold": 0
```
packgron, line 93: But this will basicaly don nothing?

crates/telio-batcher/src/batcher.rs
LukasPukenis: Okay. I will leave it as it is. This is because:
- I am not yet clear how it's gonna be used(so far only in direct connection cases through SessionKeeper)
- Batcher is under a feature flag
- Even though it never resolves on `get_actions()` if it's empty and nothing is added it will introduce no blockage into current libtelio workflow.

nat-lab/tests/test_features_builder.py
LukasPukenis: Yes, this is intentional. The idea is that the batcher touches imtimate parts of libtelio because of keep-alives. Meaning to reduce the friction we basically put it under a feature flag and just care if it doesn't break anything by just being there first. Then we care if it actually works to some degree which can be then fine-tuned through the feature flag :)

TLDR: so far we only care that batcher's existence doesn't break anything.

packgron (commented):
+1.0

tomaszklak (approved):
+1

PR #683, opened 2024-07-15
[LLT-5421] Remove rupnp crate from Telio

stalowyjez:
### Problem
Crate `rupnp` seems to be used only for an error which should be unused anyway

### Solution
Remove it

- [x] README.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/upnp.rs
+++ b/crates/telio-traversal/src/endpoint_providers/upnp.rs
@@ -12,7 +12,6 @@ use igd::{
 };
 use ipnet::Ipv4Net;
 use rand::Rng;
-use rupnp::Error::HttpErrorCode;
 use std::fmt::Debug;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4};
 use std::pin::Pin;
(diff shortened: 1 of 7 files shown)


gytsto (dismissed):
+1 lgtm

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -515,13 +514,12 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> State<Wg, I, E> {
                 }
             }
             Err(e) => {
-                if let Error::UpnpError(HttpErrorCode(http::StatusCode::INTERNAL_SERVER_ERROR)) = e
-                {
+                if let Error::NoIGDGateway = e {
```
tomaszklak, line 517: Why was this `if` section not removed, but instead a different error is checked here?

stalowyjez: I thought that this is the error which should be considered here, as this is the case in which there is no point in deleting routes when there are none - this delete)endpoint_routes would fail anyway

tomaszklak (commented):
+0.5

matislovas (approved):
+0.5

PR #686, opened 2024-07-16
LLT-5384: Handle non-UDP packets in telio-dns

mathiaspeters:
### Problem
telio-dns only actually supports UDP packets, and while we don't intend to support any other protocol, we want to handle them more gracefully. 

### Solution
The logic for UDP packets remain unchanged, new logic is added for TCP packets so that telio-dns simply returns a TCP reset, and for other protocols we produce a silent error.

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/src/nameserver.rs
+++ b/crates/telio-dns/src/nameserver.rs
@@ -10,9 +10,10 @@ use hickory_server::{
     server::{Protocol, Request},
 };
 use pnet_packet::{
-    ip::IpNextHeaderProtocols,
+    ip::{IpNextHeaderProtocol, IpNextHeaderProtocols},
     ipv4::{checksum, Ipv4Packet, MutableIpv4Packet},
     ipv6::{Ipv6Packet, MutableIpv6Packet},
+    tcp::{MutableTcpPacket, TcpFlags, TcpPacket},
     udp::{ipv4_checksum, ipv6_checksum, MutableUdpPacket, UdpPacket},
     Packet,
 };
@@ -32,6 +33,7 @@ const IPV4_HEADER: usize = 20; // bytes
 const IPV6_HEADER: usize = 40; // bytes
 const MAX_PACKET: usize = 2048;
 const UDP_HEADER: usize = 8;
+const TCP_MIN_HEADER: usize = 20;
 const MAX_CONCURRENT_QUERIES: usize = 256;
 
 /// NameServer is a server that stores the DNS records.
@@ -206,11 +208,15 @@ impl LocalNameServer {
         let resolver = Resolver::new();
         let zones = nameserver.zones().await;
 
-        let dns_request = request_info
-            .udp
-            .dns_request
-            .take()
-            .ok_or_else(|| String::from("Inexistent DNS request"))?;
+        let dns_request = match &mut request_info.payload {
+            PayloadRequestInfo::Udp {
+                ref mut dns_request,
+                ..
+            } => dns_request
+                .take()
+                .ok_or_else(|| String::from("Inexistent DNS request"))?,
+            _ => return Ok(Vec::new()),
+        };
         let dns_request = Request::new(dns_request, request_info.dns_source(), Protocol::Udp);
         telio_log_debug!("DNS request: {:?}", &dns_request);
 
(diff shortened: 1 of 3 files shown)


crates/telio-dns/src/nameserver.rs
```diff
@@ -253,13 +259,30 @@ impl LocalNameServer {
...
+
         Ok(RequestInfo {
             ip: ip.ip_request_info(),
-            udp: LocalNameServer::process_udp_packet(&ip)?,
+            payload,
         })
     }
 
-    fn process_udp_packet<'a>(ip_request: &impl IpPacket<'a>) -> Result<UdpRequestInfo, String> {
+    fn process_udp_packet<'a>(
```
Jauler, line 283: This is a kind of a nitpick, so treat it with low-prio ;). But maybe it makes sense to make `process_udp_packet()` strong-typed, and pass `UdpPacket` as an argument instead of `IpPacket` as an argument?

crates/telio-dns/src/nameserver.rs
```diff
@@ -414,23 +466,73 @@ impl RequestInfo {
...
+                    String::from("Failed to build MutableTcpPacket for response packet")
+                })?;
 
-        udp_response.set_source(self.udp.destination_port);
-        udp_response.set_destination(self.udp.source_port);
-        udp_response.set_length((UDP_HEADER + dns_response.len()) as u16);
-        udp_response.set_payload(dns_response);
-        udp_response.set_checksum(0);
-        udp_response.set_checksum(self.ip.compute_udp_checksum(&udp_response.to_immutable()));
+                tcp_response.set_source(source_port);
```
Jauler, line 507: I think we are missing sequence number for TCP

Jauler (changes requested):
IMHO it would be really nice to add integrational test which verifies, that clients attempting to connect over TCP does in-fact drop the connection immediatelly instead of waiting for timeout. A good example is running:
```bash
nc -w 5 100.64.0.2 53
```

it still times out, although expectation would be to receive an error something like this:
```
nc: connect to 100.64.0.2 port 53 (tcp) failed: Connection refused
```

tomaszklak:
> IMHO it would be really nice to add integrational test which verifies, that 3rd-party clients attempting to connect over TCP does in-fact drop the connection immediatelly instead of waiting for timeout

I second that. But maybe instead of using nat-lab it could be done in rust? I'm not sure if our code is suited to that, but starting our dns server implementation and trying to `TcpStream::connect` seems to me like a good integration test for this change.

crates/telio-dns/src/nameserver.rs
```diff
@@ -205,12 +207,16 @@ impl LocalNameServer {
...
+            PayloadRequestInfo::Udp {
+                ref mut dns_request,
+                ..
+            } => dns_request
+                .take()
+                .ok_or_else(|| String::from("Inexistent DNS request"))?,
+            _ => return Ok(Vec::new()),
+        };
+        let dns_request = Request::new(*dns_request, request_info.dns_source(), Protocol::Udp);
         telio_log_debug!("DNS request: {:?}", &dns_request);
```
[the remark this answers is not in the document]
mathiaspeters: It never gets here for other protocols. In the lines above there is 
```
match payload {
    Udp => get_info(),
    Tcp => early return,
}
```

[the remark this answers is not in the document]
tomaszklak: 🤦

crates/telio-dns/src/nameserver.rs
```diff
@@ -323,29 +346,43 @@ impl IpRequestInfo {
...
 
-struct UdpRequestInfo {
-    source_port: u16,
-    destination_port: u16,
-    dns_request: Option<MessageRequest>,
+enum PayloadRequestInfo {
+    Udp {
+        source_port: u16,
+        destination_port: u16,
+        dns_request: Option<Box<MessageRequest>>,
```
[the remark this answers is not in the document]
tomaszklak: treat it as a nitpick: I think it's better to silence clippy here instead of moving `MessageRequest` to a heap.

crates/telio-dns/src/nameserver.rs
mathiaspeters: fixed

crates/telio-dns/src/nameserver.rs
mathiaspeters: fixed

crates/telio-dns/src/nameserver.rs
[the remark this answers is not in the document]
mathiaspeters: You're right. The "big" variant is just 364 bytes so not that big

tomaszklak (approved):
+1

crates/telio-dns/src/nameserver.rs
```diff
@@ -254,18 +260,40 @@ impl LocalNameServer {
...
+                    ip.udp_checksum(udp_packet)
+                })?
+            }
+            IpNextHeaderProtocols::Tcp => {
+                let tcp_request = TcpPacket::new(ip.payload())
+                    .ok_or_else(|| String::from("Failed to build TcpPacket from request packet"))?;
+                PayloadRequestInfo::Tcp {
+                    source_port: tcp_request.get_source(),
+                    destination_port: tcp_request.get_destination(),
+                    next_sequence: tcp_request.get_sequence() + 1,
```
Jauler, line 277: :thinking: So there is interesting edge-case, as this sequence number will only work for TCP-SYN packets, or packets without any TCP data payload. Technically it is not possible to send any data without first establishing connection, which involved TCP SYN packet.

But I could imagine edge-case scenario, where a client was communicating with 100.64.0.2 before enabling DNS, and then telio-DNS gets enabled, and it first receives non-TCP-SYN packet in TCP connection, in which case, TCP RST may get ignored later. 

But this is going into so edge-casy territory, that probably adding all the extra complexity to handle this situation is net-negative instead of simply waiting until client-side application times-out. :thinking:

Jauler (approved):
+1.0

PR #687, opened 2024-07-16
[LLT-5421] Get rid of ipnetwork crate

stalowyjez:
### Problem
We are using two crates which have similar usecases, ipnet and ipnetwork, so I guess it is a good idea to get rid of less popular one.

### Solution
Get rid of ipnetwork crate

- [x] README.md is updated


gytsto (commented):
+1

Jauler (approved):
+1.5

PR #688, opened 2024-07-17
Added timestamp prefix to all lines in output from pytest process

tomaszklak:
### Problem
Not every output line produced by pytest (or its child processes) has a timestamp prefix.

### Solution
This solution prepends a timestamp to all lines produced by pytest (or its child processes). To make sure that we add a timestamp right after the line was produced both output buffering and input buffering has to be disabled (by setting `PYTHONUNBUFFERED` to `1` and `bufsize=0` respectively). Without this a 10-15s delay is observed.

- [x] README.md is updated


matislovas:
There's a bunch of `datetime.now()` in `print` in `telio.py`. Shouldn't we cleanup those as well ..?

matislovas:
+1

nat-lab/run_local.py
```diff
@@ -5,21 +5,55 @@
...
 from typing import List, Optional, Dict, Any
 
 PROJECT_ROOT = os.path.normpath(os.path.dirname(os.path.realpath(__file__)) + "/../..")
 
 TEST_TIMEOUT = 120
 
 
+def add_timestamps_to_output_lines(input_stream, output_stream):
+    for line in input_stream:
+        line = line.decode("ascii").rstrip()
```
packgron, line 19: Are you sure we will not have utf8 bytes here?

tomaszklak: To be honest I have no idea, but utf8 is indeed safer. I will change this to utf8.

packgron (approved):
+1

PR #690, opened 2024-07-17
decouple tcli from natlab

gytsto:
### Problem
tcli is not needed anymore in natlab since pyro has been introduced. and tcli is actually messing up things with logs and other stuff.

### Solution
remove tcli from natlab

- [x] README.md is updated


src/device.rs
```diff
@@ -799,6 +803,15 @@ impl Device {
         })
     }
 
+    pub fn receive_ping(&self) -> Result<String> {
```
matislovas, line 806: Wild idea: maybe we should expose/compile API methods, that are only used in tests, only in tests? I know that there was a libtelio space saving initiative :smile:

mathiaspeters: Uniffi doesn't like conditional compilation like that unfortunately

mathiaspeters:
Nit: the name of the changelog file should just be the ticket number. Since the file is empty it doesn't matter much, just to keep to convention

nat-lab/pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "nat-lab"
 
 [tool.black]
 preview = true
-exclude="tests/protobuf"
+exclude = "tests/protobuf,tests/uniffi"
```
mathiaspeters, line 6: Will this not make all the python tools skip `libtelio_remote.py` and `libtelio_proxy.py`?

src/device.rs
gytsto: Not that wild, but good idea to not expose those methods in release. But will create separate issue for this if you don't mind.

nat-lab/pyproject.toml
gytsto: you're right, will change so it would only skip generated part

src/device.rs
gytsto: LLT-5451

nat-lab/pyproject.toml
gytsto: fixed

mathiaspeters (commented):
+1

PR #692, opened 2024-07-17
Adjust conntracker limits to allow for port reuse

Jauler:
### Problem
In some cases client-side OS'es or processes may decide to reuse ports for two distinct connections, especially when UDP is involved. In this situation to conntracker in gateways or containers-under-test this will look like single UDP flow, just continuing. We are seeing this in-action in rare cases in CI/CD tests. For example today's failure:
```
[2024-07-17 05:15:04] FAILED tests/test_events.py::test_event_content_vpn_connection[WINDOWS_VM_1-WireguardGo-10.0.254.7] - AssertionError: conntracker reported out of limits {'stun': 1}
```

Keeping in mind port reuse possibility, limits could be adjusted for the worst case scenario, where all ports are reused (e.g. lower limits of `1`) and leave upper limits as-is. This PR implements it.

- [x] README.md is updated


tomaszklak (approved):
+2

PR #696, opened 2024-07-17
LLT-5349: Make docker port mapping optional

mathiaspeters:
### Problem
Port mapping doesn't work in CI. Either we specify both host and container port and get collisions, or we just specify container port and our test VMs become inaccessible. Until we can figure out the root cause, it would be good to skip port mapping in CI

### Solution
If the CI flag is set, do a simple find-and-replace in the compose file and remote the mapped port. Also cleaned up the code here around which ports are being used

- [x] README.md is updated


nat-lab/tests/utils/connection/ssh_connection.py
```diff
@@ -32,9 +31,5 @@ async def get_ip_address(self) -> tuple[str, str]:
         ip = self._connection._host  # pylint: disable=protected-access
         return (ip, ip)
 
-    async def mapped_ports(self) -> tuple[str, str]:
```
mathiaspeters, line 35: This can be removed since it's now the same as the default implementation in `connection.py`

nat-lab/docker-compose.yml
```diff
@@ -71,8 +71,7 @@ services:
     networks:
       cone-net-01:
         ipv4_address: 192.168.101.104
-    ports:
-      - "58001:58001"
+    ports: ["58001"]
```
packgron, line 74: Maybe add some comment warning that this part is modified by python, and link to the place

nat-lab/natlab.py
```diff
@@ -41,6 +41,14 @@ def start():
             "LIBTELIO_ENV_NAT_LAB_DEPS_TAG": LIBTELIO_ENV_NAT_LAB_DEPS_TAG,
         },
     )
+
+    if "GITLAB_CI" in os.environ:
+        with open("docker-compose.yml", "r", encoding="utf-8") as file:
+            filedata = file.read()
+        filedata = filedata.replace('ports: ["58001"]', "ports: []")
```
packgron, line 48: Would be nice if you add some check for 'ports: ["58001"]' to catch any modifications to it, and fail it nat-lab run with a meesage

nat-lab/tests/utils/connection/docker_connection.py
```diff
@@ -39,9 +39,7 @@ async def mapped_ports(self) -> tuple[str, str]:
         details = await self._container.show()
         ports = details["NetworkSettings"]["Ports"]
```
packgron, line 40: Don't entirely understand how is pyro port determined now

Jauler: pyro selects the port given by the OS, and then it is printed in [stdout](https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/uniffi/libtelio_remote.py#L193), and later collected by the other pytest code

nat-lab/natlab.py
mathiaspeters: I added a check with a `RuntimeError` in case the port mapping is not there

nat-lab/docker-compose.yml
mathiaspeters: Done

nat-lab/tests/utils/connection/docker_connection.py
mathiaspeters: I added a comment about this here: https://github.com/NordSecurity/libtelio/pull/696/files#diff-f1122f50402db2f1613c75eb1db5aed7bc4c9ed907b99f1bf4e5010345bfe6cdR629

packgron (approved):
+1.0

packgron (approved):

PR #698, opened 2024-07-17
Add tracing extra fields to logging callback

packgron:
### Problem
Until now some of information was lost then telio was used from ffi api, because non of the extra fields there send via telio log callback.

### Solution
Switched to tracing subscriber based implementation, as it correctly implements span logic, and will allow us to add additional systems in the future using layer approach.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5430
+++ b/.unreleased/LLT-5430
@@ -1,0 +1,1 @@
+Additional information fields are now shown in logs
--- a/src/ffi/logging.rs
+++ b/src/ffi/logging.rs
@@ -1,0 +1,224 @@
+use std::{
+    io::{self, ErrorKind},
+    str::from_utf8,
+    sync::{Arc, Mutex},
+};
+
+use tracing::{level_filters::LevelFilter, Subscriber};
+use tracing_subscriber::{
+    fmt::{self, FormatEvent, FormatFields, MakeWriter},
+    layer::SubscriberExt,
+    registry::LookupSpan,
+};
+
+use crate::{TelioLogLevel, TelioLoggerCb};
+
+/// Build a tracing subscriber for use in ffi
+pub fn build_subscriber(
+    log_level: crate::TelioLogLevel,
+    logger: Box<dyn TelioLoggerCb>,
+) -> impl Subscriber {
+    tracing_subscriber::registry()
+        .with(LevelFilter::from_level(log_level.into()))
+        .with(
+            fmt::layer()
+                .event_format(TelioEventFmt)
+                .with_ansi(false)
+                .with_writer(FfiCallback::new(logger)),
+        )
+}
+
+struct TelioEventFmt;
+
+impl<S, N> FormatEvent<S, N> for TelioEventFmt
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+    N: for<'a> FormatFields<'a> + 'static,
+{
+    fn format_event(
+        &self,
+        ctx: &fmt::FmtContext<'_, S, N>,
+        mut writer: fmt::format::Writer<'_>,
+        event: &tracing::Event<'_>,
+    ) -> std::fmt::Result {
+        let meta = event.metadata();
+        write!(
+            writer,
+            "{:?}:{} ",
+            meta.module_path().unwrap_or("<unknown module>"),
+            meta.line().unwrap_or(0),
+        )?;
+
+        ctx.format_fields(writer.by_ref(), event)?;
+
+        writeln!(writer)
+    }
+}
+
+pub struct FfiCallback {
+    callback: Arc<dyn TelioLoggerCb>,
+}
+impl FfiCallback {
+    fn new(logger: Box<dyn TelioLoggerCb>) -> Self {
+        Self {
+            callback: logger.into(),
+        }
+    }
+}
+
+impl<'a> MakeWriter<'a> for FfiCallback {
+    type Writer = FfiCallbackWriter;
+
+    fn make_writer(&self) -> Self::Writer {
+        unreachable!("`make_writer` should not be called, then `make_writer_for` is implemented")
+    }
+
+    fn make_writer_for(&self, meta: &tracing::Metadata<'_>) -> Self::Writer {
+        FfiCallbackWriter {
+            cb: self.callback.clone(),
+            level: (*meta.level()).into(),
+        }
+    }
+}
+
+pub struct FfiCallbackWriter {
+    cb: Arc<dyn TelioLoggerCb>,
+    level: TelioLogLevel,
+}
+
+impl io::Write for FfiCallbackWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // We can be certain that buf fully represents one full event
+        //
+        // See https://github.com/tokio-rs/tracing/blob/tracing-subscriber-0.3.18/tracing-subscriber/src/fmt/fmt_layer.rs#L975
+        //
+        // Additional test added to double protect from future breakages
+
+        let msg = from_utf8(buf)
+            // Trim could be avoided by removing writeln! in formatter
+            // But this makes it more univesal for the future, then we dicide to
+            // switch to other formats
+            .map(|msg| msg.trim().to_string())
+            .map_err(|e| io::Error::new(ErrorKind::InvalidData, e))?;
+
+        if let Some(filtered_msg) = filter_log_message(msg) {
+            self.cb
+                .log(self.level, filtered_msg)
+                .map_err(io::Error::other)?;
+        }
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+struct LogStatus {
+    string: String,
+    counter: u32,
+}
+
+lazy_static::lazy_static! {
+    static ref LAST_LOG_STATUS: Mutex<LogStatus> = {
+        Mutex::new(LogStatus{string: String::default(), counter: 0})
+    };
+}
+
+fn filter_log_message(msg: String) -> Option<String> {
+    let mut log_status = match LAST_LOG_STATUS.lock() {
+        Ok(status) => status,
+        Err(_) => {
+            return None;
+        }
+    };
+
+    if !log_status.string.eq(&msg) {
+        log_status.string = msg.clone();
+        log_status.counter = 0;
+        return Some(msg);
+    }
+
+    if log_status.counter > 0 && log_status.counter % 100 == 0 {
+        log_status.counter += 1;
+        return Some(format!("[repeated 100 times!] {}", msg));
+    }
+
+    if log_status.counter < 10 {
+        log_status.counter += 1;
+        return Some(msg);
+    }
+
+    log_status.counter += 1;
+    None
+}
+
+#[cfg(test)]
+mod test {
+
+    /// Added in external crate to avoid line changes as much as possible
+    use tracing::{debug, info, trace, warn};
+
+    use super::*;
+
+    use std::{
+        fmt::Debug,
+        sync::{Arc, Mutex},
+    };
+
+    use crate::{TelioLogLevel, TelioLoggerCb};
+
+    #[test]
+    fn test_trace_via_telio_cb() {
+        let start = line!() + 1;
+        let act = || {
+            trace!("first message"); // +1
+            debug!("second message"); // +2
+            info!("third\nmutiline\nmessage"); // +3
+            warn!(
+                n = 2,
+                extra = "extra info",
+                "fourth message with {}",
+                "info"
+            ); // +4
+        };
+        let mpath = module_path!();
+        let expected = [
+            (
+                TelioLogLevel::Debug,
+                format!("{:?}:{} second message", mpath, start + 2),
+            ),
+            (
+                TelioLogLevel::Info,
+                format!("{:?}:{} third\nmutiline\nmessage", mpath, start + 3),
+            ),
+            (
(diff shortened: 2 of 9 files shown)


src/ffi/logging/test.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::{
+    fmt::Debug,
+    sync::{Arc, Mutex},
+};
+
+use crate::{TelioLogLevel, TelioLoggerCb};
+
+#[test]
+fn test_trace_via_telio_cb() {
+    let expected = [(
```
mathiaspeters, line 15: It doesn't seem like this is used

packgron: Ah missed assert somehow

packgron: Forgot to commit

packgron: yeh forgot to remove, though originaly that it would be simpler to maintain, but also so  did not see tests/loger.rs,
but i prefer to have tests next to code

tests/logger.rs
```diff
@@ -26,7 +25,7 @@ mod test_module {
                 payload: String,
             ) -> FFIResult<()> {
                 assert!(matches!(log_level, telio::ffi_types::TelioLogLevel::Info));
-                assert_eq!(r#""logger::test_module":43 test message"#, payload);
+                assert_eq!(r#""logger::test_module":42 test message"#, payload);
```
matislovas, line 28: Where does this number comes from? Maybe some named const for it ..?

matislovas:
+1

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    if log_status.counter < 10 {
+        log_status.counter += 1;
+        return Some(msg);
+    }
+
+    log_status.counter += 1;
+    None
+}
```
stalowyjez, line 161: It seems that some code is just moved (like this function) - I think it makes things a lot simpler, when such moving is kept in separate commits.

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+            logs.push((level, payload));
+            Ok(())
+        }
+    }
+
+    impl Debug for Log {
+        fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            Ok(())
+        }
+    }
```
stalowyjez, line 235: As I understand you need it only for TelioLoggerCb trait?

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    #[derive(Default, Clone)]
+    struct Log(Arc<Mutex<Vec<(TelioLogLevel, String)>>>);
+    impl TelioLoggerCb for Log {
+        fn log(&self, level: TelioLogLevel, payload: String) -> crate::FFIResult<()> {
+            let mut logs = self.0.lock().expect("ok");
+            logs.push((level, payload));
+            Ok(())
+        }
+    }
```
stalowyjez, line 229: Nit: I think we usually keep such stuff before the tests, maybe it would be nice to follow that

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+            meta.line().unwrap_or(0),
+        )?;
+
+        ctx.format_fields(writer.by_ref(), event)?;
+
+        writeln!(writer)
+    }
+}
+
+pub struct FfiCallback {
```
stalowyjez, line 66: We have `FFIResult`, so I guess it would be nice to keep it consistent and name it `FFICallback`

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+            callback: logger.into(),
+        }
+    }
+}
+
+impl<'a> MakeWriter<'a> for FfiCallback {
+    type Writer = FfiCallbackWriter;
+
+    fn make_writer(&self) -> Self::Writer {
+        unreachable!("make_writer_for is used")
```
stalowyjez, line 81: This message sounds weird in this context imho,  maybe something like "make_writer is not expected to be used"?

tests/logger.rs
packgron: So this is the line of info!() trace at the top, for my test I made it relative, i will make it smarter here also

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+            // But this makes it more univesal for the future, then we dicide to
+            // switch to other formats
+            .map(|msg| msg.trim().to_string())
+            .unwrap_or_else(|err| format!("failed to decode trace as utf8 bytes: {err:?}"));
+
+        if let Some(filtered_msg) = filter_log_message(msg) {
+            let _ = self.cb.log(self.level, filtered_msg);
+        }
+
+        Ok(buf.len())
```
stalowyjez, line 116: Ehh, a nice amount of ugly boilerplate we have here

src/ffi/logging.rs
packgron: Yes that's correct, will take note next time.

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+        // See https://github.com/tokio-rs/tracing/blob/tracing-subscriber-0.3.18/tracing-subscriber/src/fmt/fmt_layer.rs#L975
+        //
+        // Additional test added to double protect from future breakages
+
+        let msg = from_utf8(buf)
+            // Trim could be avoided by removing writeln! in formatter
+            // But this makes it more univesal for the future, then we dicide to
+            // switch to other formats
+            .map(|msg| msg.trim().to_string())
+            .unwrap_or_else(|err| format!("failed to decode trace as utf8 bytes: {err:?}"));
```
stalowyjez, line 110: If this function returns an error maybe it would be nice to return it on failed decoding?

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+
+        if let Some(filtered_msg) = filter_log_message(msg) {
+            let _ = self.cb.log(self.level, filtered_msg);
```
stalowyjez, line 113: And maybe here as well?

tests/logger.rs
packgron: Ok so turning this is no an easy task here, how currently tests are design, this will have to be left as is. Once cell would i think just add additional confusion

src/ffi/logging.rs
packgron: Yes, but i see i can add a derive so will use it that

src/ffi/logging.rs
packgron: I prefer then tests are front and center, and glue is out of sight preferably, will leave as is

src/ffi/logging.rs
packgron: Well FFIResult is named againts rust naming standart :upside_down_face:

packgron: https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#fine-points

stalowyjez: Hmm, so maybe we could change that name then?

src/ffi/logging.rs
packgron: It's not that bad :D

src/ffi/logging.rs
packgron: fixed

.unreleased/LLT-5430
```diff
@@ -0,0 +1 @@
+Aditional information fields are now shown in logs
```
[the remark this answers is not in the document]
packgron: done

src/ffi/logging.rs
```diff
@@ -0,0 +1,236 @@
...
+                TelioLogLevel::Warning,
+                format!(
+                    "{:?}:{} fourth message with info n=2 extra=\"extra info\"",
+                    mpath,
+                    start + 4
+                ),
+            ),
+        ];
+        let logs = Log::default();
+        let sub = build(TelioLogLevel::Debug, Box::new(logs.clone()));
```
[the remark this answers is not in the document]
packgron: done

src/ffi/logging.rs
packgron: done

src/ffi/logging.rs
packgron: done

src/ffi/logging.rs
packgron: done

src/ffi/logging.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn format_event(
+        &self,
+        ctx: &fmt::FmtContext<'_, S, N>,
+        mut writer: fmt::format::Writer<'_>,
+        event: &tracing::Event<'_>,
+    ) -> std::fmt::Result {
+        let meta = event.metadata();
+        write!(
+            writer,
+            "{:?}:{} ",
```
mathiaspeters, line 55: Very minor nit: is `:?` needed here? It looks like it's a string that goes there and strings implement Display. Or am I missing something?

src/ffi/logging.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        let actual = logs.0.lock().unwrap().clone();
+
+        assert_eq!(&expected[..], &actual[..])
+    }
+
+    #[derive(Default, Clone, Debug)]
+    struct Log(Arc<Mutex<Vec<(TelioLogLevel, String)>>>);
+    impl TelioLoggerCb for Log {
+        fn log(&self, level: TelioLogLevel, payload: String) -> crate::FFIResult<()> {
```
stalowyjez, line 226: Hmm, it is still `FFI` instead of `Ffi`

tests/logger.rs
stalowyjez: Maybe making some const with that number can be considered, so it will be more apparent?

src/ffi/logging.rs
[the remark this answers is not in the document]
stalowyjez: idk, I still see "sub" here ^^'

src/ffi/logging.rs
```diff
@@ -0,0 +1,232 @@
...
+        // Additional test added to double protect from future breakages
+
+        let msg = from_utf8(buf)
+            // Trim could be avoided by removing writeln! in formatter
+            // But this makes it more univesal for the future, then we dicide to
+            // switch to other formats
+            .map(|msg| msg.trim().to_string())
+            .map_err(|e| io::Error::new(ErrorKind::InvalidData, e))?;
+
+        if let Some(filtered_msg) = filter_log_message(msg) {
```
stalowyjez, line 112: It seems that on `None` here we return buffer length, but log nothing

src/ffi/logging.rs
stalowyjez: Well, after reading all of this it seems to me that using this `telio-subscriber` crate introduces it more it removes and it it is less straightforward - tbh I prefer the previous implementation.

packgron: Previous implementation is fundamentaly broken, unless we implement full span id creation and management

src/ffi/logging.rs
packgron: This debug is not to break how it was displayed previously specificaly "some::path" instead of uquated "some::path"

src/ffi/logging.rs
packgron: Well it would error out looking into write_all that is used on tracing side: https://doc.rust-lang.org/src/std/io/mod.rs.html#1696-1708

src/ffi/logging.rs
packgron: I see lsp did not include #[cfg(test)] and docs also :|

tests/logger.rs
packgron: added a const

src/ffi/logging.rs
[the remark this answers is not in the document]
packgron: commited

src/ffi/logging.rs
```diff
@@ -0,0 +1,220 @@
...
+        let actual = logs.0.lock().unwrap().clone();
+
+        assert_eq!(&expected[..], &actual[..])
+    }
+
+    #[derive(Default, Clone, Debug)]
+    struct Log(Arc<Mutex<Vec<(TelioLogLevel, String)>>>);
+    impl TelioLoggerCb for Log {
+        fn log(&self, level: TelioLogLevel, payload: String) -> crate::FfiResult<()> {
+            let mut logs = self.0.lock().expect("ok");
```
stalowyjez, line 215: Nit: I don't feel like "ok" is the right message when sth panics

src/ffi/logging.rs
```diff
@@ -0,0 +1,220 @@
...
+use tracing::{level_filters::LevelFilter, Subscriber};
+use tracing_subscriber::{
+    fmt::{self, FormatEvent, FormatFields, MakeWriter},
+    layer::SubscriberExt,
+    registry::LookupSpan,
+};
+
+use crate::{TelioLogLevel, TelioLoggerCb};
+
+pub fn build(log_level: crate::TelioLogLevel, logger: Box<dyn TelioLoggerCb>) -> impl Subscriber {
```
stalowyjez, line 16: I think a public function needs a bit more descriptive name than `build`

src/ffi/logging.rs
packgron: done

src/ffi/logging.rs
packgron: done

stalowyjez (approved):
+1

PR #701, opened 2024-07-19
[LLT-5387] Print file copy progress only in case of copy failure

olekoliinyk:
### Problem
With current implementation `_file_copy_progress_handler` called each time when scp copy 16 KB which leads to enormous log file.

### Solution
Add file copy progress buffer and print it in case of copy failure.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/vm/windows_vm_util.py
+++ b/nat-lab/tests/utils/vm/windows_vm_util.py
@@ -8,7 +8,7 @@ from config import (
 )
 from contextlib import asynccontextmanager
 from datetime import datetime
-from typing import AsyncIterator
+from typing import AsyncIterator, List
 from utils.connection import Connection, SshConnection, TargetOS
 from utils.process import ProcessExecError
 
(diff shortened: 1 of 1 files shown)


Jauler (dismissed):
+1.0

nat-lab/tests/utils/vm/windows_vm_util.py
```diff
@@ -15,6 +15,7 @@
 VM_TCLI_DIR = LIBTELIO_BINARY_PATH_WINDOWS_VM
 VM_UNIFFI_DIR = UNIFFI_PATH_WINDOWS_VM
 VM_SYSTEM32 = "C:\\Windows\\System32"
+_file_copy_progress_buffer = []
```
tomaszklak, line 18: since `_copy_binaries` is `async` I think we should assume it can be run concurrently and thus this buffer shouldn't be global unique to `asyncssh.scp` call.

olekoliinyk: Do you mean that we better lock the buffer for each file?

tomaszklak: What I mean is that each call to `asyncssh.scp` should use a print callback that appends to reference to unique list.

tomaszklak: You can run this snippet to see a difference I'm talking about:

```python
import asyncio
import random

global_list = []

async def bad_async_function(id):
    for i in range(0,5):
        global_list.append(f"element from {id}: {i}")
        await asyncio.sleep(random.random())

async def bad_main():
    await asyncio.gather(
        bad_async_function(0),
        bad_async_function(1)
    )
    for l in global_list:
        print(l)

print("bad")
asyncio.run(bad_main())


async def good_async_function(id, l):
    for i in range(0,5):
        l.append(f"element from {id}: {i}")
        await asyncio.sleep(random.random())

async def good_main():
    l1 = []
    l2 = []

    async def wrapper(id, l):
        await good_async_function(id, l)

    await asyncio.gather(
        wrapper(0, l1),
        wrapper(1, l2),
    )
    for l in l1:
        print(l)
    print()
    for l in l2:
        print(l)

print("good")
asyncio.run(good_main())
```

tomaszklak (dismissed):
+1

tomaszklak (approved):

PR #702, opened 2024-07-19
LLT-5197: Make backup of events db in natlab before copying

mathiaspeters:
### Problem
For our tests, we were copying the original events db from the container to the host when checking for events. This way of doing it wouldn't take into account in-progress writes or other reasons why the db might be locked and sometimes led to the copied db being corrupted

### Solution
 As a way to get around it, we now let sqlite create a backup first, which does respect the db potentially being locked, and then we copy the backup

- [x] README.md is updated


Jauler (approved):
+1.0

packgron (approved):
+1.0

PR #705, opened 2024-07-22
Update trigger to 1.2.8 which uses same build image version for linux builder on linux and macos host

tomaszklak:
### Problem
Different images had been used when running nat-lab on linux and macos host.

### Solution
Now that the linux build is published in both amd64 and arm64 versions, same version can be used.

- [x] README.md is updated


Jauler (approved):
+2.0

PR #706, opened 2024-07-22
In case of ICMP only track type 0 and 8 connection in ConnectionTracker

tomaszklak:
### Problem
ConnectionTracker was tracking any ICMP 'connection'.

### Solution
In nat-lab, we want to only track type 0 and 8 so an additional filtering was added.

### DIFF
--- a/nat-lab/tests/test_connection_tracker.py
+++ b/nat-lab/tests/test_connection_tracker.py
@@ -1,0 +1,69 @@
+from utils.connection_tracker import parse_input, FiveTuple
+
+
+def test_connection_tracker_parse_input():
+    new_udp = parse_input(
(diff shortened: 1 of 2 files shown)


gytsto:
is 0 and 8 from ping tool?

Jauler (approved):
+1.5

Jauler:
> om ping tool?

type=0 is ICMP Echo Reply and type=8 is ICMP Echo Request. With this PR, we ignore ICMP error types in ping conntracker limits, which may cause some flakyness when ICMP Destination Unreachable returns due to unsynchronized startup order.

gytsto (approved):
+0.5

PR #707, opened 2024-07-22
Add adapter.up() retries for wireguard_nt

jjanowsk:
For some reason adapter.up() failes after sucessful adapter.create(). There is a suspicion in the team that it might be a race condition. For now let's add a few retries and investigate further the root casue.

- [x] README.md is updated


Jauler (approved):
+2.0

PR #710, opened 2024-07-22
LLT-5441 natlab preconditions

matislovas:
### Problem
Some nat-lab tests can be transfered as pre-conditions to running whole nat-lab test suite.

### Solution
Move `test_derp_client_message_forward` test to `pytest` pre-conditions.

- [x] README.md is updated


Jauler (dismissed):
+1.0

nat-lab/tests/conftest.py
```diff
@@ -1,12 +1,25 @@
...
 from utils.vm import windows_vm_util, mac_vm_util
 
+DERP_SERVER_1_ADDR = "http://10.0.10.1:8765"
+DERP_SERVER_2_ADDR = "http://10.0.10.2:8765"
+DERP_SERVER_3_ADDR = "http://10.0.10.3:8765"
+DERP_SERVER_1_SECRET_KEY = "yBTYHj8yPlG9VtMYMwJSRHdzNdyAlVXGc6X2xJkjfHQ="
+DERP_SERVER_2_SECRET_KEY = "2NgALOCSKJcDxwr8MtA+6lYbf7b98KSdAROGoUwZ1V0="
+
+SETUP_CHECKS = [
+    ("setup_check_interderp", 10.0),
```
gytsto, line 20: can't we just add function here, instead of name? wouldn't need to use `globals()` later on  :thinking:

matislovas: I was having the impression, that it is not possible in python, but will try

matislovas: Done

PR #712, opened 2024-07-24
[LLT-5453] Fix ping race which may cause two ping requests being sent-out

Jauler:
### Problem
There is a small race between ping utility receiving first response, and [.wait_for_next_ping](https://github.com/NordSecurity/libtelio/blob/v5.0.0-rc2/nat-lab/tests/utils/ping.py#L54) being called. 

So all Ping utility uses has this pattern, when using ping utility:
```python
        async with Ping(<connection>, <ping_target_ip_address>).run() as ping:
            await ping.wait_for_next_ping()
```

In this case, when [.run()](https://github.com/NordSecurity/libtelio/blob/v5.0.0-rc2/nat-lab/tests/utils/ping.py#L60) is called, `ping` process is started, and then after some time - [wait_for_next_ping()](https://github.com/NordSecurity/libtelio/blob/v5.0.0-rc2/nat-lab/tests/utils/ping.py#L55) is called, which may `.clear()` first ping response. 

In a test like [this](https://github.com/NordSecurity/libtelio/blob/v5.0.0-rc2/nat-lab/tests/test_events_link_state.py#L156) the race may prolong test execution significantly even hitting timeouts.

### Solution
Introduce a function `wait_for_any_ping()` which would not clear events at the start. Next, provide a wrapper function, which would do the pinging in one go. And refactor all tests to use this new function.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/ping.py
+++ b/nat-lab/tests/utils/ping.py
@@ -12,6 +12,13 @@ from utils.router import IPProto, REG_IPV6ADDR, get_ip_address_type
 # It should work for Linux, Windows and Mac.
 
 
+async def ping(
+    connection: Connection, ip: str, timeout: Optional[float] = None
+) -> None:
+    async with Ping(connection, ip).run() as ping_process:
+        await ping_process.wait_for_any_ping(timeout)
+
+
 class Ping:
     _ip: str
     _ip_proto: IPProto
@@ -53,6 +60,10 @@ class Ping:
 
     async def wait_for_next_ping(self, timeout: Optional[float] = None) -> None:
         self._next_ping_event.clear()
+        await asyncio.wait_for(self._next_ping_event.wait(), timeout)
+        self._next_ping_event.clear()
+
+    async def wait_for_any_ping(self, timeout: Optional[float] = None) -> None:
         await asyncio.wait_for(self._next_ping_event.wait(), timeout)
         self._next_ping_event.clear()
 
--- a/nat-lab/tests/test_connection_states.py
+++ b/nat-lab/tests/test_connection_states.py
@@ -4,7 +4,7 @@ from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_mesh_nodes
 from utils.connection_tracker import ConnectionLimits
 from utils.connection_util import generate_connection_tracker_config, ConnectionTag
-from utils.ping import Ping
+from utils.ping import ping
 
 
 @pytest.mark.asyncio
@@ -98,5 +98,4 @@ async def test_connected_state_after_routing(
         await client_alpha.connect_to_exit_node(beta.public_key)
         await client_alpha.disconnect_from_exit_node(beta.public_key)
 
-        async with Ping(conn_alpha.connection, beta.ip_addresses[0]).run() as ping:
-            await ping.wait_for_next_ping()
+        await ping(conn_alpha.connection, beta.ip_addresses[0])
--- a/nat-lab/tests/test_derp_connect.py
+++ b/nat-lab/tests/test_derp_connect.py
@@ -7,7 +7,7 @@ from helpers import SetupParameters, setup_mesh_nodes
 from telio import State
 from typing import List
 from utils.connection_util import ConnectionTag
-from utils.ping import Ping
+from utils.ping import ping
 
 DERP1_IP = str(DERP_PRIMARY["ipv4"])
 DERP2_IP = str(DERP_SECONDARY["ipv4"])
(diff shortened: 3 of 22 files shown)


packgron (commented):
+1.0

gytsto (approved):
+1

nat-lab/tests/utils/ping.py
```diff
@@ -52,7 +52,6 @@ async def execute(self) -> None:
         await self._process.execute(stdout_callback=self.on_stdout)
 
     async def wait_for_next_ping(self, timeout: Optional[float] = None) -> None:
-        self._next_ping_event.clear()
```
tomaszklak, line 55: I think this will break this situation:
```python
async with Ping(<connection>, <ping_target_ip_address>).run() as ping:
           await ping.wait_for_next_ping()
           await do_some_other_stuff_for_some_time()
           await break_network()
           await fix_network();
           await ping.wait_for_next_ping()
```

The second `wait_for_next_ping` will succeed regardless if the `fix_network()` actually worked - there will be messages in the buffer from the time `do_some_other_stuff_for_some_time` was running.

Jauler: yeah, I specifically checked all uses of `Ping(..)` to verify that this is not occurring. Although it may easy to introduce such scenario.

Maybe to make the interface "cleaner" we should merge `.run()` and `.wait_for_next_ping()`? :thinking:

tomaszklak: I think something should be added to this PR that will make it impossible to make such a mistake. I like the idea of merging those 2 methods but I worry that it might quite some time to do. If that is the case, maybe a cheaper way to go would be to stop/kill the ping process the first time the wait_for_next_ping is called and/or throw an exception the second time it is called.

gytsto: what if you were to make `run()` wait for first ping like this (not touching the code of `ping.wait_for_next_ping()`:
```
    @asynccontextmanager
    async def run(self) -> AsyncIterator["Ping"]:
        async with self._process.run(stdout_callback=self.on_stdout):
            await self._next_ping_event.wait()
            yield self
```

and then in the code we would do something like:
```
async with Ping(<connection>, <ping_target_ip_address>).run() as ping:
           await ping
```
and in the other scenarios, where @tomaszklak mentioned, we could do:
```
async with Ping(<connection>, <ping_target_ip_address>).run() as ping:
           await ping
           await do_some_other_stuff_for_some_time()
           await break_network()
           await fix_network();
           await ping.wait_for_next_ping()
```
**Note** Haven't tested it myself, not sure atm if `await ping` wouldn't just break async manager, but just an idea :thinking:

Jauler: I did some refactoring :point_right: :point_left:

nat-lab/tests/utils/ping.py
```diff
@@ -12,6 +12,13 @@
...
 
 
+async def ping(
+    connection: Connection, ip: str, timeout: Optional[float] = None
+) -> None:
+    async with Ping(connection, ip).run() as ping:
+        await ping.wait_for_any_ping()
+
+
 class Ping:
```
tomaszklak, line 22: nit: Could we move whole `Ping` class definition inside of the `ping` function? This way it would be impossible to use by accident.

tomaszklak (approved):
+1

nat-lab/tests/utils/ping.py
```diff
@@ -12,6 +12,13 @@
 # It should work for Linux, Windows and Mac.
 
 
+async def ping(
+    connection: Connection, ip: str, timeout: Optional[float] = None
+) -> None:
+    async with Ping(connection, ip).run() as ping:
+        await ping.wait_for_any_ping()
```
gytsto, line 19: timeout not passed

Jauler: good catch

nat-lab/tests/utils/ping.py
Jauler: I have left the class exposed, as conntracker is using "long-running" pings. But instead of modifying `wait_for_next_ping()` I have created new method called `wait_for_any_ping()`. 

This way we can still have this long-running pings if needed exposed via `Ping` class, but for most use-cases simply using `ping()` will be sufficient function.

tomaszklak (dismissed):

gytsto (dismissed):

tomaszklak (approved):

gytsto (approved):

PR #715, opened 2024-07-24
Remove native interfaces at the beginning of each test

tomaszklak:
### Problem
Sometimes native interfaces survive until the next test and can impact them.

### Solution
Before each test, remove all native wireguard interfaces.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_cleanup.py
+++ b/nat-lab/tests/test_cleanup.py
@@ -1,0 +1,52 @@
+import pytest
+import telio
+from contextlib import AsyncExitStack
+from helpers import SetupParameters, setup_environment
+from utils.connection_util import ConnectionTag, new_connection_raw
+from utils.vm.windows_vm_util import _get_network_interface_tunnel_keys
+
+
+@pytest.mark.asyncio
+@pytest.mark.windows
+@pytest.mark.parametrize(
+    "adapter_type, name",
+    [
+        (telio.AdapterType.WireguardGo, "Wintun Userspace Tunnel"),
+        (telio.AdapterType.WindowsNativeWg, "WireGuard Tunnel"),
+    ],
+)
+async def test_get_network_interface_tunnel_keys(adapter_type, name) -> None:
+    async with AsyncExitStack() as exit_stack:
+        connection = await exit_stack.enter_async_context(
+            new_connection_raw(ConnectionTag.WINDOWS_VM_1)
+        )
+        assert [] == await _get_network_interface_tunnel_keys(connection)
+        _env = await exit_stack.enter_async_context(
+            setup_environment(
+                exit_stack,
+                [
+                    SetupParameters(
+                        connection_tag=ConnectionTag.WINDOWS_VM_1,
+                        adapter_type=adapter_type,
+                    ),
+                    SetupParameters(
+                        connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_2,
+                        adapter_type=telio.AdapterType.LinuxNativeWg,
+                    ),
+                ],
+            )
+        )
+
+        # This function is used during test startup to remove interfaces
+        # that might have managed to survive the end of the previous test.
+        keys = await _get_network_interface_tunnel_keys(connection)
+        assert [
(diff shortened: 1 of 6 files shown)


nat-lab/tests/utils/vm/windows_vm_util.py
```diff
@@ -138,3 +141,21 @@ async def _copy_binaries(
...
+        "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}",
+        "/s",
+        "/f",
+        "DriverDesc",
+    ]).execute()
+
+    lines = result.get_stdout().splitlines()
+    keys = []
+    for i in range(len(lines)):  # pylint: disable=C0200
+        if "WireGuard Tunnel" in lines[i]:
```
gytsto, line 159: Should we care about `Wintun` here also?

nat-lab/tests/test_cleanup.py
```diff
@@ -0,0 +1,35 @@
...
+                ],
+            )
+        )
+
+        connection_alpha = env.connections[0].connection
+
+        # This function is used during test startup to remove interfaces
+        # that might have managed to survive the end of the previous test.
+        assert [
+            "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}\\0005"
```
gytsto, line 34: I'd check for name or something in DriverDesc, to be sure that it is WireguardNT adapter.

gytsto (commented):
In general, code and idea looks OK. But I think we are hidding some other issue, why this tunnel survives until next test, when libtelio instance should be already closed.

Also, maybe it would be nice to expand `test_wg_adapter_cleanup` with WireguardNT and check if it reproduces the same as WireguardGo. Maybe WireguardNT handles orphaned adapter same as our current WireguardGo implementation (cleans them itself, before creating new one)

nat-lab/tests/utils/vm/windows_vm_util.py
```diff
@@ -138,3 +141,21 @@ async def _copy_binaries(
...
+        "query",
+        "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}",
+        "/s",
+        "/f",
+        "DriverDesc",
+    ]).execute()
+
+    lines = result.get_stdout().splitlines()
+    keys = []
+    for i in range(len(lines)):  # pylint: disable=C0200
```
gytsto, line 158: Could you use `enumerate` instead of disabling pylint? :smile:

nat-lab/tests/utils/vm/windows_vm_util.py
```diff
@@ -44,9 +44,12 @@ async def new_connection(
     ) as ssh_connection:
         connection = SshConnection(ssh_connection, TargetOS.Windows)
 
+        keys = await _get_network_interface_tunnel_keys(connection)
+        for key in keys:
+            await connection.create_process(["reg", "delete", key, "/F"]).execute()
```
gytsto, line 49: nit: Don't think this is the best place to put this here. Such things shouldn't be done during test execution. Maybe not during this PR, but it should be moved to test setup/teardown :thinking:

tomaszklak: I will create a jira ticket to move this to test setup.

tomaszklak:
> In general, code and idea looks OK. But I think we are hidding some other issue, why this tunnel survives until next test, when libtelio instance should be already closed.
> 
> Also, maybe it would be nice to expand `test_wg_adapter_cleanup` with WireguardNT and check if it reproduces the same as WireguardGo. Maybe WireguardNT handles orphaned adapter same as our current WireguardGo implementation (cleans them itself, before creating new one)

As discussed on the call, I've added wintun cleanup.

gytsto (approved):
+1

matislovas (approved):
+1

PR #718, opened 2024-07-26
Log conntrack events in test logs

Jauler:
### Problem
Seems like previous root cause for `test_session_keeper` failures was incorrect, and we still have flakyness in the test. Therefore this PR allows to collect more information from the tests, to aid root-cause search.

- [x] README.md is updated


packgron (approved):
+2.0

PR #723, opened 2024-07-29
[LLT-5467] Print process Kill ID

stalowyjez:
### Problem
Currently the Kill ID for the process is not printed, so it is sometimes hard to track the processes

### Solution
Print the process Kill ID when it is created


mathiaspeters (approved):
+2

PR #725, opened 2024-07-30
Collect dmesg output

tomaszklak:
### Problem
Something is killing iptables and we don't know who it is and why.

### Solution
Collect dmesg logs with the intention to check if this is out of memory condition that is causing OOM killer to kill iptables.

- [x] README.md is updated

### DIFF
--- a/nat-lab/natlab.py
+++ b/nat-lab/natlab.py
@@ -33,31 +33,52 @@ def run_command_with_output(command, hide_output=False):
 
 
 def start():
-    original_port_mapping = 'ports: ["58001"]'
-    disabled_port_mapping = "ports: []"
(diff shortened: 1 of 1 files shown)


Jauler (changes requested):
Note that in terms of `iptables` killing failures, nat-lab tests, including telio.py are not executed yet. 

we would need to update `nat-lab/natlab.py` as well :point_right: :point_left:

tomaszklak:
> Note that in terms of `iptables` killing failures, nat-lab tests, including telio.py are not executed yet.
> 
> we would need to update `nat-lab/natlab.py` as well 👉 👈

You are right 🤦 , will move this to natlab.py in a second!

Jauler (dismissed):
+2.0

Jauler (approved):

PR #728, opened 2024-07-31
Fix race condition in test_mesh_reconnect

Jauler:
### Description
There is a time gap when meshnet is started in `test_mesh_reconnect` and test starts to wait for derp/node events coming from libtelio. If libtelio fires an even in this time gap when meshnet is already started, but test is still not waiting for event - test will attempt to wait for **second** derp/node event which will never come. This PR removes this time gap.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_reconnections.py
+++ b/nat-lab/tests/test_reconnections.py
@@ -3,6 +3,7 @@ import pytest
 import telio
 from contextlib import AsyncExitStack
 from helpers import setup_mesh_nodes, SetupParameters
+from utils import asyncio_util
 from utils.connection_tracker import ConnectionLimits
 from utils.connection_util import generate_connection_tracker_config, ConnectionTag
 from utils.ping import ping
@@ -109,12 +110,17 @@ async def test_mesh_reconnect(
             await ping(beta_connection, alpha.ip_addresses[0], 15)
 
         await client_alpha.simple_start()
-        await client_alpha.set_meshmap(api.get_meshmap(alpha.id))
 
-        await asyncio.gather(
-            client_alpha.wait_for_event_peer(beta.public_key, [telio.State.Connected]),
-            client_alpha.wait_for_event_on_any_derp([telio.State.Connected]),
-        )
+        async with asyncio_util.run_async_context(
+            asyncio.gather(
+                client_alpha.wait_for_event_peer(
+                    beta.public_key, [telio.State.Connected]
+                ),
+                client_alpha.wait_for_event_on_any_derp([telio.State.Connected]),
+            ),
+        ) as event:
+            await client_alpha.set_meshmap(api.get_meshmap(alpha.id))
+            await event
 
         await asyncio.gather(
             client_alpha.wait_for_state_peer(beta.public_key, [telio.State.Connected]),


nat-lab/tests/test_reconnections.py
```diff
@@ -109,9 +109,9 @@ async def test_mesh_reconnect(
             await ping(beta_connection, alpha.ip_addresses[0], 15)
 
         await client_alpha.simple_start()
-        await client_alpha.set_meshmap(api.get_meshmap(alpha.id))
 
         await asyncio.gather(
+            client_alpha.set_meshmap(api.get_meshmap(alpha.id)),
```
gytsto, line 114: This doesn't really fix the race condition. But on the other hand, starting to wait for event before action, might fix the race condition:

```
        async with run_async_context(
            asyncio.gather(
                client_alpha.wait_for_event_peer(
                    beta.public_key, [telio.State.Connected]
                ),
                client_alpha.wait_for_event_on_any_derp([telio.State.Connected]),
            )
        ) as events:
            await client_alpha.set_meshmap(api.get_meshmap(alpha.id))
            await events
```

mathiaspeters:
Have you verified that this works? I'm asking because I think it's a common pattern to first set the meshmap and afterwards wait for an event, and I might be mistaken but I'm pretty sure when you wait for a specific event (in this case connected), it doesn't just look at events from the moment the `wait_for_*` was called, but rather it checks the events since the previous time `wait_for_*` was called, which in this case would mean that it would see the first event and wouldn't be waiting. I could be wrong though so if you can somewhat reliably reproduce the issue before this fix and can't repro after, I'll approve the PR

gytsto:
> Have you verified that this works? I'm asking because I think it's a common pattern to first set the meshmap and afterwards wait for an event, and I might be mistaken but I'm pretty sure when you wait for a specific event (in this case connected), it doesn't just look at events from the moment the `wait_for_*` was called, but rather it checks the events since the previous time `wait_for_*` was called, which in this case would mean that it would see the first event and wouldn't be waiting. I could be wrong though so if you can somewhat reliably reproduce the issue before this fix and can't repro after, I'll approve the PR

I think you are mixing up state and event. Before uniffi changes, wait_for_event would waiy until new event is generated without looking to previously generated ones. State would return immidiately if it was in that state before even before the action of set_meshmap. But starting to wait for event, before setting meshmap, would actually solve the race condition.

nat-lab/tests/test_reconnections.py
Jauler: oh nice, thanks for taking a look :pray: 

Fixed.

nat-lab/tests/test_reconnections.py
```diff
@@ -109,17 +110,17 @@ async def test_mesh_reconnect(
...
 
-        await asyncio.gather(
-            client_alpha.wait_for_event_peer(beta.public_key, [telio.State.Connected]),
-            client_alpha.wait_for_event_on_any_derp([telio.State.Connected]),
-        )
-
-        await asyncio.gather(
-            client_alpha.wait_for_state_peer(beta.public_key, [telio.State.Connected]),
-            client_beta.wait_for_state_peer(alpha.public_key, [telio.State.Connected]),
-        )
```
gytsto, line 122: nit: Would still be nice to check if state is reported correctly on both sides after event is generated, but up to you

gytsto (dismissed):
+2, other than one nit

nat-lab/tests/test_reconnections.py
gytsto: didn't see your comment, this is useless :D

mathiaspeters (approved):

nat-lab/tests/test_reconnections.py
Jauler: generally `beta` is likely will not generate any events. So this is why I thought that asserting on them may be not the wisest thing

Jauler: but `.wait_for_state_peer` is safe probably, so I will readd it :D (changed my mind)

PR #730, opened 2024-07-31
LLT-5469 Feature config tests

mathiaspeters:
### Problem
One of the windows peeps mentioned recently while doing in-branch testing that they got incorrect values when deserializing a json feature config with the FFI function we provides for doing the deserialization, and nothing seemed to be wrong with their json. As a way to check that we aren't doing anything weird on our side, I wanted to update the test suite for the feature config.

### Solution
Add some new more comprehensive and understandable tests, and then remove tests that are either covered by the new tests or that are testing that serde itself works, which we shouldn't do in my opinion. The new tests revealed that deserializing `{}` produced different results compared to `Feature*::default()` in some cases, and those were fixed here.

For now there is a bit more code with manually implementing default for more structures and there are likely places where we can use `#[serde(default)]` instead of having functions to provide default values, but there is another PR that will hopefully be merged soon that introduces `SmartDefaults`. After that PR has been merged, we can get rid of a lot more of that kind of boilerplate, so it is untouched here in favor of a single PR later to fix it. 

Note: This PR has two separate commits that are best looked at separately. The first commit updates the tests and has actual changes that need to be reviewed. The second commit has no actual changes to the logic, it just reorders the code into a more logical order.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -16,8 +16,88 @@ pub const DEFAULT_PERSISTENT_KEEPALIVE_PERIOD: u32 = 25;
 /// Default keepalive period for direct peers
 pub const DEFAULT_DIRECT_PERSISTENT_KEEPALIVE_PERIOD: u32 = 5;
 
+/// Endpoint polling interval
+pub const DEFAULT_ENDPOINT_POLL_INTERVAL_SECS: u64 = 25;
+
 /// Type alias for UniFFI
 pub type EndpointProviders = HashSet<EndpointProvider>;
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, SmartDefault)]
+/// Encompasses all of the possible features that can be enabled
+pub struct Features {
+    /// Additional wireguard configuration
+    #[serde(default, deserialize_with = "FeatureWireguard::default_on_null")]
+    pub wireguard: FeatureWireguard,
+    /// Nurse features that can be configured for QoS
+    #[serde(default = "Features::default_nurse")]
+    #[default(Features::default_nurse())]
+    pub nurse: Option<FeatureNurse>,
+    /// Event logging configurable features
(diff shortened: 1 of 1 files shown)


crates/telio-model/src/features.rs
```diff
@@ -361,7 +361,7 @@ impl Default for FeatureValidateKeys {
 }
 
 /// Control which battery optimizations are turned on
-#[derive(Clone, Debug, Default, PartialEq, Eq, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize)]
 pub struct FeatureEndpointProvidersOptimization {
     /// Controls whether Stun endpoint provider should be turned off when there are no proxying peers
     #[serde(default = "FeatureEndpointProvidersOptimization::default_optimization_variant")]
```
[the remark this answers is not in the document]
mathiaspeters: I mention in the PR description that we can fix that in a separate PR to keep this PR clean, after the other PR that introduces `SmartDefaults` has been merged

crates/telio-model/src/features.rs
```diff
@@ -596,857 +616,374 @@ impl Features {
...
                 }
             },
             "is_test_env": true,
-            "derp":
-            {
-                "tcp_keepalive": 1,
-                "derp_keepalive": 2,
-                "enable_polling": true
+            "derp": {
+                "tcp_keepalive": 13,
```
Hasan6979, line 681: Just wondering. Why did you change the values here ?

Hasan6979 (commented):
lgtm +1

crates/telio-model/src/features.rs
mathiaspeters: I created a new test and decided to make the numbers always increase by one, and then removed the old test without really looking at what values were there

stalowyjez (approved):
Looks good, +1

PR #734, opened 2024-08-01
Natlab: scp instead of type

jjanowsk:
Do not use type command over ssh to copy file. Instead use scp

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -1008,13 +1008,12 @@ class Client:
             await asyncio.sleep(1)
 
     async def get_log(self) -> str:
-        process = (
-            self._connection.create_process(["type", "tcli.log"])
+        log_path = (
+            "tcli.log"
             if self._connection.target_os == TargetOS.Windows
-            else self._connection.create_process(["cat", "./tcli.log"])
+            else "./tcli.log"
         )
-        await process.execute()
-        return process.get_stdout()
+        return await self._connection.read_text_file(log_path)
 
     async def get_system_log(self) -> Optional[str]:
         """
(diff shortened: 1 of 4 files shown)


nat-lab/tests/telio.py
```diff
@@ -1008,13 +1008,12 @@ async def wait_for_log(self, what: str, case_insensitive: bool = True) -> None:
             await asyncio.sleep(1)
 
     async def get_log(self) -> str:
```
Jauler, line 1010: I feel like this would still generate `utf-8` error, as we are still reading log file contents into memory and parsing it as `utf-8` [here](https://github.com/NordSecurity/libtelio/pull/734/files#diff-f6b327e0eb73a3af12c4194bd89f5ddca97738b44f210eaa4c935e277bfce31fR36) (unless I have misread the code.

What is more, to my understanding, there is no need to read this file into memory, we just need to put it in the right place in `logs` directory. So maybe we can `scp` it and then `mv`'it into the right place?

jjanowsk: Yes that's intentional. It would still generate utf-8 error if the log was not properly utf-8 encoded. But the current suspicion is that the problem is with how ssh was handing the transfer of this file. The error only appered on windows and I have a feeling it may have something to do with how type's output is handled on windows side.

For your second question the get_log function is used in some tests to wait on the appearance of some messages in the log, so it isn't only used at the end of the test to collect the logfile. 

So the question is what do you want to achieve because I'm getting mixed opinions. If you want to just stop the tests from crashing then the proper way was to use error="replace" functionality in the sshconnection constructor to basically replace non utf-8 characters with '?' instead of raising exceptions. But here I hope to merge it and see over the next few days if the problem disappear because I have a hunch that it will

Jauler: Yeah, I forgot about tests querying logs. 

Ok lets attempt your variant and see whether we still have utf-8 issues.

Jauler (approved):
+2.0

PR #735, opened 2024-08-01
[LLT-5474] Cleanup natlab processes in pretest setup hook

Jauler:
### Problem
When running tests, it is important to ensure that environment is in a state expected by tests. One thing is that no leftover processes must exist from previous tests. Luckily all processes started by nat-lab has `KILL_ID` set in environment variable, used exactly for this purpose - killing them, when they become unecessary.

This PR adds a small script (must be run with elevated privileges inside of testing environment due to fact, that processes being run inside of docker containers has root privileges). which kills nat-lab processes pre-test run.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/cleanup_natlab_processes
+++ b/nat-lab/bin/cleanup_natlab_processes
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+echo "Executing natlab process cleanup script"
+
+for pid in $(ps -e -o pid=); do
+    # Skip non-testing processes
+    if ! grep --null-data --text KILL_ID /proc/${pid}/environ; then
+        continue
+    fi
+
+    # KILL
+    cmd=$(tr -d '\000' < /proc/${pid}/cmdline || echo "N/A")
+    echo "$(date) Cleaning up process: ${pid} ${cmd}"
+    kill "${pid}" || echo "Kill command failed with $?, perhaps processs already exited?"
+done
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -1,4 +1,5 @@
 import asyncio
+import os
 import pytest
 import subprocess
 import time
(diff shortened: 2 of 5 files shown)


tomasz-grz (dismissed):
LGTM +1.0

nat-lab/bin/cleanup_natlab_processes
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+echo "Executing natlab process cleanup script"
+
+for pid in $(ps -e -o pid=); do
+    # Skip non-testing processes
+    if ! grep --null-data --text KILL_ID /proc/${pid}/environ; then
```
gindoslaw, line 7: Test if `pid` exist. 
Kill command may take some time - so that here is race condition (starting from second loop iteration), grep may drop here 'no search file or directory'.

Entire loop can't be all the time safe - but lets reduce possibility of race conditions

nat-lab/bin/cleanup_natlab_processes
```diff
@@ -0,0 +1,15 @@
...
+        continue
+    fi
+
+    # KILL
+    cmd=$(cat /proc/${pid}/cmdline | sed 's/\x00//g' || echo "N/A")
```
gindoslaw, line 12: just 
sed $pattern $file

nat-lab/tests/conftest.py
```diff
@@ -159,6 +160,35 @@ async def perform_setup_checks() -> bool:
...
+    # If someone really wants cleanup - one can set
+    # ENABLE_NATLAB_PROCESS_CLEANUP envron variable
+    if (
+        not "GITLAB_CI" in os.environ
+        and not "ENABLE_NATLAB_PROCESS_CLEANUP" in os.environ
+    ):
+        return
+
+    cleanup_script_path = os.path.join(
+        os.path.dirname(__file__), "../bin/cleanup_natlab_processes"
```
gindoslaw, line 177: For the purpose of notice - it is path traversal, but so strict and environment is so restricted that I don't see window of opportunity here.

gindoslaw (approved):
+1

nat-lab/bin/cleanup_natlab_processes
tomaszklak: And `sed 's/\x00//g' $file` is an overkill when `tr -d '\000' < $file` is enough ;)

nat-lab/bin/cleanup_natlab_processes
Jauler: good point, I see two races here:
* process exits while loop is running, which would make `kill` fail. I missed this case, as it would exit the loop in such scenario. I will fix it.
* The other is that PIDs, at least theoretically, may be reused, if they wrap around. This one, IMHO, is pretty hard to deal with without introducing overkill solutions, and I am considering to leave it unsolved.

nat-lab/tests/conftest.py
Jauler: I do not see path traversal here, there is no user-controllable input (well any other user than the one who is running the script itself). Could you elaborate?

nat-lab/bin/cleanup_natlab_processes
Jauler: updated

nat-lab/bin/cleanup_natlab_processes
Jauler: fixed

gindoslaw (approved):

gindoslaw (approved):

nat-lab/tests/conftest.py
gindoslaw: fdsa

gindoslaw (approved):

gindoslaw (approved):

tomasz-grz (approved):

PR #736, opened 2024-08-01
Fix tcli `mesh set-ip` for macOS

tomasz-grz:
### Problem
Setting `mesh set-ip` on macOS fails. 

### Solution
The correct command on macOS is `ifconfig <NAME> inet <IP_ADDRESS>/10 <IP_ADDRESS>`. In addition for meshnet we also need to add the correct routes to the interface.

- [x] README.md is updated


tomaszklak (approved):
+2

dfetti (commented):
+1.0

PR #737, opened 2024-08-01
[LLT-5476] Remove unnecessary Default impls

stalowyjez:
### Problem
There is a bunch implementations of Default trait which we may get rid of easily, so this PR tries to do that.

### Solution
Get rid of them

- [x] README.md is updated


crates/telio-model/src/features.rs
```diff
@@ -19,46 +20,26 @@ pub const DEFAULT_DIRECT_PERSISTENT_KEEPALIVE_PERIOD: u32 = 5;
...
-impl Default for FeaturePersistentKeepalive {
-    fn default() -> Self {
-        Self {
-            vpn: Some(DEFAULT_PERSISTENT_KEEPALIVE_PERIOD),
-            direct: DEFAULT_DIRECT_PERSISTENT_KEEPALIVE_PERIOD,
-            proxying: Some(DEFAULT_PERSISTENT_KEEPALIVE_PERIOD),
-            stun: Some(DEFAULT_PERSISTENT_KEEPALIVE_PERIOD),
-        }
-    }
-}
```
LukasPukenis, line 60: Shouldn't these constants also be deleted in this PR ? :)

mathiaspeters:
I think this is great, but I would like to have #730 merged before this, because I think the current set of tests is a bit lacking and I want to be sure this doesn't break anything

stalowyjez:
@mathiaspeters sure :)

LukasPukenis:
I left a comment about probably-unused constants, howevr I like the approach of inlining the defaults +1

mathiaspeters:
+1 from me as well

packgron (approved):
+1 Awesome!

PR #738, opened 2024-08-02
Natlab: Do not crash on utf-8 decoding

jjanowsk:
It seems that for some reason some log files from Windows machines contain bytes that are not valid utf-8 characters. Let's not crash in such case and additionally let's dump those bytes to be able to analyze what's happening

- [x] README.md is updated


mathiaspeters (dismissed):
+1

packgron (approved):
+1

PR #742, opened 2024-08-05
copy all binaries when running locally during collection

gytsto:
### Problem
When running tests locally 1 by 1, not always there will be test name, that would include `["windows', "mac"]` strings in it. If tests name doesn't contain those string, binaries won't be copied to VMs. Doesn't apply to CI, since it is running all tests, which certainly includes those strings.

### Solution
Add separate binary copying according to `CUSTOM_ENV_GITLAB_CI` variable, indicating if it is run in CI or not.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -189,37 +189,68 @@ async def perform_pretest_cleanups():
         await cleanup()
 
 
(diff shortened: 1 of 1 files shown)


ThrasherLT:
+1.0

Jauler (approved):
+1.0

PR #745, opened 2024-08-05
nat-lab: save moose DB files 

matislovas:
### Problems
 - We do not get verbose information, when `lana` validators fail.
 - `interderp` pre-conditions timeout is too short.

### Solution
 - Print failing event, when validating and save `*-events.db` files to `logs` dir after each test.
 - Increase pre-conditions timeout to 30 secs.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -1,5 +1,6 @@
 # pylint: disable=too-many-lines
 import asyncio
+import glob
 import json
 import os
 import platform
@@ -666,6 +667,9 @@ class Client:
                     await self._router.delete_exit_node_route()
                     await self._router.delete_interface()
 
+                print(datetime.now(), "Test cleanup stage 7. Saving moose dbs")
+                await self.save_moose_db()
+
                 print(datetime.now(), "Test cleanup complete")
 
     async def simple_start(self):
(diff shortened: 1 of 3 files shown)


nat-lab/tests/telio.py
```diff
@@ -1153,6 +1161,27 @@ async def save_logs(self) -> None:
...
+        """
+        if os.environ.get("NATLAB_SAVE_LOGS") is None:
+            return
+
+        log_dir = "logs"
+        os.makedirs(log_dir, exist_ok=True)
+
+        test_name = test_name_safe_for_file_name()
+
+        moose_db_files = glob.glob("./*-events.db", recursive=False)
```
Jauler, line 1178: Arent moose DB files supposed to be inside of container? :thinking:

matislovas: Yes, and they are, with these filenames:
```
CONTAINER_EVENT_PATH = "/event.db"
CONTAINER_EVENT_BACKUP_PATH = "/event_backup.db"
```

But when downloaded to test runner, they are renamed to `{alpha,bet,gamma}-events.db`.

Jauler: Aren'nt they downloaded only if some assert is done one them? I wonder if it would not be better to copy them from all client containers at this point.

matislovas: They're downloaded with `get_moose_db_file` ... and we have the latest copy already downloaded to test runner, on which the `assert` _might_ fail, so there will be some meat for investigation.

Jauler: ok

Hasan6979 (approved):
lgtm +1.0

PR #751, opened 2024-08-07
[LLT-5492] Fix typing errors when arch is not specified

Jauler:
### Problem
A case was missed where `self.target_os` may be `None` [here](https://github.com/NordSecurity/libtelio/blob/main-2408061718-43ef9e7/ci/fetch_artifacts.py#L158) was missed causing typing errors during runtime. This PR adds filter for target_arch to be not none when attempting to download binary artifacts.

In addition requirement was added that at least one pipeline must match the filters, such that we would raise an error if download was skipped due to non-matching pipelines.

Failed run can be observed in pipeline `2379654`

- [x] README.md is updated

### DIFF
--- a/ci/fetch_artifacts.py
+++ b/ci/fetch_artifacts.py
@@ -137,6 +137,8 @@ class ArtifactsDownloader:
         tag_data = json.loads(tag_msg)
         pipeline_id = tag_data["pipeline_id"]
 
+        job_found = False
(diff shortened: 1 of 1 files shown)


tomaszklak (approved):
+1

dfetti (commented):
+1.0

PR #752, opened 2024-08-07
LLT-5106 macOS multicast integration tests

tomasz-grz:
### Problem
We want to make sure multicast implementation is working on macOS

### Solution
Add multicast nat-lab integration tests for macOS.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/multicast.py
+++ b/nat-lab/bin/multicast.py
@@ -1,5 +1,6 @@
 import argparse
 import socket
+import sys
 
 SSDP_IP: str = "239.255.255.250"
 SSDP_PORT: int = 1900
@@ -23,7 +24,14 @@ def ssdp_client(timeout: int):
 def ssdp_server(timeout: int):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     s.settimeout(timeout)
-    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    if sys.platform == "darwin":
+        # On macOS SO_REUSEPORT allows completely duplicate bindings by multiple processes
+        # if they all set SO_REUSEPORT before
+        # This option permits multiple instances of a program to each receive UDP/IP multicast or
+        # broadcast datagrams destined for the bound port
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+    else:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     s.bind(("0.0.0.0", SSDP_PORT))
 
     mreq = socket.inet_aton(SSDP_IP) + socket.inet_aton("0.0.0.0")
@@ -40,7 +48,10 @@ def ssdp_server(timeout: int):
 def mdns_client(timeout: int):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     s.settimeout(timeout)
-    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    if sys.platform == "darwin":
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+    else:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     s.bind(("0.0.0.0", MDNS_PORT))
 
     mreq = socket.inet_aton(MDNS_IP) + socket.inet_aton("0.0.0.0")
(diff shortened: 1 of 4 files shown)


nat-lab/tests/utils/multicast.py
```diff
@@ -9,6 +9,8 @@
 def _get_python(connection: Connection) -> str:
     if connection.target_os == TargetOS.Windows:
         return "python"
+    if connection.target_os == TargetOS.Mac:
```
mathiaspeters, line 12: For natlab in general we just need to specify `python3`, why do we need to specify the full path for multicast?

nat-lab/tests/test_multicast_connection.py
```diff
@@ -52,6 +52,22 @@ def generate_setup_parameter_pair(
...
+        marks=pytest.mark.mac,
+    ),
+    pytest.param(
+        generate_setup_parameter_pair([
+            (ConnectionTag.DOCKER_CONE_CLIENT_1, AdapterType.BoringTun),
+            (ConnectionTag.MAC_VM, AdapterType.BoringTun),
+        ]),
+        "mdns",
+        marks=pytest.mark.mac,
+    ),
```
tomaszklak, line 70: Why is there no linux native adapter in this whole list?

nat-lab/tests/utils/vm/mac_vm_util.py
```diff
@@ -78,6 +79,7 @@ async def _copy_binaries(
             f"{VM_UNIFFI_DIR}libtelio_remote.py",
             True,
         ),
+        (f"{LOCAL_BIN_DIR}multicast.py", f"{VM_TCLI_DIR}multicast.py", True),
```
tomaszklak, line 82: This file is copied to the mac VM, but how is it being used? I'm not seeing any mention of running it in this review.

nat-lab/tests/test_multicast_connection.py
tomasz-grz: I only adapted whatever was there for macOS, if it's just a another parameter pair, I can add it here.

nat-lab/tests/test_multicast_connection.py
```diff
@@ -61,6 +77,13 @@ async def add_multicast_route(connection: Connection) -> None:
             ["ip", "route", "add", "224.0.0.0/4", "dev", "tun10"]
         )
         await ipconf.execute()
+    elif connection.target_os == TargetOS.Mac:
+        ipconf = await connection.create_process(["route", "delete", "224/4"]).execute()
+        print(f"{ipconf.get_stdout()}")
```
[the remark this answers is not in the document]
tomasz-grz: Definitely helps in debugging, but I can remove if we want less logs.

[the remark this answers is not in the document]
mathiaspeters: We have a lot of similar diagnostics in natlab so I'm not necessarily against having it, but I don't think it should be printed twice and it should probably have something more in the print statement like timestamp, the fact that it's on mac, and maybe a quick word on what is being printed

nat-lab/tests/test_multicast_connection.py
tomaszklak: I'm not asking you to add it in this task, I'm asking out of curiosity and because quick skim of the RFC-0047 didn't answer my question.

nat-lab/tests/utils/multicast.py
tomasz-grz: we are trying to run`bin/multicast.py` from within the test, and the resulting `SshConnection` doesn't seem to find our custom python installation, instead it tries to run the system python installation and fails with `xcode-select: note: no developer tools were found`.
This ensures we are running the correct version of python, [just like when running `list_interfaces_with_router_property.py`](https://github.com/NordSecurity/libtelio/blob/ef49d4aa42e957ed5c40dfb1b99bb3130d8bbb4f/nat-lab/tests/test_vpn.py#L187)

mathiaspeters (approved):
+2

nat-lab/bin/multicast.py
```diff
@@ -56,7 +67,10 @@ def mdns_client(timeout: int):
 def mdns_server(timeout: int):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     s.settimeout(timeout)
-    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    if sys.platform == "win32":
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    else:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
```
mathiaspeters, line 73: I can see the tests are passing so must not be an issue but will this not change the behavior on linux since it's now using `SO_REUSEPORT` instead of  `SO_REUSEADDR`? Or could it be that it doesn't matter for linux and it's just important that mac and windows get the right socket options? If so, I think the different cases should be standardized to make all them look like this one `if windows: else:`

tomasz-grz: Uhh sorry, good catch, meant to do this the same as above

mathiaspeters (approved):

PR #757, opened 2024-08-08
LLT-5466 Fix malformed utf-8 byte sequence in logs

jjanowsk:
Because of the lack of lock in logger implementation two logs could
overlap leading to missing logs or in our case to malformed utf-8
sequences.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/uniffi/libtelio_remote.py
+++ b/nat-lab/tests/uniffi/libtelio_remote.py
@@ -5,6 +5,7 @@ import Pyro5.api  # type: ignore
 import Pyro5.server  # type: ignore
 import sys
 import telio_bindings as libtelio  # type: ignore # pylint: disable=import-error
+from threading import Lock
 from typing import List
 
 REMOTE_LOG = "remote.log"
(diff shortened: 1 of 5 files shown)


Jauler (approved):
nice catch! +1.0

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -75,19 +76,21 @@ def next_event(self):
...
-                    f"{datetime.datetime.now()} Failed to write logline due to error {str(e)}\n"
-                )
+        with self.lock:
+            with open(TCLI_LOG, "a", encoding="utf-8") as logfile:
+                try:
+                    logfile.write(f"{datetime.datetime.now()} {log_level} {payload}\n")
+                except IOError as e:
+                    logfile.write(
+                        f"{datetime.datetime.now()} Failed to write logline due to error {str(e)}\n"
+                    )
```
tomaszklak, line 93: nit: Maybe we should switch to [logging](https://docs.python.org/3/library/logging.html#thread-safety) which claims to be thread safe by design?

tomaszklak (approved):
+1

tomaszklak:
One additional question - i would expect that we only produce ascii logs which should be fine regardless of how many threads write them to the same file.

Is that not in fact the case, and we produce non ascii utf8 code points? 😮

jjanowsk:
> One additional question - i would expect that we only produce ascii logs which should be fine regardless of how many threads write them to the same file.
> 
> Is that not in fact the case, and we produce non ascii (utf-8? utf-16?) code points? 😮

@tomaszklak We do. WireGuard Go logs this character: "…" which is utf-8 encoded as: 0xE2 0x80 0xa6. It's possible that this is the only place and that's why only WireGuard Go tests have failed.

nat-lab/tests/uniffi/libtelio_remote.py
jjanowsk: I don't see big wins here from using it.

PR #758, opened 2024-08-08
Fix a race in test_init_lana_with_failing_moose_init_cb

tomaszklak:
### Problem
The test waited until the callback was called, not until the callback finished.

### Solution
Wait for the results of calling the callback to happen. Additionally a bug in `test_init_lana_with_failing_moose_init` is fixed - init returning true but calling callback with success would be a bug in moose.

- [x] README.md is updated


Jauler (approved):
+1.0

dfetti (approved):
+1.0

PR #760, opened 2024-08-08
LLT-5498 Add ip stack check for modules with Pinger

matislovas:
### Problem
Libtelio modules, which uses `pinger`, did not checked the IP stack, which is currently used on adapter (`IPv4`, `IPv6`, or dual `IPv4v6`) and so `pinger` can actually open the `IPv4` socket, even if the current stack is `IPv6`, but the source address would be set (by OS) to out-of-tunnel address.

### Solution
Check, what stack is used, before performing ping requests in-tunnel.

- [x] README.md is updated

### DIFF
--- a/crates/telio-nurse/src/qos.rs
+++ b/crates/telio-nurse/src/qos.rs
@@ -15,10 +15,10 @@ use telio_task::{io::mc_chan, Runtime, RuntimeExt, WaitResponse};
 use telio_wg::uapi::{AnalyticsEvent, PeerState};
 
 use telio_utils::{
-    interval, telio_log_debug, telio_log_trace, DualPingResults, DualTarget, Pinger,
+    interval, telio_log_debug, telio_log_trace, DualPingResults, DualTarget, IpStack, Pinger,
 };
 
-use crate::config::QoSConfig;
+use crate::{config::QoSConfig, data::MeshConfigUpdateEvent};
 
 /// Information about a node in the meshnet.
 #[derive(Clone)]
@@ -115,10 +115,10 @@ impl OutputData {
     #[allow(dead_code)]
     pub fn new(buckets: u32) -> Self {
         OutputData {
-            rtt: Analytics::empty_data(buckets),
-            rtt_loss: Analytics::empty_data(buckets),
-            rtt6: Analytics::empty_data(buckets),
-            rtt6_loss: Analytics::empty_data(buckets),
+            rtt: Analytics::null_data(buckets),
+            rtt_loss: Analytics::null_data(buckets),
+            rtt6: Analytics::null_data(buckets),
+            rtt6_loss: Analytics::null_data(buckets),
             tx: Analytics::empty_data(buckets),
             rx: Analytics::empty_data(buckets),
             connection_duration: String::from("0"),
@@ -142,6 +142,15 @@ impl OutputData {
         }
     }
 
+    pub fn push_rtt_data(histogram: Option<&Histogram>, buckets: u32, output: &mut String) {
+        output.push_str(&histogram.filter(|h| h.entries() > 0).map_or_else(
+            || Analytics::null_data(buckets),
+            |h| Analytics::percentile_histogram(h, buckets),
+        ));
+
+        output.push(',');
+    }
+
     fn merge_strings(a: String, b: String, separator: char) -> String {
         let mut result = String::new();
         if a.is_empty() {
@@ -161,6 +170,7 @@ impl OutputData {
 pub struct Io {
     pub wg_channel: mc_chan::Rx<Box<AnalyticsEvent>>,
     pub manual_trigger_channel: mc_chan::Rx<()>,
+    pub config_update_channel: mc_chan::Rx<Box<MeshConfigUpdateEvent>>,
 }
 
 /// Analytics data about a meshnet.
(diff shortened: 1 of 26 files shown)


nat-lab/tests/helpers.py
```diff
@@ -366,3 +366,28 @@ async def setup_mesh_nodes(
         await connection_future
 
     return env
+
+
+def connectivity_stack(node1_stack: IPStack, node2_stack: IPStack) -> Optional[IPStack]:
+    """
+    Checks, through which paths (v4, v6 or both) the nodes can communicate with each other.
+    """
+    match (node1_stack, node2_stack):
```
mathiaspeters, line 375: Python `match` statements are only supported on python 3.10 and later and I think we somewhat recently said on a different PR that we want to keep supporting python 3.9 for now, so this would need to be turned into an if-statement

matislovas: I probably missed that 3.9 support thingy, ok will downgrade to classics.

nat-lab/tests/utils/analytics/event_validator.py
```diff
@@ -341,66 +347,94 @@ def add_validator_list(self, validators: List[Validator]) -> Self:
...
+                    rtt_loss_c,
+                    rtt_loss_dnc,
+                    rtt6_c,
+                    rtt6_dnc,
+                    rtt6_loss_c,
+                    rtt6_loss_dnc,
+                ) = (None, None, None, None, None, None, None, None)
+                rtt_eq, rtt_loss_eq, rtt6_eq, rtt6_loss_eq = "", "", "", ""
+
+                match connectivity_stack(
```
mathiaspeters, line 375: The match here too will need to be an if-statement

nat-lab/tests/test_lana.py
```diff
@@ -1097,47 +1084,14 @@ async def test_lana_with_vpn_connection(
                     does_not_contain=["0:0:0:0:0:0"],
                     count=1,
                 ),
-                RttValidator(
```
mathiaspeters, line 1100: Was this removed because it's covered elsewhere?

nat-lab/tests/test_lana.py
```diff
@@ -1726,37 +1672,86 @@ def get_features_with_long_qos() -> TelioFeatures:
...
+            rtt_dnc,
+            rtt_loss_c,
+            rtt_loss_dnc,
+            rtt6_c,
+            rtt6_dnc,
+            rtt6_loss_c,
+            rtt6_loss_dnc,
+        ) = (None, None, None, None, None, None, None, None)
+        rtt_eq, rtt_loss_eq, rtt6_eq, rtt6_loss_eq = "", "", "", ""
+        match connectivity_stack(alpha_ip_stack, beta_ip_stack):
```
mathiaspeters, line 1688: `match`

mathiaspeters (commented):
+1

nat-lab/tests/helpers.py
mathiaspeters: It wasn't really an official thing, but I tried to introduce the same thing before, missing that it was only python 3.10+ and got a couple of comments on my PR that we're not ready to let go of python <3.10 yet

nat-lab/tests/test_lana.py
matislovas: It was added by error, all needed validators are added in `add_rtt_validators`

nat-lab/tests/utils/analytics/event_validator.py
matislovas: Fixed

nat-lab/tests/helpers.py
matislovas: Fixed

nat-lab/tests/test_lana.py
matislovas: Fixed

crates/telio-nurse/src/qos.rs
```diff
@@ -274,28 +285,44 @@ impl Analytics {
         for key in sorted_public_keys {
             if let Some(node) = nodes.get(key) {
                 // RTT
-                output.rtt.push_str(&Analytics::percentile_histogram(
-                    &node.rtt_histogram,
-                    buckets,
-                ));
+                if node.rtt_histogram.entries() > 0 {
```
Hasan6979, line 288: We can extract out ln 288:296 into a function and repeat it for other rtts.

crates/telio-nurse/src/qos.rs
```diff
@@ -316,16 +343,16 @@ impl Analytics {
                     .push_str(&node.connected_time.as_secs().to_string());
                 output.connection_duration.push(';');
             } else {
-                output.rtt.push_str(&Analytics::empty_data(buckets));
```
Hasan6979, line 319: Same for this. make a generic method call for all this.

matislovas: I do understand, why you want a one function for 288:296 lines, but this is already quite lean in code ...

crates/telio-nurse/src/qos.rs
matislovas: Fixed

Hasan6979 (approved):
+1.0

PR #763, opened 2024-08-12
Update conntrack trigger rule to be actually noop

Jauler:
### Problem
In Non-Initial network namespaces conntrack is deactivated until some reference is created. We have been using:
```
iptables -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
```
to add a reference, but this rule is a bit too intrusive and may cause spurious packets being accepted inside testing containers. Therefore I am proposing to change it to:
```
iptables -A INPUT -m conntrack --ctstate ESTABLISHED -j LOG --log-level debug
```
in this PR.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/client
+++ b/nat-lab/bin/client
@@ -11,8 +11,8 @@ sleep 5
 # Add conntrack states to iptables, since running dockers on ubuntu
 # conntrack doesn't find flow entries by default somehow. Works when running dockers on debian tho.
 # ref: https://serverfault.com/a/978715
-iptables -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
-ip6tables -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
+iptables -A INPUT -m conntrack --ctstate ESTABLISHED -j LOG --log-level debug
+ip6tables -A INPUT -m conntrack --ctstate ESTABLISHED -j LOG --log-level debug
 
 /libtelio/nat-lab/bin/configure_route.sh primary
 


nat-lab/bin/client
```diff
@@ -11,8 +11,8 @@ sleep 5
 # Add conntrack states to iptables, since running dockers on ubuntu
 # conntrack doesn't find flow entries by default somehow. Works when running dockers on debian tho.
 # ref: https://serverfault.com/a/978715
-iptables -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
-ip6tables -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
```
tomaszklak, line 15: Would `RETURN` be enough here?

tomaszklak (commented):
+1

nat-lab/bin/client
jjanowsk: If we use RETURN and later some test wants to add a rule to the INPUT table this rule will never be matched, so I think LOG is better suited in that case

jjanowsk (approved):
+1

nat-lab/bin/client
Jauler: `RETURN` target on "top-level" chain would stop the traversal of rules and apply the "policy" (e.g. `iptables -F ....`) target to the packet. Which would mean, that any rules further down the chain will not be evaluated. With `LOG` target, this is non-terminating target, so it is a bit more "noop".

PR #764, opened 2024-08-12
[LLT-5504] Force nc client to quit and close connection once server-side connection is closed

Jauler:
### Problem
Under "normal" circumstances closing connection from server side does not trigger `nc` to close TCP connection from client-side, which can be observed in the following screenshot:
2024-08-12_14-02-22_grim

In `test_mesh_firewall` tests there is expectation that `nc` will close connections from both sides once at least once side sends `FIN`. To match these expectations `-d` flag has been added, which tells `nc` to not read from stdin, and as a result close client-side connection immediatelly if server-side is closed.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_mesh_firewall.py
+++ b/nat-lab/tests/test_mesh_firewall.py
@@ -5,6 +5,7 @@ import config
 import pytest
 import telio
 from contextlib import AsyncExitStack
+from datetime import datetime
 from helpers import SetupParameters, setup_mesh_nodes, setup_api
 from mesh_api import Node
 from typing import Tuple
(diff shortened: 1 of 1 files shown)


mathiaspeters (dismissed):
+1

lcruz99:
+1

gytsto (approved):
+1

nat-lab/tests/test_mesh_firewall.py
```diff
@@ -601,8 +604,8 @@ async def conntrack_on_stdout(stdout: str) -> None:
             connected_event,
         )
 
-        output_notifier.notify_output("FIN_WAIT", last_ack_event)
-        output_notifier.notify_output("CLOSE_WAIT", time_wait_event)
+        output_notifier.notify_output("LAST_ACK", last_ack_event)
+        output_notifier.notify_output("TIME_WAIT", time_wait_event)
```
mathiaspeters, line 608: I'm curious about the reason for this change. I'm not against it, I would just like to hear the `why` behind it

PR #766, opened 2024-08-12
LLT-5447: Use OsRng for secret key generation

packgron:
### Problem
Then we are generating unique secret keys, there is no benefit to have additional hashing step, we gain from StdRng.

### Solution
Switched to plain OsRng

- [x] README.md is updated


LukasPukenis (commented):
According to google `OsRng` is cryptographically secure, this it makes sense. +1

mathiaspeters (approved):
+1

tomaszklak:
Given that `OsRng` [can block in early startups](https://docs.rs/getrandom/latest/getrandom/#early-boot) (ehm, nat-lab VMs?) maybe it would be better to reduce the direct dependency on teh OS source of randomness and use `thread_rng`?

packgron:
> Given that `OsRng` [can block in early startup](https://docs.rs/getrandom/latest/getrandom/#early-boot) (nat-lab VMs?) maybe it would be better to reduce the direct dependency on the OS source of randomness and use `thread_rng`?

This originally came from CheckMarx, and I though it does not realy make sence as StdRng is considered crypto safe. 
But after doing a bit of analysis, what we had previously was a hash(OsRng), so from blocking side we do not gain or loose something. We just remove additional hashing.

SecretKey should not be generated constantly (but kindof is in our test's), so I don't see this as a blocking issues in real scenarios, and OS random is typically recommended over seeded pseudo-random generators for sensitive areas.

If we encounter issues with this i would be more for adding this as a feature for use in testing, but default should be os random

tomaszklak:
> after doing a bit of analysis, what we had previously was a hash(OsRng), so from blocking side we do not gain or loose something. We just remove additional hashing.

I'm not proposing to use `StdRng` but the result of `fn thread_rng() -> ThreadRng`. `ThreadRng` implements `CryptoRng` and in addition does not constantly depend on OS (it's only periodically reseeded). So this means that with `ThreadRng` at startup we only need 32 bytes from OS to provide 64KiB of secure random bytes. 

> OS random is typically recommended over seeded pseudo-random generators for sensitive areas.

In that case, do you think that `impl RngCore + CryptoRng` in `pub fn gen_with(rng: &mut (impl RngCore + CryptoRng)) -> Self` is incorrect / not strong enough?

packgron:
> In that case, do you think that `impl RngCore + CryptoRng` in `pub fn gen_with(rng: &mut (impl RngCore + CryptoRng)) -> Self` is incorrect / not strong enough?

I would personaly prefer to trust it. CheckMarks will likely disagree (but it's more likely to be a false possitive). But will need to actually discuss this one out with AppSec.

> I'm not proposing to use StdRng but the result of fn thread_rng() -> ThreadRng. ThreadRng implements CryptoRng and in addition does not constantly depend on OS (it's only periodically reseeded). So this means that with ThreadRng at startup we only need 32 bytes from OS to provide 64KiB of secure random bytes.

Have we ever encounter issue because of this, I would agree for thread_rng being used for nonces, but it feels like it's an overkill for secret key

tomaszklak:
> Have we ever encounter issue because of this, I would agree for thread_rng being used for nonces, but it feels like it's an overkill for secret key

Hard to tell, but given that we find increasingly fun/rare problems in nat-lab now, I'm inclined to expect this to show up sooner or later :D

In all seriousness, I think that this shouldn't be a blocking issue so please consider this as nitpick :)

tomaszklak (dismissed):
+1

tomaszklak (approved):

PR #770, opened 2024-08-13
Make it possible to control using the `hide_ips` if the IP addresses are hidden

tomaszklak:
IPs should always be hidden in production, but hiding them makes it hard to investigate bugs found in the nat-lab tests. This is why a flag to control IP hiding has been added. While it defaults to hiding the IPs, for the nat-lab tests (and only for that case) it defaults to show IPs by changing pytests default value.

Additionally, as a side effect, the log censoring functionality no longer makes copies of log message strings when they haven't been modified.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -31,6 +31,9 @@ pub struct Features {
     /// Test environment (natlab) requires binding feature disabled
     /// TODO: Remove it once mac integration tests support the binding mechanism
     pub is_test_env: Option<bool>,
+    /// Controll if IP addresses should be hidden in logs
+    #[default(true)]
+    pub hide_ips: bool,
     /// Derp server specific configuration
     pub derp: Option<FeatureDerp>,
     /// Flag to specify if keys should be validated
@@ -456,6 +459,7 @@ mod tests {
                 }
             },
             "is_test_env": true,
+            "hide_ips": false,
             "derp": {
                 "tcp_keepalive": 13,
                 "derp_keepalive": 14,
@@ -537,6 +541,7 @@ mod tests {
                     }),
                 }),
                 is_test_env: Some(true),
+                hide_ips: false,
                 derp: Some(FeatureDerp {
                     tcp_keepalive: Some(13),
                     derp_keepalive: Some(14),
@@ -735,6 +740,21 @@ mod tests {
             let actual = from_str::<Features>(json).unwrap().wireguard;
             let expected = FeatureWireguard::default();
             assert_eq!(actual, expected);
+        }
+
+        #[test]
+        fn test_hide_ips() {
+            let json = r#"{}"#;
+            let actual = from_str::<Features>(json).unwrap();
+            assert_eq!(actual.hide_ips, true);
+
+            let json = r#"{"hide_ips": false}"#;
+            let actual = from_str::<Features>(json).unwrap();
+            assert_eq!(actual.hide_ips, false);
+
+            let json = r#"{"hide_ips": true}"#;
+            let actual = from_str::<Features>(json).unwrap();
+            assert_eq!(actual.hide_ips, true);
         }
     }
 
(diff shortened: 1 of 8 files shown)


crates/telio-model/src/features.rs
```diff
@@ -31,6 +31,8 @@ pub struct Features {
     /// Test environment (natlab) requires binding feature disabled
     /// TODO: Remove it once mac integration tests support the binding mechanism
     pub is_test_env: Option<bool>,
+    /// Controll if IP addresses should be hidden in logs
+    pub hide_ips: Option<bool>,
```
mathiaspeters, line 35: Does this really need to be optional? Adding a `#[serde(default)]` will make sure it's false if it's not specified in json and if doing `Features::default()`

crates/telio-model/src/features.rs
```diff
@@ -736,6 +740,25 @@ mod tests {
             let expected = FeatureWireguard::default();
             assert_eq!(actual, expected);
         }
+
+        #[test]
+        fn test_hide_ips() {
```
mathiaspeters, line 745: I recently started questioning the relevance of tests that are this granular since it's mostly testing that `serde` works as expected. The way the tests in this module is structured now is one test to set all values to non-default values in json and then tests to make sure the correct defaults are set when objects are empty. So this would just need to be added to the big json block and that should be it. At least that's my opinion

crates/telio-model/src/features.rs
mathiaspeters: After writing this I realized I got the logic backwards for this in my head, and it should of course default to true. Personally I would be in favor of still not having it be optional, and either inverting the name or using the smart defaults crate that Michal introduced

crates/telio-model/src/features.rs
tomaszklak: I find it to be way more easy to understand this test which I can actually fit in my head. The huge blob of json (and the accompanying huge struct expression) are enough for me to lose context.

crates/telio-model/src/features.rs
tomaszklak: Good point, it doesn't have to be optional. I kept the name and used the `SmartDefault` to make it default to `true`.

src/ffi/logging.rs
```diff
@@ -328,9 +341,7 @@ mod test {
         }
     }
 
-    #[test]
-    fn test_log_censor() {
-        let examples = [
+    const EXAMPLES: [(&'static str, &'static str); 11] = [
             (
                 "1999-09-09 [INFO] New endpoint (1.2.3.4:1234) created",
                 "1999-09-09 [INFO] New endpoint (IP(959535cab4852bd4):1234) created",
```
tomasz-grz, line 347: Silly idea, but what if instead of using some hash values, we had something in a more human friendly form, like the ones [generated by names](https://github.com/fnichol/names)? Would still obfuscate the address, but would be much easier to visually distinguish the devices when reading the logs.

tomasz-grz (approved):
+1.0 looks good to me

crates/telio-model/src/features.rs
mathiaspeters: I can understand the big json blob being a bit too big too fit in ones head. I'm not against testing the way you're doing but in that case I would like to see it done for all fields (obviously not in this PR though), maybe with some parameterization and json construction so we don't have to manually write these tests for all feature options. I started adding tests like that manually before settling on the current structure but it was becoming too much boilerplate after only doing a fraction of the existing fields

mathiaspeters (approved):

src/ffi/logging.rs
tomasz-grz: Hmm maybe we could have the what3words style with suffix code: `adj-adj-noun-hex`, then we would need 1024 word lists, and a 34bit suffix.

tomaszklak: Sounds like a possible solution, I will create a ticket for this improvement.

crates/telio-model/src/features.rs
tomaszklak: I will create a ticket to look into this in future.

PR #772, opened 2024-08-14
Use batcher for stun, vpn and proxied peers keepalives

dfetti:
Disable Wireguard keepalives for stun, vpn and proxied peers and use the session keeper intead because we want them to be batched for battery lifetime improvements.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5402
+++ b/.unreleased/LLT-5402
@@ -1,0 +1,1 @@
+Use batcher for stun, vpn and proxied peers persistent keepalives
--- a/crates/telio-traversal/src/session_keeper.rs
+++ b/crates/telio-traversal/src/session_keeper.rs
@@ -54,6 +54,7 @@ pub trait SessionKeeperTrait {
         threshold: Option<Duration>,
     ) -> Result<()>;
     async fn remove_node(&self, key: &PublicKey) -> Result<()>;
+    async fn get_interval(&self, key: &PublicKey) -> Option<u32>;
 }
 
 pub struct SessionKeeper {
@@ -239,6 +240,19 @@ impl SessionKeeperTrait for SessionKeeper {
         .await?;
 
         Ok(())
+    }
+
+    async fn get_interval(&self, key: &PublicKey) -> Option<u32> {
+        let pk = *key;
+        task_exec!(&self.task, async move |s| {
+            if let Some(interval) = s.batched_actions.get_interval(&pk) {
+                Ok(Some(interval))
+            } else {
+                Ok(s.nonbatched_actions.get_interval(&pk))
+            }
+        })
+        .await
+        .unwrap_or(None)
     }
 }
 struct Pingers {
--- a/crates/telio-utils/src/dual_target.rs
+++ b/crates/telio-utils/src/dual_target.rs
@@ -16,6 +16,22 @@ pub type Result<T> = std::result::Result<T, DualTargetError>;
 /// Target input with IPv4 and IPv6 addresses
 pub type Target = (Option<Ipv4Addr>, Option<Ipv6Addr>);
 
+/// Builds Target from a list of ip addresses
+pub fn build_ping_endpoint(ip_addresses: &Vec<IpAddr>, use_ipv6: bool) -> Target {
+    let mut ipv4 = None;
+    let mut ipv6 = None;
+
+    for ip in ip_addresses {
+        match (ip, ipv4, ipv6, use_ipv6) {
+            (IpAddr::V4(v4), None, _, _) => ipv4 = Some(*v4),
+            (IpAddr::V6(v6), _, None, true) => ipv6 = Some(*v6),
+            _ => (),
+        }
+    }
+
+    (ipv4, ipv6)
+}
+
 /// DualTarget wrapper
 #[derive(Clone, Copy, Debug, Default, PartialEq)]
 pub struct DualTarget {
--- a/src/device.rs
+++ b/src/device.rs
@@ -2319,8 +2319,14 @@ impl TaskRuntime for Runtime {
                                 // We have downgraded the connection. Notify cross ping check about that.
                                 direct_entities.cross_ping_check.notify_failed_wg_connection(public_key)
                                     .await?;
-                                direct_entities.session_keeper.remove_node(&public_key).await?;
                                 direct_entities.upgrade_sync.clear_accepted_session(public_key).await;
+
+                                if self.features.batching.is_none() {
+                                    // When the batcher is enabled we use session keeper for all the connections
+                                    // direct, proxy, vpn and stun. This call is guarded by the batcher feature flag
+                                    // because it can disable keepalives when we don't want to.
+                                    direct_entities.session_keeper.remove_node(&public_key).await?;
+                                }
                             } else {
                                 telio_log_warn!("Connection downgraded while direct entities are disabled");
                             }
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -11,7 +11,6 @@ use telio_dns::DnsResolver;
 use telio_firewall::firewall::{Firewall, FILE_SEND_PORT};
 use telio_model::constants::{VPN_EXTERNAL_IPV4, VPN_INTERNAL_IPV4, VPN_INTERNAL_IPV6};
 use telio_model::features::Features;
-use telio_model::mesh::NodeState::Connected;
 use telio_model::mesh::{LinkState, NodeState};
 use telio_model::EndpointMap;
 use telio_model::SocketAddr;
@@ -23,7 +22,7 @@ use telio_traversal::{
     cross_ping_check::CrossPingCheckTrait, endpoint_providers::EndpointProvider,
     SessionKeeperTrait, UpgradeSyncTrait, WireGuardEndpointCandidateChangeEvent,
 };
-use telio_utils::{telio_log_debug, telio_log_info, telio_log_warn};
+use telio_utils::{build_ping_endpoint, telio_log_debug, telio_log_info, telio_log_warn};
 use telio_wg::uapi::AnalyticsEvent;
 use telio_wg::{uapi::Peer, WireGuard};
 use thiserror::Error as TError;
@@ -59,6 +58,7 @@ struct EndpointState {
 #[derive(Debug, PartialEq, Eq, Clone)]
 struct RequestedPeer {
     peer: Peer,
+    batching_keepalive_interval: Option<u32>,
     endpoint: Option<EndpointState>,
 }
 
@@ -219,15 +219,44 @@ async fn consolidate_wg_peers<
         aggregator.report_peer_state_relayed(&event).await;
 
         wireguard_interface.del_peer(*key).await?;
+
+        // Remove from session_keeper
+        if let Some(sk) = session_keeper {
+            sk.remove_node(key).await?;
+        }
     }
 
-    for key in insert_keys {
-        let requested_peer = requested_peers.get(key).ok_or(Error::PeerNotFound)?;
-        telio_log_info!("Inserting peer: {:?}", requested_peer);
+    let batcher_threshold = features
+        .batching
+        .as_ref()
+        .map(|b| Duration::from_secs(b.direct_connection_threshold as u64));
 
-        wireguard_interface
-            .add_peer(requested_peer.peer.clone())
-            .await?;
+    for key in insert_keys {
+        telio_log_info!("Inserting peer: {:?}", requested_peers.get(key));
+        let peer = requested_peers.get(key).ok_or(Error::PeerNotFound)?;
+        let ip_addresses = peer.peer.ip_addresses.clone();
+
+        wireguard_interface.add_peer(peer.peer.clone()).await?;
+
+        // Add peer to session keeper if needed
+        match (session_keeper, peer.batching_keepalive_interval) {
+            (Some(sk), Some(keepalive_interval)) => {
+                let target = build_ping_endpoint(&ip_addresses, features.ipv6);
+                if target.0.is_some() || target.1.is_some() {
+                    sk.add_node(
+                        *key,
+                        target,
+                        Duration::from_secs(keepalive_interval as u64),
+                        batcher_threshold,
+                    )
+                    .await?;
+                } else {
+                    telio_log_warn!("Peer {:?} has no ip address", key);
+                }
+            }
+            (None, _) => telio_log_warn!("The session keeper is missing!"),
+            _ => (),
+        }
 
         if let Some(stun) = stun_ep_provider {
             if let Some(wg_stun_server) = requested_state.wg_stun_server.as_ref() {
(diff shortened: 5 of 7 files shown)


src/device/wg_controller.rs
```diff
@@ -200,12 +199,60 @@ async fn consolidate_wg_peers<
     for key in delete_keys {
         telio_log_info!("Removing peer: {:?}", actual_peers.get(key));
         wireguard_interface.del_peer(*key).await?;
+
+        // Remove from session_keeper
+        if let Some(sk) = session_keeper {
+            sk.remove_node(key).await?;
```
LukasPukenis, line 205: I am feeling cautious about this one - Libtelio currently assumes that SessionKeeper is only used by direct peers and `device.rs` has: 
```                             
if !was_proxying && is_proxying {
...
2293: direct_entities.session_keeper.remove_node(&public_key).await?;
```

So there's a risk that actually both are happening one after another(I am not sure). One would then short-circuit another and cause a failure.

Of course that might not be the case and I am just paranoid :D

src/device/wg_controller.rs
```diff
@@ -200,12 +199,60 @@ async fn consolidate_wg_peers<
...
+        }
     }
 
+    let batcher_threshold = features
+        .batching
+        .as_ref()
+        .map(|b| Duration::from_secs(b.direct_connection_threshold as u64));
```
LukasPukenis, line 212: Threshold should be configurable for each type, as each persistent-keepalive can be different, threshold cannot be a constant shared through all of them

src/device/wg_controller.rs
```diff
@@ -200,12 +199,60 @@ async fn consolidate_wg_peers<
...
+                .ok_or(Error::PeerNotFound)?
+                .peer
+                .clone();
+
+            // Persistent keepalive interval has been set by the build_requested_peers_list for
+            // every node. It considers type of connection proxy/direct and special nodes as VPN or STUN.
+            // We want to disable WG keepalives and use our session keeper instead
+            let keepalive_interval = peer
+                .persistent_keepalive_interval
+                .map_or(Duration::from_secs(25), |k| Duration::from_secs(k as u64));
```
LukasPukenis, line 229: Hardcoding `25` is a bad idea, `peer.peer.persistent_keepalive_interval` already has it :)

src/device/wg_controller.rs
```diff
@@ -200,12 +199,60 @@ async fn consolidate_wg_peers<
...
+                let target = build_target(&ip_addresses, features.ipv6);
+                if target.0.is_some() || target.1.is_some() {
+                    sk.add_node(*key, target, keepalive_interval, batcher_threshold)
+                        .await?;
+                } else {
+                    telio_log_warn!("Peer {:?} has no ip address", key);
+                }
+            }
+        } else {
+            wireguard_interface
```
LukasPukenis, line 246: Nit: It would be nice to have `wireguard_interface.add_peer()` without a conditional. Without `add_peer()` there's no connection and it only changes the `persistent-keepalive` based on batching. This can be refactored to be a bit nicer to read imho:
```
if features.batching.is_some() {
  session_keeper.add(peer);
  peer.peer.persistent_keepalive = Some(0);
}

wireguard_interface.add_peer(peer).await?;
```

This way it's less conditionals and no risk of `add_peer` not being executed due to being hidden inside of a condition

LukasPukenis (commented):
The PR lacks the configuration of the STUN, VPN, PROXY thresholds inside of a batcher configuration.

EDIT: after discussing we agreed that it makes sense to reuse the same, one, threshold option. @dfetti just in that case you should rename the option as it has `direct` in it even though it's general threshold.

src/device/wg_controller.rs
dfetti: We decided to reuse this value for now

crates/telio-utils/src/dual_target.rs
```diff
@@ -14,6 +14,29 @@ pub type Result<T> = std::result::Result<T, DualTargetError>;
...
 
+/// Builds Target from a list of ip addresses
+pub fn build_target(ip_addresses: &Vec<IpAddr>, use_ipv6: bool) -> Target {
+    let mut ipv4 = None;
+    let mut ipv6 = None;
+
+    for ip in ip_addresses {
+        match ip {
+            IpAddr::V4(v4) => {
+                if ipv4.is_none() {
```
packgron, line 25: NIT: you could avoid additional identation and merge if with mach arm. adding default as none case

src/device.rs
```diff
@@ -2294,7 +2294,12 @@ impl TaskRuntime for Runtime {
                                 direct_entities.cross_ping_check.notify_failed_wg_connection(public_key)
                                     .await?;
 
-                                direct_entities.session_keeper.remove_node(&public_key).await?;
+                                if self.features.batching.is_none() {
+                                    // When the batcher is enabled we use session keeper for all the connections
+                                    // Direct, proxy, vpn and stun. If batcher is enabled this call will disable keepalives
```
packgron, line 2299: If the backer is *disabled* this call will disable keepalives ?

tomasz-grz (commented):
+0.5 
Unfortunately I don't fully understand yet the implications it could have on wg controller to give it a full point, but otherwise it looks well done to me, just a few questions.

crates/telio-utils/src/dual_target.rs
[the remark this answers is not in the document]
dfetti: There are cases like VPN nodes that has 2 IPv4 addresses. One is enough for the session keeper / wherever the DualTarget struct is needed.

src/device/wg_controller.rs
```diff
@@ -230,9 +280,47 @@ async fn consolidate_wg_peers<
...
+
+            let peer = if features.batching.is_some() {
+                // If batching is enabled, we use the session keeper for the persistent keepalives
+                let mut peer = requested_peers
+                    .get(key)
+                    .ok_or(Error::PeerNotFound)?
+                    .peer
+                    .clone();
+
+                // Persistent keepalive interval has been set by the build_requested_peers_list for
```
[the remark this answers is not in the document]
dfetti: :bow:

tomasz-grz (approved):

src/device.rs
```diff
@@ -1575,8 +1575,12 @@ impl Runtime {
             nurse.set_private_key(*private_key).await;
         }
 
-        wg_controller::consolidate_wg_state(&self.requested_state, &self.entities, &self.features)
-            .await?;
+        Box::pin(wg_controller::consolidate_wg_state(
```
packgron, line 1578: NIT: you can use .boxed() from futures::FutureExt

src/device/wg_controller.rs
```diff
@@ -1872,6 +1931,49 @@ mod tests {
         f.consolidate_peers().await;
     }
 
+    #[tokio::test]
+    async fn when_fresh_start_then_proxy_enpoint_is_added_with_batching() {
```
packgron, line 1935:
```suggestion
    async fn when_fresh_start_and_batching_enabled_then_proxy_enpoint_is_added_with_batching() {
```

src/device/wg_controller.rs
```diff
@@ -266,63 +359,27 @@ async fn consolidate_wg_peers<
...
-                                    .into(),
-                            ),
-                            None,
-                        )
-                        .await?
-                    }
+            // If the batcher is disabled we still have to enable session keeper for the direct connections
+            if features.batching.is_none() {
+                if let Some(sk) = session_keeper {
+                    let target = build_target(&requested_peer.peer.ip_addresses, features.ipv6);
```
packgron, line 365: NIT: maybe `build_ping_endpoint` ?

packgron: ok build_session_keeper_endpoint

.unreleased/LLT-5402
```diff
@@ -0,0 +1 @@
+Use batcher for stun, vpn and proxied peers persistent keepalives
```
Jauler, line 1: As this PR is requiring additional QA cycle - lets wait for its merge to happen after v5.0.0 release

tomasz-grz (approved):

Jauler:
I know this is a bit maaaybe not supre related to this PR, but I wanted to double-check/ask. In the Device, its `requested_state` sturcture there is [feature config values added for batcher](https://github.com/NordSecurity/libtelio/blob/LLT-5402-batch-other-keepalives/src/device.rs#L231), is it intentional? Should'nt batcher feature config be in [features](https://github.com/NordSecurity/libtelio/blob/LLT-5402-batch-other-keepalives/src/device.rs#L201)?

src/device.rs
Jauler, line 1: I know this is a bit maaaybe not supre related to this PR, but I wanted to double-check/ask. In the Device, its requested_state sturcture there is [feature config values added for batcher](https://github.com/NordSecurity/libtelio/blob/LLT-5402-batch-other-keepalives/src/device.rs#L231), is it intentional? Should'nt batcher feature config be in [features](https://github.com/NordSecurity/libtelio/blob/LLT-5402-batch-other-keepalives/src/device.rs#L201)?

src/device/wg_controller.rs
```diff
@@ -219,15 +220,81 @@ async fn consolidate_wg_peers<
...
+            telio_log_debug!(
+                "Keepalive interval for {:?} is {:?}",
+                peer,
+                peer.persistent_keepalive_interval
+            );
+            let keepalive_interval = Duration::from_secs(
+                peer.persistent_keepalive_interval
+                    .map(|k| {
+                        if k == 0 {
+                            requested_state.keepalive_periods.direct
```
Jauler, line 256: :thinking: I am not sure about this part a bit :thinking: 

It seems like this forbids any kind of keepalive disable (even if carried-on by session-keeper using ICMP's), but for STUN we should be able to disable all keepalives allthogether.

src/device/wg_controller.rs
```diff
@@ -219,15 +220,81 @@ async fn consolidate_wg_peers<
...
+                        } else {
+                            k
+                        }
+                    })
+                    .unwrap_or(requested_state.keepalive_periods.direct)
+                    .into(),
+            );
+            peer.persistent_keepalive_interval = Some(0); // Disable WG persistent keepalive
```
Jauler, line 264: I am afraid that reusing persistent_keepalive_interval in `RequestedPeer` structure may be a bit confusing, if we are just simply going to override it in all cases to `0` later. Maybe lets add a separate field dedicated for holding just the session_keeper period?

src/device/wg_controller.rs
Jauler, line 1: I am afraid that reusing `persistent_keepalive_interval` in `RequestedPeer` structure may be a bit confusing, if we are just simply going to override it in all cases to `0` later. Maybe it is worth adding a separate field dedicated for holding just the session_keeper period into `RequestedPeer`?

src/libtelio.udl
```diff
@@ -556,8 +556,8 @@ dictionary Features {
 };
 
 dictionary FeatureBatching {
-    /// direct connection threshold for batching
-    u32 direct_connection_threshold;
+    /// Connection threshold for batching
+    u32 threshold;
```
Jauler, line 560: Adding reference to this: https://github.com/NordSecurity/libtelio/pull/772/files#r1751581554

crates/telio-model/src/features.rs
```diff
@@ -65,9 +65,9 @@ pub struct Features {
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, SmartDefault)]
 #[serde(default)]
 pub struct FeatureBatching {
-    /// Direct connection threshold when batching (in seconds) [default 0s]
+    /// Connection threshold when batching (in seconds) [default 0s]
     #[default(0)]
-    pub direct_connection_threshold: u32,
+    pub threshold: u32,
```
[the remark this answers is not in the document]
dfetti: Well, we decided for this PR to leave only one global threshold  for all the batched things, but in the end we will need different values for all the keepalives. So, for this PR is ok to remain only one value and I will rename it back to direct_connection_threshold to not break compatibility.

src/device.rs
```diff
@@ -2320,8 +2320,14 @@ impl TaskRuntime for Runtime {
...
                                 direct_entities.cross_ping_check.notify_failed_wg_connection(public_key)
                                     .await?;
-                                direct_entities.session_keeper.remove_node(&public_key).await?;
                                 direct_entities.upgrade_sync.clear_accepted_session(public_key).await;
+
+                                if self.features.batching.is_none() {
+                                    // When the batcher is enabled we use session keeper for all the connections
+                                    // direct, proxy, vpn and stun. This call is guarded by the batcher feature flag
+                                    // because it can disable keepalives when we don't want to.
+                                    direct_entities.session_keeper.remove_node(&public_key).await?;
```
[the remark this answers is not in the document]
dfetti: remove_node will not return any error for non-existent nodes

Jauler (approved):
+1.0

src/device/wg_controller.rs
```diff
@@ -551,7 +579,12 @@ async fn build_requested_peers_list<
                 ip_addresses.push(VPN_INTERNAL_IPV6.into());
             }
             ip_addresses.push(VPN_EXTERNAL_IPV4.into());
-            let persistent_keepalive_interval = requested_state.keepalive_periods.vpn;
+            let (persistent_keepalive_interval, batching_keepalive_interval) =
+                if features.batching.is_some() {
+                    (None, requested_state.keepalive_periods.vpn)
+                } else {
+                    (requested_state.keepalive_periods.vpn, None)
+                };
```
LukasPukenis, line 587: I am thinking about this improvement and cannot understand if it's safely implemented. Given that tests are passing it probably is, however the switching of the values is what gets me a bit paranoid, also it happens not once. One way to secure against it would be pass values straight and nearby UAPI check if we have batching then don't use the regular persistent-ka value. This of course isn't also perfect as then the logic for batching spills a bit into another area but might be a less complex solution than the current one. What are your thoughts on this one?

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -240,6 +241,19 @@ impl SessionKeeperTrait for SessionKeeper {
 
         Ok(())
     }
+
+    async fn get_interval(&self, key: &PublicKey) -> Option<u32> {
```
LukasPukenis, line 245: Speaking about `SessionKeeper` - do we need it at all at this point? Since `Batcher` is doing everything now, it's not being used IMO. References show just one reference and it's inside a unit-test :)

src/device/wg_controller.rs
```diff
@@ -252,11 +281,41 @@ async fn consolidate_wg_peers<
...
         }
 
+        // Check if we need to update sesion keeper
+        match (
+            session_keeper,
+            features.batching.is_some(),
+            requested_peer.batching_keepalive_interval,
+        ) {
+            (Some(sk), true, Some(interval)) => {
+                if sk.get_interval(key).await != Some(interval)
```
LukasPukenis, line 297: Several small questions:
- Is SessionKeeper only a storage at this point which is not really used anywhere?
- It's a bit confusing as SessionKeeper receives a normal interval but here we compare it to the batching interval which means we compare two separate things :thinking: 
- `get_interval()` is not needed at all if this comparison gets removed IMO

It might just be that I missed some details here and there and it all makes sense though :D

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: SK is used for both batched and nonbatched peers

src/device/wg_controller.rs
LukasPukenis: get_interval() is needed to compare against proxying peers that upgraded to direct ones by the means of checking their intervals(of course they might match and it simply means nothing needs to be done).

LukasPukenis:
+1

LukasPukenis (approved):

PR #775, opened 2024-08-14
natlab: Stop copying binaries locally when not needed

lcruz99:
### Problem
When running natlab locally, pytest setup functions were attempting to copy binaries to VMs even if there were no running VMs or tests to run on VMs.

### Solution
Stop trying to copying binaries if there's no need to.

- [x] README.md is updated


gytsto (commented):
Not all test have these steings in their names if they are not parameteized. That's the main problem

tomasz-grz (approved):
+1.0

lcruz99:
@gytsto that only happens with `test_get_network_interface_tunnel_keys[WireguardGo-Wintun Userspa..]`, we can either add it as parameter or also look for the adapters name (ie: WireguardGo/WindowsNativeWg)

gytsto:
> @gytsto that only happens with `test_get_network_interface_tunnel_keys[WireguardGo-Wintun Userspa..]`, we can either add the parameter to the test or also look for the adapters (ie: WireguardGo/WindowsNativeWg) name in the items name.

If that's the case, we can actually revert my commit and add parameter to test that you mentioned. But any new test that has no parameters and no name with those strings, will fail due to missing binary locally (my intention was to fix that)

lcruz99:
> > @gytsto that only happens with `test_get_network_interface_tunnel_keys[WireguardGo-Wintun Userspa..]`, we can either add the parameter to the test or also look for the adapters (ie: WireguardGo/WindowsNativeWg) name in the items name.
> 
> If that's the case, we can actually revert my commit and add parameter to test that you mentioned. But any new test that has no parameters and no name with those strings, will fail due to missing binary locally (my intention was to fix that)

I see, sooner or later that can happen indeed..  🤔

The motivation for this PR was that locally, when there's no WIN/MAC VMs, the setup function will try to connect to them with combined timeouts of ~5min. Maybe we could add a shorter timeout? Ideally we shouldn't add any time to test duration

wdyt?

gytsto:
> > > @gytsto that only happens with `test_get_network_interface_tunnel_keys[WireguardGo-Wintun Userspa..]`, we can either add the parameter to the test or also look for the adapters (ie: WireguardGo/WindowsNativeWg) name in the items name.
> > 
> > 
> > If that's the case, we can actually revert my commit and add parameter to test that you mentioned. But any new test that has no parameters and no name with those strings, will fail due to missing binary locally (my intention was to fix that)
> 
> I see, sooner or later that can happen indeed.. 🤔
> 
> The motivation for this PR was that locally, when there's no WIN/MAC VMs, the setup function will try to connect to them with combined timeouts of ~5min. Maybe we could add a shorter timeout? Ideally we shouldn't add any time to test duration. Another possible solution would be to keep this PR and also ensure that every test has the connectiontag as parameter.. wdyt?

I agree on part, that we shouldn't add any time on test duration. Maybe we just need to check if VM is up somehow. Also, when running tests from `default` VM, those ssh connection timeouts happens instantly (that's why I missed it, since I never run tests from host)

LukasPukenis (commented):
+1

nat-lab/tests/conftest.py
```diff
@@ -192,69 +192,51 @@ async def perform_pretest_cleanups():
...
-        print(e)
+    for item in items:
+        for mark in item.own_markers:
+            if mark.name == "windows" and not windows_bins_copied:
+                await _copy_vm_binaries(ConnectionTag.WINDOWS_VM_1)
+                await _copy_vm_binaries(ConnectionTag.WINDOWS_VM_2)
+                windows_bins_copied = True
+            elif mark.name == "mac" and not mac_bins_copied:
+                await _copy_vm_binaries(ConnectionTag.MAC_VM)
+                mac_bins_copied = True
```
lcruz99, line 231: Changed the logic so that we copy only based on the pytest.marks. (Also added a few marks to the tests that were missing it)

Hasan6979 (approved):
Lgtm!

nat-lab/tests/config.py
```diff
@@ -21,6 +21,8 @@ def get_root_path(path: str) -> str:
 WINDOWS_1_VM_IP = "10.55.0.13"
 WINDOWS_2_VM_IP = "10.55.0.14"
 
+VM_TCP_CONN_TIMEOUT = 10
```
mathiaspeters, line 24: It would be nice to know, either from naming or a comment, what unit this is.  I'm assuming it's in seconds?

lcruz99: correct, will add!

mathiaspeters (commented):
+1

mathiaspeters (approved):

gytsto:
Again, test w/o mark will not have binaries copied. Basically should've left as it was initially, just add that `VM_TCP_TIMEOUT` imho

PR #776, opened 2024-08-14
Sanitize nicknames from logs

lcruz99:
Sanitize nicknames making them hidden from the logs, just like hostnames, the difference is that nicknames are wrapped in an Option<>, telling if is whether or not set.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/mesh.rs
+++ b/crates/telio-model/src/mesh.rs
@@ -124,7 +124,10 @@ impl From<&PeerBase> for Node {
         Self {
             identifier: peer.identifier.clone(),
             public_key: peer.public_key,
-            nickname: peer.nickname.clone(),
+            nickname: peer
+                .nickname
+                .as_ref()
+                .map(|hidden_nick| hidden_nick.0.to_owned()),
             allowed_ips: peer
                 .ip_addresses
                 .as_ref()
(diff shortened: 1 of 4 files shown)


tomasz-grz (commented):
Nice! just a small detail

crates/telio-model/src/mesh.rs
```diff
@@ -124,7 +124,10 @@ impl From<&PeerBase> for Node {
         Self {
             identifier: peer.identifier.clone(),
             public_key: peer.public_key,
-            nickname: peer.nickname.clone(),
+            nickname: peer
```
mathiaspeters, line 127: Did these really need to be changed? I think `Option<T>` is `Clone` if `T` is `Clone`, which `Hidden` is, so just cloning should be fine. Though I'm guessing you got compiler errors for it?

lcruz99: `PeerBase.nickname` is `Option<Hidden<String>>`, while `Node.nickname` is `Option<String>` so the change was to unwrap the the inner `String` from `Hidden<T>`.

mathiaspeters: Right, makes sense!

mathiaspeters (approved):
+1

tomasz-grz (approved):
+1.0

PR #777, opened 2024-08-16
Make output notifier possible as stdout/stderr callback in natlab

gytsto:
### Problem
To be able to pass output notifier as process callback, we have to wrap it in `on_stdout` function to work, which makes no sense, when we should be able to pass `output_notifier.handle_output` as callback.

### Solution
Make output notifier handle output function possible as callback of process in natlab

- [x] README.md is updated


mathiaspeters (approved):
Looks good +1

PR #783, opened 2024-08-19
Draft: Regression test for firewall

Hasan6979:
### Problem

### Solution


.github/workflows/benchmarks.yml
```diff
@@ -5,8 +5,30 @@ permissions: {}
...
+            cargo bench --features test_utils --bench firewall_bench "64" -- --warm-up-time 1 --measurement-time 1 --save-baseline baseline
+        - name: Compare benchmarks
+          working-directory: crates/telio-firewall
+          run: |
+            result=$(critcmp current baseline -t 5)
+            output=$($output | awk '{if (NR > 2) { if ($10 == 1.00 ||  $14 == 1.00) { if ($10 < $14) {exit 1} } else { if ($13 < $17) {exit 1}}}}')
+            if [ -z "$output" ]; then
+              echo "No performance regression detected."
+            else
+              echo "Performance regression detected! Failing the build."
```
tomaszklak, line 31: This is run on the public github owned runner - we have no control over how busy is the machine any any point in time so I'm not sure that those numbers are meaningful.

Hasan6979: Small differences might not come up, but a slowdown should still be detected, something which would either way fall through as we don't regress.
Increasing the number of runs and taking average, or well self-hosting would be alternatives.

tomaszklak: **Maybe**, but it also means that this is prone to being flaky. Imo, without runner physical machine that is not being used for anything when the benchmarks are being run, this shouldn't be merged in.

.github/workflows/benchmarks.yml
```diff
@@ -5,8 +5,30 @@ permissions: {}
...
+          working-directory: crates/telio-firewall
+          run: | # v4.0.0 is just a placeholder
+            git fetch --all
+            git checkout v5.0.0-rc3
+            cargo bench --features test_utils --bench firewall_bench "64" -- --warm-up-time 1 --measurement-time 1 --save-baseline baseline
+        - name: Compare benchmarks
+          working-directory: crates/telio-firewall
+          run: |
+            result=$(critcmp current baseline -t 5)
+            output=$($output | awk '{if (NR > 2) { if ($10 == 1.00 ||  $14 == 1.00) { if ($10 < $14) {exit 1} } else { if ($13 < $17) {exit 1}}}}')
```
tomaszklak, line 27: nit: this is very unclear - I think a comment explaining the format would be a good addition

.github/workflows/benchmarks.yml
Hasan6979: We should then think about how to do that, as the same problem would come in for any other performance tests even with boringtun.

.github/workflows/benchmarks.yml
```diff
@@ -5,8 +5,30 @@ permissions: {}
...
+      env:
+        CARGO_TARGET_DIR: "/home/runner/work/libtelio/libtelio/target/criterion"
       steps:
         - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+        - name: Install critcmp for benchmark comparison
+          run: cargo install critcmp
         - name: Telio firewall benchmarks
           working-directory: crates/telio-firewall
-          run: cargo bench --features test_utils --bench firewall_bench "64" -- --warm-up-time 1 --measurement-time 1
+          run: cargo bench --features test_utils --bench firewall_bench "64" -- --warm-up-time 1 --measurement-time 1 --save-baseline current
```
tomaszklak, line 16: I think that this (and the other one) benchmark run should be run on the same physical core (and only on that one) to avoid noise from scheduling (can be done with `taskset`), but this only makes sense when we run on a known machine where we can prevent other processes from being scheduled on our selected core.

Overall, I think this is a good source of ideas for running benchmarks in CI: https://manuel.bernhardt.io/posts/2023-11-16-core-pinning/

PR #784, opened 2024-08-20
Conntracker sync

tomasz-grz:
### Problem
Nat-lab tests with ConnectionTracker `get_out_of_limits()` sometimes fail, even though in the logs the connections are present.

### Solution
- Add lock to ensure thread safe operation.
- Add mechanism to ensure we synchronized to a known event.
- Add more logs.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection_tracker.py
+++ b/nat-lab/tests/utils/connection_tracker.py
@@ -1,7 +1,6 @@
 import asyncio
 import platform
 import re
-import time
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import datetime
@@ -82,8 +81,8 @@ def parse_input(input_string) -> FiveTuple:
 
 
 class SynchronizeState(Enum):
-    NOT_SYNCHRONIZED = 0
-    WAITING_TO_SYNCHRONIZE = 1
+    NOT_SYNCED = 0
+    WAITING_TO_SYNC = 1
     RECEIVED_SYNC_PING = 2
 
 
@@ -100,7 +99,7 @@ class ConnectionTracker:
         self._config: Optional[List[ConnectionTrackerConfig]] = configuration
         self._events: List[FiveTuple] = []
         self._lock: asyncio.Lock = asyncio.Lock()
-        self._sync_state: SynchronizeState = SynchronizeState.NOT_SYNCHRONIZED
+        self._sync_state: SynchronizeState = SynchronizeState.NOT_SYNCED
         self._sync_event: asyncio.Event = asyncio.Event()
         self._sync_connection: FiveTuple = FiveTuple(
             protocol="icmp", dst_ip="127.0.0.2"
@@ -108,21 +107,17 @@ class ConnectionTracker:
 
     async def on_stdout(self, stdout: str) -> None:
         if not self._config:
-            print(datetime.now(), "ConnectionTracker on_stdout no config")
             return
 
         async with self._lock:
-            print(datetime.now(), "ConnectionTracker on_stdout lock")
             for line in stdout.splitlines():
                 connection = parse_input(line)
                 if connection is FiveTuple():
                     continue
 
-                if self._sync_state == SynchronizeState.WAITING_TO_SYNCHRONIZE:
-                    print(datetime.now(), "ConnectionTracker waiting for synchronize ping")
+                if self._sync_state == SynchronizeState.WAITING_TO_SYNC:
                     if self._sync_connection.partial_eq(connection):
                         self._sync_state = SynchronizeState.RECEIVED_SYNC_PING
-                        print(datetime.now(), "ConnectionTracker received synchronize ping")
                         continue
 
                 matching_configs = [
@@ -142,7 +137,6 @@ class ConnectionTracker:
         if platform.system() == "Darwin":
             return None
         if not self._config:
-            print(datetime.now(), "ConnectionTracker execute no config")
             return
 
         await self._process.execute(stdout_callback=self.on_stdout)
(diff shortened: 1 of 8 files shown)


nat-lab/tests/utils/connection_tracker.py
```diff
@@ -102,23 +107,35 @@ async def on_stdout(self, stdout: str) -> None:
...
-                cfg for cfg in self._config if cfg.target.partial_eq(connection)
-            ]
-            if not matching_configs:
-                continue
+                self._events.append(connection)
 
-            self._events.append(connection)
+            # we received the sync ping
+            if self._synchronize and received_sync:
+                self._synchronize = False
```
Hasan6979, line 137: Can use an enum here instead.

Hasan6979:
lgtm +1.0

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -92,8 +102,10 @@ def __init__(
         self._connection: Connection = connection
         self._config: Optional[List[ConnectionTrackerConfig]] = configuration
         self._events: List[FiveTuple] = []
+        self._lock: Lock = threading.Lock()
```
packgron, line 105: Do we need this lock here? Python asyncio default runtime on a single threaded event loop, unlike rust's multi thread pool.

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -92,8 +102,10 @@ def __init__(
...
 
         self._initialized: bool = False
+        self._synchronize: SynchronizeState = SynchronizeState.NOT_SYNCHRONIZED
         self._init_connection: FiveTuple = FiveTuple(
```
packgron, line 109: Let's rename it to sync connection?
Also maybe we should ping somthing like 127.0.0.2, to make it more unique? ()

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -102,23 +114,33 @@ async def on_stdout(self, stdout: str) -> None:
...
-            connection = parse_input(line)
-            if connection is FiveTuple():
-                continue
+        with self._lock:
+            for line in stdout.splitlines():
+                connection = parse_input(line)
+                if connection is FiveTuple():
+                    continue
+
+                if not self._initialized:
```
packgron, line 123: What is the point of this, if we are now have full state object

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -92,8 +102,10 @@ def __init__(
         self._connection: Connection = connection
         self._config: Optional[List[ConnectionTrackerConfig]] = configuration
         self._events: List[FiveTuple] = []
+        self._lock: Lock = threading.Lock()
 
         self._initialized: bool = False
+        self._synchronize: SynchronizeState = SynchronizeState.NOT_SYNCHRONIZED
```
packgron, line 108: I would recommend to use https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock
or https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event
This has a good explanation: https://stackoverflow.com/questions/25799576/whats-python-asyncio-lock-for

nat-lab/tests/utils/connection_tracker.py
tomasz-grz: Would you suggest using `asyncio.Lock()` instead then or just remove it all together?
my intention was to make sure we wait for all of the stdout lines to be processed before checking the events in `get_out_of_limits`, with the added bonus of being thread-safe

nat-lab/tests/utils/connection_tracker.py
tomasz-grz: Good idea

nat-lab/tests/utils/connection_tracker.py
tomasz-grz: I kept the old initialize mechanism in, but we could also re-use synchronize() for it.

nat-lab/tests/utils/connection_tracker.py
packgron: that or Event

nat-lab/tests/utils/connection_tracker.py
tomasz-grz: There's just one downside of using `asyncio.Event`, is that it has no `timeout`, so it could get locked for ever.. or until the test fails after 120s.
[threading.Event](https://docs.python.org/3/library/threading.html#event-objects) on the other hand does have a `timeout`.

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -128,42 +141,72 @@ async def execute(self) -> None:
...
-            async with Ping(self._connection, "127.0.0.1").run():
-                start_time = time.time()
-                while not self._initialized:
-                    if time.time() - start_time >= 1:
-                        self._initialized = True
-                        break
-                    await asyncio.sleep(0.1)
-
+            # initialization is just waiting for first conntrack event
+            await asyncio.sleep(0.1)  # take a nap to settle things in
```
packgron, line 210: This should not be needed, a noop sleep is tipicaly usefull then you have no other awaits in some long loop.
And now you also have wait also.

Or you tested that this was needed?

nat-lab/tests/utils/connection_tracker.py
packgron: you can always wrap async method with timeout(some awaitable) and get a timeout if you need

packgron: timeout are never an issue in async code

tomasz-grz: Wrapped the event.wait() method in timeout

nat-lab/tests/utils/connection_tracker.py
tomasz-grz: this was tested and does not without it unfortunately.. for some reason it takes a moment before conntrack on_stdout is ready to process the events

packgron (approved):
+2.0, merge and cleanup commits

PR #786, opened 2024-08-20
Nat-lab: Check logs for errors

jjanowsk:
Introduce checking for errors in our logs. It is possible that some tests like e.g. test_adapter_gone_event will emit error logs and this is expected. For that reason self._allowed_errors was introduced in telio.py. But most happy path tests should not emit them. Tests that should be checked and cleaned up are marked with LLT-5532 tag and the jira task is created for it. 

- [x] README.md is updated


mathiaspeters (commented):
+1

Jauler (approved):
+2.0

I guess this may be rather a source of flakyness, but in the long-term, it is likely, that after going through initial "hurdle" - it may be help resolve many internal issues. :rocket:

PR #787, opened 2024-08-21
LLT-5533: Use feature config from generated code

mathiaspeters:
We have several types from libtelio reimplemented in natlab, meaning that have double the maintenance work for the types, and we aren't fully dogfooding our generated bindings. This is the first in a series of PRs to start using generated types instead of custom types, where possible, in natlab. This first PR deals with the feature config.

There are two commits with separate concerns that are best reviewed separately.

The first commit adds serialization logic so that Pyro5 knows how to pass the objects back and forth. This is basically just serializing all our generated types to and from python dicts, done recursively. The classes to (de)serialize could be specified manually, but doing so would mean extra work whenever new types are added, like a new feature config struct. In the current setup, all of our own types in libtelio are automatically possible to (de)serialize. There is also a bindings utils file added where convenience functions and such can be added, and it has a wildcard import for all types in the generated bindings so that stuff can be imported from `utils.bindings` without having to manually add it to the list of imports there, and in doing so the imports elsewhere can be made more consistent since it all comes from `utils.bindings`, rather than some coming from `utils.bindings` and some coming from `uniffi.telio_bindings`.

The second commit is what actually starts using the generated bindings for the feature config, and it's pretty straightforward. It fully removes the old `TelioFeatures` along with `test_telio_features.py` and all references to the feature config replaced with that of the generated bindings. The diff looks big and messy, but there's no logic changes or anything like that. If the diff is checked in the terminal with something like `difftastic`, the diff is quite small and focused.

- [x] README.md is updated


nat-lab/tests/uniffi/serialization.py
```diff
@@ -0,0 +1,62 @@
+import inspect
+import Pyro5.api  # type: ignore
+
+
+def init_serialization(libtelio):
```
stalowyjez, line 5: Hmm, how is this serialization used?

nat-lab/tests/utils/bindings.py
```diff
@@ -0,0 +1,49 @@
...
+def features_with_endpoint_providers(
+    providers: Optional[list[EndpointProvider]],
+) -> Features:
+    features = test_features(lambda builder: builder.enable_direct())
+    assert features.direct
+    features.direct.providers = providers
+    return features
+
+
+def test_features(
```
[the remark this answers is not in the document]
mathiaspeters: I renamed it to `default_features`

nat-lab/tests/utils/bindings.py
```diff
@@ -0,0 +1,49 @@
+from typing import Callable, Optional
+
+from uniffi.telio_bindings import *  # pylint: disable=wildcard-import, unused-wildcard-import  # isort: skip
```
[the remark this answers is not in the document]
mathiaspeters: The idea here is that this file exports all the types from the generated bindings so that anywhere you want to use those types you can import them from this file (cleans up the imports a bit since there are utilities here and then you don't have separate imports for "direct from generated bindings" and "from binding utilities"). It would definitely work without the wildcard but then if you want to use a type that wasn't explicitly exported already you have to add it here manually, to what would become a very big list of imports. I generally don't like wildcard imports but I think in this case it's the better choice

nat-lab/tests/utils/bindings.py
```diff
@@ -0,0 +1,49 @@
...
+def feature_nurse(
+    heartbeat_interval: int = 3600,
+    initial_heartbeat_interval: int = 300,
+    qos: Optional[FeatureQoS] = FeatureQoS(
+        rtt_interval=300, rtt_tries=3, rtt_types=[RttType.PING], buckets=5
+    ),
+    enable_nat_type_collection: bool = False,
+    enable_relay_conn_data: bool = True,
+    enable_nat_traversal_conn_data: bool = True,
+    state_duration_cap: int = 24 * 60 * 60,
```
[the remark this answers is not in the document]
mathiaspeters: This was a leftover I had forgotten to fix, but it's much simplified now

nat-lab/tests/utils/bindings.py
```diff
@@ -0,0 +1,49 @@
+from typing import Callable, Optional
```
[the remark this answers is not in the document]
mathiaspeters: It's going to have utilities for more of the bindings than just features, so I wouldn't make the name more specific, and since it's already in the `utils` folder, I think adding `helpers` to the name is a bit redundant

nat-lab/tests/uniffi/libtelio_proxy.py
```diff
@@ -129,5 +132,5 @@ def receive_ping(self) -> str:
     def probe_pmtu(self, host: str) -> int:
         return self.handle_remote_error(lambda r: r.probe_pmtu(host))
 
-    def get_nat(self, ip: str, port: int) -> NatType:
+    def get_nat(self, ip: str, port: int) -> libtelio.NatType:
```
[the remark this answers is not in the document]
mathiaspeters: We were inconsistent across natlab with which way we were doing this sort of thing so I decided to make them the same, and I went with the option I prefer :shrug:

nat-lab/tests/test_direct_connection.py
```diff
@@ -43,20 +42,24 @@
...
-    left: Tuple[ConnectionTag, List[str], Optional[Batching]],
-    right: Tuple[ConnectionTag, List[str], Optional[Batching]],
+    left: Tuple[ConnectionTag, List[EndpointProvider], Optional[FeatureBatching]],
+    right: Tuple[ConnectionTag, List[EndpointProvider], Optional[FeatureBatching]],
 ) -> List[SetupParameters]:
+    def features(
+        providers: list[EndpointProvider], batching: Optional[FeatureBatching]
+    ) -> Features:
+        features = test_features(lambda builder: builder.enable_direct())
+        features.wireguard.persistent_keepalive.direct = 10
```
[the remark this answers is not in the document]
mathiaspeters: I didn't notice that before but you're right, and it's fixed

nat-lab/tests/test_dns.py
```diff
@@ -373,7 +373,7 @@ async def test_dns_after_mesh_off(alpha_ip_stack: IPStack) -> None:
                             ConnectionTag.DOCKER_CONE_CLIENT_1
                         ),
                         derp_servers=[],
-                        features=TelioFeatures(ipv6=True),
+                        features=test_features(lambda builder: builder.enable_ipv6()),
```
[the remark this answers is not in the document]
mathiaspeters: I changed the `test_features` (now `default_features`) to be more like the old way of calling these

nat-lab/tests/test_dns.py
```diff
@@ -867,6 +867,11 @@ async def test_dns_ttl_value() -> None:
 
         EXPECTED_TTL_VALUE = 1234
 
+        def features() -> Features:
+            features = test_features()
+            features.dns = FeatureDns(exit_dns=None, ttl_value=EXPECTED_TTL_VALUE)
+            return features
+
```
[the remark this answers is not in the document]
mathiaspeters: In this specific case I think it was okay since the function was defined inside of the test and the only usage was a few lines later, but the function was unnecessary so I defined the features directly instead of wrapping them in a function

nat-lab/tests/test_events_link_state.py
```diff
@@ -26,14 +31,17 @@ def _generate_setup_paramete_pair(
     else:
         count = 0
 
+    features = test_features(lambda builder: builder.enable_link_detection())
+    features.link_detection = FeatureLinkDetection(
+        rtt_seconds=1, no_of_pings=count, use_for_downgrade=False
+    )
+    features.wireguard = long_persistent_keepalive_periods()
```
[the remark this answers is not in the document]
mathiaspeters: Good catch, fixed

nat-lab/tests/uniffi/serialization.py
mathiaspeters: I added a comment to the function to explain a but, but basically it's called at the start of  `libtelio_proxy.py` and `libtelio_remote.py` which makes Pyro5 able to (de)serialize our types on both the test host and in the test containers/VMs

stalowyjez:
Looks good now, +1

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -5,12 +5,17 @@
...
+from serialization import (  # type: ignore # pylint: disable=import-error
+    init_serialization,
+)
 from threading import Lock
 from typing import List
 
 REMOTE_LOG = "remote.log"
 TCLI_LOG = "tcli.log"
 
+init_serialization(libtelio)
```
Hasan6979, line 17: Could you add a small comment here and in `libtelio_proxy` on what this function is doing. A small 1-liner so that anyone going working on the code can understand as just by reading it doesn't feel exactly clear what this would do.

nat-lab/tests/uniffi/serialization.py
```diff
@@ -0,0 +1,62 @@
...
+    all_members = inspect.getmembers(libtelio, inspect.isclass)
+    public_members = []
+    for name, cls in all_members:
```
Hasan6979, line 17: Maybe this is okay, I don't know for sure. But we are going only 1 level deep here. What if we switched to calling this recursively ? IMO this feels a bit rigid.

nat-lab/tests/uniffi/serialization.py
```diff
@@ -0,0 +1,62 @@
...
+                if not name.startswith("_"):
+                    public_members.append(cls)
+
+    classes = tuple(
+        filter(
+            lambda m: str(m).startswith("<class ") and libtelio.__name__ in str(m),
+            public_members,
+        )
+    )
+    enums = tuple(filter(lambda m: str(m).startswith("<enum "), public_members))
```
Hasan6979, line 31: For filtering out enums and classes, did you explore inspect.issubclass ? Just wondering why a string based filtering instead instead of a similar approach as above.

nat-lab/tests/uniffi/serialization.py
```diff
@@ -0,0 +1,62 @@
...
+            class_name = type(obj).__name__
+            out = {
+                "__class__": class_name,
+                "__dict__": {k: serialize_obj(v) for k, v in obj.__dict__.items()},
+            }
+            return out
+        if isinstance(obj, enums):
+            enum_name = type(obj).__name__
+            return {"__enum__": enum_name, "value": obj.name}
+        if isinstance(obj, list):
```
Hasan6979, line 44: What if an object is neither ?

nat-lab/tests/test_direct_connection.py
```diff
@@ -300,31 +309,35 @@ async def test_direct_working_paths_are_reestablished_and_correctly_reported_in_
...
             )
-            param.features.lana = Lana(prod=False, event_path="/event.db")
+            param.features.lana = FeatureLana(prod=False, event_path="/event.db")
         env = await setup_mesh_nodes(exit_stack, setup_params)
         alpha, beta = env.nodes
         alpha_client, beta_client = env.clients
 
-        def fix_provider_name(name):
-            return "UPnP" if name == "upnp" else name.title()
+        def get_provider_name(provider: EndpointProvider):
```
Hasan6979, line 320: I would rename this as `determine_provider_name` or `resolve_provider_name`

nat-lab/tests/test_pmtu.py
```diff
@@ -14,6 +14,13 @@
 from utils.router import IPStack
 
 
+def features() -> Features:
+    f = test_features(lambda builder: builder.enable_pmtu_discovery())
```
Hasan6979: Would `enable_xxx` be required if we want to add any feature other than what is there by default ?

nat-lab/tests/uniffi/serialization.py
mathiaspeters: When I switch to calling it recursively the tests don't start because it enters an infinite loop. Basically, everything is an object, `inspect.isclass` returns more than I think it should, so there is no end to getting nested types

nat-lab/tests/uniffi/serialization.py
mathiaspeters: Then it it passed over the Pyro5 boundary as-is. The intention for it is to only apply to primitive types. The only type which may become relevant to add here at some point is `dict`, but we don't have any such cases in the udl file right now so I skipped adding it here

nat-lab/tests/test_pmtu.py
mathiaspeters: This pattern was changed to not use lambda+builder like here, but I guess the question still stands. 

Using `pmtu_discovery` as an example: `pmtu_discovery is disabled by default so if you want to enable it, you need to set `enable_pmtu_discovery=True`, which would then enable it with the default value for `response_wait_timeout_s` which is 5. If you wanted a different value for `response_wait_timeout_s`, you could skip `enable_pmtu_discovery=True` and instead do `features.pmtu_discovery=FeaturePmtuDiscovery(<your value here>)` to set it afterwards. But if we instead think about `nurse`, which has many more fields, then it makes more sense to `enable_nurse=True` so you get the default values for all of its fields and then update the fields you want to change, because otherwise you have to create the whole `FeatureNurse` object from scratch

nat-lab/tests/uniffi/serialization.py
Hasan6979: Idk if it will work, but can you try it out with `issubclass` and `not issubclass` and see ? If it works thats great, if not then keep it as it is!

Hasan6979 (approved):
+1

PR #789, opened 2024-08-21
Lukasp/enable batching natlab direct conns

LukasPukenis:
Previously batcher was only used in a disabled form(feature was set to `None`) or with threshold of 0, effectively simulating no batching at all. This PR adds non-zero batcher to direct connection tests to prove a point that it doesn't break direct connections when enabled.

The threshold and keepalive adhere to the rule of threshold being half the period.

- [x] README.md is updated


tomaszklak (approved):
+2

PR #794, opened 2024-08-23
natlab: Trigger garbage collection when dropping node

lcruz99:
### Problem
This test starts one of the nodes without any EP and expects to accomplish relay path connection. Then it recreates the node to add the stun server as EP, expecting then a direct path connection.

It was failing locally because although the `telio.Client` instance was dropped it was ran inside the main exit stack, so not collected by the python's gc while inside the exit stack. This resulted in a failed second `telio.Client` init with a `Address already in use` OS error when attempting the Pyro5 connection.

### Solution
This commit replaces the `setup_mesh()` function by individual exit stacs, so that the dropped node is fully cleaned by the python's garbage collector before recreating the libtelio remote connection.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_events.py
+++ b/nat-lab/tests/test_events.py
@@ -4,11 +4,16 @@ import pytest
 import timeouts
 from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_environment, setup_mesh_nodes, setup_api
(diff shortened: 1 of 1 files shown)


mathiaspeters (approved):
+1

PR #796, opened 2024-08-23
natlab: Add comments differenciating primary routes switch

lcruz99:
### Problem
On natlab machines, switching to primary/secondary network has different implementations wether we're working with the Windows/Mac VM or the docker containers.
This is implicit knowledge, however the developer might not be aware of these differences and it can be a bug source in the future.

### Solution
Add function documentation to state what each network switch does.

- [x] README.md is updated


Hasan6979 (dismissed):
🚀 
+2.0

I had exactly this problem recently!

Hasan6979 (approved):

PR #798, opened 2024-08-26
natlab: Remove VM connection timeout

lcruz99:
### Problem
This connection timeout was added to avoid having the natlab run locked for 5min (ie: default tcp timeout) if any problem arises. It happens that this timeout was too short and for now the best solution is to remove it.

### Solution
Remove the timeout.

- [x] README.md is updated


packgron (approved):
+2.0

PR #802, opened 2024-08-27
[LLT-5519] Fix roaming vs cross_ping_check racing

packgron:
### Problem

Until now it was possible that wg event comming into device.rs and
tick happening in device.rs, could lead to invalid race condition.

If in that situation, select pick's consilidation step, while
old endpoint exists in cross ping check, and peer just downgraded
to relay do to roaming, wg_controller would again put that
peer to an invalid direct connection.

### Solution
Eposed Instant information, next to compared endpoints to ensure correct causality order

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5519
+++ b/.unreleased/LLT-5519
@@ -1,0 +1,1 @@
+Fixed a low-probability race-condition which may cause unnecessary path transitions
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -711,9 +711,9 @@ async fn build_requested_meshnet_peers_list<
     for (public_key, requested_peer) in requested_peers.iter_mut() {
         // Gather required information
         let actual_peer = actual_peers.get(public_key);
-        let time_since_last_endpoint_change = wireguard_interface
-            .time_since_last_endpoint_change(*public_key)
-            .await?;
+        let time_since_last_endpoint_change = actual_peer
+            .and_then(|p| p.endpoint_changed_at)
+            .map(|t| t.elapsed());
         let checked_endpoint = checked_endpoints.get(public_key);
         let proxy_endpoint = proxy_endpoints.get(public_key);
         let upgrade_request_endpoint = upgrade_request_endpoints
@@ -766,6 +766,7 @@ async fn build_requested_meshnet_peers_list<
                 None => &[],
             },
             &upgrade_request_endpoint,
+            cross_ping_check,
         )
         .await?;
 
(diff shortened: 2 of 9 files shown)


src/device/wg_controller.rs
```diff
@@ -2016,6 +2043,57 @@ mod tests {
         f.consolidate_peers().await;
     }
 
+    #[tokio::test]
+    async fn when_cross_check_vailidated_is_old_then_keep_proxying() {
```
Jauler, line 2047: Nit:
```suggestion
    async fn when_cross_check_validated_is_old_then_keep_proxying() {
```

src/device/wg_controller.rs
```diff
@@ -711,9 +712,9 @@ async fn build_requested_meshnet_peers_list<
     for (public_key, requested_peer) in requested_peers.iter_mut() {
         // Gather required information
         let actual_peer = actual_peers.get(public_key);
-        let time_since_last_endpoint_change = wireguard_interface
-            .time_since_last_endpoint_change(*public_key)
-            .await?;
+        let time_since_last_endpoint_change = actual_peer
+            .and_then(|p| p.endpoint_changed_at)
+            .map(|t| Instant::now() - t);
```
matislovas, line 717: nit: can be changed to `t.elapsed()`

.unreleased/LLT-5519
Jauler, line 1: Maybe it is worth adding something in changelog or this :thinking:, as this is while, yeah, quite an edge-case, but still bugfix :). something like:

> Fixed a low-probability race-condition which may cause unnecessary path transitions

tomasz-grz (commented):
+0.5

matislovas:
+1

src/device/wg_controller.rs
```diff
@@ -841,12 +841,24 @@ async fn select_endpoint_for_peer<'a>(
...
 ) -> Result<(Option<SocketAddr>, Option<EndpointState>)> {
     // Retrieve some helper information
-    let actual_endpoint = actual_peer.clone().and_then(|p| p.endpoint);
+    let actual_endpoint = actual_peer.as_ref().and_then(|p| p.endpoint);
+
+    let is_actual_endpoint_newer = {
+        let changed_at = actual_peer.as_ref().and_then(|p| p.endpoint_changed_at);
+        let checked_at = checked_endpoint.as_ref().map(|e| e.changed_at);
+        match (changed_at, checked_at) {
+            (Some(wg), Some(cpc)) => wg >= cpc,
```
Jauler, line 855: I just realized one edge-case, which is a bit sus. Therefore would like to discuss this.

Lets imagine, that for some reason, telio-proxy decides to change it's mapping and thus relay->relay endpoint change occurs. I think this will not trigger endpoints to be invalidated in device/mod.rs, meaning that CPC will not attempt to publish any endoints, but this conditions will prevent to use them for the forseeable future.

Jauler: To fix it, in we may add this condition, that in case CPC is holding checked endpoints, which are older than currently used relay endpoint -> we must invalidate them. 

This may replace existing `was_proxing == false && is_proxying == true` condition. As the one based on `changed_at` is more general and independant of execution cycle :+1:

packgron: It would be very rare, the only reason for proxy to change is socket failure.

The wailure scenario would be that we would stop not do endpoint discovery for some time?

packgron: Yeh, I think it's fair, we will be independent of other places

Jauler: > The wailure scenario would be that we would stop not do endpoint discovery for some time?

Failure scenario would be that no connection upgrades to direct would occur until restart or network change.

Jauler (dismissed):
+1.0

Jauler (approved):

PR #803, opened 2024-08-28
Fix incorrect entries in nat_traversal_conn_info

jjanowsk:
Previusly report_peer_state_direct and report_peer_state_relayed were triggered by different events. Because of that there was a possibility for the reorder of those events causing incorrect reporting. Now we report it always from a single place in the wg_controller so there is no risk of a race condition.

- [x] README.md is updated


src/device.rs
```diff
@@ -2297,12 +2295,8 @@ impl TaskRuntime for Runtime {
                                 // We have downgraded the connection. Notify cross ping check about that.
                                 direct_entities.cross_ping_check.notify_failed_wg_connection(public_key)
                                     .await?;
-
                                 direct_entities.session_keeper.remove_node(&public_key).await?;
-
-                                let mut event = AnalyticsEvent::from_event(&mesh_event);
-                                event.peer_state = PeerState::Connected;
-                                self.entities.aggregator.change_peer_state_relayed(&event).await;
+                                direct_entities.upgrade_sync.clear_accepted_session(public_key).await;
```
packgron, line 2299: We only clear session on downgrade? Should we not also clear it on connection to direct, or is that not detectable currently?

packgron:
You will need a rebase

crates/telio-nurse/src/aggregator.rs
```diff
@@ -257,7 +257,25 @@ impl ConnectivityDataAggregator {
         .await
     }
 
-    async fn change_peer_state_common(&self, event: &AnalyticsEvent, endpoints: PeerEndpointTypes) {
+    fn log_new_state(event: &AnalyticsEvent, endpoints: &PeerEndpointTypes) {
+        if endpoints.local_ep == EndpointType::Relay && endpoints.remote_ep == EndpointType::Relay {
+            telio_log_debug!(
+                "Relayed peer state change for {} to {:?} will be reported",
```
Jauler, line 263: IMHO we can bump these to be `info` level.

src/device/wg_controller.rs
```diff
@@ -236,7 +255,31 @@ async fn consolidate_wg_peers<
...
+            dual_ip_addresses: actual_peer.get_dual_ip_addresses(),
+            tx_bytes: actual_peer.tx_bytes.unwrap_or_default(),
+            rx_bytes: actual_peer.rx_bytes.unwrap_or_default(),
+            peer_state: actual_peer.state(),
+            timestamp: Instant::now(),
+        };
+        if is_actual_peer_proxying {
+            aggregator.report_peer_state_relayed(&event).await;
+        } else if let Some(us) = upgrade_sync {
+            if let Some(accepted_direct_session) = us.get_accepted_session(*key).await {
```
Jauler, line 271: :thinking: I am afraid there might be edge-case here.

Lets imagine we have accepted some upgrade, and upgraded connection to direct. Later, for some reason, another upgrade sync request is sent, it gets accepted, but because we are already in direct, no endpoint change in wireguard will occur. This would report new upgrade sync info, but keep old wireguard endpoint for, potentially, indefinite amount of time.

In order to fix it, one possible approach would be to expand accepted_session with LRU cache, which would hold some bounded amount of upgrade sync sessions, keyed by endpoint (`IP:port`). In here we would get accepted session by pubic key as well as currently "installed" endpoint in wireguard. If no such session could be found, we can report "Unknown:Unknown" as endpoint provider (**Note:** [this](https://github.com/NordSecurity/libtelio/blob/v5.0.0-rc4/crates/telio-nurse/src/aggregator.rs#L45) would have to be expanded with "Unknown" case.)

Jauler: oh

Actually I am wrong. As new upgrade request would get honored even if we are already in direct. And in practice, if we would receive upgrade syn request, and are "planning" to ignore it, it is probably better to reject it instead of accepting it.

Ok, so this edge-case turned out to be not real.

crates/telio-nurse/src/aggregator.rs
```diff
@@ -1635,7 +1639,7 @@ mod tests {
         current_peer_event.tx_bytes += 2000;
         current_peer_event.peer_state = NodeState::Disconnected;
         env.connectivity_data_aggregator
-            .change_peer_state_direct(
+            .report_peer_state_direct(
                 &current_peer_event,
                 EndpointProvider::Upnp,
                 EndpointProvider::Local,
```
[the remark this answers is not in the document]
jjanowsk: There already is one: `test_aggregator_multiple_events_single_segment`

crates/telio-traversal/src/upgrade_sync.rs
```diff
@@ -477,16 +468,10 @@ impl Runtime for State {
                 match msg.decision {
                     Decision::Accepted => {
                         if let Some(data) = self.pending_direct_sessions.remove(&msg.session) {
-                            if let Some(event) = self.make_nat_event(public_key).await {
-                                self.connectivity_data_aggregator.
-                                    change_peer_state_direct(&event, data.local_ep, data.remote_ep).await;
-                            } else {
-                                telio_log_warn!("Received upgrade decision from peer {public_key:?} to which we didn't send a request: {msg:?}");
-                            }
+                            self.accepted_direct_sessions.insert(public_key, SessionData {local_ep: data.local_ep, remote_ep: data.remote_ep});
```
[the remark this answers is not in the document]
jjanowsk: Sorry, I don't understand the question

src/device/wg_controller.rs
```diff
@@ -4,16 +4,17 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::iter::FromIterator;
 use std::net::IpAddr;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
```
packgron, line 7: It's better to use tokio::time::Instant, it reacts to tokio mock'ed time (pause, unpause, advance etc) and is aready used in most of telio codebase

packgron: I guess it comes from previous code, but probably would be nice to switch :sweat_smile:

packgron (commented):
+1.0

packgron (commented):
Reaprove from my side: +1

Jauler (approved):
+1.0

PR #807, opened 2024-09-03
Llt 5483 big batching test

LukasPukenis:
### Problem
Batching tests, as per design session, should be star-like with multiple different nodes and idling. Then we should observe batching over the course of the time and see that it's working.

### Solution
Setup preferred meshnet topology, launch tcpdump on nodes of interest, then aggregate packet captures by using filters into histograms and display them.

Existing configuration in this PR is pretty minimal, just few directly connected nodes. Windows, MacOS, VPN, Proxied peers will be added next.

As a result currently we can see some histograms but we still need to:
- add proper configuration
- add various filters to observation while we don't have fully implemented the batching

- [x] README.md is updated

### DIFF
--- a/nat-lab/Pipfile.lock
+++ b/nat-lab/Pipfile.lock
@@ -953,6 +953,14 @@
             "markers": "python_version >= '3.8'",
             "version": "==2.32.3"
         },
+        "scapy": {
+            "hashes": [
+                "sha256:5b260c2b754fd8d409ba83ee7aee294ecdbb2c235f9f78fe90bc11cb6e5debc2"
+            ],
+            "index": "pypi",
+            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3' and python_version < '4'",
+            "version": "==2.5.0"
+        },
         "serpent": {
             "hashes": [
                 "sha256:0407035fe3c6644387d48cff1467d5aa9feff814d07372b78677ed0ee3ed7095",
--- a/nat-lab/tests/test_batching.py
+++ b/nat-lab/tests/test_batching.py
@@ -1,0 +1,189 @@
+# Packet captures and histograms are used to visually and in-code observe the batching in action.
+import asyncio
+import os
+import pytest
+import telio
+from contextlib import AsyncExitStack
+from helpers import setup_mesh_nodes, SetupParameters
+from itertools import product
+from scapy.layers.inet import IP, UDP  # type: ignore
+from scapy.layers.inet6 import ICMPv6EchoRequest, IPv6  # type: ignore
+from telio import PathType, State, AdapterType
+from telio_features import (
+    Batching,
+    LinkDetection,
+    PersistentKeepalive,
+    TelioFeatures,
+    Direct,
+    Wireguard,
+)
+from typing import List
+from utils.batching import capture_traffic, save_histogram, generate_histogram_from_pcap
+from utils.connection import DockerConnection
+from utils.connection_util import DOCKER_GW_MAP, ConnectionTag, container_id
+
+
+def _generate_setup_parameters(
+    conn_tag: ConnectionTag,
+    adapter: telio.AdapterType,
+    endpoint_providers: List[str],
+    direct_enabled: bool,
+    batching_enabled: bool,
+) -> SetupParameters:
+    features = TelioFeatures(
+        direct=Direct(providers=endpoint_providers) if direct_enabled else None,
+        link_detection=LinkDetection(
+            use_for_downgrade=True, rtt_seconds=1, no_of_pings=1
+        ),
+        wireguard=Wireguard(
+            persistent_keepalive=PersistentKeepalive(
+                direct=70,
+                proxying=70,
+                stun=70,
+                vpn=70,
+            )
+        ),
+        batching=Batching(direct_connection_threshold=35) if batching_enabled else None,
+    )
+
+    item = SetupParameters(
+        connection_tag=conn_tag,
+        adapter_type=adapter,
+        features=features,
+    )
+
+    return item
+
+
+@pytest.mark.batching
+@pytest.mark.asyncio
+@pytest.mark.timeout(200)
+@pytest.mark.parametrize(
+    "setup_params,misalign_sleep_amount",
+    [
+        pytest.param(
+            [
+                _generate_setup_parameters(
+                    ConnectionTag.DOCKER_CONE_CLIENT_1,
+                    AdapterType.LinuxNativeWg,
+                    ["stun", "upnp", "local"],
+                    direct_enabled=True,
+                    batching_enabled=True,
+                ),
+                _generate_setup_parameters(
+                    ConnectionTag.DOCKER_CONE_CLIENT_2,
+                    AdapterType.LinuxNativeWg,
+                    ["stun", "upnp", "local"],
+                    direct_enabled=True,
+                    batching_enabled=True,
+                ),
+                _generate_setup_parameters(
+                    ConnectionTag.DOCKER_UPNP_CLIENT_1,
+                    AdapterType.LinuxNativeWg,
+                    ["stun", "upnp", "local"],
+                    direct_enabled=True,
+                    batching_enabled=True,
+                ),
+                _generate_setup_parameters(
+                    ConnectionTag.DOCKER_UPNP_CLIENT_2,
+                    AdapterType.LinuxNativeWg,
+                    ["stun", "upnp", "local"],
+                    direct_enabled=True,
+                    batching_enabled=True,
+                ),
+            ],
+            6,
+            marks=[pytest.mark.small_batch],
+        ),
+    ],
+)
+async def test_batching_star(
+    setup_params: List[SetupParameters], misalign_sleep_amount: int
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        capture_duration = 100
+
+        env = await setup_mesh_nodes(exit_stack, setup_params)
+
+        await asyncio.gather(*(
+            client.wait_for_state_peer(
+                peer_node.public_key, [State.Connected], [PathType.Direct]
+            )
+            for client, peer_node in product(env.clients, env.nodes)
+            if not client.is_node(peer_node)
+        ))
+
+        # Capturing the traffic can be done on nodes or gateways
+        # On gateways there's a guarantee that the traffic has arrived
+        # however we can't decrypt it easily.
+        # On nodes there's no guarantee the packet left the network adapter
+        # but we can inspect the packet before it hit the TUN device.
+        # Here we capture on both
+        gateways = [DOCKER_GW_MAP[param.connection_tag] for param in setup_params]
+        gateway_container_names = [container_id(conn_tag) for conn_tag in gateways]
+        conns = [client.get_connection() for client in env.clients]
+        node_container_names = [
+            conn.container_name()
+            for conn in conns
+            if isinstance(conn, DockerConnection)
+        ]
+
+        container_names = gateway_container_names + node_container_names
+        # TODO: assert these are docker connections
+        print("Will capture batching on containers: ", container_names)
+        cnodes = zip(env.clients, env.nodes)
+
+        # the plan is to misalign the peers for better batcing observation. WIthout misalignment
+        # peers will form a meshnet almost immediately, already achieving the desired batching effect.
+        for client in env.clients:
+            await client.stop_device()
+
+        # TODO: this is possible to refactor and use `start_tcpdump` and `stop_tcpdump`
+        pcap_capture_tasks = []
+        for name in container_names:
+            task = asyncio.create_task(
+                capture_traffic(
+                    name,
+                    capture_duration,
+                )
+            )
+            pcap_capture_tasks.append(task)
+
+        # misalign the peers by sleeping some before starting each node again
+        for client, node in cnodes:
+            await asyncio.sleep(misalign_sleep_amount)
+            await client.simple_start()
+            await client.set_meshmap(env.api.get_meshmap(node.id))
+
+        await asyncio.gather(*(
+            client.wait_for_state_peer(
+                peer_node.public_key, [State.Connected], [PathType.Direct]
+            )
+            for client, peer_node in product(env.clients, env.nodes)
+            if not client.is_node(peer_node)
+        ))
+
+        pcap_paths = await asyncio.gather(*pcap_capture_tasks)
+
+        filters = [
+            None,
+            ("icmpv6echorequest", lambda p: p.haslayer(ICMPv6EchoRequest)),
+            (
+                "ipv46 udp",
+                lambda p: (
+                    (p.haslayer(IP) or p.haslayer(IPv6)) if p.haslayer(UDP) else True
+                ),
+            ),
+        ]
+
+        for container, pcap_path in zip(container_names, pcap_paths):
+            for filt in filters:
+                filter_name = filt[0] if filt else "none"
+                hs = generate_histogram_from_pcap(
+                    pcap_path, capture_duration, filt[1] if filt else None
+                )
+                title = f"{container}-filter({filter_name})"
+                save_histogram(title, hs, "logs", max_height=12)
+
+        for p in pcap_paths:
+            os.unlink(p)
--- a/nat-lab/tests/utils/batching.py
+++ b/nat-lab/tests/utils/batching.py
@@ -1,0 +1,138 @@
+import asyncio
+import math
+import os
+import pytest
+import subprocess
+import typing
+from scapy.all import PcapReader, Packet  # type: ignore
+from typing import Callable, List, Optional
+
+
+def _generate_histogram(
+    data: list[int], buckets: int, bucket_size: int = 1
+) -> List[int]:
+    assert len(data) > 0
+    max_val = max(data)
+
+    if max_val >= buckets * bucket_size:
+        raise ValueError(
+            f"Histogram doesn't fit the data({max_val}). The max value is {max_val} but the histogram has {buckets} buckets with a width of"
+            f" {bucket_size}, the maximum value it can fit is {buckets * bucket_size}"
+        )
+
+    hs = [0] * buckets
+
+    for v in data:
+        bucket_index = int(v / bucket_size)
+        hs[bucket_index] += 1
+
+    return hs
+
+
+async def capture_traffic(container_name: str, duration_s: int) -> str:
+    cmd_rm = f"docker exec --privileged {container_name} rm /home/capture.pcap"
+    os.system(cmd_rm)
+
+    iface = "any"
+    capture_path = "/home/capture.pcap"
+
+    cmd = f"docker exec -d --privileged {container_name} tcpdump -i {iface} -U -w {capture_path}"
+    res = os.system(cmd)
+    if res != 0:
+        raise RuntimeError(f"Failed to launch tcpdump on {container_name}")
+
+    await asyncio.sleep(duration_s)
+
+    local_path = f"{container_name}.pcap"
+    subprocess.run([
+        "docker",
+        "cp",
+        container_name + ":" + "/home/capture.pcap",
+        local_path,
+    ])
+
+    cmd_rm = f"docker exec --privileged {container_name} pkill tcpdump"
+    os.system(cmd_rm)
+
+    return local_path
+
+
+# Render ASCII histogram drawing for visual inspection
+def save_histogram(name: str, data: List[int], save_dir: str, max_height=None):
+    output = []
+    if not data:
+        output.append(f"No data provided for {name}")
+        return
+
+    max_value = max(data)
+
+    if max_height is None:
+        max_height = max_value
+
+    scaled_data = [math.ceil((value / max_value) * max_height) for value in data]
+    for row in range(max_height, 0, -1):
+        line = ""
+        for value in scaled_data:
+            if value >= row:
+                line += "█"
+            else:
+                line += " "
+        line = "|" + line
+        output.append(line)
+
+    output.append(f"+{'-' * (len(data))}")
+    output.append(f"0{' ' * (len(data)-1)}{len(data)}")
+    output.append(f"^-Histogram of {name}")
+
+    # try to save `name.hist` and warn if it already exists by asserting false
+    file_name = f"{name}.hist"
+    file_path = os.path.join(save_dir, file_name)
+
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write("\n".join(output))
+
+    print(f"Histogram saved successfully as '{file_name}' in '{save_dir}'.")
+
+
+def generate_histogram_from_pcap(
+    pcap_path: str,
+    buckets: int,
+    packet_filter: Optional[Callable[[Packet], bool]],
+) -> typing.List[int]:
+    print("Looking for a pcap at", pcap_path)
+
+    first_packet_time = None
+    timestamps = []
+
+    with PcapReader(pcap_path) as pcap_reader:
+        first_packet = True
+        for pkt in pcap_reader:
+            if first_packet:
+                first_packet_time = pkt.time
+                first_packet = False
+
+            if packet_filter and not packet_filter(pkt):
+                continue
+
+            timestamps.append(pkt.time - first_packet_time)
+
+    # we either filtered out everything or didn't receive any traffic
+    if len(timestamps) == 0:
+        return []
+
+    return _generate_histogram(timestamps, buckets)
+
+
+@pytest.mark.asyncio
+async def test_histogram():
+    data = []
+    for _ in range(10):
+        data.append(2)
+        data.append(3)
+
+    for _ in range(50):
+        data.append(4)
+
+    data.append(9)
+
+    assert _generate_histogram(data, 10, 1) == [0, 0, 10, 10, 50, 0, 0, 0, 0, 1]
--- a/nat-lab/Pipfile
+++ b/nat-lab/Pipfile
@@ -49,7 +49,7 @@ psutil = "*"
 types-psutil = "*"
 asyncssh = "==2.14.2"
 aiodocker = "*"
-
+scapy="*"
 [dev-packages]
 
 [requires]
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -526,6 +526,15 @@ class Client:
         if telio_features.nurse is not None and telio_features.lana is not None:
             self._fingerprint = telio_features.lana.event_path, fingerprint
 
+    def is_node(self, node: Node) -> bool:
+        return self._node == node
+
+    def get_connection(self):
+        return self._connection
+
+    def get_proxy_port(self):
+        return self._proxy_port
+
     @asynccontextmanager
     async def run(self, meshmap: Optional[Meshmap] = None) -> AsyncIterator["Client"]:
         if isinstance(self._connection, DockerConnection):


.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v2.2.2 # REMEMBER to also update in .gitlab-ci.yml
+          triggered-ref: lukasp/batching-tests-long
```
LukasPukenis, line 34: This will be reverted back, but for now it allows me to trigger the gitlab pipeline as the test contains `batching` pytest mark.

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,281 @@
...
+        for name, histograms in zip(container_names, futures):
+            for filter_name, hs in histograms:
+                title = f"{name}-filter({filter_name})-id({request.node.callspec.id})"
+                draw_histogram(title, hs, max_height=12)
+                silence_streak = longest_zero_streak(hs)
+                print(f"Longest silence for {title} is {silence_streak}\n")
+
+        assert (
+            False
+        ), "TODO: implement a proper assert(this assert also forced the dump of output)"
```
LukasPukenis, line 281: So far the results are not yet as we expect. There's Pyro5 traffic, there's STUN traffic from the DERP server, etc. For now the assertion is always False in order to dump the test output and in turn observe the histograms rendered in ascii

.github/workflows/gitlab.yml
LukasPukenis: at the moment, becasue of pytest.mark it will not be executed alongside regular libtelio tests and will not delay those unecessarily

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class Histogram:
+    def __init__(self) -> None:
+        self._values: typing.List[float] = []
+
+    def add_value(self, value: float) -> None:
+        self._values.append(value)
+
+    def get(self, bin_count: int, bin_width: int) -> List[int]:
```
mathiaspeters, line 24: Nit: I think it's more common to talk about buckets than about bins when it comes to histograms

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,162 @@
...
+        if len(self._values) == 0:
+            return []
+
+        sorted_values = sorted(self._values)
+        print("len(sorted_values) -> ", len(sorted_values))
```
mathiaspeters, line 29: Is this planned diagnostics or a leftover from development?

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,162 @@
...
+        max_val = sorted_values[len(sorted_values) - 1]
```
mathiaspeters, line 30: Since this is python, it would be enough to do `sorted_values[-1]`

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,281 @@
...
+            gamma_client.wait_for_state_peer(
+                beta.public_key,
+                [State.Connected],
+                [PathType.Direct],
+            ),
+        )
+
+        gateway_container_names = []
+        for conn_tag in gateways:
+            gateway_container_names.append(container_id(conn_tag))
```
mathiaspeters, line 189: I would move this block to where you defined `gateways` (or move the definition of `gateways` to just above this block), to make it a little bit easier to read. I would also suggest changing this to `gateway_container_names = [container_id(conn_tag) for conn_tag in gateways]`

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,281 @@
...
+        conns = [client._connection for client in env.clients]
+        node_container_names = [
+            conn.container_name()
+            for conn in conns
+            if isinstance(conn, DockerConnection)
+        ]
+
+        container_names = gateway_container_names + node_container_names
+        # TODO: so far assuming these are docker containers
+        print("container names: ", container_names)
```
mathiaspeters, line 200: Planned diagnostic or leftover?

mathiaspeters:
I left a few nits but I also have a couple of more general questions:

1. Why add custom tcpdump collection for this test? We already have infrastructure in place to capture pcaps on all docker containers and I'm probably missing something but why not rely on that?

2. I'm a little bit concerned with adding this to our test suite since these tests will add a lot of execution time to our already long-running pipelines. Maybe it would be an idea to mark the test as `long`? Although that brings with it the problem that it wouldn't ever be run automatically since we never run the `long` tests in any pipeline

LukasPukenis:
@mathiaspeters good concerns, let me clarify:
- I came to realise about `tcpdump` already collecting captures a bit later afterwards, that's the whole explanation :) I still think it's nice to capture separately and process as our captures might introduce a filter or so and ruin the results. So far I think it shouldn't pose a problem and we can remove it if unecessary in the near future.
- This test so far should be running separately, triggered by a nightly pipeline, not on each branch push, hence why I've added the `batching` marker for `pytest`, so it would basically not be even captured by `long` as this one should have a big timeout. In our discussions we clarified that it would be nice to run it for around 5 minutes in total while idling to see what is happening. tldr: this shouldn't run even if merged alongside other tests neither nightly nor regular.

nat-lab/tests/test_batching.py
LukasPukenis: Nice to see which containers are gonna receive `tcpdump` treatment :)

nat-lab/tests/utils/batching.py
LukasPukenis: leftover, thanks for pointing out, removed

nat-lab/tests/utils/batching.py
LukasPukenis: Good point to be honest. I've renamed all `bin` occurences into `bucket`

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,162 @@
...
+
+        if max_val >= bin_count * bin_width:
+            raise ValueError(
+                f"Histogram doesn't fit value of {max_val}. Has {bin_count} with width of"
+                f" {bin_width}"
```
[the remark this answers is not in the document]
LukasPukenis: Rewrote it

nat-lab/tests/test_batching.py
LukasPukenis: fixed

mathiaspeters:
+1

Looks good! If the tcpdump solution that is used for this test starts being used anywhere else, I think we should try to think of a more general solution, but until then, all good :)

nat-lab/Pipfile.lock
```diff
@@ -953,6 +953,14 @@
             "markers": "python_version >= '3.8'",
             "version": "==2.32.3"
         },
+        "scapy": {
+            "hashes": [
+                "sha256:5b260c2b754fd8d409ba83ee7aee294ecdbb2c235f9f78fe90bc11cb6e5debc2"
+            ],
+            "index": "pypi",
+            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3' and python_version < '4'",
```
stalowyjez, line 961: Why such python versions? Especially why python 4 is bad?

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,161 @@
+# There's a dillemma in how to best capture the traffic. Capturing on the node itself
+# provides insights into the packets themselves coming to and from the tunnel device,
+# however capturing locally doesn't mean the packet actually left the machine.
+# An alternative is to capture on gateways which guarantees the packet arrival
+# however there's no way to easily pry open the packets and observe their insides.
```
stalowyjez, line 5: I guess it would be nice if this comment would actually say which way of capturing them is implemented here.

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,161 @@
...
+import os
+import pytest
+import subprocess
+import typing
+from scapy.all import PcapReader  # type: ignore
+from scapy.layers.inet import IP, ICMP  # type: ignore
+from typing import Any, Callable, List, Tuple
+
+
+class Histogram:
```
stalowyjez, line 17: This looks more like a HistogramFactory tbh, as sth which could be called really a histogram is generated by `get` method, here you have just a list of floats

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,161 @@
...
+    os.unlink(local_path)
+
+    cmd_rm = f"docker exec --privileged {container_name} pkill tcpdump"
+    os.system(cmd_rm)
+
+    return histograms
+
+
+# Render ASCII histogram drawing for visual inspection
+def draw_histogram(name: str, data: List[int], max_height=None):
```
stalowyjez, line 88: Why isn't it a method of `Histogram`?

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def filter(self, packet) -> bool:
+        return self._filter(packet)
+
+    def __str__(self):
+        return f"PacketFilter({self._name})"
+
+
+def _generate_histogram(
+    pcap_path: str, buckets: int, bucket_width: int, filter: PacketFilter
```
stalowyjez, line 127: Hmm, what is the purpose of wrapping this `Callable` with `PacketFilter`?

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,278 @@
...
+            ),
+        )
+
+        futures = await asyncio.gather(*tasks)
+        for name, histograms in zip(container_names, futures):
+            for filter_name, hs in histograms:
+                title = f"{name}-filter({filter_name})-id({request.node.callspec.id})"
+                draw_histogram(title, hs, max_height=12)
+                silence_streak = longest_zero_streak(hs)
+                print(f"Longest silence for {title} is {silence_streak}\n")
```
stalowyjez, line 274: It counts the empty buckets in a row, but buckets may vary in size, is it really ok? Wouldn't it be better to just get the largest difference between consecutive packets during histogram generation?

stalowyjez: Ok, it seems that it is set to 1, but I guess if we depend on that it would be good to make it just hardcoded, or at least make it default or sth, searching for the place where it is decided might be painful

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,278 @@
...
+    Direct,
+    Wireguard,
+)
+from typing import List
+from utils.batching import PacketFilter, capture_histograms, draw_histogram
+from utils.connection import DockerConnection
+from utils.connection_util import DOCKER_GW_MAP, ConnectionTag, container_id
+
+
+def longest_zero_streak(nums):
```
stalowyjez, line 24: This could be moved to the histogram (i.e. "batching") module

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,278 @@
...
+                ),
+                _generate_setup_parameters(
+                    ConnectionTag.DOCKER_UPNP_CLIENT_2,
+                    AdapterType.LinuxNativeWg,
+                    ["stun", "upnp", "local"],
+                    direct_enabled=True,
+                    batching_enabled=False,
+                ),
+            ],
+            id="no_batching",
```
stalowyjez, line 107: Is it used anywhere?

nat-lab/Pipfile.lock
LukasPukenis: I didn't think of it, just used `pip install scapy` and extracted relevant part into Pipfile.lock

nat-lab/tests/utils/batching.py
LukasPukenis: That's actually a good call. No need for Histogram to be anything else except a free function. I've refactored it out into a such function now.

nat-lab/tests/utils/batching.py
LukasPukenis: This becomes irrelevant since I removed the `Histogram` class entirely and renamed `get` into `generate_histogram -> list[int]`

nat-lab/tests/utils/batching.py
LukasPukenis: No big reason except `PacketFilter` also has a `_name` property to it which I find nicer to use instead of passing a tuple of `(str, Callable)`

nat-lab/tests/test_batching.py
LukasPukenis: That's a very valid point. To be honest this metric was not really that much useful so instead of fixing it I just removed it entirely. Your comment on finding largest consecutive difference was good though.

nat-lab/tests/test_batching.py
LukasPukenis: Without the `id` field, the tests look as such:
```
      <Coroutine test_batching_star[CONE_CLIENT_1-LinuxNativeWg-stun-upnp-local-CONE_CLIENT_2-LinuxNativeWg-stun-upnp-local-UPNP_CLIENT_1-LinuxNativeWg-stun-upnp-local-UPNP_CLIENT_2-LinuxNativeWg-stun-upnp-local]>
      <Coroutine test_batching_star[CONE_CLIENT_1-LinuxNativeWg-stun-upnp-local-batch-35-CONE_CLIENT_2-LinuxNativeWg-stun-upnp-local-batch-35-UPNP_CLIENT_1-LinuxNativeWg-stun-upnp-local-batch-35-UPNP_CLIENT_2-LinuxNativeWg-stun-upnp-local-batch-35]>
```

Batching tests should come basically down to one big test that batches, another one which doesn't so I felt it's nice to separate them more clearly

LukasPukenis: These are also incomplete names, as it's only a very small network yet

nat-lab/tests/utils/batching.py
LukasPukenis: I though its a pretty nice disclaimer inside of `batching.py` however making promises there about how we use it is pretty wrong since it's disconnected and the file contains general purpose code. As a middle ground I moved the comment inside of the testcase itself and mentioned that both are used.

nat-lab/tests/utils/batching.py
LukasPukenis: So I actually just replaced it with `Callable` and that went _fine_. It definitely fixed a funny looking part of `filter.filter(pkt)`.

nat-lab/tests/test_batching.py
LukasPukenis: Additionally, I've encountered windows failing with pytest due to too long filenames when saving the logs already in the past

nat-lab/tests/test_batching.py
LukasPukenis: I removed it entirely for now. The goal of this PR is to push something through the door and then augment with necessary metrics

nat-lab/tests/utils/batching.py
LukasPukenis: I refactored code a bit around so now `histogram` is only mentioned when it's generated from data. I think it's leaner that way and became easier to follow

nat-lab/tests/test_batching.py
LukasPukenis: An alternative might be to save the drawings alongside artifacts but I think ascii provides a smooth way of doing the same at the cost of `-v` flag on `run_local.py`

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,207 @@
...
+        # TODO: assert these are docker connections
+        print("Will capture batching on containers: ", container_names)
+        cnodes = zip(env.clients, env.nodes, [c.connection for c in env.connections])
+
+        # the plan is to misalign the peers for better batcing observation. WIthout misalignment
+        # peers will form a meshnet almost immediately, already achieving the desired batching effect.
+        for client in env.clients:
+            await client.stop_device()
+
+        tasks = []
```
stalowyjez, line 185: Some name indicating what these tasks are doing would be nice imho

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,207 @@
...
+                    capture_duration,
+                )
+            )
+            tasks.append(task)
+
+        api = env.api
+
+        # misalign the peers by sleeping some before starting each node again
+        for client, node, _ in cnodes:
+            await asyncio.sleep(6)
```
stalowyjez, line 199: Uhh, it waits 24 seconds by design? Maybe it could be a bit less? Isn't 1 second difference enough?

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,207 @@
...
+            for client, peer_node in product(clients, peer_nodes)
+            if not client.is_node(peer_node)
+        ))
+
+        pcap_paths = await asyncio.gather(*tasks)
+
+        # pylint: disable=consider-using-set-comprehension
+        pyro5_ports = list(set([client.get_proxy_port() for client in env.clients]))
+        assert len(pyro5_ports) == 1
+        pyro5_port = pyro5_ports[0]
```
stalowyjez, line 216: Maybe you can just take this [0] immediately, I think the error will be quite obvious if there is none (or is it a check for multiple ports?)

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,248 @@
...
+                    f"{container}-filter({filter_name})-id({request.node.callspec.id})"
+                )
+                draw_histogram(title, hs, max_height=12)
+
+        for p in pcap_paths:
+            os.unlink(p)
+
+        assert (
+            False
+        ), "TODO: implement a proper assert(this assert also forced the dump of output)"
```
stalowyjez, line 248: So, as I understand this test will always fail on this line - I guess it would break CI when merged, soo, what is the plan to do with it? Mark it as flaky or sth like that?

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,126 @@
...
+from scapy.all import PcapReader, Packet  # type: ignore
+from typing import Callable, List, Optional
+
+
+def generate_histogram(data: list[int], buckets: int, bucket_size: int) -> List[int]:
+    if len(data) == 0:
+        return []
+
+    sorted_values = sorted(data)
+    max_val = sorted_values[-1]
```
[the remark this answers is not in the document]
LukasPukenis: It's being used to throw an exception in case data doesn't fit in. Instead of checking each element if it fits, I simply sort and select the biggest one. Granted I could just do `max(data)` though

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,126 @@
...
+import pytest
+import subprocess
+import typing
+from scapy.all import PcapReader, Packet  # type: ignore
+from typing import Callable, List, Optional
+
+
+def generate_histogram(data: list[int], buckets: int, bucket_size: int) -> List[int]:
+    if len(data) == 0:
+        return []
```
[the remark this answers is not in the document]
LukasPukenis: You gave me some ideas, so I made thse changes:
- simplified the code around `sorted_values` and eliminated that unecessary variable
- moved `len()==0` check into the call site. It will still throw ValueError in the `max()` function however the call site will protect against that which is expected since there might be no data captured or the filter was too harsh

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,126 @@
...
+
+    sorted_values = sorted(data)
+    max_val = sorted_values[-1]
+
+    if max_val >= buckets * bucket_size:
+        raise ValueError(
+            f"Histogram doesn't fit the data({max_val}). The max value is {max_val} but the histogram has {buckets} buckets with a width of"
+            f" {bucket_size}, the maximum value it can fit is {buckets * bucket_size}"
+        )
```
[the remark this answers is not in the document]
LukasPukenis: Possible, but I find this approach nicer - first validate, then do the work. I don't see much drawbacks

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,126 @@
...
+import math
+import os
+import pytest
+import subprocess
+import typing
+from scapy.all import PcapReader, Packet  # type: ignore
+from typing import Callable, List, Optional
+
+
+def generate_histogram(data: list[int], buckets: int, bucket_size: int) -> List[int]:
```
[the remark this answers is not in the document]
LukasPukenis: done

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,126 @@
...
+
+    print(f"+{'-' * (len(data))}")
+    print(f"0{' ' * (len(data)-1)}{len(data)}")
+    print(f"^-Histogram of {name}")
+
+
+def generate_histogram_from_pcap(
+    pcap_path: str,
+    buckets: int,
+    bucket_width: int,
```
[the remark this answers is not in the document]
LukasPukenis: I can answer that partially - I do not see any possibility of this value being higher than 1, since we can just simulate, but I can see a usage of `<1` bucket size. That would of course require seconds->milliseconds conversion as well.

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,126 @@
...
+    hs = [0] * buckets
+
+    for v in sorted_values:
+        bucket_index = int(v / bucket_size)
+        hs[bucket_index] += 1
+
+    return hs
+
+
+async def capture_traffic(container_name: str, duration_s: int) -> str:
```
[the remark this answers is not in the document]
LukasPukenis: I think `start_tcpdump()` and `stop_tcpdump()` could be reused as they are essentially the same thing, except they find unique path for saving however:
- those functions work only under an env variable. Filling it on my side would be a bad taste, adding overriding argument as well, basically very error-prone.
- Refactoring of those functions is possible but I do not really want to do that, this PR is already slow enough to push through
- I felt it's much simpler to just maintain the code I need for batching separately and maybe refactor it out once it proves to be at least working how I expect :)

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,207 @@
...
+        capture_duration = 30
+
+        env = await setup_mesh_nodes(exit_stack, setup_params)
+
+        alpha_client, beta_client, gamma_client, delta_client = env.clients
+        alpha, beta, gamma, delta = env.nodes
+
+        clients = [alpha_client, beta_client, gamma_client, delta_client]
+        peer_nodes = [alpha, beta, gamma, delta]
+
```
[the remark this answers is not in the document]
LukasPukenis: no need, nicer to just iterate straight awai through `env.*`. I fixed it already

nat-lab/tests/test_batching.py
LukasPukenis: This is totally hand-crafted. Could maybe be parametrized? Dunno if that makes sense. The idea is that if no delay is made, then packets already will appear perfectly batched locally as all the peers will _most probably_ connect immediately. I want some delay, but it should be something very small like 1second but also not something that might have a common denominator - such as 5. If you think it's worth it I can parametrize it but more or less this value doesn't matter while it's not too big and not too small

nat-lab/tests/test_batching.py
LukasPukenis: I am not yet sure how Pyro5 works and based on my understanding it chooses mapped Docker's ports. My understanding is that it might be different for each client, however from the observation I see that it's always the same.

Making those filters unique based on each `pyro5_ports` requires extra work so I just went the lazy way - assert they're all equal and grab the first one

nat-lab/tests/test_batching.py
LukasPukenis: Good point. So I wanted the quickest way to see it in action, hence I just `print` stuff. Printing works fine and it's visible when the test fails, so I need the failure while we work on it.

If I put it in the nightly, it will always fail and I am not sure if that's a good thing, might be an annoyance, instead I plan immediately afterwards to introduce another job into the `libtelio-build` pipeline which would also run nightly just for this.

With no assert, `pytest` assumes it all went find and doesn't emit the captured stdout and stderr.

An alternative is possible to save those diagrams to files or something like that however this PR already feels like walking through a mud so I opted out for simplicity

nat-lab/tests/test_batching.py
LukasPukenis: I renamed the variable into `pcap_capture_tasks`

nat-lab/tests/test_batching.py
LukasPukenis: I reworked this one and now the assert is gone. It rather renders histograms into textfiles at `logs/`.

nat-lab/tests/utils/batching.py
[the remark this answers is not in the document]
LukasPukenis: Refactored slighly that the call site doesn't pass the value of 1 but the function internally has a default of such, as a compromise

nat-lab/tests/test_batching.py
LukasPukenis: I parametrized the value so now it's obvious that it's a test parameter

nat-lab/tests/test_batching.py
```diff
@@ -0,0 +1,200 @@
...
+    setup_params: List[SetupParameters], misalign_sleep_amount: int
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        capture_duration = 100
+
+        env = await setup_mesh_nodes(exit_stack, setup_params)
+
+        clients = env.clients
+        peer_nodes = env.nodes
+
```
stalowyjez, line 110: Nit: you still copy them, why? ^^'

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,137 @@
...
+import os
+import pytest
+import subprocess
+import typing
+from scapy.all import PcapReader, Packet  # type: ignore
+from typing import Callable, List, Optional
+
+
+def _generate_histogram(
+    data: list[int], buckets: int, bucket_size: int = 1
```
stalowyjez, line 12: Maybe float is more relevant if you consider batch sizes smaller than 1?

nat-lab/tests/utils/batching.py
```diff
@@ -0,0 +1,137 @@
...
+) -> List[int]:
+    max_val = max(data)
```
stalowyjez, line 14: Maaybe some comment that you treat this `max` also for guarding against `len(data) == 0` would be nice, so someone  won't add `default=0` to it as a fix in the future

stalowyjez (commented):
It looks quite ok to me, just a few minor comments

nat-lab/tests/utils/batching.py
LukasPukenis: Having sub 1second would require floats but that would also require more work to convert seconds to milliseconds so it would work, thus I think it's unjustified at the moment, basically don't have it because we don't need it yet.

nat-lab/tests/test_batching.py
LukasPukenis: No reason, leftovers, removed them :)

nat-lab/tests/utils/batching.py
LukasPukenis: Not sure if I understood correctly but `len(data) == 0` now happens at the call site. For all this function knows about - it expects non-zero data list. Do you mean maybe I should assert for `assert len(data) > 0`?

stalowyjez: Assert would be ok

LukasPukenis: Surely, added it

stalowyjez (approved):
+1

nat-lab/tests/test_batching.py
mathiaspeters: The ports are not all the same when running in CI, but will be the same on the container side (but not host side) when running locally

nat-lab/tests/test_batching.py
mathiaspeters: I think the `report.html` still has the output even for passed tests. Printing to files is also more than fine

mathiaspeters (approved):

PR #808, opened 2024-09-03
LLT-5586: Conntracker TCP states

tomasz-grz:
### Problem
There are tests (like `test_mesh_firewall.py`) where we are manually starting `conntrack` in order to observe TCP state changes. Occasionally they fail because they might be missing the new connetion events, and `conntrack` does not report on subsequent states in that case. We need to make sure that `conntrack` is ready to process events before proceeding with the test.

### Solution

- Added mechanism to the existing `ConnectionTracker` class to track TCP states alongside connection limits, in order to utilize it's synchronization mechanism.
- Updated tests to use `ConnectionTracker` instead of manually starting `conntrack` process.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_mesh_firewall.py
+++ b/nat-lab/tests/test_mesh_firewall.py
@@ -10,7 +10,13 @@ from helpers import SetupParameters, setup_mesh_nodes, setup_api
 from mesh_api import Node
 from typing import Tuple
 from utils import testing, stun
-from utils.connection_tracker import ConnectionLimits
+from utils.connection_tracker import (
+    ConnectionLimits,
+    ConnectionTrackerConfig,
+    FiveTuple,
+    TcpState,
+    ConnectionTracker,
+)
 from utils.connection_util import generate_connection_tracker_config, ConnectionTag
 from utils.output_notifier import OutputNotifier
 from utils.ping import ping
@@ -579,12 +585,6 @@ async def test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_server_s
                 print(datetime.now(), f"nc output: {line}")
             await output_notifier.handle_output(stdout)
 
-        async def conntrack_on_stdout(stdout: str) -> None:
-            for line in stdout.splitlines():
-                if f"src={CLIENT_BETA_IP} dst={CLIENT_ALPHA_IP}" in line:
-                    print(datetime.now(), f"Conntrack event: {line}")
-                    await output_notifier.handle_output(line)
-
         listening_start_event = asyncio.Event()
         sender_start_event = asyncio.Event()
         connected_event = asyncio.Event()
@@ -605,16 +605,20 @@ async def test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_server_s
             connected_event,
         )
 
-        output_notifier.notify_output("LAST_ACK", last_ack_event)
-        output_notifier.notify_output("TIME_WAIT", time_wait_event)
+        async with ConnectionTracker(
+            connection_beta,
+            [
+                ConnectionTrackerConfig(
+                    "nc",
+                    ConnectionLimits(),
+                    FiveTuple(protocol="tcp", dst_ip=CLIENT_ALPHA_IP, dst_port=PORT),
+                )
+            ],
+            True,
+        ).run() as conntrack:
+            conntrack.notify_on_tcp_state(TcpState.LAST_ACK, last_ack_event)
+            conntrack.notify_on_tcp_state(TcpState.TIME_WAIT, time_wait_event)
 
-        async with connection_beta.create_process([
-            "conntrack",
-            "--family",
-            "ipv4" if CLIENT_PROTO == IPProto.IPv4 else "ipv6",
-            "-E",
-        ]).run(stdout_callback=conntrack_on_stdout) as conntrack_proc:
-            await conntrack_proc.wait_stdin_ready()
             # registering on_stdout callback on both streams, cuz most of the stdout goes to stderr somehow
             async with connection_alpha.create_process([
                 "nc",
@@ -640,7 +644,7 @@ async def test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_server_s
                 await connected_event.wait()
 
             # kill server and check what is happening in conntrack events
-            # if everything is correct -> conntrack should show FIN_WAIT -> CLOSE_WAIT
+            # if everything is correct -> conntrack should show LAST_ACK -> TIME_WAIT
             # if something goes wrong, it will be stuck at LAST_ACK state
             await last_ack_event.wait()
             await time_wait_event.wait()
@@ -714,11 +718,6 @@ async def test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_client_s
 
         output_notifier = OutputNotifier()
 
-        async def conntrack_on_stdout(stdout: str) -> None:
-            for line in stdout.splitlines():
-                if f"src={CLIENT_ALPHA_IP} dst={CLIENT_BETA_IP}" in line:
-                    await output_notifier.handle_output(line)
-
         listening_start_event = asyncio.Event()
         sender_start_event = asyncio.Event()
         connected_event = asyncio.Event()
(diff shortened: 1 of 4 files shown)


Jauler (dismissed):
+1.0

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -174,17 +268,17 @@ async def get_out_of_limits(self) -> Optional[Dict[str, int]]:
...
 
         # wait to synchronize over a known event
         async with Ping(self._connection, "127.0.0.2").run():
             print(datetime.now(), "ConnectionTracker waiting for _sync_event")
             try:
-                await asyncio.wait_for(self._sync_event.wait(), timeout=10.0)
+                await asyncio.wait_for(self._sync_event.wait(), timeout=5.0)
                 print(datetime.now(), "ConnectionTracker got _sync_event")
-            except TimeoutError:
+            except (TimeoutError, asyncio.TimeoutError):
```
jjanowsk, line 281: asyncio.TimeoutError is just an alias for standard TimeoutError

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -174,17 +268,17 @@ async def get_out_of_limits(self) -> Optional[Dict[str, int]]:
...
+    async def _synchronize(self):
         if not self._config:
             return None
 
         # wait to synchronize over a known event
         async with Ping(self._connection, "127.0.0.2").run():
             print(datetime.now(), "ConnectionTracker waiting for _sync_event")
             try:
-                await asyncio.wait_for(self._sync_event.wait(), timeout=10.0)
+                await asyncio.wait_for(self._sync_event.wait(), timeout=5.0)
```
jjanowsk, line 279: Leftover from some debugging?

.unreleased/LLT-5586_conntracker_TCP_states
```diff
@@ -0,0 +1 @@
+Add mechanism to the existing `ConnectionTracker` class to track TCP states.
```
jjanowsk, line 1: Noone outside libtelio care about that. Should be empty file IMO

nat-lab/tests/test_mesh_firewall.py
```diff
@@ -781,5 +774,5 @@ async def conntrack_on_stdout(stdout: str) -> None:
                 # kill client and check what is happening in conntrack events
                 # if everything is correct -> conntrack should show LAST_ACK -> TIME_WAIT
                 # if something goes wrong, it will be stuck at LAST_ACK state
-                await last_ack_event.wait()
-                await time_wait_event.wait()
+                assert await conntrack.event_with_tcp_state_exists(TcpState.LAST_ACK)
+                assert await conntrack.get_last_event_tcp_state() == TcpState.TIME_WAIT
```
jjanowsk, line 778: You are introducing a race here. Previously we had a wait here for the given event. Now you just check the list of currently received events. It's possible that those LAST_ACK event has not been yet registered by conntrack. Please add a wait for it otherwise we'll see a flakiness from it

nat-lab/tests/test_vpn.py
```diff
@@ -472,11 +474,10 @@ async def conntrack_on_stdout(stdout: str) -> None:
...
             )
 
-            # if everything is correct -> conntrack should show FIN_WAIT -> CLOSE_WAIT
-            # or our connection killing mechanism will reset connection resulting in CLOSE output.
+            # under normal circumstances -> conntrack should show FIN_WAIT -> CLOSE_WAIT
+            # But our connection killing mechanism will reset connection resulting in CLOSE output.
             # Wait for close on both clients
-            await close_wait_event.wait()
-            await close_wait_event.wait()
+            assert await conntrack.get_count_event_with_state(TcpState.CLOSE) == 2
```
jjanowsk, line 480: Same here. We need to wait for them

jjanowsk (changes requested):

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -44,21 +84,29 @@ def get_key(self) -> str:
         return self.key
 
 
-def parse_input(input_string) -> FiveTuple:
+def parse_input(input_string) -> ConntrackerEvent:
+    event = ConntrackerEvent()
     five_tuple = FiveTuple()
```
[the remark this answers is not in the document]
tomasz-grz: Good point! 💯

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -28,6 +29,45 @@ def partial_eq(self, comparison: "FiveTuple") -> bool:
         )
 
 
+class EventType(str, Enum):
```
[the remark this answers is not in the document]
tomasz-grz: Indeed, not needed

nat-lab/tests/test_mesh_firewall.py
tomasz-grz: Hmm you're right, I was afraid this might be the case.. Internally get events call waits for the synchronize event, since takes a bit of time everything appears to be arriving in the correct order, according to logs. Although theoretically there is no guarantee that sync event will arrive after LAST_ACK event, I'll try to think of a more elegant way to implement this.

nat-lab/tests/utils/connection_tracker.py
jjanowsk: Yeah, it seems so. From 3.11 docs: 
```
exception asyncio.TimeoutError
A deprecated alias of [TimeoutError](https://docs.python.org/3.11/library/exceptions.html#TimeoutError), 
raised when the operation has exceeded the given deadline.

Changed in version 3.11: This class was made an alias of 
[TimeoutError](https://docs.python.org/3.11/library/exceptions.html#TimeoutError).
```

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -199,5 +293,5 @@ async def run(self) -> AsyncIterator["ConnectionTracker"]:
             await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
tomasz-grz: Unfortunately conntrack takes a bit of time to start, without the sleep the sync event might arrive before stdout_callback is ready to process it.. An alternative I guess would be to ping it in a loop.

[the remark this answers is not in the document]
jjanowsk: I checked what is the underlying problem. It's not that ping is not sending pings continuously, but our conntrack wrapper reacts only on [NEW] events. Because all pings sent from the same command execution have the same id (different seq number but the same id) they are treated as one connection. Because of that it's possible that we do not catch this first ping and hance we are missing NEW event. Your solution now should work but I think it can be simplified to something like this (not tested. Written in the browser ;)):
```
while not self._sync_event.is_set():
    await ping(self._connection, "127.0.0.2")
```

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -128,21 +175,67 @@ async def execute(self) -> None:
...
+            return []
+
+        await self._synchronize()
+
+        async with self._lock:
+            return self._events
+
+    async def wait_for_event_with_tcp_state(
+        self, state: TcpState, timeout: float | None = None
+    ):
```
jjanowsk, line 192: 2 things that I see that could be improved:
1. Easy: get rid of timeout parameter. We don't need to handle timeout ourselves this way. If someone needs a timeout on an async method they can always use `asyncio.wait_for` to wrap it.
2. More work: I think this would be nicer if it had worked as our `OutputNotifier`. Meaning you would have a method `notify_on_tcp_state(TcpState)` and later you would wait on an event which is going to be set when this tcp state is reported. This way you don't need to create this busy loop.

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -128,21 +175,67 @@ async def execute(self) -> None:
...
+        """Returns a list of the reported events that matched the ConnectionTrackerConfig"""
+        if platform.system() == "Darwin":
+            return []
+        if not self._config:
+            return []
+
+        await self._synchronize()
+
+        async with self._lock:
+            return self._events
```
jjanowsk, line 188: You are returning a reference to a mutable type which you protect with a lock. Later in count_events_with_tcp_state you access it without locking.

jjanowsk (changes requested):

nat-lab/tests/utils/connection_tracker.py
tomasz-grz: Hmm good point about the timeout.
Thanks, I'll take a look into the `OutputNotifier`.

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -112,37 +156,54 @@ async def on_stdout(self, stdout: str) -> None:
...
+            return None
 
         await self._process.execute(stdout_callback=self.on_stdout)
 
+    def notify_on_tcp_state(self, state: TcpState, event: asyncio.Event) -> None:
+        """Register an Event to be notified when a specific TCP state is reported"""
+        if state not in self._tcp_state_events:
+            self._tcp_state_events[state] = [event]
+        else:
+            self._tcp_state_events[state].append(event)
```
jjanowsk, line 188: Nit: This is quite common pattern that you can solve by using defaultdict.
```
self._tcp_state_events = defaultdict(list) #instead of {}
... later you can just do this:
self._tcp_state_events[state].append(event)
without worying about creating the key
```

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -174,30 +235,41 @@ async def get_out_of_limits(self) -> Optional[Dict[str, int]]:
...
-    async def synchronize(self):
+    def _check_tcp_state_for_events(self, conntracker_event: ConntrackerEvent) -> None:
+        if tcp_state := conntracker_event.tcp_state:
+            if notify_events := self._tcp_state_events.get(tcp_state):
+                for event in notify_events:
+                    # set only the first unset event
+                    if not event.is_set():
+                        print(datetime.now(), "ConnectionTracker setting event", event)
+                        event.set()
+                        break
```
jjanowsk, line 246: Coninuation of the defaultdict nit. Another nice benefit is that you can simplify this to:
```
if tcp_state := conntracker_event.tcp_state:
    try:
        self._tcp_state_events[tcp_state].pop(0).set()
    except IndexError:
        pass
```

jjanowsk (dismissed):
+1 Looks good to me. I checked and actually _lock can be removed from connection_tracker so if you want you can do this with this PR but I understand it you don't :D

jjanowsk (approved):

PR #812, opened 2024-09-09
LLT-5552: Create teliod daemon

stalowyjez:
### Problem
We need some skeleton for a new Telio daemon to be run in OpenWRT

### Solution
Add a skeleton of a new Teliod daemon

- [x] README.md is updated


clis/daemon-common/src/daemon.rs
```diff
@@ -0,0 +1,122 @@
+//! Code for general daemon handling utilities and abstracting away related dependencies.
+//!
+
+cfg_if::cfg_if! {
+    if #[cfg(unix)] {
```
lcruz99, line 5: Instead of disabling the code we could configure the manifest file to only support this target

lcruz99: this also for comms.rs file

stalowyjez: That would be great, unfortunately I have no idea how to do that, this is the simplest thing I found :/
Do you have some resources on this topic?

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+                }
+                cmd => {
+                    if TeliodDaemon::is_running()? {
+                        let response = DaemonSocket::send_command(
+                            &TeliodDaemon::get_ipc_socket_path()?,
+                            &serde_json::to_string(&cmd)?,
+                        )?;
+
+                        if response.as_str() == "OK" {
+                            println!("Command exectued successfully!");
```
lcruz99, line 84: exectued -> executed

lcruz99: exclamation mark should be removed to be coherent with other logs

clis/tcli/Cargo.toml
```diff
@@ -58,9 +52,7 @@ telio-utils.workspace = true
 telio-wg.workspace = true
 
 [target.'cfg(unix)'.dependencies]
-# Used to avoid the manual work of daemonizing the libtelio process.
-# Also is the only reason why Windows isn't supported.
-daemonize = { version = "0.5.0", optional = true }
+daemon-common = { version = "0.1.0", path = "../daemon-common", optional = true }
 
 [[bin]]
 name = "tclid"
```
lcruz99, line 58: this should be added to the docs. ie: building tclid requires this feature:
```
cargo build -p tcli --bin tclid --features="tclid"
```

clis/tcli/src/bin/tclid/main.rs
```diff
@@ -8,27 +8,26 @@ use std::path::PathBuf;
```
lcruz99, line 8: ^^ unused import

clis/tcli/src/bin/tclid/main.rs
```diff
@@ -8,27 +8,26 @@ use std::path::PathBuf;
...
 use clap::Parser;
+use daemon_common::daemon::{Daemon, ProcessType};
 use tokio::time::Duration;
 
-mod coms;
-mod daemon;
 mod tclid;
 mod telio_cli_interface;
 
 use coms::DaemonSocket;
```
lcruz99, line 17: daemon-common::coms

lcruz99: get_ipc_socket_path should be imported as well

clis/daemon-common/src/coms.rs
```diff
@@ -0,0 +1,143 @@
+//! Code for handling inter process communication and abstracting away it's dependencies.
```
lcruz99, line 1: file name "coms.rs" is ambiguous, if it's for communications it should be "comms.rs"/"comm.rs"

lcruz99:
I'm wondering: shouldn't this be compilable? at least linters should have been run?

clis/daemon-common/src/daemon.rs
stalowyjez: @lcruz99 It doesn't work for me too well, maybe I'm doing it wrong somehow, but it seems that it still tries to compile the dependencies and fails on the last step, which is not that much better. Also, this PR is not really about TCLI, so I guess it would be better to make another issue for improving this. The `daemon-common` crate doesn't need this conditions if we exclude it in windows testing (as it is unix-only crate). And for `teliod` maybe the cleanest way would be making separate files with OS-specific implementation, what do you think?

ThrasherLT:
+1

clis/daemon-common/src/daemon.rs
stalowyjez: Oh, in the end I just removed that conditions, as the code is cleaner and CI works with that

clis/daemon-common/src/daemon.rs
```diff
@@ -69,33 +72,34 @@ impl Daemon {
...
-            }
+    fn start() -> Result<ProcessType> {
+        use std::fs::File;
+        let wd_path = get_wd_path(Self::NAME)?;
+
+        std::fs::create_dir_all(&wd_path)?;
+        let stderr = File::create(wd_path.join(format!("{}.err", Self::NAME)))?;
+
+        let daemonize = Daemonize::new()
+            .pid_file(format!("{}.pid", Self::NAME)) // Every method except `new` and `start`
```
lcruz99, line 83: please make block comments vertical

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+    D: Deserializer<'de>,
+{
+    Deserialize::deserialize(deserializer).and_then(|s: String| {
+        LevelFilter::from_str(&s).map_err(|_| {
+            de::Error::unknown_variant(&s, &["error", "warn", "info", "debug", "trace", "off"])
+        })
+    })
+}
+
+/// Simple telio daemon for Linux and OpenWRT
```
lcruz99, line 45: looks like this comment is out of place?

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+            Ok(())
+        }
+        Ok(ProcessType::Daemon) => daemon_event_loop(config),
+        Err(e) => Err(e),
+    }
+}
+
+fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<()> {
+    // Create logger
+    let (non_blocking_writer, _tracing_worker_guard) =
```
lcruz99, line 128: shouldn't the logger be initialized right at the beginning of main()?

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+
+            Ok(())
+        }
+        Ok(ProcessType::Daemon) => daemon_event_loop(config),
+        Err(e) => Err(e),
+    }
+}
+
+fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<()> {
+    // Create logger
```
lcruz99, line 127: trivial comment

clis/teliod/src/main.rs
lcruz99: Nvm, I've missed the logic behind this.. Perhaps the API process can have its own logger instance?

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Starts the Teliod daemon which will run actual Telio.
+///
+/// # Arguments
+///
+/// * `config` - Daemon config.
+fn start_daemon(config: TeliodDaemonConfig) -> Result<()> {
+    println!("Starting Teliod daemon...");
+    match TeliodDaemon::start() {
+        Ok(ProcessType::Api) => {
+            // Verifying that the server had been started, but we have no way of synchronizing
```
lcruz99, line 101: I think this solution using time arrays can be improved. We can use a one shot channel or even a bool flag for synchronization.. otherwise a exponential backoff mechanism can also be used.

stalowyjez: Something like this maybe?

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+            } else {
+                eprintln!("Teliod daemon is not running");
+            }
+        }
+    };
+
+    Ok(())
+}
+
+const DAEMON_START_MSG: &'static str = "ready";
```
lcruz99, line 94: consts should be defined at the beginning of the file (after imports)

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+///
+/// * `config` - Daemon config.
+fn start_daemon(config: TeliodDaemonConfig) -> Result<()> {
+    println!("Starting Teliod daemon...");
+
+    let (api_fd, daemon_fd) = pipe()?;
+
+    match TeliodDaemon::start() {
+        Ok(ProcessType::Api) => {
+            let _ = close(daemon_fd.as_raw_fd());
```
lcruz99, line 108: value returned by `close()` will be dropped when it goes out of scope, so you should consume `daemon_fd` with             `close(daemon_fd.into_raw_fd());` or call drop directly as done below

clis/teliod/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+For OpenWRT you might need a bit more complex command, including your router architecture and the fact the OpenWRT is MUSLE-based, for example:
+
+```CARGO_TARGET_ARMV7_UNKNOWN_LINUX_MUSLEABIHF_LINKER=rust-lld cargo build --target armv7-unknown-linux-musleabihf```
+
+### Using Teliod daemon
+
+TODO: Update it when Teliod will be ready
+
+Teliod runs Telio library in the background and provied a simple CLI tool to manage it.
```
Jauler, line 15:
```suggestion
Teliod runs Telio library in the background and provides a simple CLI tool to manage it.
```

clis/daemon-common/src/daemon.rs
```diff
@@ -2,10 +2,10 @@
...
 
-use anyhow::{self, Result};
+use anyhow::Result;
 use sysinfo::{ProcessRefreshKind, RefreshKind, System};
 
-use crate::TCLID_WD_PATH;
+use daemonize::Daemonize;
 
 /// Enum for figuring out the current process after the daemon is forked away
 /// from the API by the daemonization.
```
Jauler, line 11: I wonder, does it really make sense to use the same process (binary) for daemon and for client? :thinking:

clis/daemon-common/src/daemon.rs
```diff
@@ -69,33 +71,36 @@ impl Daemon {
...
+    fn start() -> Result<ProcessType> {
+        use std::fs::File;
+        let wd_path = get_wd_path(Self::NAME)?;
+
+        std::fs::create_dir_all(&wd_path)?;
+        let stderr = File::create(wd_path.join(format!("{}.err", Self::NAME)))?;
+
+        // Every method except `new` and `start` is optional, see `Daemonize` documentation for
+        // default behavior.
+        let daemonize = Daemonize::new()
```
Jauler, line 83: Turns out, this daemonization should be conditional (e.g. on config or cli argument) because according to OpenWRT docs, their [init system](https://openwrt.org/docs/guide-developer/procd-init-scripts)  expects daemons to be running in foreground and the procd itself takes care of daemonization.

This means that we need ability to start a daemon in "daemon" mode without daemonization (forking).

clis/daemon-common/src/daemon.rs
```diff
@@ -19,21 +19,23 @@ pub enum ProcessType {
...
 
-impl Daemon {
     /// Checks whether the daemon is currently running.
     /// Won't work if pid file had been tampered with or it's location changed.
     ///
     /// # Returns
     ///
     /// Result containing a bool. True means that the daemon is running.
-    pub fn is_running() -> Result<bool> {
+    fn is_running() -> Result<bool> {
```
Jauler, line 32: I have not yet double-checked this, but I wonder, maybe it is possible to check `.is_running` by the existance or non-existance of unix socket?

Note that in OpenWRT, while we can have `teliod start` and `teliod stop` commands, we also need ability to start teliod process in  daemon mode directly, and then the client should fail if unix socket does not exist instead of spawning new daemon. This is because, `procd` is responsible for process management in OpenWRT, and I guess it would be nicer to cooperate with it, instead of spawning "unmanaged" processes.

clis/daemon-common/src/daemon.rs
```diff
@@ -104,11 +109,11 @@ impl Daemon {
 /// # Returns
 ///
 /// Result containing the path of the daemons work directory.
-fn get_wd_path() -> Result<PathBuf> {
+fn get_wd_path(daemon_name: &str) -> Result<PathBuf> {
```
Jauler, line 112: This looks strange to me as per generic `teliod`, we are getting current path to "client" binary, and later expecting to get something with the `*.sock` as unix socket. And later adding `/var/run/xxx` here? Or am I missing something?

clis/daemon-common/src/daemon.rs
stalowyjez: Okk, I guess we can move all of this daemonization stuff under some feature? I guess we might need it for different OSes? Or we won't, so maybe I should remove it in the end? ^^'

clis/daemon-common/src/daemon.rs
stalowyjez: Yeah, it does look strange to me too, it was some idea to make a private helper function for a public interface.

clis/daemon-common/src/daemon.rs
Jauler: IMHO, we can leave the start/stop commands, for standard `tclid` usage, like it was before. But add something (e.g. probably a command or argument, like `teliod --daemon` or similar, whatever makes the most sense to you) that allows to start daemon and skip the forking part with the assumption, that forking is done by procd/systemd/etc.

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+            match command {
+                Cmd::Start { .. } => {
+                    warn!("Start command should never be passed to the daemon");
+                    connection.respond("ERR".to_owned())?;
+                }
+                Cmd::HelloWorld { name } => {
+                    info!("Hello {}!", name);
+                    connection.respond("OK".to_owned())?;
+                }
+                Cmd::Stop => {
```
Jauler, line 182: IMHO, it would be nice for the daemon to react to `SIGTERM` instead of specific command, as this is standard method for signaling for process, that it needs to shutdown by standard managers like dumb-init/procd/systemd etc.

stalowyjez: Hmm, maybe both?

clis/daemon-common/src/daemon.rs
stalowyjez: Hmm, but `tclid` or `teliod`?

clis/daemon-common/src/daemon.rs
Jauler: Maybe lets us `/run` :thinking: as per FHS standard https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html and add the fallback to `/var/run` just in case (e.g. if `/run` does not exist)

> System programs that maintain transient UNIX-domain sockets must place them in this directory or an appropriate subdirectory as outlined above.

This should also be compatible with MacOS as well :crossed_fingers:.

clis/daemon-common/src/daemon.rs
stalowyjez: Hmm, is checking it by the pid worse than checking it by the socket? Sure, I will add some option to spawn it directly.

clis/daemon-common/src/daemon.rs
Jauler: I mean, that `tclid` would have commands to start stop daemon, and `teliod` does not really need them.

clis/teliod/src/main.rs
Jauler: for `teliod` specifically it would be rather strange to have stop. As under normal circumstances - `teliod stop` would cause the process to stop, which later would be treated as daemon crash by process manager (dumb-init/systemd/procd) which, under normal circumstances, would restart it. Thus `teliod stop` command would be more like `teliod restart` :D which I guess could be a bit confusing.

clis/daemon-common/src/daemon.rs
Jauler: I was thinking about, because generally we would have two kind of indicators about existance of daemon (pid file and unix socket) therefore I thought this may be a bit redundant, thats all. But fundamentally - no problems, maybe just that we should store the pid file somewhere in standartized location, e.g. like `/var/run` or `/run` or similar.

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+                }
+            },
+            signal = signals.next() => {
+                match signal {
+                    Some(SIGTERM) => {
+                        info!("Recieved SIGTERM signal, exiting");
+                        break None;
+                    }
+                    Some(_) => {
+                        info!("Recieved unexpected signal, ignoring");
```
lcruz99, line 221: nit: "Received"

lcruz99: and above the same

clis/teliod/src/daemon.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// Daemon name, used for naming daemon files
+    const NAME: &'static str;
+
+    /// Helper function for getting the path to the daemon's working directory.
+    /// The directory is used for storing *.pid and *.socket files.
+    ///
+    /// # Returns
+    ///
+    /// Result containing the path of the daemons work directory.
+    fn get_wd_path() -> Result<PathBuf>;
```
Jauler, line 21: nitpick: I guess this function name is a bit confusing now, after refactoring. As it is just returning `/run/` or `/var/run`, and does not really have anything to od with working directory ;). Maybe lets rename it, for example: `get_runtime_data_directory()` or similar.

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+struct TeliodDaemon;
+
+impl Daemon for TeliodDaemon {
+    const NAME: &'static str = "teliod";
+
+    fn get_wd_path() -> Result<PathBuf> {
+        if Path::new("/run").exists() {
+            Ok(PathBuf::from("/run/"))
+        } else {
+            Ok(PathBuf::from("/var/run/"))
```
Jauler, line 36: I guess we should check that `/var/run` exists as well, and return error if nor `/run` nor `/var/run` exists.

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+    )]
+    HelloWorld { name: String },
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    match Cmd::parse() {
+        Cmd::Daemon { config_path } => {
+            if TeliodDaemon::is_running()? {
+                eprintln!("Teliod is already running");
```
Jauler, line 87: Note that in case daemon exits with error condition -> we should make sure that its exit code is non-zero. I guess we can use `anyhow::bail` or `return Err(...)` which nicely "represents" error condition.

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+                    }
+                    Err(err_str) => {
+                        error!(err_str);
+                        break Some(err_str);
+                    }
+                }
+            },
+            signal = signals.next() => {
+                match signal {
+                    Some(SIGTERM) => {
```
Jauler, line 216: Do not forget to cleanup pidfile and unix socket here.

clis/teliod/src/comms.rs
```diff
@@ -0,0 +1,120 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - Address (more specifically path, but left address for clarity) of the IPC socket.
+    /// * `cmd` - Command to be sent to the daemon.
+    ///
+    /// # Returns
+    ///
+    /// A Result containing a response string.
+    pub async fn send_command(addr: &Path, cmd: &str) -> Result<String> {
```
Jauler, line 72: To some extent we depend on `cmd` not having any newlines in itself. Maybe it is worth veryfing it and enforcing this, e.g. by returning an error if there is a newline inside of cmd?

clis/teliod/src/daemon.rs
```diff
@@ -0,0 +1,66 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Result containing a bool. True means that the daemon is running.
+    fn is_running() -> Result<bool> {
+        let system = System::new_with_specifics(
+            RefreshKind::new().with_processes(ProcessRefreshKind::everything()),
+        );
+
+        match std::fs::read_to_string(Self::get_pid_path()?) {
```
Jauler, line 34: I guess this part is inherently racy, e.g. if `teliod` crashes or is killed abruptly, pid file will be left there, and then there is always a chance that some other process will take previous pid. While I agree that this part is quite an edge-case, but maybe there is some usefulness to rely on unix socket for that. E.g. if new daemon process is being started and unix socket exists -> issue an error. In case of a client process - failure to open unix socket could indicate that daemon is not running.

This of course means, that we do need to handle unix socket cleanup.

clis/teliod/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+    info!("Entering event loop");
+
+    let err_str = loop {
+        select! {
+            maybe_cmd = cmd_listener.get_command() => {
+                match maybe_cmd {
+                    Ok(EventLoopCommand::HelloWorld(name)) => {
+                        info!("Hello {}", name);
+                    }
+                    Err(err_str) => {
```
Jauler, line 208: I do not think we should exit the daemon if client sends invalid command. IMHO, lets just log the error, and respond to a client with some kind of error response.

clis/teliod/src/main.rs
stalowyjez: Socket is already cleaned by RAII, I'll do the same with with PID file

clis/teliod/src/comms.rs
stalowyjez: But here? Only JSONed commands are sent here, I don't think they can contain any newlines. But I can add some check, sure ;)

clis/teliod/src/daemon.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Struct for packaging together the functions for managing the daemon.
+/// Currently does not need any members, since most of the data is stored
+/// in files anyway.
+pub struct TeliodDaemon {
+    pid_path: PathBuf,
+}
+
+impl TeliodDaemon {
+    /// Initializes all of the needed daemon stuff, currently just creates a PID file
+    pub fn new() -> Result<Self> {
```
Jauler, line 22: I think pid-file is not used anywhere anymore

clis/teliod/README.md
```diff
@@ -0,0 +1,28 @@
+### Building Teliod
+
+For typical Linux environment it might be build using simply:
```
Jauler, line 3:
```suggestion
For typical Linux environment it might be built using simply:
```

Jauler (approved):
+1.0, just I guess pid-file creation and removal is not needed anymore, so we can remove it.

clis/teliod/src/daemon.rs
stalowyjez: I thought the deamon just should have some pidfile, is used in the test for killing and maybe for such purposes but I can probably use `ps aux` instead. Idk, I can remove it.

ThrasherLT (approved):
Reviewed again after changes, everything seems sorted out, so +1.0

Jauler (approved):
+1.0

PR #815, opened 2024-09-10
Minimal implemenation of the notification center mqtt service

tomaszklak:
### Problem
To test both the esp32/esphome and openwrt we need an environment that has a working fake of the notification center.

### Solution
This adds a mqtt broker and a small fake implementation of the notification center. The http endpoints are just as an example used to manually test the fake (those will be implemented for real in LLT-5563).

- [x] README.md is updated


dfetti (commented):
+1.0

nat-lab/bin/core-api.py
```diff
@@ -0,0 +1,211 @@
...
+
+def main():
+    print("Starting core api simulator")
+    mqttc = mqtt.Client(
+        mqtt.CallbackAPIVersion.VERSION2, client_id="sender", protocol=mqtt.MQTTv311
+    )
+    mqttc.on_connect = on_connect
+    mqttc.on_message = on_message
+    mqttc.on_subscribe = on_subscribe
+    mqttc.connect("10.0.80.85", port=1883, keepalive=60)
```
olekoliinyk, line 198: I think it would be nice to add some re-tries to connect (it might be the reason why core-api container failed to create in the pipeline, i.e. it cannot connect to the mqtt broker).

nat-lab/docker-compose.yml
```diff
@@ -582,6 +582,28 @@ services:
     healthcheck:
       test: "ls /ready"
 
+  mqtt-broker:
```
olekoliinyk, line 585: Is it worth to add a health check for the mqtt-broker container?

nat-lab/bin/mqtt_listener.py
```diff
@@ -0,0 +1,18 @@
...
+def on_message(_client, _userdata, message):
+    print(f"{message.payload.decode()}")
+    sys.exit(0)
+
+
+mqttc = mqtt.Client(client_id="reciever", protocol=mqtt.MQTTv311)
+
+mqttc.on_message = on_message
+
+# TODO: add credentials when LLT-5604 will be ready
```
Jauler, line 14: IMHO for credentials, for MVP version, and keeping in mind that this is testing-only environment, we can hardcode something then make sure, the same is returned by API, and verify that those hardcoded values are provided by the clients in here.

nat-lab/tests/test_notification_center.py
```diff
@@ -0,0 +1,120 @@
...
+    async with AsyncExitStack() as exit_stack:
+        # Setup connections
+        connection_alpha = await exit_stack.enter_async_context(
+            new_connection_by_tag(ConnectionTag.DOCKER_CONE_CLIENT_1)
+        )
+        connection_beta = await exit_stack.enter_async_context(
+            new_connection_by_tag(ConnectionTag.DOCKER_CONE_CLIENT_2)
+        )
+
+        mqtt_process = await run_mqtt_listener(exit_stack, connection_alpha)
```
Jauler, line 65: I think you should be able to establish MQTT connection directly from this test instead of running it in separate process. I guess it does not really have to be run in one of the `DOCKER_CONE_CLIENT_X` networks.

nat-lab/bin/core-api.py
Jauler, line 1: I guess as we are adding MQTT and a bit of API in the same PR, it would be really nice to add `notifications/v1/tokens` endpoint to API, so we would have "normal" flow. E.g.:
1. [this is a new step] Issue POST request to `notifications/v1/tokens`, retreive broker address, and authentication info (auth may be hardcoded)
2. [this is existing step, slitghtly modified] create mqtt client and connect to the broker retrieved in step 1
3. [this step already exists] register machine and check if notification is received.

nat-lab/bin/mqtt_listener.py
Jauler: Here is a bit more info about API returning credentials: https://github.com/NordSecurity/libtelio/pull/815/files#r1766352606

nat-lab/bin/core-api.py
Jauler: Oh we agreed to make this as separate PR/task. NVM then.

nat-lab/bin/mqtt_listener.py
Jauler: Oh we agreed on making this a separate PR/task. NVM then.

Jauler (dismissed):
So other then https://github.com/NordSecurity/libtelio/pull/815/files#r1766348529, +1.0

nat-lab/bin/core-api.py
stalowyjez: Adding docker-compose dependency sounds good enough IMO

Jauler (approved):
+1.0

PR #817, opened 2024-09-11
Fix starcast transport component failing to open transport socket when meshnet is started before tunnel interface IP is configured

ThrasherLT:
### Problem
If Telio is run via TCLI and `mesh on <tun_name>` without calling `dev start` and configuring the tunnel interface with meshnet IP, starcast transport socket cannot be bound to meshnet IP.

### Solution
Starcast now retries binding transport socket to the meshnet IP every 5 seconds until it succeeds. Until then multicast messages captured by the starcast vpeer are buffered in the mpsc channel or ignored if it is full.

- [x] README.md is updated

### DIFF
--- a/crates/telio-starcast/src/transport.rs
+++ b/crates/telio-starcast/src/transport.rs
@@ -1,6 +1,7 @@
 use std::{
     net::{IpAddr, SocketAddr},
     sync::Arc,
+    time::Duration,
 };
 
 use async_trait::async_trait;
@@ -15,7 +16,13 @@ use telio_task::{
     io::{wait_for_tx, Chan},
     task_exec, Runtime, RuntimeExt, Task, WaitResponse,
 };
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn};
+use telio_utils::{
+    exponential_backoff::{
+        Backoff, Error as ExponentialBackoffError, ExponentialBackoff, ExponentialBackoffBounds,
+    },
+    PinnedSleep,
+};
+use telio_utils::{telio_log_debug, telio_log_info, telio_log_warn};
 
 use telio_model::constants::{
     IPV4_MULTICAST_NETWORK, IPV4_STARCAST_ADDRESS, IPV6_MULTICAST_NETWORK, IPV6_STARCAST_ADDRESS,
@@ -61,6 +68,12 @@ pub enum Error {
     /// Task execution failed
     #[error(transparent)]
     TaskError(#[from] telio_task::ExecError),
+    /// Transport socket isn't open yet, probably because tunnel interface hadn't been configured yet.
+    #[error("Transport socket isn't open yet")]
+    TransportSocketNotOpen,
+    /// Exponential backoff error
+    #[error(transparent)]
+    ExponentialBackoffError(#[from] ExponentialBackoffError),
 }
 
 /// A peer in the meshnet
@@ -91,7 +104,7 @@ impl Transport {
     /// Starts the transport component
     ///
     /// Parameters:
-    /// * meshnet_ip - the meshnet IP of the node on which this component is currently running
+    /// * meshnet_ip - The meshnet IP of the node on which this component is currently running
     /// * socket_pool - To create the transport socket
     /// * packet_chan - A channel to send packets to and receive packets from the virtual peer component
     pub async fn start(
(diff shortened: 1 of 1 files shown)


mathiaspeters:
I'm fine with this as is, but it might be a good idea to use an exponential backoff instead of constant 5 seconds. If it seems like a small thing to add, let's do it now, otherwise let's create a ticket for it in the multicast epic. Either way, +1

mathiaspeters:
What's the reason the expontential backoff is set as a parameter? Do you see it as something we might want to have control over from the outside?

ThrasherLT:
> What's the reason the expontential backoff is set as a parameter? Do you see it as something we might want to have control over from the outside?

I wrote it according to the other implementations and exposed it as something that MIGHT be useful to configure in the future, but I can change it back, if you think it's not worth exposing.

mathiaspeters:
> I wrote it according to the other implementations and exposed it as something that MIGHT be useful to configure in the future, but I can change it back, if you think it's not worth exposing.

To me it makes more sense to create the `ExpontentialBackoff` inside the transport component, or potentially even in the `State` (whichever makes the most sense to you)

ThrasherLT:
> > I wrote it according to the other implementations and exposed it as something that MIGHT be useful to configure in the future, but I can change it back, if you think it's not worth exposing.
> 
> To me it makes more sense to create the `ExpontentialBackoff` inside the transport component, or potentially even in the `State` (whichever makes the most sense to you)

Removed exponential backoff from the arguments to the Transport component.

mathiaspeters (dismissed):

mathiaspeters:
+1

crates/telio-starcast/src/transport.rs
```diff
@@ -138,13 +167,32 @@ impl Transport {
     }
 }
 
+async fn open_transport_port(
```
stalowyjez, line 170: Why is it `*_port` not `*_socket`?

crates/telio-starcast/src/transport.rs
```diff
@@ -228,6 +281,29 @@ impl Runtime for State {
...
+                    MULTICAST_TRANSPORT_PORT,
+                )
+                .await
+                else {
+                    telio_log_warn!(
+                        "Starcast transport socket still not opened, will retry in {:?}",
+                        self.exponential_backoff.get_backoff()
+                    );
+                    self.exponential_backoff.next_backoff();
+                    return Self::next();
```
stalowyjez, line 300: What does it do, just runs `wait` function another time? If yes, I guess you could simplify it by just either waiting for sockets or waiting on sockets.

crates/telio-starcast/src/transport.rs
```diff
@@ -118,6 +144,9 @@ impl Transport {
                 recv_buffer: vec![0; MAX_PACKET_SIZE],
                 nat: StarcastNat::new(IPV4_STARCAST_ADDRESS, IPV6_STARCAST_ADDRESS),
                 peers: Vec::new(),
+                socket_pool,
+                meshnet_ip,
+                exponential_backoff,
```
stalowyjez, line 149: Is there any way in which we may lose the socket? If not maybe we could drop all of this after getting one? (E.g. Making an enum which either keeps socket or these guys?)

crates/telio-starcast/src/transport.rs
```diff
@@ -228,6 +281,29 @@ impl Runtime for State {
...
     async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
+        let transport_socket = match &self.transport_socket {
+            Some(transport_socket) => transport_socket.to_owned(),
+            None => {
+                PinnedSleep::new(self.exponential_backoff.get_backoff(), ()).await;
+                let Ok(transport_socket) = open_transport_port(
+                    self.socket_pool.clone(),
+                    self.meshnet_ip,
+                    MULTICAST_TRANSPORT_PORT,
+                )
```
stalowyjez, line 292: Do you assign it somehow to `self.transport_socket`?

stalowyjez (commented):
Looks quite good, few minor comments

crates/telio-starcast/src/transport.rs
ThrasherLT: Lost that part during refactoring. Fixed now.

crates/telio-starcast/src/transport.rs
ThrasherLT: Seems like an overcomplication, since these types don't really take up large amounts of memory or anything, unless I'm missing something.

crates/telio-starcast/src/transport.rs
ThrasherLT: I'd rather not block the wait loop. Especially, if external actions will need to be handled in the future.

crates/telio-starcast/src/transport.rs
ThrasherLT: Fixed

crates/telio-starcast/src/transport.rs
stalowyjez: Hmm, maybe I'm missing sth, but it seems to me as the same amount of blocking it is currently, it would be just a bit simpler

crates/telio-starcast/src/transport.rs
stalowyjez: I don't insist, but I guess that would be a cleaner design, instead of matching on `Some(socket)` or `None`, you would match on `SocketPreparation(socket_pool, ..)` or `SocketReady(socket)` - I think it would be easier to catch some kinds of errors (like the one below) and the `State` wouldn't be polluted with some fields which are actually used only once for this socket creation.

crates/telio-starcast/src/transport.rs
stalowyjez: Well, maybe you could add some test for this, so it won't be lost during the next refactor? ;)

crates/telio-starcast/src/transport.rs
ThrasherLT: I'd really rather leave it as it is, if it's not causing any issues.

crates/telio-starcast/src/transport.rs
ThrasherLT: Added a test for it

crates/telio-starcast/src/transport.rs
ThrasherLT: As I said, if in the future we needed to call `task_exec!` while we're still waiting for the transport socket to be opened, we'd have to revert your suggested changes, so I'd rather leave it as it is, since the simplicity here is kind of subjective.

crates/telio-starcast/src/transport.rs
stalowyjez: Sure

crates/telio-starcast/src/transport.rs
stalowyjez: Ok

stalowyjez (approved):
Looks good, +1

PR #819, opened 2024-09-16
Remove redudant meshnet peers list

Hasan6979:
### Problem
There is a duplicate list of peer public keys in derp config. Hashset of public key is used for checking to which peers to accept from, and vector of public keys which is used for derp polling. 

### Solution
Derp polling list can be made from the hashset and the vector of public keys is redundant.


gytsto (dismissed):
+1

src/device.rs
```diff
@@ -1811,7 +1811,7 @@ impl Runtime {
             let derp_config = DerpConfig {
                 secret_key,
                 servers: SortedServers::new(config.derp_servers.clone().unwrap_or_default()),
-                allowed_pk: peers,
+                meshnet_peers: peers,
```
mathiaspeters, line 1814: This is probably out of scope for this PR and even more probably not significant, but you can optimize the creation of this `peers` set by doing this:
```
let peers: HashSet<PublicKey> = config
            .as_ref()
            .map(|c| {
                c.peers
                    .as_ref()
                    .map(|peers| peers.iter().map(|peer| peer.base.public_key).collect())
            })
            .flatten()
            .unwrap_or_default();
```
It would prevent a relatively expensive clone, but it could also turn out to have no benefit whatsoever

crates/telio-proto/src/packet/control/derppoll.rs
```diff
@@ -42,11 +47,16 @@ pub struct DerpPollRequestMsg {
 
 impl DerpPollRequestMsg {
     /// Creates new request message from `peers`.
-    pub fn new(session: Session, peers: &[PublicKey]) -> Self {
+    pub fn new(session: Session, peers: &std::collections::HashSet<PublicKey>) -> Self {
```
mathiaspeters, line 50: I would import instead of fully qualifying the name here, especially since we're already importing from `std::collections`

crates/telio-proto/src/packet/control/derppoll.rs
```diff
@@ -42,11 +47,16 @@ pub struct DerpPollRequestMsg {
...
         Self {
             session,
             msg: DerpPollRequest {
-                peers: RepeatedField::from_vec(peers.iter().map(|peer| peer.to_string()).collect()),
+                peers: RepeatedField::from_vec({
+                    let mut sorted_peers: Vec<String> =
+                        peers.iter().map(|peer| peer.to_string()).collect();
+                    sorted_peers.sort();
```
mathiaspeters, line 57: `sort_unstable` is to be preferred unless the you need equal items to be in the same order as they initially were, because it's faster and uses less memory (probably not significant here, but still)

mathiaspeters: Also, it looks like these fields weren't sorted before. Is there a reason they are sorted now? Or did I miss the sort somewhere?

mathiaspeters:
Some nits, but +1

crates/telio-proto/src/packet/control/derppoll.rs
Hasan6979: Yes. Copy-pasted, forgot to clean up.

crates/telio-proto/src/packet/control/derppoll.rs
Hasan6979: Since earlier there was a vector coming in from Config, and the test was based on vector that may have been incidentally sorted. There was some part where we do have sorted peers but maybe its not this one and sorting isn't really needed here at all!

mathiaspeters (approved):
+1

PR #820, opened 2024-09-16
LLT-5540: macOS python path

tomasz-grz:
### Problem
On the vagrant macOS box we have a manual installing of python. When running python commands in nat-lab, `connection.create_process()` doesn’t see the new python installation path, since `/etc/profile` is not sourced. And tries to run the default system python, which fails with:
```
xcode-select: error: no developer tools were found at '/Applications/Xcode.app', and no install could be requested (perhaps no UI is present), please install manually from 'developer.apple.com'
```

### Solution
Simplify the way python scripts are executed on macOS, by providing a helper function that gives the full path to the python installation.

- [x] README.md is updated


LukasPukenis (commented):
+1 overall a nice refactoring

mathiaspeters (approved):
Looks great! +2

PR #822, opened 2024-09-17
LLT-5542: Reduce number of direct tests

jjanowsk:
TODO


nat-lab/tests/test_direct_connection.py
```diff
@@ -39,265 +40,227 @@
...
             )
-            await exit_stack.enter_async_context(
-                beta_client.get_router().break_tcp_conn_to_host(str(server["ipv4"]))
+            for client, server in itertools.product(env.clients, DERP_SERVERS)
+        ])
+
+        await asyncio.gather(*[
+            client.wait_for_state_on_any_derp([State.Connecting, State.Disconnected])
+            for client in env.clients
+        ])
```
jjanowsk, line 131: Prepare wait futures before breaking the connection. Same for State.Connected. Should be created before restoring the connection

nat-lab/tests/test_direct_connection.py
```diff
@@ -457,295 +419,84 @@ async def test_direct_working_paths_stun_ipv6() -> None:
...
-            alpha_client.wait_for_state_on_any_derp([State.Connected]),
-            beta_client.wait_for_state_on_any_derp([State.Connected]),
-        )
-
-        async with AsyncExitStack() as temp_exit_stack:
-            await temp_exit_stack.enter_async_context(
-                alpha_client.get_router().disable_path(
-                    alpha_client.get_endpoint_address(beta.public_key)
-                )
+        await asyncio.gather(*[
```
jjanowsk, line 464: same here

nat-lab/tests/test_direct_connection.py
```diff
@@ -457,295 +419,84 @@ async def test_direct_working_paths_stun_ipv6() -> None:
...
 
-        await asyncio.gather(
-            alpha_client.wait_for_state_peer(
-                beta.public_key, [State.Connected], [PathType.Direct]
-            ),
-            beta_client.wait_for_state_peer(
-                alpha.public_key, [State.Connected], [PathType.Direct]
-            ),
-        )
+        await asyncio.gather(*[
```
jjanowsk, line 478: same here

nat-lab/tests/test_direct_connection.py
```diff
@@ -800,53 +559,43 @@ async def test_direct_working_paths_with_pausing_upnp_and_stun(
...
-        await asyncio.sleep(5)
-
-        if stun_enabled:
-            await alpha_client.wait_for_log(
-                "Skipping getting endpoint via STUN endpoint provider(ModulePaused)"
-            )
-        if upnp_enabled:
-            await alpha_client.wait_for_log(
-                "Skipping getting endpoint via UPNP endpoint provider(ModulePaused)"
+        await asyncio.gather(*[
```
jjanowsk, line 569: same here

nat-lab/tests/test_direct_connection.py
gytsto: Since you are checking for state, you can do that after the fact. Doesn't really matter if future is created before imo. If you were to check for event, then yeah, waiting for event beforehand would make sense.

nat-lab/tests/test_direct_connection.py
```diff
@@ -39,265 +40,227 @@
...
+
+            async def ping_timeout(connection, node):
+                try:
+                    await ping(connection, node.ip_addresses[0], 10)
+                except asyncio.TimeoutError:
+                    pass
+                else:
+                    # if no timeout exception happens, this means, that peers connected through relay
+                    # faster than we expected, but if no relay event occurs, this means, that something
+                    # else was wrong, so we asserts
```
gytsto, line 236: I think random letter `s` appeared at the end of the comment :smile:, change `asserts` to `assert`

nat-lab/tests/test_direct_connection.py
```diff
@@ -861,10 +610,79 @@ async def test_direct_working_paths_with_pausing_upnp_and_stun(
...
+]
+
+
+@pytest.mark.asyncio
+@pytest.mark.long
+@pytest.mark.timeout(timeouts.TEST_DIRECT_FAILING_PATHS_TIMEOUT)
+@pytest.mark.parametrize("setup_params", UHP_FAILING_PATHS)
+# Not sure this is needed. It will only be helpful to catch if any
+# libtelio change would make any of these setup work.
+async def test_direct_failing_paths(setup_params: List[SetupParameters]) -> None:
```
gytsto, line 667: We could actually get rid of this test in general or atleast add `mark.skip`, this test shouldn't be run in CI

nat-lab/tests/test_direct_connection.py
```diff
@@ -39,265 +40,227 @@
...
-@pytest.mark.parametrize("setup_params", UHP_FAILING_PATHS)
-# Not sure this is needed. It will only be helpful to catch if any
-# libtelio change would make any of these setup work.
-async def test_direct_failing_paths(setup_params: List[SetupParameters]) -> None:
-    async with AsyncExitStack() as exit_stack:
-        env = await setup_mesh_nodes(exit_stack, setup_params, is_timeout_expected=True)
-        _, beta = env.nodes
-        alpha_client, beta_client = env.clients
-        alpha_connection, _ = [conn.connection for conn in env.connections]
+async def _ping_between_all_nodes(env: Environment) -> None:
```
gytsto, line 109: might be nice to move this to `helpers.py`, so we could use this in other tests, kinda nice util

gytsto (dismissed):
Other than few nits, great work. +1

nat-lab/tests/test_direct_connection.py
jjanowsk: Ohh, I was not aware of that distinction.  I checked the code now, and you're right. Thanks!

nat-lab/tests/test_direct_connection.py
jjanowsk: It's marked as long, and since we have: `PYTEST_MARKS: "not long"` in our nat-lab job it's skipped anyway

tomasz-grz (approved):
Great work +1

nat-lab/tests/test_direct_connection.py
```diff
@@ -39,265 +40,227 @@
...
     async with AsyncExitStack() as exit_stack:
         env = await setup_mesh_nodes(exit_stack, setup_params)
-        _, beta = env.nodes
-        alpha_client, beta_client = env.clients
-        alpha_connection, _ = [conn.connection for conn in env.connections]
 
-        for server in DERP_SERVERS:
-            await exit_stack.enter_async_context(
-                alpha_client.get_router().break_tcp_conn_to_host(str(server["ipv4"]))
+        print("Test direct connection")
```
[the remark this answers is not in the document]
jjanowsk: We do have timestamps in logs

gytsto (approved):

PR #823, opened 2024-09-17
natlab: Add support for faster debug builds

lcruz99:
By default release builds are used on build scripts, and release build profile is configured to use compiler size optimization flags which slows down compilation, taking around double the time than without optimizations. Also, `tcli` is no longer required to run nat-lab, therefore it shouldn't be built by default when running build scripts.

This commit adapts build scripts to make use of debug binaries and stop building tcli implicitly. 

Tcli is now built when `--tcli` flag is passed and debug builds are now fully supported for natlab and other build scripts functionalities.

Benchmarking results for release vs. optimized dev profile:

Linux:      **6m46s -> 1m50s**
Win:        **8m24s -> 1m40s**
macOS:  **9m33s -> 43s**
Android: **2m47s -> 1m05s**

- [x] README.md is updated

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v2.4.5 # REMEMBER to also update in .gitlab-ci.yml
+          triggered-ref: v2.4.6 # REMEMBER to also update in .gitlab-ci.yml
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v2.4.5 # REMEMBER to also update in .github/workflows/gitlab.yml
+    branch: v2.4.6 # REMEMBER to also update in .github/workflows/gitlab.yml
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -206,6 +206,11 @@ lto = true
 codegen-units = 1
 debug = "full"
 
+[profile.dev]
+debug = 0
+strip = "debuginfo"
+overflow-checks = false
+
 #TODO: remove this when anyone starts using telio-starcast
 [workspace.metadata.cargo-udeps.ignore]
 normal = ["telio-starcast"]
--- a/ci/build_libtelio.py
+++ b/ci/build_libtelio.py
@@ -263,6 +263,7 @@ LIBTELIO_CONFIG = {
     "macos": {
         "packages": {
             "tcli": {"tcli": "tcli"},
+            "teliod": {"teliod": "teliod"},
             NAME: {f"lib{NAME}": f"lib{NAME}.dylib"},
         },
         "post_build": [post_copy_darwin_debug_symbols_to_distribution_dir],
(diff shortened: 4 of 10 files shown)


.github/workflows/gitlab.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           project-id: ${{ secrets.PROJECT_ID }}
           schedule: ${{ github.event_name == 'schedule' }}
           cancel-outdated-pipelines: ${{ github.ref_name != 'main' }}
-          triggered-ref: v2.4.3 # REMEMBER to also update in .gitlab-ci.yml
+          triggered-ref: LLT-5468_speedup_dev_builds # REMEMBER to also update in .gitlab-ci.yml
```
Hasan6979, line 34: Is this change temporary ?

lcruz99: yes is part of test branch commit https://github.com/NordSecurity/libtelio/pull/823/commits/24feffc72391763b385318c57705ef11b7fbf7f0

Cargo.toml
```diff
@@ -206,6 +206,11 @@ lto = true
 codegen-units = 1
 debug = "full"
 
+[profile.dev]
```
packgron, line 209: there is this porfile used, in development.
As if so, I would with we expecialy need all of these things

ci/build_libtelio.py
```diff
@@ -302,6 +304,16 @@ def main() -> None:
...
+    build_parser.add_argument(
+        "--tcli",
+        action="store_true",
+        help="Include tcli package",
+    )
+    lipo_parser.add_argument(
+        "--tcli",
+        action="store_true",
+        help="Include tcli package",
+    )
```
packgron, line 316: Why are these line duplicated?

lcruz99: they add the tcli option for both build and lipo parsers. So that we can use the `--tcli` arg both for build and lipo commands.

Cargo.toml
lcruz99: `dev` profile is already optimized for speed, I could have created a custom profile from it but thought that having `debug=0` could be the default.

lcruz99: it makes the linking stage considerably faster

nat-lab/tests/utils/vm/mac_vm_util.py
```diff
@@ -62,12 +63,10 @@ async def _copy_binaries(
                 raise exception
         await connection.create_process(["mkdir", "-p", directory]).execute()
 
-    DIST_PATH = "dist/darwin/macos/release/x86_64/"
+    DIST_PATH = f"dist/darwin/macos/{os.getenv('TELIO_BIN_PROFILE')}/x86_64/"
```
packgron, line 66: How will this act if TELIO_BIN_PROFILE is missing?

lcruz99: it raises an exception [here](https://github.com/NordSecurity/libtelio/pull/823/files#diff-a91e5f9ea38f2b6c0e059f4f4ad014acb21345cc7a344656df63eb5379d51afaR61) and asks user to define it.

.gitlab-ci.yml
```diff
@@ -15,4 +15,4 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v2.4.3 # REMEMBER to also update in .github/workflows/gitlab.yml
+    branch: LLT-5468_speedup_dev_builds # REMEMBER to also update in .github/workflows/gitlab.yml
```
Hasan6979, line 18: Adding as a reminder to fix this.

Hasan6979 (commented):
Generally lgtm +1

packgron (approved):
+1.0

LukasPukenis:
The time savings :rocket: :rocket: :rocket:

PR #824, opened 2024-09-19
[LLT-4929] traversal: Remove SM crate from cross ping check

lcruz99:
### Problem
[SM crate ](https://crates.io/crates/sm) is deprecated and thus no longer maintaned. We've been using it to define a endpoint state machine for cross ping check component of telio-traversal:
```
 sm! {
     EndpointState {
         InitialStates { Disconnected }

         SendCallMeMaybeRequest {
             Disconnected => EndpointGathering
         }

         ReceiveCallMeMaybeResponse {
             EndpointGathering => Ping
         }

         Publish {
             Ping => Published
         }

         Timeout {
             EndpointGathering => Disconnected
             Ping => Disconnected
         }

         EndpointGone {
             Published => Disconnected
         }
     }
 }
```
Because of this dependency we've been required to keep old versions of `quote`, `proc-macro2` and `syn` crates.

### Solution
This PR replaces the FSM defined using the SM crate and removes that crate from libtelio workspace. 

- [x] README.md is updated

### DIFF
--- a/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
+++ b/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
@@ -1,16 +1,17 @@
 use super::{Error, WireGuardEndpointCandidateChangeEvent};
 use crate::{
+    do_state_transition,
     endpoint_providers::{
         EndpointCandidate, EndpointCandidatesChangeEvent, EndpointProvider, EndpointProviderType,
         Error as EndPointError, PongEvent,
     },
+    endpoint_state::{EndpointState, EndpointStateMachine, Event},
     last_rx_time_provider::{is_peer_alive, TimeSinceLastRxProvider},
     ping_pong_handler::PingPongHandler,
 };
 use async_trait::async_trait;
 use enum_map::EnumMap;
 use futures::Future;
-use sm::sm;
 use std::fmt::Debug;
 use std::sync::Arc;
 use std::time::Duration;
@@ -32,56 +33,6 @@ use tokio::time::{Instant, Interval};
 const CPC_TIMEOUT: Duration = Duration::from_secs(10);
 const UPGRADE_TIMEOUT: Duration = Duration::from_secs(60);
 const MAX_SESSION_CANDIDATES: usize = 512;
-
-// Cross ping state machine definintion
-sm! {
-    EndpointState {
-        InitialStates { Disconnected }
-
-        SendCallMeMaybeRequest {
-            Disconnected => EndpointGathering
-        }
-
-        ReceiveCallMeMaybeResponse {
-            EndpointGathering => Ping
-        }
-
-        Publish {
-            Ping => Published
-        }
-
-        Timeout {
-            EndpointGathering => Disconnected
-            Ping => Disconnected
-        }
-
-        EndpointGone {
-            Published => Disconnected
-        }
-    }
-}
-use EndpointState::Variant::*;
-use EndpointState::*;
-macro_rules! do_state_transition {
-    ($machine: expr, $event: expr, $ep: expr) => {{
-        if $machine.state() == EndpointState::Published || $event == EndpointState::Publish {
-            do_state_transition!($machine, $event, $ep, telio_log_info);
-        } else {
-            do_state_transition!($machine, $event, $ep, telio_log_debug);
-        }
-    }};
-    ($machine: expr, $event: expr, $ep: expr, $log: ident) => {{
-        $log!(
-            "Node's {:?} EP {:?} state transition {:?} -> {:?}",
-            $ep.public_key,
-            $ep.local_endpoint_candidate.udp,
-            $machine.state(),
-            $event
-        );
-        $ep.last_state_transition = Instant::now();
-        $ep.state = $machine.transition($event).as_enum();
-    }};
-}
 
 #[cfg_attr(any(test, feature = "mockall"), mockall::automock)]
 #[async_trait]
@@ -278,8 +229,8 @@ impl CrossPingCheckTrait for CrossPingCheck {
                 Ok(s.endpoint_connectivity_check_state
                     .iter()
                     .filter_map(
-                        |(session, v)| match (v.state.clone(), v.last_validated_endpoint) {
-                            (PublishedByPublish(_), Some(ep)) => Some((
+                        |(session, v)| match (v.state.get(), v.last_validated_endpoint) {
+                            (EndpointState::Published, Some(ep)) => Some((
                                 v.public_key,
                                 WireGuardEndpointCandidateChangeEvent {
                                     public_key: v.public_key,
@@ -312,7 +263,7 @@ impl CrossPingCheckTrait for CrossPingCheck {
                 .values_mut()
                 .filter(|v| v.public_key == public_key);
             for session in sessions {
-                session.handle_endpoint_succesfull_notification();
+                session.handle_endpoint_successful_notification();
             }
             Ok(())
         })
@@ -398,7 +349,7 @@ impl<E: Backoff> State<E> {
                     let session = EndpointConnectivityCheckState {
                         public_key: added_node,
                         local_endpoint_candidate: endpoint.clone(),
-                        state: Machine::new(Disconnected).as_enum(),
+                        state: EndpointStateMachine::default(),
                         last_state_transition: Instant::now(),
                         last_validated_endpoint: None,
                         last_rx_time_provider: self.last_rx_time_provider.clone(),
@@ -463,7 +414,7 @@ impl<E: Backoff> State<E> {
                     public_key: node,
                     local_endpoint_candidate: added_endpoint.clone(),
                     local_session: session_id,
-                    state: Machine::new(Disconnected).as_enum(),
+                    state: EndpointStateMachine::default(),
                     last_state_transition: Instant::now(),
                     last_validated_endpoint: None,
                     last_rx_time_provider: self.last_rx_time_provider.clone(),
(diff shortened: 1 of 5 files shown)


crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,120 @@
...
+                self.state = EndpointState::Ping;
+            }
+
+            (EndpointState::Ping, Event::Publish) => {
+                self.last_event = Some(event);
+                self.state = EndpointState::Published;
+            }
+
+            (EndpointState::EndpointGathering, Event::Timeout)
+            | (EndpointState::Ping, Event::Timeout) => {
```
mathiaspeters, line 47: I hope I'm getting the syntax right here, but you're supposed to be able to do something like this
```suggestion
            (EndpointState::EndpointGathering|EndpointState::Ping, Event::Timeout) => {
```

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -1155,7 +1104,7 @@ mod tests {
     }
 
     fn prepare_test_session_in_state(
-        state: EndpointState::Variant,
+        state: (EndpointState, Option<Event>),
```
mathiaspeters, line 1107: Why is this not `EndpointStateMachine`?

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -792,18 +743,18 @@ impl<E: Backoff> EndpointConnectivityCheckState<E> {
...
             );
-            do_state_transition!(m, EndpointGone, self);
+            do_state_transition!(self, Event::EndpointGone);
         }
         Ok(())
     }
 
-    fn handle_endpoint_succesfull_notification(&mut self) {
-        if let PublishedByPublish(_) = self.state.clone() {
+    fn handle_endpoint_successfull_notification(&mut self) {
```
mathiaspeters, line 756: Nit: 
```suggestion
    fn handle_endpoint_successful_notification(&mut self) {
```

crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,120 @@
+use std::fmt;
+use telio_utils::telio_log_warn;
+
+#[derive(Debug, Default, PartialEq)]
+pub enum EndpointState {
```
mathiaspeters, line 5: Since this doesn't hold any data, I think you should derive `Copy` (and `Clone`) and pass it around by value instead of by reference

crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[default]
+    Disconnected,
+    EndpointGathering,
+    Ping,
+    Published,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Event {
```
mathiaspeters, line 14: Same thing here

crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    pub fn get(&self) -> &EndpointState {
+        &self.state
+    }
+
+    pub fn last_event(&self) -> &Option<Event> {
+        &self.last_event
+    }
```
mathiaspeters, line 69: Because then these two don't have to have the reference (at least I'm 95% sure that `Option<T>` is `Copy` when `T` is `Copy`

crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn eq(&self, other: &(EndpointState, Option<Event>)) -> bool {
+        self.state == other.0 && self.last_event == other.1
+    }
+}
+
+impl fmt::Display for EndpointStateMachine {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self.state)
+    }
+}
```
mathiaspeters, line 88: Was this automatically implemented by the `sm` macro? And if so, does this differ from the old implementation? I could very well be mistaken because I haven't looked into it at all, but I would imagine that if this existed before, it would include the event that led to the current state just by nature of how `sm` creates the enum(s)

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
lcruz99: done

crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
```diff
@@ -949,40 +900,40 @@ impl<E: Backoff> EndpointConnectivityCheckState<E> {
         let duration_in_state = Instant::now() - self.last_state_transition;
         let timeout = CPC_TIMEOUT; // TODO: make configurable
 
-        match self.state.clone() {
-            EndpointGatheringBySendCallMeMaybeRequest(m) => {
+        match (self.state.get(), self.state.last_event()) {
+            (EndpointState::EndpointGathering, _) => {
```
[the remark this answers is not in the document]
lcruz99: exactly

crates/telio-traversal/src/endpoint_state.rs
lcruz99: you're right https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

:D

crates/telio-traversal/src/endpoint_state.rs
lcruz99: I'm not sure how `sm` handled this but I've implemented this trait so that the logs end up unchanged with what was before:
`2024-09-03T13:28:54.881583Z  INFO telio_traversal::connectivity_check::cross_ping_check: 800: Node's "JRAb...n0s=" EP 200.225.115.103:24020 state transition Published -> EndpointGone`

crates/telio-traversal/src/endpoint_state.rs
lcruz99: fair enough, will change

crates/telio-traversal/src/endpoint_state.rs
lcruz99: agreed

crates/telio-traversal/src/endpoint_state.rs
mathiaspeters: Okay, that's all good then

mathiaspeters:
+1

crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,119 @@
...
+        $log!(
+            "Node's {:?} EP {:?} state transition {:?} -> {:?}",
+            $ep.public_key,
+            $ep.local_endpoint_candidate.udp,
+            $ep.state,
+            $event
+        );
+        $ep.last_state_transition = Instant::now();
+        $ep.state.handle_event($event);
+    }};
```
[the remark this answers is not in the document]
lcruz99: I'm not sure if I follow. If we are going to or coming from `EndpointState::Published` state, we should use `telio_log_info` else `telio_log_debug`. I can replace this macro but would still need to do that check.

what do you suggest?

crates/telio-traversal/src/endpoint_state.rs
```diff
@@ -0,0 +1,119 @@
...
+    ReceiveCallMeMaybeResponse,
+    Publish,
+    Timeout,
+    EndpointGone,
+}
+
+#[derive(Default, Debug)]
+pub struct EndpointStateMachine {
+    state: EndpointState,
+    last_event: Option<Event>,
```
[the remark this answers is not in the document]
lcruz99: I think additional states will further increase complexity.

Yes, transitions are handled a bit different in `sm` crate, I think underneath they use dyn traits, consuming the state machine on each transition. This makes the build fail if an invalid transition is attempted. On our case we'll only get a warning in runtime. Do you suggest we should handling it differenttly?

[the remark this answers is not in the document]
LukasPukenis: `sm` crate seemed to do a really good job if it didn't allow invalid transitions(i didn't know that) :).

My point about `last_event` was that it was used together with the current state to decide about things. This introduces complexity but worse I think it allows for bugs to appear more easily.

As an improvement I suggest to remove `last_event` and represent each unique state separately. There shouldn't be that much of a rework, basically convert two variables into a unique enum.

For example `Disconnected` state is at least issued from these events that are then stored together: `EndpointGone`, `Timeout`. I suggest:
```
enum State {
    Disconnected(DisconnectedReason),
}

enum DisconnectedReason {
    Timeout,
    EndpointGone
}
```

Does that sound reasonable?

crates/telio-traversal/src/endpoint_state.rs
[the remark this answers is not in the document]
LukasPukenis: It's fine. I was initially skeptical that it's needed but I see that I do not have a better suggestion, except maybe general dislike for macros :D

crates/telio-traversal/src/endpoint_state.rs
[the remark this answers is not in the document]
lcruz99: > Does that sound reasonable?

actually more than reasonable, it sounds better 😅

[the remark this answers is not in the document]
lcruz99: 🤔  instead of using `last_state` do evaluate the transition we would use `DisconnectedReason` enum, in the end will be the same logic.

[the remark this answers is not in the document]
lcruz99: And actually `last_state` is giving information that otherwise we would have to implement using the <Disconnected, Ping, EndpointGathering>Reason enums

packgron:
There is one more thing to consider, this can for now be a reading, but there is a state machine patter, there you use type system to enforce what transitions can happen at compile time: https://hoverbear.org/blog/rust-state-machine-pattern/

If I'm not mistaken sm create was basically doing this, so we could rol out our manual version, but not sure if all the boiler plate is truly worth it :sweat_smile:

LukasPukenis (approved):

PR #827, opened 2024-09-20
LLT-5534: Use generated meshnet config in natlab

mathiaspeters:
This is the second PR to start using types generated from UniFFI instead of defining new ones, where possible, in natlab. This PR deals with the meshnet config.

Some notes:
- The `(get|set)_meshmap` functions have been renamed to `(get|set)_meshnet_config`
- Strong types instead of dicts means that we can do things like `derp.ipv4` instead of `derp["ipv4"]`

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_telio_version_compatibility.py
+++ b/nat-lab/tests/test_telio_version_compatibility.py
@@ -3,10 +3,18 @@ import json
 import pytest
 import shlex
 import telio
+from config import DERP_SERVERS
 from contextlib import AsyncExitStack
 from mesh_api import API
+from typing import Any, List, Dict
 from utils import testing
-from utils.bindings import features_with_endpoint_providers, EndpointProvider
+from utils.bindings import (
+    features_with_endpoint_providers,
+    EndpointProvider,
+    Config,
+    Peer,
+    Server,
+)
 from utils.connection_util import (
     ConnectionTag,
     ConnectionLimits,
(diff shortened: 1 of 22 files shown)


jjanowsk (dismissed):
+1

nat-lab/tests/test_telio_version_compatibility.py
```diff
@@ -29,6 +37,60 @@
...
+            "region_code": derp.region_code,
+            "name": derp.name,
+            "hostname": derp.hostname,
+            "ipv4": derp.ipv4,
+            "relay_port": derp.relay_port,
+            "stun_port": derp.stun_port,
+            "stun_plaintext_port": derp.stun_plaintext_port,
+            "public_key": derp.public_key,
+            "weight": derp.weight,
+            "use_plain_text": derp.use_plain_text,
```
stalowyjez, line 66: It looks to me that it could be done in a bit simpler way using `vars`. If you need to remove `conn_state` field, I get the same results with this:
```
def backport_derp_v2(derp: Server) -> Dict[str, Any]:
    result = vars(derp)
    result.pop("conn_state")
    return result
```

stalowyjez (dismissed):
Overall it looks good, just one small comment (not necessary though), +1

nat-lab/tests/test_telio_version_compatibility.py
stalowyjez: Hmm, maybe some comment why do you omit it would be also nice, maybe some test? idk

mathiaspeters: I do like the idea in general, but in this case I would prefer to not make that change for two reasons:
- most important reason, if we change the structure of `Server` we might get cryptic errors but with the way of doing the conversion we'll get a clear type error that derp doesn't contain a field
- less important, I think it's more consistent since we have these bigger dictionary creations for the other parts of the backporting functions

stalowyjez (approved):

PR #828, opened 2024-09-20
[LLT-5563] Minimal hosted version of API in nat-lab

olekoliinyk:
### Problem
This is a minimal hosted version of API in nat-lab. 
It implements all /machine API methods.
Some of the data is mocked and the actual configuration of the peers is not implemented.
This will be taken care of with further PRs as redesigning the current mesh_api implementation is required. 

### Solution

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/core-api.py
+++ b/nat-lab/bin/core-api.py
@@ -3,11 +3,29 @@
 import json
 import paho.mqtt.client as mqtt  # type: ignore # pylint: disable=import-error
 from dataclasses import asdict, dataclass
+from enum import Enum
+from http import HTTPStatus
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from itertools import count
-from pprint import pprint
-from typing import Optional
+from typing import Dict, Optional
 from uuid import uuid4
+
+DERP_SERVER = {
+    "region_code": "nl",
+    "name": "Natlab #0001",
+    "hostname": "derp-01",
+    "ipv4": "10.0.10.1",
+    "relay_port": 8765,
+    "stun_port": 3479,
+    "stun_plaintext_port": 3478,
+    "public_key": "qK/ICYOGBu45EIGnopVu+aeHDugBrkLAZDroKGTuKU0=",  # NOTE: this is hardcoded key for transient docker container existing only during the tests
+    "weight": 1,
+}
+
+
+class CoreApiErrorCode(Enum):
+    MACHINE_ALREADY_EXISTS = 101117
+    MACHINE_NOT_FOUND = 101102
 
 
 @dataclass
@@ -16,7 +34,7 @@ class MachineCreateRequest:
     hardware_identifier: str
     os: str
     os_version: str
-    device_type: str = "other"
+    device_type: Optional[str] = None
     app_user_uid: Optional[str] = None
     nickname: Optional[str] = None
     discovery_key: Optional[str] = None
@@ -42,10 +60,9 @@ class Node:
 class CoreServer(HTTPServer):
     def __init__(self, server_address, RequestHandlerClass, mqttc: mqtt.Client) -> None:
         super().__init__(server_address, RequestHandlerClass)
-        # NOTE: In future this will most likely need to become a map from uuids to nodes so that it's possible
-        # to implement 'GET /v1/meshnet/machines/{machineIdentifier}/map'
-        self._known_machines: set[str] = set()
+        self._known_machines: Dict[str, Node] = {}
         self._mqttc = mqttc
+        self._id_counter = count(1)
 
     def _send_notification(self):
         message = {
@@ -65,28 +82,53 @@ class CoreServer(HTTPServer):
         msg_info.wait_for_publish()
 
     def add_machine(self, node):
-        self._known_machines.add(node.identifier)
+        self._known_machines[node.identifier] = node
         self._send_notification()
 
     def remove_machine(self, uid):
-        self._known_machines.remove(uid)
-        self._send_notification()
+        if self._known_machines.pop(uid, None) is not None:
+            self._send_notification()
+            return True
+        return False
+
+    def get_machines(self):
+        return self._known_machines
+
+    def next_id(self):
+        return next(self._id_counter)
 
 
 class CoreApiHandler(BaseHTTPRequestHandler):
     def __init__(self, request, client_address, server: CoreServer):
         self.server: CoreServer
         self.machines_path = "/v1/meshnet/machines"
-        self.next_id_generator = count(1)
         super().__init__(request, client_address, server)
 
-    def _set_headers(self):
-        self.send_response(200)
-        self.send_header("Content-type", "text/html")
+    def _set_headers(
+        self, content_type="application/json", status_code: int = HTTPStatus.OK
+    ):
+        self.send_response(status_code)
+        self.send_header("Content-type", content_type)
         self.end_headers()
 
+    def _write_response(self, response, status_code: int = HTTPStatus.OK):
+        self._set_headers(status_code=status_code)
+        self.wfile.write(json.dumps(response).encode("utf-8"))
+
+    def _send_error_response(
+        self, error_code: CoreApiErrorCode, message: str, status_code: int
+    ):
+        error_response = {"errors": {"code": error_code.value, "message": message}}
+        self._write_response(error_response, status_code)
+
     def do_GET(self):
-        self._set_headers()
+        if self.path == "/":
+            self.handle_root_path()
+        elif self.path == self.machines_path:
+            self.handle_get_machines()
+        elif self.path.split("/")[-1] == "map":
+            machine_id = self.path.split("/")[-2]
+            self.handle_get_machine_map(machine_id)
 
     def do_HEAD(self):
         self._set_headers()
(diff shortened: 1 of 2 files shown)


nat-lab/bin/core-api.py
```diff
@@ -132,31 +170,94 @@ def add_node(self, req):
...
-            self.server.remove_machine(uid)
-            print(f"{uid} removed")
-            self.send_response(204)
-        except KeyError:
-            print(f"{uid} unknown")
-            self.send_response(400)
+        if self.server.remove_machine(uid):
+            self._set_headers(status_code=204)
+        else:
+            self._send_error_response(101102, "Machine not found", 404)
```
LukasPukenis, line 190: Maybe there's a library for the HTTP status codes we could use? They are pretty descriptive in other languages as an enum, would be nice to use those instead of just numbers. This would reduce the cognitive load slightly :)

gytsto (dismissed):
+1

nat-lab/tests/test_core_api.py
```diff
@@ -0,0 +1,294 @@
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "machine_data",
+    [[
+        create_machine_data(
+            "test-public-key-1", "88cc55d2c6f64ff680aaa7235374411d", "linux", "5.10"
+        ),
+        create_machine_data(
+            "test-public-key-2", "98dd66e2d7f74ff780bbb8236485522e", "windows", "5.11"
+        ),
```
LukasPukenis, line 107: What's the format of a hash and how it was created? If it's required to be derived from the key then it would be better to use `func('test-public-key-1')` instead of a hash :)

nat-lab/bin/core-api.py
```diff
@@ -5,18 +5,29 @@
...
+    "region_code": "mock",
+    "name": "mock",
+    "hostname": "mock-01",
+    "ipv4": "1.1.1.1",
+    "relay_port": 1111,
+    "stun_port": 1111,
+    "stun_plaintext_port": 1111,
+    "public_key": "mock",
+    "weight": 1,
+}
```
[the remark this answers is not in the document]
olekoliinyk: It is fake data. 
As I mentioned in the PR description:
#Some of the data is mocked and the actual configuration of the peers is not implemented. This will be taken care of with further PRs as redesigning the current mesh_api implementation is required.

This is only to allow basic nat-lab tests to pass in case we will have an emulator with tests running in nat-lab earlier than full implementation. I.e. we will be able to register the device and get the meshmap, but of course not the pings.
The main issue is that we cannot re-use the methods/ variables/ classes from nat-lab/tests according to nat-lab architecture (here core-api is in /bin folder).
With further PRs it will be addressed.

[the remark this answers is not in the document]
LukasPukenis: Why not just return empty data list instead, that way we could simply ignore it. Splitting the PR makes perfect sense, but the change could be introduced alongside other PR. Of course if the PR is from another repo and it's not ready yet, that might cause a problem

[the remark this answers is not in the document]
olekoliinyk: Excellent question. There is validation on esp32 side when getting meshmap: if derp_servers == 0 then Fail, otherwise parse derp ip, pub key, port etc.

nat-lab/tests/test_core_api.py
```diff
@@ -0,0 +1,294 @@
...
+@pytest.mark.parametrize(
+    "machine_data",
+    [[
+        create_machine_data(
+            "test-public-key-1", "88cc55d2c6f64ff680aaa7235374411d", "linux", "5.10"
+        )
+    ]],
+    indirect=True,
+)
+# pylint: disable=unused-argument
```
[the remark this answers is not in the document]
olekoliinyk: I think it would be not possible since there is indirect parametrization. Unused data is machine_data, which is used in fixture_register_machine

nat-lab/bin/core-api.py
```diff
@@ -132,31 +170,94 @@ def add_node(self, req):
...
+            "os_version": node.os_version,
+            "device_type": node.device_type,
+            "nickname": node.nickname,
+            "ip_addresses": node.ip_addresses,
+            "discovery_key": node.discovery_key,
+            "relay_address": node.relay_address,
+            "traffic_routing_supported": node.traffic_routing_supported,
+            "peers": peers,
+            "dns": {"domains": [], "hosts": {}},
+            "derp_servers": [MOCKED_DERP_SERVER],
```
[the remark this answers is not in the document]
olekoliinyk: Please, check my reply under this discussion: https://github.com/NordSecurity/libtelio/pull/828#discussion_r1784469127

nat-lab/bin/core-api.py
```diff
@@ -132,31 +170,94 @@ def add_node(self, req):
...
             os=req.os,
             os_version=req.os_version,
-            device_type=req.device_type,
-            nickname=req.nickname,
-            discovery_key=req.discovery_key,
-            relay_address=req.relay_address,
+            device_type=req.device_type or "",
+            nickname=req.nickname or "",
+            discovery_key=req.discovery_key or "",
+            relay_address=req.relay_address or "",
```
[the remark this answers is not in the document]
olekoliinyk: Those fields are optional in POST request. However, in the response, they are presented and all should have the type: string.

nat-lab/bin/core-api.py
```diff
@@ -132,31 +186,90 @@ def add_node(self, req):
...
-        try:
-            self.server.remove_machine(uid)
-            print(f"{uid} removed")
-            self.send_response(204)
-        except KeyError:
-            print(f"{uid} unknown")
-            self.send_response(400)
+        if self.server.remove_machine(uid):
+            self._set_headers(status_code=HTTPStatus.NO_CONTENT)
+            self._set_headers(status_code=HTTPStatus.NO_CONTENT)
```
stalowyjez, line 205: Why do you need to call it twice?

nat-lab/tests/test_core_api.py
```diff
@@ -0,0 +1,413 @@
...
+        return registered_machines
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "machine_data",
+    [[
+        create_machine_data(
+            generate_wireguard_like_public_key(),
+            generate_hardware_identifier(),
```
stalowyjez, line 182: Nit: I'm not sure about this, but maybe it would be better to fix sth here? I guess test runs should be as similar as possible.

stalowyjez (commented):
Looks quite good!, +1

nat-lab/tests/test_core_api.py
olekoliinyk: Well, we don't have any validation on keys. From API perspective string type is expected.

nat-lab/bin/core-api.py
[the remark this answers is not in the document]
LukasPukenis: Using real, already existing container is a great idea :) makes reviewing a breeze as well. I am happy with how this was resolved.

nat-lab/bin/core-api.py
LukasPukenis: I see usage of HTTP_STATUS :rocket:

nat-lab/tests/test_core_api.py
```diff
@@ -0,0 +1,356 @@
...
+        )
+
+        response_data = await send_http_request(
+            connection,
+            f"{server_host}/v1/meshnet/machines",
+            method="POST",
+            data=payload,
+        )
+
+        assert response_data["errors"]["code"] == 101117
```
LukasPukenis, line 352: This also could benefit from an enum value

nat-lab/tests/test_core_api.py
```diff
@@ -0,0 +1,356 @@
...
+            connection,
+            f"{server_host}/v1/meshnet/machines/{machine['identifier']}",
+            method="DELETE",
+            expect_response=False,
+        )
+
+
+linux_vm = MachineData(
+    public_key="zIdfzkxAhNKLj5ayl232xO/cnFm+v/9BK47mOdAxB1c=",
+    hardware_identifier="01c191a655e30be58dbe48c6a1c1508a",
```
LukasPukenis, line 122: This still raises me a question. Maybe it makes sense though but as addressed in the previous comment:
> Well, we don't have any validation on keys. From API perspective string type is expected.

If these are not validated in any way but just need to be strings, I think it might be beneficial to use obvious strings, something like `"linux_vm_public_key_str"`.

My main motivation is to reduce complexity, seeing real-looking public key looks like it's generated somehow or must correlate with something :)

LukasPukenis (commented):
MY comments were addressed +1

nat-lab/tests/test_core_api.py
olekoliinyk: It make sense. I think I misunderstood your previous comment.

nat-lab/bin/core-api.py
```diff
@@ -132,31 +186,89 @@ def add_node(self, req):
...
+        if not machine:
+            self._send_error_response(
+                CoreApiErrorCode.MACHINE_NOT_FOUND,
+                "Machine not found",
+                HTTPStatus.NOT_FOUND,
+            )
+            return
 
-        self.end_headers()
+        content_length = int(self.headers["Content-Length"])
```
Jauler, line 222: Note that I do not thing this header is required. And it is possible that it may be skipped. Although if both, esp32 and teliod includes it -> we may expect that it will exist here.

nat-lab/bin/core-api.py
```diff
@@ -132,31 +186,89 @@ def add_node(self, req):
...
+        patch_data = self.rfile.read(content_length)
```
Jauler, line 223: Note that `Content-Length` header is not required, and I am not sure we include them in ESP32 and/or teliod implementations. Therefore I guess it would be better not to depend on it. One way to do that is:
```suggestion
        if "Content-Length" not in self.headers:
            patch_data = self.rfile.read()
        else:
            content_length = int(self.headers["Content-Length"])
            patch_data = self.rfile.read(content_length)
```

Jauler (approved):
+1.0, just lets handle Content-Length being missing.

nat-lab/bin/core-api.py
olekoliinyk: I think Content-Length is required by HTTP/1.1 protocol which default protocol in [esp_http_client](esp_http_client https://github.com/espressif/esp-idf/blob/master/components/esp_http_client/esp_http_client.c#L161).
But when there is request being send with the payload, Content-Length is calculated by CURL (or esp_http_client in our case). 
For example executing below command:
>  curl -v -X POST http://localhost:8080/<endpoint> \
>    -H "Content-Type: application/json" \
>    -d '{"key": "test_value"
>        }'

Results to:
> POST /v1/meshnet/machines HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.88.1
> Accept: */*
> Content-Type: application/json
> **Content-Length: 238**

nat-lab/bin/core-api.py
```diff
@@ -3,20 +3,38 @@
...
+from enum import Enum
+from http import HTTPStatus
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from itertools import count
-from pprint import pprint
-from typing import Optional
+from typing import Dict, Optional
 from uuid import uuid4
 
+DERP_SERVER = {
```
gytsto, line 13: Can [this](https://github.com/NordSecurity/libtelio/blob/326a972824ee7c92f211e05acc085905113081de/nat-lab/tests/config.py#L142) be used here?

gytsto (approved):
other than one nit, +1

nat-lab/bin/core-api.py
olekoliinyk: With current set-up I think no. We cannot do direct import from nat-lab/tests/config.py because we're not copying nat-lab/tests to the containers which runs from /bin. Even if we copy them, python dependencies may be not met for nat-lab/tests in the core-api docker container.
I believe we can address it later on with re-designing mesh_api.py and connecting it to core-api.py

gytsto: ah yeah, makes sense, missed that part. all good then

PR #830, opened 2024-09-24
LLT-4104: nc wrapper

tomasz-grz:
### Problem
Tests from `test_mesh_firewall.py` and `test_vpn.py` use netcat (nc), however until now only Linux containers were used to run this tests. To be able to run them on Windows/Mac VMs it's needed to adapt these commands.

nc comand is present on macOS but is not behaving exactly as the Linux counterpart, on macOS nc does not report any events in verbose mode when listening (-v and -l flags), and events in client mode are printed in a different format.

### Solution
This PR adds a wrapper class to help running tests with netcat instead of creating the process manually.
Additionally it adds a clone of the nc command written in python, to be used on macOS and hopefully on Windows.
Finally the `test_fullcone_nat.py` tests were removed as they were not used for anything functional.

- [x] README.md is updated


nat-lab/tests/utils/netcat.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class NetCatServer(NetCat):
+    """Helper class for NC server"""
+
+    def __init__(
+        self,
+        connection: Connection,
+        port: int,
+        udp: bool = False,
```
tomasz-grz, line 125: ipv6

tomasz-grz: detached

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,192 @@
...
+import sys
+import time
+from typing import TextIO, List, Union
+
+UDP_SCAN_COUNT: int = 5
+RECV_SIZE = 4096
+
+
+class NetCat:
+    """A netcat clone written in Python"""
```
tomasz-grz, line 14: Alternative would have been to build BSD netcat from source on macOS, but then we would still need to find a solution for Windows

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,192 @@
...
+            self.sock.send(b"X")
+            time.sleep(1)
+
+    def _readwrite(self):
+        """Loop reading though sockets and stdin"""
+        inputs: List[Union[socket.socket, TextIO]] = [self.sock]
+        if not self.args.d:
+            inputs.append(sys.stdin)
+        while True:
+            readable, _, _ = select.select(inputs, [], [])
```
tomasz-grz, line 106: On Windows `select()` only works for sockets and not `stdin` 😞

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+                    for key, _ in events:
+                        callback = key.data
+                        callback()
+        except KeyboardInterrupt:
+            print("")
+            sys.exit(2)
+        except OSError as e:
+            if e.errno in (errno.EBADF, errno.ECONNREFUSED):
+                # connection closed
+                sys.exit(0)
```
Jauler, line 160: This will introduce silent failures, where we expect NC to work, but for some reason it does not. 

Is there any reason why this condition should be silenced and then process have to exit with success?

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+                self._register_socket()
+                self._register_stdin()
+                while not self.should_close:
+                    events = self.selector.select(timeout=None)
+                    for key, _ in events:
+                        callback = key.data
+                        callback()
+        except KeyboardInterrupt:
+            print("")
+            sys.exit(2)
```
Jauler, line 156: Is there a need to handle keyboard interrupt silently?

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+            self.selector.unregister(sys.stdin)
+            self.should_close = True
+
+    def run(self):
+        try:
+            if self.args.l:
+                self._listen()
+            else:
+                self._connect()
+            if not self.args.z:
```
Jauler, line 146: -z seems to be unsupported. Maybe it is worth simply failing the startup with argparse error? IMHO it would make it clearer to see what is happening.

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+            # Bind to a local port or source address if specified
+            if self.args.p or self.args.s:
+                local_address: str = self.args.s if self.args.s else ""
+                local_port: int = self.args.p if self.args.p else 0
+
+                self.sock.bind((local_address, local_port))
+
+            self.sock.connect((hostname, port))
+            if self.udp:
+                self._udptest()
```
Jauler, line 58: I am afraid this may generate unexpected traffic.

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+            self._vprint(
+                f"Connection to {hostname} {port} port [{self.sock_type}/*] succeeded!"
+            )
+        except OSError as e:
+            if e.errno == errno.EBADF:
+                # connection closed
+                sys.exit(0)
```
Jauler, line 65: This may silently hide conditions where the connection is closed by remote peer. Maybe it is worth signaling this condition somehow to tests?

nat-lab/tests/test_mesh_firewall.py
```diff
@@ -449,67 +448,30 @@ async def test_mesh_firewall_file_share_port(
             with pytest.raises(asyncio.TimeoutError):
                 await ping(connection_beta, CLIENT_ALPHA_IP, 15)
 
-        output_notifier = OutputNotifier()
-        listening_start_event = asyncio.Event()
-        sender_start_event = asyncio.Event()
-        connected_event = asyncio.Event()
+        async with NetCatServer(
+            connection_alpha, PORT, udp=True, ipv6=CLIENT_PROTO == IPProto.IPv6
```
Jauler, line 452: I think source_IP got lost somewhere for `NetCatServer()` processes.

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+        """Listen for incoming connections"""
+        # Set SO_REUSEPORT option
+        try:
+            self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        except AttributeError:
+            # SO_REUSEPORT may not be available on all systems, fall back to SO_REUSEADDR
+            self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+
+        port = self.args.port
+        bind_addr = "::" if self.ipv6 else "0.0.0.0"
```
Jauler, line 82: I think you are missing `source_ip` here. If it is set for server mode -> we should bind server-side sockets to some specific source IP as well.

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+                sys.stdin, selectors.EVENT_READ, self._read_from_stdin
+            )
+
+    def _read_from_socket(self):
+        """Handle incoming data from the socket"""
+        data, addr = self.sock.recvfrom(RECV_SIZE)
+        if self.udp and self.listen and not self.client_addr:
+            self._vprint(f"Connection received on {addr[0]} {addr[1]}")
+            self.client_addr = addr
+            self.sock.connect(addr)
```
Jauler, line 118: I guess for our purposes -> I am not sure if we really need connected sockets. Maybe it is worth simplifying our nc variant, and just always use unconnected sockets for UDP?

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+    parser.add_argument("-6", dest="ipv6", action="store_true", help="Use IPv6")
+    parser.add_argument(
+        "-4", dest="ipv4", action="store_true", help="Use IPv4 addresses"
+    )
+    parser.add_argument("-v", action="store_true", help="Verbose mode")
+    parser.add_argument("-n", action="store_true", help="Do not resolve hostnames")
+    parser.add_argument("-l", action="store_true", help="Listen mode (server)")
+    parser.add_argument("-u", action="store_true", help="UDP mode")
+    parser.add_argument("-d", action="store_true", help="Do not read from stdin")
+    parser.add_argument(
```
Jauler, line 180: IMHO we can remove `-z` as it is not supported, and passing it, causes silent closure.

nat-lab/bin/netcat.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+    parser.add_argument("-p", type=int, help="Bind to local port number")
+    parser.add_argument("-s", type=str, help="Bind to local source address")
+    parser.add_argument("hostname", nargs="?", help="Hostname or IP address")
+    parser.add_argument("port", type=int, help="Port number")
+    args = parser.parse_args()
+
+    if args.z and args.l:
+        parser.error("Cannot use -z in listen mode (-l)")
+    if args.s and args.l:
```
Jauler, line 191: I think we need to support listening on specified interface (`-l`&`-s` combination).

Jauler: Oh wait, on BSD you specify listening IP with just positional argument. My Bad.

nat-lab/bin/netcat.py
tomasz-grz: Hmmm what do you mean, -z seems to be unsupported?
I tried to re-crate the behavior [form](https://sources.debian.org/src/netcat-openbsd/1.195-2/netcat.c/#L915) as closely as I could, and it seems to be the same from my tests and pcaps.

It tries to connect to the listener and then immediately closes and exits with 0 if it does, without reading anything from stdin, otherwise it "fails" with 1  from `Connection refused`

nat-lab/bin/netcat.py
tomasz-grz: This is also according to the ["original" behavior](https://sources.debian.org/src/netcat-openbsd/1.195-2/netcat.c/#L1852)

nat-lab/bin/netcat.py
tomasz-grz: This is kind of what is also happening in the original `nc`, when the remote closes the connection it just silently exits with 0. Otherwise the tests fail because the listener process returns a non 0 code when the client process finishes first. 🤔

nat-lab/bin/netcat.py
tomasz-grz: Thanks, good catch!

nat-lab/bin/netcat.py
tomasz-grz: Hmm, I tried here to implement [this bit](https://sources.debian.org/src/netcat-openbsd/1.195-2/netcat.c/#L774), without the `connect()` the process will not finish when trying to send data if the listener is gone.

nat-lab/bin/netcat.py
tomasz-grz: I'll double check this, I might have over compensated for the other EBADF in connect 😰

nat-lab/bin/netcat.py
tomasz-grz: I tried to make the stdout and stderr look the same as the `nc` command, otherwise it prints a bunch of python errors

nat-lab/bin/netcat.py
tomasz-grz: I think that is the expected behavior, no? -z will exit "successfully" only if there is an open port and it succeeds connecting, otherwise it fails with connection refused.

nat-lab/bin/netcat.py
tomasz-grz: Right, this actually covered for the case when UDP listener was gone

nat-lab/bin/netcat.py
Jauler: Oopsie sorry, somehow I completely mixed up the meaning of `-z` flag. Yeah, you are right. I am resolving this comment.

nat-lab/bin/netcat.py
Jauler: sorry, I got mixed up with the meaning of `-z`

nat-lab/bin/netcat.py
tomasz-grz: You're right, this is one is an extra here, I probably forgot it when refactoring things

nat-lab/bin/netcat.py
Jauler: :thinking: ok, I was not aware of this `nc` behavior :O. Anyways, resolving the issue.

nat-lab/bin/netcat.py
Jauler: Yeah, I guess if we are trying to create honest copy of `nc` it would make sense. But in our situation, I think having more verbose errors will be beneficial down the line, even if `nc` was not producing the same ones.

nat-lab/bin/netcat.py
Jauler: good point about exiting. Missed it.

nat-lab/bin/netcat.py
Jauler: got it. Thanks for clarifying.

nat-lab/bin/netcat.py
tomasz-grz: So I wouldn't say unexpected. But what it does for sure is slow down the test because of those sleeps 😪

nat-lab/bin/netcat.py
tomasz-grz: Moved the error log above, so it always logs the OSError

nat-lab/bin/netcat.py
tomasz-grz: Added a more descriptive here

nat-lab/tests/test_mesh_firewall.py
tomasz-grz: added `bind_ip`

nat-lab/tests/utils/netcat.py
```diff
@@ -0,0 +1,220 @@
...
+from typing import Optional, AsyncIterator
+from utils.connection import Connection, TargetOS
+from utils.output_notifier import OutputNotifier
+from utils.process import Process
+from utils.python import get_python_binary
+
+
+def _get_netcat_script_path(connection: Connection) -> str:
+    if connection.target_os == TargetOS.Windows:
+        return LIBTELIO_BINARY_PATH_WINDOWS_VM + "netcat.py"
```
jjanowsk, line 14: Please either copy it to the windows vm or raise NotImplementedError here

nat-lab/tests/utils/netcat.py
```diff
@@ -0,0 +1,220 @@
...
+            flags += "u"
+        if detached:
+            flags += "d"
+        if port_scan and not listen:
+            flags += "z"
+
+        # use the built in netcat command on linux
+        if connection.target_os == TargetOS.Linux:
+            command = ["nc", flags, str(port)]
+        else:
```
jjanowsk, line 73: I think this if/else magic should be in _get_netcat_script_path. Now you have it in two places

nat-lab/tests/test_netcat.py
```diff
@@ -0,0 +1,56 @@
...
+from contextlib import AsyncExitStack
+from helpers import SetupParameters, setup_environment
+from utils.connection_util import ConnectionTag
+from utils.netcat import NetCatServer, NetCatClient
+
+TEST_STRING = "test_data"
+PORT = 12345
+
+
+@pytest.mark.asyncio
```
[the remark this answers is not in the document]
jjanowsk: Let's leave it. Should be relatively quick anyway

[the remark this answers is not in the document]
jjanowsk: But maybe let's just do one test with linux <-> mac connection to check interoperability of those two implementations? And just parameterize it for udp/tcp?

nat-lab/tests/test_vpn.py
```diff
@@ -436,45 +428,35 @@ async def connect(
...
             # Wait for both netcat processes
-            await sender_start_event.wait()
-            await sender_start_event.wait()
+            await asyncio.gather(
+                nc_client_1.connection_succeeded(), nc_client_2.connection_succeeded()
+            )
+
+            # exchange some data
+            await nc_client_2.send_data("GET")
+            await nc_client_2.receive_data()
```
jjanowsk, line 459: What is being received here by nc_client_2?

jjanowsk (commented):
Clean one! Nice! Few nits and one possible bug hiding somewhere, unless I've missed something.

nat-lab/tests/test_netcat.py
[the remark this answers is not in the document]
tomasz-grz: Good idea!

nat-lab/tests/utils/netcat.py
tomasz-grz: Yup, I think this was leftover from when I was testing it on windows 😞

nat-lab/tests/utils/netcat.py
tomasz-grz: Good idea!

nat-lab/tests/test_vpn.py
tomasz-grz: Just waiting to receive the HTTP reply from the PHOTO_ALBUM server. It's not necessary for the test, I think it was just leftover checking if the method works, I can remove

Jauler (approved):
+1.0

nat-lab/tests/test_netcat.py
```diff
@@ -0,0 +1,55 @@
...
+    "setup_params",
+    [
+        pytest.param(
+            SetupParameters(
+                connection_tag=ConnectionTag.MAC_VM,
+            ),
+            marks=pytest.mark.mac,
+        ),
+    ],
+)
```
jjanowsk, line 24: You can just get rid of `setup_params` parameter, and hardcode MAC_VM the same way you did for CONE_CLINET

tomasz-grz: Then it's just gonna be an easy copy paste to add windows :D

PR #831, opened 2024-09-26
Make STUN and UPnP providers optimization enabled by default

dfetti:
I runned some tests with optimization enabled/disabled to measure the relay to direct recovery time.

The scenario used for tests:
- Node A and B are directly connected
- Cut the connection and expect downgrade to relay
- Undo the cut and wait for path upgrade to direct

Times measured between relay event to the direct event showed that the recover process is faster with the optimization enabled.

For STUN is with 3-4 seconds faster.
For UPNP is with 10 seconds faster.

Nevertheless this is a battery lifetime improvement feature and we want to push apps to use it leaving them the option to disable it if needed.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-4963
+++ b/.unreleased/LLT-4963
@@ -1,0 +1,1 @@
+Make STUN and UPnP providers optimization enabled by default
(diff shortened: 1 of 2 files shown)


lcruz99 (approved):
+2

PR #833, opened 2024-09-26
Add trigger ability for batcher

LukasPukenis:
Add triggering ability to batcher so it could
evaluate deadlines and thresholds on demand.

The approach is simple - any activity on sent
or received data on any peer will trigger the batcher.

This is built on assumption that triggering on incoming
data would sync the batchers between two devices. However
triggering only between two devices would leave other devices
unsynced, thus simplified approach works even better
to "sync the clocks" across all the nodes.

Side effects:
consider having 3 interconnected peers: A, B and C.
Peer C is idling and A streams data to B.
Now A and B is each triggered on every packet and in turn
send premature keepalives to C at T_new=T_orig-threshold.

Triggering will happen mostly on wg_consolidate() which happens
every second.

- [x] README.md is updated


crates/telio-traversal/src/session_keeper.rs
```diff
@@ -62,7 +63,10 @@ pub struct SessionKeeper {
 }
 
 impl SessionKeeper {
-    pub fn start(sock_pool: Arc<SocketPool>) -> Result<Self> {
+    pub fn start(
+        sock_pool: Arc<SocketPool>,
+        activity_recorder: Option<Arc<dyn ActivityRecorder>>,
```
LukasPukenis, line 68: I would like not to introduce an interface just for one function which is not gonna be used anywhere else, however I failed to extract it to a lambda. Suggestions very welcome :)

crates/telio-wg/src/wg.rs
```diff
@@ -44,6 +44,26 @@ use std::{
     time::Duration,
 };
 
+/// Interface for retrieving stats about network activity
+#[async_trait]
+pub trait ActivityRecorder: Sync + Send {
+    /// Retrieve global peer network activity timestamps
+    async fn get_latest_peer_network_activity(&self) -> Result<Option<TxRxTimestampPair>, Error>;
+}
```
LukasPukenis, line 52: This trait could be omitted, however my rust-fu is limited here. session_keeper::start() is the place to look at for more context

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -274,6 +289,25 @@ impl Runtime for State {
...
+            if let Ok(Some(timestamps)) = wg.get_ts().await {
+                tx_has_changed = self.last_tx_ts.map_or(false, |ts| timestamps.tx_ts > ts);
+                self.last_tx_ts = Some(timestamps.tx_ts);
+
+                rx_has_changed = self.last_rx_ts.map_or(false, |ts| timestamps.rx_ts > ts);
+                self.last_rx_ts = Some(timestamps.rx_ts);
+            }
+        }
+
+        if tx_has_changed || rx_has_changed {
```
stalowyjez, line 306: Hmm, shouldn't it be only `rx_has_changed` ? And is `SessionKeeper` the best entity to handle the triggering? I thought it should be also done when the connection is in `Relayed` state?

LukasPukenis: From real life experiments both sending and receiving data causes the drainage for a while, thus we care about both of those things happening.

Now about the SessionKeeper - it **is** the best place since it was the one used for keepalives(both batched and non-batched). Batcher is only being used inside of it. For now only because we dealt with direct keepalives that were handled in SessionKeeper.

I would say that SessionKeeper outlived it's purpose and probably warrant a rename at this point though :).

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -15,7 +15,7 @@ use telio_task::{task_exec, BoxAction, Runtime, Task};
 use telio_utils::{
     dual_target, repeated_actions, telio_log_debug, telio_log_warn, DualTarget, RepeatedActions,
 };
-
+use telio_wg::NetworkActivityGetter;
```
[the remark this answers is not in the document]
LukasPukenis: I believe `cargo fmt` should deal with it

crates/telio-batcher/src/batcher.rs
```diff
@@ -37,56 +50,76 @@ where
         Self {
             actions: HashMap::new(),
             notify_add: tokio::sync::Notify::new(),
```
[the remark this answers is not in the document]
LukasPukenis: I personally like not to import stuff that's only used once in the file, it simplifies things since there's less to think about what we import and what we use from the imports. In this case there's no import of `Notify` so it's clear it might be only occasionally used.

Of course, this is more of a matter of discipline, I could use `tokio::sync::Notify` multiple times no problem :)

crates/telio-wg/src/wg.rs
```diff
@@ -786,10 +813,17 @@ impl State {
...
-                        ),
+                        Ok(s) => {
+                            s.update(
+                                new.rx_bytes.unwrap_or_default(),
+                                new.tx_bytes.unwrap_or_default(),
+                            );
+
+                            if let (Some(tx_ts), Some(rx_ts)) = (s.get_tx_ts(), s.get_rx_ts()) {
+                                self.network_activity_ts = Some(TxRxTimestampPair { tx_ts, rx_ts });
+                            }
```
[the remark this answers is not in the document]
LukasPukenis: It's either this, or traversing `self.stats` and retrieving timestamps on each `get_ts()` call. Both should be fine, there shouldn't be too much stats to traverse to be honest, but this is slightly more optimized in terms of micro-optimisation :))

crates/telio-batcher/src/batcher.rs
```diff
@@ -99,11 +132,25 @@ where
...
-            "adding item to batcher with key({:?}), interval({:?}), threshold({:?})",
+            "Adding item to batcher with key({:?}), interval({:?}), threshold({:?})",
             key,
             interval,
             threshold,
         );
+
+        let threshold = {
+            if threshold >= interval {
+                let capped_threshold = interval / 2;
```
Hasan6979, line 143: If the interval is greater than the threshold, why is it capped at half the interval, and not lets say any other value (like 1/4 interval for eg). Just want to understand the reasoning here.

LukasPukenis: No worries, could have commented that, but no real reason except that based on our design sessions and RFCs it seems that the threshold of half the period should work the best as it should batch everything eventually that way but also is not too big to needlessly send data (batcher is currently replacement for wireguard's persistent-keepalives)

It would probably even be better if this check would be more verbose, since if difference is miniscule but still not enough to pass the conditional
it will enter a spiral where:
- A sends data to B
- B get's triggered, sends what's batched to A
- A get's trigger in turn, so sends it's batched things to A
- repeat

crates/telio-wg/src/wg.rs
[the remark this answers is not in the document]
LukasPukenis: Actually I think you're right. Consolidation event will be triggered every second or on some events, same for session keeper's update, meaning to calculate and store it only introduces possibility for a bug to appear. I will inline the code in this function instead.

crates/telio-batcher/src/batcher.rs
LukasPukenis: @LukasPukenis note to self: introduce a minimum threshold, such as 5seconds to avoid spiral of death. However even 5 seconds might be too much if the interval is for example 3 seconds, so the check should be for a difference:

```
if interval - threshold < HARD_THRESHOLD_LIMIT {
  threshold = floor(interval / 2); // as per rfc this value is our guess to be the best value, so we default to that
}
```

crates/telio-batcher/src/batcher.rs
```diff
@@ -37,56 +50,76 @@ where
...
 
-                // TODO: This can be optimized by early breaking and precollecting items beforehand
-                if let Some(closest_entry) = actions.values().min_by_key(|entry| entry.0.deadline) {
-                    tokio::select! {
-                        _ = self.notify_add.notified() => {
-                            // Item was added, we need to immediately emit it
-                        }
-                        _ = sleep_until(closest_entry.0.deadline) => {
-                            // Closest action should now be emitted
+                let active_trigger = self
```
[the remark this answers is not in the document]
LukasPukenis: renamed

crates/telio-batcher/src/batcher.rs
```diff
@@ -37,56 +50,76 @@ where
...
                 }
 
                 return batched_actions;
             } else {
-                let _ = self.notify_add.notified().await;
+                _ = self.notify_add.notified().await;
             }
         }
     }
-
```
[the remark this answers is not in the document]
LukasPukenis: No reason at all and didn't even notice. Very surprised `cargo fmt` didn't pick up on that. Anyway manually fixed it :)

crates/telio-wg/src/wg.rs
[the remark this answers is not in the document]
LukasPukenis: @stalowyjez so the actual reason(I remembered when I went to refactor) was that every entry inside of `stats` is under a mutex(not the hashmap itself but rather the values).

This means every time this is called, a mutex is locked, and in real life this is gonna be called on every consolidation which is not that often to be honest.

However it also means that in case of a bug where batcher would resolve prematurely it will call over and over and over again which would be a problem on itself, however with mutexes it would increase the possible damage significantly due to contention.

For the time being I am preferring a simpler approach so I reworked it as per your suggestion.

crates/telio-batcher/src/batcher.rs
LukasPukenis: @Hasan6979 your question triggered me(pun not intended) to work a bit more on it and I've added minimal checks into the code as well:
- each item has a timestamp it last emitted action which is then checked against
- threshold is capped at maximum of half the interval instead of the previous check with the reason being that it's much more secure that way and easier to verify the correct behaviour.

LukasPukenis:
Note to self: The existing PR has an issue where it will trigger itself once batched as there will be data sent(and then received for icmp pings) later on. This menas triggering mechanism will always work and basically became the same thing as iterating through the jobs every second, meaning every task will start emitting at a period of T-threshold .

The solution is to make the trigger a state machine with a cooldown, meaning that once trigger is in cooldown, it doesn't have an effect. How big is the cooldown is hard to say but it should be:
- longer than 1 second because that's the period of wg consolidation, meaning after batching we can expect the trigger within 1second.
- long enough to incorporate responses back such as ICMP ECHO RESPONSE (for our requests triggered by the batcher).
I would say hardcoding a value such as 30 should be a good start.

-----------
Batcher's threshold throttling should be split into a separate PR for easier review

PR #834, opened 2024-09-26
Add allow_multicast flags from mesh config to starcast integration

ThrasherLT:
### Problem
`allow_multicast` and `peer_allow_multicast` flags are present in mesh config, but not implemented in starcast yet.

### Solution
Add code that removes peers from multicast peer list if we have `allow_multicast` set to false for them and prevent starcast from sending multicast packets to peers who have `peer_allow_multicast` set to false, but still receive multicast packets from those peers.

- [x] README.md is updated

### DIFF
--- a/crates/telio-starcast/src/transport.rs
+++ b/crates/telio-starcast/src/transport.rs
@@ -83,15 +83,19 @@ pub struct Peer {
     pub addr: SocketAddr,
     /// The public key of the peer
     pub public_key: PublicKey,
+    /// Whether our node accepts multicast messages from the peer or not.
+    pub allow_multicast: bool,
+    /// Whether the peer node accepts multicast messages from our node or not.
+    pub peer_allows_multicast: bool,
 }
 
 /// Config for transport component
 /// Contains fields that can change at runtime
 pub enum Config {
     /// Simple transport config, has IP of peer but not port
-    Simple(Vec<(PublicKey, IpAddr)>),
+    Simple(Vec<(PublicKey, IpAddr, bool, bool)>),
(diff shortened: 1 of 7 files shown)


packgron (commented):
+1.0

crates/telio-starcast/src/transport.rs
```diff
@@ -351,23 +394,49 @@ mod tests {
...
     }
 
     impl Scaffold {
         async fn start() -> Self {
             let transport_socket = Arc::new(Self::bind_local_socket().await);
-            let mut peers = Vec::with_capacity(3);
-            for _ in 0..3 {
-                peers.push((SecretKey::gen().public(), Self::bind_local_socket().await));
-            }
+            // Peers with all the different possible meshnet configurations:
```
Hasan6979, line 403: This code be made concise for eg wtih
```
for i in 0..3 {
    let allow_multicast = i % 2 == 0;  // true for i = 0, 2; false for i = 1, 3
    let peer_allow_multicast = i < 2;      // true for i = 0, 1; false for i = 2, 3

    peers.push((public_key, socket, is_first_true, is_second_true));
}
```

Hasan6979 (commented):
Other than that...  +1

crates/telio-starcast/src/transport.rs
ThrasherLT: I think your point is valid, but the way it is now is easier to read and conveys intention a bit better, so I think these are both valid approaches, so I'd rather not change the current code, unless there's somehting wrong with it.

Hasan6979 (approved):

PR #835, opened 2024-09-26
LLT-5535: Use generated events in natlab

mathiaspeters:
This is PR 3/4 to replace bespoke types in natlab with those in the generated python bindings. This PR deals with using strongly typed events, which brings with it the need to start using some related types from the bindings, making the PR a bit on the bigger side. Most of the changes should be conceptually quite simple with mostly changes like `State.Connected` -> `NodeState.CONNECTED` 

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_events.py
+++ b/nat-lab/tests/test_events.py
@@ -5,9 +5,17 @@ import timeouts
 from contextlib import AsyncExitStack
 from helpers import SetupParameters, setup_environment, setup_mesh_nodes, setup_api
 from mesh_api import API
-from telio import AdapterType, PathType, PeerInfo, State, Client
+from telio import AdapterType, Client
 from utils import stun
-from utils.bindings import features_with_endpoint_providers, EndpointProvider
+from utils.bindings import (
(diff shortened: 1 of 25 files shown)


nat-lab/tests/test_events.py
```diff
@@ -18,6 +26,41 @@
 from utils.router import IPStack
 
 
+def node_cmp(left: TelioNode, right: TelioNode):
```
lcruz99, line 29: nit: It sounds like this could be part of the TelioNode class or maybe a fn from /utils?

mathiaspeters: `TelioNode` comes from the generated bindings so we unfortunately can't add it there, and I did consider adding it to `utils/bindings.py` but it's only used in this file so I decided to put it here. If you think it's still a better idea to move it to `utils/bindings.py` I can do that, I don't have any strong opinions on the matter

lcruz99: fair enough, let's leave it here

lcruz99 (approved):
LGTM +2

PR #837, opened 2024-09-27
[LLT-5635] Fix Teliod test

stalowyjez:
### Problem
Teliod test is flaky

### Solution
Removes sleeps and timeouts, ensures that the socket is created before running the next Teliod instance, uses waiting killall instead of kill.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_teliod.py
+++ b/nat-lab/tests/test_teliod.py
@@ -1,4 +1,5 @@
 import asyncio
+import pytest
 from contextlib import AsyncExitStack
 from helpers import setup_connections
 from utils.connection_util import ConnectionTag
@@ -6,6 +7,7 @@ from utils.process.process import ProcessExecError
 
 TELIOD_EXEC_PATH = "/libtelio/dist/linux/release/x86_64/teliod"
 CONFIG_FILE_PATH = "/etc/teliod/config.json"
+SOCKET_FILE_PATH = "/run/teliod.sock"
 
 TELIOD_START_PARAMS = [
     TELIOD_EXEC_PATH,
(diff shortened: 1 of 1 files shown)


nat-lab/tests/test_teliod.py
```diff
@@ -45,41 +52,21 @@ async def test_teliod() -> None:
...
 
         assert not teliod_process.is_executing()
 
         # Run the hello-world command again - this time it should fail
         try:
-            await asyncio.wait_for(
-                connection.create_process(TELIOD_HELLO_WORLD_PARAMS).execute(),
-                1,
-            )
+            await connection.create_process(TELIOD_HELLO_WORLD_PARAMS).execute()
```
jjanowsk, line 70: You forgot `assert False` in the try so if the execute does not fail you won't know about it. That's why it's probably better to use pytest.raises like this:
```
with pytest.raises(ProcessExecError) as err:
    await connection.create_process(TELIOD_HELLO_WORLD_PARAMS).execute()
assert err.stderr == "Error: DaemonIsNotRunning"
```

jjanowsk (commented):
+1 Looks good, just please change those two try/except constructions into pytest.raises

nat-lab/tests/test_teliod.py
```diff
@@ -30,56 +40,31 @@ async def test_teliod() -> None:
         )
 
         # Let the daemon start
-        await asyncio.sleep(1)
+        while not await is_teliod_running(connection):
```
Hasan6979, line 43: What if the daemon does not start due to some reason ? It could get stuck here.

stalowyjez: Yeah, that would be an error and in such case the test timeout will kill it anyway and the test will fail - I was told that this is the current way of doing things i such cases

Hasan6979: Sounds good!

Hasan6979 (approved):
+1

PR #838, opened 2024-09-27
Fix In libtelio compressed files as symlinks are not protected from path traversal

andrisecops:
### Problem
The server protects static routes from path traversal outside the root directory when `follow_symlinks=False` (default). It does this by resolving the requested URL to an absolute path and then checking that path relative to the root. However, these checks are not performed when looking for compressed variants in the FileResponse class, and symbolic links are then automatically followed when performing `Path.stat()` and `Path.open()` to send the file.

### Solution
[WeaknessCWE-61](https://cwe.mitre.org/data/definitions/61.html)
CVE-2024-42367
Update fix `libtelio compressed files as symlinks are not protected from path traversal`


LukasPukenis (commented):
Hello :) Thanks for the Pull Request on Libtelio and for pointing out a CVE!

I see you've updated the `Pipfile.lock` however reflective changes are not inside of `Pipfile` diff. Could you add that as well? This would also make it easy to verify that the Lockfile contains good changes.(`pipenv lock` should produce no changes).

I reproduce different hashes then in the pull request assuming that `aiohttp` was updated to `3.10.2`:

```
$ cd nat-lab && sed -i 's/aiohttp = "==3.9.5"/aiohttp = "==3.10.2"/g' Pipfile && pipenv lock
$ git diff | wc -l
923
```

There might also be an issue in my approach, feel free to correct it.

ThrasherLT:
Hey. I've noticed that 'aiohappyeyeballs' version 2.4.2 is yanked. Was there some specific reason why this version was chosen?

LukasPukenis:
Hello. Due to no responses to our questions we've moved on with the issue ourselves by updating `aiohttp` manually to the patched version in the PR: https://github.com/NordSecurity/libtelio/pull/887.

Thanks for contributing and we hope this is not the last PR :)

PR #839, opened 2024-09-27
LLT-5547: Use generated adapter type

mathiaspeters:
This is the final PR migrating from bespoke types in natlab to types from generated python bindings. This PR consists of almost only changing from the old `AdapterType` (defined in `telio.py`)  to `TelioAdapterType` which is in the generated bindings. 

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -4,14 +4,12 @@ import glob
 import os
 import platform
 import re
-import uniffi.telio_bindings as libtelio
 import uuid
 import warnings
 from collections import Counter
 from config import DERP_SERVERS
 from contextlib import asynccontextmanager
 from datetime import datetime
-from enum import Enum
 from mesh_api import Node, start_tcpdump, stop_tcpdump
 from typing import AsyncIterator, List, Optional, Set
 from uniffi.libtelio_proxy import LibtelioProxy, ProxyConnectionError
@@ -29,6 +27,7 @@ from utils.bindings import (
     NodeState,
     RelayState,
     LinkState,
+    TelioAdapterType,
 )
 from utils.command_grepper import CommandGrepper
 from utils.connection import Connection, DockerConnection, TargetOS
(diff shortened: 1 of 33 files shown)


nat-lab/tests/telio.py
```diff
@@ -1158,11 +1142,3 @@ async def collect_core_dumps(self):
                     f" {container_name}:{file_path} {core_dump_destination}"
                 )
                 os.system(cmd)
-
-
-def generate_secret_key() -> str:
```
tomasz-grz, line 1163: Why were these removed?

nat-lab/tests/test_mesh_plus_vpn.py
```diff
@@ -268,37 +269,39 @@ async def test_mesh_plus_vpn_both_peers(
...
             ConnectionTag.WINDOWS_VM_1,
-            AdapterType.WireguardGo,
+            TelioAdapterType.WIREGUARD_GO_TUN,
             "10.0.254.7",
             marks=pytest.mark.windows,
         ),
         pytest.param(
             ConnectionTag.MAC_VM,
-            AdapterType.Default,
+            None,
```
tomasz-grz, line 295: What happens if the adapter is set to `None`? Is the default one used for each platform? AFAIK only BoringTun works on macOS

tomasz-grz (commented):
Nice work +1.0, just a few questions from my side

nat-lab/tests/telio.py
mathiaspeters: All this function is doing is calling a function in the bindings, so instead of having this function, the call sites can just call that function from the bindings directly

nat-lab/tests/test_mesh_plus_vpn.py
mathiaspeters: The logic for that is handled here: https://github.com/NordSecurity/libtelio/pull/839/files#diff-f1122f50402db2f1613c75eb1db5aed7bc4c9ed907b99f1bf4e5010345bfe6cdR793 and there is a comment for the `SetupParams` class here https://github.com/NordSecurity/libtelio/pull/839/files#diff-9fa2bd5404cf7ee9aeb4e031cc531e8d22f729d73f6891181350bfab552fa181R43 that I updated. Basically, `None` tells natlab to use the default adapter for each platform which is `WireguardGo` on windows and `BoringTun` on all other platforms

nat-lab/tests/telio.py
```diff
@@ -813,6 +790,13 @@ def get_features(self) -> Features:
         assert self._telio_features
         return self._telio_features
 
+    def get_adapter_type(self) -> TelioAdapterType:
+        if self._adapter_type is not None:
+            return self._adapter_type
+        if isinstance(self.get_router(), WindowsRouter):
+            return TelioAdapterType.WIREGUARD_GO_TUN
+        return TelioAdapterType.BORING_TUN
```
stalowyjez, line 798: I guess these two are written BoringTun and WireGuard, so why one of them is written with underscore and the other one without?

stalowyjez (commented):
Looks good, +1

nat-lab/tests/telio.py
mathiaspeters: If you check in the `.udl` it's `WireguardGoTun` and `BoringTun`, hence `WIREGUARD_GO_TUN` and `BORING_TUN`

nat-lab/tests/test_mesh_plus_vpn.py
tomasz-grz: In that sense the `AdapterType.Default` pattern was a little bit more descriptive of what is actually going on.
What do you think about naming the parameter `override_adapter_type` instead then?

tomasz-grz (dismissed):

tomasz-grz (approved):
Nice 🥇 +1

PR #846, opened 2024-10-03
Fix bad adapter type in natlab

mathiaspeters:
The old AdapterType was removed in favor of the generated TelioAdapterType, but between the last commit on that migration PR being pushed and the pipeline being green so it could be merged, another PR added a new test relying on the old AdapterType. because of that, the PRs were incompatible but had no code conflicts, and got the repo into a bad state

- [x] README.md is updated


LukasPukenis (approved):
+2

PR #847, opened 2024-10-03
Move sync blocking code to thread

gytsto:
### Problem
Some libtelio code is synchronous blocking, which might block main thread.

### Solution
Wrap all libtelio_proxy functions with asyncio.to_thread so those calls wouldn’t block async threads.

- [x] README.md is updated


gytsto:
> I would suggest looking into if this can be implemented with a decorator, like `serialize_error` is done in `libtelio_remote.py`

Great suggestion. Check out the implementation please

mathiaspeters (approved):
Looks great! +2

PR #849, opened 2024-10-03
Reset exponential backoff on succesfull endpoint verification

Jauler:
### Description
It is not guaranteed that `wg_controller` will use every published endpoint, which means, that existing [exponential backoff reset](https://github.com/NordSecurity/libtelio/blob/main/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs#L758) may not occur. But in any case, if endpoint has passed the validation - we may reset exponential backoff already, as we know, that required information has flown succesfully between two meshnet nodes.

- [x] README.md is updated

### DIFF
--- a/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
+++ b/crates/telio-traversal/src/connectivity_check/cross_ping_check.rs
@@ -820,6 +820,10 @@ impl<E: Backoff> EndpointConnectivityCheckState<E> {
                         };
 
                         do_state_transition!(self, Event::Publish);
+
+                        // Reset exponential backoff on succesfull endpoint verification
+                        self.exponential_backoff.reset();
+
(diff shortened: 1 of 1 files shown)


LukasPukenis:
+2 nice one

LukasPukenis (dismissed):

gytsto:
Maybe it would be better to add backoff control to `do_transition` macro? Other than that, lgtm +1

Jauler:
> Maybe it would be better to add backoff control to `do_state_transition` macro? Other than that, lgtm +1

Well there is only one place where transition to published occurs, and it already have event sending, state update (e.g. storing validated endpoints), which IMHO belongs to the same category, therefore either all of them would have to be moved or none. And I guess the macro would expand a lot then :thinking:

gytsto (approved):

PR #850, opened 2024-10-03
traversal: Set random ports between ephemeral range

lcruz99:
### Problem
UPNP external port range was being set at IGD crate criteria, which is RNG assigned between 32768..65535.
IANA standardized this range on 49152..65535 and this is what should be used. (see: https://www.rfc-editor.org/rfc/rfc6335.html#section-6)

### Solution
Leave the port random generation to telio, using 49152..65535.

- [x] README.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/upnp.rs
+++ b/crates/telio-traversal/src/endpoint_providers/upnp.rs
@@ -34,18 +34,24 @@ use mockall::automock;
 const MAX_SUPPORTED_PACKET_SIZE: usize = 1500;
 const GET_INTERFACE_TIMEOUT_S: Duration = Duration::from_secs(2);
 
+const EPHEMERAL_PORT_RANGE: std::ops::RangeInclusive<u16> = 49152..=65535;
+
+const DEFAULT_LEASE_DURATION_S: u32 = 600;
+
 type Result<T> = std::result::Result<T, Error>;
 
 #[cfg_attr(test, automock)]
 #[async_trait]
 pub trait UpnpEpCommands: Send + Default + 'static {
     async fn check_endpoint_routes(&self, proxy_port: u16, wg_port: u16) -> Result<bool>;
-    async fn add_any_endpoint_routes(
+    async fn add_endpoint_routes(
         &self,
         ip_addr: Ipv4Addr,
-        proxy_port: u16,
-        wg_port: u16,
-    ) -> Result<(u16, u16)>;
+        proxy_port_internal: u16,
+        proxy_port_external: u16,
+        wg_port_internal: u16,
+        wg_port_external: u16,
+    ) -> Result<()>;
     async fn extend_endpoint_duration(
         &self,
         ip_addr: Ipv4Addr,
@@ -90,7 +96,7 @@ impl UpnpEpCommands for IgdGateway {
         while let Ok(resp) = gw.get_generic_port_mapping_entry(i).await {
             i += 1;
 
-            if resp.lease_duration < 10 * 60 {
+            if resp.lease_duration < DEFAULT_LEASE_DURATION_S {
                 extend_timeout = true;
             }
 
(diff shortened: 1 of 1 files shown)


LukasPukenis:
Would be great to link relevant RFC part for the record :)

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -34,18 +34,25 @@ use mockall::automock;
 const MAX_SUPPORTED_PACKET_SIZE: usize = 1500;
 const GET_INTERFACE_TIMEOUT_S: Duration = Duration::from_secs(2);
 
+const MIN_EPHEMERAL_PORT_RANGE: u16 = 49152;
+const MAX_EPHEMERAL_PORT_RANGE: u16 = 65535;
+
+const DEFAULT_LEASE_DURATION_S: u32 = 600;
```
LukasPukenis, line 40: It's nice that it's postfixed with `_S` but I think it could benefit from being `time::Duration`, this way it would be more type-safe

LukasPukenis (commented):
Looks legitimate +1

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: fair enough, will change

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -545,9 +553,11 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> State<Wg, I, E> {
     }
 
     fn create_random_endpoint_ports(&mut self) {
-        self.wg_port_mapping.external = rand::thread_rng().gen_range(1024..49151);
+        self.wg_port_mapping.external =
+            rand::thread_rng().gen_range(MIN_EPHEMERAL_PORT_RANGE..MAX_EPHEMERAL_PORT_RANGE);
```
stalowyjez, line 557: Hmm, maybe this could be just a const range?

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -747,17 +756,17 @@ mod tests {
     }
 
     fn check_endpoint_route(proxy: u16, wg: u16) -> Result<bool> {
-        if ENDPOINT.lock().wg.port() == proxy && ENDPOINT.lock().udp.port() == wg {
+        if ENDPOINT.lock().wg.port() == wg && ENDPOINT.lock().udp.port() == proxy {
```
stalowyjez, line 759: So there was a bug here all this time? o_O

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -876,8 +887,18 @@ mod tests {
...
+
+        assert_eq!(udp_int_port, 55555);
+        assert_eq!(wg_int_port, 55345);
+
+        let epc = upnp.get_endpoint_candidate().await.unwrap();
+        let epc_wg_port = epc.wg.port();
+        let epc_udp_port = epc.udp.port();
+
+        assert!(MIN_EPHEMERAL_PORT_RANGE < epc_wg_port && MAX_EPHEMERAL_PORT_RANGE > epc_wg_port);
+        assert!(MIN_EPHEMERAL_PORT_RANGE < epc_udp_port && MAX_EPHEMERAL_PORT_RANGE > epc_udp_port);
```
stalowyjez, line 901: Isn't low range bound inclusive?

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: yes, went unnoticed all this time

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: Yes it could, however I don't think there is any benefit, if not for worse in case of readability/flexibility in accessing each value individually..

stalowyjez: Is `MIN_EPHEMERAL_PORT_RANGE..MAX_EPHEMERAL_PORT_RANGE` more readable than simply `EPHEMERAL_PORT_RANGE`? I don't think so, but this is a nitpick anyway.

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: changed constqnt to be of inclusive range type.

stalowyjez (approved):
Looks good, +1

PR #853, opened 2024-10-07
Core API suport for teliod

Hasan6979:
### Problem
To run a meshnet App for OpenWRT a thin cli layer on top of telio library is to be designed. This wrapper would call the Core API which is normally done by the Apps

### Solution
Add support to call Core API to register and update device, get meshmap.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/README.md
+++ b/clis/teliod/README.md
@@ -23,6 +23,9 @@ There is a command for running the daemon:
      - `trace`
      - `off`
    - `log_file_path` - a path to the daemon's logging file (relative paths are starting in Teliod's working directory)
+   - `authentication_token` - Token from Nord VPN account to authenticate API calls
+   - `app_user_uid` - A unique number for each user of the application
+   - `interface_name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
 
 And following cli commands:
  - `teliod hello-world <NAME>` - simple command for testing purposes, logs "Hello NAME!", used to ensure daemon proper startup, should be erased in the future and replaced by some more serious one
--- a/clis/teliod/src/comms.rs
+++ b/clis/teliod/src/comms.rs
@@ -113,6 +113,17 @@ impl DaemonSocket {
     }
 }
 
+impl Drop for DaemonSocket {
+    fn drop(&mut self) {
+        // Normally this should be cleaned up, but in some cases it is not
+        // The ipc path here should be valid, as the daemon socket is created
+        // based on whether the path is valid or not.
+        if let Ok(socket_path) = DaemonSocket::get_ipc_socket_path() {
+            let _ = fs::remove_file(&socket_path);
+        }
+    }
+}
+
 /// Struct for tracking a single IPC connection and keeping the stream
 /// Mostly needed for keeping the stream while the server is generating a reply.
 pub struct DaemonConnection {
--- a/clis/teliod/src/config.rs
+++ b/clis/teliod/src/config.rs
@@ -1,9 +1,12 @@
 use std::{num::NonZeroU64, path::PathBuf, str::FromStr};
 
-use serde::{de, Deserialize, Deserializer};
+use serde::{de, Deserialize, Deserializer, Serialize};
 use smart_default::SmartDefault;
-use tracing::level_filters::LevelFilter;
+use std::fs;
+use tracing::{debug, info, level_filters::LevelFilter};
 use uuid::Uuid;
+
+use telio::crypto::SecretKey;
 
 #[derive(PartialEq, Eq, Clone, Copy, Debug, SmartDefault)]
 #[repr(transparent)]
@@ -50,11 +53,40 @@ fn reconnect_after_expiry_default() -> Percentage {
     Percentage(90)
 }
 
+#[derive(Serialize, Deserialize, Debug, Default)]
+pub struct DeviceIdentity {
+    pub hw_identifier: String,
+    pub private_key: SecretKey,
+    pub machine_identifier: String,
+}
+
+pub fn generate_hw_identifier(private_key: SecretKey) -> String {
+    // Generate hw_identifier
+    let public_key = private_key.public();
+    debug!("Generating hw identifier");
+    format!("{}.{}", public_key, "teliod")
+}
+
+impl DeviceIdentity {
+    pub fn from_file(identity_path: &PathBuf) -> Option<DeviceIdentity> {
+        info!("Fetching config");
+
+        if let Ok(file) = fs::File::open(identity_path) {
+            debug!("found existing config.");
+            if let Ok(c) = serde_json::from_reader(file) {
+                return Some(c);
+            }
+        }
+        None
+    }
+}
+
 #[derive(PartialEq, Eq, Deserialize, Debug)]
 pub struct TeliodDaemonConfig {
     #[serde(deserialize_with = "deserialize_log_level")]
     pub log_level: LevelFilter,
     pub log_file_path: String,
+    pub interface_name: String,
 
     pub app_user_uid: Uuid,
 
@@ -124,6 +156,7 @@ mod tests {
             log_level: LevelFilter::INFO,
             log_file_path: "test.log".to_owned(),
             app_user_uid: Uuid::from_str("2ba97921-38d7-4736-9d47-261cf3e5c223").unwrap(),
+            interface_name: "utun10".to_owned(),
             authentication_token:
                 "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa".to_owned(),
             http_certificate_file_path: None,
@@ -140,6 +173,7 @@ mod tests {
             "log_level": "Info",
             "log_file_path": "test.log",
             "app_user_uid": "2ba97921-38d7-4736-9d47-261cf3e5c223",
+            "interface_name": "utun10",
             "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
             }"#;
 
@@ -151,6 +185,7 @@ mod tests {
                 "log_level": "Info",
                 "log_file_path": "test.log",
                 "app_user_uid": "2ba97921-38d7-4736-9d47-261cf3e5c223",
+                "interface_name": "utun10",
                 "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
                 "mqtt": {}
             }"#;
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -1,0 +1,372 @@
+use crate::config::generate_hw_identifier;
+use crate::DeviceIdentity;
+use base64::{prelude::*, DecodeError};
+use dirs;
+use reqwest::{header, Client, StatusCode};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::fs::{create_dir_all, OpenOptions};
+use std::os::unix::fs::OpenOptionsExt;
+use std::sync::Arc;
+use telio::crypto::SecretKey;
+use telio::telio_utils::exponential_backoff::{
+    Backoff, Error as BackoffError, ExponentialBackoff, ExponentialBackoffBounds,
+};
+use telio::{crypto::PublicKey, telio_model::config::Config as MeshMap};
+use thiserror::Error;
+use tokio::time::Duration;
+use tracing::{debug, error, info, warn};
+
+const API_BASE: &str = "https://api.nordvpn.com/v1";
+#[cfg(target_os = "macos")]
+const OS_NAME: &str = "macos";
+#[cfg(target_os = "linux")]
+const OS_NAME: &str = "linux";
+
+type MachineIdentifier = String;
+const MAX_RETRIES: usize = 10;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    ReqwestError(#[from] reqwest::Error),
+    #[error(transparent)]
+    DeserializeError(#[from] serde_json::Error),
+    #[error("Unable to fetch Machine Identifier: {0}")]
+    FetchingIdentifierError(StatusCode),
+    #[error("Unable to register due to Error: {0}")]
+    PeerRegisteringError(StatusCode),
+    #[error(transparent)]
+    DecodeError(#[from] DecodeError),
+    #[error("Invalid response recieved from server")]
+    InvalidResponse,
+    #[error("Unable to update machine due to Error: {0}")]
+    UpdateMachineError(StatusCode),
+    #[error("Max retries exceeded for connection: {0}")]
+    ExpBackoffTimeout(String),
+    #[error("No local data directory found")]
+    NoDataLocalDir,
+    #[error("Device not registered with server")]
+    DeviceNotFound,
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error(transparent)]
+    BackoffBoundsError(#[from] BackoffError),
+}
+
+#[derive(Default, Serialize, Deserialize)]
+struct MeshConfig {
+    public_key: PublicKey,
+    hardware_identifier: String,
+    os: String,
+    os_version: String,
+    device_type: String,
+    traffic_routing_supported: bool,
+}
+
+fn build_backoff() -> Result<ExponentialBackoff, Error> {
+    let backoff_bounds = ExponentialBackoffBounds {
+        initial: Duration::from_secs(1),
+        maximal: Some(Duration::from_secs(180)),
+    };
+    Ok(ExponentialBackoff::new(backoff_bounds)?)
+}
+
+pub async fn init_with_api(
+    auth_token: &str,
+    interface_name: &str,
+) -> Result<DeviceIdentity, Error> {
+    let mut identity_path = dirs::data_local_dir().ok_or(Error::NoDataLocalDir)?;
+    identity_path.push("teliod");
+    if !identity_path.exists() {
+        let _ = create_dir_all(&identity_path);
+    }
+    identity_path.push(&format!("{interface_name}.json"));
+
+    let mut device_identity = match DeviceIdentity::from_file(&identity_path) {
+        Some(identity) => identity,
+        None => {
+            let private_key = SecretKey::gen();
+            let hw_identifier = generate_hw_identifier(private_key);
+
+            let machine_identifier =
+                match fetch_identifier_with_exp_backoff(auth_token, private_key.public()).await {
+                    Ok(id) => id,
+                    Err(e) => match e {
+                        Error::DeviceNotFound => {
+                            info!("Unable to load identifier due to {e}. Registering ...");
+                            register_machine_with_exp_backoff(
+                                &hw_identifier,
+                                private_key.public(),
+                                auth_token,
+                            )
+                            .await?
+                        }
+                        _ => return Err(e),
+                    },
+                };
+
+            DeviceIdentity {
+                private_key,
+                hw_identifier,
+                machine_identifier,
+            }
+        }
+    };
+
+    if let Err(e) = update_machine_with_exp_backoff(auth_token, &device_identity).await {
+        if let Error::UpdateMachineError(status) = e {
+            if status == StatusCode::NOT_FOUND {
+                debug!("Unable to update. Registering machine ...");
+                device_identity.machine_identifier = register_machine_with_exp_backoff(
+                    &device_identity.hw_identifier,
+                    device_identity.private_key.public(),
+                    auth_token,
+                )
+                .await?;
+            } else {
+                // exit daemon
+                return Err(Error::UpdateMachineError(status));
+            }
+        } else {
+            return Err(e);
+        }
+    }
+
+    let mut options = OpenOptions::new();
+    // User has read/write but others have none
+    options.mode(0o600);
+    let mut file = options.create(true).write(true).open(identity_path)?;
+    serde_json::to_writer(&mut file, &device_identity)?;
+    Ok(device_identity)
+}
+
+async fn update_machine_with_exp_backoff(
+    auth_token: &str,
+    device_identity: &DeviceIdentity,
+) -> Result<(), Error> {
+    let mut backoff = build_backoff()?;
+    let mut retries = 0;
+
+    loop {
+        match update_machine(&device_identity, auth_token).await {
+            Ok(_) => return Ok(()),
+            Err(e) => {
+                if let Error::UpdateMachineError(_) = e {
+                    return Err(e);
+                }
+                let _ = wait_with_backoff_delay(&mut backoff, &mut retries, "update machine", e)
+                    .await?;
+            }
+        }
+    }
+}
+
+async fn fetch_identifier_with_exp_backoff(
+    auth_token: &str,
+    public_key: PublicKey,
+) -> Result<MachineIdentifier, Error> {
+    let mut backoff = build_backoff()?;
+    let mut retries = 0;
+
+    loop {
+        match fetch_identifier_from_api(auth_token, public_key).await {
+            Ok(id) => return Ok(id),
+            Err(e) => match e {
+                Error::FetchingIdentifierError(_) | Error::DeviceNotFound => return Err(e),
+                _ => {
+                    let _ =
+                        wait_with_backoff_delay(&mut backoff, &mut retries, "fetch identifier", e)
+                            .await?;
+                }
+            },
+        }
+    }
+}
+
+async fn register_machine_with_exp_backoff(
+    hw_identifier: &str,
+    public_key: PublicKey,
+    auth_token: &str,
+) -> Result<MachineIdentifier, Error> {
+    let mut backoff = build_backoff()?;
+    let mut retries = 0;
+
+    loop {
+        match register_machine(hw_identifier, public_key, auth_token).await {
+            Ok(id) => return Ok(id),
+            Err(e) => match e {
+                Error::PeerRegisteringError(_) => return Err(e),
+                _ => {
+                    let _ =
+                        wait_with_backoff_delay(&mut backoff, &mut retries, "register machine", e)
+                            .await?;
+                }
+            },
+        }
+    }
+}
+
+async fn wait_with_backoff_delay(
+    backoff: &mut ExponentialBackoff,
+    retries: &mut usize,
+    action: &str,
+    error: Error,
+) -> Result<(), Error> {
+    warn!(
+        "Failed to {action} due to {error}, will wait for {:?} and retry",
+        backoff.get_backoff()
+    );
+    tokio::time::sleep(backoff.get_backoff()).await;
+    if *retries > MAX_RETRIES {
+        return Err(Error::ExpBackoffTimeout(action.to_string()));
+    }
+    backoff.next_backoff();
+    *retries += 1;
+
+    Ok(())
+}
+
+async fn fetch_identifier_from_api(
+    auth_token: &str,
+    public_key: PublicKey,
+) -> Result<MachineIdentifier, Error> {
+    debug!("Fetching machine identifier");
+    let client = Client::new();
+    let response = client
+        .get(&format!("{}/meshnet/machines", API_BASE))
+        .header(
+            header::AUTHORIZATION,
+            format!("Bearer token:{}", auth_token),
+        )
+        .header(header::ACCEPT, "application/json")
+        .send()
+        .await?;
+
+    let status = response.status();
+    if status == StatusCode::OK {
+        let json_data: Value = serde_json::from_str(&response.text().await?)?;
+
+        for node in json_data.as_array().ok_or(Error::InvalidResponse)? {
+            // Get the public_key and identifier from each item
+            if let Some(recvd_pk) = node.get("public_key").and_then(|k| k.as_str()) {
+                if BASE64_STANDARD
+                    .decode(recvd_pk)?
+                    .as_slice()
+                    .cmp(&public_key.0)
+                    .is_eq()
+                {
+                    if let Some(identifier) = node.get("identifier").and_then(|i| i.as_str()) {
+                        info!("Match found! Identifier: {}", identifier);
+                        return Ok(identifier.to_owned());
+                    }
+                }
+            }
+        }
+        return Err(Error::DeviceNotFound);
+    }
+
+    error!(
+        "Unable to fetch identifier due to {:?}",
+        response.text().await
+    );
+    Err(Error::FetchingIdentifierError(status))
+}
+
+async fn update_machine(device_identity: &DeviceIdentity, auth_token: &str) -> Result<(), Error> {
+    debug!("Updating machine");
+    let client = Client::new();
+    let status = client
+        .patch(&format!(
+            "{}/meshnet/machines/{}",
+            API_BASE, device_identity.machine_identifier
+        ))
+        .header(
+            header::AUTHORIZATION,
+            format!("Bearer token:{}", auth_token),
+        )
+        .header(header::CONTENT_TYPE, "application/json")
+        .header(header::ACCEPT, "application/json")
+        .json(&MeshConfig {
+            public_key: device_identity.private_key.public(),
+            hardware_identifier: device_identity.hw_identifier.clone(),
+            os: OS_NAME.to_owned(),
+            os_version: "teliod".to_owned(),
+            device_type: "other".to_owned(),
+            traffic_routing_supported: true,
+        })
+        .send()
+        .await?
+        .status();
+
+    if status == StatusCode::OK {
+        return Ok(());
+    }
+    Err(Error::UpdateMachineError(status))
+}
+
+pub async fn get_meshmap(
+    device_identity: Arc<DeviceIdentity>,
+    auth_token: Arc<String>,
+) -> Result<MeshMap, Error> {
+    debug!("Getting meshmap");
+    let client = Client::new();
+    Ok(serde_json::from_str(
+        &client
+            .get(&format!(
+                "{}/meshnet/machines/{}/map",
+                API_BASE, device_identity.machine_identifier
+            ))
+            .header(
+                header::AUTHORIZATION,
+                format!("Bearer token:{}", auth_token),
+            )
+            .header(header::ACCEPT, "application/json")
+            .send()
+            .await?
+            .text()
+            .await?,
+    )?)
+}
+
+async fn register_machine(
+    hw_identifier: &str,
+    public_key: PublicKey,
+    auth_token: &str,
+) -> Result<MachineIdentifier, Error> {
+    info!("Registering machine");
+    let client = Client::new();
+    let response = client
+        .post(&format!("{}/meshnet/machines", API_BASE))
+        .header(
+            header::AUTHORIZATION,
+            format!("Bearer token:{}", auth_token),
+        )
+        .header(header::CONTENT_TYPE, "application/json")
+        .header(header::ACCEPT, "application/json")
+        .json(&MeshConfig {
+            public_key,
+            hardware_identifier: hw_identifier.to_owned(),
+            os: OS_NAME.to_owned(),
+            os_version: "teliod".to_owned(),
+            device_type: "other".to_owned(),
+            traffic_routing_supported: true,
+        })
+        .send()
+        .await?;
+
+    let status = response.status();
+    // Save the machine identifier received from API
+    if status == StatusCode::CREATED {
+        let response: Value = serde_json::from_str(&response.text().await?)?;
+        if let Some(machine_identifier) = response.get("identifier").and_then(|i| i.as_str()) {
+            info!("Machine Registered!");
+            return Ok(machine_identifier.to_owned());
+        } else {
+            return Err(Error::InvalidResponse);
+        }
+    }
+
+    error!("Unable to register due to {:?}", response.text().await);
+    Err(Error::PeerRegisteringError(status))
+}
--- a/clis/teliod/src/main.rs
+++ b/clis/teliod/src/main.rs
@@ -8,21 +8,29 @@ use nix::sys::signal::Signal;
 use serde::{Deserialize, Serialize};
 use serde_json::error::Error as SerdeJsonError;
 use signal_hook_tokio::Signals;
+use std::time::Duration;
 use std::{
     fs::{self, File},
     sync::Arc,
 };
+use telio::{
+    crypto::SecretKey,
+    device::{Device, DeviceConfig, Error as DeviceError},
+    telio_model::{config::Config as MeshMap, features::Features},
+    telio_utils::select,
+    telio_wg::AdapterType,
+};
 use thiserror::Error as ThisError;
-use tokio::task::JoinError;
-use tracing::{debug, error, info, warn};
-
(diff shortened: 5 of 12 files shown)


clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,162 @@
...
+const OS_NAME: &str = "windows";
+#[cfg(target_os = "macos")]
+const OS_NAME: &str = "macos";
+#[cfg(target_os = "linux")]
+const OS_NAME: &str = "linux";
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Rewquest(#[from] reqwest::Error),
```
gytsto, line 21: Reqwest instead of Rewquest

Hasan6979: ✅

matislovas:
Other than that ... +1

clis/teliod/src/main.rs
```diff
@@ -153,8 +331,27 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
...
-    let telio_task_handle = tokio::task::spawn_blocking(telio_task);
+    // Tx is unused here, but this channel can be used to communicate with the
+    // telio task
+    let (tx, rx) = mpsc::channel(10);
+
+    // TODO: This is temporary to be removed later on when we have proper integration
+    // tests with core API. This is to not look for tokens in a test environment
+    // right now as the values are dummy and program will not run as it expects
+    // real tokens.
+    if !config.authentication_token.eq("abcd1234") {
```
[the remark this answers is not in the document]
Hasan6979: Running with an invalid token results in the api calls failing and the subsequent integration test failing as well.

clis/teliod/src/main.rs
```diff
@@ -129,19 +192,134 @@ impl CommandListener {
...
+    Ok(())
 }
 
-async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError> {
+fn update_meshmap(tokens: &Tokens, telio: &Device) -> Result<(), TeliodError> {
+    let meshmap: MeshMap = serde_json::from_str(&get_meshmap(tokens)?)?;
+    trace!("Meshmap {:#?}", meshmap);
+    telio.set_config(&Some(meshmap))?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
Hasan6979: But to wait on that channel we would still have to do `blocking_recv` which comes back to the same problem. Doesn't it ? Or do you have something else in mind ?

clis/teliod/src/main.rs
```diff
@@ -129,19 +192,134 @@ impl CommandListener {
...
+    let mut tokens = fetch_tokens(config).await;
+
+    let mut retries = 0;
+    loop {
+        let status = update_machine(&tokens).await?;
+        match status {
+            StatusCode::OK => break,
+            StatusCode::FORBIDDEN | StatusCode::UNAUTHORIZED | StatusCode::BAD_REQUEST => {
+                // exit daemon
+                panic!("{:?} response recieved. Exiting", status);
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -129,19 +192,134 @@ impl CommandListener {
...
 
-async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError> {
+fn update_meshmap(tokens: &Tokens, telio: &Device) -> Result<(), TeliodError> {
+    let meshmap: MeshMap = serde_json::from_str(&get_meshmap(tokens)?)?;
+    trace!("Meshmap {:#?}", meshmap);
+    telio.set_config(&Some(meshmap))?;
+    Ok(())
+}
+
+async fn fetch_tokens(config: &mut TeliodDaemonConfig) -> Tokens {
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -153,8 +331,27 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
...
+    let (tx, rx) = mpsc::channel(10);
+
+    // TODO: This is temporary to be removed later on when we have proper integration
+    // tests with core API. This is to not look for tokens in a test environment
+    // right now as the values are dummy and program will not run as it expects
+    // real tokens.
+    if !config.authentication_token.eq("abcd1234") {
+        init_api(&mut config, config_path).await?;
 
+        let tokens = Tokens {
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -129,19 +194,169 @@ impl CommandListener {
     }
 }
 
+async fn init_api(config: &mut TeliodDaemonConfig, config_path: String) -> Result<(), TeliodError> {
+    let mut client_config = ClientConfig::new(config).await?;
```
stalowyjez, line 198: I think it's a bit strange to hide registration call inside some function or method - maybe it would be nicer to call it directly here? And you could skip UPDATE on successful registration.

clis/teliod/src/main.rs
```diff
@@ -129,19 +194,169 @@ impl CommandListener {
...
+        start_telio(&mut telio, client_config.private_key)?;
+        if let Err(e) = update_meshmap(&client_config, &telio) {
+            error!("Unable to set meshmap due to {e}");
+        }
+
+        while let Some(cmd) = rx_channel.blocking_recv() {
+            info!("Got command {:?}", cmd);
+            match cmd {
+                TelioTaskCmd::UpdateMeshmap => {
+                    if let Err(e) = update_meshmap(&client_config, &telio) {
```
stalowyjez, line 272: As I said before, I think Telio task should block just on the commands and Telio API, but maybe it's not that but, I don't insist on moving it outside (but I think it would improve the quality of this code)

stalowyjez (commented):
It looks quite good, just a few minor comments

clis/teliod/src/main.rs
```diff
@@ -129,19 +194,169 @@ impl CommandListener {
...
+        let machine_identifier = if let Some(machine_id) = &config.machine_identifier {
+            machine_id.to_string()
+        } else {
+            if let Ok(id) = load_identifier_from_api(&config.authentication_token, public_key).await
+            {
+                id
+            } else {
+                debug!("Machine not yet registered");
+                // If registering fails. Close the daemon
+                register_machine(&hw_identifier, public_key, &config.authentication_token).await?
```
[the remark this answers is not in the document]
Hasan6979: You are right. I think registering here can be skipped completely and after initializing tokens go to updating the machine. It would return error as identifier is invalid, and simply register the machine then.

clis/teliod/src/main.rs
Hasan6979: If we use async to get_meshmap we would still have to use block_recv to get the meshmap part, which would in the end come back to the same problem wouldn't it ?

stalowyjez: Telio task would wait on the whole channel and could just use the received meshmap when it comes, so if any other thing  comes from the channel when meshmap is being downloaded, it could just use it

clis/teliod/src/main.rs
```diff
@@ -129,19 +196,185 @@ impl CommandListener {
...
+        }
+    });
+}
+
+fn update_meshmap(map: &str, telio: &Device) -> Result<(), TeliodError> {
+    let meshmap: MeshMap = serde_json::from_str(map)?;
+    trace!("Meshmap {:#?}", meshmap);
+    telio.set_config(&Some(meshmap))?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -129,19 +196,185 @@ impl CommandListener {
...
+
+#[allow(mpsc_blocking_send)]
+fn get_meshmap(client_config: Arc<ClientConfig>, tx: mpsc::Sender<TelioTaskCmd>) {
+    let config_clone = client_config.clone();
+    tokio::spawn(async move {
+        let result = get_meshmap_from_server(config_clone).await;
+        match result {
+            Ok(map) => {
+                if let Err(e) = tx.send(TelioTaskCmd::UpdateMeshmap(map)).await {
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -129,19 +196,185 @@ impl CommandListener {
...
+        } else {
+            debug!("Generating hw identifier");
+            format!("{}.{}", public_key, "openWRT")
+        };
+
+        // Copy or generate machine_identifier
+        let machine_identifier = if let Some(machine_id) = &config.machine_identifier {
+            machine_id.to_string()
+        } else {
+            if let Ok(id) = load_identifier_from_api(&config.authentication_token, public_key).await
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -129,19 +194,169 @@ impl CommandListener {
...
+                if retries < MAX_RETRIES {
+                    let backoff_time = BASE_DELAY_MS * 2_u64.pow(retries);
+                    let sleep_time = min(backoff_time, MAX_BACKOFF_TIME);
+                    debug!("Retrying after {} ms...", sleep_time);
+                    tokio::time::sleep(Duration::from_millis(sleep_time)).await;
+                    retries += 1;
+                } else {
+                    // If max retries exceeded, exit daemon
+                    return Err(TeliodError::UpdateMachineTimeoutError);
+                    // panic!("Max retries reached. Exiting.");
```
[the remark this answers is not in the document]
Hasan6979: ✅

nat-lab/data/teliod/config.json
```diff
@@ -1,4 +1,5 @@
 {
     "log_level": "trace",
-    "log_file_path": "teliod_natlab.log"
+    "log_file_path": "teliod_natlab.log",
+    "authentication_token": "abcd1234"
```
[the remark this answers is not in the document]
Hasan6979: ✅

clis/teliod/src/main.rs
```diff
@@ -129,19 +207,193 @@ impl CommandListener {
...
+                }
+            }
+            Err(e) => error!("Getting meshmap failed due to {e}"),
+        }
+    });
 }
 
-async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError> {
+fn start_telio(telio: &mut Device, private_key: SecretKey) -> Result<(), DeviceError> {
+    if !telio.is_running() {
```
stalowyjez, line 334: Can Telio be running at this point? It might look reasonable locally, but this function is used only once when Telio was just created, so I guess you don't need to check that (I don't think this function is particularly helpful tbh)

clis/teliod/src/main.rs
```diff
@@ -129,19 +207,193 @@ impl CommandListener {
...
+            ..Default::default()
+        };
+
+        telio.start(&device_config)?;
+
+        info!(
+            "started telio with {:?}:{}...",
+            AdapterType::BoringTun,
+            private_key
+        );
```
stalowyjez, line 348: I don't think printing private key in logs is the best practice. As the adapter will be always the same, do we really need this log?

clis/teliod/src/main.rs
```diff
@@ -129,19 +207,193 @@ impl CommandListener {
...
 
+async fn init_api(
+    config: &mut TeliodDaemonConfig,
+    config_path: String,
+) -> Result<ClientConfig, TeliodError> {
+    let mut client_config = ClientConfig::new(config).await?;
+
+    let mut retries = 0;
+    loop {
+        let status = update_machine(&client_config).await?;
```
stalowyjez, line 218: It seems that on the first run of this loop machine id is "", so you just make update request on incorrect URL, what is the point of that? You can have machine id in config, but it would be nice to check it before using it.

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,175 @@
...
+use telio::crypto::PublicKey;
+use thiserror::Error;
+use tracing::{debug, info};
+
+const API_BASE: &str = "https://api.nordvpn.com/v1";
+
+#[cfg(windows)]
+const OS_NAME: &str = "windows";
+#[cfg(target_os = "macos")]
+const OS_NAME: &str = "macos";
```
stalowyjez, line 16: I don't think there is any point of making this for Windows. Maybe it is ok for MacOS?

clis/teliod/src/main.rs
Hasan6979: But machine id is in config, and if it is there, it is loaded here https://github.com/NordSecurity/libtelio/pull/853/files/d07eb05a2b9994218dec09da44e790f0461d355f#diff-adacd55ecee4795f10ea1a14eea3906b343e77c9c9d8a717761e1f82f2f2e7faR375

stalowyjez: Yeah, but it might not be there

stalowyjez: And also machine ID feels to me like a bit internal thing - I'm not sure if we should bother end user with it (soo, I'm not sure if it really should be in config)

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,175 @@
...
+        for item in items {
+            // Get the public_key and identifier from each item
+            if let Some(recvd_pk) = item.get("public_key").and_then(|k| k.as_str()) {
+                if BASE64_STANDARD
+                    .decode(recvd_pk)?
+                    .as_slice()
+                    .cmp(&public_key.0)
+                    .is_eq()
+                {
+                    if let Some(machine_identifier) =
```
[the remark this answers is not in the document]
Hasan6979: Changed it to `Nodes` and `Node`

clis/teliod/src/main.rs
Hasan6979: I would keep this log, but changed to debug!

clis/teliod/src/core_api.rs
Hasan6979: Removed Windows.

clis/teliod/src/main.rs
Hasan6979: I would prefer to keep this function, though modified it a bit. A little less clutter in the main function.

clis/teliod/src/main.rs
stalowyjez: But private key? ^^'

clis/teliod/src/main.rs
Hasan6979: Yes. But that is how it is supposed to work.
And regarding machine ID being in config, I mean, the end user isn't really bothered with it, because it would download it if not present, and if it is there then 🤷

clis/teliod/src/main.rs
Hasan6979: Removed private key as well.

clis/teliod/src/main.rs
```diff
@@ -178,6 +451,7 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
                 match signal {
                     Some(SIGTERM) => {
                         info!("Received SIGTERM signal, exiting");
+                        fs::remove_file("/var/run/teliod.sock")?;
```
stalowyjez, line 454: It should be closed on any kind of exit, not only on such signal

Hasan6979: Moved this to the end.

tomaszklak:
Please remember to clean up the commit history before merging :)

clis/teliod/src/main.rs
```diff
@@ -179,6 +409,9 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
                 match signal {
                     Some(s @ SIGHUP | s @ SIGTERM | s @ SIGINT | s @ SIGQUIT) => {
                         info!("Received signal {:?}, exiting", Signal::try_from(s));
+                        if let Err(e) = tx.try_send(TelioTaskCmd::Quit) {
+                            error!("Channel closed or telio not yet started. {e}");
+                        };
```
stalowyjez, line 414: This comment sounds incorrect, if Telio is just not started it will be added to the channel's buffer. There is another rare case in which it may fail, though (and I think this is why I think it would be better to use normal `send` here):
```
This method differs from [send](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.send) by returning immediately if the channel’s buffer is full or no receiver is waiting to acquire some data. Compared with [send](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.send), this function has two failure cases instead of one (one for disconnection, one for a full buffer).```

File: clis/teliod/src/main.rs [hunk 22]
```diff
@@ -115,17 +152,188 @@ impl CommandListener {
...
+            }
+        }
+    }
+
+    config.private_key = Some(client_config.private_key.clone());
+    config.hw_identifier = Some(client_config.hw_identifier.clone());
+    config.machine_identifier = Some(client_config.machine_identifier.clone());
+
+    let mut file = File::create(config_path)?;
+    serde_json::to_writer(&mut file, &config)?;
```
stalowyjez, line 220: It's rather strange for a program to modify it's own config - I think you should either move it to another file, or just don't save it at all (@Jauler has some smarter ideas how to do it, maybe you can create some task for researching the topic)

clis/teliod/src/main.rs
```diff
@@ -115,17 +152,188 @@ impl CommandListener {
...
+            {
+                Ok(id) => id,
+                Err(e) => {
+                    debug!("Unable to load identifier due to {e}. Registering ...");
+                    register_machine(
+                        &client_config.hw_identifier,
+                        client_config.public_key,
+                        &client_config.auth_token,
+                    )
+                    .await?
```
stalowyjez, line 174: I'm not sure if it makes sense to try to register after any kind of failure - when it is `FetchingIdentifierError` then sure, but if you encountered some connectivity issue I guess it would be better to make some retries, as you do with update? And tbh to register the same applies

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,173 @@
...
+    public_key: PublicKey,
+    hardware_identifier: String,
+    os: String,
+    os_version: String,
+    device_type: String,
+    traffic_routing_supported: bool,
+}
+
+pub async fn load_identifier_from_api(
+    auth_token: &String,
```
tomasz-grz, line 47: Why not simply `&str`?

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,173 @@
...
+    let response = client
+        .post(&format!("{}/meshnet/machines", API_BASE))
+        .header(
+            header::AUTHORIZATION,
+            format!("Bearer token:{}", auth_token),
+        )
+        .header(header::CONTENT_TYPE, "application/json")
+        .header(header::ACCEPT, "application/json")
+        .json(&MeshConfig {
+            public_key: public_key,
```
tomasz-grz, line 151: You can use the shorthand `public_key,`

clis/teliod/src/config.rs
```diff
@@ -50,9 +52,21 @@ fn reconnect_after_expiry_default() -> Percentage {
     Percentage(90)
 }
 
-#[derive(PartialEq, Eq, Deserialize, Debug)]
+#[derive(Deserialize, Debug, Default)]
+pub struct ClientConfig {
+    pub hw_identifier: String,
```
tomasz-grz, line 57: why would this be configurable? I think on the apps this was calculated as some hash from the UDID,  maybe we can use [something like this](https://crates.io/crates/mid) or how was this set for in tcli?

clis/teliod/src/config.rs
```diff
@@ -50,9 +52,21 @@ fn reconnect_after_expiry_default() -> Percentage {
...
+    pub auth_token: String,
+    pub private_key: SecretKey,
+    pub public_key: PublicKey,
```
tomasz-grz, line 60: If public key is derived from private key, why do we need to provide both? Maybe we could use libtelio `generate_public_key(secret_key: SecretKey)` function.
What do you think? this could prevent the user accidentally giving a wrong pair?

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,173 @@
...
+            hardware_identifier: hw_identifier.clone(),
+            os: OS_NAME.to_owned(),
+            os_version: "unknown".to_owned(),
```
tomasz-grz, line 154: We could set this or `device_type` to `teliod`, so that we could quickly distinguish it in the list, like we do in tclid: `"os_version": "macos tcli"`

clis/teliod/src/main.rs
```diff
@@ -167,6 +392,11 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
                     Ok(ClientCmd::HelloWorld{name}) => {
                         info!("Hello {}", name);
                     }
+                    Ok(ClientCmd::GetMeshmap) => {
+                        if let Err(e) = tx.try_send(TelioTaskCmd::GetMeshmap) {
```
tomasz-grz, line 393: When trying this I only get `Command executed successfully`but not a mesnet map.
Also how can I tell that the device was successfully registered? Unfortunately I don't see it when getting the meshnet map manually from the API

clis/teliod/src/main.rs
```diff
@@ -115,17 +152,188 @@ impl CommandListener {
...
         None,
-    );
+    )?;
+
+    let config_ptr = std::sync::Arc::new(client_config);
+    // TODO: This is temporary to be removed later on when we have proper integration
+    // tests with core API. This is to not look for tokens in a test environment
+    // right now as the values are dummy and program will not run as it expects
+    // real tokens.
+    if !config_ptr.auth_token.eq("") {
```
tomasz-grz, line 243: Would be nice to also get (or override) this from env variable `NORD_TOKEN` if present, like tcli does.

clis/teliod/src/main.rs
Hasan6979: This is removed now. It was there as a placeholder before.

clis/teliod/src/config.rs
Hasan6979: PublicKey is generated from PrivateKey. We don't save the keypair.

clis/teliod/src/core_api.rs
Hasan6979: `device_type` cant be `teliod`. It is not a valid expected value. `os_version` could be changed yes!

clis/teliod/src/config.rs
Hasan6979: tcli just joins publickey.name as the hw_identifier and I had used to same logic for this.

clis/teliod/src/main.rs
tomasz-grz: Right, and status reporting is part of a different ticket 😅

clis/teliod/src/config.rs
tomasz-grz: So the user doesn't have to provide a `hw_identifier` to register their device? Then maybe it shouldn't be part of the config file?

clis/teliod/src/core_api.rs
Hasan6979: ✅

clis/teliod/src/main.rs
Hasan6979: Done!

clis/teliod/src/config.rs
Hasan6979: Removed from user config and made part of a separate file to be read/written but not user defined.

clis/teliod/src/config.rs
Hasan6979: Removed.

clis/teliod/src/main.rs
```diff
@@ -166,6 +434,11 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
                     Ok(ClientCmd::HelloWorld{name}) => {
                         info!("Hello {}", name);
                     }
+                    Ok(ClientCmd::GetMeshmap) => {
+                        if let Err(e) = tx.try_send(TelioTaskCmd::GetMeshmap) {
+                            error!("Channel closed or telio not yet started. {e}");
+                        };
+                    }
```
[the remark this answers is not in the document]
Hasan6979: Removed.

clis/teliod/src/main.rs
Hasan6979: Changed to `send` here.

clis/teliod/src/main.rs
Hasan6979: Moved to its own file in local dir

clis/teliod/src/main.rs
Hasan6979: Applied exp backoff to register and fetching identifier

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,234 @@
...
+            .header(header::ACCEPT, "application/json")
+            .send()
+            .await?
+            .text()
+            .await?,
+    )?)
+}
+
+async fn register_machine(
+    hw_identifier: &String,
```
tomasz-grz, line 197: Same with `&String`, according to clippy 
> writing `&String` instead of `&str` involves a new object where a slice will do

clis/teliod/src/main.rs
```diff
@@ -115,14 +161,193 @@ impl CommandListener {
...
+    }
+    identity_path.push(&format!("utun10.json"));
+
+    let mut device_identity = DeviceIdentity::new(&identity_path).await;
+
+    let backoff_bounds = ExponentialBackoffBounds {
+        initial: Duration::from_secs(1),
+        maximal: Some(Duration::from_secs(300)),
+    };
+    let mut backoff = ExponentialBackoff::new(backoff_bounds)?;
```
tomasz-grz, line 178: I'm a bit puzzled by this, why is this `mut`, but we `.clone()` it subsequently, is the backoff state shared `Arc`?

clis/teliod/src/main.rs
```diff
@@ -115,14 +161,193 @@ impl CommandListener {
...
+
+    if device_identity.machine_identifier.eq("") {
+        device_identity.machine_identifier = match load_identifier_with_exp_backoff(
+            auth_token,
+            device_identity.public_key,
+            backoff.clone(),
```
tomasz-grz, line 184: Is `load_identifier_with_exp_backoff` is modifying a shared state of `backoff`? Shouldn't it be taken as `&mut` then? or is this cloning a reference to `backoff`?

clis/teliod/src/main.rs
```diff
@@ -115,14 +161,193 @@ impl CommandListener {
     }
 }
 
+async fn init_api(auth_token: &str) -> Result<DeviceIdentity, TeliodError> {
+    let mut identity_path = dirs::data_local_dir().ok_or(TeliodError::NoDataLocalDir)?;
+    identity_path.push("teliod");
+    if !identity_path.exists() {
+        let _ = fs::create_dir_all(&identity_path);
+    }
+    identity_path.push(&format!("utun10.json"));
```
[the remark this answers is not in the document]
Hasan6979: I am not sure about that. In tcli we get the interface name from the cli. i mean, this doesn't have to be named after the tunnel interface tbh.

clis/teliod/src/main.rs
Hasan6979: This cloning is not a reference to the `backoff` but a clone of backoff as we want each backoff to be independent. I mean, if it is a reference and `load_identifier` takes some retires, then the next backoff that `register_machine` gets would follow up right ?

clis/teliod/src/main.rs
Hasan6979: This is `mut` because this `backoff` is used [here](https://github.com/NordSecurity/libtelio/pull/853/files/4d995b85a8ebe7432a72a5ed2649eeec57929151#diff-adacd55ecee4795f10ea1a14eea3906b343e77c9c9d8a717761e1f82f2f2e7faR228)

clis/teliod/src/main.rs
tomasz-grz: Yes exactly, in that case if `load_identifier` only needs an independent instance of `backoff`, that lives only through-out it's scope, then why do we need to pass it a `backoff` at all? Maybe in that case load_identifier should just create it's own instance of an `ExponentialBackoff`?

What do you think? We could have a sane `Default` implementation for ExponentialBackoff, or just a helper function that creates one? or if we want it to be configurable, we could just pass the `max_duration: usize` for example instead of passing `mut backoff: ExponentialBackoff` which would be a bit more straight forward in terms of ownership 😅

Hasan6979: I think a helper function to create a default backoff sounds nicer!

clis/teliod/src/main.rs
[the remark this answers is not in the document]
tomasz-grz: No it doesn't, but how is the tunnel interface named? quite often `utun10` can already be taken

clis/teliod/src/config.rs
```diff
@@ -53,21 +54,46 @@ fn reconnect_after_expiry_default() -> Percentage {
...
+        }
+
+        // Generate Secret/public key pair
+        debug!("Generating secret key!");
+        let private_key = SecretKey::gen();
+        let public_key = private_key.public();
+
+        // Generate hw_identifier
+        debug!("Generating hw identifier");
+        let hw_identifier = format!("{}.{}", public_key, "openWRT");
```
tomasz-grz, line 81: This most certainly would also be used for QNAP and docker, so maybe just "teliod"?

nat-lab/data/teliod/config.json
```diff
@@ -2,5 +2,6 @@
     "log_level": "trace",
     "log_file_path": "teliod_natlab.log",
     "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
-    "app_user_uid": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+    "app_user_uid": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
+    "interface_name": "teliod"
```
tomasz-grz, line 6: We'll need to remember to add in the readme that for macOS this has to be `utun#`

clis/teliod/src/main.rs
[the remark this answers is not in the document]
Hasan6979: Yes. I have made this name as configurable via the config. In tcli you have to specify which `tun` you would want, so following similar pattern here.

clis/teliod/src/core_api.rs
```diff
@@ -44,62 +62,146 @@ struct MeshConfig {
     traffic_routing_supported: bool,
 }
 
-pub async fn load_identifier_with_exp_backoff(
+fn build_exp_backoff() -> Result<ExponentialBackoff, Error> {
```
tomasz-grz, line 65: Awesome!

clis/teliod/README.md
```diff
@@ -23,6 +23,9 @@ There is a command for running the daemon:
      - `trace`
      - `off`
    - `log_file_path` - a path to the daemon's logging file (relative paths are starting in Teliod's working directory)
+   - `authentication_token` - Token from Nord VPN account to authenticate API calls
+   - `app_user_uid` - A unique number for each user of the application
+   - `interface_name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
```
tomasz-grz, line 28: I think would be nice to also update `example_teliod_config.json`, now it fails with `ParsingError(Error("missing field `interface_name`", line: 6, column: 1))`

Hasan6979: Ah forgot that! I was using my custom config 😅

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,343 @@
...
+
+async fn register_machine_with_exp_backoff(
+    hw_identifier: &str,
+    public_key: PublicKey,
+    auth_token: &str,
+) -> Result<MachineIdentifier, Error> {
+    let mut retries = 0;
+    let mut backoff = build_exp_backoff()?;
+    loop {
+        if let Ok(id) = register_machine(hw_identifier, public_key, auth_token).await {
```
tomasz-grz, line 176: Please brake the retry loop if we receive the `400 - Bad request` error `101120` `Maximum machine count reached`. Or maybe in fact any error from `Bad request` or `Invalid credentials`? Since it makes no sense to retry any amount of times as it will never work.

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+async fn fetch_identifier_with_exp_backoff(
+    auth_token: &str,
+    public_key: PublicKey,
+) -> Result<MachineIdentifier, Error> {
+    let mut retries = 0;
+    let mut backoff = build_exp_backoff()?;
+    loop {
+        match fetch_identifier_from_api(auth_token, public_key).await {
```
tomasz-grz, line 154: Maybe same here? Not much sense in retrying if we get `Invalid authorization header` or `Invalid credentials`

clis/teliod/src/config.rs
```diff
@@ -50,11 +53,47 @@ fn reconnect_after_expiry_default() -> Percentage {
...
+        let public_key = private_key.public();
+
+        // Generate hw_identifier
+        debug!("Generating hw identifier");
+        let hw_identifier = format!("{}.{}", public_key, "teliod");
+
+        DeviceIdentity {
+            hw_identifier,
+            private_key,
+            machine_identifier: None,
```
stalowyjez, line 86: I think it shouldn't be an option - a ready `DeviceIdentity` should have all of the fields ready - soo, I think you can just have two constructors? `from_path` and `new(machine_id)` (or maybe `generate(machine_id)`?). If you fail to get data from file, you can then fetch identifier from File and generate a new one.

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,343 @@
...
+                        device_identity.private_key.public(),
+                        auth_token,
+                    )
+                    .await?,
+                );
+            }
+            _ => {
+                // Retry with exp back-off
+                let _ =
+                    retry_after_backoff_delay(&mut backoff, &mut retries, "update machine").await?;
```
stalowyjez: Tbh Telio's exponential backoff might not be the best tool for this job - it was intended to be used with Telio's state machines (it could be done in this way, but probably not worth it) - I think using https://docs.rs/backoff/0.4.0/backoff/ you could greatly simplify this code - probably with `[with_max_elapsed_time](https://docs.rs/backoff/0.4.0/backoff/exponential/struct.ExponentialBackoffBuilder.html#method.with_max_elapsed_time)` you wouldn't even need `retries`

clis/teliod/src/main.rs
```diff
@@ -115,14 +161,193 @@ impl CommandListener {
...
+
+    let mut device_identity = DeviceIdentity::new(&identity_path).await;
+
+    let backoff_bounds = ExponentialBackoffBounds {
+        initial: Duration::from_secs(1),
+        maximal: Some(Duration::from_secs(300)),
+    };
+    let mut backoff = ExponentialBackoff::new(backoff_bounds)?;
+
+    if device_identity.machine_identifier.eq("") {
```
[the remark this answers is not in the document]
Hasan6979: Modified the logic of this and the conditional nor option isn't needed anymore

clis/teliod/src/core_api.rs
Hasan6979: Done!

clis/teliod/src/core_api.rs
Hasan6979: Done!

clis/teliod/src/comms.rs
```diff
@@ -113,6 +113,13 @@ impl DaemonSocket {
     }
 }
 
+impl Drop for DaemonSocket {
+    fn drop(&mut self) {
+        let socket_path = DaemonSocket::get_ipc_socket_path().unwrap();
+        let _ = fs::remove_file(&socket_path);
```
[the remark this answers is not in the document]
Hasan6979: Added a check in nat-lab test after closing teliod to check whether the socket is present or not

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,343 @@
...
+        if let Ok(id) = register_machine(hw_identifier, public_key, auth_token).await {
+            return Ok(id);
+        } else {
+            let _ =
+                retry_after_backoff_delay(&mut backoff, &mut retries, "register machine").await?;
+        }
+    }
+}
+
+async fn retry_after_backoff_delay(
```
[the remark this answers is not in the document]
Hasan6979: I think a function with a closure is maybe a bit overkill ? I changed the name of the function, hopefully which conveys the intent a bit better.

clis/teliod/src/core_api.rs
Hasan6979: Changed to `backoff` crate

clis/teliod/src/config.rs
Hasan6979: Added `from_path` constructor

clis/teliod/src/core_api.rs
```diff
@@ -0,0 +1,338 @@
...
+    serde_json::to_writer(&mut file, &device_identity)?;
+    Ok(device_identity)
+}
+
+async fn fetch_identifier_with_exp_backoff(
+    auth_token: &str,
+    public_key: PublicKey,
+) -> Result<MachineIdentifier, Error> {
+    let mut backoff = ExponentialBackoffBuilder::new()
+        .with_max_elapsed_time(Some(Duration::from_secs(300)))
```
tomasz-grz, line 143: maybe we could define `300` as a global static?

clis/teliod/src/comms.rs
```diff
@@ -113,6 +113,17 @@ impl DaemonSocket {
     }
 }
 
+impl Drop for DaemonSocket {
+    fn drop(&mut self) {
+        // Normally this should be cleaned up, but in some cases it is not
+        // The ipc path here should be valid, as the daemon socket is created
+        // based on whether the path is valid or not.
+        #[allow(unwrap_check)]
+        let socket_path = DaemonSocket::get_ipc_socket_path().unwrap();
```
tomasz-grz, line 122: how about?
```rust
if let Ok(socket_path) = DaemonSocket::get_ipc_socket_path() {
    let _ = fs::remove_file(&socket_path);
}
```

tomasz-grz (commented):
+1 🚀 Great work, some final knits from me 😄 feel free to ignore

clis/teliod/src/comms.rs
Hasan6979: ✅

clis/teliod/src/core_api.rs
Hasan6979: Made it into a file local const

clis/teliod/src/main.rs
```diff
@@ -63,17 +91,35 @@ enum TeliodError {
...
             if DaemonSocket::get_ipc_socket_path()?.exists() {
                 Err(TeliodError::DaemonIsRunning)
             } else {
-                let file = File::open(config_path)?;
-                let config: TeliodDaemonConfig = serde_json::from_reader(file)?;
+                let file = File::open(&config_path)?;
+                let mut config: TeliodDaemonConfig = serde_json::from_reader(file)?;
+                let token = std::env::var("NORD_TOKEN").ok();
+                if let Some(t) = token {
+                    debug!("Overriding token from env");
```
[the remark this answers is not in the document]
Hasan6979: The idea was so that in development someone doesn't accidentally end up committing their token. I mean, in that sense it would be better to have an option to load it from the env.

stalowyjez (approved):
lgtm, +1

PR #854, opened 2024-10-07
Update rust_build_utils ref to fix dylib rpath

LukasPukenis:
Dylib's encoded absolute path for rpath in
rust_build_utils which made the linker look for
dependencies in the absolute path. This of course
broke because the paths were build-machine-relative paths.

This bump should be enough to force libtelio produce dylibgs with id of @rpath/libtelio.dylib

### Problem

### Solution


lcruz99 (approved):
+2

PR #855, opened 2024-10-07
LLT-5663: Fix TTC for wg peers

gytsto:
Use real rx/tx stats that are updated in uapi

### Problem
CONNECTED/CONNECTING status is based on `time_since_last_rx` value which is cached per peer and normally updated each second based on `self.stats` which contains timestamp of last received data and last sent data. But the `self.stats` is only updated later in the same iteration. Meaning, that `self.time_since_last_rx` will not get updated till the next iteration. Causing TTC for wg peers to increase with extra second.

### Solution
Update `self.stats` at the start of iteration.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5663
+++ b/.unreleased/LLT-5663
@@ -1,0 +1,1 @@
+Fix time to connect delay for wg peers
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -720,8 +720,6 @@ impl State {
                 "Disconnected peer missing from old list",
             ))?;
 
-            self.stats.remove(key);
-
             // Remove all disconnected peers from no link detection mechanism
             if let Some(link_detection) = self.link_detection.as_mut() {
                 link_detection.remove(key);
@@ -743,16 +741,14 @@ impl State {
                 .get(key)
                 .ok_or(Error::InternalError("New peer missing from new list"))?;
 
-            let bytes_and_ts = Arc::new(Mutex::new(BytesAndTimestamps::new(
-                peer.rx_bytes,
-                peer.tx_bytes,
-            )));
-
-            self.stats.insert(*key, bytes_and_ts.clone());
+            let stats = self
+                .stats
+                .get(key)
+                .ok_or(Error::InternalError("No stats available for peer"))?;
 
             // Node is new and default LinkState is down. Save it before sending the event
             if let Some(link_detection) = self.link_detection.as_mut() {
-                link_detection.insert(key, bytes_and_ts.clone());
+                link_detection.insert(key, stats.clone());
             }
 
             self.send_event(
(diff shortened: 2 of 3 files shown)


crates/telio-wg/src/wg.rs
```diff
@@ -906,6 +890,32 @@ impl State {
 
     #[allow(mpsc_blocking_send)]
     async fn update(&mut self, mut to: uapi::Interface, push: bool) -> Result<bool, Error> {
+        for (pk, peer) in &mut to.peers {
```
tomasz-grz, line 893: There's a similar loop below
```rust
        for (pk, peer) in &mut to.peers {
            peer.time_since_last_rx = self.time_since_last_rx(*pk);
        }
```

Could they be combined?

crates/telio-wg/src/wg.rs
```diff
@@ -746,16 +744,14 @@ impl State {
...
-            self.stats.insert(*key, bytes_and_ts.clone());
+            let stats = self
+                .stats
+                .get(key)
+                .ok_or(Error::InternalError("No stats available for peer"))?;
 
             // Node is new and default LinkState is down. Save it before sending the event
             if let Some(link_detection) = self.link_detection.as_mut() {
-                link_detection.insert(key, bytes_and_ts.clone());
+                link_detection.insert(key, stats.clone());
```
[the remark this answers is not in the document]
gytsto: stats is `&Arc<...>`, where insert requires `Arc<...>`, therefore the clone (*compiler suggestion*)

crates/telio-wg/src/wg.rs
```diff
@@ -906,6 +890,32 @@ impl State {
...
+        for (pk, peer) in &mut to.peers {
+            match self.stats.get_mut(pk) {
+                Some(stats) => {
+                    if let Ok(mut s) = stats.lock() {
+                        s.update(
+                            peer.rx_bytes.unwrap_or_default(),
+                            peer.tx_bytes.unwrap_or_default(),
+                        );
+                    } else {
+                        telio_log_error!("poisoned lock for peer - {:?}", pk);
```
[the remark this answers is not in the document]
gytsto: yes, it did... changed

.unreleased/LLT-5663
```diff
@@ -0,0 +1 @@
+Fix TTC for wg peers
```
packgron, line 1: Expant TCC so it would be clear in the future

crates/telio-wg/src/wg.rs
```diff
@@ -1377,6 +1384,13 @@ pub mod tests {
         // Connect
         peer.time_since_last_handshake = Some(Duration::from_secs(15));
         ifa.peers.insert(pkc, peer.clone());
+
+        let _ = task_exec!(&wg.task, async move |s| {
```
packgron, line 1388: Why are we goint into internals of implemenation to update stats, this will make this test very fragile.
Can you do this through update, and test a more realistic case

crates/telio-wg/src/wg.rs
```diff
@@ -1445,7 +1458,13 @@ pub mod tests {
 
         // Connects
         peer.time_since_last_handshake = Some(Duration::from_secs(94));
-        peer.rx_bytes = Some(1);
+
+        let _ = task_exec!(&wg.task, async move |s| {
```
packgron, line 1462: Again this will very easily hide issues in code, it should be done throug update, (poll of wg adapter)

packgron (changes requested):
Then rytis was originaly fixing it, it seemed that issue was in wg_controller update, why was it fully moved to wg.rs, or am I missing smthing

.unreleased/LLT-5663
gytsto: Fixed

crates/telio-wg/src/wg.rs
gytsto: Fixed

crates/telio-wg/src/wg.rs
gytsto: Fixed

packgron (approved):
+1.0

PR #858, opened 2024-10-09
Fix link detection fluctuations

dfetti:
At some point the logic in is_link_up has been changed and it no longer used the WG_KEEPALIVE and RTT.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -198,12 +198,9 @@ impl BytesAndTimestamps {
 
     /// Checks if the link is up based on the round-trip time (RTT) and WireGuard keepalive interval.
     pub fn is_link_up(&self, rtt: Duration) -> bool {
-        if let (Some(rx_ts), Some(tx_ts)) = (self.rx_ts, self.tx_ts) {
-            tx_ts <= rx_ts
-        } else if let Some(rx_ts) = self.rx_ts {
-            rx_ts.elapsed() < WG_KEEPALIVE + rtt
-        } else {
-            false
+        match (self.rx_ts, self.tx_ts) {
+            (Some(rx_ts), Some(tx_ts)) => tx_ts <= rx_ts || rx_ts.elapsed() < WG_KEEPALIVE + rtt,
+            _ => false,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


LukasPukenis (commented):
Looks like the code from before however we do not know exactly the cause of the issue and should be able to include a testcase producing it(and fixing with this PR)

crates/telio-wg/src/wg.rs
```diff
@@ -1819,4 +1816,38 @@ pub mod tests {
...
+            rx_ts,
+            tx_ts,
+        };
+
+        let now = Instant::now();
+        let a_second_ago = now.checked_sub(Duration::from_secs(1)).unwrap();
+        let eleven_seconds_ago = now.checked_sub(Duration::from_secs(11)).unwrap();
+        let rtt = Duration::from_secs(0);
+
+        let missing_ts_1 = build(None, None);
```
ThrasherLT, line 1834: I think this might be simplified by using test parametrization from rstest crate, but this is just a suggestion so feel free to resolve this yourself if you think it's not worth the hassle to implement.

ThrasherLT:
lgtm +1.0

crates/telio-wg/src/wg.rs
```diff
@@ -1819,4 +1817,32 @@ pub mod tests {
...
+        let rtt = Duration::from_secs(0);
+
+        let stats = BytesAndTimestamps {
+            rx_bytes: 5,
+            tx_bytes: 5,
+            rx_ts,
+            tx_ts,
+        };
+        assert_eq!(stats.is_link_up(rtt), expect_link_up);
+    }
```
LukasPukenis, line 1847: Could you add additional:
```
#[case(Some(0), Some(100), true)]
#[case(Some(0), Some(1000), true)]
```

Or some other big number just to show that without any traffic it will not change states?

crates/telio-wg/src/wg.rs
```diff
@@ -1819,4 +1817,32 @@ pub mod tests {
...
+    #[case(Some(11), Some(0), false)]
+    fn test_bytes_and_timestamps(
+        #[case] rx_sec_ago: Option<u64>,
+        #[case] tx_sec_ago: Option<u64>,
+        #[case] expect_link_up: bool,
+    ) {
+        let now = Instant::now();
+        let rx_ts = rx_sec_ago.and_then(|v| now.checked_sub(Duration::from_secs(v)));
+        let tx_ts = tx_sec_ago.and_then(|v| now.checked_sub(Duration::from_secs(v)));
+        let rtt = Duration::from_secs(0);
```
LukasPukenis, line 1838: Ideally `rtt` could also be extracted into the `rstest` above, so it would also be configurable :), it can be explicitly set to zero in existing testcases but additional ones with non-zero value would be nice to have covered and should be pretty easy :)

lcruz99 (approved):
+1

PR #859, opened 2024-10-09
[LLT-5674] Collect interderpcli logs in case of process failure

olekoliinyk:
### Problem
When interderpcli failed to execute there is only INTERNALERROR exception of the process printed out to the execution trace which not helps to identify the actual root cause.

### Solution
In case of exception raised during execution of interderpcli process save_logs method can be called to save interderpcli.log content.

- [x] README.md is updated


Jauler (approved):
+2.0

PR #860, opened 2024-10-10
Add missing subdomain www.google.com in nat-lab's dns

tomaszklak:
### Problem
Request to subdomain `www.google.com` are being made but the dns is lacking an entry for it.

### Solution
Add new entry for `www.google.com`.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/dns-server
+++ b/nat-lab/bin/dns-server
@@ -6,10 +6,12 @@ from twisted.names import dns, server
 KNOWN = {
     dns.AAAA: {
         b"google.com": dns.Record_AAAA(address="2a00:1450:400e:80c::200e"),
+        b"www.google.com": dns.Record_AAAA(address="2a00:1450:400e:80c::200e"),
         b"error-with-noerror-return-code.com": None,
     },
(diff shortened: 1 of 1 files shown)


nat-lab/bin/dns-server
```diff
@@ -6,10 +6,12 @@ from twisted.names import dns, server
 KNOWN = {
     dns.AAAA: {
         b"google.com": dns.Record_AAAA(address="2a00:1450:400e:80c::200e"),
+        b"www.google.com": dns.Record_AAAA(address="2a00:1450:400e:80c::200e"),
```
LukasPukenis, line 9: Just an idea - maybe instead of two entries we could use `*.google.com` ?

LukasPukenis (approved):
+2 no complaints

nat-lab/bin/dns-server
tomaszklak: This is very basic server and the logic is just a lookup in a hashtable - I don't think there is a simple way to handle suffix match like that.

PR #861, opened 2024-10-10
LLT-5677: print api nodes on setup

gytsto:
### Problem
it feels like sometimes mesh_api gives wrong or empty values, therefore further actions in test fails

### Solution
for debugging purposes print out api/node in natlab when setup is done

- [x] README.md is updated


packgron (approved):
+2.0

PR #864, opened 2024-10-11
NVA-7973_android_page_size_16kb

zapekk:
### Problem
Page size 16kB researching

### Solution
For temporary test the request was created


build.rs
```diff
@@ -109,7 +109,7 @@ fn main() -> Result<()> {
     if target_os == "android" {
         let pkg_name = env!("CARGO_PKG_NAME");
         let soname = format!("lib{}.so", pkg_name);
-        println!("cargo:rustc-cdylib-link-arg=-Wl,-soname,{}", soname);
+        println!("cargo:rustc-cdylib-link-arg=-Wl,-z,max-page-size=16384,-soname,{}", soname);
```
tomaszklak, line 112: Please add a link in comment that explains why we are setting `max-page-size`. I've found this: https://developer.android.com/guide/practices/page-sizes#update-packaging which looks relevant, but I'm not familiar with android so maybe there is a better link.

Also, what about `common-page-size=16384` - are we sure that the toolchain is new and we don't need it?

zapekk: https://source.android.com/docs/core/architecture/16kb-page-size/16kb
Per the guidelines:
"To build shared libraries that aren't part of [android project](https://cs.android.com/android/platform/superproject/main), you need to pass this linker flag:
-Wl,-z,max-page-size=16384"

tomaszklak: The page you linked to says:

> 16 KB pages are only supported on arm64 targets with 16 KB kernels. However, there is also an option to [simulate 16 KB userspace on x86_64](https://source.android.com/docs/core/architecture/16kb-page-size/getting-started-cf-x86-64-pgagnostic) for Cuttlefish.

Here we set the `max-page-size` for any android target but we build libtelio on android for `i686`, `x86_64`, `armv7`, `aarch64`. I think you should add it conditionally, only for supported architecture(s).

PR #865, opened 2024-10-11
Increase timeout in interderpcli to 10s

tomaszklak:
### Problem
Timeouts in interderpcli are suspected to be to short which is causing nat-lab failures.

### Solution
Increase timeouts in interderpcli from 3s to 10s

- [x] README.md is updated

### DIFF
--- a/clis/interderpcli/src/main.rs
+++ b/clis/interderpcli/src/main.rs
@@ -17,6 +17,8 @@ use tracing_appender;
 use tracing_subscriber;
 use url::{Host, Url};
 
+const TIMEOUT: Duration = Duration::from_secs(10);
+
 #[derive(Parser)]
 #[command(version)]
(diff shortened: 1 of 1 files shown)


clis/interderpcli/src/main.rs
```diff
@@ -179,7 +184,10 @@ async fn ping_each_other(
         }
     })
     .await
-    .context("Second client did not receive the ping message from the first client in 10s")??;
+    .context(format!(
+        "First client did not receive the ping message from the second client in {:?}",
```
gytsto, line 188: so is it first or second? :smile:

tomaszklak: Omg, defeated by copy-pasting code... thanks for spotting this! :)

gytsto (approved):
+2

PR #866, opened 2024-10-11
[LLT-5563] Add TLS and domain names to core-api and mqtt

olekoliinyk:
### Problem
In context of esp32-meshnet nat-lab integration tests, device emulator is going to send requests to `https://api.nordvpn.com`.  So far, nat-lab core-api was not supporting tls connection and core-api domain. 

### Solution
- Add self-signed certificate with `CN=*.nordvpn.com"` record. Certificate is used for core-api as well as for mqtt broker.
- Add core-api and mqtt broker DNS records
- Refractor core-api and notification center tests to use self-signed certificate when sending API requests to core-api
- Add core-api extra-hosts record to `cone-client-01` in `docker-compose.yml`. This is only for core-api and notification center tests. esp32 emulator is going to use nat-lab DNS server to resolve requested domains.
- For [test_notification_center.py](https://github.com/NordSecurity/libtelio/pull/866/files#diff-546eb346acc06e5621b7b4b8708c4343db16dd6a7485dd3939009bbbb6745d88) self-singed certificate is not used for connection to mqtt broker due to the fact that test is executed on the host and there is no `mqtt.nordvpn.com` DNS records. In context of this test, it might be not so relevant as the purpose of the test is to verify mqtt messages. If there is any issue with tls connection to the mqtt-broker, it will be failing then either on docker compose level or core-api tests.

- [x] README.md is updated


nat-lab/tests/test_notification_center.py
```diff
@@ -64,7 +66,7 @@ async def test_nc_register():
     async with AsyncExitStack() as exit_stack:
         # Setup connections
         connection_beta = await exit_stack.enter_async_context(
```
Jauler, line 68: For this test, I do not think you need client container, you should be able to do it from the test itself directly.

ThrasherLT:
lgtm, but I'm not an expert on python and https, so +0.5

gytsto (dismissed):
+1

nat-lab/tests/helpers.py
```diff
@@ -421,3 +423,38 @@ async def ping_between_all_nodes(env: Environment) -> None:
...
+        curl_command.extend(["-d", data])
+
+    process = await connection.create_process(curl_command).execute()
+    response = process.get_stdout()
+    if expect_response:
+        try:
+            return json.loads(response)
+        except json.JSONDecodeError:
+            assert False, f"Expected JSON response but got: {response}"
+    return None
```
tomaszklak, line 452: Why not use requests ([example](https://requests.readthedocs.io/en/stable/user/advanced/#ssl-cert-verification))? It's already in our pipfile

nat-lab/data/core_api/README.md
```diff
@@ -0,0 +1,4 @@
+# Self-Signed Certificate (SSC)
+
+This folder contains **self-signed certificate**. \
+Please note that this certificate is intended **only for testing purposes** in nat-lab environment.
```
tomaszklak, line 4: nit: It would be nice to add here the exact command(s) used to generate the certificates.

tomaszklak (dismissed):
+1, nice :)

tomaszklak (approved):

PR #867, opened 2024-10-11
LLT-5554: Add MQTT notification center support to `teliod`

tomaszklak:
### Problem
Teliod needs to be notified about meshnet changes which means it has to have a connection to Notification Center.

### Solution
Implement Notification Center connection using rumqttc. After #853 is merged, the placeholder callback(s) will be replaced by triggering of meshmap update.

Integration tests will be done once core api support is added to nat-lab.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/comms.rs
+++ b/clis/teliod/src/comms.rs
@@ -58,6 +58,7 @@ impl DaemonSocket {
         let _ = fs::remove_file(ipc_socket_path);
         let socket = ListenerOptions::new()
             .name(ipc_socket_path.to_fs_name::<FilesystemUdSocket>()?)
+            .reclaim_name(true)
             .create_tokio()?;
 
         Ok(Self { socket })
--- a/clis/teliod/src/main.rs
+++ b/clis/teliod/src/main.rs
@@ -5,6 +5,7 @@ use nix::libc::{SIGHUP, SIGINT, SIGQUIT, SIGTERM};
 use signal_hook_tokio::Signals;
 use std::{
     fs::{self, File},
+    path::PathBuf,
     str::FromStr,
     sync::Arc,
 };
@@ -31,10 +32,15 @@ struct TeliodDaemonConfig {
     log_level: LevelFilter,
     log_file_path: String,
 
-    app_user_id: Uuid,
+    app_user_uid: Uuid,
 
     #[serde(deserialize_with = "deserialize_authentication_token")]
     authentication_token: String,
+
+    /// Path to a http pem certificate to be used when connecting to CoreApi
+    http_certificate_file_path: Option<PathBuf>,
+    /// Path to a mqtt pem certificate to be used when connecting to Notification Center
+    mqtt_certificate_file_path: Option<PathBuf>,
 }
 
 fn deserialize_log_level<'de, D>(deserializer: D) -> Result<LevelFilter, D::Error>
@@ -163,7 +169,7 @@ fn telio_task() {
 
 async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError> {
     let (non_blocking_writer, _tracing_worker_guard) =
-        tracing_appender::non_blocking(fs::File::create(config.log_file_path)?);
+        tracing_appender::non_blocking(fs::File::create(&config.log_file_path)?);
     tracing_subscriber::fmt()
         .with_max_level(config.log_level)
         .with_writer(non_blocking_writer)
@@ -176,8 +182,7 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
     let cmd_listener = CommandListener { socket };
 
     let nc = NotificationCenter::new(
-        &config.authentication_token,
-        config.app_user_id,
+        &config,
         vec![Arc::new(|am| {
             println!("Example callback registered at the start: {am:?}")
         })],
--- a/clis/teliod/src/main.rs
+++ b/clis/teliod/src/main.rs
@@ -3,15 +3,17 @@
 use clap::Parser;
 use nix::libc::{SIGHUP, SIGINT, SIGQUIT, SIGTERM};
 use signal_hook_tokio::Signals;
+use smart_default::SmartDefault;
 use std::{
     fs::{self, File},
+    num::NonZeroU64,
     path::PathBuf,
     str::FromStr,
     sync::Arc,
 };
 use thiserror::Error as ThisError;
 use tokio::task::JoinError;
-use tracing::{error, info, level_filters::LevelFilter, warn};
+use tracing::{debug, error, info, level_filters::LevelFilter, warn};
 use uuid::Uuid;
 
 mod comms;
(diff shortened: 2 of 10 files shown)


gytsto (commented):
lgtm, but I can only give +0.5

clis/teliod/src/nc.rs
```diff
@@ -0,0 +1,533 @@
...
+
+        start_mqtt(nc_config).await?;
+
+        Ok(Self { callbacks })
+    }
+
+    pub async fn add_callback(&self, callback: Callback) {
+        let mut callbacks = self.callbacks.lock().await;
+        callbacks.push(callback);
+    }
```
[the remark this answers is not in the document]
tomaszklak: I'm not sure. I will see how it goes when integrating the result of #853 and remove one of the two way of registering a callback then.

clis/teliod/src/nc.rs
```diff
@@ -0,0 +1,533 @@
...
+    debug!("mqtt incoming publish: {p:?}: {text:?}");
+
+    ensure!(
+        !notification.is_acked(),
+        "received {:?} which is already acked",
+        notification.message_id()
+    );
+
+    send_delivery_confirmation(client, notification.message_id()).await?;
+    send_acknowledgement(client, notification.message_id()).await?;
```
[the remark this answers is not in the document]
tomaszklak: According to the NC documentation we need to send both. Also, the linux client is sending both: https://github.com/NordSecurity/nordvpn-linux/blob/9a5764e0c9f8aafad60c3f157550783a93268eab/nc/nc.go#L368-L378

clis/teliod/src/nc.rs
```diff
@@ -0,0 +1,533 @@
...
+        callbacks.push(callback);
+    }
+}
+
+async fn start_mqtt(nc_config: NCConfig) -> Result<(), Error> {
+    let backoff_bounds = ExponentialBackoffBounds {
+        initial: Duration::from_secs(1),
+        maximal: Some(Duration::from_secs(300)),
+    };
+    let backoff = ExponentialBackoff::new(backoff_bounds)?;
```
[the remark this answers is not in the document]
tomaszklak: This would mean we construct new backoff multiple times **and** have to deal with the fact that this construction can fail in the async task. I think that keeping it here makes error handling much simpler than having to deal with error propagation from the task.

clis/teliod/src/main.rs
```diff
@@ -176,12 +217,24 @@ async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError
...
-                    Some(_) => {
-                        info!("Received unexpected signal, ignoring");
+                    Some(SIGINT) => {
+                        info!("Received SIGINT signal, exiting");
+                        break Ok(());
+                    }
+                    Some(SIGQUIT) => {
+                        info!("Received SIGQUIT signal, exiting");
+                        break Ok(());
+                    }
```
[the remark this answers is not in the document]
tomaszklak: I think you are thinking about https://docs.rs/nix/latest/nix/sys/signal/enum.Signal.html but we are using here constants for `libc` that are just plain `c_int`. Either way I've made it less verbose at the cost of slightly less clear log:
```
2024-10-16T10:22:13.469834Z  INFO teliod: 221: Received signal 2, exiting
```

clis/teliod/src/comms.rs
```diff
@@ -58,6 +58,7 @@ impl DaemonSocket {
         let _ = fs::remove_file(ipc_socket_path);
         let socket = ListenerOptions::new()
             .name(ipc_socket_path.to_fs_name::<FilesystemUdSocket>()?)
+            .reclaim_name(true)
```
Jauler, line 61: Is'nt the reclamation enabled by default?

clis/teliod/src/nc.rs
```diff
@@ -0,0 +1,533 @@
...
+    pub async fn add_callback(&self, callback: Callback) {
+        let mut callbacks = self.callbacks.lock().await;
+        callbacks.push(callback);
+    }
+}
+
+async fn start_mqtt(nc_config: NCConfig) -> Result<(), Error> {
+    let backoff_bounds = ExponentialBackoffBounds {
+        initial: Duration::from_secs(1),
+        maximal: Some(Duration::from_secs(300)),
```
Jauler, line 116: maybe we should make the exponential backoff configurable?

clis/teliod/src/nc.rs
```diff
@@ -0,0 +1,533 @@
...
+    let backoff = ExponentialBackoff::new(backoff_bounds)?;
+
+    tokio::spawn(async move {
+        loop {
+            let (client, mut eventloop, expires_at) =
+                Box::pin(connect_to_nc_with_backoff(&nc_config, backoff.clone())).await;
+
+            loop {
+                select! {
+                    _ = tokio::time::sleep_until(expires_at) => {
```
Jauler, line 127: Maybe lets, just to be safe, recreate the connection at least some amount of time before the expiration? we can have it in ratio, e.g. after reaching 90% of expiration period, or we can simply have fixed amount, like a few minutes or so.

Whatever way we choose to go, I guess it would be nice to have the threshold configurable, but with sane default value.

:pray:

clis/teliod/src/nc.rs
```diff
@@ -0,0 +1,533 @@
...
+                );
+                tokio::time::sleep(backoff.get_backoff()).await;
+                backoff.next_backoff();
+            }
+        }
+    }
+}
+
+async fn connect_to_nc(nc_config: &NCConfig) -> Result<(AsyncClient, EventLoop, Instant), Error> {
+    let credentials = request_nc_credentials(&nc_config).await?;
```
Jauler, line 176: Note that there is one failure path, which is a bit nasty...

If `nc` credentials are successfully received, but then MQTT connection fails - we should at least attempt to revoke them, as otherwise it is possible to pile them up on the server-side.

clis/teliod/src/comms.rs
tomaszklak: It is - I'm not really sure why I've added it and will remove it in a sec 😕

clis/teliod/src/nc.rs
matislovas: I tried following scenario using `Postman` and `mosquitto_sub`:

1. Fetching NC credentials with POST `/v1/notifications/tokens`.
2. Subscribing to MQTT with fetched creds.
3. Sending PATCH request to `/v1/meshnet/machines`. Changes are seen in MQTT `/meshnet` topic.
4. Fetching NC credentials with POST `/v1/notifications/tokens` once again.
5. After that, I am _not_ disconnected from MQTT, but if I do disconnect manually, I am not able to use creds from step 2.

I am forcefully dropped from MQTT broker, if another connection with same identifier (and valid creds) joins in.

matislovas: So we're good there. If new creds are issued, the previous ones are revoked.

tomaszklak: @Jauler do you want me to remove the revocation?

Jauler: Yeah, I talked with the Notification Center devs, and they mentioned, that it is safe to issue multiple tokens with the same `app-user-id` meaning that my initial comment is not really relevant. 

I guess the original flow is much simpler, therefore maybe it makes sense to revert back?

Sorry for the confusion.

tomaszklak: Removed :)

Jauler (dismissed):
+1.5

stalowyjez (dismissed):
Looks good, added 3 comments on things I think could look a bit better, but anyway +1

stalowyjez (approved):

PR #868, opened 2024-10-11
Make upnp lease duration configurable

lcruz99:
### Problem
User is not able to set a custom UPnP lease duration.

### Solution
This PR adds a UPnP telio feature with lease duration field.


crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -36,7 +36,7 @@ const GET_INTERFACE_TIMEOUT_S: Duration = Duration::from_secs(2);
 
 const EPHEMERAL_PORT_RANGE: std::ops::RangeInclusive<u16> = 49152..=65535;
 
-const DEFAULT_LEASE_DURATION_S: u32 = 600;
+const LEASE_RENEWAL_LIMIT_S: u32 = 600;
```
packgron, line 39: What does this const controll, we have what we set as wanted, but i don't entirely get there this comes from.
At what poin we should update lease ?

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -521,17 +527,18 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> State<Wg, I, E> {
...
             Err(e) => {
-                if let Error::NoIGDGateway = e {
-                    telio_log_info!("Failed to check route: route does not exist")
-                } else {
-                    telio_log_info!("Error: {}", e);
+                telio_log_info!("Error validating endpoint: {}", e);
 
-                    telio_log_info!("Deleting an old Upnp endpoint");
-                    let detete = self.igd_gw.delete_endpoint_routes(
+                telio_log_info!("Dropping Upnp endpoint");
```
tomaszklak, line 532: 1. seems to me this should be `telio_log_error!`
2. is there a reason not to merge those two logs into one?

crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -521,17 +527,18 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> State<Wg, I, E> {
...
+                    .igd_gw
+                    .delete_endpoint_routes(
                         self.proxy_port_mapping.external,
                         self.wg_port_mapping.external,
-                    );
-                    let _ = detete.await;
+                    )
+                    .await
+                {
+                    telio_log_info!("Error dropping Upnp endpoint: {}", e);
```
tomaszklak, line 541: This looks like it should be an error log, not info.

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: We have `FeaturesUPnP::lease_duration_s` which sets the actual lease duration for the UPnP port mappings. This `LEASE_RENEWAL_LIMIT_S` is the value which when the lease duration falls below it, we should renew the mappings by sending a request to the UPnP device. 

E.g: We set the mappings lease duration to 3600s, when we check the lease duration left - every upnp_interval tick time - and there's less than 600s left for them to expire, we simple renew the lease for another `FeaturesUPnP::lease_duration_s` which is 3600s.

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: This error could be simply the mappings were already dropped by the UPnP device, as the comment above, I don't consider it worth to be logged as a error, as the runtime is not affected and the mappings won't exist anyway.

src/device.rs
```diff
@@ -1352,6 +1352,7 @@ impl Runtime {
 
             // Create Upnp Endpoint Provider
             let upnp_endpoint_provider = if has_provider(Upnp) {
+                let direct_features = self.features.direct.clone().unwrap_or_default();
```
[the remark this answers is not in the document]
lcruz99: good catch, I missed that

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: Thinking this better, maybe this const can be correlated with the `FeaturesUPnP::lease_duration_s`, something like 1/5 of `FeaturesUPnP::lease_duration_s`

crates/telio-traversal/src/endpoint_providers/upnp.rs
tomaszklak: 1. in that case this sounds like it should be `telio_log_warn!`.
2. imo, lets merge it

tomaszklak:
I think that 4bb57eb85cac982cd20876d22b9b2be544a891d6 shouldn't be a separate commit but a part of b5caf1cd39e8af3e21848be78b4bc295dbd73cef

crates/telio-traversal/src/endpoint_providers/upnp.rs
packgron: Yeh, that was what i was thinking also, maybe min(1/10, 1)

tomaszklak:
> > I think that [4bb57eb](https://github.com/NordSecurity/libtelio/commit/4bb57eb85cac982cd20876d22b9b2be544a891d6) shouldn't be a separate commit but a part of [b5caf1c](https://github.com/NordSecurity/libtelio/commit/b5caf1cd39e8af3e21848be78b4bc295dbd73cef)
> 
> You're right, the commits prefixed with `review:` will be squashed before merging the PR. I've been doing that to avoid messing up the git history for the reviewers and segregate what's been changed during the review.

In that case, when is the good time to review the commit history?

lcruz99:
> > > I think that [4bb57eb](https://github.com/NordSecurity/libtelio/commit/4bb57eb85cac982cd20876d22b9b2be544a891d6) shouldn't be a separate commit but a part of [b5caf1c](https://github.com/NordSecurity/libtelio/commit/b5caf1cd39e8af3e21848be78b4bc295dbd73cef)
> > 
> > 
> > You're right, the commits prefixed with `review:` will be squashed before merging the PR. I've been doing that to avoid messing up the git history for the reviewers and segregate what's been changed during the review.
> 
> In that case, when is the good time to review the commit history?

I'd say after every issue is resolved and +2 acquired. Then commit history is cleaned and approval by the reviewers unlocked.

crates/telio-traversal/src/endpoint_providers/upnp.rs
lcruz99: It has to be higher than the tick time (`upnp_interval`) though, otherwise it will get expired. I'll try to come up with something.

tomaszklak:
> > > > I think that [4bb57eb](https://github.com/NordSecurity/libtelio/commit/4bb57eb85cac982cd20876d22b9b2be544a891d6) shouldn't be a separate commit but a part of [b5caf1c](https://github.com/NordSecurity/libtelio/commit/b5caf1cd39e8af3e21848be78b4bc295dbd73cef)
> > > 
> > > 
> > > You're right, the commits prefixed with `review:` will be squashed before merging the PR. I've been doing that to avoid messing up the git history for the reviewers and segregate what's been changed during the review.
> > 
> > 
> > In that case, when is the good time to review the commit history?
> 
> I'd say after every issue is resolved and +2 acquired. Then commit history is cleaned and approval by the reviewers unlocked.

I don't think we agreed to this way of working - I would like to see a clean git history (in the state you intend to merge it) **before** I add any points.

lcruz99:
> > > > > I think that [4bb57eb](https://github.com/NordSecurity/libtelio/commit/4bb57eb85cac982cd20876d22b9b2be544a891d6) shouldn't be a separate commit but a part of [b5caf1c](https://github.com/NordSecurity/libtelio/commit/b5caf1cd39e8af3e21848be78b4bc295dbd73cef)
> > > > 
> > > > 
> > > > You're right, the commits prefixed with `review:` will be squashed before merging the PR. I've been doing that to avoid messing up the git history for the reviewers and segregate what's been changed during the review.
> > > 
> > > 
> > > In that case, when is the good time to review the commit history?
> > 
> > 
> > I'd say after every issue is resolved and +2 acquired. Then commit history is cleaned and approval by the reviewers unlocked.
> 
> I don't think we agreed to this way of working - I would like to see a clean git history (in the state you intend to merge it) **before** I add any points.

I understand your POV. I propose once the review is done I would clean git history and then get the eventual reviewer points.

tomaszklak (dismissed):
+1

lcruz99:
it looks like the test is flaky, I have to investigate it and probably make some changes

lcruz99:
> it looks like the test is flaky, I have to investigate it and probably make some changes
test is fixed, I've had to adjust endpoint time variables on the test itself.

stalowyjez (approved):
lgtm, +1

PR #870, opened 2024-10-14
add tcpdump for interderp check

gytsto:
### Problem
interderp check fails and we don't know really know why

### Solution
- print interderp logs
- catch tcpdump dumps for derp servers

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -2,13 +2,13 @@ import asyncio
 import os
 import pytest
 import subprocess
-from contextlib import AsyncExitStack
-from helpers import SetupParameters, setup_connections
+from helpers import SetupParameters
 from interderp_cli import InterDerpClient
+from itertools import combinations
 from mesh_api import start_tcpdump, stop_tcpdump
 from utils.bindings import TelioAdapterType
 from utils.connection import DockerConnection
-from utils.connection_util import ConnectionTag, LAN_ADDR_MAP
+from utils.connection_util import ConnectionTag, LAN_ADDR_MAP, new_connection_raw
 from utils.router import IPStack
 from utils.vm import windows_vm_util, mac_vm_util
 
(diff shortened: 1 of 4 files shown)


nat-lab/tests/conftest.py
```diff
@@ -104,45 +104,35 @@ def pytest_make_parametrize_id(config, val):
...
+                    connection,
+                    server1,
+                    server2,
+                    DERP_SERVER_1_SECRET_KEY,
+                    DERP_SERVER_2_SECRET_KEY,
+                    idx,
+                )
+                await derp_test.execute()
+                await derp_test.save_logs()
+        except:
```
packgron, line 133: Why is this not in finlay?

nat-lab/tests/conftest.py
```diff
@@ -104,45 +104,35 @@ def pytest_make_parametrize_id(config, val):
...
+                combinations(
+                    [DERP_SERVER_1_ADDR, DERP_SERVER_2_ADDR, DERP_SERVER_3_ADDR], 2
+                )
+            ):
+                derp_test = InterDerpClient(
+                    connection,
+                    server1,
+                    server2,
+                    DERP_SERVER_1_SECRET_KEY,
+                    DERP_SERVER_2_SECRET_KEY,
```
LukasPukenis, line 128: Don't these keys cause na issue since there are now combinations for them?

nat-lab/tests/conftest.py
gytsto: oh yeah, you are right.... had another stop_tcpdump after try/except statement, but forgot to push it. But `finally` is event better

nat-lab/tests/conftest.py
gytsto: These same keys were used for each combo previously, now it is nothing different. What do you mean, can you elaborate more?

LukasPukenis:
+1

packgron (approved):
+1.0

PR #872, opened 2024-10-14
Add ping argument information

packgron:
### Problem
Then test times out due to global timeout, and ping is the failure reason, it's may be unclear what was not reached

### Solution
Attempt to add this information as task name.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/ping.py
+++ b/nat-lab/tests/utils/ping.py
@@ -16,7 +16,10 @@ async def ping(
     connection: Connection, ip: str, timeout: Optional[float] = None
 ) -> None:
     async with Ping(connection, ip).run() as ping_process:
-        await ping_process.wait_for_any_ping(timeout)
+        await asyncio.create_task(
+            ping_process.wait_for_any_ping(timeout),
+            name=f"ping({connection}, {ip}, {timeout})",
+        )
 
 
 class Ping:


nat-lab/tests/utils/ping.py
```diff
@@ -16,7 +16,10 @@ async def ping(
     connection: Connection, ip: str, timeout: Optional[float] = None
 ) -> None:
     async with Ping(connection, ip).run() as ping_process:
-        await ping_process.wait_for_any_ping(timeout)
+        await asyncio.create_task(
```
Jauler, line 19: :thinking: :thinking: :thinking: 

Does'nt this mean, that we `await` only for the creation of the task, but not the ping process?

packgron: I added here specifcaly, because it was there it shown that ping was stuck in, other place is most likely in docker_process, but that is a bit more generic

packgron:
```
Stack for <Task cancelling name='Task-4272' coro=<ping() running at /builds/vagrant/low-level-hacks/vpn/client/libtelio-build/libtelio/nat-lab/tests/utils/ping.py:19> wait_for=<Future cancelled> cb=[gather.<locals>._done_callback() at /usr/lib/python3.11/asyncio/tasks.py:754]> (most recent call last):
```

Jauler: But I mean it will break tests, because ping normally awaits for its results. And now the function returns immediatelly.

Jauler: oh appearently it seems like I might be wrong. So nevermind. Resolving this.

Jauler (approved):
+2.0

PR #879, opened 2024-10-17
PoC: Rust-based integration test PoC

mathiaspeters:
This PR is a PoC that *should not be merged*.

The purpose of this PR is to showcase how we could implement rust-based integration tests for libtelio, to make discussing a related RFC easier. 

The implementation is based around having process isolation for different libtelio devices and localizing any libtelio-external dependencies to also be run directly as part of the test process, with the exception of the VPN server for the VPN test. For this PoC, the VPN server is run in a docker container, which has to be started before trying to run the test, but would be internalized before any actual implementation is merged.

The goal is to eventually be cross-platform but for this PoC, only linux is supported.

`cargo test test_poc` will run the tests. Using standard `#[test]` was selected here for convenience, but is by no means set in stone.


systests/src/utils/interface_helper.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    pub fn new_ifc_name(&mut self) -> String {
+        let tun_name = format!("tun{}", self.next_tun_num);
+        self.next_tun_num += 1;
+        tun_name
+    }
+
+    pub fn configure_ifc(ifc_name: &str, ip: IpAddr) {
+        run_command(
```
tomasz-grz, line 21: I guess this wouldn't work running the tests on macOS because it uses `ifconfig` for interface management. as far as I know there is no one cross-platform crate for doing so, only retrieving the configuration..

systests/src/utils/test_client.rs
```diff
@@ -0,0 +1,149 @@
...
+        self.dev
+            .start(&DeviceConfig {
+                private_key: self.private_key,
+                adapter: AdapterType::BoringTun,
+                fwmark: None,
+                name: Some(self.ifc_name.to_owned()),
+                tun: None,
+            })
+            .unwrap();
+        self.dev.set_fwmark(11673110).unwrap();
```
tomasz-grz, line 86: This also fails with 
```
  --> systests/src/utils/test_client.rs:86:18
   |
86 |         self.dev.set_fwmark(11673110).unwrap();
   |                  ^^^^^^^^^^ method not found in `Device`
```

systests/Cargo.toml
```diff
@@ -0,0 +1,59 @@
+[package]
+name = "systests"
```
tomasz-grz, line 2: Would be nice to have a `readme.md` with some minimal instructions

systests/src/main.rs
```diff
@@ -0,0 +1,9 @@
+use tests::{meshnet::test_meshnet_poc, vpn::test_vpn_poc};
+
+mod tests;
+mod utils;
+
+fn main() {
```
tomasz-grz, line 6: why don't we use the tests approach here? 
```rust
#[cfg(test)]
mod tests_vpn {
    use super::*;

    #[test]
    fn test_vpn_poc_works() {
        test_vpn_poc();
    }
}
```
then we could get nicer test result reports like
```rust
thread 'test_vpn::test_vpn_poc_works' panicked at systests/src/tests/vpn.rs:51:5:
assertion failed: false
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Successfully executed 'ip link delete tun10'
Successfully executed 'ip rule del priority 32111'


failures:
    test_vpn::test_vpn_poc_works

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.22s
```

tomasz-grz (commented):
Nice work! Some quick initial thoughts before taking a deeper look

mathiaspeters:
Due to changes in product direction, this piece work is not going to be continued, though it might prove useful for further work

PR #880, opened 2024-10-17
Llt 5593 sanitize dns names

tomaszklak:
### Problem
DNS forwarding can lead to logging of the domain names used by the user.

### Solution
Since domain names are logged as part of structures not owned by libtelio, instead of using `Hidden` the same approach as IP hiding is used - `LogCensor`.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5593
+++ b/.unreleased/LLT-5593
@@ -1,0 +1,1 @@
+Domain names in the logs are now hashed in the same way as IPs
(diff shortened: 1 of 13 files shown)


tomaszklak:
> I've tested and it works as expected! One thing though, shouldn't `Features::hide_ips` be renamed at this point?

Thanks for testing! Missing renaming is a good point, thanks!

PR #881, opened 2024-10-17
Update macOS clippy runner and remove actions-rs

tomasz-grz:
### Problem
macOS clippy workflow is [occasionally failing](https://github.com/NordSecurity/libtelio/actions/runs/11365707093/job/31614492082?pr=868) with `(signal: 6, SIGABRT: process abort signal)`.
MacOS-12 runners are [deprecated by GitHub](https://github.com/actions/runner-images/issues/10721) and will be removed.
`actions-rs` Rust GitHub actions are [unmaintained and archived](https://github.com/actions-rs).

### Solution
Update the macOS workflow runners to `macos-13` for `x86_64` targets and `macos-14` for `aarch64`.
Remove `actions-rs` dependency for macOS clippy workflow, as the GitHub Runners already come with all the Rust [tools preinstalled](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#rust-tools).

- [x] README.md is updated

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,13 +22,13 @@ jobs:
         include:
           - arch: x86_64
             target_os: macos
-            runner: macos-12
+            runner: macos-13
           - arch: aarch64
             target_os: macos
-            runner: macos-12
+            runner: macos-14
           - arch: aarch64
             target_os: ios
-            runner: macos-12
+            runner: macos-13
           - arch: x86_64
             target_os: windows
             runner: windows-2022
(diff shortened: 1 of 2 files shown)


mathiaspeters:
Is there a reason `macos-14` couldn't be used for `x86` as well?

mathiaspeters:
> > Is there a reason `macos-14` couldn't be used for `x86` as well?
> 
> Hmm it could be, for the `build.yml` you mean? I just followed what we have already in `tests.yml`. But by default `macos-14` is an `aarch64` runner. We would have to additionally install the `x86_64` tool-chain, either by converting `rust-toolchain` to `rust-toolchain.toml` and specifying it there, or just have an extra workflow step.
> 
> If we want to use `macos-14 aarch64` for `x86_64` tests, that would probably have to use `Rosetta` emulation, which I don't see a benefit of.
> 
> We cloud also just run clippy on `aarch64` I doubt it would matter that much for linting.
> 
> What do you have in mind? Do you have any concerns about using `macos-13` runners?

My "concern" was mostly about using two different version if not necessary, but I think your implementation is the right choice given that we'd either need two different `macos-14` for x86 and aarch or install stuff in `macos-14`. So all good from my side

mathiaspeters:
+2

jjanowsk (approved):

PR #883, opened 2024-10-20
Fix UNIX Symbolic Link (Symlink) Following not protected from untrusted code injection

codingfromanywhere:
### Problem
The server protects static routes from path traversal outside the root directory when `follow_symlinks=False` (default). It does this by resolving the requested URL to an absolute path and then checking that path relative to the root. However, these checks are not performed when looking for compressed variants in the `FileResponse` class, and symbolic links are then automatically followed when performing `Path.stat()` and `Path.open()` to send the file. Servers with static routes that contain compressed variants as symbolic links, pointing outside the root directory, or that permit users to upload or create such links, are impacted.

### Solution
Upgrade new package as `fix` :
```js
                # Do not follow symlinks and ignore any non-regular files.
                st = compressed_path.lstat()
                if S_ISREG(st.st_mode):
                    return compressed_path, st, file_encoding
```


LukasPukenis:
Hello. I assume it's an alternative account for `andrisecops` since PRs happened at similar time and have identical description. 

Due to no responses to our questions on another PR, we've moved on with the issue ourselves by updating aiohttp manually to the patched version in the PR: https://github.com/NordSecurity/libtelio/pull/887.

Thanks for contributing and for the future I kindly suggest to follow up on discussions in order to have your PR approved.

PR #884, opened 2024-10-21
Detect long blocking tasks

tomaszklak:
### Problem
We suspect that tokio threads are blocked by some blocking operations.

### Solution
Print a log when some thread goes from unparked to parked after more than 1s - telio has no place where it expects to be doing work for that long without a break. To make it simpler to match with what was happening thread ID will also be printed in the logs.

- [x] README.md is updated

### DIFF
--- a/crates/telio-utils/src/tokio.rs
+++ b/crates/telio-utils/src/tokio.rs
@@ -6,7 +6,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-const ALERT_DURATION: Duration = Duration::from_secs(60);
+const ALERT_DURATION: Duration = Duration::from_secs(10);
 const UNPARKED_THRESHOLD: Duration = Duration::from_secs(1);
 
 #[derive(PartialEq, Eq, Debug, Clone, Copy)]
@@ -19,7 +19,7 @@ enum ThreadStatus {
 
 /// ThreadTracker will track changes of state of tokio's threads.
 pub struct ThreadTracker {
-    statuses: FxHashMap<ThreadId, ThreadStatus>,
+    statuses: FxHashMap<ThreadId, (ThreadStatus, Instant)>,
     last_change: Instant,
 }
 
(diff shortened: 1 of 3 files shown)


tests/logger.rs
```diff
@@ -13,7 +13,7 @@ mod test_module {
     #[test]
     fn test_logger() {
         // Line number of tracing::info! location
-        const INFO_LINE: u32 = 50;
+        const INFO_LINE: u32 = 51;
```
LukasPukenis, line 16: What's the meaning of this constant?

tomaszklak: This is the line number in this file of the `tracing::info!` macro call in this test.

crates/telio-utils/src/tokio.rs
```diff
@@ -33,8 +34,15 @@ impl Default for ThreadTracker {
...
-        self.statuses.insert(current().id(), status);
+        let now = Instant::now();
+        let delta = now - self.last_change;
+        self.last_change = now;
+        if let Some(ThreadStatus::Unparked) = self.statuses.insert(current().id(), status) {
+            if status == ThreadStatus::Parked && delta > UNPARKED_THRESHOLD {
+                let tid = std::thread::current().id();
+                telio_log_debug!("Thread {tid:?} was unparked for too long: {delta:?}");
+            }
+        }
```
LukasPukenis, line 45: :thinking: if I understand correctly, this will not provide us specific information into what's blocking but rather the fact that it happens.(checks intervals between [unparked; parked]).

What do you think about a possible solution:
- Log when `tid` was `parked`. When `unpark`ing check how long it was actually parked meaning doing job we suspect might be too long.
- Hopefully stacktrace is useful in that case and we can track it down. Without stacktrace I suppose this is the same as existing implementation :)

The way I see it, in the current implementation we lose the information about the owner of the long task

Jauler (changes requested):
In current implementation, I guess nothing will be printed if the thread is being blocked forever (as it never gets parked again). I guess this is not super likely, that we are experiencing it right now, but it would be cool if we could catch such condition as well. 

For that I guess it should be enough to alter https://github.com/NordSecurity/libtelio/blob/main/crates/telio-utils/src/tokio.rs#L81 to print for unparked threads as well.

Do you think it is worth doing that?

tomaszklak:
> In current implementation, I guess nothing will be printed if the thread is being blocked forever (as it never gets parked again). I guess this is not super likely, that we are experiencing it right now, but it would be cool if we could catch such condition as well.
> 
> For that I guess it should be enough to alter https://github.com/NordSecurity/libtelio/blob/main/crates/telio-utils/src/tokio.rs#L81 to print for unparked threads as well.
> 
> Do you think it is worth doing that?

I've added something like that but to be sure that it's logged during tests, I've reduced the `ALARM_DURATION` from 60s to 10s.

crates/telio-utils/src/tokio.rs
LukasPukenis: I had a wrong view of the park/unpark. Per my previous understanding parking thread meant _actually doing the work_. The whole comment is moot :)
In your case it's exactly how it should be it seems.

LukasPukenis:
+1

Jauler:
> > In current implementation, I guess nothing will be printed if the thread is being blocked forever (as it never gets parked again). I guess this is not super likely, that we are experiencing it right now, but it would be cool if we could catch such condition as well.
> > For that I guess it should be enough to alter https://github.com/NordSecurity/libtelio/blob/main/crates/telio-utils/src/tokio.rs#L81 to print for unparked threads as well.
> > Do you think it is worth doing that?
> 
> I've added something like that but to be sure that it's logged during tests, I've reduced the `ALARM_DURATION` from 60s to 10s.

Cool, I was also thinking, that 60s may be too long ;) :P

Jauler (approved):
+1.0

PR #885, opened 2024-10-21
Fix magic DNS not working under PQ VPN

matszczygiel:
### Problem
Magic DNS breaks when a PQ connection is established

### Solution
Magic DNS relies on the stable public key of the WG interface. When connecting to the PQ server the keys change and are rotated every few minutes. The fix is to notify the DNS about this change.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_pq.py
+++ b/nat-lab/tests/test_pq.py
@@ -12,6 +12,7 @@ from utils.connection_util import (
     ConnectionTag,
     new_connection_by_tag,
 )
+from utils.dns import query_dns
 from utils.ping import ping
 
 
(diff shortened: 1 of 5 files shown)


nat-lab/tests/test_pq.py
```diff
@@ -263,3 +264,111 @@ async def test_pq_vpn_rekey(
...
+
+        await client.connect_to_vpn(
+            str(wg_srv["ipv4"]),
+            int(wg_srv["port"]),
+            str(wg_srv["public_key"]),
+            pq=True,
+        )
+        await ping(client_conn, config.PHOTO_ALBUM_IP)
+
+        # Expect this to work as well after the secrate key change
```
mathiaspeters, line 373: Nit:
```suggestion
        # Expect this to work as well after the secret key change
```

mathiaspeters:
Just out of curiosity, did you run the new natlab test without the fix in place to verify it was able to reproduce the bug?

matszczygiel:
@mathiaspeters Yes, I did. It was failing before the fix

nat-lab/tests/test_pq.py
matszczygiel: Fixed

mathiaspeters:
+1

matislovas (approved):
+1

PR #886, opened 2024-10-22
[LLT-5634] Stop endpoints invalidating themselves

stalowyjez:
### Problem
Currently we notify endpoints gone when they were overwritten by Wireguard, which is fine, but we also do that when the endpoint is set by Telio, which doesn't make a lot of sense, because in this way endpoint removes itself.

### Solution
This PR adds info about what (i.e. Telio or Wireguard) changed the endpoint and then using it to claim that the endpoint is gone only if it was changed by Wireguard.


crates/telio-wg/src/uapi.rs
```diff
@@ -28,6 +28,15 @@ pub enum Error {
     ParsingError(&'static str, String),
 }
 
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+/// Possiblie reason of the endpoint change:
+pub enum EndpointChangeReason {
```
lcruz99, line 33: can we use this enum on [update fn](https://github.com/NordSecurity/libtelio/pull/886/files#diff-e155115f7ca084b5dc3db3e636dd963f8174c17b6d08ec4f7c871df22363f5b0R914) arguments instead of bool?

stalowyjez:
@lcruz99 I added some test - I think it should catch it efficiently in the future, nat-lab tests are expensive, I think we should avoid them if sth can be tested with unit tests

lcruz99:
@stalowyjez super! you could reproduce it with the test you've added?

crates/telio-wg/src/uapi.rs
stalowyjez: Added this refactor in a separate commit - I guess `UpdateReason` makes more sense when we talk about using it for the update function everywhere. I'm not sure whether it looks better or not tbh, I'm happy with both versions.

mathiaspeters (approved):
+1

lcruz99 (approved):

PR #888, opened 2024-10-22
Move moose flush to spawn blocking

tomaszklak:
### Problem
In natlab tests, calls to `lana!(flush_changes)` took more than 1s. Since it's non-async, it blocks the thread on which current async task is running.

### Solution
To fix this, it's wrapped in `spawn_blocking` which moves it to a separate thread where tokio is expecting blocking to occur.

Also fixed the CI job for unused-features by locking in the dependency versions.

- [x] README.md is updated

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -151,10 +151,8 @@ jobs:
           toolchain: 1.80.0
           override: true
           default: true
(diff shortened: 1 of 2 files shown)


Jauler (approved):
+2.0

tomasz-grz (approved):
+1.0

PR #892, opened 2024-10-23
[LLT-5722] revert debug builds

Jauler:
### Problem
With optimization for time to run dev pipelines, it was accidentally pushed that nightly pipelines are being built using debug builds. As those changes were not intentional, this PR reverts those changes.


tomaszklak (approved):
+2

PR #893, opened 2024-10-23
[LLT-5725] avoid opening and closing log file on every log line

Jauler:
### Problem
`libtelio_remote.py` when consuming logs via callback would open and close log file on every single log line, which is really innefficient. This PR reuses the same open file handle for the duration of `libtelio_remote.py` process itself.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -444,9 +444,6 @@ class Client:
                         await self.set_meshnet_config(meshnet_config)
                     yield self
             finally:
-                print(datetime.now(), "Test cleanup: Saving logs")
-                await self.save_logs()
-
                 print(
                     datetime.now(),
                     "Test cleanup: Stopping tcpdump and collecting core dumps",
@@ -474,6 +471,7 @@ class Client:
 
                 print(datetime.now(), "Test cleanup: Shutting down")
                 if self._libtelio_proxy:
+                    await self.get_proxy().flush_logs()
                     await self.get_proxy().shutdown(self._connection.target_name())
                 else:
                     print(
@@ -490,7 +488,10 @@ class Client:
                 await self.save_moose_db()
 
                 print(datetime.now(), "Test cleanup: Checking logs")
-                await self.check_logs_for_errors()
+                await self._check_logs_for_errors()
+
+                print(datetime.now(), "Test cleanup: Saving logs")
+                await self._save_logs()
 
                 print(datetime.now(), "Test cleanup complete")
 
@@ -892,6 +893,15 @@ class Client:
             await asyncio.sleep(1)
 
     async def get_log(self) -> str:
+        await self.flush_logs()
+        return await self._get_log_without_flush()
+
+    async def _get_log_without_flush(self) -> str:
+        """
+        This function retrieves telio logs without flushing them. It may be needed to do that
+        if log retrieval is requested after process has already exited. In such a case there is
+        nothing to flush and attempting to do so will cause errors.
+        """
         process = (
             self._connection.create_process(["type", "tcli.log"])
             if self._connection.target_os == TargetOS.Windows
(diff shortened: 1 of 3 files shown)


nat-lab/tests/telio.py
```diff
@@ -986,7 +974,7 @@ async def get_network_info(self) -> str:
         return ""
 
     async def check_logs_for_errors(self) -> None:
```
tomaszklak, line 976: Since both `check_logs_for_errors` and `save_logs` now depend on the fact that they are called only in this file and after the call to `flush_logs` I think it would be good to add something to make sure we don't accidentally call them in the wrong context. Maybe at least some sort of prefix could be added to the names, like `_` (or `_unsafe_`) to indicate those are implementation details not meant to be called from outside. Also a docstring explaining they don't flush would be good.

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -51,38 +52,40 @@ def next_event(self) -> libtelio.Event:
...
-                try:
-                    logfile.write(f"{datetime.datetime.now()} {log_level} {payload}\n")
-                except IOError as e:
-                    logfile.write(
-                        f"{datetime.datetime.now()} Failed to write logline due to error {str(e)}\n"
-                    )
+            lock_acquired_ts = time.time()
+            self.logfile.write(f"{datetime.datetime.now()} {log_level} {payload}\n")
+            write_ts = time.time()
+        end_ts = time.time()
```
tomaszklak, line 66: nit: I'm not sure we need `write_ts` if we already have `end_ts`

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -176,23 +179,35 @@ def probe_pmtu(self, host: str) -> int:
...
 
 def main():
     object_name = sys.argv[1]
     container_ip = sys.argv[2]
     port = int(sys.argv[3])
 
+    # Cleanup old log files if any exists
+    try:
+        os.remove(TCLI_LOG)
+        os.remove(REMOTE_LOG)
```
tomaszklak, line 195: Is that needed, I thought only `TCLI_LOG` is impacted. Or is this some additional fix/improvement?

tomaszklak (commented):
+1, ~~one real comment and two nitpicks but if we want to have it merged today I'm fine with doing name changes/docstrings in a future PR.~~

nat-lab/tests/uniffi/libtelio_remote.py
Jauler: TBH, this just replicates old behavior:
* https://github.com/NordSecurity/libtelio/blob/c296e61379e9cdab89009119314e082b63209368/nat-lab/tests/uniffi/libtelio_remote.py#L57
* https://github.com/NordSecurity/libtelio/blob/c296e61379e9cdab89009119314e082b63209368/nat-lab/tests/uniffi/libtelio_remote.py#L77

nat-lab/tests/uniffi/libtelio_remote.py
Jauler: Initially I added it just to check whether unlocking is not causing delays. But it does not, and I guess it is reasonably safe to assume, that it will not.

Hence: Fixed

nat-lab/tests/telio.py
Jauler: Updated

tomaszklak (approved):

nat-lab/tests/telio.py
```diff
@@ -937,23 +947,6 @@ async def clear_system_log(self) -> None:
                     f"Clear-EventLog -LogName {log_name}",
                 ]).execute()
 
-    async def get_log_lines(self, regex: Optional[str] = None) -> List[str]:
```
Hasan6979, line 940: This isn't used anymore, but maybe we can keep this as a util for any future test ? Or is there another alternative for this now ?

Jauler: IMHO, whenever we actually have a need for it, we can add it. Right now there are no usages of it.

Hasan6979 (approved):
+1

PR #895, opened 2024-10-23
Call logging callback on a separate thread

tomaszklak:
To avoid blocking callers, when logging is slow, the call to the logging callback is moved to the separate therad. Now, logging anything only causes non blocking try_send that will fail if there is no more space in the queue of not yet processed log messages.

To have some visibility into how this system performs, count of dropped logs is printed along side one of the periodic logs of libtelio.

- [x] README.md is updated

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -99,7 +99,7 @@ pub use wg::{
 #[cfg(test)]
 use wg::tests::AdapterExpectation;
 
-use crate::logging::LOG_CENSOR;
+use crate::logging::{logs_dropped_until_now, LOG_CENSOR};
 
 #[derive(Debug, TError)]
 pub enum Error {
@@ -2422,7 +2422,7 @@ impl TaskRuntime for Runtime {
             },
 
             _ = self.polling_interval.tick() => {
-                telio_log_debug!("WG consolidation triggered by tick event");
+                telio_log_debug!("WG consolidation triggered by tick event, total logs dropped: {}", logs_dropped_until_now());
                 wg_controller::consolidate_wg_state(&self.requested_state, &self.entities, &self.features)
                     .boxed()
                     .await
--- a/src/device.rs
+++ b/src/device.rs
@@ -99,7 +99,7 @@ pub use wg::{
 #[cfg(test)]
 use wg::tests::AdapterExpectation;
 
-use crate::logging::{logs_dropped_until_now, LOG_CENSOR};
+use crate::logging::{logs_dropped_since_last_checked, logs_dropped_until_now, LOG_CENSOR};
 
 #[derive(Debug, TError)]
 pub enum Error {
@@ -2423,6 +2423,10 @@ impl TaskRuntime for Runtime {
 
             _ = self.polling_interval.tick() => {
                 telio_log_debug!("WG consolidation triggered by tick event, total logs dropped: {}", logs_dropped_until_now());
+                let dropped = logs_dropped_since_last_checked();
+                if dropped > 0 {
+                    telio_log_warn!("New logs dropped: {dropped}");
+                }
                 wg_controller::consolidate_wg_state(&self.requested_state, &self.entities, &self.features)
                     .boxed()
                     .await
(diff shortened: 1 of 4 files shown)


src/ffi/logging.rs
```diff
@@ -17,18 +20,26 @@ use tracing_subscriber::{
 
 use crate::{TelioLogLevel, TelioLoggerCb};
 
+const LOG_BUFFER: usize = 20;
```
tomaszklak, line 23: What should be the value here? Should this be somehow configurable?

src/device.rs
```diff
@@ -2410,7 +2410,7 @@ impl TaskRuntime for Runtime {
             },
 
             _ = self.polling_interval.tick() => {
-                telio_log_debug!("WG consolidation triggered by tick event");
+                telio_log_debug!("WG consolidation triggered by tick event, total logs dropped: {}", logs_dropped_until_now());
```
LukasPukenis, line 2413: nit: do we expect it in normal circumstances? Maybe `telio_log_warn!` is more appropriate

src/ffi/logging.rs
Jauler: I guess we can make FeatureConfig value for it with the default to something like 128 or so. What do you think?

tests/logger.rs
```diff
@@ -13,24 +16,32 @@ mod test_module {
...
         }
         impl TelioLoggerCb for TestLogger {
             fn log(
                 &self,
                 log_level: telio::ffi_types::TelioLogLevel,
                 payload: String,
             ) -> FfiResult<()> {
+                if payload == "Starting async logger thread" {
+                    return Ok(());
+                }
```
LukasPukenis, line 36: Nit: this is detached from the actual message, such upon change, this conditional will break. It poses no harm but will eventually become a question "what is this log line in my logs". I suggest making a constant of it or just removing the check altogether :)

src/ffi/logging.rs
```diff
@@ -164,6 +187,48 @@ fn filter_log_message(msg: String) -> Option<String> {
...
+enum LogMessage {
+    Message(TelioLogLevel, String),
+    Quit,
+}
+
+fn start_async_logger(cb: Box<dyn TelioLoggerCb>, buffer_size: usize) -> SyncSender<LogMessage> {
+    let (sender, receiver) = sync_channel::<LogMessage>(buffer_size);
+    let handle = Builder::new()
+        .name("libtelio-logger".to_owned())
+        .spawn(move || {
```
LukasPukenis, line 199: Moving logging into a thread is a nice move, especially since we don't control the callback :+1: 

Derivative question if you can answer it - is this the same for event callbacks, are those in a separate thread?

src/ffi/logging.rs
tomaszklak: The problem is that the features are passed in for each [telio/device constructor](https://github.com/NordSecurity/libtelio/blob/main/src/libtelio.udl#L193) while the logger is configured[ completely independently and possibly only once](https://github.com/NordSecurity/libtelio/blob/main/src/libtelio.udl#L157).

LukasPukenis: To me personally, it feels nice to have a big value in here as it feels safer, even though I understand that if we hit even `20` in here it probably means that the log sink is lagging behind and is already problematic.

In theory, once it starts lagging to some degree, probably it will continue the trend and will increase.

I would however feel a bit safer if this value would be slightly higher in case the sink would have a hiccup or two, say value of 1024(purely emotional choice).

LukasPukenis (commented):
Left some non-blocking questions. Looks nice +1

src/device.rs
tomaszklak: Not sure I understand, but this is not the delta since last call, but a total number of dropped logs (strictly monotonic) so the log should stay debug.

LukasPukenis: I mean if the count of skipped events is greater than 0 then we could warn! instead of debug!, however this would require keeping the previous skip count to detect this.

basically my point is to make it more visible as debug messages might get lost in a sea of other logs :)

tomaszklak: You're right, with buffer size at 1024, it should be fine to issue a warning about new dropped logs. I've added it and some tests to actually prove the dropping is working as expected :)

matislovas (approved):
+1

PR #896, opened 2024-10-24
Fix broken (unexpected 'name()' call) wg_controller tests

tomaszklak:
### Problem
Some of the tests in wg_controrller started to fail on main. After local reproduction, it turns out that `EndpointProvider::name()` was called on a mock instance without any expectation set for that function.

The snippet that requires `expect_name` is this: https://github.com/NordSecurity/libtelio/blob/main/src/device/wg_controller.rs#L198-L206

### Solution
Add `expect_name` call.

- [x] README.md is updated


Jauler (approved):
+2.0

PR #898, opened 2024-10-25
Llt 5617 collect dns server logs

tomaszklak:
### Problem
Logs from dns server (and tcpdump pcaps) are not collected from dns servers, but are needed for nat-lab investigation.

### Solution
Dns logs are now timestamped and are collected once per the run_local.py run. The pcaps are collected per test.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -250,16 +250,38 @@ def pytest_sessionfinish(session, exitstatus):
         return
 
     if not session.config.option.collectonly:
-        num_containers = 3
+        collect_nordderper_logs()
+        collect_dns_server_logs()
 
-        for i in range(1, num_containers + 1):
-            container_name = f"nat-lab-derp-{i:02d}-1"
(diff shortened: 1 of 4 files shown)


nat-lab/tests/conftest.py
```diff
@@ -244,22 +244,55 @@ def pytest_runtest_setup():
     asyncio.run(perform_pretest_cleanups())
 
 
+# pylint: disable=unused-argument
+def pytest_runtest_call(item):
```
gytsto, line 248: Isn't this called before each test? But I might be interpreting this wrong https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=pytest_runtest_call#pytest.hookspec.pytest_runtest_call

gytsto: ah nvm, I now see that pcaps are for each test

gytsto (approved):
lgtm +1

mathiaspeters (approved):
+1

PR #899, opened 2024-10-25
Add support for qnap library build

lcruz99:
This PR adds qnap build support to the build scripts.

It builds a statically linked `tcli` binary and then wraps it into qnap's QPKG.

Currently only `x86_64` arch is supported.

- [x] README.md is updated


ci/build_libtelio.py
```diff
@@ -260,6 +290,22 @@ def post_copy_darwin_debug_symbols_to_distribution_dir(config, args):
             NAME: {f"lib{NAME}": f"lib{NAME}.so"},
         },
     },
+    "qnap": {
+        "archs": {
+            "x86_64": {
+                "env": {
+                    "RUSTFLAGS": (
+                        f" -L {PROJECT_ROOT}/3rd-party/libmoose/{LIBTELIO_ENV_MOOSE_RELEASE_TAG}/bin/common/linux/{MOOSE_MAP['x86_64']}  -C target-feature=+crt-static",
```
tomaszklak, line 298: I don't think we can do this. This forces static linking of the standard library, but without changing the standard library we will link to glibc. Even ignoring the technical reasons why it might not be such a good idea to link statically to glibc, this means that we would link to statistically LGPL library. This forces us to release **all** code under (l)gpl.

lcruz99: QNAP has 10 years old glibc version, hence the idea to statically link with it.

lcruz99: Libtelio statically links glibc without modifying it, would it still be considered derivated work?
Also, libtelio uses GPL3 which has already a strong copyleft, is LGPL relevant here? Libmoose is still dynamically linked. I'm not sure about this...

lcruz99: And not every 3rd party libraries support `musl`, so an extra effort is needed to use it

tomaszklak: Maybe it's fine, but even if we right now don't link anything statically with libtelio this will expose us to potential risks. If we accidentally release something statically linked in future this will have to be opensourced.

If this has been already discussed and confirmed as fine with legal than ignore me :)

Jauler: I think we cannot statically link glibc, as it, unless we know, that we are not using `libnss`. Otherwise we will very likely encounter crashes during runtime due to this problem: https://bugzilla.redhat.com/show_bug.cgi?id=111298#c2

If you want to have static libc for qnap -> I think musl is simpler choice from technical perspective.

lcruz99: @tomaszklak @Jauler now building targetting `musl`.

tomaszklak: Where can I see it build for musl (or how exactly to do this locally) ?

gytsto (commented):
I am not sure about qnap building process itself, but scripts looks fine, therefore +0.5

ci/build_libtelio.py
lcruz99: You're right, I've removed `+crt-static`.

You can take a look at gitlab's `LLT-5729_add_qnap_support` branch, there's a green build on the latest pipeline.

ci/build_libtelio.py
```diff
@@ -101,6 +101,36 @@ def post_copy_darwin_debug_symbols_to_distribution_dir(config, args):
...
 
 
+def post_qnap_build_wrap_binary_on_qpkg(config, args):
+    packages = LIBTELIO_CONFIG[config.target_os].get("packages", None)
+    if packages:
+        for _, bins in packages.items():
+            for _, bin in bins.items():
+                src_path = os.path.join(
+                    PROJECT_CONFIG.get_distribution_path(
+                        config.target_os, config.arch, "", config.debug
```
tomaszklak, line 111: Why are we packaging the debug bulid and not the release build?

ci/build_libtelio.py
tomaszklak: This doesn't seem to produce a static library:
```
$ ldd ~/dev/tcli
        linux-vdso.so.1 (0x00007ffd3c17d000)
        libsqlite3.so.0 => /lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007a5f38f2a000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007a5f36319000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007a5f36000000)
        /lib64/ld-linux-x86-64.so.2 (0x00007a5f3908b000)
```

lcruz99: It looks like rust_build_utils hash was outdated, fixed it

lcruz99:
We are still dynamically linking `libsqlite3.so`, I'll request `libsqlite3.a` to moose team and soon update this PR

tomaszklak (commented):
Please clean up the commit history

tomaszklak (approved):
+1

PR #900, opened 2024-10-25
[LLT-5742] Update endpoint invalidation logic

Jauler:
## Problem
Recent changes to endpoint invalidation logic missed a scenario, where connection is downgraded from direct to relayed one, leaving old endpoints in a valid state. This may cause spurious upgrade request, and unnecessary relay->direct->relay cycles.

This PR refactors the logic a bit. First, it exports this invalidation logic into separate consolidation function, and defines (in comments) specific conditions when endpoints should be invalidated.

What is more, it changes when endpoints are considered for invalidation, to be able to catch both cases, when we are roaming from one EP to another (including downgrade-by-roaming) and when we forcefully downgrading by inactivity.

### Failing tests analysis
Recently `test_direct_working_paths` started failing sometimes. I was digging into root cause, and the following is what I managed to find:
1. According to test scenario, all the nodes are created and establishes direct connection :checkmark:
2. According to test scenario, all nodes direct connection is verified to be truly direct :checkmark:
3. Then according to test scenario - verifying direct->relay transition commences
    3.1 iptables rules are inserted to break direct connections
    3.2 Test is waiting for all nodes to report "relayed" connection
    3.3 A link between eta and beta specifically reports relay, but before iptables rules (blocking direct connections) are cleared,  a spurious upgrade to direct is performed. This Spurious update to upgrade generates two events, one claiming that the connection is direct and later, after libtelio realizes, that upgrade is unsuccessful, again relay event.  :suscat: :suscat: :suscat: :suscat: :suscat: :suscat: :suscat:
    3.4 All nodes finishes reporting relay and direct->relay transition testing finishes :checkmark: . Note that events generated by spurious upgrade/downgrade are still "in queue".
4. Then, according to test scenario, all nodes should transition back to direct, and test attempts to verify that
    4.1 First, tests waits for all nodes to report direct connection. At this moment beta and eta the event generated by spurious upgrade is taken as transition to direct. While in-fact this link is in "relay".
    4.2 At this moment, as part of verification if the connection is truly direct -> connection to DERPs is cut off. Because eta and beta was still in relayed state -> they loose meshnet connection completely now.
    4.3 As part of "checking if connection is truly direct" now, from each node attempts to ping each other node, and tests waits until all has received pinging response. As beta and eta has its meshnet connection broken - pings cannot go through, and without DERPs, they cannot reestablish any connection. Hence the test times out :x:

### Spurious upgrade
Spurious upgrade is caused by the fact, that after a downgrade the Endpoint is not invalidated due to update reason being `UpdateReason::Push` and thus not satisfying the following conditions: https://github.com/NordSecurity/libtelio/blob/38348ee809f867c7cf8e22dcd6d2795295f96143/src/device/wg_controller.rs#L978.

## Resolution
1. This PR refactors endpoint invalidation logic into separate stateless consolidate call. 
2. This PR updates the conditions which triggers invalidation, mainly:
    2.1 If we have downgraded - invalidate any endpoints published before downgrade
    2.2 If we have roamed - invalidate any endpoints which were published before roaming

Note: If we are doing a "standard" upgrade to direct endpoint and thus - writing direct endpoints into WG, the `UpdateReason` will be `Push`. And in this case we do *not* want to invalidate endpoints, as there is no indication, that hose are invalid.

- [x] README.md is updated

### DIFF
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -81,6 +81,12 @@ pub async fn consolidate_wg_state(
     )
     .await?;
     consolidate_wg_fwmark(requested_state, &*entities.wireguard_interface).await?;
+    consolidate_endpoint_invalidation(
+        &*entities.wireguard_interface,
+        entities.meshnet.left().map(|m| &*m.proxy),
+        entities.cross_ping_check(),
+    )
+    .await?;
     consolidate_wg_peers(
         requested_state,
         &*entities.wireguard_interface,
@@ -107,6 +113,13 @@ pub async fn consolidate_wg_state(
     )
     .boxed()
     .await?;
+    consolidate_endpoint_invalidation(
+        &*entities.wireguard_interface,
+        entities.meshnet.left().map(|m| &*m.proxy),
+        entities.cross_ping_check(),
+    )
+    .await?;
+
     let starcast_pub_key = if let Some(starcast_vpeer) = entities.starcast_vpeer() {
         Some(starcast_vpeer.get_peer().await?.public_key)
     } else {
@@ -151,6 +164,104 @@ async fn consolidate_wg_fwmark<W: WireGuard>(
     if actual_fwmark != requested_fwmark {
         wireguard_interface.set_fwmark(requested_fwmark).await?;
     }
+    Ok(())
+}
+
+async fn consolidate_endpoint_invalidation<W: WireGuard, P: Proxy, C: CrossPingCheckTrait>(
+    wireguard_interface: &W,
+    proxy: Option<&P>,
+    cross_ping_check: Option<&Arc<C>>,
+) -> Result {
+    // For each peer, we have to notify CPC and Endpoint Providers if
+    // some endpoint failed.
+    //    Endpoint failure is considered to be one of two things:
+    //    * Endpoint has been downgraded for any reason
+    //    * Remote node has roamed to different endpoint
+    //
+    //    Endpoint is *not* considered failed if endpoint has changed in
+    //    WireGuard by writing to it.
+    //
+    // Note:  Invalidation must be done before WireGuard peer consolidation,
+    //        as we must *not* consider endpoints-to-be-invalidated as eligible
+    //        for upgrade
+    // Note2: But, on the other hand, if WireGuard Peers consolidation makes
+    //        a decision to downgrade its connection - running this endpoint
+    //        invalidation just before WG peer consolidation would delay endpoint
+    //        invalidation until the next consolidation cycle. This is undesirable,
+    //        therefore endpoint invalidation checks should be run before *and*
+    //        after WireGuard peer consolidation.
+
+    // Retreive some helper values
+    let (proxy_endpoints, cpc, checked_endpoints) = match (proxy, cross_ping_check) {
+        (Some(proxy), Some(cpc)) => (
+            proxy.get_endpoint_map().await?,
+            cpc,
+            cpc.get_validated_endpoints().await?,
+        ),
+        _ => {
+            return Ok(()); // Meshnet is disabled
+        }
+    };
+    let actual_peers = wireguard_interface.get_interface().await?.peers;
+
+    // Each peer endpoint invalidation is processed individually
+    for (public_key, actual_peer) in actual_peers {
+        let is_proxying = is_peer_proxying(&actual_peer, &proxy_endpoints);
+        let actual_endpoint = actual_peer.endpoint;
+        let changed_at = actual_peer.endpoint_changed_at;
+        let published_endpoint = checked_endpoints.get(&public_key);
+        let published_at = published_endpoint.map(|endpoint| endpoint.changed_at);
+        let mut should_invalidate_endpoints = false;
+
+        // At debug level -> log every endpoint invalidation consideration. For info - only once a
+        // decision to invalidate has been taken
+        telio_log_debug!(
+            "Considering endpoint invalidation.\
+                         Decision info: {is_proxying:?} \
+                         {actual_endpoint:?} {changed_at:?} \
+                         {published_endpoint:?} {published_at:?}"
+        );
+
+        // If endpoint has never changed in wireguard or we have no published endpoints - there is
+        // nothing to invalidate.
+        let (changed_at, published_at) = match (changed_at, published_at) {
+            (Some(changed_at), Some(published_at)) => (changed_at, published_at),
+            _ => continue,
+        };
+
+        // Any endpoint published before an actual downgrade has occured is considered failed
+        if is_proxying && (published_at < changed_at.0) {
+            let change_reason = changed_at.1;
+            telio_log_info!(
+                "Invalidating published EPs for {public_key:?} after downgrade.\
+                            Decision info: {actual_endpoint:?} {changed_at:?} {change_reason:?}\
+                            {published_endpoint:?} {published_at:?}"
+            );
+            should_invalidate_endpoints = true;
+        }
+
+        // In direct connection if we have roamed from one direct EP to another - old checked EP is
+        // considered not valid anymore.
+        // Note, that endpoint change caused by libtelio writing new endpoint into WireGuard interface
+        // is *not* considered roaming.
+        if !is_proxying && changed_at.1 == UpdateReason::Pull && (published_at < changed_at.0) {
+            telio_log_info!(
+                "Invalidating published EPs for {public_key:?} after roaming.\
+                            Decision info: {actual_endpoint:?} {changed_at:?} \
+                            {published_endpoint:?} {published_at:?}"
+            );
+            should_invalidate_endpoints = true;
+        }
+
+        if should_invalidate_endpoints {
+            if let Err(err) = cpc.notify_failed_wg_connection(public_key).await {
+                telio_log_warn!(
+                    "Failed to send connection failed report for {public_key}: {err:?}"
+                );
+            }
+        }
+    }
+
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


src/device/wg_controller.rs
```diff
@@ -2189,128 +2313,269 @@ mod tests {
...
-        f.requested_state.keepalive_periods.direct = direct_keepalive_period;
 
-        let cpc_endpoint_at = Instant::now();
-        let changed_endpoint_at = cpc_endpoint_at - Duration::from_secs(1);
+        let mapped_port = 12;
+        let changed_endpoint_at = Instant::now();
+        let cpc_endpoint_at = changed_endpoint_at - Duration::from_secs(1);
 
-        f.when_requested_meshnet_config(vec![(pub_key, allowed_ips.clone())]);
+        //Arrange proxy mappings
```
gytsto, line 2435:
```suggestion
        // Arrange proxy mappings
```

src/device/wg_controller.rs
```diff
@@ -2189,128 +2313,269 @@ mod tests {
...
+        let pub_key = SecretKey::gen().public();
+        let allowed_ips = Fixture::create_any_allowed_ips();
+        let remote_wg_endpoint = SocketAddr::from(([192, 168, 0, 1], 13));
+        let local_wg_endpoint = SocketAddr::from(([192, 168, 0, 2], 15));
+
+        let mapped_port = 12;
+        let changed_endpoint_at = Instant::now();
+        let cpc_endpoint_at = changed_endpoint_at - Duration::from_secs(1);
+
+        //Arrange proxy mappings
```
gytsto, line 2481:
```suggestion
        // Arrange proxy mappings
```

src/device/wg_controller.rs
```diff
@@ -2189,128 +2313,269 @@ mod tests {
...
 
-        let direct_keepalive_period = 1234;
-        f.requested_state.keepalive_periods.direct = direct_keepalive_period;
-
-        let cpc_endpoint_at = Instant::now();
-        // changed_endpoint >= cpc_changed_endpoint
-        let changed_endpoint_at = cpc_endpoint_at;
-
-        f.when_requested_meshnet_config(vec![(pub_key, allowed_ips.clone())]);
+        //Arrange proxy mappings
```
gytsto, line 2525:
```suggestion
        // Arrange proxy mappings
```

src/device/wg_controller.rs
```diff
@@ -154,6 +167,103 @@ async fn consolidate_wg_fwmark<W: WireGuard>(
...
+    //        as we must *not* consider endpoints-to-be-invalidated as eligible
+    //        for upgrade
+    // Note2: But, on the other hand, if WireGuard Peers consolidation makes
+    //        a decision to downgrade its connection - running this endpoint
+    //        invalidation just before WG peer consolidation would delay endpoint
+    //        invalidation until the next consolidation cycle. This is undesirable,
+    //        therefore endpoint invalidation checks should be run before *and*
+    //        after WireGuard peer consolidation.
+
+    //Retreive some helper values
```
gytsto, line 194:
```suggestion
    // Retreive some helper values
```

src/device/wg_controller.rs
```diff
@@ -154,6 +167,103 @@ async fn consolidate_wg_fwmark<W: WireGuard>(
...
+                            {published_endpoint:?} {published_at:?}"
+            );
+            should_invalidate_endpoints = true;
+        }
+
+        // In direct connection if we have roamed from one direct EP to another - old checked EP is
+        // considered not valid anymore.
+        // Note, that endpoint change caused by libtelio writing new endpoint into WireGuard interface
+        // is *not* considered roaming.
+        if !is_proxying && changed_at.1 == UpdateReason::Pull && (published_at < changed_at.0) {
```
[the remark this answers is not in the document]
Jauler: I had this implemented with variables first, but wanted to add specific reason (roaming vs downgrade) in logs which triggers invalidation. And then it becomes harder to avoid if's :/

src/device/wg_controller.rs
```diff
@@ -154,6 +167,103 @@ async fn consolidate_wg_fwmark<W: WireGuard>(
...
+        );
+
+        // If endpoint has never changed in wireguard or we have no published endpoints - there is
+        // nothing to invalidate.
+        let (changed_at, published_at) = match (changed_at, published_at) {
+            (Some(changed_at), Some(published_at)) => (changed_at, published_at),
+            _ => continue,
+        };
+
+        // Any endpoint published before a actual downgrade has occured is considered failed
```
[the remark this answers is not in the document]
Jauler: fixed

src/device/wg_controller.rs
```diff
@@ -2189,128 +2313,269 @@ mod tests {
...
+            let mut f = Fixture::new();
+            f.features.ipv6 = true;
+
+            let pub_key = SecretKey::gen().public();
+            let allowed_ips = Fixture::create_any_allowed_ips();
+            let remote_wg_endpoint = SocketAddr::from(([192, 168, 0, 1], 13));
+            let changed_endpoint_at = Instant::now();
+            let cpc_endpoint_at = changed_endpoint_at + Duration::from_secs(1);
+
+            //Arrange proxy mappings
```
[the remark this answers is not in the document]
Jauler: fixed

src/device/wg_controller.rs
Jauler: fixed

src/device/wg_controller.rs
Jauler: fixed

src/device/wg_controller.rs
Jauler: fixed

src/device/wg_controller.rs
```diff
@@ -2189,128 +2313,269 @@ mod tests {
...
+            let allowed_ips = Fixture::create_any_allowed_ips();
+            let remote_wg_endpoint = SocketAddr::from(([192, 168, 0, 1], 13));
+            let local_wg_endpoint = SocketAddr::from(([192, 168, 0, 2], 15));
+
+            let mapped_port = 12;
+            let proxy_endpoint = SocketAddr::from(([127, 0, 0, 1], mapped_port));
+            let changed_endpoint_at = Instant::now();
+            let cpc_endpoint_at = changed_endpoint_at - Duration::from_secs(1);
+
+            //Arrange proxy mappings
```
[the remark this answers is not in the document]
Jauler: fixed

src/device/wg_controller.rs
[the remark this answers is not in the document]
Jauler: oh, but I missed the "roaming" word itself :D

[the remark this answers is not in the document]
Jauler: Updated logging lines. Although only subtly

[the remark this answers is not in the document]
gytsto: missed that diff before :D

gytsto (approved):
+1

src/device/wg_controller.rs
```diff
@@ -2189,128 +2314,269 @@ mod tests {
...
+        }
+    }
+
+    #[tokio::test]
+    async fn when_validated_ep_is_pushed_into_wg_produce_no_failed_notification() {
+        // EndpointGone notification should be produced regardless wether we
+        // have pushed the endpoint into to the WG ourselves, or read it from
+        // wireguard_interface after roaming.
+        //
+        // Test for both cases
```
mathiaspeters, line 2423: This comment doesn't seem like it applies here

src/device/wg_controller.rs
```diff
@@ -2189,128 +2314,269 @@ mod tests {
...
+            f.then_cross_ping_check_not_notfied_about_failed_connections(vec![pub_key]);
+
+            // Act
+            f.consolidate_endpoint_invalidation().await;
+        }
+    }
+
+    #[tokio::test]
+    async fn when_ep_is_validated_before_downgrade_produce_failed_notification() {
+        // EndpointGone notification should be produced regardless wether we
```
mathiaspeters, line 2368: Incredibly minor nit, only fix if fixing something else:
```suggestion
        // EndpointGone notification should be produced regardless whether we
```

src/device/wg_controller.rs
```diff
@@ -2189,128 +2314,269 @@ mod tests {
...
+        // Prepare assertions
+        f.then_cross_ping_check_not_notfied_about_failed_connections(vec![pub_key]);
+
+        // Act
+        f.consolidate_endpoint_invalidation().await;
+    }
+
+    #[tokio::test]
+    async fn when_roaming_from_direct_to_direct_produce_failed_notification() {
+        // EndpointGone notification should be produced regardless wether we
```
mathiaspeters, line 2465: Same nit as before:
```suggestion
        // EndpointGone notification should be produced regardless whether we
```

mathiaspeters: Also, the test only covers the `Pull` case, so maybe whole comment is wrong

src/device/wg_controller.rs
```diff
@@ -2189,128 +2314,269 @@ mod tests {
...
+        f.consolidate_endpoint_invalidation().await;
+    }
+
+    #[tokio::test]
+    async fn when_meshnet_disabled_produce_no_failed_notification() {
+        // EndpointGone notification should be produced regardless wether we
+        // have pushed the endpoint into to the WG ourselves, or read it from
+        // wireguard_interface after roaming.
+        //
+        // Test for both cases
```
mathiaspeters, line 2554: Here too

mathiaspeters (approved):
Ship it! +1

PR #901, opened 2024-10-25
Spawn event callback calling task loop on a separate thread

tomaszklak:
### Problem
Analysis of `test_event_link_state_peer_goes_offline` failure suggests that event generation happened in rust code but was not visible on the python side.

### Solution
To make sure that nothing is blocking events processing, new separate thread is started only for that purpose. To 
to make sure that the event was indeed generated, new debug log is added.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -707,13 +707,17 @@ impl State {
             }
         });
 
+        let event = Event {
+            state,
+            link_state,
+            peer,
+            old_peer,
+        };
+
+        telio_log_debug!("Sending new event {event:?}");
+
         self.event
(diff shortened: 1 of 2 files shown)


Hasan6979:
lgtm +1
Just wondering why couldn't we use `spawn_blocking` for this as that would also be clear of the async runtime ?

tomaszklak:
> lgtm +1 Just wondering why couldn't we use `spawn_blocking` for this as that would also be clear of the async runtime ?

We could, but this way we can name the thread and since this is going to be a long running thread it might be useful to distinguish it from other tokio threads.

tomasz-grz (approved):
+1 ship it!

PR #907, opened 2024-10-28
Add some retry logic to upnpc

matislovas:
### Problem
`upnpc` command is prone to race condition, because we're resetting `upnpd` before that.

### Solution
Add retry logic for `upnpc`.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_upnp_connection.py
+++ b/nat-lab/tests/test_upnp_connection.py
@@ -14,7 +14,21 @@ from utils.bindings import (
 )
 from utils.connection_util import ConnectionTag
 from utils.ping import ping
+from utils.process import ProcessExecError
 from utils.router import new_router, IPStack
+
+
+async def execute_upnpc_with_retry(connection, timeout=10.0):
+    start_time = asyncio.get_event_loop().time()
+    while True:
+        try:
+            return await connection.create_process(["upnpc", "-i", "-l"]).execute()
+        except ProcessExecError as e:
+            if asyncio.get_event_loop().time() - start_time >= timeout:
+                print(f"Timeout while waiting for upnpc to start: {str(e)}")
+                raise
+
+        await asyncio.sleep(1.0)
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


Jauler (approved):
+1

mathiaspeters (approved):
+1

nat-lab/tests/test_upnp_connection.py
```diff
@@ -17,6 +17,22 @@
 from utils.router import new_router, IPStack
 
 
+async def execute_upnpc_with_retry(connection, timeout=10.0):
+    start_time = asyncio.get_event_loop().time()
+    while True:
+        cmd = await connection.create_process(["upnpc", "-i", "-l"]).execute()
```
gytsto, line 23: This would raise an exception, breaking your loop and test in general. Instead of checking for return code, you have to catch `ProcessExecError` and then only try retry logic.

lcruz99:
Originally I've chosen 40s/5s of `lease_duration_s`/`endpoint_interval` thinking it would avoid these kind of errors. With this alternative of yours you can reduce those values to something like 15s/3s.

nat-lab/tests/test_upnp_connection.py
matislovas: Wrapped

matislovas:
> Originally I've [chosen](https://github.com/NordSecurity/libtelio/pull/907/files#diff-d3cfb426a998ac2e72864f389c4d78fe2d26538c87cbe7fd388fe4d73d1f9875R238) 40s/5s of `lease_duration_s`/`endpoint_interval` thinking it would avoid these kind of errors. With this alternative of yours you can reduce those values to something like 15s/3s.

But wouldn't this increase the risk of failing test, when running on slow VM ..?

nat-lab/tests/test_upnp_connection.py
```diff
@@ -14,9 +14,31 @@
...
 from utils.ping import ping
+from utils.process import ProcessExecError
 from utils.router import new_router, IPStack
 
 
+async def execute_upnpc_with_retry(connection, timeout=10.0):
+    start_time = asyncio.get_event_loop().time()
+    while True:
+        try:
+            cmd = await connection.create_process(["upnpc", "-i", "-l"]).execute()
```
gytsto, line 25: I think we can simplify it abit. No need to check return code, since it will be zero or 137 (sigkill which is mostly caused by leaving stack == test cleanup), otherwise exception will be raised.
```
async def execute_upnpc_with_retry(connection, timeout=10.0):
    start_time = asyncio.get_event_loop().time()
    while True:
        try:
            return await connection.create_process(["upnpc", "-i", "-l"]).execute()
        except ProcessExecError as e:
            if asyncio.get_event_loop().time() - start_time >= timeout:
                print(f"Timeout while waiting for upnpc to start: {str(e)}")
                raise

        await asyncio.sleep(1.0)
```

matislovas: Fixed

gytsto (approved):
+1

PR #908, opened 2024-10-28
[LLT-5604] Notifications token support in CoreApi

olekoliinyk:
### Problem
In the ESP32 mesh network (meshnet) integration tests within the nat-lab environment, the v1/notifications/token endpoint must be supported by the nat-lab CoreApi to provide essential connection details for the MQTT message broker. This endpoint requires Basic authentication for access control, while all other CoreApi endpoints use Bearer token authentication.

### Solution
1) Implement `v1/notifications/token` Endpoint: Add the `v1/notifications/token` endpoint to provide connection details for the MQTT broker, based on the specifications outlined in the `v1/notifications/token` documentation.
2) Enable Basic Authentication for `v1/notifications/token`: Add Basic authentication support for the `v1/notifications/token` endpoint. Additionally, create and add test user accounts to validate authentication.
3) Add Bearer Token Authentication for Other Endpoints: For all endpoints except `v1/notifications/token`, add Bearer token authentication support. Generate a set of reusable tokens for testing purposes.
4) Configure MQTT Broker Credentials: Set up two credentials for the MQTT broker: one for CoreApi access and another for a subscriber, which will be returned by the `v1/notifications/token` endpoint.
5) Update Existing Tests for Authentication: Modify the existing tests to incorporate either Bearer token or Basic authentication as appropriate.
6) Relocate MQTT Listener for Testing Compatibility: Move the MQTT listener from `test_notification_center.py` to run within the cone-client to facilitate proper notification center testing. Since the data from `v1/notifications/token ` includes a broker domain name that the test execution host cannot resolve, this adjustment allows cone-client (with necessary Docker setup) to handle resolution.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/core-api.py
+++ b/nat-lab/bin/core-api.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-
+import base64
 import json
 import paho.mqtt.client as mqtt  # type: ignore # pylint: disable=import-error
 import ssl
@@ -25,10 +25,26 @@ DERP_SERVER = {
 
 CERTIFICATE_PATH = "/etc/ssl/server_certificate/server.pem"
 
+MQTT_BROKER_HOST = "mqtt.nordvpn.com"
+MQTT_BROKER_PORT = 8883
+
+# Below credentials are intended only for testing purposes in nat-lab environment.
+CORE_API_CREDENTIALS = {
+    "username": "token",
+    "password": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
+}
+MQTT_CREDENTIALS = {
+    "username": "mqtt_broker",
+    "password": "9-A'.:vUM3FPTCABorsK}J4mM}/3898_",
+}
+
 
 class CoreApiErrorCode(Enum):
     MACHINE_ALREADY_EXISTS = 101117
     MACHINE_NOT_FOUND = 101102
+    INVALID_CREDENTIALS = 100104
+    AUTHORIZATION_HEADER_NOT_PROVIDED = 100105
+    AUTHORIZATION_HEADER_INVALID = 100106
 
 
 @dataclass
@@ -105,6 +121,7 @@ class CoreApiHandler(BaseHTTPRequestHandler):
     def __init__(self, request, client_address, server: CoreServer):
         self.server: CoreServer
         self.machines_path = "/v1/meshnet/machines"
+        self.notifications_path = "/v1/notifications/tokens"
         super().__init__(request, client_address, server)
 
     def _set_headers(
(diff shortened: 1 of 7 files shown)


nat-lab/tests/test_core_api.py
```diff
@@ -49,6 +51,8 @@ class CoreApiErrorCode(Enum):
     "weight": int,
 }
 
+bearer_authorization_header = "Bearer IronKey"
```
stalowyjez, line 54: Afaik it should be more like `Bearer token:IronKey`

nat-lab/bin/mqtt-listener.py
```diff
@@ -0,0 +1,51 @@
...
+    mqttc.loop_forever()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="MQTT Client")
+
+    parser.add_argument("mqtt_broker_host", type=str, help="MQTT broker host")
+    parser.add_argument("mqtt_broker_port", type=int, help="MQTT broker port")
+    parser.add_argument("mqtt_broker_user", type=str, help="MQTT broker user")
+    parser.add_argument("mqtt_broker_password", type=str, help="MQTT broker password")
```
stalowyjez, line 43: Maybe it would be simpler to just make this `notify/tokens` HTTP call here? I think it would be a small change which would make it both simpler and closer to the desired state.

nat-lab/tests/test_core_api.py
```diff
@@ -345,3 +359,114 @@ async def not_able_to_register_same_machine_twice(server_host):
...
+            response_data = await send_https_request(
+                connection,
+                url,
+                method,
+                CORE_API_CA_CERTIFICATE_PATH,
+                authorization_header=authorization_header,
+                data=payload,
+            )
+        assert response_data["errors"]["code"] == CoreApiErrorCode.INVALID_CREDENTIALS
+        assert response_data["errors"]["message"] == "Authorization header not provided"
```
stalowyjez, line 444: Shouldn't it be "Invalid credentials"?

nat-lab/bin/core-api.py
```diff
@@ -19,16 +19,61 @@
...
+
+
+@dataclass
+class NotificationTokenRequest:
+    user_hash = str
+    app_user_id = str
+    platform_id = int
+    pns_id: Optional[int] = None
+    pns_token: Optional[str] = None
+    protocol = str
```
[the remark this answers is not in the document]
olekoliinyk: Yep, missed it.

nat-lab/tests/test_notification_center.py
```diff
@@ -24,6 +23,14 @@ class MachineResponse:
     traffic_routing_supported: bool
 
 
+bearer_authorization_header = "Bearer StellarPass"
```
matislovas, line 26: Maybe `Bearer token:StellarPass`?

nat-lab/bin/mqtt-listener.py
```diff
@@ -0,0 +1,51 @@
+import argparse
```
matislovas, line 1: Maybe it is worth adding shebang here, and run it with default python ..?

nat-lab/bin/core-api.py
```diff
@@ -19,16 +19,61 @@
...
+ALLOWED_BEARER_TOKENS = ["IronKey", "StellarPass", "ShadowCipher"]
+MQTT_USERS = {
+    "coreApi": {
+        "username": "core_api",
+        "password": "core@pi",
+    },
+    "subscriber": {
+        "username": "subscriber",
+        "password": "topicSubscriber",
+    },
```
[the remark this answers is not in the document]
matislovas: You cannot send the message to specified `username` or `user_id` in MQTT. I guess in prod case, there should be a broker instance per meshnet, where core-api sends a message two, and it distributes it to all clients ..?

nat-lab/tests/test_core_api.py
olekoliinyk: I think that's the case for basic authentication.
In case of bearer token the header should be as following:
`"Authorization: Bearer <token>"`

nat-lab/bin/mqtt-listener.py
olekoliinyk: IMHO, I would prefer to keep `v1/notifications/token` call in `test_notification_center` since we're doing the validation of the response there and it's supposed to be a part of the test

nat-lab/bin/core-api.py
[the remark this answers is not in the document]
olekoliinyk: From core-api nat-lab perspective having one account for MQTT would be sufficient at that moment. I think we're going operate within one user account for the esp32-meshnet.

nat-lab/tests/test_notification_center.py
olekoliinyk: Same comment and discussion here:
https://github.com/NordSecurity/libtelio/pull/908#discussion_r1840489068

matislovas:
+1

nat-lab/tests/test_notification_center.py
```diff
@@ -35,23 +42,56 @@ def verify_mqtt_payload(payload):
...
+        authorization_header=basic_authorization_header,
+    )
+    assert "endpoint" in response
+    assert "username" in response
+    assert "password" in response
+    assert "expires_in" in response
+
+    mqtt_data = response["endpoint"].replace("//", "").split(":")
+    host = mqtt_data[1]
+    port = mqtt_data[2]
```
stalowyjez, line 91: Nit: tbh you could just do `_, host, port = ` (what is in 0, though?)

nat-lab/tests/test_core_api.py
olekoliinyk: You're right. It's `Bearer token:<token>`

nat-lab/tests/test_notification_center.py
```diff
@@ -35,23 +42,56 @@ def verify_mqtt_payload(payload):
...
+            "python3",
+            "/opt/bin/mqtt-listener.py",
+            mqtt_broker_host,
+            mqtt_broker_port,
+            mqtt_broker_user,
+            mqtt_broker_password,
+        ]).run()
+    )
+    await sleep(1)
+    return mqtt_process
```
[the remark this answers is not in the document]
olekoliinyk: Agree that previous solution was cleaner. However the data returned by `v1/notifications/token` contains domain name which should be mapped to the nat-lab mqtt broker container ip and locally it can be done via editing the host  (which might be not a good solution).

nat-lab/tests/test_notification_center.py
olekoliinyk: 0 - is protocol

nat-lab/tests/test_notification_center.py
```diff
@@ -35,23 +50,96 @@ def verify_mqtt_payload(payload):
...
+            mqtt_broker_host,
+            mqtt_broker_port,
+            mqtt_broker_user,
+            mqtt_broker_password,
+        ]).run()
+    )
+    return await wait_for_mqtt_executing(mqtt_process)
+
+
+def wait_with_timeout(max_wait=5, initial_wait_time=0.1, factor=2, on_timeout=None):
```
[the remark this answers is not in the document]
olekoliinyk: Yeah, it may look overcomplicated for such a simple test, but we will have a clear indication of what failed. If we have just basic sleep, then the pytest session may fail with the timeout in case process doesn't return anything.

nat-lab/bin/core-api.py
```diff
@@ -124,16 +153,89 @@ def _send_error_response(
...
+        return False
+
+    @staticmethod
+    def requires_basic_authentication(func):
+        def wrapper(self, *args, **kwargs):
+            if not self.validate_basic_authorization():
+                return None
+            return func(self, *args, **kwargs)
+
+        return wrapper
```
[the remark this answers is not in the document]
olekoliinyk: There might be a requirement to add some other endpoints that require basic authentication (just an assumption as of now) and this decorator can be re-used then.

nat-lab/tests/test_notification_center.py
```diff
@@ -35,23 +50,96 @@ def verify_mqtt_payload(payload):
...
+
+@wait_with_timeout(on_timeout="MQTT process didn't start within max wait time.")
+async def wait_for_mqtt_executing_condition(mqtt_process):
+    if mqtt_process.is_executing():
+        return mqtt_process
+    return None
+
+
+async def wait_for_mqtt_output(mqtt_process, target_string, **kwargs):
+    return await wait_for_mqtt_output_condition(mqtt_process, target_string, **kwargs)
```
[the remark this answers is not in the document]
olekoliinyk: They were removed :)

nat-lab/tests/test_notification_center.py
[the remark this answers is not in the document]
olekoliinyk: After discussion offline, we decided not to overcomplicate this test and use basic `while with sleep`

stalowyjez (approved):
Looks good to me, +1

PR #909, opened 2024-10-29
LLT-5683: Increase constrains in link state testcases

dfetti:
Previously the link state testcase scenarios were structured like setup, do some interaction between peers, sleep for a while and check the reported link states. This approach did not consider the timings of the events.

Proposed solution consists in checking the state constantly during the scenario, not only at the end of it.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5683
+++ b/.unreleased/LLT-5683
@@ -1,1 +1,0 @@
-Increase constrains in link state testcases
--- a/.unreleased/LLT-5683
+++ b/.unreleased/LLT-5683
@@ -1,0 +1,1 @@
+Increase constrains in link state testcases
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -130,10 +130,11 @@ class Runtime:
                 return
             await asyncio.sleep(0.1)
 
-    async def notify_link_state(self, public_key: str, states: List[LinkState]) -> None:
+    async def notify_link_state(self, public_key: str, state: LinkState) -> None:
+        """Wait until a link_state event matching the `state` for `public_key` is available."""
         while True:
             peer = self.get_peer_info(public_key)
-            if peer and peer.link_state in states:
+            if peer and peer.link_state == state:
                 return
             await asyncio.sleep(0.1)
 
@@ -165,11 +166,11 @@ class Runtime:
                 return
             await asyncio.sleep(0.1)
 
-    def get_link_state_events(self, public_key: str) -> List[Optional[LinkState]]:
+    def get_link_state_events(self, public_key: str) -> List[LinkState]:
         return [
             peer.link_state
             for peer in self._peer_state_events
-            if peer and peer.public_key == public_key
+            if peer and peer.public_key == public_key and peer.link_state is not None
         ]
 
     def get_peer_info(self, public_key: str) -> Optional[TelioNode]:
@@ -273,9 +274,10 @@ class Events:
     async def wait_for_link_state(
         self,
         public_key: str,
-        state: List[LinkState],
+        state: LinkState,
         timeout: Optional[float] = None,
     ) -> None:
+        """Wait until a link_state event matching the `state` for `public_key` is available."""
         await asyncio.wait_for(
             self._runtime.notify_link_state(public_key, state), timeout
         )
@@ -294,7 +296,7 @@ class Events:
             timeout,
         )
 
-    def get_link_state_events(self, public_key: str) -> List[Optional[LinkState]]:
+    def get_link_state_events(self, public_key: str) -> List[LinkState]:
         return self._runtime.get_link_state_events(public_key)
 
     async def wait_for_state_derp(
@@ -570,12 +572,13 @@ class Client:
     async def wait_for_link_state(
         self,
         public_key: str,
-        states: List[LinkState],
+        state: LinkState,
         timeout: Optional[float] = None,
     ) -> None:
+        """Wait until a link_state event matching the `state` for `public_key` is available."""
         await self.get_events().wait_for_link_state(
             public_key,
-            states,
+            state,
             timeout,
         )
 
@@ -601,7 +604,7 @@ class Client:
         )
         print(datetime.now(), f"[{self._node.name}]: got event {event_info}")
 
-    def get_link_state_events(self, public_key: str) -> List[Optional[LinkState]]:
+    def get_link_state_events(self, public_key: str) -> List[LinkState]:
         return self.get_events().get_link_state_events(public_key)
 
     async def wait_for_state_derp(
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -130,6 +130,13 @@ class Runtime:
                 return
             await asyncio.sleep(0.1)
 
+    async def notify_link_state(self, public_key: str, states: List[LinkState]) -> None:
+        while True:
+            peer = self.get_peer_info(public_key)
+            if peer and peer.link_state in states:
+                return
+            await asyncio.sleep(0.1)
+
     async def notify_peer_event(
         self,
         public_key: str,
@@ -261,6 +268,16 @@ class Events:
                 public_key, state, paths, is_exit, is_vpn, link_state
             ),
             timeout,
+        )
+
+    async def wait_for_link_state(
+        self,
+        public_key: str,
+        state: List[LinkState],
+        timeout: Optional[float] = None,
+    ) -> None:
+        await asyncio.wait_for(
+            self._runtime.notify_link_state(public_key, state), timeout
         )
 
     async def wait_for_event_peer(
@@ -548,6 +565,18 @@ class Client:
             is_vpn,
             timeout,
             link_state,
+        )
+
+    async def wait_for_link_state(
+        self,
+        public_key: str,
+        states: List[LinkState],
+        timeout: Optional[float] = None,
+    ) -> None:
+        await self.get_events().wait_for_link_state(
+            public_key,
+            states,
+            timeout,
         )
 
     async def wait_for_event_peer(
--- a/nat-lab/tests/test_events_link_state.py
+++ b/nat-lab/tests/test_events_link_state.py
@@ -135,6 +135,16 @@ FEATURE_DISABLED_PARAMS = [
 ]
 
 
+async def wait_for_any_with_timeout(tasks, timeout: float):
+    done_tasks, _pending_tasks = await asyncio.wait(
+        tasks,
+        timeout=timeout,
+        return_when=asyncio.FIRST_COMPLETED,
+    )
+    if len(done_tasks) == 0:
+        raise asyncio.TimeoutError
+
+
 @pytest.mark.asyncio
 @pytest.mark.parametrize("setup_params", FEATURE_ENABLED_PARAMS)
 async def test_event_link_state_peers_idle_all_time(
@@ -145,14 +155,26 @@ async def test_event_link_state_peers_idle_all_time(
         alpha, beta = env.nodes
         client_alpha, client_beta = env.clients
 
-        # Expect no link event while peers are idle
-        await asyncio.sleep(20)
-        alpha_events = client_beta.get_link_state_events(alpha.public_key)
-        beta_events = client_alpha.get_link_state_events(beta.public_key)
+        await client_alpha.wait_for_link_state(beta.public_key, [LinkState.UP])
+        await client_beta.wait_for_link_state(alpha.public_key, [LinkState.UP])
 
-        # 1 down event when Connecting, 1 up event when Connected
-        assert alpha_events == [LinkState.DOWN, LinkState.UP]
-        assert beta_events == [LinkState.DOWN, LinkState.UP]
+        # Expect no link event while peers are idle
+        with pytest.raises(asyncio.TimeoutError):
+            await wait_for_any_with_timeout(
+                [
+                    asyncio.create_task(
+                        client_alpha.wait_for_link_state(
+                            beta.public_key, [LinkState.DOWN]
+                        )
+                    ),
+                    asyncio.create_task(
+                        client_beta.wait_for_link_state(
+                            alpha.public_key, [LinkState.DOWN]
+                        )
+                    ),
+                ],
+                timeout=25,
+            )
 
 
 @pytest.mark.asyncio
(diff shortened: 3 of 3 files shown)


nat-lab/tests/test_events_link_state.py
```diff
@@ -140,14 +150,26 @@ async def test_event_link_state_peers_idle_all_time(
         alpha, beta = env.nodes
         client_alpha, client_beta = env.clients
 
-        # Expect no link event while peers are idle
-        await asyncio.sleep(20)
-        alpha_events = client_beta.get_link_state_events(alpha.public_key)
-        beta_events = client_alpha.get_link_state_events(beta.public_key)
+        await client_alpha.wait_for_link_state(beta.public_key, [LinkState.UP])
+        await client_beta.wait_for_link_state(alpha.public_key, [LinkState.UP])
```
LukasPukenis, line 154: These can be executed in parallel as if `alpha` will emit link state event sooner, say at t=0 and at t=10 `beta` emits, then `alpha` will be delayed

nat-lab/tests/telio.py
```diff
@@ -130,6 +130,13 @@ async def notify_peer_state(
                 return
             await asyncio.sleep(0.1)
 
+    async def notify_link_state(self, public_key: str, states: List[LinkState]) -> None:
+        while True:
+            peer = self.get_peer_info(public_key)
+            if peer and peer.link_state in states:
```
tomasz-grz, line 136: Would this return for any first matching state?

nat-lab/tests/test_events_link_state.py
```diff
@@ -195,21 +214,54 @@ async def test_event_link_state_peers_exchanging_data_then_idling_then_resume(
...
+                            beta.public_key, [LinkState.DOWN]
+                        )
+                    ),
+                    asyncio.create_task(
+                        client_beta.wait_for_link_state(
+                            alpha.public_key, [LinkState.DOWN]
+                        )
+                    ),
+                ],
+                timeout=25,
```
LukasPukenis, line 238: Just noting that `pytest.raises(asyncio.TimeoutError)` are good to be left with an explicit `timeout` value. On Windows which is producing delays this is still gonna pass and on containers that have no delays it will  assert for correctness(things don't happen earlier than expected)

nat-lab/tests/test_events_link_state.py
```diff
@@ -168,18 +190,15 @@ async def test_event_link_state_peers_exchanging_data_for_a_long_time(
...
             await ping(connection_beta, alpha.ip_addresses[0])
-
-        # Expect no nolink event while peers are active
-        alpha_events = client_beta.get_link_state_events(alpha.public_key)
-        beta_events = client_alpha.get_link_state_events(beta.public_key)
-
-        # 1 down event when Connecting, 1 up event when Connected
-        assert alpha_events == [LinkState.DOWN, LinkState.UP]
-        assert beta_events == [LinkState.DOWN, LinkState.UP]
+            await client_alpha.wait_for_link_state(beta.public_key, [LinkState.UP], 1)
```
LukasPukenis, line 200: Just putting in writing what we already talked about `wait_for_link_state` should have no explicit timeout set or else it will flake on Windows test runners :/

nat-lab/tests/test_events_link_state.py
```diff
@@ -225,24 +277,28 @@ async def test_event_link_state_peer_goes_offline(
...
+                beta.public_key, [LinkState.DOWN], 10
+            )
 
-        # 1 down event when Connecting, 1 up event when Connected
-        assert alpha_events == [LinkState.DOWN, LinkState.UP]
-        # 1 down event when Connecting, 1 up event when Connected, 1 down event when client is stopped
-        assert beta_events == [LinkState.DOWN, LinkState.UP, LinkState.DOWN]
+        # Expect the link down event
+        # It should arrive in 11-15 seconds after the link is cut and ping mod disabled
+        # And 22-25 seconds if the ping mod is enabled
```
LukasPukenis, line 298: I am mentioning for general context and also because these comments are a bit ambiguous since the values are not exact but rather ranges which might be confusing now looking at it :)

Link detection is based on 10seconds which is a guarantee from WireGuard(passive-keepalive) and additional configurable RTT - see `_generate_setup_parameter_pair()` which now is `1` so `11` second in total.

If enhanched detection is used then `icmp echo` aka ping is also used additionally before reporting the state which adds some extra time(forgot how much).

nat-lab/tests/test_events_link_state.py
```diff
@@ -225,24 +277,28 @@ async def test_event_link_state_peer_goes_offline(
...
         await client_beta.stop_device()
 
-        await asyncio.sleep(1)
-
         with pytest.raises(asyncio.TimeoutError):
-            await ping(connection_alpha, beta.ip_addresses[0], 5)
+            await ping(connection_alpha, beta.ip_addresses[0], 3)
 
-        await asyncio.sleep(25)
-        alpha_events = client_beta.get_link_state_events(alpha.public_key)
```
tomasz-grz, line 239: Maybe we would still want to call `get_link_state_events()` right after `notify_link_state()` just to make sure we have the correct sequence of events [LinkState.DOWN, LinkState.UP, LinkState.DOWN]?

nat-lab/tests/test_events_link_state.py
```diff
@@ -320,24 +376,28 @@ async def test_event_link_state_peer_doesnt_respond(
...
+
         async with ICMP_control(connection_beta):
             with pytest.raises(asyncio.TimeoutError):
-                await ping(connection_alpha, beta.ip_addresses[0], 8)
-
-            alpha_events = client_beta.get_link_state_events(alpha.public_key)
-            beta_events = client_alpha.get_link_state_events(beta.public_key)
-
-            # The connection is normal and events should be: initial down, then several up events but no more down events
-            assert alpha_events.count(LinkState.DOWN) == 1
```
tomasz-grz, line 331: Personally I think checking the count in tests is really good practice

nat-lab/tests/test_events_link_state.py
```diff
@@ -145,14 +155,26 @@ async def test_event_link_state_peers_idle_all_time(
         alpha, beta = env.nodes
         client_alpha, client_beta = env.clients
 
-        # Expect no link event while peers are idle
-        await asyncio.sleep(20)
-        alpha_events = client_beta.get_link_state_events(alpha.public_key)
-        beta_events = client_alpha.get_link_state_events(beta.public_key)
+        await client_alpha.wait_for_new_link_state(beta.public_key, [LinkState.UP])
```
tomasz-grz, line 158: This could also be some shared precondition for every test case to follow DRY principles

nat-lab/tests/test_events_link_state.py
```diff
@@ -168,18 +190,22 @@ async def test_event_link_state_peers_exchanging_data_for_a_long_time(
             conn.connection for conn in env.connections
         ]
 
+        await client_alpha.wait_for_new_link_state(beta.public_key, [LinkState.UP])
+        await client_beta.wait_for_new_link_state(alpha.public_key, [LinkState.UP])
+
         for _ in range(0, 40):
             await asyncio.sleep(1)
```
tomasz-grz, line 197: 40 seconds of sleep 🛌

nat-lab/tests/test_events_link_state.py
olekoliinyk: We just omitting the very first event (LinkState.DOWN), which I believe the default one when the node just created?

nat-lab/tests/test_events_link_state.py
olekoliinyk: Should we have this sleep here?

nat-lab/tests/telio.py
olekoliinyk: Yes, looks like to match multiple event types

nat-lab/tests/test_events_link_state.py
tomasz-grz: I'm not so familiar with the link state feature.. @LukasPukenis do you know if one second of inactivity could trigger something?

nat-lab/tests/test_events_link_state.py
tomasz-grz: Yes seems like we are assuming that the initial event is always there.

nat-lab/tests/telio.py
```diff
@@ -263,6 +276,17 @@ async def wait_for_state_peer(
             timeout,
         )
 
+    async def wait_for_new_link_state(
+        self,
+        public_key: str,
+        state: List[LinkState],
+        timeout: Optional[float] = None,
+    ) -> None:
+        """Wait until a new link_state event matching one of the `states` for `public_key` is available."""
```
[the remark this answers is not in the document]
tomasz-grz: Hmm It could also be returning the `LinkState` so we could await for any event, and then check if it's the expected one 🤔

.unreleased/LLT-5683
```diff
@@ -0,0 +1 @@
+Increase constrains in link state testcases
```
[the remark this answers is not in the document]
tomasz-grz: :D actually now we are decreasing them

nat-lab/tests/test_events_link_state.py
```diff
@@ -195,21 +221,54 @@ async def test_event_link_state_peers_exchanging_data_then_idling_then_resume(
...
+                        client_beta.wait_for_new_link_state(
+                            alpha.public_key, [LinkState.DOWN]
+                        )
+                    ),
+                ],
+                timeout=5,
+            )
+
+        # Expect the links are still UP
+        assert client_alpha.get_last_link_state_event(beta.public_key) == LinkState.UP
```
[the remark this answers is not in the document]
tomasz-grz: It was meant to replicate the functionality of the original `wait_for_link_state`.

Replaced it now with checking the order of the events

nat-lab/tests/test_events_link_state.py
```diff
@@ -145,14 +155,26 @@ async def test_event_link_state_peers_idle_all_time(
...
-        # 1 down event when Connecting, 1 up event when Connected
-        assert alpha_events == [LinkState.DOWN, LinkState.UP]
-        assert beta_events == [LinkState.DOWN, LinkState.UP]
+        # Expect no link event while peers are idle
+        with pytest.raises(asyncio.TimeoutError):
+            await wait_for_any_with_timeout(
+                [
+                    asyncio.create_task(
+                        client_alpha.wait_for_new_link_state(
+                            beta.public_key, [LinkState.DOWN]
```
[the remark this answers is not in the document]
tomasz-grz: Converted this now to check single LinkState

nat-lab/tests/test_events_link_state.py
tomasz-grz: Added helper

nat-lab/tests/telio.py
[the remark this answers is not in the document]
tomasz-grz: converted to `state: LinkState`

nat-lab/tests/telio.py
tomasz-grz: Converted to `state: LinkState`

nat-lab/tests/test_events_link_state.py
tomasz-grz: Reverted back checking the order

olekoliinyk:
+1.0

nat-lab/tests/test_events_link_state.py
```diff
@@ -16,6 +16,14 @@
 from utils.connection_util import ConnectionTag
 from utils.ping import ping
 
+RTT_SECONDS = 1
+PING_TIMEOUT = 3
+# The interval at which link-state events are emitted.
+# Determined by WG-PASSIVE_KEEPALIVE + RTT delay
+LINK_STATE_INTERVAL = 10 + RTT_SECONDS
+# The time by which we expect to emit at least one link-state event.
+EXPECTED_LINK_STATE_EVENT_TIMEOUT = 2 * LINK_STATE_INTERVAL
```
tomasz-grz, line 25: Rename to IDLE_TIMEOUT

nat-lab/tests/test_events_link_state.py
```diff
@@ -225,24 +308,38 @@ async def test_event_link_state_peer_goes_offline(
...
 
-        # 1 down event when Connecting, 1 up event when Connected
-        assert alpha_events == [LinkState.DOWN, LinkState.UP]
-        # 1 down event when Connecting, 1 up event when Connected, 1 down event when client is stopped
-        assert beta_events == [LinkState.DOWN, LinkState.UP, LinkState.DOWN]
+        # Expect the link to still be UP for the fist 10 + RTT seconds
+        with pytest.raises(asyncio.TimeoutError):
+            await client_alpha.wait_for_new_link_state(
+                beta.public_key, LinkState.DOWN, LINK_STATE_INTERVAL - PING_TIMEOUT
+            )
```
tomasz-grz, line 327: use gather here

LukasPukenis:
+1

Hasan6979 (approved):

nat-lab/tests/telio.py
```diff
@@ -53,6 +53,7 @@ class Runtime:
     _error_events: List[ErrorEvent]
     _started_tasks: List[str]
     _stopped_tasks: List[str]
+    _peer_update_event: asyncio.Event
```
mathiaspeters, line 56: Is this field really necessary?

tomasz-grz: Not strictly speaking, but I thought it would be good practice to make this event driven.
What do you have in mind?

.unreleased/LLT-5683
```diff
@@ -0,0 +1 @@
+Increase constraints in link state testcases
```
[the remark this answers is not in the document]
tomasz-grz: Good point

nat-lab/tests/telio.py
mathiaspeters: With other events we keep a list of the events and we wait until the event we're listening for appears in that list, which I would argue is also "event driven". I'm not against the kind of event driven solution you have here but I would prefer to have one solution instead of two

mathiaspeters: You shouldn't have to go back to sleeping in the test, it should be enough to do the polling solution we use in, for example, `wait_for_event_peer` to make it event-based. I do like the idea of `await`-ing new events arriving rather than doing the polling but I don't like the inconsistency. For now, maybe the best solution is to create a ticket for using this new method for other places that wait for events, and go ahead with this as-is

tomasz-grz: Good point, I made it like `wait_for_event_peer` and crated a jira ticket.

mathiaspeters (approved):
+1

PR #915, opened 2024-10-29
[LLT-5565] teliod status reporting

tomasz-grz:
### Problem
`teliod` is envisioned as rather headless daemon, which is expected to operate in places where there are no direct human interaction. But in some cases one may need to retrieve current state of meshnet, online/offline status of nodes etc.

### Solution
This PR adds a channel for returning results as `CommandResponse` back from `telio_task` that processes`TelioTaskCmd`s, and a `CommandListener` that handles commands from the client and passes it back the `CommandResponse` result.

Additionally this PR:
- Refactor `daemon_event_loop()` to exit sooner in case `telio_task()` fails with an error.
- Move error type definitions to `error.rs`
- Move daemon specific code to `daemon.rs`

- [x] README.md is updated

### DIFF
--- a/clis/teliod/README.md
+++ b/clis/teliod/README.md
@@ -28,4 +28,4 @@ There is a command for running the daemon:
    - `interface_name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
 
 And following cli commands:
- - `teliod hello-world <NAME>` - simple command for testing purposes, logs "Hello NAME!", used to ensure daemon proper startup, should be erased in the future and replaced by some more serious one
+ - `teliod get-status` - returns the status of teliod and a peer list
--- a/clis/teliod/src/command_listener.rs
+++ b/clis/teliod/src/command_listener.rs
@@ -1,0 +1,209 @@
+use crate::{
+    comms::DaemonConnection, daemon::TelioTaskCmd, ClientCmd, DaemonSocket, TelioStatusReport,
+    TeliodError,
+};
+use serde::{Deserialize, Serialize};
+use telio::telio_task::io::chan;
+use tokio::sync::oneshot;
+use tracing::{error, info};
+
+/// Command response type used to communicate between `telio runner -> daemon -> client`
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub enum CommandResponse {
+    Ok,
+    StatusReport(TelioStatusReport),
+    Err(String),
+}
+
+impl CommandResponse {
+    pub fn serialize(&self) -> String {
+        serde_json::to_string(self).unwrap_or("Err serializing".to_string())
+    }
+
+    pub fn deserialize(input: &str) -> Result<CommandResponse, TeliodError> {
+        serde_json::from_str::<CommandResponse>(input)
+            .map_err(|e| TeliodError::InvalidResponse(format!("{}", e)))
+    }
+}
+
+pub struct CommandListener {
+    socket: DaemonSocket,
+    /// Channel to send commands to telio task
+    telio_task_tx: chan::Tx<TelioTaskCmd>,
+}
+
+impl CommandListener {
+    pub fn new(socket: DaemonSocket, telio_task_tx: chan::Tx<TelioTaskCmd>) -> CommandListener {
+        CommandListener {
+            socket,
+            telio_task_tx,
+        }
+    }
+
+    async fn process_command(
+        &mut self,
+        command: &ClientCmd,
+        connection: &mut DaemonConnection,
+    ) -> Result<(), TeliodError> {
+        match command {
+            ClientCmd::GetStatus => {
+                info!("Reporting telio status");
+                let (response_tx, response_rx) = oneshot::channel();
+                #[allow(mpsc_blocking_send)]
+                self.telio_task_tx
+                    .send(TelioTaskCmd::GetStatus(response_tx))
+                    .await
+                    .map_err(|e| {
+                        error!("Error sending command: {}", e);
+                        TeliodError::CommandFailed(ClientCmd::GetStatus)
+                    })?;
+                // wait for a response from telio runner
+                if let Ok(report) = response_rx.await {
+                    // send the response to teliod client
+                    connection
+                        .respond(CommandResponse::StatusReport(report).serialize())
+                        .await?;
+                }
+            }
+        }
+        Ok(())
+    }
+
+    pub async fn handle_client_connection(&mut self) -> Result<ClientCmd, TeliodError> {
+        let mut connection = self.socket.accept().await?;
+        let command_str = connection.read_command().await?;
+
+        if let Ok(command) = serde_json::from_str::<ClientCmd>(&command_str) {
+            self.process_command(&command, &mut connection).await?;
+            Ok(command)
+        } else {
+            error!("Received invalid command from client: {}", command_str);
+            connection
+                .respond(
+                    CommandResponse::Err(format!("Invalid command: {}", command_str)).serialize(),
+                )
+                .await?;
+            Err(TeliodError::InvalidCommand(command_str))
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{ClientCmd, CommandResponse, DaemonSocket};
+    use std::{
+        path::Path,
+        time::{SystemTime, UNIX_EPOCH},
+    };
+    use tokio::{
+        io::{AsyncReadExt, AsyncWriteExt},
+        net::UnixStream,
+        sync::mpsc,
+        task,
+    };
+
+    const TEST_SOCKET_PATH: &str = "test_socket";
+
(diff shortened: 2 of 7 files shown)


clis/teliod/src/command_listener.rs
```diff
@@ -0,0 +1,153 @@
...
+                connection.respond(response.serialize()).await?;
+                self.connection = None;
+                Ok(())
+            }
+            None => Err(TeliodError::BrokenSignalStream),
+        }
+    }
+
+    pub async fn handle_client_command(&mut self) -> Result<(), TeliodError> {
+        match self.get_command().await {
```
tomasz-grz, line 71: expand `get_command()` here, so we don't have to deal with `self.connection`?

clis/teliod/src/command_listener.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub fn deserialize(input: &str) -> Result<CommandResponse, TeliodError> {
+        serde_json::from_str::<CommandResponse>(input)
+            .map_err(|e| TeliodError::InvalidResponse(format!("{}", e)))
+    }
+}
+
+pub struct CommandListener {
+    socket: DaemonSocket,
+    /// Channel to send commands to telio task
+    telio_task_tx: mpsc::Sender<TelioTaskCmd>,
```
matislovas, line 31: nit: `mpsc::Sender/Receiver` is defined in `telio-task/src/io.rs` (`Rx` and `Tx`) so maybe it is worth using them.

clis/teliod/src/command_listener.rs
```diff
@@ -0,0 +1,200 @@
...
+    async fn test_command_response_serialization() {
+        let response = CommandResponse::Ok;
+        assert_eq!(response.serialize(), "\"Ok\"");
+
+        let error_response = CommandResponse::Err("Test error".to_string());
+        assert_eq!(error_response.serialize(), "{\"Err\":\"Test error\"}");
+    }
+
+    #[tokio::test]
+    #[serial]
```
packgron, line 151: Any reason not to just create unix socket in temp dir with some random key, vs making all tests rerial?

clis/teliod/src/command_listener.rs
```diff
@@ -92,40 +69,69 @@ impl CommandListener {
...
         let (_tx, rx) = mpsc::channel(1);
 
         let socket = DaemonSocket::new(Path::new(TEST_SOCKET_PATH)).unwrap();
+
         CommandListener::new(socket, tx.clone(), rx)
     }
 
-    async fn client_send_command(cmd: &str) -> std::io::Result<()> {
+    // Simulate client sending command and waiting for response
+    async fn client_send_command(cmd: &str) -> Result<CommandResponse, std::io::Error> {
```
stalowyjez, line 123: Nit: Isn't it just `std::io::Result<CommandResponse>`?

stalowyjez (commented):
Looks good! +1

clis/teliod/src/command_listener.rs
tomasz-grz: Yeah that's probably a better approach

clis/teliod/src/command_listener.rs
tomasz-grz: It is indeed 😅

clis/teliod/src/command_listener.rs
tomasz-grz: Good catch, I'll take a look

matislovas:
I wouldn't move the `TeliodError` to a separate file, because it is only a single enum there, but if everyone else agrees on it, I will not block it :smile:

matislovas:
+0.5

stalowyjez:
> I wouldn't move the `TeliodError` to a separate file, because it is only a single enum there, but if everyone else agrees on it, I will not block it 😄

We currently have 3 error types here (of similar size), so moving one of them maybe doesn't make that much sense? Maybe moving all of them or merging them somehow would be nicer? idk

clis/teliod/src/daemon.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+        }
+    }
+    Ok(())
+}
+
+fn task_retrieve_meshmap(
+    device_identity: Arc<DeviceIdentity>,
+    auth_token: Arc<String>,
+    tx: mpsc::Sender<TelioTaskCmd>,
```
packgron, line 95: feels to me like this should be a oneshot ?

clis/teliod/src/daemon.rs
```diff
@@ -0,0 +1,242 @@
...
+    let mut cmd_listener = CommandListener::new(socket, tx.clone(), response_rx);
+
+    // TODO: This if condition and ::default call is temporary to be removed later
+    // on when we have proper integration tests with core API.
+    // This is to not look for tokens in a test environment right now as the values
+    // are dummy and program will not run as it expects real tokens.
+    let mut identity = DeviceIdentity::default();
+    if !config
+        .authentication_token
+        .eq("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
```
packgron, line 158: nit: make it a EMPTY_TOKEN const?

clis/teliod/src/daemon.rs
tomasz-grz: Good point! I didn't change this bit besides moving it, but I'll see if I can quickly convert it

clis/teliod/src/daemon.rs
tomasz-grz: Same as above, but good idea, I can quickly refactor this too

clis/teliod/src/daemon.rs
tomasz-grz: Hmmm or maybe not, if I understand correctly. Because the `telio_task` is running in a loop, the `TelioTaskCmd` `rx` part would be dropped after the first command was processed leaving the daemon unable to process further commands. Maybe @Hasan6979 can comment better as he designed this part?

What could possibly become a `oneshot` channel is the `CommandResponse` part, as every client command is expecting to have only one reply.  But in that case every task command would have to pass their own new `response_tx` part to the `telio_task`. Then the TelioTaskCmd would look something like this
```rust
TelioTaskCmd::GetStatus { response_tx: oneshot::Sender<TelioStatusReport> }
```
instead of 
```rust
TelioTaskCmd::GetStatus
```

My idea with `CommandResponse` channel was that it could be reused by all of the commands, for example if we want to return `Ok` status from `UpdateMeshmap`. What do you think?

tomasz-grz: Or is it only the Sender that gets dropped after the first message in a `oneshot`?

tomasz-grz: If I got it right

>  [oneshot channel](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html) supports sending a **single value** from a single producer to a single consumer. This channel is usually used to send the result of a computation to a waiter.

clis/teliod/src/command_listener.rs
tomasz-grz: Updated to use `chan::{Rx, Tx}`

clis/teliod/src/command_listener.rs
tomasz-grz: Fixed by adding current time `nanoseconds` to the socket name

clis/teliod/src/command_listener.rs
tomasz-grz: Fixed

clis/teliod/src/daemon.rs
tomasz-grz: Moved to `EMPTY_TOKEN`

clis/teliod/src/daemon.rs
Hasan6979: Sender is consumed at `send` and it doesn't implement a `clone` trait as well. If `try_recv` is attempted again it would result in an error since the tx part is dropped.
We would want it to not be OneShot as it would still receive update commands from MQTT.

Cargo.lock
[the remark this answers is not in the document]
tomasz-grz: Hmm interesting, I removed it but not sure why it's still in here.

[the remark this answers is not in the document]
tomasz-grz: Oh looks like it's also used by `telio-lana` and `telio-network-monitors` 🤔 so it bumped the version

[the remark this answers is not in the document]
packgron: oh then it's ok

clis/teliod/README.md
```diff
@@ -29,3 +29,4 @@ There is a command for running the daemon:
 
 And following cli commands:
  - `teliod hello-world <NAME>` - simple command for testing purposes, logs "Hello NAME!", used to ensure daemon proper startup, should be erased in the future and replaced by some more serious one
+ - `teliod get-status` - returns the status of teliod and a peer list
```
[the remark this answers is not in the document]
tomasz-grz: Never mind, removed :)

packgron (approved):
+1.0

PR #917, opened 2024-10-30
scapy has changed it's api, making mypy fail

LukasPukenis:
### Problem
scapy dependency had _any_ version in Pipfile however was v2.5 in Pipfile.lock previously. After recent updates it got updated to v2.6 which introduces api breakage. 

### Solution
This PR replaces `*` to `v2.5` in Pipflie

- [x] README.md is updated


gytsto (approved):
+2 if all pipeline checks are green :smile:

PR #918, opened 2024-10-30
Add warning to run_local.py script

gytsto:
### Problem
Running natlab locally with run_local.py and building libtelio from it, requires atleast 16GBs of RAM. Which would fail if it was run from within `default` natlab VM, which has only 4GBs of RAM.

### Solution
Adding a warning, so people won't get confused why this is happening.

- [x] README.md is updated

### DIFF
--- a/nat-lab/run_local.py
+++ b/nat-lab/run_local.py
@@ -82,13 +82,35 @@ def main() -> int:
         verify_setup_correctness()
 
     if not args.nobuild:
-        run_build_command("linux", args)
-        # Run windows tests on WinVM
-        if args.windows:
-            run_build_command("windows", args)
-        # Run nat-lab natively on macOS (TODO: Add windows support)
-        if args.o == "darwin":
-            run_build_command("darwin", args)
+        print("\u001b[33m")
+        print("|=======================================================|")
+        print("| WARNING! Running builds requires atleast 16GBs of RAM |")
+        print("|=======================================================|")
+        print("\u001b[0m")
+        try:
+            run_build_command("linux", args)
+            # Run windows tests on WinVM
+            if args.windows:
+                run_build_command("windows", args)
+            # Run nat-lab natively on macOS (TODO: Add windows support)
+            if args.o == "darwin":
+                run_build_command("darwin", args)
+        except subprocess.CalledProcessError:
+            print("\u001b[31m")
+            print(
+                "|===================================================================|"
+            )
+            print(
+                "| ERROR! If build failed by getting SIGKILL, it might               |"
+            )
+            print(
+                "| ERROR! be due to lack of RAM. Build requires atleast 16GBs of RAM |"
+            )
+            print(
+                "|===================================================================|"
+            )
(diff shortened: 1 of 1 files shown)


nat-lab/run_local.py
```diff
@@ -82,13 +82,35 @@ def main() -> int:
...
-        # Run nat-lab natively on macOS (TODO: Add windows support)
-        if args.o == "darwin":
-            run_build_command("darwin", args)
+        print("\u001b[33m")
+        print("|=======================================================|")
+        print("| WARNING! Running builds requires atleast 16GBs of RAM |")
+        print("|=======================================================|")
+        print("\u001b[0m")
+        try:
+            run_build_command("linux", args)
```
tomasz-grz, line 91: Does this mean that it's gonna be always run twice, do we want that?

nat-lab/run_local.py
```diff
@@ -82,13 +82,35 @@ def main() -> int:
...
+            # Run windows tests on WinVM
+            if args.windows:
+                run_build_command("windows", args)
+            # Run nat-lab natively on macOS (TODO: Add windows support)
```
tomasz-grz, line 95: I guess windows is already added, so we can remove the TODO 😄

tomasz-grz (commented):
Nice, one though tough, is it worth to investigate why building requires 16gb in the first place and if there is a way to reduce it? Or alternatively increase the `default` VM memory so it doesn't fail?

nat-lab/run_local.py
gytsto: So script works the same as it used to. It builds linux, if needed builds windows and darwin separately, since they all are built in different environments. Meaning that it will be ran twice. In CI it won't build anything, bcuz `--nobuild` flag is passed and binaries a fetched from artifacts.

nat-lab/run_local.py
gytsto: I don't really know what this TODO is about. Don't really know if this works natively on windows tho :smile: Therefore leaving it as it is

gytsto:
> Nice, one thought tough, is it worth to investigate why building requires 16gb in the first place and if there is a way to reduce it? Or alternatively increase the `default` VM memory so it doesn't fail?

First of all this doesn't affect CI. Secondly, I think this high amount of RAM needed is because of parallel linking or some other thing called LTO which I didn't really get into. And the last, most of libtelio devs are building binaries outside `default` VM, it's just an edge case, where someone rarely wants to build it inside VM, so increasing VM max RAM to 16GBs is not really needed. Natlab itself doesn't require that much memory to run. That's why I've added a warning instead of increasing memory itself.

jjanowsk (commented):
+1 although what's wrong with the regular prints :) ?

jjanowsk (approved):

gytsto:
> +1 although what's wrong with the regular prints :) ?

Just wanted them to standout, I can remove the coloring of you don't like it :smile:

tomasz-grz (approved):
+1

PR #921, opened 2024-10-30
Fix upnp expiration testing

matislovas:
### Problem
Lease duration expiration was tested extensively.

### Solution
Shouldn't do that, added 1 additional seconds of waiting, and 1 attempt of trying.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_upnp_connection.py
+++ b/nat-lab/tests/test_upnp_connection.py
@@ -234,6 +234,7 @@ async def test_upnp_port_lease_duration(
         # once the endpoint interval expires, the endpoint is validated and if
         # the lease duration left is less than 25% the port mappings are renewed.
         lease_duration_s = 40
+        jitter_duration_s = 5
 
         assert alpha_setup_params.features.direct
         assert alpha_setup_params.features.direct.upnp_features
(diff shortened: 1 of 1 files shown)


nat-lab/tests/test_upnp_connection.py
```diff
@@ -264,7 +264,7 @@ async def test_upnp_port_lease_duration(
             await temp_exit_stack.enter_async_context(alpha_gw_router.reset_upnpd())
 
         # this check should be done before endpoint interval triggers a new mapping (optimal > 5s)
-        upnpc_cmd = await execute_upnpc_with_retry(alpha_conn)
+        upnpc_cmd = await alpha_conn.create_process(["upnpc", "-i", "-l"]).execute()
```
gytsto, line 267: why reverting back from `execute_upnpc_with_retry`. Won't we just hit previous issue we had again?

gytsto: Or maybe here we should get specific exit status?

matislovas: This should check if the lease does _does not_ exists (hence `assert re.search("^ [0-9]+ UDP", ....) is None`). So we shouldn't check it for 10 seconds. The `execute_upnpc_with_retry` to repeat, if we actually want to find the lease.

gytsto: ok, but won't this call throw some sort of `ProcessExecError` and then fail the test if `upnpd` is not started yet?

matislovas: In case of not working `upnpd` - yes it _should_ fail, but than the test should fail, because we're expecting to be running, only those leases should be expired.

lcruz99:
Since we're using retries and the endpoint interval is set to 5s, I suspect `lease_duration_s` doesn't need to be 40s long. Can you try to reduce it something like 10s?

PR #922, opened 2024-10-30
Added test to to verify correctness of periodic qos collection

tomaszklak:
- [x] README.md is updated


mathiaspeters:
I'm not entirely sure what the new test is actually testing, so it would be nice with a comment or maybe a more descriptive name

mathiaspeters:
The name is definitely better. A last question, with the increase RTT interval we don't typically collect that data during our tests and this test sets it to something shorter and checks that it's collected as expected?

mathiaspeters:
+1

matszczygiel:
+1

PR #924, opened 2024-10-31
Tag each ping in nat lab with random payload

tomaszklak:
### Problem
It's not always easy to map from the ping packet in the pcap to the test log.

### Solution
Tag each ping packet sent from tests with a random bytes to make it easy to map back from pcap to test log. In case of the docker connections, `KILL_ID` is used as the random string.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/ping.py
+++ b/nat-lab/tests/utils/ping.py
@@ -34,17 +34,20 @@ class Ping:
         self._ip = ip
         self._connection = connection
         self._ip_proto = testing.unpack_optional(get_ip_address_type(ip))
+        kill_id = secrets.token_hex(8).upper()
 
         if connection.target_os == TargetOS.Windows:
             self._process = connection.create_process(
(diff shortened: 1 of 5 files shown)


LukasPukenis (commented):
+1

nat-lab/tests/utils/ping.py
```diff
@@ -33,18 +34,29 @@ def __init__(self, connection: Connection, ip: str) -> None:
         self._ip = ip
         self._connection = connection
         self._ip_proto = testing.unpack_optional(get_ip_address_type(ip))
+        kill_id = secrets.token_hex(8).upper()
 
         if connection.target_os == TargetOS.Windows:
             self._process = connection.create_process(
                 ["ping", ("-4" if self._ip_proto == IPProto.IPv4 else "-6"), "-t", ip]
```
gytsto, line 41: what about windows?

tomaszklak: Afaics, windows has no support for such thing: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/ping :(

PR #927, opened 2024-11-04
Fix direct connection wait

packgron:
### Problem
Direct connection test fails, as some nodes can report direct events, before we fully downgrade to relayed connections

### Solution
Because python futures are eagerly executed, we should not collect
events for two different steps at the same time.

Moved relay and direct event gathering after a change in network
configurations, and only after each step is confirmed

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_direct_connection.py
+++ b/nat-lab/tests/test_direct_connection.py
@@ -8,6 +8,7 @@ import timeouts
 from collections import defaultdict
 from config import DERP_SERVERS
 from contextlib import AsyncExitStack, asynccontextmanager
+from datetime import datetime
 from helpers import (
     Environment,
     ping_between_all_nodes,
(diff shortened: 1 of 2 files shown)


tomaszklak (dismissed):
+1

mathiaspeters (dismissed):
+1

nat-lab/tests/test_direct_connection.py
```diff
@@ -189,27 +191,30 @@ async def test_direct_working_paths(
...
+                exit_stack.enter_async_context(
+                    run_async_context(
+                        client.wait_for_event_peer(
+                            node.public_key, [NodeState.CONNECTED], [PathType.DIRECT]
+                        )
+                    )
+                )
+                for client, node in itertools.product(env.clients, env.nodes)
+                if not client.is_node(node)
+            ])
```
mathiaspeters, line 210: Isn't this in the wrong scope now? It looks from the indentation that it's part of the `_disable_direct_connection` blocks whereas the `await gather` call for it is outside of that scope. At least that's what it looks like

packgron: Yeh this is the anoying part, its future in futures:
```
async def show():
    async def fake_wait():
        print("done")
        return "done"
        
        x = exit_stack.enter_async_context(
            run_async_context(
                fake_wait()
            )
        )
        print("x", x)
        
        y = await x
        print("y", y)

        z = await y
        print("z", z)

asyncio.run(show())
```

outputs:
```
x <coroutine object AsyncExitStack.enter_async_context at 0x77509351c040>
y <Task pending name='Task-2' coro=<show.<locals>.fake_wait() running at /home/etg/etg/src/telio-build/libtelio/nat-lab/./tests/test_direct_connection.py:702>>
done
z done
```

packgron: basicaly enter_async_context is also a future, so needs to be awaited, then returns a wrapped coroutine

tomaszklak (approved):

PR #928, opened 2024-11-05
Added moose logs collection in nat-lab

tomaszklak:
- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -32,6 +32,7 @@ from utils.bindings import (
 from utils.command_grepper import CommandGrepper
 from utils.connection import Connection, DockerConnection, TargetOS
 from utils.connection_util import get_uniffi_path
+from utils.moose import MOOSE_LOGS_DIR
 from utils.output_notifier import OutputNotifier
 from utils.process import Process, ProcessExecError
 from utils.python import get_python_binary
@@ -1040,6 +1041,17 @@ class Client:
                 f.write("\n\n\n\n--- SYSTEM LOG ---\n\n")
                 f.write(system_log_content)
 
+        moose_traces = await find_files(
+            self._connection, MOOSE_LOGS_DIR, "moose_trace.log*"
+        )
+        for trace_path in moose_traces:
+            copy_file(self._connection, trace_path, log_dir)
+            file_name = os.path.basename(trace_path)
+            os.rename(
+                os.path.join(log_dir, file_name),
+                os.path.join(log_dir, f"{container_id}-{file_name}"),
+            )
+
     async def save_moose_db(self) -> None:
         """
         Check if any the moose db files exists ("*-events.db"),
@@ -1129,11 +1141,9 @@ class Client:
 
         coredump_folder, file_prefix = self.get_coredump_folder()
 
-        # find all core dump files
-        process = await self._connection.create_process(
-            ["find", coredump_folder, "-maxdepth", "1", "-name", f"{file_prefix}*"]
-        ).execute()
-        dump_files = process.get_stdout().strip().split()
+        dump_files = await find_files(
+            self._connection, coredump_folder, f"{file_prefix}*"
+        )
 
         coredump_dir = "coredumps"
         os.makedirs(coredump_dir, exist_ok=True)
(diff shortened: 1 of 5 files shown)


packgron:
I only see singular moose log file then multiple clients are used, and it's unclear from witch client those logs are comming

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -13,6 +14,7 @@
 
 REMOTE_LOG = "remote.log"
 TCLI_LOG = "tcli.log"
+MOOSE_LOG_DIR = "moose_logs"
```
LukasPukenis, line 17: Can this constant be used instead of other occurrences of `moose_logs`?

nat-lab/tests/test_lana.py
```diff
@@ -2041,6 +2041,17 @@ async def test_lana_same_meshnet_id_is_reported_after_a_restart(
                 CONTAINER_EVENT_BACKUP_PATH,
                 events_path,
             )
+            moose_traces = await find_files(
```
LukasPukenis, line 2044: Nitpicky but apparently we just expect one file thus `find_file` that would guarantee one result at max would be more correct, iterating through a loop later on gives impression that it's a list of files, but this runs per-connection(beta in this case), thus only will resolve one file.

As an easy alternative below could just assert that the length is one and just use the 0th item

tomaszklak: No, we don't expect one file. Moose is doing daily rolling logs, so for a test that happens to be running over midnight it will result in two logs, eg: `moose_trace.log.2024-11-06` and `moose_trace.log.2024-11-07.`

nat-lab/tests/test_lana.py
```diff
@@ -2041,6 +2041,17 @@ async def test_lana_same_meshnet_id_is_reported_after_a_restart(
...
+                connection_beta, "/moose_logs", "moose_trace.log*"
```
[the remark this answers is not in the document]
tomaszklak: Yes, moose adds the date to the end of it's logs.

nat-lab/tests/telio.py
```diff
@@ -1156,6 +1160,37 @@ async def collect_core_dumps(self):
...
+async def find_files(connection, where, name_pattern):
+    """Wrapper for 'find' command over the connection"""
+
+    try:
+        process = await connection.create_process(
+            ["find", where, "-maxdepth", "1", "-name", name_pattern]
+        ).execute()
+        return process.get_stdout().strip().split()
+    except ProcessExecError:
+        # Expected when 'where' doesn't exist
```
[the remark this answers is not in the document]
tomaszklak: If you call this function on directory which doesn't exist it will return empty list instead of raising an exception. This happens for most of our tests and the moose directory (most of our tests have moose disabled).

nat-lab/tests/uniffi/libtelio_remote.py
tomaszklak: I'm having a hard time forcing python to import `MOOSE_LOGS_DIR` from `moose.py` in `libtelio_remote.py` so I'm afraid this is the only place where this string has to be duplicated.

tomaszklak:
> I only see singular moose log file then multiple clients are used, and it's unclear from witch client those logs are comming

Yeah, I was so focused on the 'same meshnet id' test that I forgot to add separate prefixes for the various containers. It's now fixed, eg:
```
logs
├── derp_01_relay.log
├── derp_02_relay.log
├── derp_03_relay.log
├── dns_server_1.log
├── dns_server_2.log
└── test_lana_with_same_meshnet_IPv4_IPv4
    ├── alpha-events.db
    ├── beta-events.db
    ├── cone-client-01-moose_trace.log.2024-11-06
    ├── cone-client-01.log
    ├── cone-client-02-moose_trace.log.2024-11-06
    ├── cone-client-02.log
    ├── gamma-events.db
    ├── nat-lab-cone-client-01-1.pcap
    ├── nat-lab-cone-client-02-1.pcap
    ├── nat-lab-dns-server-1-1.pcap
    ├── nat-lab-dns-server-2-1.pcap
    ├── nat-lab-symmetric-client-01-1.pcap
    ├── symmetric-client-01-moose_trace.log.2024-11-06
    └── symmetric-client-01.log
```

packgron (dismissed):
+1.0

nat-lab/tests/uniffi/libtelio_remote.py
mathiaspeters: This should be possible to solve but the current solution might be cleaner. I'm guessing your issue with importing is happening on windows and mac but kinda works on linux? In that case, in `(mac|windows)_vm_util.py` if you copy `moose.py` to a folder "next to" the `uniffi` folder in the VM you should be able to import it with a relative file import. As in, a partial file tree view on the VMs should look like
```
..
  - uniffi
    - libtelio_remote.py
  - utils
    - moose.py 
```

Though I should say that I haven't tried this so maybe it doesn't work :shrug:

mathiaspeters:
+1

nat-lab/tests/uniffi/libtelio_remote.py
tomaszklak: To be honest, I'm not convinced that complexity of copying that file just to avoid copying string constant is justifiable.

matszczygiel (approved):
+1

nat-lab/tests/uniffi/libtelio_remote.py
Jauler: Agree

Jauler (approved):
+1.0

PR #932, opened 2024-11-07
Fix rustls crash on TLS connection

matszczygiel:
### Problem
rustls crashes when the default crypto provider is not defined

### Solution
Use one of the rustls crate features (`ring`) to define the default crypto provider.
According to https://docs.rs/rustls/latest/rustls/index.html#crate-features defining `ring` feature and NOT defining `custom-provider` is enough to implicitly use the ring provider without the need to install it manually.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5770
+++ b/.unreleased/LLT-5770
@@ -1,0 +1,1 @@
+Fix rustls crash on derp TLS connection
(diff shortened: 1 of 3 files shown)


Jauler (approved):
+2.0

PR #937, opened 2024-11-11
Allow log flushing to fail in nat-lab

ThrasherLT:
### Problem
log flushing can someitmes fail and cause issues for nat-lab tests

### Solution
Allowed log flushing to fail


mathiaspeters:
I'm wondering, does it make sense to always allowing flushing to fail or just in this specific case?

mathiaspeters:
I was more thinking, is it just during test cleanup that we want to allow it to fail, or does it make sense to add it to `libtelio_remote.py` instead so an error while flushing never causes an issue

nat-lab/tests/telio.py
```diff
@@ -472,7 +472,15 @@ async def on_stderr(stderr: str) -> None:
...
+                    # flush_logs() is allowed to fail here:
+                    try:
+                        await self.get_proxy().flush_logs()
+                    except:
+                        print(
+                            datetime.now(),
+                            "Test cleanup: log flush failed",
+                        )
+
                     await self.get_proxy().shutdown(self._connection.target_name())
```
gytsto, line 484: I also think that shutdown should be inside `try` block, because if we get error on `flush_logs`, it is likely that we will get error on `shutdown`. Also would be nice if we could print what exception we got for more info.

gytsto:
> I was more thinking, is it just during test cleanup that we want to allow it to fail, or does it make sense to add it to `libtelio_remote.py` instead so an error while flushing never causes an issue

it was meant more for specific case, like test cleanup. So telio logs would be saved in any case

nat-lab/tests/telio.py
gytsto: nvm, don't add shutdown, it is already handled :smile:, but print exception as mentioned before

Jauler (dismissed):
+2.0

mathiaspeters:
+1

mathiaspeters (approved):

PR #939, opened 2024-11-11
Move Batcher under SessionKeeper

LukasPukenis:
View commit details for more info

- [x] README.md is updated


crates/telio-traversal/src/batcher.rs
```diff
@@ -172,14 +204,14 @@ mod tests {
...
-        for ac in batcher.get_actions().await {
+        for ac in batcher.get_actions().await.unwrap() {
             ac.1(&mut test_checker).await.unwrap();
         }
 
         assert!(test_checker.values.len() == 3);
         assert!(test_checker.values[2].1.duration_since(start_time) == Duration::from_secs(650));
     }
-
+    //
```
tomasz-grz, line 214: a suspenseful comment 😅

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -62,7 +62,12 @@ pub struct SessionKeeper {
 }
 
 impl SessionKeeper {
-    pub fn start(sock_pool: Arc<SocketPool>) -> Result<Self> {
+    pub fn start(
+        sock_pool: Arc<SocketPool>,
+
+        // This is a way to pass mocked batcher instance when testing
+        #[cfg(test)] batcher: Box<dyn BatcherTrait<PublicKey, State>>,
```
tomasz-grz, line 69: Is this a parameter that is only present during testing? And I thought you couldn't "overload" functions in rust 😀  part genius part insanity 🤯

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -307,7 +319,10 @@ impl Runtime for State {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::batcher::MockBatcherTrait;
```
tomasz-grz, line 322: Nice!

tomasz-grz (commented):
Quick first look seems good +0.5, more points after a closer look with fresh 👀

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: You guessed right - `#[cfg(test)]` enables the next expression only when testing `cargo test` :).

The reason I used it boils down to some things:
- I wanted to move batcher to SessionKeeper since it's just a container, same as RepeatedTask(used by SessionKeeper).
- `wg_controller::consolidate()` calls `SessionKeeper` but based on the threshold parameter it might need to call batcher instead of `RepeatedTasks`.
- I want to somehow test that actually batcher was called and the only way to do that is to provide a trait since we use `mockall` that allows mocked traits. This way I could test that the API call went to batcher :)

To be honest there's no harm in leaving out the `cfg(test)` and simply having the parameter always present but I've already seen such pattern several times so I am thinking it might advertise the intent slightly better :)

tomasz-grz (commented):
Still looks good so here's another +0.5 :)

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -307,7 +319,10 @@ impl Runtime for State {
...
+    use mockall::Sequence;
```
lcruz99, line 323: unused import

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -391,4 +406,60 @@ mod tests {
...
+            .once()
+            .with(eq(pk))
+            .return_once(|_| ());
+
+        // it's hard to mock the exact return since it involves a complex type, however we
+        // can at least verify that the batcher's actions were queried
+        batcher
+            .expect_get_actions()
+            .times(..)
+            .returning(|| Err(BatcherError::NoActions));
```
lcruz99, line 446: use of undeclared type `BatcherError`

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: You're totally right to be honest, I was just flexing my freshly gained Rust skills :))

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: Thanks for spotting, got left out during refactoring :) Fixed

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: Removed it :)

crates/telio-traversal/src/batcher.rs
```diff
@@ -32,57 +56,59 @@ where
...
+                    action.0.deadline = now + action.0.interval;
+                    batched_actions.push((key.clone(), action.1.clone()));
                 }
+            }
 
-                return batched_actions;
-            } else {
-                let _ = self.notify_add.notified().await;
+            if batched_actions.is_empty() {
+                telio_log_error!("Batched jobs incorrectly. At least one job should be emitted.");
```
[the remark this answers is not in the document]
LukasPukenis: Thanks, that is a really nice suggestion and I've added it :)

crates/telio-traversal/src/batcher.rs
LukasPukenis: Removed it :)

lcruz99 (approved):
+1

PR #945, opened 2024-11-13
LLT-5618 wait for kills

tomaszklak:
### Problem
Sometimes upnp module managed to run periodic code that recreates mappings in between the `upnpd` restart and `upnpc` call that verifies no mappings are present.

### Solution
All calls to `kill` or `killall` now wait for the process to terminate. After that change there is no need to for the above assertion - we know that we have restarted the `upnpd`.

- [x] README.md is updated


mathiaspeters:
+1

packgron (approved):
+1.0

PR #946, opened 2024-11-14
bump libtelio build version and save macos system logs

gytsto:
### Problem
- no macos logs
- poor macos vm provision

### Solution
- bump libtelio-build version with new macos vm provision
- move dmesg host logs collection to before and after pytest
- add macos system and dmesg logs

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -232,9 +232,54 @@ async def _copy_vm_binaries_if_needed(items):
                 return
 
 
+def save_dmesg_from_host():
+    try:
+        result = subprocess.run(
+            ["sudo", "dmesg", "-d", "-T"],
+            capture_output=True,
+            text=True,
+            check=True,
+        ).stdout
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing dmesg: {e}")
+        return
+
+    if result:
+        with open(
+            os.path.join("logs", "dmesg.txt"), "w", encoding="utf-8"
+        ) as f:
+            f.write(result)
+
+
+def save_audit_log():
+    try:
+        source_path = "/var/log/audit/audit.log"
+        if os.path.exists(source_path):
+            shutil.copy2(source_path, "logs/audit.log")
+        else:
+            print(f"The audit file {source_path} does not exist.")
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        print(f"An error occurred when processing audit log: {e}")
+
+async def collect_kernel_logs(items):
+    if os.environ.get("NATLAB_SAVE_LOGS") is None:
+        return
+
+    is_ci = os.environ.get("CUSTOM_ENV_GITLAB_CI") is not None and os.environ.get(
+        "CUSTOM_ENV_GITLAB_CI"
+    )
+
+    log_dir = "logs"
+    os.makedirs(log_dir, exist_ok=True)
+
+    save_dmesg_from_host()
+    save_audit_log()
+
+
 def pytest_runtestloop(session):
     if not session.config.option.collectonly:
         asyncio.run(_copy_vm_binaries_if_needed(session.items))
+        asyncio.run(collect_kernel_logs(session.items))
 
         if not asyncio.run(perform_setup_checks()):
             pytest.exit("Setup checks failed, exiting ...")
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -1,6 +1,7 @@
 import asyncio
 import os
 import pytest
+import shutil
 import subprocess
 from helpers import SetupParameters
 from interderp_cli import InterDerpClient
@@ -232,7 +233,7 @@ async def _copy_vm_binaries_if_needed(items):
                 return
 
 
-def save_dmesg_from_host():
+def save_dmesg_from_host(prefix):
     try:
         result = subprocess.run(
             ["sudo", "dmesg", "-d", "-T"],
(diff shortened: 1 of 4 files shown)


LukasPukenis (commented):
Is it possible to split the PR into two:
- necessary refactoring with no functional changes
- actual MacOS log collection
?

It's pretty hard to see the actual, functional changes

gytsto:
> Is it possible to split the PR into two:
> 
>     * necessary refactoring with no functional changes
> 
>     * actual MacOS log collection
>       ?
> 
> 
> It's pretty hard to see the actual, functional changes

there are no refactoring changes, besides moving `save_dmesg_from_host` and `save_audit_logs` to conftest.py, but I guess I can split them in 2 commits if it's ok for you

gytsto:
@LukasPukenis it should be more clear for you know I hope, split to 3 commits (bump, move, add) :smile:

gytsto:
Note: while this was in review, new version of libtelio-build was released, related to original changes, therefore second version bump :)

nat-lab/tests/conftest.py
```diff
@@ -232,7 +233,7 @@ async def _copy_vm_binaries_if_needed(items):
                 return
 
 
-def save_dmesg_from_host():
+def save_dmesg_from_host(prefix):
```
mathiaspeters, line 236: Nit: Wouldn't this technically be a suffix, since it's `dmesg-{prefix}` rather than `{prefix}-dmesg`?

gytsto: you are right... mixed up the words :laughing: will fix it in a second

gytsto: fixed

mathiaspeters:
+1

nat-lab/tests/conftest.py
```diff
@@ -232,9 +233,75 @@ async def _copy_vm_binaries_if_needed(items):
...
+        with open(
+            os.path.join("logs", f"dmesg-macos-{suffix}.txt"), "w", encoding="utf-8"
+        ) as f:
+            f.write(dmesg_proc.get_stdout())
+    except ProcessExecError as e:
+        print(f"Failed to collect dmesg logs {e}")
+
+
+async def collect_kernel_logs(items, suffix):
+    if os.environ.get("NATLAB_SAVE_LOGS") is None:
```
LukasPukenis, line 278: Minor: Would it be possible to move environment access to call sites instead and simplify functions? That would be a very easy change but the functions would become less dependent on the global state. Though it doesn't bring much benefit but I think making these util functions as self-contained as possible is a nice thing. `is_ci` would also become a function parameter. 

Actually this would also enable unit testing in natlab code itself easier

nat-lab/tests/conftest.py
```diff
@@ -232,9 +233,75 @@ async def _copy_vm_binaries_if_needed(items):
...
+    save_audit_log(suffix)
+
+    for item in items:
+        if any(mark.name == "mac" for mark in item.own_markers):
+            try:
+                async with mac_vm_util.new_connection() as conn:
+                    await _save_macos_logs(conn, suffix)
+            except OSError as e:
+                if is_ci:
+                    raise e
```
LukasPukenis, line 298: I wonder why do we only care about the exceptions raised wen we're in CI? It is useful information locally as well imho

nat-lab/tests/conftest.py
```diff
@@ -232,9 +233,75 @@ async def _copy_vm_binaries_if_needed(items):
...
+            print(f"The audit file {source_path} does not exist.")
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        print(f"An error occurred when processing audit log: {e}")
+
+
+async def _save_macos_logs(conn, suffix):
+    try:
+        dmesg_proc = await conn.create_process(["dmesg"]).execute()
+        with open(
+            os.path.join("logs", f"dmesg-macos-{suffix}.txt"), "w", encoding="utf-8"
```
LukasPukenis, line 270: question: these logs are stored on the mac machine itself, but how do the same filename not clash with other log files from other mac machines for example when collecting those? If you could explain that part a bit that would be :heart:

nat-lab/tests/conftest.py
```diff
@@ -232,9 +233,75 @@ async def _copy_vm_binaries_if_needed(items):
...
+
+    is_ci = os.environ.get("CUSTOM_ENV_GITLAB_CI") is not None and os.environ.get(
+        "CUSTOM_ENV_GITLAB_CI"
+    )
+
+    log_dir = "logs"
+    os.makedirs(log_dir, exist_ok=True)
+
+    save_dmesg_from_host(suffix)
+    save_audit_log(suffix)
```
LukasPukenis, line 289: I think the naming is a bit inconsistent, both are saving logs from the machine, so maybe just `save_dmesg_log` and `save_audit_log` would be enough? :thinking:

nat-lab/tests/conftest.py
gytsto: while running locally, we except to not have all VMs running, therefore we expect to get `OSError`. But in CI we expect all VMs to be running and if `OSError` happens, that means, something is wrong, therefore we raise an exception.

nat-lab/tests/conftest.py
gytsto: There is no other mac machine, only one macos test vm. `dmesg` command is executed from `linux vm` to `macos vm` via ssh, then stdout output is flushed to file in `linux vm`. During all pytest session it will create only 1 before and after file.

nat-lab/tests/conftest.py
gytsto: I'll change `save_audit_log_from_host`, adding `from_host` suffix makes it more clear where the log is comming from

nat-lab/tests/conftest.py
gytsto: moved

nat-lab/tests/conftest.py
LukasPukenis: Not sure if I got it right but in case there will appear a second mac machine, does it mean this code needs fixes then?

gytsto: maybe yes, but there is no plans for 2nd mac vm in any near or far future :smile:

tomaszklak (approved):

PR #948, opened 2024-11-14
LLL-5772 handle upnp lease expiry when battery optimizations are on

tomaszklak:
### Problem
UPNP lease is not renewed when battery optimizations are turned on and UPNP endpoint provider is paused.

### Solution
Ignore battery optimizations when it't time (at least 75% of the whole lease duration) to renew the lease.

- [x] README.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/upnp.rs
+++ b/crates/telio-traversal/src/endpoint_providers/upnp.rs
@@ -21,11 +21,13 @@ use telio_crypto::PublicKey;
 use telio_proto::{Session, WGPort};
 use telio_sockets::External;
 use telio_task::{io::chan::Tx, task_exec, BoxAction, Runtime, Task};
+use telio_utils::interval;
 use telio_utils::{
     exponential_backoff::{Backoff, ExponentialBackoff, ExponentialBackoffBounds},
     telio_log_debug, telio_log_info, telio_log_warn, PinnedSleep,
 };
 use telio_wg::{DynamicWg, WireGuard};
+use tokio::time::Instant;
 use tokio::{net::UdpSocket, pin, sync::Mutex};
 
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -653,7 +653,7 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> Runtime for State<Wg, I, E> {
     {
         pin!(updated);
 
-        if self.is_endpoint_provider_paused {
+        if !self.igd_gw.lease_needs_renew() && self.is_endpoint_provider_paused {
```
Jauler, line 656: I think we may have a bit of an issue with pausing and UPnP module. (And we had it for as long as UPnP and pausing module existed, but I guess it would be nice to fix it.

As pausing is done on the assumption, that endpoint will stay "active" without any actions, which is true for local endpoints and STUN endpoints (as long as traffic is flowing), but it is *not* true for UPnP, as the gateway have every right to cut off the traffic flow once lease expires. 

In our case, we will check whether there are no pending renews required and we will check if module is paused -> and will go to indefinite sleep. During which the lease may expire and gateway could cutoff communication. 

So I wonder if pausing even makes sense for UPnP at all :thinking:.

tomaszklak: Yeah, I was also worried about that but this is not exactly how it will behave. Every 5 seconds we will run periodic wg consolidation. This will pause the endpoint providers even if they are already pause which other branch of the `select!` and returns. Which means that we will notice the need for renew with 5s granularity.

Jauler: Hmm, this requirement to call "pause()" every once in a while seems like really non-obvious requirement. I know that this is not directly related to your task, but are there any ways to make this requirement more explicit?

tomaszklak: We could ignore the question of how to enforce pausing, by replacing this [line](https://github.com/NordSecurity/libtelio/blob/b92ad778b7a7eb3092dfc63bdc255725081cbe25/crates/telio-traversal/src/endpoint_providers/upnp.rs#L659) with a await for sleep(5s)

Jauler: yeah, that could work.

On the other hand, `lease_needs_renew()` does not when exactly renew will be needed, so maybe it is worth `::sleep_until()`'ing instead of polling?

tomaszklak: Good point. I've updated the code so that we sleep until the time when renew is needed.

Jauler (approved):
+1.0

PR #949, opened 2024-11-15
Prepare VPN containers only whenever needed

Jauler:
### Problem
VPN containres are being overprepared for every container/every node on every test. This is excessive, 

### Solution
This PR introduces an indicator whether VPN containers are needed in `setup_environment()` and skips VPN container preparation when not needed. What is more, it only prepares containers in setup_environment once instead of doing it for every node separately.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/helpers.py
+++ b/nat-lab/tests/helpers.py
@@ -217,6 +217,7 @@ async def setup_environment(
     exit_stack: AsyncExitStack,
     instances: List[SetupParameters],
     provided_api: Optional[API] = None,
+    prepare_vpn: bool = False,
 ) -> AsyncIterator[Environment]:
     """Sets up the basic environment based on the given parameters.
 
@@ -224,6 +225,7 @@ async def setup_environment(
     * `setup_api` (creates mocked core API (if not provided)),
     * `setup_connections` (sets up the connections to the docker containers) and
     * `setup_clients` (creates clients using the given configuration)
+    * `prepare_vpn` - Indicates whether VPN containers are needed for the test.
     which results in the fully prepared test configuration.
     It also checks whether the number of connections is correct with the conntrackers.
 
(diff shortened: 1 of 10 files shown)


nat-lab/tests/helpers.py
```diff
@@ -217,13 +217,15 @@ async def setup_environment(
...
     provided_api: Optional[API] = None,
+    prepare_vpn: bool = False,
 ) -> AsyncIterator[Environment]:
     """Sets up the basic environment based on the given parameters.
 
     It basically combines functionalities of the few functions:
     * `setup_api` (creates mocked core API (if not provided)),
     * `setup_connections` (sets up the connections to the docker containers) and
     * `setup_clients` (creates clients using the given configuration)
+    * conditionally configure VPN containers
```
gytsto, line 228:
```suggestion
    * `prepare_vpn` - Indicates whether VPN containers are needed for the test.
```

nat-lab/tests/helpers.py
```diff
@@ -310,6 +313,9 @@ async def setup_environment(
         ),
     )
 
+    if prepare_vpn:
```
gytsto, line 316: Maybe move this before `setup_clients`, just in case somebody will add wg_server as peer

gytsto (commented):
other then few comments, great PR

nat-lab/tests/helpers.py
Jauler: This makes it slightly more complicated, as current VPN configuration is refering to the list of the nodes in API to configure required public keys to be accepted in VPN containers.

But `.setup_client`s establishes this list.

nat-lab/tests/helpers.py
Jauler: fixed

nat-lab/tests/helpers.py
gytsto: `setup_api` does that, `setup_clients` already uses created meshmap, so it would be appropriate to run `prepare_vpn` before `setup_clients`. Unless I am not getting the idea what you are saying :smile:

Jauler: oh yeah, you are right. Fixed.

gytsto (approved):
+2

PR #953, opened 2024-11-15
[LLT-5751] Configure teliod tunnel interface

tomasz-grz:
### Problem

### Solution


clis/teliod/src/command_listener.rs
```diff
@@ -59,22 +55,91 @@ impl CommandListener {
...
+                if let Ok(response) = response_rx.await {
+                    match response {
+                        Ok(()) => {
+                            return Ok(CommandResponse::Ok);
+                        }
+                        Err(e) => {
+                            return Ok(CommandResponse::Err(e.to_string()));
+                        }
+                    }
+                }
```
lcruz99, line 82: We can replace this daemon response blocks for a function/closure. (including `ClientCmd::GetStatus` case)

clis/teliod/src/command_listener.rs
```diff
@@ -59,22 +55,91 @@ impl CommandListener {
...
+                        }
+                        Err(e) => {
+                            return Ok(CommandResponse::Err(e.to_string()));
+                        }
+                    }
                 }
             }
         }
-        Ok(())
+        Ok(CommandResponse::Err(
```
lcruz99, line 131: is this reachable?

lcruz99:
I wonder if the `set-ip` command is useful. What if we do set the interfaces whenever we enable/disable the meshnet?

clis/teliod/src/daemon.rs
```diff
@@ -47,32 +65,112 @@ fn telio_task(
...
+    // Keep track of current meshnet states
+    let mut state = TelioTaskStates::default();
+    // Mock the interface configuration for tests
+    let sys_config: Box<dyn SystemConfigurator> = if auth_token.as_str().eq(EMPTY_TOKEN) {
+        // And some fake IP address
+        state.interface_ip_address = Some(IpAddr::V4(Ipv4Addr::new(100, 100, 100, 100)));
+        Box::new(MockSystemConfigurator::new(false))
+    } else {
+        Box::new(IfconfigSystemConfigurator)
+    };
```
lcruz99, line 77: auth token is defining if we're running tests?

wouldn't conditional compilation fit?

clis/teliod/src/daemon.rs
```diff
@@ -47,32 +65,112 @@ fn telio_task(
...
+                        meshnet_enabled: state.meshnet_is_enabled,
+                        meshnet_ip: state.interface_ip_address,
+                        external_nodes: telio.external_nodes()?,
                     };
                     debug!("Telio status: {:#?}", status_report);
                     if response_tx_channel.send(status_report).is_err() {
                         error!("Telio task failed sending status report")
                     }
                 }
+                TelioTaskCmd::SetMeshnet(enabled, response_tx_channel) => {
```
lcruz99, line 128: Why are we setting/unsetting the meshmap? Shouldn't this command kill or start telio?

clis/teliod/src/daemon.rs
```diff
@@ -25,19 +31,31 @@ pub enum TelioTaskCmd {
...
     node_identity: Arc<DeviceIdentity>,
     auth_token: Arc<String>,
     mut rx_channel: mpsc::Receiver<TelioTaskCmd>,
     tx_channel: mpsc::Sender<TelioTaskCmd>,
     interface_name: &str,
+    auto_start_meshnet: bool,
+    auto_configure_interface: bool,
 ) -> Result<(), TeliodError> {
     debug!("Initializing telio device");
     let mut telio = Device::new(
```
lcruz99, line 61: shouldn't this `telio` instance be created only when Enable meshnet?

clis/teliod/src/daemon.rs
```diff
@@ -47,32 +65,112 @@ fn telio_task(
...
     // tests with core API. This is to not look for tokens in a test environment
     // right now as the values are dummy and program will not run as it expects
     // real tokens.
     if !auth_token.as_str().eq("") {
         start_telio(&mut telio, node_identity.private_key, interface_name)?;
-        task_retrieve_meshmap(node_identity, auth_token, tx_channel);
+
+        // Auto enable meshnet for the first time
+        if auto_start_meshnet {
+            task_enable_meshnet(tx_channel.clone());
```
lcruz99, line 88: this sounds a bit confusing to me. 

we have:
- `task_enable_meshnet` which actually loops a `SetMeshnet` command back to the daemon. I was expecting a `telio = Device::new(..)` and `telio.start()` instead?
- `TelioTaskCmd::SetMeshnet` should enable/disable libtelio, instead is just updating the meshmap?
- `task_retrieve_meshmap` not only retrieves the meshmap but also updates it, also enabling meshnet?

clis/teliod/src/daemon.rs
```diff
@@ -47,32 +65,112 @@ fn telio_task(
...
+        }
 
         while let Some(cmd) = rx_channel.blocking_recv() {
             info!("Got command {:?}", cmd);
             match cmd {
                 TelioTaskCmd::UpdateMeshmap(map) => {
```
lcruz99, line 94: this command doesn't exist on client side, so why have it?

clis/teliod/src/daemon.rs
lcruz99: This is not a very objective comment, I just have the feeling that this is overengineered and could be simplified by using functions with a more straightforward definition. 

I'm finding that is confusing if we are starting telio or setting a valid meshnet config.

clis/teliod/src/daemon.rs
```diff
@@ -47,32 +65,112 @@ fn telio_task(
...
+                        if let Err(e) = telio.set_config(&None) {
+                            error!("Unable to disable meshnet due to {e}");
+                            result = Err(TeliodError::DeviceError(e));
+                        }
+                    }
+                    if response_tx_channel.send(result).is_err() {
+                        error!("Telio task set meshnet failed sending result")
+                    }
+                }
+                TelioTaskCmd::SetIp(response_tx_channel) => {
```
lcruz99, line 151: Isn't this a bit trivial? If we start meshnet couldn't the interface be automatically configured? the same for the opposite

clis/teliod/src/daemon.rs
```diff
@@ -25,19 +31,31 @@ pub enum TelioTaskCmd {
...
     Quit,
 }
 
 const EMPTY_TOKEN: &str = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
 
+#[derive(Debug, Default)]
+pub struct TelioTaskStates {
+    interface_ip_address: Option<IpAddr>,
+    meshnet_is_enabled: bool,
+}
```
lcruz99, line 48: do we really need to keep track of this?

we can get the ip address from the meshmap, and check if telio is running through `telio.is_running()`

nat-lab/tests/test_teliod.py
```diff
@@ -66,7 +67,99 @@ async def test_teliod() -> None:
...
         assert not await is_teliod_running(connection)
 
-        # Run the hello-world command again - this time it should fail
+        # Run the get-status command again - this time it should fail
         with pytest.raises(ProcessExecError) as err:
             await connection.create_process(TELIOD_STATUS_PARAMS).execute()
         assert err.value.stderr == "Error: DaemonIsNotRunning"
+
+
+async def test_teliod_meshnet_toggle() -> None:
```
lcruz99, line 76: this test is testing the behaviour of the status reporting and not truly if the meshnet is working

nat-lab/tests/test_teliod.py
```diff
@@ -66,7 +67,99 @@ async def test_teliod() -> None:
...
+            '"meshnet_enabled": true'
+            in (
+                await connection.create_process(TELIOD_STATUS_PARAMS).execute()
+            ).get_stdout()
+        )
+
+        assert teliod_process.is_executing()
+
+
+async def test_teliod_set_ip() -> None:
```
lcruz99, line 126: Same as the previous test.

This is testing the status reporting and not really if the interface is correctly configured by the `set-ip` command.

I think these two tests can be made unit-tests (@ teliod sources), and I'd expect from integration tests it should test a more high-level-close-to-reality scenario. eg: Enable meshnet > set-ip and check that the interface is correctly configured and there is meshnet connection.

clis/teliod/src/system_configurator.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn remove_ip(&self, adapter_name: &str) -> Result<(), TeliodError> {
+        info!("Removing assigned IP from {}", adapter_name);
+
+        self.execute(Command::new("ifconfig").args([adapter_name, "delete"]))?;
+        self.execute(Command::new("ifconfig").args([adapter_name, "down"]))
+    }
+}
+
+/// Mock implementation of `SystemConfigurator` for testing purposes.
+pub struct MockSystemConfigurator {
```
lcruz99, line 86: This should be part of an eventual tests mod 🤔 

Also, you've defined a mock for the `SystemConfigurator` but looks like the tests are missing.

clis/teliod/src/daemon.rs
lcruz99: or as said above, we can use generics instead

nat-lab/tests/test_teliod.py
tomasz-grz: Hmm good point, but since we are not getting any API responses in NAT-LAB, we have no IP to set, we could set the fake IP though 🤔

clis/teliod/src/system_configurator.rs
tomasz-grz: The tests are in `[nat-lab/tests/test_teliod.py](https://github.com/NordSecurity/libtelio/pull/953/files/950ef94273060333616143318a543dffbd6878af#diff-ad0957b95a003082412bb199243caf3ab119b9af35cd68110a91332239101c73)`

clis/teliod/src/daemon.rs
```diff
@@ -25,19 +31,31 @@ pub enum TelioTaskCmd {
...
+
 // From async context Telio needs to be run in separate task
 fn telio_task(
     node_identity: Arc<DeviceIdentity>,
     auth_token: Arc<String>,
     mut rx_channel: mpsc::Receiver<TelioTaskCmd>,
     tx_channel: mpsc::Sender<TelioTaskCmd>,
     interface_name: &str,
+    auto_start_meshnet: bool,
+    auto_configure_interface: bool,
```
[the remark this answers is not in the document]
tomasz-grz: Good point, we could probably just pass the whole `TeliodDaemonConfig`

nat-lab/tests/test_teliod.py
tomasz-grz: That is correct, since we don't have an API token and any actual peers for the account, I am not exactly sure how we could test it. What do you propose?

clis/teliod/src/daemon.rs
tomasz-grz: `telio.is_running()` does not return the status of meshnet, but whole libtelio runtime.

Yes, in that case we would need to keep in memory the whole `meshmap` but since we only need the `IpAddr` I thought it would be an overkill.

clis/teliod/src/daemon.rs
tomasz-grz: Mainly because we would want it to be automatically updated on changes, and not a manual action from the client, for example when we receive a MQTT notification after a new peer was added.

clis/teliod/src/daemon.rs
tomasz-grz: Starting and stopping telio itself is probably much more costly in terms of operations (starting the runtime creating the tunnel interface, sockets etc) than simply setting the meshnet map. 
Correct me if I'm wrong, but I recall the [recommended](https://github.com/NordSecurity/libtelio/blob/main/src/doc/introduction.md#creating-a-meshnet) way of enabling and disabling meshnet was through `set_config()`, with disabling simply being `device.set_config(&None)`.

Also if we couple starting telio device to enabling meshnet, we would not be able to support the use case of connecting to an VPN exit node without meshnet being enabled first. But maybe we should clarify with PO if we would ever want to support this use case for teliod, at least in the desktop apps it is a standard use-case.

tomasz-grz: Hmm `task_retrieve_meshmap` is also set as the callback to the `NotificationCenter` so it has to retrieve and update meshmap in one step. Though there is the check to prevent it from setting meshmap if meshnet is disabled.
```
if state.meshnet_is_enabled {
  telio.set_config(&Some(map)) {}
}
```

I tried to decouple it in the beginning but then it would require even more loop back commands that made the flow even more complicated to follow.. What changes would make it clearer?

clis/teliod/src/daemon.rs
tomasz-grz: Please refer to this [comment](https://github.com/NordSecurity/libtelio/pull/953#discussion_r1853572042)

clis/teliod/src/daemon.rs
tomasz-grz: Let's discuss this further under this [thread](https://github.com/NordSecurity/libtelio/pull/953#discussion_r1853572042)

clis/teliod/src/command_listener.rs
tomasz-grz: Good idea, I'll take a look how to make it cleaner.

tomasz-grz:
> I wonder if the `set-ip` command is useful or has any usecase. What if we set the interfaces whenever we enable/disable the meshnet?

Not necessarily, I went back and forth between having it separate and not, based on the conflicting feedback and discussions we had so far. So in the end chose to keep it separate for greater flexibility, the client side command can be easily removed.

clis/teliod/src/daemon.rs
tomasz-grz: It actually does, `UpdateMeshmap` has this call
```rust
if address_changed && auto_configure_interface {
  task_set_ip(tx_channel.clone());
}
```

clis/teliod/src/daemon.rs
lcruz99: The tests are not the main topic of this PR, however I don't see why we define a different behaviour for integration tests. Production code should not "be aware" if we are doing any integration tests. Therefore, I think this should be removed and if there's any mocking to be done should be to help unit testing.

clis/teliod/src/daemon.rs
lcruz99: > Starting and stopping telio itself is probably much more costly in terms of operations (starting the runtime creating the tunnel interface, sockets etc) than simply setting the meshnet map.
Correct me if I'm wrong, but I recall the [recommended](https://github.com/NordSecurity/libtelio/blob/main/src/doc/introduction.md#creating-a-meshnet) way of enabling and disabling meshnet was through set_config(), with disabling simply being device.set_config(&None).

The usecase I see to stop libtelio through `teliod` is for instace if we stop the meshnet to enable a TelioFeature or change the config, in that case it will be not possible as you'll have to restart `telio::Device`. I think the cost of doing that should be somewhat trivial in this scenario.

> Also if we couple starting telio device to enabling meshnet, we would not be able to support the use case of connecting to an VPN exit node without meshnet being enabled first. But maybe we should clarify with PO if we would ever want to support this use case for teliod, at least in the desktop apps it is a standard use-case.

For QNAP usecase, VPN/exit-node feature must be disabled. Nevertheless, in that scenario I can't see why we shouldn't be able to do both `start telio > connect to exit node` and `start telio > connect to meshnet`.

clis/teliod/src/daemon.rs
lcruz99: makes sense.

clis/teliod/src/daemon.rs
```diff
@@ -47,32 +65,112 @@ fn telio_task(
...
+                        .ip_addresses
+                        .as_deref()
+                        .and_then(|addresses| addresses.first().cloned());
+                    let address_changed = state.interface_ip_address != new_ip_address;
+                    state.interface_ip_address = new_ip_address;
+
+                    if state.meshnet_is_enabled {
+                        match telio.set_config(&Some(map)) {
+                            Ok(_) => {
+                                // configure the interface ip address
```
lcruz99, line 105: would be nice to have also a log when meshmap is updated successfully.

nat-lab/tests/test_teliod.py
lcruz99: API is mocked for every nat-lab tests that use telio client (`tcli`), check: [API](https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/mesh_api.py#L208) and [clients](https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/helpers.py#L216) setup.

I propose a real scenario, launching `teliod` and testing the meshnet connection between two or more peers.

nat-lab/tests/test_teliod.py
lcruz99: > since we are not getting any API responses in NAT-LAB

We are, check my[ previous comment](https://github.com/NordSecurity/libtelio/pull/953#discussion_r1853691588).

I guess we'll need to be able to set in `teliod` the meshmap the nat-lab API provides us.

clis/teliod/src/system_configurator.rs
lcruz99: As mentioned [here](https://github.com/NordSecurity/libtelio/pull/953#discussion_r1853636572), tests from [nat-lab/tests/test_teliod.py](https://github.com/NordSecurity/libtelio/pull/953/files/950ef94273060333616143318a543dffbd6878af#diff-ad0957b95a003082412bb199243caf3ab119b9af35cd68110a91332239101c73) can be made unit tests. We shouldn't be modifying production code for integration testing.

clis/teliod/src/daemon.rs
tomasz-grz: Hmm `TelioFeature` is part of `Device::new()` and not `telio.start()`, so we would have to re-create the whole device not just re-start it. We would have to keep track of more states, I'm worried if that would make things even harder to follow.

Is it part of the requirements that we dynamically change the `TelioFeatures` during run time? 

Also the config is passed as an argument to `teliod`, so we would have to restart whole of `teliod` either way if the config changes.

clis/teliod/example_teliod_config.json
```diff
@@ -3,5 +3,6 @@
     "log_file_path": "example_log_file.log",
     "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
     "app_user_uid": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
-    "interface_name": "utun10"
+    "interface_name": "utun10",
+    "interface_config_provider": "manual"
```
packgron, line 7: love the idea of this, but maybe it makes more sence to use:
```json
"interface": {
   "provider": "manual",
   "name": "utun10",
   ...
}
```
Or is there a requirement to keep this flat ?

clis/teliod/src/command_listener.rs
```diff
@@ -105,13 +99,10 @@ mod tests {
...
 
-    // Create a "random" socket path for the test, since tests run in parallel they can deadlock
+    // Create a random socket path for the test, since tests run in parallel they can deadlock
     fn make_socket_path() -> String {
-        let nanos = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .subsec_nanos();
-        format!("{}_{}", TEST_SOCKET_PATH, nanos)
+        let mut rng = rand::thread_rng();
```
packgron, line 104: got a collision ?
Maybe it makes more secnce to have some lazy_static! BUSY_LIST ?

clis/teliod/src/interface_configurator.rs
```diff
@@ -0,0 +1,120 @@
+use crate::TeliodError;
+use serde::Deserialize;
+use std::net::IpAddr;
+use std::process::Command;
+use tracing::{error, info};
+
+pub trait InterfaceConfigurator {
```
packgron, line 7: NIT: trait name should tell more about functionality provided, not describe an object, thing From, Into vs Converter, Diplay vs Fromatter etc. So i would recommend to name this "ConfigureInterface" vs some "Xinator"

clis/teliod/src/interface_configurator.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Configure the IP address and routes for a given interface
+    fn set_ip(&self, _adapter_name: &str, _ip_address: &IpAddr) -> Result<(), TeliodError> {
+        Ok(())
+    }
+
+    /// Helper function to execute a system command
+    fn execute(&self, command: &mut Command) -> Result<(), TeliodError> {
```
packgron, line 14: This has nothing to do with the trait it is defined on, move it to a free function or a seperate trait if it has a need for specilization

clis/teliod/src/interface_configurator.rs
```diff
@@ -0,0 +1,120 @@
...
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            error!("Error executing command {:?}: {:?}", command, stderr);
+            Err(TeliodError::SystemCommandFailed(stderr.into()))
+        }
+    }
+}
+
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum InterfaceConfigurationProvider {
```
packgron, line 31: NIT: might save some boiler plate: https://lib.rs/crates/enum_dispatch

clis/teliod/src/interface_configurator.rs
tomasz-grz: Good point

clis/teliod/example_teliod_config.json
tomasz-grz: Good idea, i don't think there is a requirement, but I'll try this out

clis/teliod/src/interface_configurator.rs
tomasz-grz: I like the idea, but unfortunately it's creating a bit of a conflict with `#[derive(Deserialize)]` that would take me a bit of time to sort out. I added a note to follow up on this.

clis/teliod/src/command_listener.rs
tomasz-grz: I was using time nanos as "random" value to avoid additional dependencies, but they were not precise enough, 1/10 times having two test starting almost instantaneously. `rand` is simple enough to get this sorted for now.

clis/teliod/src/command_listener.rs
tomasz-grz: Added `handle_response` helper function to handle responses in a more generic way.

clis/teliod/src/daemon.rs
tomasz-grz: Moved these out of this PR

clis/teliod/src/daemon.rs
tomasz-grz: Moved enabling meshnet out of this PR

nat-lab/tests/test_teliod.py
tomasz-grz: Removed these tests from this PR

clis/teliod/src/system_configurator.rs
tomasz-grz: Removed the "Mock" implementation in favor of "Manual" interface config

clis/teliod/src/daemon.rs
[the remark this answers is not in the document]
tomasz-grz: Instead of giving the full `TeliodDaemonConfig` for now this is taking only an `InterfaceConfig`

clis/teliod/src/daemon.rs
tomasz-grz: Added log

clis/teliod/src/interface_configurator.rs
tomasz-grz: Renamed trait to `ConfigureInterface` as well as implementations.

packgron (commented):
+1.0
I feel like it would be nice so that `"interface": {} would also be a valid way to signal to setup interface (with platform specific defaults, name/provider). But i think it can be part of seperate PR

lcruz99 (approved):
+1

PR #955, opened 2024-11-18
Collect DiagnosticReports from Mac VM after all test are completed

tomaszklak:
### Problem
It's not always clear if/when/why a process crashed on mac vm in nat-lab.

### Solution
Collect diagnostic reports from mac vm which contain crashes.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -8,7 +8,7 @@ from interderp_cli import InterDerpClient
 from itertools import combinations
 from mesh_api import start_tcpdump, stop_tcpdump
 from utils.bindings import TelioAdapterType
-from utils.connection import DockerConnection, SshConnection
+from utils.connection import DockerConnection
 from utils.connection_util import (
     ConnectionTag,
     container_id,
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -8,7 +8,7 @@ from interderp_cli import InterDerpClient
 from itertools import combinations
 from mesh_api import start_tcpdump, stop_tcpdump
 from utils.bindings import TelioAdapterType
-from utils.connection import DockerConnection
+from utils.connection import DockerConnection, SshConnection
 from utils.connection_util import (
     ConnectionTag,
     container_id,
@@ -349,6 +349,7 @@ def pytest_sessionfinish(session, exitstatus):
         collect_nordderper_logs()
         collect_dns_server_logs()
         asyncio.run(collect_kernel_logs(session.items, "after_tests"))
+        asyncio.run(collect_mac_diagnostic_reports())
 
 
 def collect_nordderper_logs():
@@ -382,3 +383,15 @@ def copy_file_from_container(container_name, src_path, dst_path):
         )
     except subprocess.CalledProcessError:
         print(f"Error copying log file {src_path} from {container_name} to {dst_path}")
+
+
+async def collect_mac_diagnostic_reports():
+    print("Collect mac diagnostic reports")
+    async with mac_vm_util.new_connection(reenable_nat=False) as connection:
+        if isinstance(connection, SshConnection):
+            await connection.download(
+                "/Library/Logs/DiagnosticReports", "logs/system_diagnostic_reports"
+            )
+            await connection.download(
+                "/root/Library/Logs/DiagnosticReports", "logs/user_diagnostic_reports"
+            )
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -386,12 +386,19 @@ def copy_file_from_container(container_name, src_path, dst_path):
 
 
 async def collect_mac_diagnostic_reports():
+    is_ci = "GITLAB_CI" in os.environ
+    if not (is_ci or "NATLAB_COLLECT_MAC_DIAGNOSTIC_LOGS" in os.environ):
+        return
     print("Collect mac diagnostic reports")
-    async with mac_vm_util.new_connection(reenable_nat=False) as connection:
-        if isinstance(connection, SshConnection):
+    try:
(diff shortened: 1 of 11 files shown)


nat-lab/tests/conftest.py
```diff
@@ -363,3 +364,15 @@ def copy_file_from_container(container_name, src_path, dst_path):
         )
     except subprocess.CalledProcessError:
         print(f"Error copying log file {src_path} from {container_name} to {dst_path}")
+
+
+async def collect_mac_diagnostic_reports():
+    print("Collect mac diagnostic reports")
+    async with mac_vm_util.new_connection(reenable_nat=False) as connection:
+        if isinstance(connection, SshConnection):
```
gytsto, line 372: if you were to add `download` method to Connection:ABC and docker_connection, there would be no need to check if this is `SshConnection`. Creating `new_connection` from mac vm util I automatically assume I should get `SshConnection`

nat-lab/tests/conftest.py
```diff
@@ -363,3 +364,15 @@ def copy_file_from_container(container_name, src_path, dst_path):
         )
     except subprocess.CalledProcessError:
         print(f"Error copying log file {src_path} from {container_name} to {dst_path}")
+
+
+async def collect_mac_diagnostic_reports():
+    print("Collect mac diagnostic reports")
+    async with mac_vm_util.new_connection(reenable_nat=False) as connection:
```
gytsto, line 371: Why this `reenable_nat` is really needed? Maybe it has to be set to `False` in some other places too.

nat-lab/tests/conftest.py
mathiaspeters: I like this idea in general, but I think that would make sense in that case to add an OS check in that function. Yes, a mac connection will always be `SshConnection` but an `SshConnection` can also be a windows connection. As long as it's only called with a mac connection it's of course not a functional issue, but I think it would be clearer if there was either a `if is_mac:` check or a comment in the function saying it's only for mac

nat-lab/tests/conftest.py
tomaszklak: To be honest I don't know, and I would prefer to keep the behaviour changes limited to what is needed for collecting mac diagnostic reports.

nat-lab/tests/conftest.py
tomaszklak: Done

nat-lab/tests/conftest.py
```diff
@@ -363,3 +364,21 @@ def copy_file_from_container(container_name, src_path, dst_path):
         )
     except subprocess.CalledProcessError:
         print(f"Error copying log file {src_path} from {container_name} to {dst_path}")
+
+
+async def collect_mac_diagnostic_reports():
+    if not (
+        "GITLAB_CI" in os.environ or "NATLAB_COLLECT_MAC_DIAGNOSTIC_LOGS" in os.environ
```
gytsto, line 371: maybe `NATLAB_SAVE_LOGS` would be enough, instead of introducing `NATLAB_COLLECT_MAC_DIAGNOSTIC_LOGS`

nat-lab/tests/utils/vm/mac_vm_util.py
```diff
@@ -14,14 +15,18 @@
 
 VM_TCLI_DIR = LIBTELIO_BINARY_PATH_MAC_VM
 VM_UNIFFI_DIR = UNIFFI_PATH_MAC_VM
+CONNECTION_TIMEOUT = 10
```
gytsto, line 18: are we sure that 10 seconds will be enough? also why this is really needed, at some point `asyncssh.connect` will timeout itself. This is kinda handled by system's default TCP connect timeout or can be configured by setting `connect_timeout` at `SSHClientConnectionOptions`

nat-lab/tests/conftest.py
```diff
@@ -363,3 +364,21 @@ def copy_file_from_container(container_name, src_path, dst_path):
...
+    ):
+        return
+    print("Collect mac diagnostic reports")
+    try:
+        async with mac_vm_util.new_connection(reenable_nat=False) as connection:
```
gytsto, line 376: while running this locally, if mac VM is not present, I'll get an exception in here. Maybe just add `if CI:` check and then raise an exception, otherwise just ignore and proceed.

nat-lab/tests/conftest.py
gytsto: And what happens if we set `reenable_nat=True` here?

tomaszklak: I don't need to disable/enable nat so even if it would work for me to do it, I don't want to do it here.

nat-lab/tests/conftest.py
tomaszklak: I suspect that most of the time, people running nat-lab don't have mac VM running, so I want to give them an option to not run this part and wait for some timeout to terminate diagnostic reports collection.

nat-lab/tests/conftest.py
gytsto: That is true in most places, that's why I was asking, maybe we don't need to do that in other places :smile: anyway, this was more of a discussion thread, not a request change other places.

nat-lab/tests/conftest.py
gytsto: :ok_hand:

nat-lab/tests/utils/vm/mac_vm_util.py
tomaszklak: I've decided to restore this to original and rely on the builtin tcp timeout.

nat-lab/tests/conftest.py
tomaszklak: Done

gytsto (approved):
Looks great +1

gytsto (approved):

gytsto (approved):

PR #957, opened 2024-11-18
[LLT-5479] Redirect go logs to Telio logger

stalowyjez:
### Problem
There are some logs logged by wireguard-go (especially wintun) which are not forwarded to libtelio logger

### Solution
Log the golang default logger output with libtelio logger, so these logs are not lost

- [x] README.md is updated

### DIFF
--- a/wireguard-go-rust-wrapper/wireguard-go/main.go
+++ b/wireguard-go-rust-wrapper/wireguard-go/main.go
@@ -22,6 +22,7 @@ import "C"
 import (
 	"errors"
 	"fmt"
+	"log"
 	"math"
 	"math/rand"
 	"runtime"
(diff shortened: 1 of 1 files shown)


wireguard-go-rust-wrapper/wireguard-go/main.go
```diff
@@ -97,6 +98,13 @@ func tracef(format string, args ...interface{}) {
 	trace_logger.Printf(format, args...)
 }
 
+type DevLoggerWriter struct {}
+
+func (t DevLoggerWriter) Write(p []byte) (n int, err error) {
+	debugf(string(p[:]))
```
matislovas, line 104: Does this only have a debug logs? Or can we differentiate?

stalowyjez: I'm afraid that it provides only redirection.  It seems that wintun has only INFO and ERROR logs and I felt like making them DEBUG makes some sense, replacing both of them with DEBUG sounds better to me than making INFO -> ERROR or ERROR -> INFO (they are not our logs and we add them now mostly for debugging purposes anyway).

matislovas: Yeah, maybe we can leave the `debug` only, but maybe than mark them as `wintun-log` or sth similar? Do distinguish them from the rest of the logs.

stalowyjez: @matislovas added some tag

mathiaspeters (approved):
+1

PR #958, opened 2024-11-19
[LLT-5789] PQ: impose timeout on all socket recv() calls

matszczygiel:
### Problem
PQ connection can hang if the server is not responding - this happens when the session limit is hit

### Solution
Impose timeout to socket `recv()` calls to force active reconnection. The timeout was imposed but not in a strongly typed manner and there was one place where `recv()`  was not protected by the timeout and according to the logs that was the place where PQ connection was hanging

- [x] README.md is updated


lcruz99:
LGTM, +1.5

lcruz99:
I haven't found any tests for this fetching/rekeying behaviours, maybe this blocking calls would have been caught.

Can you add some?

matszczygiel:
@lcruz99 there are multiple tests
* unit tests for packet parsing - https://github.com/NordSecurity/libtelio/blob/main/crates/telio-pq/src/proto.rs#L400
* fuzzing - https://github.com/NordSecurity/libtelio/tree/main/crates/telio-pq/fuzz
* and nat-lab tests: https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/test_pq.py

This specific issue occurs in rare conditions that are particularly hard to encapsulate in a test. It is observed when connecting rapidly to the production server multiple times, and as a result, the server starts rate-limiting the session.
It cannot be reproduced in nat-lab or unit tests because it requires an instance of a production server. Well maybe it is doable but that couldn't have been foreseen while writing synthetic tests IMHO.

crates/telio-pq/src/proto.rs
```diff
@@ -187,13 +184,13 @@ async fn handshake(
     }
 
     // The response should be 92, so the buffer is sufficient
-    let read = sock.recv(&mut pkgbuf).await?;
+    let pkg = sock.recv(&mut pkgbuf).await?;
```
tomaszklak, line 187: This is the place that was missing the timeout? It can be tested by starting a udp server that will ignore the handshake that is sent above and will not reply to it in any way. This way you can check in a test that after ~RECV_TIMEOUT the `async fn handshake` will return with an error.

matszczygiel: Yes that was the place

tomaszklak:
Why not also add a send timeout?

crates/telio-pq/src/proto.rs
matszczygiel: Do you mean nat-lab test? The `handshake` function is not exposed

tomaszklak: I mean pure rust test.

matszczygiel:
@tomaszklak any reason to add a `send()` timeout? With `recv()` is not up to the software in question when it returns

To give some context - this change ensures that the PQ requests are send in a loop. Imagine the server drops the request but will answer for the second one and the client is still stuck waiting for the first one.
This change is to prevent such a situation.

Jauler:
A quick question, I wonder maybe using [set_read_timeout](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.set_read_timeout) would be simpler when compared to adding additional wrapper?

matszczygiel:
@Jauler its tokio socket. It's using async runtime (`epoll()`) and doesn't provide timeout - it sort of doesn't make any sense in async context

Jauler:
> @lcruz99 there are multiple tests
> 
>     * unit tests for packet parsing - https://github.com/NordSecurity/libtelio/blob/main/crates/telio-pq/src/proto.rs#L400
> 
>     * fuzzing - https://github.com/NordSecurity/libtelio/tree/main/crates/telio-pq/fuzz
> 
>     * and nat-lab tests: https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/test_pq.py
> 
> 
> This specific issue occurs in rare conditions that are particularly hard to encapsulate in a test. It is observed when connecting rapidly to the production server multiple times, and as a result, the server starts rate-limiting the session. It cannot be reproduced in nat-lab or unit tests because it requires an instance of a production server. Well maybe it is doable but that couldn't have been foreseen while writing synthetic tests IMHO.
> 
> What kind of specific test was on your mind?

We could try probably to make the server never respond and verify in the test that client retries the request after some time

Jauler:
> @Jauler its tokio socket. It's using async runtime (`epoll()`) and doesn't provide timeout - it sort of doesn't make any sense in async context

Oh yeah, you are right. NVM with `set_read_timeout()` then.

lcruz99:
@matszczygiel I had in mind tests mostly for the key/rekeying mechanism. Idk how hard it would be, I guess the handshake had to be done and then test the rest of the possible behaviours.

crates/telio-pq/src/proto.rs
matszczygiel: Let me think of something

matszczygiel:
I'll try to add some tests.
I'd like to express my opinion that the type system now ensures the timeout is used, so testing this behavior is superfluous.
I was sensible to test this behavior before changing the type system as it was not ensured, but I've never considered silent servers to be an issue.
Anyway that's my opinion

matszczygiel:
@tomaszklak @lcruz99 Ok. I did add tests for the timeout. They are in a separate commit

crates/telio-pq/src/proto.rs
matszczygiel: I've added tests for `fetch_keys` and `rekey` since those are the interface functions

tomaszklak:
> I'll try to add some tests. I'd like to express my opinion that the type system now ensures the timeout is used, so testing this behavior is superfluous. It was sensible to test this behavior before changing the type system as it was not ensured, but I've never considered silent servers to be an issue. Anyway that's my opinion

How do you know that what you did is doing what you **think** it's doing? First, there is (was) no test that this newtype is enforcing timeout. Second, it maybe enforced now but this doesn't prevent future work (new features in this area, or refactorings) from unknowingly breaking that behaviour and there is no test that ensures timeout is applied when receiving handshake response (not to mention other places, but let's say this is out of scope for this review 😄 ).

tomaszklak (approved):
+1

matszczygiel:
Oh, @lcruz99 I thought you gave +1.5 before and I clicked "merge" after seeing +1 from Tomasz. Is the current state of the MR OK for you? or should I revert?

PR #961, opened 2024-11-19
Add new ffi function to unset the global logger

tomaszklak:
The unsetting of the global logger will wait for all the messages (that have been in the queue) to be logged, when it was called. This should prevent the user provided logging callback to be ever called after the `unset_global_logger` call finished.

- [x] README.md is updated

### DIFF
--- a/src/ffi/logging.rs
+++ b/src/ffi/logging.rs
@@ -4,9 +4,10 @@ use std::{
     sync::{
         atomic::AtomicUsize,
         mpsc::{sync_channel, RecvError, SyncSender},
-        Mutex,
+        Arc, Barrier, Mutex,
     },
     thread::Builder,
+    time::Instant,
 };
 
 use once_cell::sync::Lazy;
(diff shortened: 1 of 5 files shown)


src/ffi/logging.rs
```diff
@@ -310,9 +364,62 @@ mod test {
     }
 
     #[test]
-    fn slow_telio_cb_handling() {
-        const LOGS_TO_DROP: usize = 5;
+    fn test_trace_via_slow_telio_cb() {
```
packgron, line 367: I would like to see a test there we call a stopper in middle of an act, as we are not testing functionality actually used by apps. as FfiCalback drop will not be practically used then using global logger: (see [example]( https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f7a8eed276096ff8f62c8819f5c4aa40) )

But it's used then using with_default.

src/ffi/logging.rs
```diff
@@ -25,6 +26,44 @@ pub const START_ASYNC_LOGGER_MSG: &str = "Starting async logger thread";
...
+    }
+
+    pub fn stop(&self) {
+        match *self.sender.lock() {
+            Some(ref sender) => {
+                let start = Instant::now();
+                let b = Arc::new(Barrier::new(2));
+                match sender.try_send(LogMessage::Quit(b.clone())) {
+                    Ok(()) => {
+                        b.wait();
```
packgron, line 53: I'm thinking could two threads calling this cause a deadlock, this place feel's safeish, as sender is exclusive to one thread, and you wait for a barier. but if somehow we get a both FfiCallback drop and stop called at the same time from differnet threads, on barier will never be awaited for no ?

packgron: On the other hand you are either using drop, or you are using with_default, but not both at the same time

packgron: Well for test's we could be using stop + drop + with_default, but it's a more controlled enviroment

packgron: Making queue size of 1 would ensure safety, but we would drop to many messages :/

packgron: Oh maybe not as queue is per sender, if i remember correctly

src/ffi/logging.rs
tomaszklak: I think the update `test_logger` integration test does that.

src/ffi/logging.rs
packgron: hmm, i guess it's finish. Although i feel it would be safer to simply have a lock on barrier it self, basicly trying to ensure that only one thread can wait hold it at a time. But it's strart feeling like over compication for non reachable code :sweat_smile:, leaving to you to decide. and resolve this.

packgron (commented):
+1.0

packgron (approved):

PR #967, opened 2024-11-20
Collect tcpdumps from gateways

tomaszklak:
### Problem
We don't collect tcpdumps from all gateways.

### Solution
tcpdump pcaps are collected from all gateways, once per session. To make it faster, it's done concurrently.

- [x] README.md is updated


Jauler (approved):
+2.0

PR #973, opened 2024-11-21
[LLT-5607] Reduce derp warning logs test

stalowyjez:
THIS IS ONLY A TEST

### Problem
Too much derp warnings when connection is just broken

### Solution
Print only some of them

### DIFF
--- a/crates/telio-relay/src/derp.rs
+++ b/crates/telio-relay/src/derp.rs
@@ -265,7 +265,16 @@ impl State {
                         break (server, conn);
                     }
                     Err(err) => {
-                        telio_log_warn!("({}) Failed to connect: {}", Self::NAME, err);
+                        if config.servers.current_server_num == config.servers.servers.len() {
+                            telio_log_warn!(
+                                "({}) Failed to connect to any of {} servers: {}",
+                                Self::NAME,
+                                config.servers.servers.len(),
+                                err
+                            );
+                        } else if config.servers.current_server_num == 1 {
+                            telio_log_warn!("({}) Failed to connect: {}", Self::NAME, err);
+                        }
                         last_disconnection_reason = err.into();
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


crates/telio-relay/src/derp.rs
```diff
@@ -273,7 +273,16 @@ impl State {
                         break (server, conn);
                     }
                     Err(err) => {
-                        telio_log_warn!("({}) Failed to connect: {}", Self::NAME, err);
+                        if config.servers.current_server_num == config.servers.servers.len() {
+                            telio_log_warn!(
+                                "({}) Failed to connect to any of {} servers: {}",
+                                Self::NAME,
```
packgron, line 279: I think we should add start adding a tracing span with it's name eneter upon wait_with_update, to remove this Self::NAME here, and make it a default.
But probably as a queue'd issue.

packgron (commented):
+1.0

crates/telio-relay/src/derp.rs
```diff
@@ -273,7 +273,16 @@ impl State {
                         break (server, conn);
                     }
                     Err(err) => {
-                        telio_log_warn!("({}) Failed to connect: {}", Self::NAME, err);
+                        if config.servers.current_server_num == config.servers.servers.len() {
```
mathiaspeters, line 276: This if-statement is confusing to me, could you explain the conditions a bit?

stalowyjez: The idea behind that is to print this log only after going through the whole list of servers - i.e. when all of the servers from the list failed. The number after using the last server is one higher than the last server, so equal to number of servers.

mathiaspeters: I wasn't looking at the code beyond what has changed. I' wondering what benefit there is to knowing more detailed why we can't connect to the first server, but not the others. Maybe I'm missing something but it seems to me like it would be enough with the `if`, without the `else if`

stalowyjez: Just like this?

crates/telio-relay/src/derp.rs
stalowyjez: Added task for this one: https://nordsec.atlassian.net/browse/LLT-5960

PR #974, opened 2024-11-22
Add DERP userspace keepalives

dfetti:
This PR adds DERP userspace keepalive to libtelio. (Unbatched for now)

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -327,6 +327,10 @@ pub struct FeatureDerp {
     pub tcp_keepalive: Option<u32>,
     /// Derp will send empty messages after this many seconds of not sending/receiving any data [default 60s]
     pub derp_keepalive: Option<u32>,
+    /// Poll Keepalive: Application level keepalives meant to replace the TCP keepalives
+    /// They will use derp_keepalive as interval
+    #[serde(default)]
+    pub poll_keepalive: bool,
     /// Enable polling of remote peer states to reduce derp traffic
     pub enable_polling: Option<bool>,
     /// Use Mozilla's root certificates instead of OS ones [default false]
@@ -536,6 +540,7 @@ mod tests {
             "derp": {
                 "tcp_keepalive": 13,
                 "derp_keepalive": 14,
+                "poll_keepalive": true,
                 "enable_polling": true,
                 "use_built_in_root_certificates": true
             },
@@ -626,6 +631,7 @@ mod tests {
                     derp: Some(FeatureDerp {
                         tcp_keepalive: Some(13),
                         derp_keepalive: Some(14),
+                        poll_keepalive: true,
                         enable_polling: Some(true),
                         use_built_in_root_certificates: true,
                     }),
--- a/crates/telio-relay/src/derp.rs
+++ b/crates/telio-relay/src/derp.rs
@@ -127,16 +127,21 @@ struct State {
 /// Keepalive values that help keeping Derp connection in conntrack alive,
 /// so server can send traffic after being silent for a while
 /// *derp_keepalive* is also used as an interval for retrieving remote peer states.
-#[derive(Debug, Clone, PartialEq, Eq)]
+/// PollKeepalive is a feature that is meant to replace TCP keepalives with
+/// application level keepalives (DerpPollRequest).
+/// More info on PollKeepalive can be found in RFC-LLT-0070
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct DerpKeepaliveConfig {
     tcp_keepalive: u32,
     derp_keepalive: u32,
+    poll_keepalive: bool,
 }
 
 impl From<&Option<FeatureDerp>> for DerpKeepaliveConfig {
     fn from(derp: &Option<FeatureDerp>) -> Self {
         let mut tcp_keepalive = proto::DERP_TCP_KEEPALIVE_INTERVAL;
         let mut derp_keepalive = proto::DERP_KEEPALIVE_INTERVAL;
+        let mut poll_keepalive = false;
         if let Some(derp) = derp {
             if let Some(tcp_ka) = derp.tcp_keepalive {
                 tcp_keepalive = tcp_ka;
@@ -144,11 +149,13 @@ impl From<&Option<FeatureDerp>> for DerpKeepaliveConfig {
             if let Some(derp_ka) = derp.derp_keepalive {
                 derp_keepalive = derp_ka;
             }
+            poll_keepalive = derp.poll_keepalive;
         }
 
         DerpKeepaliveConfig {
             tcp_keepalive,
             derp_keepalive,
+            poll_keepalive,
         }
     }
 }
@@ -156,6 +163,7 @@ impl From<&Option<FeatureDerp>> for DerpKeepaliveConfig {
 const DEFAULT_SERVER_KEEPALIVE_CONFIG: DerpKeepaliveConfig = DerpKeepaliveConfig {
     tcp_keepalive: proto::DERP_TCP_KEEPALIVE_INTERVAL,
     derp_keepalive: proto::DERP_KEEPALIVE_INTERVAL,
+    poll_keepalive: false,
 };
 
 /// Derp configuration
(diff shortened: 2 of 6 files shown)


crates/telio-relay/src/derp.rs
```diff
@@ -127,35 +127,43 @@ struct State {
 /// Keepalive values that help keeping Derp connection in conntrack alive,
 /// so server can send traffic after being silent for a while
 /// *derp_keepalive* is also used as an interval for retrieving remote peer states.
+/// PollKeepalive is a feature that is meant to replace TCP keepalives with
+/// application level keepalives (DerpPollRequests).
```
LukasPukenis, line 131:
```suggestion
/// application level keepalives (DerpPollRequest).
```

crates/telio-relay/src/derp.rs
```diff
@@ -726,8 +734,9 @@ impl Runtime for State {
                     },
                     // On tick send derp poll request to derp stream
                     Some((permit, _)) = wait_for_tx(&c.comms_direct.tx, poll_timer_tick) => {
-                        if config.enable_polling {
+                        if config.enable_polling || config.server_keepalives.poll_keepalive.is_some() {
```
LukasPukenis, line 737: I think that maybe, for simplification of things, it makes sense to leave `enable_polling` responsible just for polling(as it is currently on mainline) but the new flag just control which endpoint on DERP it connects to and controls server->client logic instead?

This way the same parameter for polling could be reused when connecting to derp/v1 and derp/v2 and the flags would be separated, resulting in this simple understanding of behaviour:
- `enable_polling` - enables polling optimisation
- `poll_keepalive` - connects to derp/v2 turning off derp->libtelio keepalives(this has no effect on libtelio)
- `enable_polling` + `poll_keepalive` -> enable polling optimisation and connect to derp/v2 to disable derp->libtelio optimisation


This should make both flags separated and they would work separately and together.
This would result in simpler logic. For example now `poll_keepalives` take preference over `enable_polling` which is hidden from the API user.

What's your opinion on this?

crates/telio-relay/src/derp/http.rs
```diff
@@ -154,6 +202,7 @@ async fn connect_and_start<RW: AsyncRead + AsyncWrite + Send + 'static>(
     secret_key: SecretKey,
     server_keepalives: DerpKeepaliveConfig,
     host: &str,
+    derp_v2: bool,
```
LukasPukenis, line 205: I suggest using an enum in this case, boolena is not very descriptive when speaking about versions and it also makes lookup more difficult. Having something like `DerpVersion::V2` being passed would make it very clear what this parameter is about

nat-lab/tests/test_derp_connect.py
```diff
@@ -15,14 +15,40 @@
...
+) -> List[SetupParameters]:
+    features = default_features()
+    if enable_poll:
+        assert features
+        features.derp = FeatureDerp(
+            tcp_keepalive=120,
+            derp_keepalive=120,
+            poll_keepalive=15,
+            enable_polling=False,
+            use_built_in_root_certificates=False,
```
LukasPukenis, line 29: it has `if enable_poll` and `enable_polling=False`. Natlab test in this case doesn't test the new feature which is understandable since derp image is still of v1, however with the separation I've written in another comment, natlab test could test with bth polling enabled and disabled(as it would only result in server not pinging libtelio). I think it's worth pursuing.

LukasPukenis (commented):
In general looks good but needs some improvements

crates/telio-relay/src/derp.rs
dfetti: Changed as we discussed

crates/telio-model/src/features.rs
```diff
@@ -536,6 +540,7 @@ mod tests {
             "derp": {
                 "tcp_keepalive": 13,
                 "derp_keepalive": 14,
+                "poll_keepalive": true,
```
tomaszklak, line 543: add a test that verifies what will be deserialized if the field is missing

crates/telio-relay/src/derp/http.rs
```diff
@@ -89,21 +94,67 @@ pub async fn connect_http_and_start(
...
+                    config.connect(server_name, stream).await?,
+                    addr,
+                    derp_config.secret_key.clone(),
+                    derp_config.server_keepalives.clone(),
+                    &hostport,
+                    DerpVersion::V2,
+                )
+                .await
+            }
+        };
```
tomaszklak, line 145: This section seems to be mostly copy-pasted from below. Please extract it to avoid duplication.

crates/telio-relay/src/derp/http.rs
```diff
@@ -204,24 +260,38 @@ async fn connect_http<R: AsyncRead + Unpin, W: AsyncWrite + Unpin>(
...
-                server_keepalives.derp_keepalive,
-            )
-            .as_bytes(),
-        )
-        .await?;
+                Poll-Keepalive: {}\r\n\r\n",
+            telio_utils::version_tag(),
+            std::env::consts::OS,
+            server_keepalives.derp_keepalive,
+        ),
```
tomaszklak, line 277: This is duplicated with below - only path changes. Please remove the duplication.

crates/telio-relay/src/derp/http.rs
```diff
@@ -89,21 +94,67 @@ pub async fn connect_http_and_start(
...
-        keepalive_idle: Some(TCP_KEEPALIVE_IDLE),
-        keepalive_intvl: Some(TCP_KEEPALIVE_INTERVAL),
-        nodelay_enable: Some(true),
-        user_timeout: Some(TCP_USER_TIMEOUT),
-        buf_size: SocketBufSizes {
-            tx_buf_size: Some(SOCK_BUF_SZ),
-            rx_buf_size: Some(SOCK_BUF_SZ),
-        },
-    }))?;
+    // First try to use derp v2, with poll keepalives it the feature is enabled
```
tomaszklak, line 97: "it" => "if" ?

crates/telio-relay/src/derp/http.rs
```diff
@@ -89,21 +94,67 @@ pub async fn connect_http_and_start(
...
+                )
+                .await
+            }
+        };
+
+        if connection.is_ok() {
+            // We managed to successfully connect to derp v2
+            telio_log_debug!("Successfully connected to derp v2");
+            return connection;
+        }
```
tomaszklak, line 151: please add a warning if the connection is no ok - completely silently falling back to v1 is probably not what we want.

crates/telio-model/src/features.rs
dfetti: It already exists `test_empty_derp`

tomaszklak (commented):
+1

crates/telio-relay/src/derp/http.rs
```diff
@@ -204,24 +232,38 @@ async fn connect_http<R: AsyncRead + Unpin, W: AsyncWrite + Unpin>(
...
 ) -> Result<Vec<u8>, Error> {
-    writer
-        .write_all(
-            format!(
-                "GET /derp HTTP/1.1\r\n\
+    let request = match derp_version {
+        // We try to connect to derp V2
+        DerpVersion::V2 => format!(
+            "GET /v2/derp HTTP/1.1\r\n\
                 Host: {host}\r\n\
```
lcruz99, line 241: nitpick: Different identation from the `DerpVersion::V1` case, is it on purpose?

crates/telio-relay/src/derp/http.rs
```diff
@@ -240,51 +282,151 @@ async fn connect_http<R: AsyncRead + Unpin, W: AsyncWrite + Unpin>(
...
-                request.headers().get("keep-alive"),
-                Some(&HeaderValue::from_static("tcp=1, derp=2"))
-            );
-            Ok(Response::builder()
-                .body(hyper::Body::from(RESPONSE_BODY))
-                .unwrap())
+    const RESPONSE_BODY: &str = "test body";
+    const HOST: &str = "hostname";
+
+    fn default_asserts(request: &Request<Body>) -> () {
```
lcruz99, line 317: nitpick: Uneeded `-> ()` return value.

lcruz99:
+1

lcruz99 (approved):

PR #980, opened 2024-11-25
Assign random tcp port range at the start of each test for WIN VMs

matislovas:
### Problem
Derp resuable port numbers migrates to next test, therefore messes up the derp server connections.

### Solution
Have a concrete tcp port range for each test.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -1,6 +1,7 @@
 import asyncio
 import os
 import pytest
+import random
 import shutil
 import subprocess
 from helpers import SetupParameters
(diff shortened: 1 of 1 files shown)


nat-lab/tests/conftest.py
```diff
@@ -67,6 +68,52 @@ def event_loop():
...
+            print(f"win_ports_{vm_tag} process execution failed: {e}")
+            return False
+
+    return True
+
+
+@pytest.fixture(autouse=True)
+@pytest.mark.windows
+def setup_windows_ports(request):
+    test_name = request.node.name
```
gytsto, line 101: I'd suggest using `request.session.items` then you could check for `marks`.

```
@pytest.fixture(autouse=True)
@pytest.mark.windows
def setup_windows_ports(request):
    items = request.session.items
    for item in items:
        for mark in item.own_markers:
            if mark.name != "windows":
                continue

            for tag in [ConnectionTag.WINDOWS_VM_1, ConnectionTag.WINDOWS_VM_2]:
                if not asyncio.run(win_ports(tag)):
                    print(f"Failed to setup ports for {tag}")

            return
```

nat-lab/tests/conftest.py
```diff
@@ -67,6 +68,52 @@ def event_loop():
...
+                "ipv4",
+                "set",
+                "dynamic",
+                "tcp",
+                "start=" + str(start_port),
+                "num=" + str(num_ports),
+            ]).execute(on_output, on_output)
+        except ProcessExecError as e:
+            print(f"win_ports_{vm_tag} process execution failed: {e}")
+            return False
```
gytsto, line 93: nit: I'd suggest of getting rid return booleans and exception handling. Would move all exception handing in the fixture itself. That way it would be more python like.

lcruz99:
+0.5

nat-lab/tests/conftest.py
matislovas: I incorporated your suggestion. So if there's a name defined test params - than only for that VM the command exectued. If there isn't - than for both.

nat-lab/tests/conftest.py
matislovas: Fixed

nat-lab/tests/conftest.py
gytsto: :+1:

gytsto (approved):
+2

PR #988, opened 2024-11-27
[LLT-5811] qnap: Add cgi script handling REST API

lcruz99:
This PR introduces a cgi script that the QNAP web UI can execute in order to manage the `teliod` daemon process.

Note: Configuration update and authentication features are going to be tackled in a future PR.

- [x] README.md is updated

### DIFF
--- a/qnap/shared/teliod.cgi
+++ b/qnap/shared/teliod.cgi
@@ -1,0 +1,131 @@
+#!/bin/bash
+
+echo "Content-Type: application/json"
+echo ""
+
+# Varibles set by the HTTP web server
+#
+# $REQUEST_METHOD
+# $QUERY_STRING
+# $CONTENT_LENGTH
+
+POST_DATA=""
+
+if [[ "$REQUEST_METHOD" == "POST" || "$REQUEST_METHOD" == "PATCH" ]]; then
+    read -n $CONTENT_LENGTH POST_DATA
+fi
+
+QPKG_DIR="/share/CACHEDEV1_DATA/.qpkg/NordSecurityMeshnet"
+APP_TMP_DIR="/tmp/nordsecuritymeshnet"
+APP_CMD="$QPKG_DIR/teliod"
+CONFIG_FILE="$QPKG_DIR/teliod.cfg"
+LOG_FILE="$APP_TMP_DIR/teliod.log"
+PID_FILE="$APP_TMP_DIR/teliod.pid"
+
+is_application_running() {
+    if [ -f ${PID_FILE} ]; then
+        PID=$(cat ${PID_FILE})
+        if [ -d /proc/${PID}/ ]; then
+            # application is running
+            return 0
+        fi
+    fi
+    # application is NOT running
+    return 1
+}
+
+send_response() {
+    local code=$1
+    local message=$2
+    echo "{\"code\": $code, \"message\": \"$message\"}"
+    exit $code
+}
+
+start_daemon() {
+    if is_application_running; then
+        send_response 400 "Application is already running."
+    fi
+    $APP_CMD daemon $CONFIG_FILE > $LOG_FILE 2>&1 &
+    if [[ $? -eq 0 ]]; then
+        echo $! > ${PID_FILE}
+        send_response 200 "Application started successfully."
+    else
+        echo $! > ${PID_FILE}
+        send_response 500 "Failed to start the application."
+    fi
+}
+
+stop_daemon() {
+    if [ -f ${PID_FILE} ]; then
+        PID=$(cat ${PID_FILE})
+        kill ${PID} || true
+        rm -f ${PID_FILE}
+        send_response 200 "Application stopped successfully."
+    else
+        send_response 400 "Application PID not found."
+    fi
+}
+
+update_config() {
+    NEW_CONFIG=$(echo "$POST_DATA" | jq -r '.config')
+    if [[ -z "$NEW_CONFIG" ]]; then
+        send_response 400 "No configuration provided."
+    fi
+    echo "$NEW_CONFIG" > $CONFIG_FILE
+    if [[ $? -eq 0 ]]; then
+        send_response 200 "Configuration updated successfully."
+    else
+        send_response 500 "Failed to update configuration."
+    fi
+}
+
+get_status() {
+    if is_application_running; then
+        STATUS=$($APP_CMD get-status)
+        echo "{\"code\": 200, \"status-report\": $STATUS}"
+        exit 0
+    else
+        send_response 400 "Application is not running."
+    fi
+}
+
+get_logs() {
+    if [[ -f $LOG_FILE ]]; then
+        LOGS=$(tail -n 100 $LOG_FILE | jq -R -s '.')
+        echo "{\"code\": 200, \"logs\": $LOGS}"
+        exit 0
+    else
+        send_response 404 "Log file not found."
+    fi
+}
+
+case "$REQUEST_METHOD" in
+    POST)
+        if [[ "$QUERY_STRING" == "action=start" ]]; then
+            start_daemon
+        elif [[ "$QUERY_STRING" == "action=stop" ]]; then
+            stop_daemon
+        else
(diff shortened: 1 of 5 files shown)


qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,126 @@
...
+# $CONTENT_LENGTH
+
+POST_DATA=""
+
+if [[ "$REQUEST_METHOD" == "POST" || "$REQUEST_METHOD" == "PATCH" ]]; then
+    read -n $CONTENT_LENGTH POST_DATA
+fi
+
+QPKG_DIR="/share/CACHEDEV1_DATA/.qpkg/NordSecurityMeshnet"
+APP_TMP_DIR="/tmp/nordsecuritymeshnet"
```
gytsto, line 19: just curious, shouldn't this be `/tmp/nordsecurity/meshnet` ?

qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,126 @@
...
+APP_CMD="$QPKG_DIR/teliod"
+CONFIG_FILE="$QPKG_DIR/teliod.cfg"
+LOG_FILE="$APP_TMP_DIR/teliod.log"
+PID_FILE="$APP_TMP_DIR/teliod.pid"
+
+is_application_running() {
+    if [ -f ${PID_FILE} ]; then
+        PID=$(cat ${PID_FILE})
+        if [ -d /proc/${PID}/ ]; then
+            echo 1
```
gytsto, line 29: nit: it's abit weird, that we return 1 on success, while in `start_application` for other success we check if return is 0. But not sure on how to proceed with this, since my logic wants to see 1 if running, but also, wants to see 0 if success (meaning it's running)... So not really an action point, but just a comment for thinking

qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,126 @@
...
+    exit $code
+}
+
+start_application() {
+    if [ is_application_running -eq 1 ]; then
+            send_response 400 "Application is already running."
+    fi
+    $APP_CMD daemon $CONFIG_FILE &
+    if [[ $? -eq 0 ]]; then
+        echo $! > ${PID_FILE}
```
gytsto, line 47: nit: `echo $! > ${PID_FILE}` maybe move this before if statement, since we are doing same thing in both cases.

qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,126 @@
...
+    else
+        echo $! > ${PID_FILE}
+        send_response 500 "Failed to start the application."
+    fi
+}
+
+stop_application() {
+    if [ -f ${TELIOD_PID_FILE} ]; then
+        PID=$(cat ${TELIOD_PID_FILE})
+        kill -9 ${PID} || true
```
gytsto, line 58: why would kill fail here? maybe we need to do something here if process is not successfully killed, instead of leaving it hanging? asking this, since I'm not sure if removing PID file will actually remove process

qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,126 @@
...
+    echo "$NEW_CONFIG" > $CONFIG_FILE
+    if [[ $? -eq 0 ]]; then
+        send_response 200 "Configuration updated successfully."
+    else
+        send_response 500 "Failed to update configuration."
+    fi
+}
+
+get_status() {
+    if [[ -f $PID_FILE ]]; then
```
gytsto, line 80: why not use here `is_application_running`

qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,126 @@
...
+        if [[ "$QUERY_STRING" == "action=get-status" ]]; then
+            get_status
+        elif [[ "$QUERY_STRING" == "action=get-logs" ]]; then
+            get_logs
+        else
+            send_response 400 "Invalid action for GET."
+        fi
+        ;;
+    *)
+       send_response 405 "Method not allowed."
```
gytsto, line 124: feels like indentation is off here abit :smile:

lcruz99: fixed

qnap/shared/teliod.cgi
lcruz99: That also crossed my mind, what would make sense here, if we're running a bash command normally is how you described, but the return value for `is_application_running()` made sense to me to be true/false.

qnap/shared/teliod.cgi
lcruz99: in that case `$?` would return the exit code of that echo command instead of the `$APP_CMD`

qnap/shared/teliod.cgi
gytsto: whatever you decide, good for me, just wanted to point out :smile:

qnap/shared/teliod.cgi
lcruz99: removing PID file wouldn't kill the process by itself. `kill` command might fail if the process is not running.

qnap/shared/teliod.cgi
lcruz99: you're right, I'll rework this

qnap/shared/teliod.cgi
gytsto: oh yeah, true, missed that

qnap/shared/teliod.cgi
gytsto: But in this case, if process is not actually running, shouldn't we send 400 as response like in the else statement. Also abit weird that PID file exists, but process doesn't. If this behaviour expected, add some comment saying that PID file can exist, but process might be dead. Idk how to express myself on this, but it just raises questions :smile:

lcruz99: > Also abit weird that PID file exists, but process doesn't.

When `teliod` crashes the PID file is not cleaned, that is handled by the start/stop script, when the user enable/disable the app on the QTS (QNAP OS).

gytsto: ok, this part seems to be OK. Can the file cleanup itself be handled from `teliod` itself, whenever crash happens? is it some sort of bug or expected crash behavior which we cannot do anything about?

qnap/shared/teliod.cgi
```diff
@@ -77,10 +77,12 @@ update_config() {
 }
 
 get_status() {
-    if [[ -f $PID_FILE ]]; then
-        send_response 200 "Application is running."
+    if [[ is_application_running -eq 1 ]]; then
+        STATUS=($APP_CMD get-status)
```
gytsto, line 81: not sure, but maybe `$` needed?
```suggestion
        STATUS=$($APP_CMD get-status)
```

gytsto (commented):
other than those leftover questions, which are not really relevant to final result, everything looks good +1

qnap/shared/teliod.cgi
lcruz99: the file only stores the PID, doesn't necessarily mean that the process is running. I doesn't have to.. for that we can check with /proc/<PID> 

not sure if we can/should do it on `teliod`

qnap/shared/teliod.cgi
lcruz99: yes, will change

qnap/shared/teliod.cgi
```diff
@@ -0,0 +1,131 @@
...
+send_response() {
+    local code=$1
+    local message=$2
+    echo "{\"code\": $code, \"message\": \"$message\"}"
+    exit $code
+}
+
+start_daemon() {
+    if is_application_running; then
+            send_response 400 "Application is already running."
```
mathiaspeters, line 46: Nit: inconsistent indentation

mathiaspeters (dismissed):
+1

qnap/shared/teliod.cgi
lcruz99: fixed

lcruz99: fixed

mathiaspeters (approved):

PR #989, opened 2024-11-28
[LLT-5824] When shutting down natlab's telio-proxy, allow ConnectionRefusedError

Jauler:
### Problem
During shutdown via RPC, ConnectionRefusedError may happen naturally. Do not panic in case of this error.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/uniffi/libtelio_proxy.py
+++ b/nat-lab/tests/uniffi/libtelio_proxy.py
@@ -56,7 +56,7 @@ class LibtelioProxy:
                 "Unknown" if container_or_vm_name is None else container_or_vm_name,
             )
 
-        except Pyro5.errors.ConnectionClosedError as e:
+        except (Pyro5.errors.ConnectionClosedError, ConnectionRefusedError) as e:
             # Shutting down the server via client request is naturally racy,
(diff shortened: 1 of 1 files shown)


LukasPukenis (commented):
+1 nice

tomaszklak (dismissed):
+1

nat-lab/tests/uniffi/libtelio_proxy.py
```diff
@@ -56,7 +56,7 @@ def shutdown(self, container_or_vm_name: Optional[str] = None):
                 "Unknown" if container_or_vm_name is None else container_or_vm_name,
             )
 
-        except Pyro5.errors.ConnectionClosedError as e:
+        except (Pyro5.errors.ConnectionClosedError, Pyro5.errros.ConnectionRefusedError) as e:
```
gytsto, line 59: Does this really work with `||` or will it expect to catch tuple

tomaszklak: Seems like correct syntax: https://docs.python.org/3/tutorial/errors.html#handling-exceptions

gytsto (approved):

PR #991, opened 2024-11-28
[LLT-5815] pq vaidate that pq get response is coming from the upgrader

matszczygiel:
### Problem
While waiting for the handshake packet we can receive some random traffic packets not related to the PQ handshake

### Solution
By checking its src IP and port, validate that the handshake response packet comes from the PQ upgrader service.
Additionally, this PR changes the timeout logic from per `recv()` calls to imposing the timeout on the whole handshake procedure

- [x] README.md is updated


crates/telio-pq/src/conn.rs
```diff
@@ -27,29 +27,30 @@ impl ConnKeyRotation {
...
+                ));
+
+                match tokio::time::timeout(request_retry, fetch_keys).await {
+                    Ok(Ok(keys)) => {
                         telio_log_debug!("PQ keys fetched");
                         break keys;
                     }
-                    Err(err) => telio_log_warn!("Failed to fetch PQ keys: {err}"),
+                    Ok(Err(err)) => telio_log_warn!("Failed to fetch PQ keys: {err}"),
+                    Err(_timeout) => telio_log_warn!("Failed to fetch PQ keys: TIMEOUT"),
```
LukasPukenis, line 53: very minor - could you implement the timeout value in seconds in the error?

crates/telio-pq/src/conn.rs
```diff
@@ -79,11 +82,13 @@ impl ConnKeyRotation {
                         #[allow(mpsc_blocking_send)]
                         let _ = chan.send(super::Event::Rekey(wg_keys)).await;
                     }
-                    Err(err) => {
+                    Ok(Err(err)) => {
```
LukasPukenis, line 85: I didn't look into the call site, maybe it makes sense, but can't we get rid of double indirection? If `Ok()` then it should be a legit value.

nat-lab/tests/test_pq.py
```diff
@@ -372,3 +378,106 @@ async def test_dns_with_pq(
...
+
+        wg_server = config.WG_SERVER  # use non PQ server
+        await client.connect_to_vpn(
+            str(wg_server["ipv4"]),
+            int(wg_server["port"]),
+            str(wg_server["public_key"]),
+            pq=True,
+        )
+
+        await client.wait_for_log("Failed to fetch PQ keys: TIMEOUT")
```
LukasPukenis, line 421: question: what happens in the current libtelio with this testcase? there was `recv` timeout, wonder if it would just swallow the timeout or produce the error message as well?

LukasPukenis (commented):
+1 testcases show that it works, left few questions

crates/telio-pq/src/conn.rs
matszczygiel: Do you mean like printing the timeout?

crates/telio-pq/src/conn.rs
matszczygiel: The outer `Result<>` is imposed by `tokio::time::timeout`. I don't see a way to change it

nat-lab/tests/test_pq.py
matszczygiel: Semantically it should behave the same. However, there's no such log in the current released version. I've added these tests mostly to compensate for removing some tests in the `proto.rs` module

Jauler: You mean non-PQ WG connection to VPN servers? 

In that case retry logic and timeouts are handled by WireGuard implementations (neptun on apple/android, wg-nt on windows, linux-kernel-module on Linux) which generally employs constant backoff strategy (e.g. retries handshake every 5s)

crates/telio-pq/src/conn.rs
Jauler: TBH, it does represent the actual situation kind of nicely IMHO.

Because we have two results, one coming from task execution (e.g. OK means executed Err means some execution error). If we get OK, which means, that task executed till its end -> the task itself may have failed for some inner reason.

But IMHO it would be nice to print out the value of `_timeout`  on line 90. While it probably will ever print `Elapsed`, it is still kind of nice confirmation :D

crates/telio-pq/src/proto.rs
```diff
@@ -203,20 +221,52 @@ async fn handshake(
...
+fn validate_get_response_ip(ip: &Ipv4Packet) -> Result<(), &'static str> {
+    if ip.get_source() != REMOTE_IP {
+        return Err("invalid src IP");
+    }
+    if ip.get_destination() != LOCAL_IP {
+        return Err("invalid dst IP");
+    }
+    if ip.get_next_level_protocol() != ip::IpNextHeaderProtocols::Udp {
+        return Err("invalid protocol");
+    }
```
Jauler, line 255: Could we additionally validate the checksum.

I believe it should be enough just to compare [chacksum()](https://docs.rs/pnet/latest/pnet/packet/ipv4/fn.checksum.html) against [get_checksum()](https://docs.rs/pnet/latest/pnet/packet/ipv4/struct.Ipv4Packet.html#method.get_checksum)

matszczygiel: Sure

crates/telio-pq/src/conn.rs
matszczygiel: `_timeout` is of type `Elapsed` which is empty (doesn't contain the duration). But I can print the duration because I have access to it anyway

crates/telio-pq/src/proto.rs
```diff
@@ -203,20 +221,52 @@ async fn handshake(
...
+}
+
+fn validate_get_response_udp(udp: &UdpPacket, local_port: u16) -> Result<(), &'static str> {
+    if udp.get_source() != SERVICE_PORT {
+        return Err("PQ keys UDP: invalid src port");
+    }
+    if udp.get_destination() != local_port {
+        return Err("PQ keys UDP: invalid dst port");
+    }
+
```
Jauler, line 266: Could we additionally validate UDP checksum?

I believe it should be enough just to compare [ipv4_chacksum()](https://docs.rs/pnet/latest/pnet/packet/udp/fn.ipv4_checksum.html) against [get_checksum()](https://docs.rs/pnet/latest/pnet/packet/udp/struct.UdpPacket.html#method.get_checksum)

crates/telio-pq/src/conn.rs
matszczygiel: I've added the print

crates/telio-pq/src/proto.rs
matszczygiel: Sure

matszczygiel: Done

Jauler (dismissed):
+2.0

nat-lab/tests/test_pq.py
```diff
@@ -372,3 +378,108 @@ async def test_dns_with_pq(
...
+    ],
+)
+async def test_pq_vpn_upgrade_from_non_pq(
+    alpha_setup_params: SetupParameters,
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        alpha_setup_params.features.post_quantum_vpn.handshake_retry_interval_s = 1
+
+        env = await exit_stack.enter_async_context(
+            setup_environment(exit_stack, [alpha_setup_params])
```
Jauler, line 455: you need to add `prepare_vpn = True` here, as after [this PR](https://github.com/NordSecurity/libtelio/pull/949) the container preparation has been changed to be done on-deman. It save quite measurable amount of time, as otherwise containers are being prepared 1000 times or so during test-suite.

nat-lab/tests/test_pq.py
```diff
@@ -372,3 +378,108 @@ async def test_dns_with_pq(
...
+    ],
+)
+async def test_pq_vpn_silent_pq_upgrader(
+    setup: SetupParameters,
+) -> None:
+    async with AsyncExitStack() as exit_stack:
+        setup.features.post_quantum_vpn.handshake_retry_interval_s = 1
+
+        env = await exit_stack.enter_async_context(
+            setup_environment(exit_stack, [setup])
```
Jauler, line 406: Same us: https://github.com/NordSecurity/libtelio/pull/991/files#r1863529347

matszczygiel: Done. Waiting for CI to finish

PR #994, opened 2024-11-29
Print last stdout in CommandGrepper, if timeout happens

matislovas:
### Problem
We cannot see, if there's anything irregular, when using `CommandGrepper` and timeout happens.

### Solution
Print last stdout, if timeout happens.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/command_grepper.py
+++ b/nat-lab/tests/utils/command_grepper.py
@@ -8,6 +8,7 @@ class CommandGrepper:
     _connection: Connection
     _check_cmd: List[str]
     _timeout: Optional[float]
+    _last_stdout: Optional[str]
 
     def __init__(
         self,
@@ -26,6 +27,7 @@ class CommandGrepper:
         self._connection = connection
         self._check_cmd = check_cmd
         self._timeout = timeout
+        self._last_stdout = None
 
     async def check_exists(
         self, exp_primary: str, exp_secondary: Optional[List[str]] = None
@@ -49,7 +51,7 @@ class CommandGrepper:
             print(f"Process exec error: {e}")
             raise
         except TimeoutError as e:
-            print(f"Timeout error: {e}")
+            print(f"Timeout error: {e}, last stdout: {self._last_stdout}")
             return False
         except Exception as e:
             print(f"Some other exception happened: {e}")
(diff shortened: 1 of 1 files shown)


nat-lab/tests/utils/command_grepper.py
```diff
@@ -49,7 +51,7 @@ async def check_exists(
             print(f"Process exec error: {e}")
             raise
         except TimeoutError as e:
-            print(f"Timeout error: {e}")
+            print(f"Timeout error: {e}, last stdout: {self._last_stdout}")
```
gytsto, line 54: nit: maybe we can print `_last_stdout` on all exceptions, not only timeout

nat-lab/tests/utils/command_grepper.py
```diff
@@ -88,15 +90,14 @@ async def _run_check(
     ) -> bool:
         while True:
             process = await self._connection.create_process(self._check_cmd).execute()
+            self._last_stdout = process.get_stdout()
```
LukasPukenis, line 93: Might be irrelevant but question - does it return full or partial stdout via `get_stdout()`? If it's full, then the `last` part is no necessary :)

nat-lab/tests/utils/command_grepper.py
```diff
@@ -88,15 +90,14 @@ async def _run_check(
...
 
             if exists:
-                if self._check_if_exists(
-                    process.get_stdout(), exp_primary, exp_secondary
-                ):
+                if self._check_if_exists(self._last_stdout, exp_primary, exp_secondary):
```
LukasPukenis, line 96: minor: the `return` can be just inlined

nat-lab/tests/utils/command_grepper.py
matislovas: It does return full, just named it last, because it is, well, the last one received :smile:

nat-lab/tests/utils/command_grepper.py
matislovas: Yeah, I though about that, decided not, but what the hell, let's do it :smile:

gytsto: it will not hurt to have extra logs anyway :laughing:

LukasPukenis:
+1 just the lastname would be nice to rename

PR #996, opened 2024-11-29
Detach batcher and direct feature

LukasPukenis:
Batching relies on Sessionkeeper(is baked into it) which is not present when direct feature is disabled or if SessionKeeper::start fails.

This breaks cases where batching is enabled but direct feature is not for VPN reestablishment for example(added testcases fail on `main`). This is a limitation of the decision to keep SessionKeeper instead of introducing a separate entity for batching.

The fix is to override the batching feature when either SessionKeeper or Direct are None which is done here.

However I did an extra step and moved SessionKeeper outside the Direct entities as I think it doesn't belong there anymore due to being repurposed for more stuff(vpn, proxy, stun keepalives when batching, derp keepalives in progress). This moving out provided very minimal changes, though the diff looks huge.

Please see individual commits for more details

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_batching.py
+++ b/nat-lab/tests/test_batching.py
@@ -279,8 +279,9 @@ def proxying_peer_parameters(clients: List[ConnectionTag]):
         )
     ],
 )
-@pytest.mark.timeout(60)
-async def test_proxying_peer_keepalive(setup_params: List[SetupParameters]) -> None:
+async def test_proxying_peer_batched_keepalive(
+    setup_params: List[SetupParameters],
+) -> None:
     # Since batching keepalives are performed on application level instead of Wireguard
     # backend we need to ensure that proxying peers are receiving the keepalives. To test
     # for that we can enable link detection that guarantees quick detection if there's no corresponding
@@ -306,13 +307,10 @@ async def test_proxying_peer_keepalive(setup_params: List[SetupParameters]) -> N
 
         _, beta_node = env.nodes
 
-        # 20 seconds should be enough since:
-        # * 10 seconds is for WireGuard's Passive-Keepalive
-        # * RTT is configured to be 2 seconds
-        # * no pings to validate
-        # * PersistentKeepalive configured to 5seconds
-        # Given all these it should be worst case latency until detection: 10+2+5=17
-        await asyncio.sleep(20)
-
-        link_events = alpha.get_link_state_events(beta_node.public_key)
-        assert link_events[-1] == LinkState.DOWN
+        await alpha.wait_for_state_peer(
+            beta_node.public_key,
+            [NodeState.CONNECTED],
+            [PathType.RELAY],
+            timeout=30,
+            link_state=LinkState.DOWN,
+        )
--- a/nat-lab/tests/test_batching.py
+++ b/nat-lab/tests/test_batching.py
@@ -2,7 +2,7 @@ import asyncio
 import itertools
 import pytest
 from contextlib import AsyncExitStack
-from helpers import SetupParameters, setup_environment
+from helpers import SetupParameters, setup_environment, setup_mesh_nodes
 from itertools import zip_longest
 from scapy.layers.inet import TCP, UDP, ICMP  # type: ignore
 from scapy.layers.l2 import ARP  # type: ignore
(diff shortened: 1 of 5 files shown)


gytsto (commented):
+0.5

src/device/wg_controller.rs
```diff
@@ -429,38 +429,37 @@ async fn consolidate_wg_peers<
...
+
+        if let Some(sk) = session_keeper {
+            if !is_actual_peer_proxying && is_requested_peer_proxying && features.batching.is_none()
+            {
+                sk.remove_node(key).await?;
+            } else if let Some(interval) = requested_peer.batching_keepalive_interval {
+                if let Some(sk) = session_keeper {
+                    if sk.get_interval(key).await != Some(interval)
+                        || (requested_peer.peer.ip_addresses != actual_peer.ip_addresses)
+                        || (is_actual_peer_proxying != is_requested_peer_proxying)
```
LukasPukenis, line 443: Note to self: experiment with UPNP peers by not providing UPNP provider. Local testing appears to keep it oscillating

LukasPukenis: After investigation, this is not an issue but was rather caused by a bug in batcher which will need to be fixed. So this is a symptom, not the root cause.

The bug is out of scope of this PR but it is basically that using direct connections with batching will cause immediate downgrade due to overwritten keepalive value to 0

src/device.rs
```diff
@@ -2372,13 +2386,6 @@ impl TaskRuntime for Runtime {
                                 direct_entities.cross_ping_check.notify_failed_wg_connection(public_key)
                                     .await?;
                                 direct_entities.upgrade_sync.clear_accepted_session(public_key).await;
-
```
dfetti, line 2375: Why did you removed that?

LukasPukenis: I moved it into wg_controller so the logic around adding or removing stuff to/from SessionKeeper would be in one place. consolidation in wg_controller also does checks for downgraded peers so I moved this one there :)

jjanowsk (commented):
Looks good in general. Very nice commit structure. But I don't like this approach to testing. It does not scale. It's ok in UTs, we can leave it there, but I would really rework the nat-lab part

src/device.rs
```diff
@@ -1494,12 +1476,22 @@ impl Runtime {
...
         });
 
+        let session_keeper = Arc::new(SessionKeeper::start(
+            self.entities.socket_pool.clone(),
+            self.features.batching.unwrap_or_default(),
+            self.entities
+                .wireguard_interface
+                .subscribe_to_network_activity()
+                .await?,
+        )?);
```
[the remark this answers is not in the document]
jjanowsk: That's outdated. I looked commit by commit, and later that part was redone.

nat-lab/tests/test_derp_connect.py
```diff
@@ -15,13 +15,46 @@
...
+        ],
+        [
+            SetupParameters(
+                connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_1,
+                features=generate_features(batching=True),
+            ),
+            SetupParameters(
+                connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_2,
+                features=generate_features(batching=True),
+            ),
```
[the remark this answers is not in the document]
LukasPukenis: That's a fair argument. I've added those to be sure it really works. At the moment it looked like a good call. Now I suppose only one adiditonal test is enough, will refactor that way :)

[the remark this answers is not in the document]
LukasPukenis: @jjanowsk your comment about natlab was spot on. I went on refactoring the testcases until I realized a much better testcase exist. I've added last commit that removes existing ones and adds a new one. Please take a look

nat-lab/tests/test_batching.py
```diff
@@ -240,3 +242,77 @@ async def start_node_manually(client, node, sleep_s):
...
+@pytest.mark.parametrize(
+    "setup_params",
+    [
+        proxying_peer_parameters(
+            [ConnectionTag.DOCKER_CONE_CLIENT_1, ConnectionTag.DOCKER_CONE_CLIENT_2]
+        )
+    ],
+)
+@pytest.mark.timeout(60)
+async def test_proxying_peers_reconnect(setup_params: List[SetupParameters]) -> None:
```
jjanowsk, line 283: There is no reconnection in this test ;)

nat-lab/tests/test_batching.py
```diff
@@ -240,3 +242,77 @@ async def start_node_manually(client, node, sleep_s):
...
+
+        _, beta_node = env.nodes
+
+        # 20 seconds should be enough since:
+        # * 10 seconds is for WireGuard's Passive-Keepalive
+        # * RTT is configured to be 2 seconds
+        # * no pings to validate
+        # * PersistentKeepalive configured to 5seconds
+        # Given all these it should be worst case latency until detection: 10+2+5=17
+        await asyncio.sleep(20)
```
jjanowsk, line 315: Maybe instead of sleep + assert it would be better to actively wait for LinkState.DOWN? I'm afraid the current approach can cause some unnecessary flakiness.

nat-lab/tests/test_batching.py
LukasPukenis: Whoops, that was initial experiment. Fixed that :)

jjanowsk (approved):

PR #999, opened 2024-12-02
Revert "Overwrite SessionKeeper entry on Proxy->Direct upgrade"

LukasPukenis:
Reverts NordSecurity/libtelio#992

The fix was incomplete: it rather rewrote entries in SessionKeeper no matter what when iterating through `update_keys` so it fixed the alignment issue when upgrading from Proxying to Direct, however now will excessively overwrite those. Need a better solution

- [x] README.md is updated


dfetti (dismissed):
+2.0

dfetti (approved):

PR #1001, opened 2024-12-02
Fix 'overflow when adding duration to instant' error in upnp

tomaszklak:
As can be seen here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=051502d4fa626f2a1615bfd06d589ec9

using duration max with instants can lead to overflows. Instead lets use a duration that is long enough that in practice is 'indistinguishable' from infinity.

- [x] README.md is updated


LukasPukenis:
+2

LukasPukenis (approved):

tomasz-grz (approved):
+1 🚀

PR #1005, opened 2024-12-02
Print backtrace on panic

tomaszklak:
### Problem
Only the inner most backtrace frame is printed when we handle panic

### Solution
Print whole backtrace (on debug level) when panic happens. Also re-enable integer overflow checks in debug builds.

- [x] README.md is updated

### DIFF
--- a/crates/telio-utils/src/backtrace.rs
+++ b/crates/telio-utils/src/backtrace.rs
@@ -1,0 +1,31 @@
+use crate::telio_log_debug;
+use backtrace;
+
+/// Print to debug log current backtrace
+pub fn log_current_backtrace() {
+    let mut counter = 0;
+    backtrace::trace(|frame| {
+        let ip = frame.ip();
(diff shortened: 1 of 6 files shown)


crates/telio-utils/src/backtrace.rs
```diff
@@ -0,0 +1,31 @@
...
+                .unwrap_or_else(|| format!("{symbol_address:?}"));
+
+            let filename = symbol
+                .filename()
+                .map(|filename| filename.to_string_lossy().into_owned())
+                .unwrap_or_else(|| "uknown.file".to_owned());
+
+            let lineno = symbol.lineno().unwrap_or(0);
+
+            telio_log_debug!("{counter:3} {ip:?} {filename}:{lineno} {name}");
```
matislovas, line 25: nit: I would add to this string some indication that this is "Backtrace".

matislovas (approved):
+2

PR #1006, opened 2024-12-02
Emit 'connecting' event from PQ handshake

matszczygiel:
### Problem
When issuing a PQ connection request, the node state was reported as 'CONNECTING' only when the PQ handshake was finished. This led to no state being reported when the handshake was failing.

### Solution
Emit a 'CONNECTING' event upon PQ handshake


nat-lab/tests/test_pq.py
```diff
@@ -410,18 +411,44 @@ async def test_pq_vpn_silent_pq_upgrader(
         client, *_ = env.clients
 
         wg_server = config.WG_SERVER  # use non PQ server
+
```
mathiaspeters, line 414: What do you think about having this test also check that `DISCONNECTING` is being emitted?

matszczygiel: You might be surprised but the call [`disconnect_from_vpn()`](https://github.com/NordSecurity/libtelio/pull/1006/files#diff-d71adf294c60fb47c1fdaabf09dfb47c83a4566748a530105ca35e310f7f49fdR451) below does the check inside

crates/telio-pq/src/entity.rs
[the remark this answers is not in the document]
matszczygiel: Yeaaaaah, I agree, this is not nice

src/device.rs
```diff
@@ -2563,6 +2579,26 @@ fn set_tunnel_interface(socket_pool: &Arc<SocketPool>, config: &DeviceConfig) {
...
+        hostname: None,
+        allow_incoming_connections: false,
+        allow_peer_send_files: false,
+        path: PathType::Direct,
+        allow_multicast: false,
+        peer_allows_multicast: false,
+        ..Default::default()
+    }
+}
+
```
LukasPukenis, line 2601: I wonder a bit about this code, a lot of fields have hardcoded values such as `allow_multicast: false` as an example. Is this the right conversion?

Nitpick: wouldn't it be more rustacean way to have a `From<T>` implemented and then use that instead?

matszczygiel: This is a helper function local to this module. This specific conversion was used, and I needed to make the same conversion somewhere else (within the module), so my idea was to extract this conversion into a helper function.
It isn't meant to be fully general though maybe it can be made so.

Regarding `From<>`
* both types are from a different crate (telio-model) - you cannot implement a trait on the foreign type, though it should be possible in `telio-model` crate
* as mentioned it wasn't meant to be general

So IDK how "correct" this conversion is so that it can be promoted to general and `From<>` be implemented. @Jauler any opinion on that?

Jauler: IMHO when implementing `From<X>` care must be taken to either:
* Make sure that transition is possible strictly into one direction (e.g. ExitNode -> Node, but not `Node -x-> ExitNode`. But you can always add more `From<T>` implementations, hence this is really hard to ensure.
or:
* Make sure that transition ExitNode -> Node -> ExitNode does not loose any information (note that the chain may be longer than two types).

Previously we had `From<T>` implemented for many pairs of types, especially various `Node` types, but that resulted in multitude bugs, where some unexpected transformation chain would loose some information. 

Therefore, if transformation is purpose-specific, like @matszczygiel mentioned -> we should explicitly signal that. Hence in my opinion purpose-specific helper function is suitable here and makes it less likely for unexpected transformation chains to form accidentally.

mathiaspeters: Small correction: "you cannot implement a trait on the foreign type" is not true, but you cannot implement a *foreign* trait on a foreign type. It doesn't really change the discussion, just wanted to add the correction

matszczygiel: Good then. Leaving this as is, since this is very specific conversion

src/device.rs
```diff
@@ -2312,6 +2308,21 @@ impl Runtime {
             }
         }
     }
+
+    fn is_dublicated_event(&mut self, node: &Node) -> bool {
```
LukasPukenis, line 2312:
```suggestion
    fn is_duplicated_event(&mut self, node: &Node) -> bool {
```

LukasPukenis: Why can't this function be just `*node == *last_node` check? I fear a bit that the implicit `state` checking hides a possible bug since if there was no previous event it means the passed event is guaranteed not to be a duplicate.

src/device.rs
```diff
@@ -2426,6 +2437,38 @@ impl TaskRuntime for Runtime {
...
+                            let _ = self.event_publishers.libtelio_event_publisher.send(
+                                Box::new(Event::Node {body: node.clone()})
+                            );
+                            self.remember_last_transmitted_node_event(node);
+                        }
+                    },
+                    (telio_pq::Event::Disconnected(pubkey), _, Some(last_exit)) if last_exit.public_key == *pubkey  => {
+                        let node = Node {
+                            state: PeerState::Disconnected,
+                            link_state: if self.features.link_detection.is_some() {Some(LinkState::Down)} else {None},
```
LukasPukenis, line 2458:
```suggestion
                            link_state: link_state: self.features.link_detection.map(|_| LinkState::Down),
```

src/device.rs
```diff
@@ -2426,6 +2437,38 @@ impl TaskRuntime for Runtime {
             Some(pq_event) = self.event_listeners.post_quantum_subscriber.recv() => {
                 telio_log_debug!("WG consolidation triggered by PQ event");
 
+                match (&pq_event, &self.requested_state.exit_node, &self.requested_state.last_exit_node) {
+                    (telio_pq::Event::Connecting(pubkey), Some(exit), _) if exit.public_key == *pubkey  => {
+                        let node = Node {
+                            state: PeerState::Connecting,
+                            link_state: if self.features.link_detection.is_some() {Some(LinkState::Down)} else {None},
```
LukasPukenis, line 2444: same as below

LukasPukenis (commented):
I am missing natlab testcase which checks the happy path where peer receives all expected events when correctly connecting.

I see there's `test_pq_vpn_silent_pq_upgrader` however that tests the non-pq server - what about the happy case?

src/device.rs
Jauler: One could do the `node == last_node` check, but that would require hashmap `last_transmitted_event` entries to never be removed. This is obviously undesirable, therefore whenever `Disconnect` event is fired - the `last_transmitted_event` entry is cleared from hashmap. 

This slightly changes the semantics, because event missing from hasmap now can mean one of two things:
* Either there was never an event fired for specified `public_key`
* Or the last event fired for specified `public_key` is Disconnected.

Because `Disconnected` event can never be first event fired for the node -> we can safely assume, that if event has state `disconnected` and there is no entry in hashmap -> the entry is dublicated `Disconnect` event.

src/device.rs
Jauler: I do not see any obvious benefit of using `.map()` in this case, especially when we are ignoring mappped type.

Jauler:
> I am missing natlab testcase which checks the happy path where peer receives all expected events when correctly connecting.
> 
> I see there's `test_pq_vpn_silent_pq_upgrader` however that tests the non-pq server - what about the happy case?

Every test using `_connect_to_vpn_pq` is enforcing that connecting events are coming, and `disconnect_from_vpn()` in telio.py is waiting for `disconnect` event.

src/device.rs
LukasPukenis: Ok that makes sense

LukasPukenis (commented):
+1

src/device.rs
LukasPukenis: It is functionally the same code just more rusty if I got it right.

src/device.rs
mathiaspeters: It should still be renamed `dublicated` -> `duplicated`, but I'm also okay with that happening in the follow-up

src/device.rs
mathiaspeters: I agree with Lukas on this. At least in my opinion, the expected transformation is a conditional transformation for `Option<T>`, not just `T`, so making this a conditional transformation to `Some(_)` is what map is for. Also, `thing.map(|_| whatever)` is IMO clearer than `if thing.is_some() { Some(whatever) } else { None }`

mathiaspeters: Though I don't see a functional issue at all with the current state of it so in the interest of merging quickly, it can be done in the follow up PR instead

mathiaspeters (approved):
+1

A couple of comments I would like to see in the follow-up PR, but this is good to go in my opinion

src/device.rs
Jauler: The followup PR (https://github.com/NordSecurity/libtelio/pull/1021) would remove the need for this kind of event building altogether. Hence this code is planned to be removed as well.

src/device.rs
Jauler: same as above

PR #1008, opened 2024-12-03
Collect tcpdumps from all gateways

tomaszklak:
Previously only gateways which were the default gateways for some node had the tcpdumps collected.


mathiaspeters:
+1

gytsto (approved):
+1

PR #1010, opened 2024-12-04
Fix crash in neli (netlink library)

tomaszklak:
It was failing with:

    range end index 4294836240 out of range for slice of length 1444

It seems we don't need dummy struct to wait for this socket to be ready. It's enough to try to read any bytes from it. In such situation it seems that recv returns each time network interface appears (or disappear).

- [x] README.md is updated


matszczygiel:
+1

PR #1012, opened 2024-12-04
[LLT-5831] QNAP: move cgi logic to teliod

lcruz99:
### Problem
CGI consists in a bash script that handles http requests and directly control teliod daemon execution.
As in the near future more features (eg: authentication) are to be added, after internal discussions it was decided that would be more secure and easy to maintain/implement if the handling of the requests is moved to the teliod binary. 

### Solution
Make the CGI script redirect http requests to teliod.

- [x] README.md is updated


tomaszklak:
Can we use instead something like https://docs.rs/cgi2/latest/cgi/ , so that we avoid implementing irrelevant details of cgi/http, and focus on our business logic?

clis/teliod/src/main.rs
```diff
@@ -44,6 +46,9 @@ enum Cmd {
     Daemon { config_path: String },
     #[clap(flatten)]
     Client(ClientCmd),
+    #[cfg(feature = "qnap")]
+    #[clap(about = "Receive and parse http requests")]
+    Cgi,
```
stalowyjez, line 51: I don't get it tbh - why is it added here?  It seems it communicates with the rest of the teliod code through cli interface, it looks like completely separate tool, to me it seems that there is no reason to keep it together.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+enum Action {
+    Start,
+    Stop,
+    UpdateConfig,
+    GetStatus,
+    GetTeliodLogs,
+    GetMeshnetLogs,
+    Invalid,
+}
+
```
stalowyjez, line 80: What is the reason of creating this intermediate enum? It looks to me that you just make two `match` statements instead of one, without any visible gains.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+fn is_teliod_running() -> bool {
+    if let Ok(pid) = fs::read_to_string(PathConfig::pid()) {
+        if fs::metadata(format!("/proc/{}", pid.trim())).is_ok() {
+            return true;
+        }
+    }
+    false
```
stalowyjez, line 127: It could be just `.map` + `unwrap_or(false)`

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+impl PathConfig {
+    const APP_TMP_DIR: &'static str = "/tmp/nordsecuritymeshnet";
+    const QPKG_DIR: &'static str = "/share/CACHEDEV1_DATA/.qpkg/NordSecurityMeshnet";
+
+    pub fn teliod_bin() -> String {
+        format!("{}/teliod", Self::QPKG_DIR)
+    }
+
+    pub fn teliod_cfg() -> String {
+        format!("{}/teliod.cfg", Self::QPKG_DIR)
```
stalowyjez, line 22: Teliod knowing where is the config file is a bit weird idea imho, in such case I would refactor it to exclude that argument and hardcode it in the rest of the code as well.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+        if matches!(self, HttpMethod::Post | HttpMethod::Patch) {
+            let content_length = env::var("CONTENT_LENGTH")
+                .ok()
+                .and_then(|len| len.parse::<usize>().ok());
+            if let Some(length) = content_length {
+                let _ = io::stdin().take(length as u64).read_to_string(&mut data);
+            }
+        }
+        data
+    }
```
stalowyjez, line 67: This looks like a bad design, the struct doesn't really own these data, it just loads it from the variable, so I see no reason to make it a `get` method on self

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+
+impl HttpMethod {
+    fn from_method(method: &str) -> Self {
+        match method {
+            "GET" => HttpMethod::Get,
+            "POST" => HttpMethod::Post,
+            "PATCH" => HttpMethod::Patch,
+            _ => HttpMethod::Unsupported,
+        }
+    }
```
stalowyjez, line 54: It looks like sth which should be loaded from the external library imho, maybe use `Method` from `http` crate or sth

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+        {
+            Ok(output) => {
+                let status = String::from_utf8_lossy(&output.stdout).to_string();
+                send_response(200, json!({ "status-report": status }));
+            }
+            Err(_) => send_response(500, "Failed to retrieve status."),
+        }
+    } else {
+        send_response(400, "Application is not running.");
+    }
```
stalowyjez, line 224: Maybe it would be better to directly send it to the socket?

stalowyjez (commented):
imho looks strange, like someone wanted to write a shell script, but had to do it in rust for some reason

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+    // {
+    //     Some(new_config) => {
+    //         if fs::write(PathConfig::cfg(), new_config).is_ok() {
+    //             send_response(200, "Configuration updated successfully.");
+    //         } else {
+    //             send_response(500, "Failed to update configuration.");
+    //         }
+    //     }
+    //     None => send_response(400, "Invalid configuration data."),
+    // }
```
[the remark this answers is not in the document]
lcruz99: There's a TODO on the line above:
```
// TODO: LLT-5712
```

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    pub fn meshnet_log() -> String {
+        format!("{}/meshnet.log", Self::QPKG_DIR)
+    }
+
+    pub fn pid() -> String {
+        format!("{}/teliod.pid", Self::APP_TMP_DIR)
+    }
+}
```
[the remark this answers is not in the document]
lcruz99: Just to express that files are relative to `APP_TMP_DIR` and `QPKG_DIR`.

clis/teliod/src/main.rs
lcruz99: you're right it could have been a separate tool. The team has decided the pros didn't outweight the cons and decided to have this design. What's your rationale in splitting it into a different binary?

clis/teliod/src/qnap.rs
lcruz99: It looks like that, but I thought it would be good to have something to expand on in the future.

clis/teliod/src/qnap.rs
lcruz99: can you provide an example?

clis/teliod/src/qnap.rs
lcruz99: sorry I didn't understand, can you explain?

clis/teliod/src/qnap.rs
lcruz99: > the struct doesn't really own these data
The method behaviour depends on the enum type, so it made sense to make it an enum method. If I turn it into a function I'd still check it against Post/Patch

clis/teliod/src/qnap.rs
lcruz99: will do

clis/teliod/src/qnap.rs
lcruz99: ? I didn't understand, can you explain?

lcruz99:
> imho looks strange, like someone wanted to write a shell script, but had to do it in rust for some reason

@stalowyjez that was exactly what happened. What improvements do you suggest?

lcruz99:
> Can we use instead something like https://docs.rs/cgi2/latest/cgi/ or https://docs.rs/cgi/0.6.0/cgi/ , so that we avoid implementing irrelevant details of cgi/http, and focus on our business logic?

@tomaszklak will do

clis/teliod/src/qnap.rs
lcruz99: in the end I'll change it with the `http` crate objects as you suggested below

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
stalowyjez: Maybe you could have a look at https://docs.rs/const_format/latest/const_format/macro.concatcp.html ?

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
stalowyjez: TODO is ok when you have some sound temporary workaround, this looks just like a code during debugging, not sth what should be pushed master at all

clis/teliod/src/main.rs
stalowyjez: Well, it doesn't really share any code, I see no arguments for bundling it together. The cons are quite obvious, you create false impression that these pieces of code are somehow connected, you cannot compile them separately (when compiling for qnap), but most importantly, as function shouldn't do job of two function based on some boolean, a binary shouldn't be, in fact, two separate binaries, glued with some `cgi` flag.

clis/teliod/src/qnap.rs
stalowyjez: Currently teliod daemon starts with config file path provided as an argument - the situation where this path is at the same time hardcoded provided to and hardcoded in teliod doesn't feel right.

clis/teliod/src/qnap.rs
stalowyjez: Soo, currently this methods is teliod running teliod which sends message to another teliod through the socket. I guess you could send this message directly instead.

stalowyjez:
>@stalowyjez that was more or less what happened, but what do you think it looks strange besides other comments?

I mean it just doesn't look like a right tool for the job to me

clis/teliod/src/qnap.rs
stalowyjez: Sth like this, I guess:
```fn is_teliod_running() -> bool {
    fs::read_to_string(PathConfig::pid())
        .and_then(|pid| fs::metadata(format!("/proc/{}", pid.trim())))
        .is_ok()
}
```

But tbh we agreed to use existance of the socket instead: https://github.com/NordSecurity/libtelio/blob/c60eedeeeca8e5f0928bc04ea29c9f720806cdff/clis/teliod/src/main.rs#L96 soo, maybe it should be made consistent?

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: I'll delete this commented code.

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: fair enough, will change.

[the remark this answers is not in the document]
lcruz99: fair enough, will do

clis/teliod/src/main.rs
lcruz99: I think you can use the same reasoning for the teliod as a daemon or client. It was decided to keep them under the same binary, and the same decision was made for qnap's cgi.

> you cannot compile them separately

Why would you insist on having them separated? There's this CGI feature that is only used for qnap usecase and therefore excluded for other platform builds.

> as function shouldn't do job of two function based on some boolean, a binary shouldn't be, in fact, two separate binaries, glued with some cgi flag.

What is the job of two functions? We call qnap::parse_cgi() to handle the http request, that's one usecase only. And this isn't a function that's enabled by a boolean, it's included or excluded from the binary depending on the flag you pass to the compiler at compile-time so there's no code/symbols on binaries for other platforms. That's what cargo features are about, it's literally used by almost all crates. 🤔

clis/teliod/src/qnap.rs
lcruz99: Other platforms certainly have a different path for the config file whereas for QNAP this is a static path, hence the idea of having it hardcoded.

clis/teliod/src/qnap.rs
lcruz99: thanks for the example, I'll use it.

> But tbh we agreed to use existance of the socket instead: 

I don't see the benefits of using it, we still need to keep track of the PID to be able to stop the application. Maybe that suits the checks before socket communications but not here I think.

clis/teliod/src/main.rs
stalowyjez: >I think you can use the same reasoning for the teliod as a daemon or client

TBH if I would start from scratch, I probably would make them separate, but they actually share some code and communicate in some reasonable manner, while this new part treats the rest of teliod like a blackbox

>What is the job of two functions? We call qnap::parse_cgi() to handle the http request, that's one usecase only. And this isn't a function that's enabled by a boolean, it's included or excluded from the binary depending on the flag you pass to the compiler at compile-time so there's no code/symbols on binaries for other platforms. That's what cargo features are about, it's literally used by almost all crates. 🤔

I'm not talking about compilation flag, but about `cgi` parameter - you pass it and run some program which is completely disjoint with the rest. I'm not insisting on separating them, but I think there are two reasonable options:
 1. Integrate it, so use the socket, use consistent checks of the stuff, errors, etc.
 2. Keep it as a separate script like now, but in such case bundling it together has no benefits imho

clis/teliod/src/main.rs
```diff
@@ -134,5 +139,10 @@ async fn main() -> Result<(), TeliodError> {
                 Err(TeliodError::DaemonIsNotRunning)
             }
         }
+        #[cfg(feature = "qnap")]
+        Cmd::Cgi => {
```
packgron, line 143: So now we have this quite specialized for qnap, so i think it would make sense to also call it a bit more concretely: like telio `qnap-cgi` or leave this as generic cgi, with qnap auth as also a seperate sub part

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+};
+
+use serde_json::json;
+
+use crate::config::TeliodDaemonConfig;
+
+struct PathConfig;
+
+impl PathConfig {
+    const APP_TMP_DIR: &'static str = "/tmp/nordsecuritymeshnet";
```
packgron, line 14: NIT: Would be nice to have them configurable via parameters, but leave default's as is.

clis/teliod/src/main.rs
packgron: Yeh this came down via vote, and all cases are equivalent.

But I would be for not calling `teliod client` from `teliod cgi` but actualy reasuing socket logic to interact with deamon directly, like teliod client.

Metal model being that there are two client impl's for teliod daemon:
- cli interface
- web-cgi interface

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
packgron: In thease cases in built `concat!` would be enough

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+            "action=update-config" => Action::UpdateConfig,
+            "action=get-status" => Action::GetStatus,
+            "action=get-teliod-logs" => Action::GetTeliodLogs,
+            "action=get-meshnet-logs" => Action::GetMeshnetLogs,
+            _ => Action::Invalid,
+        }
+    }
+}
+
+pub(crate) fn parse_cgi() {
```
packgron, line 95: It would be nice to have cgi api described somthere in details, also maybe we should make it a bit more REST'full ?

So:
Start daemon:
```
POST /
{ ... config ... }
```

Reconfigure and restart the daemon
```
PATCH /
{ ... config ... }
```

Stop the deamon
```
DELETE /
```

And then only have query params for GET's
So let's for now leave `GET /` for possibly getting the index.html in authed way

`GET /?info=status` for mesh status

and
`GET /?info=log` for log

qnap/shared/teliod.cgi
```diff
@@ -1,131 +1,7 @@
-#!/bin/bash
+#!/bin/sh
```
packgron, line 1: add: `set -eu`

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: `concat!` only supports literals, `concatcp!` as @stalowyjez works

clis/teliod/src/qnap.rs
ghztomash: On the other had `cgi` would still have to call the teliod binary to start the daemon.

clis/teliod/src/qnap.rs
ghztomash: There's also the [hyper](https://crates.io/crates/hyper) crate for http/rest stuff, maybe it has something useful

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub fn teliod_bin() -> String {
+        format!("{}/teliod", Self::QPKG_DIR)
+    }
+
+    pub fn teliod_cfg() -> String {
+        format!("{}/teliod.cfg", Self::QPKG_DIR)
+    }
+
+    pub fn teliod_log() -> String {
+        "/var/log/teliod.log".to_owned()
```
ghztomash, line 26: Shouldn't these be retrieved from the config?

qnap/shared/teliod.cgi
lcruz99: added

clis/teliod/src/qnap.rs
lcruz99: Reused daemon socket where aplicable. As @ghztomash mentioned for daemon start we still have to run `teliod` as a command.

clis/teliod/src/qnap.rs
lcruz99: used external `http` crate from `rust-cgi`.

clis/teliod/src/main.rs
lcruz99: Added socket logic instead of issuing the bash command.

clis/teliod/src/main.rs
lcruz99: changed

qnap/shared/teliod.cgi
packgron: Not added to this script

clis/teliod/src/qnap.rs
lcruz99: It's one possibility. Do you think other file paths should also be set on the config? or just `teliod.log`?

clis/teliod/src/qnap.rs
lcruz99: qweqwe

clis/teliod/QNAP.md
```diff
@@ -0,0 +1,127 @@
...
+
+This REST API allows interaction with the Teliod daemon. It provides endpoints for managing the daemon, updating its configuration, and retrieving logs and status information.
+
+### Endpoints
+
+#### 1. **Start the Daemon**
+- **Endpoint**: `/`
+- **Method**: `POST`
+- **Description**: Starts the Teliod Daemon in the background.
+- **Request Body**: None
```
packgron, line 23: Why are we not sending configuration together with start?

clis/teliod/QNAP.md
```diff
@@ -0,0 +1,127 @@
...
+- **Responses**:
+  - **200 OK**: Daemon started successfully.
```
packgron, line 25: Should be 201 Created

clis/teliod/QNAP.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### 5. **Get Meshnet Logs**
+- **Endpoint**: `/?info=get-meshnet-logs`
+- **Method**: `GET`
+- **Description**: Retrieves the latest logs of the Meshnet.
+- **Request Body**: None
+- **Responses**:
+  - **200 OK**: Log content in JSON format.
+    ```json
+    {
```
packgron, line 81: This is not valid json, just return text/plain with full log file

clis/teliod/QNAP.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### 6. **Get Teliod Logs**
+- **Endpoint**: `/?info=get-teliod-logs`
+- **Method**: `GET`
+- **Description**: Retrieves the latest logs of the Teliod Daemon.
+- **Request Body**: None
+- **Responses**:
+  - **200 OK**: Log content in JSON format.
+    ```json
+    {
```
packgron, line 95: Just return text/plain, with a log file

packgron: What is difference between teliod and meshnet logs?

clis/teliod/QNAP.md
```diff
@@ -0,0 +1,127 @@
...
+    "log_level": "info",
+    "log_file_path": "/new/path/to/log.log"
+  }
+  ```
+  - **Responses**:
+  - **200 OK**: Configuration updated successfully
+  - **400 Bad Request**: Invalid JSON payload or configuration fields.
+  - **500 Internal Server Error**: Failed to update configuration.
+
+#### 4. **Get Meshnet Status**
```
packgron, line 56: For ui, i will probably will also need get-config, as persisting it in browser is probably not a good idea. It should return current config of teliod

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+        (&Method::POST, _) => start_daemon(),
+        (&Method::DELETE, _) => stop_daemon(),
+        (&Method::PATCH, _) => {
+            let body = match String::from_utf8(request.into_body()) {
+                Ok(body) => body,
+                Err(_) => return text_response(400, "Invalid UTF-8 in request body."),
+            };
+            update_config(&body)
+        }
+        (&Method::GET, Some("info=get-status")) => get_status(),
```
packgron, line 42: Will you add proper query parsing on next PR ?

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+            teliod_log_file
+                .try_clone()
+                .expect("Failed to open teliod log file."),
+        ))
+        .stderr(Stdio::from(teliod_log_file))
+        .spawn()
+    {
+        Ok(child) => {
+            let pid = child.id();
+            if fs::write(PID_FILE, pid.to_string()).is_ok() {
```
tomaszklak, line 109: This should be done by the deamon process, not by the cgi process.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
+use std::{
+    fs, io,
+    path::PathBuf,
+    process::{Command, Stdio},
+    time::Duration,
+};
+
+use const_format::concatcp;
+use rust_cgi::{http::Method, text_response, Request, Response};
```
tomaszklak, line 9: There are many places in this file where you call `text_response` with a numeric value, please convert them to https://docs.rs/http/1.0.0/http/status/struct.StatusCode.html

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+
+    let update_values: serde_json::Value = match serde_json::from_str(body) {
+        Ok(updates) => updates,
+        Err(e) => {
+            eprintln!("Error parsing config json: {}", e);
+            return text_response(400, "Invalid JSON payload");
+        }
+    };
+
+    if let serde_json::Value::Object(update_map) = update_values {
```
tomaszklak, line 150: Please define a type that represents valid body and use (de)serialisation derived for it instead of manual json value traversal.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+            if let Ok(Ok(daemon_reply)) = socket_query {
+                if let Ok(CommandResponse::StatusReport(status)) =
+                    CommandResponse::deserialize(&daemon_reply)
+                {
+                    return text_response(
+                        200,
+                        serde_json::to_string_pretty(&status).unwrap_or_default(),
+                    );
+                }
+            }
```
tomaszklak, line 227: Would be nice to know that some errors happened - please add some sort of logging or return response with a useful string that includes the error.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+        }
+    }
+
+    text_response(500, "Failed to retrieve status.")
+}
+
+fn get_teliod_logs() -> Response {
+    match fs::read_to_string(TELIOD_LOG) {
+        Ok(logs) => text_response(200, logs),
+        Err(_) => text_response(500, "Log file not found."),
```
tomaszklak, line 237: Why drop the error? It might be useful.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+}
+
+fn get_meshnet_logs() -> Response {
+    match fs::read_to_string(MESHNET_LOG) {
```
tomaszklak, line 242: Why drop the error? It might be useful.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+            mqtt_default_cfg.backoff_maximal
+        );
+        assert_eq!(
+            updated_config.mqtt.reconnect_after_expiry,
+            mqtt_default_cfg.reconnect_after_expiry
+        );
+        assert_eq!(
+            updated_config.mqtt.certificate_file_path,
+            mqtt_default_cfg.certificate_file_path
+        );
```
tomaszklak, line 329: nit: I think it would be more readable if you would create the expected struct and compare it against the `updated_config`, instead of manually doing it field by field.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+                "config_provider": "manual"
+            },
+            "http_certificate_file_path": "/http/certificate/path/"
+        }
+        "#;
+        fs::write(TELIOD_CFG, initial_config).unwrap();
+
+        let update_body = r#"
+        {
+            "authentication_token": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
```
tomaszklak, line 351: Please add tests for all the fields.

tomaszklak (changes requested):

clis/teliod/QNAP.md
lcruz99: configuration is being read from file. User should pass configuration values if he wish so?

clis/teliod/QNAP.md
lcruz99: changed

clis/teliod/QNAP.md
packgron: It probably would be nice, but will see, maybe I will change this a bit then implementing ui

clis/teliod/QNAP.md
lcruz99: fixed

clis/teliod/QNAP.md
lcruz99: teliod logs are app logs (eg: Wg task started, Stacktrace errors, etc.), meshnet logs are logs from libtelio

clis/teliod/src/qnap.rs
lcruz99: Nice! changed

clis/teliod/src/config.rs
```diff
@@ -138,7 +161,20 @@ fn deserialize_authentication_token<'de, D: Deserializer<'de>>(
     }
 }
 
-#[derive(PartialEq, Eq, Deserialize, Debug)]
+fn serialize_authentication_token<S>(auth_token: &str, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    if auth_token.len() == 64 && auth_token.chars().all(|c| c.is_ascii_hexdigit()) {
```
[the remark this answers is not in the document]
lcruz99: In this case I still have to check ArrayString length, the difference is we'll have fixed capacity, or am I missing something?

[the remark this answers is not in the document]
lcruz99: checks that it shouldn't exceed the capacacity limits. Shouldn't we assure it's 0 xor 64, as done [here](https://github.com/NordSecurity/libtelio/blob/LLT-5831_move_cgi_logic_to_teliod/clis/teliod/src/config.rs#L156)?

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+use tests::TELIOD_CFG;
+
+pub(crate) fn handle_request(request: Request) -> Response {
+    match (request.method(), request.uri().query()) {
+        (&Method::POST, _) => start_daemon(),
+        (&Method::DELETE, _) => stop_daemon(),
+        (&Method::PATCH, _) => {
+            let body = match String::from_utf8(request.into_body()) {
+                Ok(body) => body,
+                Err(_) => return text_response(400, "Invalid UTF-8 in request body."),
```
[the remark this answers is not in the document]
lcruz99: changed

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+const TELIOD_CFG: &str = concatcp!(QPKG_DIR, "/teliod.cfg");
+#[cfg(test)]
+use tests::TELIOD_CFG;
+
+pub(crate) fn handle_request(request: Request) -> Response {
+    match (request.method(), request.uri().query()) {
+        (&Method::POST, _) => start_daemon(),
+        (&Method::DELETE, _) => stop_daemon(),
+        (&Method::PATCH, _) => {
+            let body = match String::from_utf8(request.into_body()) {
```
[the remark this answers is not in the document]
lcruz99: changed

clis/teliod/Cargo.toml
```diff
@@ -34,6 +34,11 @@ anyhow.workspace = true
 smart-default = "0.7.1"
 base64 = "0.22.1"
 dirs = "4.0.0"
+const_format = "0.2.33"
+rust-cgi = "0.7.1"
```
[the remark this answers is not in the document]
lcruz99: fixed

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+#[cfg(not(test))]
+const TELIOD_CFG: &str = concatcp!(QPKG_DIR, "/teliod.cfg");
+#[cfg(test)]
+use tests::TELIOD_CFG;
+
+pub(crate) fn handle_request(request: Request) -> Response {
+    match (request.method(), request.uri().query()) {
+        (&Method::POST, _) => start_daemon(),
+        (&Method::DELETE, _) => stop_daemon(),
+        (&Method::PATCH, _) => {
```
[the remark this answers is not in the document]
lcruz99: verification of path?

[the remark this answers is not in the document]
tomaszklak: For example in the documentation we say:
```
- **Endpoint**: `/`
- **Method**: `POST`
```
But in code we start daemon after any post, regardless of the path. We should verify that the path is exacly `/`. Similar for `DELETE` and `PATCH` methods.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+        (&Method::GET, Some("info=get-teliod-logs")) => get_teliod_logs(),
+        (&Method::GET, Some("info=get-meshnet-logs")) => get_meshnet_logs(),
+        (_, _) => text_response(400, "Invalid request."),
+    }
+}
+
+fn is_teliod_running() -> bool {
+    fs::read_to_string(PID_FILE)
+        .and_then(|pid| fs::metadata(format!("/proc/{}", pid.trim())))
+        .is_ok()
```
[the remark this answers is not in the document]
lcruz99: that sounds simple and effective, thanks!

clis/teliod/src/qnap.rs
lcruz99: asdasd

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: the Endpoint path `/` is relative to the teliod.cgi path. So `/` is equivalent to `/cgi-bin/qpkg/teliod.cgi`. I can check against this hardcoded path..

clis/teliod/src/qnap.rs
lcruz99: added more specific error handling

clis/teliod/src/qnap.rs
lcruz99: added error log

clis/teliod/src/qnap.rs
lcruz99: added error log

clis/teliod/src/qnap.rs
lcruz99: Agreed, changed

clis/teliod/src/qnap.rs
lcruz99: wdym? Add tests for each field partial update?

clis/teliod/src/qnap.rs
lcruz99: This is a QNAP specific file, I can do it conditionally when the `qnap` feature is enabled. wdyt?

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
tomaszklak: Is user able to control the path parameter?

clis/teliod/src/qnap.rs
tomaszklak: The important part is that the whole production code is covered - whether it's in one test or many is less important and I will leave this up to you to judge which is better in this case.

clis/teliod/src/qnap.rs
tomaszklak: Ok :)

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: technically yes, but if they change the path they won't be running `teliod.cgi` 🤔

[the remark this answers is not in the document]
tomaszklak: So what is the point of checking path few lines below in:
```rust
(&Method::GET, Some("info=get-status"))
```
?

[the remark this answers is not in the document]
lcruz99: That value is returned from `uri().query()` whereas the path is retrieved from `uri().path()`:

see [rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2fc86337eca752f3f5dd20b803789f40)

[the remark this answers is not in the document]
tomaszklak: If we expect that it will always be the same value (`/cgi-bin/qpkg/teliod.cgi`?) than there should be no reason not to check - if we are mistaken we want to know about this.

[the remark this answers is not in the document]
lcruz99: fair enough, added

clis/teliod/src/qnap.rs
lcruz99: added

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+fn teliod_socket_exists() -> bool {
+    DaemonSocket::get_ipc_socket_path()
+        .and_then(|path| path.metadata())
+        .is_ok()
+}
+
+fn kill_teliod_process() -> Result<(), io::Error> {
+    match fs::read_to_string(PID_FILE) {
+        Ok(pid) => {
+            let _ = Command::new("kill").arg("-9").arg(pid.trim()).status();
```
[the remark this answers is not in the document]
lcruz99: added

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+            if teliod_socket_exists() {
+                let _ = DaemonSocket::get_ipc_socket_path().and_then(fs::remove_file);
```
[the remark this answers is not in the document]
lcruz99: This can occur if the daemon process has crashed for some reason and the socket fd wasn't cleaned

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: Referring to the above comment, this doesn't guarantee that the process being killed is effectively the telio daemon process right?

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,398 @@
...
+            }
+            fs::remove_file(PID_FILE)
```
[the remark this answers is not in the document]
lcruz99: This can occur if the daemon process has crashed for some reason and the socket fd wasn't cleaned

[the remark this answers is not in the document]
lcruz99: If the daemon failed to cleanup, the `PID_FILE` or the socket might still exist

clis/teliod/src/qnap.rs
lcruz99: yes, I'll leave it for the next PR

clis/teliod/QNAP.md
```diff
@@ -0,0 +1,127 @@
...
+### Example usage with curl
+
+#### Start Teliod daemon:
+```bash
+curl -X POST http://<NAS-IP>:8080/
+```
+
+#### Stop Teliod daemon:
+```bash
+curl -X POST http://<NAS-IP>:8080/
```
packgron, line 116: -X DELETE?

clis/teliod/Cargo.toml
```diff
@@ -34,6 +34,13 @@ anyhow.workspace = true
 smart-default = "0.7.1"
 base64 = "0.22.1"
 dirs = "4.0.0"
+const_format = { version = "0.2.33", optional = true }
+rust-cgi = { version = "0.7.1", optional = true }
+arrayvec = "0.7.6"
```
tomaszklak: This doesn't seem to be used anywhere - please remove it.

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
tomaszklak: Ugh.... I meant `query` not `path` 🤦

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
tomaszklak: Yeah, I'm thinking that maybe we shouldn't be calling `kill` at all - it might turn out it's possible somehow to trick our cgi handler to kill arbitrary processes on qnap.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,458 @@
...
+    }
+}
+
+fn stop_daemon() -> Response {
+    if shutdown_teliod().is_ok() {
+        text_response(StatusCode::OK, "Application stopped successfully.")
+    } else {
+        match kill_teliod_process() {
+            Ok(_) => text_response(StatusCode::OK, "Application stopped successfully."),
+            _ => text_response(StatusCode::GONE, "Application not found."),
```
tomaszklak, line 185: Might be good to log the error somewhere, or include it in the response.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,458 @@
...
+        ),
+    }
+}
+
+fn stop_daemon() -> Response {
+    if shutdown_teliod().is_ok() {
+        text_response(StatusCode::OK, "Application stopped successfully.")
+    } else {
+        match kill_teliod_process() {
+            Ok(_) => text_response(StatusCode::OK, "Application stopped successfully."),
```
tomaszklak, line 184: I think it would be best not to use the same exact message here and in the ok case.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,458 @@
...
+            StatusCode::INTERNAL_SERVER_ERROR,
+            format!("Failed to start the application: {error}"),
+        ),
+    }
+}
+
+fn stop_daemon() -> Response {
+    if shutdown_teliod().is_ok() {
+        text_response(StatusCode::OK, "Application stopped successfully.")
+    } else {
```
tomaszklak, line 182: Let's not ignore this error and include it in below messages or log it somehow.

clis/teliod/QNAP.md
lcruz99: Missed that, updated

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    pub const TELIOD_CFG: &str = "/tmp/teliod_config.json";
+
+    lazy_static! {
+        static ref TEST_SEQ_MUTEX: Mutex<()> = Mutex::new(());
+    }
+
+    #[test]
+    fn test_update_config() {
+        let _lock = TEST_SEQ_MUTEX.lock().unwrap();
```
tomaszklak, line 296: This can be replaced with `serial_test`, which we already use across other crates in libtelio.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,458 @@
...
+    use tracing::level_filters::LevelFilter;
+    use uuid::Uuid;
+
+    use super::{update_config, TeliodDaemonConfig};
+    use crate::{
+        config::{InterfaceConfig, MqttConfig, Percentage},
+        configure_interface::InterfaceConfigurationProvider,
+    };
+
+    pub const TELIOD_CFG: &str = "/tmp/teliod_config.json";
```
tomaszklak, line 288: I think this will fail on windows - can you change it to use https://crates.io/crates/tempfile ?

clis/teliod/Cargo.toml
lcruz99: removed

clis/teliod/src/qnap.rs
packgron: I think this is out of scope for now, we can make this as an issue, to make cgi standalone and cross platform

tomaszklak: This is not about making cgi cross platform - it's about making unit tests pass for developer(s?) on windows.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,458 @@
...
+            DaemonSocket::send_command(
+                &socket_path.unwrap_or_default(),
+                &serde_json::to_string(&$command).unwrap_or_default(),
+            ),
+        ))
+    }};
+}
+
+pub(crate) fn handle_request(request: Request) -> Response {
+    if request.uri().path() != "/cgi-bin/qpkg/teliod.cgi" {
```
packgron, line 66: Is this really correct ?

Path should be string after "/cgi-bin/qpkg/teliod.cgi'

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
packgron: Hey let's not do path verification in this PR, let's leave it after we have a web page logic
also path() should only contain string afeter cgi-bin/qpkg/teliod.cgi,soforexample if cgi-bin/qpkg/teliod.cgi/xxx/x would be request, path should be /xxx/x.

In my setup i will rely that '/' path is return index page, and '/style.css' will return css and so on, so let's leave path out of this pr.

Query tipycaly should not be fragile, we disccuseed with luis already, but proper parsing should be done so for example:
?some-id=1&some-other=2 should be equivalent to ?some-other=2&some-id=1
We will add this with auth logic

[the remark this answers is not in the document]
lcruz99: will remove path verification. Query will indeed be validated on a next PR

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: Removed `kill` process measure

clis/teliod/src/qnap.rs
lcruz99: Added shutdown error log

clis/teliod/src/qnap.rs
lcruz99: done

clis/teliod/src/qnap.rs
lcruz99: How would windows devs run these unit tests?

clis/teliod/src/qnap.rs
lcruz99: path verification is removed

clis/teliod/src/qnap.rs
tomaszklak: Hm, you right - there is no configurability in this module and all paths are completely hardcoded. It's probably relatively cheap to make paths be configurable with a prefix (which would be empty in production) and set it to temporary dir in tests. But fell free to treat this thread as nitpick and ignore it if you want.

clis/teliod/src/qnap.rs
lcruz99: `kill_teliod_process` function removed

clis/teliod/src/qnap.rs
lcruz99: `kill_teliod_process` function removed

clis/teliod/src/command_listener.rs
```diff
@@ -76,6 +76,19 @@ impl CommandListener {
                 })
                 .await
             }
+            ClientCmd::QuitDaemon =>
```
Jauler, line 79: IMHO it would be nice to log commands executed (e.g. `debug!()` or maybe `info!()`). This helps puts things into perspective when debugging various issues.

Obviously - we should log all commands not just `QuitDaemon`. Although it does not have to be this PR, it may be separate one

clis/teliod/src/configure_interface.rs
```diff
@@ -24,9 +24,10 @@ fn execute(command: &mut Command) -> Result<(), TeliodError> {
...
 }
 
-#[derive(Debug, Deserialize, PartialEq, Eq)]
+#[derive(Default, Debug, Deserialize, Serialize, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
 pub enum InterfaceConfigurationProvider {
+    #[default]
     Manual,
     Ifconfig,
     Iproute,
```
Jauler, line 33: This is just an idea, so you may resolve this message once you have read it.

Eventually it would be really nice to use netlink for network configs instead of spawning shell commands :).

clis/teliod/src/qnap.rs
Jauler, line 1: To make sure that feature `qnap` does not accidentally leak out of the feature, would it make sense to mark all of this file as `#[cfg(feature = "qnap")]`?

clis/teliod/src/main.rs
```diff
@@ -96,13 +105,21 @@ async fn main() -> Result<(), TeliodError> {
             if DaemonSocket::get_ipc_socket_path()?.exists() {
                 Err(TeliodError::DaemonIsRunning)
             } else {
+                #[cfg(feature = "qnap")]
+                let _pid_guard = qnap::PidFile::new().await?;
```
Jauler, line 109: I wonder do we have a need for pidfiles?

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,431 @@
...
+    let stderr = match teliod_log_file.try_clone() {
+        Ok(file) => Stdio::from(file),
+        Err(error) => {
+            return text_response(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                format!("Failed to start the application: {error}"),
+            )
+        }
+    };
+    match Command::new("setsid")
```
stalowyjez, line 140: imho it would be nicer to fork than using cli

clis/teliod/src/main.rs
```diff
@@ -96,13 +105,21 @@ async fn main() -> Result<(), TeliodError> {
...
                 let file = File::open(&config_path)?;
+
                 let mut config: TeliodDaemonConfig = serde_json::from_reader(file)?;
-                let token = std::env::var("NORD_TOKEN").ok();
-                if let Some(t) = token {
+
+                if let Ok(token) = std::env::var("NORD_TOKEN") {
                     debug!("Overriding token from env");
-                    config.authentication_token = t;
+                    if token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit()) {
```
Jauler, line 116: It would be nice to make AuthorizationToken type which would enforce these environments. e.g. something on the lines like:

```rust
struct AuthorizationToken {
...
};

impl AuthorizationToken {
    pub fn new(String token) -> Self {...}
}
```

And then every place, which needs access to AuthorizationToken would use `AuthorizationToken` struct instead of String. The benefit of such approach is that we can be sure (in a type safe manner) that invariants (e.g. length and symbol requirements) are always enforced for such type, and thus it is significantly harder to miss validation.

Jauler: I had pretty much the same thought here: https://github.com/NordSecurity/libtelio/pull/1012/files#r1894030576

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,431 @@
...
+    match fs::read_to_string(TELIOD_LOG) {
+        Ok(logs) => text_response(StatusCode::OK, logs),
+        Err(error) => text_response(
+            StatusCode::BAD_GATEWAY,
+            format!("Error reading teliod log file: {}", error),
+        ),
+    }
+}
+
+fn get_meshnet_logs() -> Response {
```
Jauler, line 242: Is there a good reason to separate `teliod` and `meshnet` logs? maybe it would be worth to interleave them, like apps do? In that case it is slightly easier to get them from users, as there is single file instead of two.

clis/teliod/src/command_listener.rs
lcruz99: Every command received by the daemon is logged [here](https://github.com/NordSecurity/libtelio/pull/1012/files#diff-c053616e6c57a08149244bd73aba33e23957c3eb3174c20b8ee7a02d2c95a593L254).

clis/teliod/src/configure_interface.rs
lcruz99: I think is a good idea but out of scope of this PR :)

clis/teliod/src/config.rs
```diff
@@ -138,12 +188,73 @@ fn deserialize_authentication_token<'de, D: Deserializer<'de>>(
...
+                &["error", "warn", "info", "debug", "trace", "off"],
+            )
+        }),
+        _ => Ok(None),
+    }
+}
+
+fn deserialize_partial_authentication_token<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<Option<String>, D::Error> {
```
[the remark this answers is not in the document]
lcruz99: That was my initial thought, I've changed my mind though when I realized the effort and started to go out-of-scope of the PR. This is definitely a TODO for the next PRs.

clis/teliod/src/qnap.rs
lcruz99: Didn't fully understand what you mean with "leak out of the future". Anyways, this tree structure will change in the next PR so I think is trivial

clis/teliod/src/main.rs
lcruz99: That's the pattern used for QNAP applications, for start/stop routines (eg: [qnap/shared/NordSecMeshnet.sh](https://github.com/NordSecurity/libtelio/blob/main/qnap/shared/NordSecMeshnet.sh)).

good point, initially we were killing the process from the cgi script, which was in the meantime removed and we don't need the Pidfile now.. so will remove Pidfile from the app side.

clis/teliod/src/main.rs
lcruz99: Agreed! However this will be part of the next PR.

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+}
+
+impl Drop for PidFile {
+    fn drop(&mut self) {
+        if let Err(e) = std::fs::remove_file(PID_FILE) {
+            warn!("Failed to remove PID file: {}", e);
+        }
+    }
+}
```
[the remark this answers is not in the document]
lcruz99: @Jauler also questioned this. It's a good catch since we no longer manage the app on a process level.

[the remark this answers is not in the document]
lcruz99: will remove

clis/teliod/src/qnap.rs
```diff
@@ -0,0 +1,431 @@
...
+    config::{TeliodDaemonConfig, TeliodDaemonConfigPartial},
+    ClientCmd, DaemonSocket, TeliodError, TIMEOUT_SEC,
+};
+
+const TELIOD_TMP_DIR: &str = "/tmp/nordsecuritymeshnet";
+const PID_FILE: &str = concatcp!(TELIOD_TMP_DIR, "/teliod.pid");
+const QPKG_DIR: &str = "/share/CACHEDEV1_DATA/.qpkg/NordSecurityMeshnet";
+const TELIOD_BIN: &str = concatcp!(QPKG_DIR, "/teliod");
+const MESHNET_LOG: &str = concatcp!(QPKG_DIR, "/meshnet.log");
+const TELIOD_LOG: &str = "/var/log/teliod.log";
```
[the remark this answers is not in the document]
lcruz99: Hm.. what do you mean with "teliod logs with meshnet logger"?

clis/teliod/src/qnap.rs
lcruz99: `fork()` inherits the parent session id which makes the process quit after the parent session ends. Instead `setsid` creates a new session id for the child process.

https://man7.org/linux/man-pages/man2/setsid.2.html

clis/teliod/src/qnap.rs
lcruz99: In my POV I'd prefer to keep them separated simply because meshnet logs will bury `teliod` logs which also convey a different category of information.

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
stalowyjez: I mean it uses tracing logs as Telio, so all of them go to the same file. Here there is the initialization for both meshnet and teliod: https://github.com/NordSecurity/libtelio/blob/26fe43ad2da750dd73d2159e4cb1d7f8ffd1a925/clis/teliod/src/daemon.rs#L184
it doesn't really use stdout for logging (or it shouldn't at least)

clis/teliod/src/qnap.rs
stalowyjez: I meant sth like we had in Tcli, it worked well

clis/teliod/src/qnap.rs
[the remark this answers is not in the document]
lcruz99: We have meshnet logs (from libtelio) which use the tracing crate and then we have teliod logs which log the stdout/stderr of the teliod process.

The link you shared has tracing object initlization, where is teliod?

clis/teliod/src/qnap.rs
lcruz99: Can you be more specific pls?

clis/teliod/src/qnap.rs
Jauler: ok

tomaszklak (approved):
+1

clis/teliod/src/qnap.rs
Jauler: I was just thinking, that maybe it is worth skipping `qnap.rs` compilation if `qnap` feature was not enabled.

Jauler (approved):
+1.0

clis/teliod/src/qnap.rs
lcruz99: this is already being done here: https://github.com/NordSecurity/libtelio/blob/39fd537855e80e74b082d691eb113c07d799c0b5/clis/teliod/src/main.rs#L22

clis/teliod/src/qnap.rs
lcruz99: To be continued on a next PR

PR #1014, opened 2024-12-05
Automate day to day rust development tasks with Just(file)

tomaszklak:
### Problem
There are many shorter and longer commands that we (sometimes) want to also run locally. Most of them are long enough that I don't remember them and have to pull them out of shell history or even worse from CI config.

### Solution
Use [just](https://just.systems/) to automate this. `Just` is a recent rewrite/replacement for make. You can run it like that:

```
$ just # will run the default 'test' / 'cargo test...' command
$ just clippy # will run the clippy target
$ just d # will run the 'deny' target' (cargo deny check) using the 'd' alias
```

Second commits replaces our CI jobs which started those various commands directly, with `just` invocations so that we can easily keep them in sync with what we run locally.

- [x] README.md is updated

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -28,30 +28,26 @@ jobs:
       runs-on: ubuntu-22.04
       steps:
         - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
-        - uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # v1.0.6
+        - uses: extractions/setup-just@dd310ad5a97d8e7b41793f8ef055398d51ad4de6 # v2.0.0
           with:
-            components: clippy
-        - uses: actions-rs/cargo@ae10961054e4aa8b4aa7dffede299aaf087aa33b # v1.0.3
-          with:
-            command: clippy
-            args: --target "x86_64-unknown-linux-gnu" --lib --color always -- --deny warnings --allow unknown-lints -W clippy::expect_used -W clippy::panic -W clippy::unwrap_used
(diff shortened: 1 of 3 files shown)


Hasan6979:
Great work! 🚀 

Could you also add running the firewall benchmarks command in the Just file ?

mathiaspeters:
This is a great initiative! 

I think we should be mindful about the aliases so they don't get used in checked-in code, maybe we can add a comment where the aliases are defined to only use them for local usage?

tomaszklak:
> Could you also add running the firewall benchmarks command in the Just file ?

That's a good idea, but in this first PR all I want to do is automate the most common tasks related to pushing code for review. In future PR any team member will be able to easily add more and we will (might) slowly convert everything. On the other hand we might also find that we don't want to use `just` so converting to much before we have a chance to use it is not the best idea.

mathiaspeters:
+1

Hasan6979 (approved):
+1

.github/workflows/linters.yml
```diff
@@ -28,30 +28,26 @@ jobs:
       runs-on: ubuntu-22.04
       steps:
         - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
-        - uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # v1.0.6
-          with:
-            components: clippy
-        - uses: actions-rs/cargo@ae10961054e4aa8b4aa7dffede299aaf087aa33b # v1.0.3
+        - uses: extractions/setup-just@v2
```
jjanowsk, line 31: Please sha-pin

jjanowsk (approved):

PR #1021, opened 2024-12-11
[LLT-5814] refactor events reporting from libtelio

Jauler:
### Problem
Event reporting in libtelio is rather scattered around, and with introduction of PQ now we need to take a look at the state of multiple modules before we can decide on what events should be fired. For this reason some refactoring is required, and this PR attempts to perform it.

### Solution
General convention of events is that they should be fired whenever any field in the reported state changes. This allows for, IMHO, rather fool-proof aproach of having some module responsible for event reporting, which gets notified whenever something changes. This module takes a look at current state of affairs by querying relevant modules, builds up necessary events, and if any field is different -> fires an event.


Jauler:
close as irrelevenat

PR #1023, opened 2024-12-12
LLT-5866: Remove app_user_uid from the config.

tomasz-grz:
### Problem
Currently `app_user_id` used for notification center is part of the user provided config for teliod. It shouldn’t be user configurable, we can reuse the the meshnet `hardware_id` for it instead which should be generated when the device is registered.

### Solution
Remove `app_user_uid` from the user config.
Use a randomly generated UUID instead of the public key for `hw_identifier`.
Use the `hw_identifier` as `app_user_uid` in notification center.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -71,22 +71,19 @@ fn build_backoff() -> Result<ExponentialBackoff, Error> {
     Ok(ExponentialBackoff::new(backoff_bounds)?)
 }
 
-pub async fn init_with_api(
-    auth_token: &str,
-    interface_name: &str,
-) -> Result<DeviceIdentity, Error> {
+pub async fn init_with_api(auth_token: &str) -> Result<DeviceIdentity, Error> {
     let mut identity_path = dirs::data_local_dir().ok_or(Error::NoDataLocalDir)?;
     identity_path.push("teliod");
     if !identity_path.exists() {
         let _ = create_dir_all(&identity_path);
     }
-    identity_path.push(&format!("{interface_name}.json"));
+    identity_path.push("data.json");
 
     let mut device_identity = match DeviceIdentity::from_file(&identity_path) {
         Some(identity) => identity,
         None => {
             let private_key = SecretKey::gen();
-            let hw_identifier = generate_hw_identifier(private_key.clone());
+            let hw_identifier = generate_hw_identifier();
 
             let machine_identifier =
                 match fetch_identifier_with_exp_backoff(auth_token, private_key.public()).await {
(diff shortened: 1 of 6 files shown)


clis/teliod/src/nc.rs
```diff
@@ -86,12 +86,15 @@ struct NCConfig {
...
+    pub async fn new(
+        config: &super::TeliodDaemonConfig,
+        hw_identifier: &Uuid,
+    ) -> Result<Self, Error> {
         let callbacks = Arc::new(Mutex::new(vec![]));
 
         let nc_config = NCConfig {
             authentication_token: config.authentication_token.clone(),
-            app_user_uid: config.app_user_uid,
+            app_user_uid: *hw_identifier,
```
matislovas, line 97: nit: maybe rename `NCConfig`'s `app_user_uid` to `hw_id`?

matislovas:
+1

clis/teliod/src/nc.rs
ghztomash: From MQTT API point of view this field is still called `app_user_uid`. I could change the `NCConfig` but `TokenHttpRequest`would still have to remain `app_user_uid` otherwise it would break the HTTP request.

I would rather change the `new()` signature.

Hasan6979:
What is the difference between `app_user_id` and `hw_identifier` ? 
From what I understand is that, one is more device specific, and the other is user specific right ? 1 user can have many devices.

tomasz-grz:
@Hasan6979 One is used to identify the device on meshnet, and the other to identify the device for MQTT subscriptions.

> The parameter `app_user_uid` has to be an identifier that is unique per device and user.

clis/teliod/src/core_api.rs
```diff
@@ -71,22 +71,19 @@ fn build_backoff() -> Result<ExponentialBackoff, Error> {
...
-    interface_name: &str,
-) -> Result<DeviceIdentity, Error> {
+pub async fn init_with_api(auth_token: &str) -> Result<DeviceIdentity, Error> {
     let mut identity_path = dirs::data_local_dir().ok_or(Error::NoDataLocalDir)?;
     identity_path.push("teliod");
     if !identity_path.exists() {
         let _ = create_dir_all(&identity_path);
     }
-    identity_path.push(&format!("{interface_name}.json"));
+    identity_path.push("data.json");
```
lcruz99, line 80: what's the reason behind this change?

tomasz-grz: Based from the discussions on slack, we want to destroy the old config if we switch users.
Currently, if you change the interface name in the config file, it will create a new identity file leaving the old one behind.

Hasan6979:
+1

clis/teliod/src/nc.rs
tomasz-grz: Updated the signature.

packgron (approved):
+1.0

PR #1029, opened 2024-12-16
Disable software update downloads on the Mac VM

tomaszklak:
### Problem
We suspect that some failures that happen when sw update is happening are caused by high network traffic during that time.

### Solution
Disable not only doing software update but also downloading it.


matislovas:
I thought it is more _correct_ to put that to provision scripts ..?

PR #1032, opened 2024-12-17
Check that the VMs can ping each other

tomaszklak:
### Problem
There is no point in running tests when networking is broken

### Solution
Check that VMs can reach each other, and also at least one selected docker host.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -4,11 +4,13 @@ import pytest
 import random
 import shutil
 import subprocess
+from config import DERP_PRIMARY
 from datetime import datetime
 from helpers import SetupParameters
 from interderp_cli import InterDerpClient
 from itertools import combinations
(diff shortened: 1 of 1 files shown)


nat-lab/tests/conftest.py
```diff
@@ -196,6 +202,52 @@ def pytest_make_parametrize_id(config, val):
...
+    }
+    for source, destinations in test_nodes.items():
+        for dest_ip in destinations:
+            try:
+                async with new_connection_with_conn_tracker(source, None) as (
+                    connection,
+                    _,
+                ):
+                    await ping(connection, dest_ip, 5)
+                    results[source].append((dest_ip, True))
```
LukasPukenis, line 237: Very minor: can we do reverse ip lookup and insert hostname instead in here to avoid having map the IP to the machine?

LukasPukenis (approved):
+1

LukasPukenis (approved):

PR #1034, opened 2024-12-17
Fix FiveTuple printing

Jauler:
### Problem
FiveTuple from connection tracker, when printed would produce the following output:
```
In nlx_1 there has been 3 connections to <class 'utils.connection_tracker.FiveTuple'> which is more then max limit of 2
```
Obviously `<class 'utils.connection_Tracker.FiveTuple>` is not desirable, hence this PR adds `def __str__` implementation to the class.

This is the result:
2024-12-17_17-20-44_grim

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection_tracker.py
+++ b/nat-lab/tests/utils/connection_tracker.py
@@ -41,6 +41,9 @@ class FiveTuple:
             (self.protocol, self.src_ip, self.dst_ip, self.src_port, self.dst_port)
         )
 
+    def __str__(self) -> str:
+        return f"{self.protocol} {self.src_ip}:{self.src_port} -> {self.dst_ip}:{self.dst_port}"
+
 
 class EventType(Enum):
     """Event type reported by conntrack"""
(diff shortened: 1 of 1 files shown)


nat-lab/tests/utils/connection_tracker.py
```diff
@@ -41,6 +41,9 @@ def __hash__(self):
             (self.protocol, self.src_ip, self.dst_ip, self.src_port, self.dst_port)
         )
 
+    def __str__(self) -> str:
+        return f"{self.protocol} {self.src_ip}:{self.dst_ip} -> {self.dst_ip}:{self.dst_port}"
```
mathiaspeters, line 45:
```suggestion
        return f"{self.protocol} {self.src_ip}:{self.src_port} -> {self.dst_ip}:{self.dst_port}"
```

mathiaspeters:
Tiny sorta-unrelated nit/request: Can we change `In nlx_1 there has been 3 connections to <class 'utils.connection_tracker.FiveTuple'> which is more then max limit of 2` so that it says `than` instead of `then`. #sorrynotsorry

Jauler:
> Tiny sorta-unrelated nit/request: Can we change `In nlx_1 there has been 3 connections to <class 'utils.connection_tracker.FiveTuple'> which is more then max limit of 2` so that it says `than` instead of `then`. #sorrynotsorry

nice catch. Fixed.

nat-lab/tests/utils/connection_tracker.py
Jauler: Fixed

mathiaspeters (approved):
+2

PR #1035, opened 2024-12-17
Add timestamp in logs (also) on libtelio side

tomaszklak:
By default timestamp are not added, but tests-like environments might want to add timestamps on libtelio side.

This is enabeld for nat-lab / libtelio-remote tests.

- [x] README.md is updated


src/ffi/logging.rs
```diff
@@ -258,7 +263,17 @@ fn start_async_logger(cb: Box<dyn TelioLoggerCb>, buffer_size: usize) -> SyncSen
             let _ = cb.log(TelioLogLevel::Debug, START_ASYNC_LOGGER_MSG.to_owned());
             loop {
                 match receiver.recv() {
-                    Ok(LogMessage::Message(level, msg)) => {
+                    Ok(LogMessage::Message(level, msg, timestamp)) => {
+                        let msg = if TIMESTAMPS_IN_LOGS.load(std::sync::atomic::Ordering::Relaxed) {
```
LukasPukenis, line 267: Question - do we need to use atomic in here since we just write to it once in the beginning so there should be no races possible for it. Treat it as a question rather than a nitpick :)

LukasPukenis (commented):
Nice +1

dfetti (approved):
+1.0

src/ffi/logging.rs
tomaszklak: It can't be const bool because we don't have special testing build for nat-lab, it the same artifact as used in production. It can't be static bool because `add_timestamps_to_logs` and this load are not happening on the same (`libtelio-logger`) thread. This means that it needs to be either `Mutex<bool>` or `AtomicBool` to allow for correct r/w access from multiple threads.

PR #1036, opened 2024-12-18
Refactor batcher and inline into SessionKeeper

LukasPukenis:
Refactor batcher and inline into SessionKeeper
    
    This commit does these things:
    - Inlines batching flag to SessionKeeper, removing batcher entirely
    as a separate entity with all of its logic(threshold, traffic trigger).
    - Emits all actions from SessionKeeper if batching is enabled when
    an action is added. This works nicely if we can guarantee that actions
    are gonna be a multiple of some T which is the case as currently we
    target T=70s for proxy, direct, vpn, stun keepalives. This also makes
    thresholds and the whole logic around them irrelevant.
    
    Traffic triggered batching implementation was doing more harm than good
    since it triggers on _any_ traffic, meaning even when alignment was
    achieved, it might misalign since any traffic triggers it. This needs to
    be done via separate channel which is out of scope for this PR
    
    Feature flags are not removed in this commit so not to push for
    version update of libtelio, so now they are no-op.

### Problem

### Solution

- [x] README.md is updated


crates/telio-traversal/src/session_keeper.rs
```diff
@@ -194,113 +170,81 @@ impl SessionKeeperTrait for SessionKeeper {
...
+                        }
+                        Ok(())
+                    })
+                }),
+            ))
+        })
+        .await
+        .map_err(Error::Task)?
+        .map_err(Error::RepeatedActionError)
+        .map(|_| ())?;
```
jjanowsk, line 213: `.map(|_| ())?;` I think this is a no-op and can be removed

crates/telio-utils/src/repeated_actions.rs
```diff
@@ -171,6 +166,111 @@ mod tests {
...
+            .unwrap();
+
+        // at this point in time there's 7seconds until action_0 and 10seconds until action_1
+
+        ctx.actions.select_action().await.unwrap().1(&mut ctx)
+            .await
+            .unwrap();
+        ctx.actions.select_action().await.unwrap().1(&mut ctx)
+            .await
+            .unwrap();
```
jjanowsk, line 238: Why special case for ts_10 and ts_13? You can just add them to the vector and loop 6 times instead of 4

crates/telio-utils/src/repeated_actions.rs
```diff
@@ -171,6 +166,111 @@ mod tests {
...
+            loop {
+                if values_to_expect.len() == 0 {
+                    break;
+                }
+                ctx.actions.select_action().await.unwrap().1(&mut ctx)
+                    .await
+                    .unwrap();
+
+                assert_eq!(ctx.get(), values_to_expect.pop().unwrap());
+            }
```
jjanowsk, line 253: Maybe just:
```
let values_to_expect = vec!["ts_10", "ts_13", "ts_20", "ts_23", "ts_30", "ts_33"];
for time in values_to_expect {
    ctx.actions.select_action().await.unwrap().1(&mut ctx)
        .await
        .unwrap();
    assert_eq!(ctx.get(), time);
}
```

crates/telio-utils/src/repeated_actions.rs
```diff
@@ -171,6 +166,111 @@ mod tests {
...
```
jjanowsk, line 272: Same here

src/device/wg_controller.rs
```diff
@@ -366,6 +366,11 @@ async fn consolidate_wg_peers<
     let actual_keys: HashSet<&PublicKey> = actual_peers.keys().collect();
     let delete_keys = &actual_keys - &requested_keys;
     let insert_keys = &requested_keys - &actual_keys;
+    telio_log_debug!(
+        "actual_keys: {}, requested_keys: {}",
+        actual_peers.keys().len(),
+        requested_peers.keys().len(),
+    );
```
jjanowsk, line 373: Leftover from debugging?

src/device/wg_controller.rs
```diff
@@ -767,6 +757,10 @@ async fn build_requested_peers_list<
                 (_, pq_keys) => {
                     let preshared_key = pq_keys.map(|pq| pq.pq_shared);
 
+                    telio_log_debug!(
+                        "Will insert with keepalive of {:?}",
+                        persistent_keepalive_interval
+                    );
```
jjanowsk, line 763: Also look like some leftover?

jjanowsk (commented):
❤️ ❤️ ❤️ 
image

Only some minor nits from my side

crates/telio-traversal/src/session_keeper.rs
LukasPukenis: You're right, removed it. Probably previously it was inlined directly into `Ok()` so it made sense

crates/telio-utils/src/repeated_actions.rs
LukasPukenis: did so :+1:

src/device/wg_controller.rs
LukasPukenis: Good catch, thanks!

src/device/wg_controller.rs
LukasPukenis: Removed it, it was a leftover

jjanowsk (approved):
+1

src/device.rs
```diff
@@ -1341,10 +1336,13 @@ impl Runtime {
...
         // functionality will break when offloading actions to it, thus we disable the feature
-        if session_keeper.is_none() && self.features.batching.is_some() {
-            telio_log_warn!(
-                "Batching feature is enabled but SessionKeeper failed to start. Disabling batching."
-            );
+        if session_keeper.is_none() {
+            if self.features.batching.is_some() {
+                // Batching feature enables batching for various kinds of keepalives provided by
+                // SessionKeeper. As a fallback, disable batcher in case SessionKeeper fails to
+                // start to preerve maximum operatibility of libtelio
```
ghztomash, line 1343: `preserve`?

ghztomash (approved):
+1 looks good

src/device.rs
LukasPukenis: whoops :D

PR #1042, opened 2024-12-20
Make sure that build scripts use rerun-if-changed in a correct way

tomaszklak:
For lana, this script is self contained so it should explicitly depend only on itself. For toplevel, we don't need to check changes in crates and src if the bypass is in place.

After this change, subsequent invocations of `run_local.py` or `build.sh`, will not rebuild/relink anything if there are no changes.

- [x] README.md is updated

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -73,9 +73,6 @@ fn check_git_hooks() -> Result<()> {
 }
 
 fn verify_llt_secrets() {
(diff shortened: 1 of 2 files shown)


LukasPukenis (approved):
+2

LukasPukenis (approved):

PR #1044, opened 2024-12-20
Misalign batch testcase peers and add scoring

LukasPukenis:
### Problem
Batching testcase was too ideal, just two peers and they interconnect immediately

### Solution
- Add 2 additional peers
- Misalign all of their starting times and also their direct connection formation by blocking direct link for some time

This required to proceed with peer initialization manually instead of using the regular helpers

- [x] README.md is updated


nat-lab/tests/test_batching.py
```diff
@@ -29,17 +29,19 @@
     render_chart,
     generate_packet_distribution_histogram,
     generate_packet_delay_histogram,
+    get_ordered_histogram_score,
 )
 
-BATCHING_MISALIGN_S = 7
-BATCHING_CAPTURE_TIME = 120  # Tied to TEST_BATCHING_TIMEOUT
+BATCHING_CAPTURE_TIME = 130
+DOCKER_CONE_GW_2_IP = "10.0.254.2"
```
dfetti, line 36: These IP is also defined in test_direct_connection.py. Maybe you can put them all in config.py or somewhere to avoid redefinitions.

dfetti (commented):
+1.0

nat-lab/tests/test_batching.py
LukasPukenis: I think it's a very valid point but to make it properly the whole natlab could go through a refactor to extract IPs, which actually would be a good initiative.

Doing so in this PR is a bit out of scope as it's too big, so I will ticket that as a natlab improvement so it could be properly implemented through the whole natlab instead

dfetti (approved):
+1.0

PR #1046, opened 2024-12-23
Make pollKeepalive optional

tomaszklak:
### Problem
`pool_keepalive` was made a required field. This is causing failures in unaligned clients (the fact that it has a default serde value in the json doesn't help).

### Solution
Make the field optional in the udl.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -330,7 +330,7 @@ pub struct FeatureDerp {
     /// Poll Keepalive: Application level keepalives meant to replace the TCP keepalives
     /// They will use derp_keepalive as interval
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


tomaszklak:
> +1
> 
> Would we bump the major version now because we changed the UDL? or is it backwards compatible because of the optional?

We didn't release version with this new field so as far as I understand, if we release 5.2 after this PR it will be just like the addition of non-optional didn't happen. Which means that we added optional field which is afaik compatible with applications using previous (5.1.5) version.

mathiaspeters (approved):
+2

This would not be a breaking change even if it was released and actively used by apps. Breaking changes, by definition, are when we make a change that makes it not work for someone, but this just allows omitting the field without any other effects, so not a breaking change

PR #1049, opened 2025-01-07
LLT-5884: PQ VPN breaks after some time of nonet

mathiaspeters:
### Problem
If a client is connected to PQ VPN and then doesn't do a new handshake within wireguard's 180s reject connection timeout (can happen for example during nonet), the client would not be able to establish a new handshake since it's supplying a preshared key even though the server doesn't expect one, leading to a broken VPN connection

### Solution
If postquantum is active and time since last handshake exceeds wireguards reject threshold (in `uapi::Peer::time_since_last_handshake` this is represented as `None`), restart postquantum

- [x] README.md is updated

### DIFF
--- a/crates/telio-pq/src/entity.rs
+++ b/crates/telio-pq/src/entity.rs
@@ -1,12 +1,15 @@
 use std::net::SocketAddr;
 use std::sync::Arc;
 
+use parking_lot::Mutex;
+
 use telio_model::features::FeaturePostQuantumVPN;
 use telio_task::io::chan;
 use telio_utils::telio_log_debug;
 
 struct Peer {
     pubkey: telio_crypto::PublicKey,
+    wg_secret: telio_crypto::SecretKey,
     addr: SocketAddr,
     /// This is a key rotation task guard, its `Drop` implementation aborts the task
     _rotation_task: super::conn::ConnKeyRotation,
@@ -17,16 +20,16 @@ pub struct Entity {
     features: FeaturePostQuantumVPN,
     sockets: Arc<telio_sockets::SocketPool>,
     chan: chan::Tx<super::Event>,
-    peer: Option<Peer>,
+    peer: Mutex<Option<Peer>>,
 }
 
 impl crate::PostQuantum for Entity {
     fn keys(&self) -> Option<crate::Keys> {
-        self.peer.as_ref().and_then(|p| p.keys.clone())
+        self.peer.lock().as_ref().and_then(|p| p.keys.clone())
     }
 
     fn is_rotating_keys(&self) -> bool {
-        self.peer.is_some()
+        self.peer.lock().is_some()
     }
 }
 
@@ -40,44 +43,50 @@ impl Entity {
             features,
             sockets,
             chan,
-            peer: None,
+            peer: Mutex::new(None),
         }
     }
 
-    pub fn on_event(&mut self, event: super::Event) {
-        let Some(peer) = &mut self.peer else {
-            return;
-        };
-
-        match event {
-            super::Event::Handshake(addr, keys) => {
-                if peer.addr == addr {
-                    peer.keys = Some(keys);
-                }
-            }
-            super::Event::Rekey(super::Keys {
-                wg_secret,
-                pq_shared,
-            }) => {
-                if let Some(keys) = &mut peer.keys {
-                    // Check if we are still talking to the same exit node
-                    if keys.wg_secret == wg_secret {
-                        // and only then update the preshared key,
-                        // otherwise we're connecting to different node already
-                        keys.pq_shared = pq_shared;
-                    } else {
-                        telio_log_debug!(
-                            "PQ secret key does not match, ignoring shared secret rotation"
-                        );
+    pub fn on_event(&self, event: super::Event) {
+        if let Some(peer) = self.peer.lock().as_mut() {
+            match event {
+                super::Event::Handshake(addr, keys) => {
+                    if peer.addr == addr {
+                        peer.keys = Some(keys);
                     }
                 }
+                super::Event::Rekey(super::Keys {
+                    wg_secret,
+                    pq_shared,
+                }) => {
+                    if let Some(keys) = &mut peer.keys {
+                        // Check if we are still talking to the same exit node
+                        if keys.wg_secret == wg_secret {
+                            // and only then update the preshared key,
+                            // otherwise we're connecting to different node already
+                            keys.pq_shared = pq_shared;
+                        } else {
+                            telio_log_debug!(
+                                "PQ secret key does not match, ignoring shared secret rotation"
+                            );
+                        }
+                    }
+                }
+                _ => (),
             }
-            _ => (),
         }
     }
 
-    pub async fn stop(&mut self) {
-        if let Some(peer) = self.peer.take() {
+    pub fn peer_pubkey(&self) -> Option<telio_crypto::PublicKey> {
+        self.peer.lock().as_ref().map(|p| p.pubkey)
+    }
+
+    // There is some temporary event filtering around disconnected events for postquantum
+    // That filtering depends heavily on the peer being removed before emitting Disconnected
+    // here
+    pub async fn stop(&self) {
+        let peer = self.peer.lock().take();
+        if let Some(peer) = peer {
             #[allow(mpsc_blocking_send)]
             let _ = self
                 .chan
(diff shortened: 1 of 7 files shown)


nat-lab/tests/test_pq.py
```diff
@@ -136,7 +137,7 @@ async def test_pq_vpn_connection(
 ) -> None:
     async with AsyncExitStack() as exit_stack:
         env = await exit_stack.enter_async_context(
-            setup_environment(exit_stack, [alpha_setup_params])
+            setup_environment(exit_stack, [alpha_setup_params], prepare_vpn=True)
```
mathiaspeters, line 140: Existing bug in this test. Without this fix the test only passes if some other test that does prepare for VPN is run before this test in the same test session

nat-lab/tests/test_pq.py
```diff
@@ -238,7 +239,7 @@ async def test_pq_vpn_rekey(
 
     async with AsyncExitStack() as exit_stack:
         env = await exit_stack.enter_async_context(
-            setup_environment(exit_stack, [alpha_setup_params])
+            setup_environment(exit_stack, [alpha_setup_params], prepare_vpn=True)
```
mathiaspeters, line 242: Existing bug in this test. Without this fix the test only passes if some other test that does prepare for VPN is run before this test in the same test session

crates/telio-pq/src/entity.rs
```diff
@@ -17,16 +20,16 @@ pub struct Entity {
     features: FeaturePostQuantumVPN,
     sockets: Arc<telio_sockets::SocketPool>,
     chan: chan::Tx<super::Event>,
-    peer: Option<Peer>,
+    peer: Mutex<Option<Peer>>,
```
mathiaspeters, line 23: This peer needs to be restarted (requires mutation) in places where we currently didn't have a mutable reference to it. To make the change transparent to all places where `entities.postquantum_wg` is used, this was wrapped in a mutex to add interior mutability

crates/telio-pq/src/entity.rs
```diff
@@ -40,44 +43,47 @@ impl Entity {
...
-        let Some(peer) = &mut self.peer else {
-            return;
-        };
-
-        match event {
-            super::Event::Handshake(addr, keys) => {
-                if peer.addr == addr {
-                    peer.keys = Some(keys);
+    pub fn on_event(&self, event: super::Event) {
+        if let Some(peer) = self.peer.lock().as_mut() {
```
mathiaspeters, line 51: `let Some(peer) = self.peer.lock() else { return; }` sadly doesn't work so we'll have to deal with an extra layer of indentation here

crates/telio-pq/src/entity.rs
```diff
@@ -111,4 +118,16 @@ impl Entity {
...
+
+    pub async fn restart(&self) {
+        let params = self.peer.lock().as_ref().map(|peer| {
+            let addr = peer.addr;
+            let pubkey = peer.pubkey;
+            let wg_secret = peer.wg_secret.clone();
+            (addr, pubkey, wg_secret)
+        });
+        if let Some((addr, pubkey, wg_secret)) = params {
+            self.start(addr, wg_secret, pubkey).await;
```
mathiaspeters, line 130: `start()` calls `stop()` internally so it shouldn't need to be called manually

nat-lab/tests/test_pq.py
```diff
@@ -510,3 +511,40 @@ async def test_pq_vpn_upgrade_from_non_pq(
...
+        ip = await stun.get(client_conn, config.STUN_SERVER)
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        await _connect_vpn_pq(
+            client_conn,
+            client_alpha,
+        )
+
+        async with client_alpha.get_router().break_udp_conn_to_host(str(config.NLX_SERVER["ipv4"])), client_alpha.get_router().break_tcp_conn_to_host(str(config.NLX_SERVER["ipv4"])):
+            await asyncio.sleep(195)
```
mathiaspeters, line 546: The reject threshold in wireguard is 180s but for some reason this sleep had to be at least 195s to trigger the bug

src/device/wg_controller.rs
```diff
@@ -147,6 +149,42 @@ pub async fn consolidate_wg_state(
...
+    let should_restart_pq = match entities.postquantum_wg.peer_addr() {
+        Some(addr) => {
+            let time_since_last_handshake = entities
+                .wireguard_interface
+                .get_interface()
+                .await
+                .ok()
+                .and_then(|ifc| {
+                    ifc.peers.iter().find_map(|(_, peer)| {
+                        if peer.endpoint.is_some_and(|e| e == addr) {
```
Jauler, line 172: A small thought, could we match peer not by its endpoint address, but by public key? I am afraid that endpoint addresses does not really provide any uniqueness guarantees and keys has that quite a bit better.

nat-lab/tests/test_pq.py
```diff
@@ -510,3 +511,43 @@ async def test_pq_vpn_upgrade_from_non_pq(
...
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        await _connect_vpn_pq(
+            client_conn,
+            client_alpha,
+        )
+
+        async with client_alpha.get_router().break_udp_conn_to_host(
+            str(config.NLX_SERVER["ipv4"])
+        ), client_alpha.get_router().break_tcp_conn_to_host(
```
Jauler, line 548: I do not believe there is a need to break tcp connections to VPN server. WG is UDP only.

nat-lab/tests/test_pq.py
```diff
@@ -510,3 +511,43 @@ async def test_pq_vpn_upgrade_from_non_pq(
...
+            str(config.NLX_SERVER["ipv4"])
+        ):
+            await asyncio.sleep(195)
+
+        await ping(client_conn, config.PHOTO_ALBUM_IP, timeout=10)
```
Jauler, line 553: Note that pinging here does not guarantee that new PQ session has in-fact occured. In practice the important factor to reproduce the issue is that server-side abandons the session due to inactivity, which does not happen within 195s normally, meaning that old session likely can resurrect itself.

Also sleeping in tests is often the source of flakyness.

But I do have a suggestion. IMHO instead of waiting for there server-side to abandon the session we can force that by either explicitly deleting the peer (by using `nlx peer <pubkey> remove` command, or by restarting whole server) on the server side. And then waiting for client-side to recover from such situation. This will make test event based, will ensure that server-side has abandoned the peer, so we are testing for the correct scenario.

But do not consider this comment as a blocker for merging fix. If apps QA confirm in-branch, that the sleep issue is fixed - merge it, and we can do hotfix release. The test is here to avoid the same future regressions, so it can be done in separate PR.

crates/telio-pq/src/entity.rs
Jauler: Can we add the comment in the `.stop()` callsite (in `.start()` function) that `restart()` depends on `.stop()` being there. Just in case, as it might be not obvious why `.stop()` is required there.

src/device/wg_controller.rs
mathiaspeters: Good idea, fixed

nat-lab/tests/test_pq.py
mathiaspeters: I didn't like the sleep, especially not such a long one, but didn't want to dig too much into how to trigger it without the sleep. I will start working on this, if I finish it before the fix has been fully validated it will be included in this PR, otherwise in a follow-up

mathiaspeters: I've been thinking about this and I think what you're suggesting would be much harder to implement the bug fix for. The test and fix as currently written are based on the fact that wireguard has a hardcoded 180s timeout (since last handshake) before a connection is considered inactive and packets are rejected. I fully agree that writing a test this way can be a source of flakiness but I do think the fix is sound, although it would make sense to update the condition to be `Some(None) || Some(duration) if duration > 180s` to be more complete. As for the test, I 100% agree that a sleep-based test like this could be a likely source of flakiness but if we replaced the sleep by, for example, `nlx peer <pubkey> remove`, the fix wouldn't actually do anything until the 180s has passed anyways, and detecting arbitrarily that the server has abandoned the session would not be easy.

Me an Lukas did mention though that if the server can provide reason why a connection was rejected, that would be something that would be better basis for this test and fix. When that information is available we should definitely add the option of restarting postquantum (or whatever else might need it) based on that, but until then, I don't see a way for us to solve it differently (not saying a different way doesn't exist, just that it would be significantly harder and I can't think of a solution right now)

LukasPukenis: Deleting the peer and expecting the failure - we cannot detect that just by using wireguard's timestamps quicker than in 180seconds. Client will maintain the last handshake timestamp and will assume everything is fine up until we reach WG's REJECT-AFTER-KEY, which is the condition of flow restart for pq which looks correct here.

Without an additional active probing to aid in no-net detection or expected pings from the server towards the client(similar as with direct connection downgrade) we can't really do anything, so deleting `nlx` peer will not achieve the test goal and the sleep is the only solution I see here working in this case, unfortunately of course :/

I think it might be worth exploring if we can fast-forward time in docker containers instead of sleeping though :)

crates/telio-pq/src/entity.rs
LukasPukenis: Hot take - I think it would be easier instead of comments, to just make it explicit and call `stop()`. `start()` should just start. There's also `fn restart()` so currently it's pretty confusing :)

crates/telio-pq/src/entity.rs
LukasPukenis: Makes perfect sense

nat-lab/tests/test_pq.py
mathiaspeters: From a short investigation it is possible to fast-forward time so I can investigate that, but I would do that as a follow-up PR

crates/telio-pq/src/entity.rs
mathiaspeters: To make this conversation moot, `restart` no longer depends on `stop` at all

LukasPukenis (approved):
+2 Android confirmed the events are fixed as well and connectivity is restored.

For the time fast-forwarding I suggest creating a ticket but it should be a generic one as other testcases also use some sleep :)

PR #1056, opened 2025-01-13
Llt 5865 cgi web UI

packgron:
### Problem
Qnap need web ui, based on cgi

### Solution
A htmx + tailscale based solution, that satisfies major part of web ui requirements.


clis/teliod/assets/telio.js
```diff
@@ -0,0 +1,19 @@
...
+  validateToken: input => {
+    const valid = /^[0-9a-f]{64}$/.test(input.value);
+    if (valid) {
+      input.setCustomValidity("");
+    } else {
+      input.setCustomValidity("Token must be a valid 64char hex number");
+    }
+  },
+  
+  validateTunnel: input =>  {
```
matislovas, line 11: Do we have access from script side, if the actual interface specified here exists?

lcruz99: No, this is the input the user chooses for the interface name. This script will validate the chosen name against the regex expr. below.

lcruz99: If the interface exists and its creation are tasks on telio side.

clis/teliod/src/cgi/mod.rs
```diff
@@ -54,12 +60,14 @@ impl Deref for CgiRequest {
     }
 }
 
+#[allow(dead_code)]
```
matislovas, line 63: Why are those not removed?

contrib/http_root/cgi-bin/teliod.cgi
```diff
@@ -10,7 +10,10 @@ fail() {
 DEV_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 BIN_DIR="$(realpath $DEV_DIR/../../../target/debug/)"
 
-cargo build -F cgi -p teliod >"$DEV_DIR/cargo.log" 2>&1 || fail
+export BYPASS_LLT_SECRETS=1
```
matislovas, line 13: Why is it needed here?

contrib/http_root/cgi-bin/teliod.cgi
```diff
@@ -10,7 +10,10 @@ fail() {
...
+BUILD_USER=${SUDO_USER:-$(cat $DEV_DIR/builder)}
+
+sudo --preserve-env=BYPASS_LLT_SECRETS -u $BUILD_USER -- cargo build -F cgi -p teliod >"$DEV_DIR/cargo.log" 2>&1 || fail
```
matislovas, line 16: Is this called everytime user enters teliod section in qnap?

matislovas:
+1

contrib/http_root/cgi-bin/teliod.cgi
lcruz99: Everytime the user makes a request, ie. runs teliod.cgi script.

LukasPukenis:
I observed that it was working when you demonstrated it which means at least to some degree it's proven to be working and other things we will just find out. There's nothing obvious sticking out except maybe for `BYPASS_LLT_SECRETS`.

There's one caveat though - because QNAP UI lives in the same repository as libtelio and doesn't use libtelio as a submodule instead, it is bound to be broken when incompatible libtelio changes appear since we can't tag specific libtelio version which we are sure is compatible. This might break it by accident(libtelio introduces a breaking change, we push a new qnap build to the appstore).

Another way to address this is to have some sort of a test for it. Since it's a Web UI test there are plenty of options to choose from but still, it's extra work to set up the pipelines and the tests themselves.

Regardless the tests or libtelio locking they can be done separately and do not block this PR, but once merged we should address it.

+1

LukasPukenis (approved):
+1

PR #1058, opened 2025-01-13
LLT-5886: Protected RTT QoS pings

tomasz-grz:
### Problem
Due to the nature of Apple's NetworkExtension control protocol, packets originating from the NetworkExtension process are routed outside of the tunnel. 
Because of this on macOS AppStore and iOS platforms, raw ICMP packets from QoS RTT and Link detection components destined for the tunnel interface, leave the main interface instead.

### Solution
Use the `SocketPool` component in order to bind the ping raw socket to the tunnel interface. The Pinger component had to moved into it's own crate in order to avoid circular dependencies between Utils and Sockets.
LinkDetection component was also updated to use the new Pinger.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-5886_protected_pinger
+++ b/.unreleased/LLT-5886_protected_pinger
@@ -1,0 +1,2 @@
+Use the SocketPool component in order to bind the Pinger raw socket to the tunnel interface on macOS.
+LinkDetection component was also updated to use the new Pinger.
--- a/crates/telio-pinger/src/lib.rs
+++ b/crates/telio-pinger/src/lib.rs
@@ -5,7 +5,8 @@ use std::time::Duration;
 use std::{convert::TryInto, net::IpAddr};
 use surge_ping::{Client, Config as PingerConfig, PingIdentifier, PingSequence, ICMP};
 
-use crate::{telio_log_debug, telio_log_error, DualTarget};
+use telio_sockets::SocketPool;
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_trace, DualTarget};
 
 /// Information needed to check the reachability of endpoints.
 ///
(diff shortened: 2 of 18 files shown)


.unreleased/LLT-5886_protected_pinger
Jauler, line 1: This change I guess will be relevant to apple teams in changelogs, so maybe we could add some message here.

crates/telio-wg/src/wg.rs
```diff
@@ -345,11 +345,18 @@ impl DynamicWg {
...
             adapter,
             link_detection,
-            cfg,
+            cfg.try_clone()?,
             ipv6_enabled,
+            Some(cfg.socket_pool),
         ));
         #[cfg(windows)]
-        return Ok(Self::start_with(io, adapter, link_detection, ipv6_enabled));
+        return Ok(Self::start_with(
```
Jauler, line 353: Socket pool is available on all platforms, maybe it would be simpler/easier to just use it regardless of platforms? Less conditionals makes testing easier, and less edge cases :thinking: WDYT?

crates/telio-pinger/src/lib.rs
```diff
@@ -42,20 +43,37 @@ pub struct DualPingResults {
...
+    /// * `socket_pool` - Optional SocketPool used to protect the sockets.
+    pub fn new(
+        no_of_tries: u32,
+        ipv6: bool,
+        socket_pool: Option<Arc<SocketPool>>,
+    ) -> std::io::Result<Self> {
         let client_v6 = if ipv6 {
-            Some(Arc::new(Self::build_client(ICMP::V6)?))
+            let client_v6 = Arc::new(Self::build_client(ICMP::V6)?);
+            if let Some(ref socket_pool) = socket_pool {
```
Jauler, line 60: Maybe it makes sense to add non-optional dependency? e.g. to unconditionally create sockets using `telio-sockets`, as it is a common place which handles various socket protection mechanisms, binding monitoring etc. And it makes it easier and less edge-casy, if we reduce number of conditionals? WDYT?

.unreleased/LLT-5886_protected_pinger
tomasz-grz: Good point, added a comment

crates/telio-wg/src/wg.rs
tomasz-grz: Good idea, converted the `socket_pool` to no-optional and it's passed in all cases.
Also added a mock for the Protector.

crates/telio-pinger/src/lib.rs
tomasz-grz: Thank you, addressed this.

LukasPukenis (commented):
Very nice PR.

I checked out the `telio-sockets/protector/*` and it doesn't contain the information of what it does except the `apple.rs` which mentions what it tries to do.

While we're at it, would it be possible to add a brief comment into `protector.rs` about it's purpose?

Anyway, that's not a blocker :) +1

tomasz-grz:
Also added rust docs to the protector module.

Jauler (approved):
+1.0

PR #1061, opened 2025-01-15
Temporary disable tcpdump for windows

gytsto:
### Problem
windump might be causing issues for windows vms and make tests fail

### Solution
temporary disable windump

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/tcpdump.py
+++ b/nat-lab/tests/utils/tcpdump.py
@@ -162,6 +162,10 @@ async def make_tcpdump(
     try:
         async with AsyncExitStack() as exit_stack:
             for conn in connection_list:
+                # TODO(gytsto): temporary disable windows tcpdump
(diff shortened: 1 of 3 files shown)


nat-lab/tests/utils/tcpdump.py
```diff
@@ -162,6 +162,10 @@ async def make_tcpdump(
     try:
         async with AsyncExitStack() as exit_stack:
             for conn in connection_list:
+                # TODO(gytsto): temporary disable windows tcpdump
```
LukasPukenis, line 165: Adding a ticket number would allow us not to forget it was disabled

LukasPukenis (dismissed):
+2

nat-lab/tests/utils/tcpdump.py
gytsto: added

LukasPukenis (approved):

PR #1062, opened 2025-01-15
Add faster WG polling after update

ThrasherLT:
### Problem
Current WireGuard polling mechanism is inefficient and thus slowing down connection times for VPN.

### Solution
Connection times can be sped up by speeding up polling period after an update had been made to WireGuard.
Implements RFC LLT-0083.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/wg.rs
+++ b/crates/telio-wg/src/wg.rs
@@ -14,8 +14,8 @@ use telio_model::{
 use telio_sockets::{NativeProtector, SocketPool};
 use telio_utils::{
     dual_target::{DualTarget, DualTargetError},
-    get_ip_stack, interval, telio_err_with_log, telio_log_debug, telio_log_error, telio_log_trace,
-    telio_log_warn, IpStack,
+    get_ip_stack, interval, interval_at, telio_err_with_log, telio_log_debug, telio_log_error,
+    telio_log_trace, telio_log_warn, IpStack,
 };
 use thiserror::Error as TError;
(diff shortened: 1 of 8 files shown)


Jauler (dismissed):
+1.0, Nice work!!!

crates/telio-wg/src/wg.rs
```diff
@@ -1293,6 +1331,54 @@ pub mod tests {
...
+        adapter.expect_send_uapi_cmd_generic_call(1).await;
+        // Calling `.update()` on `State` without any changes.
+        cause_update(&wg).await;
+        // The update above should have caused the polling period to become faster.
+        assert_eq!(
+            get_polling_period(&wg).await,
+            Duration::from_millis(DEFAULT_POLLING_PERIOD_AFTER_UPDATE_MS)
+        );
+        // But after the original period passess, polling period should become slower again.
+        tokio::time::advance(Duration::from_millis(DEFAULT_POLLING_PERIOD_MS)).await;
```
tomaszklak, line 1369: Afaics, this test verifies that the internal fields change to expected values, but doesn't check that the expected behavior is happening. I think that we should verify that during the `DEFAULT_POLLING_PERIOD_MS` after the update, our code will read from the uapi at least `DEFAULT_POLLING_PERIOD_MS / DEFAULT_POLLING_PERIOD_AFTER_UPDATE_MS` times.

tomaszklak (approved):
+1

PR #1066, opened 2025-01-17
Update libtelio dependencies

jjanowsk:
- [x] README.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -3170,7 +3170,8 @@ pub mod tests {
             name: "eth0".to_string(),
             addr: IfAddr::V4(Ifv4Addr {
                 ip: Ipv4Addr::new(192, 168, 1, 10),
-                netmask: Ipv4Addr::new(192, 168, 1, 0),
+                netmask: Ipv4Addr::new(255, 0, 0, 0),
+                prefixlen: 8,
                 broadcast: None,
             }),
             index: Some(12),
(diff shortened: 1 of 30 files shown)


crates/telio-firewall/src/firewall.rs
```diff
@@ -3170,7 +3170,8 @@ pub mod tests {
             name: "eth0".to_string(),
             addr: IfAddr::V4(Ifv4Addr {
                 ip: Ipv4Addr::new(192, 168, 1, 10),
-                netmask: Ipv4Addr::new(192, 168, 1, 0),
+                netmask: Ipv4Addr::new(255, 0, 0, 0),
+                prefixlen: 8,
```
LukasPukenis, line 3174: Judging by the diff, the netmask was incorrect but the unit test was not failing. Is it possible to make it unit-tested as well while we're at it?

LukasPukenis (commented):
Missing `.unreleased` entry. Though the PR itself looks OK and the linked pipeline reveals only one failed job due to Windows VM.

One thing that needs to be addressed - `pqcrypto-internals` update, it fails on our current Mac builders and I've manually reverted it not long ago so you will need to do that or else it will not build for apple with our runners :). This one is blocking

However without it +1

crates/telio-firewall/src/firewall.rs
jjanowsk: It was not failing because the code does not care about this netmask at all, and there is not reason to test it. If there was a Default implementation for Ifv4Addr we could use it, but there is not, so we need to initialize it with some dummy data

LukasPukenis (approved):
+1

PR #1072, opened 2025-01-22
LLT-5939: Fix test after windows VM provisioning fix

mathiaspeters:
### Problem
We recently changed the network model for the windows VMs for natlab, which makes a test fail because the adapter is different

### Solution
Update the test

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_cleanup.py
+++ b/nat-lab/tests/test_cleanup.py
@@ -41,7 +41,7 @@ async def test_get_network_interface_tunnel_keys(adapter_type, name) -> None:
         # that might have managed to survive the end of the previous test.
         keys = await _get_network_interface_tunnel_keys(connection)
         assert [
-            "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}\\0005"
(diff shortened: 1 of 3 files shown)


nat-lab/tests/test_cleanup.py
```diff
@@ -6,6 +6,8 @@
 from utils.vm.windows_vm_util import _get_network_interface_tunnel_keys
 
 
+# This xfail is highly temporary and will very soon be removed
+@pytest.mark.xfail
```
stalowyjez, line 10: Maybe mark this as TODO?

.unreleased/LLT-5939
```diff
@@ -0,0 +1 @@
+
```
stalowyjez, line 1: Why isn't it just empty?

tomasz-grz (commented):
+1 but please add the TODO

.unreleased/LLT-5939
mathiaspeters: Helix automatically added the newline when I saved the file. It should be empty now

nat-lab/tests/test_cleanup.py
mathiaspeters: Added

stalowyjez (dismissed):
Well, I thought more about making this comment a `TODO`, but that's ok too, +1

jjanowsk (dismissed):
+1 But IMO you can  just merge failing '-build' MR and commit this change together with version update to not waste time

jjanowsk (approved):

PR #1074, opened 2025-01-23
LLT-5886: Protected RTT QoS pings

tomasz-grz:
### Problem
Due to the nature of Apple's NetworkExtension control protocol, packets originating from the NetworkExtension process are routed outside of the tunnel.
Because of this on macOS AppStore and iOS platforms, raw ICMP packets from QoS RTT and Link detection components destined for the tunnel interface, leave the main interface instead.

### Solution
Use the SocketPool component in order to bind the ping raw socket to the tunnel interface. The Pinger component had to moved into it's own crate in order to avoid circular dependencies between Utils and Sockets.
LinkDetection component was also updated to use the new Pinger.

Note: on iOS the requested binding is not kept, and we have to rebind the socket before every ping.

The PR has been broken up in to two parts: second refactoring apart https://github.com/NordSecurity/libtelio/pull/1135

- [x] README.md is updated

### DIFF
--- a/crates/telio-pinger/src/lib.rs
+++ b/crates/telio-pinger/src/lib.rs
@@ -5,7 +5,8 @@ use std::time::Duration;
 use std::{convert::TryInto, net::IpAddr};
 use surge_ping::{Client, Config as PingerConfig, PingIdentifier, PingSequence, ICMP};
 
-use crate::{telio_log_debug, telio_log_error, DualTarget};
+use telio_sockets::SocketPool;
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_trace, DualTarget};
 
 /// Information needed to check the reachability of endpoints.
 ///
@@ -42,20 +43,33 @@ pub struct DualPingResults {
 impl Pinger {
     const PING_TIMEOUT: Duration = Duration::from_secs(5);
 
-    /// Create new instance of `Ping`.
+    /// Create new instance of `Ping` with a socket pool.
     ///
     /// # Arguments
     ///
     /// * `no_of_tries` - How many pings should be sent.
-    pub fn new(no_of_tries: u32, ipv6: bool) -> std::io::Result<Self> {
+    /// * `ipv6` - Enable IPv6 support.
+    /// * `socket_pool` - Optional SocketPool used to protect the sockets.
+    pub fn new(
+        no_of_tries: u32,
+        ipv6: bool,
+        socket_pool: Arc<SocketPool>,
+    ) -> std::io::Result<Self> {
         let client_v6 = if ipv6 {
-            Some(Arc::new(Self::build_client(ICMP::V6)?))
+            let client_v6 = Arc::new(Self::build_client(ICMP::V6)?);
+            telio_log_trace!("Making pinger IPv6 socket internal");
+            socket_pool.make_internal(client_v6.get_socket().get_native_sock())?;
+            Some(client_v6)
         } else {
             None
         };
 
+        let client_v4 = Arc::new(Self::build_client(ICMP::V4)?);
+        telio_log_trace!("Making pinger IPv4 socket internal");
+        socket_pool.make_internal(client_v4.get_socket().get_native_sock())?;
+
         Ok(Self {
-            client_v4: Arc::new(Self::build_client(ICMP::V4)?),
+            client_v4,
             client_v6,
             no_of_tries,
         })
@@ -179,5 +193,44 @@ impl Pinger {
             }
         }
         Client::new(&config_builder.build())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use telio_sockets::protector::MockProtector;
+
+    // Basic constructor test
+    #[tokio::test]
+    async fn test_pinger_new_v6_sock_pool() {
+        let mut protect = MockProtector::default();
+        protect.expect_make_internal().returning(|_| Ok(()));
+
+        let pinger = Pinger::new(1, true, Arc::new(SocketPool::new(protect)))
+            .expect("Failed to create Pinger");
+        assert!(pinger.client_v4.get_socket().get_native_sock() > 0);
+        assert!(pinger.client_v6.is_some());
+        assert_eq!(pinger.no_of_tries, 1);
+    }
+
+    // Basic ping test
+    #[tokio::test]
+    async fn test_ping_localhost() {
+        let mut protect = MockProtector::default();
+        protect.expect_make_internal().returning(|_| Ok(()));
+
+        let pinger = Pinger::new(2, false, Arc::new(SocketPool::new(protect)))
+            .expect("Failed to create Pinger");
+
+        let target =
+            DualTarget::new(("127.0.0.1".parse().ok(), None)).expect("Failed to create target");
+
+        let result = pinger.perform(target).await;
+        assert!(
+            result.v4.unwrap().successful_pings > 0,
+            "Expected at least one successful ping to 127.0.0.1"
+        );
+        assert!(result.v6.is_none());
     }
 }
--- a/crates/telio-pinger/src/lib.rs
+++ b/crates/telio-pinger/src/lib.rs
@@ -6,16 +6,20 @@ use std::{convert::TryInto, net::IpAddr};
 use surge_ping::{Client, Config as PingerConfig, PingIdentifier, PingSequence, ICMP};
 
 use telio_sockets::SocketPool;
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_trace, DualTarget};
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_trace, telio_log_warn, DualTarget};
+
+const PING_PAYLOAD_SIZE: usize = 56;
 
 /// Information needed to check the reachability of endpoints.
 ///
 /// Can be used with both IPv4 and IPv6 addresses.
+#[derive(Clone)]
 pub struct Pinger {
     client_v4: Arc<Client>,
     client_v6: Option<Arc<Client>>,
     /// Number of tries
     pub no_of_tries: u32,
+    socket_pool: Arc<SocketPool>,
 }
 
 /// Information gathered after a ping action
@@ -44,12 +48,13 @@ impl Pinger {
     const PING_TIMEOUT: Duration = Duration::from_secs(5);
 
     /// Create new instance of `Ping` with a socket pool.
+    /// For performing pings inside of the tunnel.
     ///
     /// # Arguments
     ///
     /// * `no_of_tries` - How many pings should be sent.
     /// * `ipv6` - Enable IPv6 support.
-    /// * `socket_pool` - Optional SocketPool used to protect the sockets.
+    /// * `socket_pool` - SocketPool used to protect the sockets.
     pub fn new(
         no_of_tries: u32,
         ipv6: bool,
@@ -72,6 +77,7 @@ impl Pinger {
             client_v4,
             client_v6,
             no_of_tries,
+            socket_pool,
         })
     }
 
@@ -126,21 +132,20 @@ impl Pinger {
             ..Default::default()
         };
 
-        telio_log_debug!("Trying to ping {:?} host", host);
+        let ping_id = PingIdentifier(rand::random());
+        telio_log_debug!("Trying to ping {:?} host, {:#06x}", host, ping_id.0);
 
-        let mut pinger = match host {
-            IpAddr::V4(_) => {
-                self.client_v4
-                    .clone()
-                    .pinger(host, PingIdentifier(rand::random()))
-                    .await
-            }
+        let (mut pinger, socket) = match host {
+            IpAddr::V4(_) => (
+                self.client_v4.pinger(host, ping_id).await,
+                self.client_v4.get_socket().get_native_sock(),
+            ),
             IpAddr::V6(_) => {
                 if let Some(client) = &self.client_v6 {
-                    client
-                        .clone()
-                        .pinger(host, PingIdentifier(rand::random()))
-                        .await
+                    (
+                        client.pinger(host, ping_id).await,
+                        client.get_socket().get_native_sock(),
+                    )
                 } else {
                     return None;
                 }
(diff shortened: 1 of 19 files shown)


crates/telio-traversal/src/session_keeper.rs
```diff
@@ -313,14 +233,17 @@ mod tests {
...
             .unwrap()
-            .get_socket();
+            .get_client(ICMP::V4)
+            .unwrap();
+
+        let sock = pinger_client.get_socket();
 
         // Drop the pinger client explicitly, for it to stop listening on ICMP socket
-        drop(sess_keep.get_pinger_client(ICMP::V4).await.unwrap());
+        drop(pinger_client);
```
LukasPukenis, line 246: Could we provide a function in `Pinger` that would consume `self` by value and return the socket instead? Maybe even do so with `fn get_socket(self) -> Socket`, this way we wouldn't need to drop anything explicitly

crates/telio-pinger/src/lib.rs
```diff
@@ -42,22 +49,39 @@ pub struct DualPingResults {
...
 
-    /// Create new instance of `Ping`.
+    /// Create new instance of `Ping` with a socket pool.
     ///
     /// # Arguments
     ///
     /// * `no_of_tries` - How many pings should be sent.
-    pub fn new(no_of_tries: u32, ipv6: bool) -> std::io::Result<Self> {
+    /// * `ipv6` - Enable IPv6 support.
+    /// * `socket_pool` - Optional SocketPool used to protect the sockets.
```
LukasPukenis, line 58: This is not `Optional` :)

crates/telio-pinger/src/lib.rs
```diff
@@ -42,22 +49,39 @@ pub struct DualPingResults {
...
+        no_of_tries: u32,
+        ipv6: bool,
+        socket_pool: Arc<SocketPool>,
+        payload_data: Option<&str>,
+    ) -> std::io::Result<Self> {
         let client_v6 = if ipv6 {
-            Some(Arc::new(Self::build_client(ICMP::V6)?))
+            let client_v6 = Arc::new(Self::build_client(ICMP::V6)?);
+            telio_log_trace!("Making pinger IPv6 socket internal");
+            socket_pool.make_internal(client_v6.get_socket().get_native_sock())?;
```
LukasPukenis, line 69: I am trying to think if there are situations where this component might be used to do outside-the-tunnel pings :thinking: if that's ever a case it would be nice to immediately know that this is a wrong component to do so at the moment(`make_internal` in the constructor and in `ping_action` for iOS would need to be removed or added under a flag).

I think it's worth adding a disclaimer for it near `new` so it would be immediately obvious that. Though this is minimal risk probably, still, details get lost in time :)

LukasPukenis (commented):
I tried to give it a good look and left some comments. It appears as the pinger should work and rebind the sockets for iOS every time and all of: link detection, session keeper, qos - should use the new pinger instead of surge.

I haven't found obvious issues

+0.75

crates/telio-pinger/src/lib.rs
tomasz-grz: You're right, this is a leftover from when it was optional 😅

crates/telio-pinger/src/lib.rs
```diff
@@ -5,16 +5,23 @@ use std::time::Duration;
 use std::{convert::TryInto, net::IpAddr};
 use surge_ping::{Client, Config as PingerConfig, PingIdentifier, PingSequence, ICMP};
 
-use crate::{telio_log_debug, telio_log_error, DualTarget};
+use telio_sockets::SocketPool;
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_trace, telio_log_warn, DualTarget};
+
+const PING_PAYLOAD_SIZE: usize = 56;
```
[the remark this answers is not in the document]
tomasz-grz: Indeed it is :D

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -281,6 +200,7 @@ impl Runtime for State {
 mod tests {
     use super::*;
     use std::net::{Ipv4Addr, Ipv6Addr};
+    use surge_ping::ICMP;
```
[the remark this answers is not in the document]
tomasz-grz: for `get_client(ICMP::V4)`, but it can also be made to just return a tuple of both clients.

crates/telio-wg/Cargo.toml
```diff
@@ -46,6 +47,7 @@ tokio = { workspace = true, features = ["test-util"] }
 
 telio-firewall.workspace = true
 telio-task = { workspace = true, features = ["test-util"] }
+telio-sockets = { workspace = true, features = ["mockall"] }
```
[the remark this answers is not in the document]
tomasz-grz: Removed

crates/telio-traversal/src/session_keeper.rs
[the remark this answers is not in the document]
tomasz-grz: Refactored `get_client()` into `take_sockets()`

crates/telio-pinger/src/lib.rs
```diff
@@ -112,21 +136,19 @@ impl Pinger {
             ..Default::default()
         };
 
-        telio_log_debug!("Trying to ping {:?} host", host);
+        let ping_id = PingIdentifier(rand::random());
+        telio_log_debug!("Trying to ping {:?} host, {:#06x}", host, ping_id.0);
+        let socket;
 
         let mut pinger = match host {
```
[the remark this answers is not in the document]
tomasz-grz: Updated

crates/telio-traversal/src/session_keeper.rs
tomasz-grz: Updated to `get_sockets()`

crates/telio-pinger/src/lib.rs
tomasz-grz: Updated

crates/telio-pinger/src/lib.rs
tomasz-grz: Added a disclaimer.

crates/telio-traversal/src/session_keeper.rs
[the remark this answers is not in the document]
tomasz-grz: But forgot to remove the `use` :D

crates/telio-pinger/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+                    )
+                } else {
+                    return None;
+                }
+            }
+        };
+
+        pinger.timeout(Self::PING_TIMEOUT);
+        #[cfg(test)]
+        pinger.timeout(Duration::from_millis(10));
```
mathiaspeters, line 179: When I read this it seems to me like the intention is something like 
```
#[cfg(not(test)]
normal timeout
#[cfg(test)]
test timeout
```
but that's not what it looks like. Is that intended?

crates/telio-pinger/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+    fn build_client(proto: ICMP) -> std::io::Result<Client> {
+        let mut config_builder = PingerConfig::builder().kind(proto);
+        if cfg!(any(target_os = "macos", target_os = "tvos",)) {
+            config_builder = config_builder.sock_type_hint(Type::RAW);
+        }
+        // Raw sockets are not allowed on iOS, instead use
+        // socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP)
+        //
+        // TODO: LLT-5991: Refactor writing ICMP packets directly into the tunnel
+        // TODO: Check if raw sockets are supported on tvos
```
mathiaspeters, line 232: If we don't know if raw sockets are supported on tvos, is the previous `if` a good idea then? Is there a better "default" solution that we use until we determine that? Like reusing the ios solution instead?

crates/telio-pinger/src/lib.rs
tomasz-grz:
```rust
#[cfg(test)]
pinger.timeout(Duration::from_millis(10));
```
Would override the previous `pinger.timeout()` so IMHO there's no need to compile it out 🤷 (probably gets optimized out anyway).

But I do agree maybe something like this would probably be cleaner
```rust
        pinger.timeout(if cfg!(test) {
            Duration::from_millis(10)
        } else {
            Self::PING_TIMEOUT
        });
```
Or even better, make it runtime configurable so we don't have conditional compilation in "critical" parts

mathiaspeters: I think `#[cfg()]` is better than `if cfg!()` though. And I just didn't dig deep enough in what the timeout is doing, so all good as-is

crates/telio-pinger/src/lib.rs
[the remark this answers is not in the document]
tomasz-grz: This part has been moved to the refactoring PR

LukasPukenis (commented):
The code itself looks good, however I have concerns about the correctness of it and this mostly comes from the fact there are no iOS and TvOS tests for this.

There are multiple solutions bundled in one go(which is very nice):
- SessionKeeper, Nurse, Link detection not using `make_internal`
- `make_internal` being done repeatedly for iOS
- Raw sockets are not allowed on iOS thus it falls back to DGRAM

The SK, LinkDetection, Nurse using socket_pool is a good standalone change and could be merged in immediately with no extra work.

However for socket rebinding and RAW vs DGRAM it would be nice to somehow ensure correctness of it since we had(and have now more) conditionals around tvOS, iOS. Though this might not be easily solvable, since we do not possess tvOS or iOS device in runners so the only solution is to in-branch test and mention in the PR that it was done and it seems to work :)

crates/telio-pinger/src/lib.rs
tomasz-grz: tvOS testing was requested, but it will take some time before we receive any results. So for now we need to make some assumptions and move on.

tomasz-grz: For now our assumption is that existing implementation for tvOS works correctly since we only received bug-reports for iOS.

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -195,6 +210,23 @@ impl SessionKeeperTrait for SessionKeeper {
                 interval,
                 Arc::new(move |c| {
                     Box::pin(async move {
+                        // This is a temporary solution for iOS due to NECP re-binding the socket to
```
Jauler, line 213: TBH, if the re-binding is only required for ICMP - IMHO this is not likely to be temporarely solution/workaround. Therefore we can probably cancel LLT-5991 and remove this comment.

What do you think?

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -79,21 +79,25 @@ impl SessionKeeper {
...
-            target_os = "macos",
-            target_os = "tvos",
-        )) {
+        if cfg!(any(target_os = "macos", target_os = "tvos",)) {
             config_builder = config_builder.sock_type_hint(Type::RAW);
         }
+        // Raw sockets are not allowed on iOS, instead use
+        // socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP)
+        //
+        // TODO: LLT-5991: Refactor writing ICMP packets directly into the tunnel
```
Jauler, line 96: If this NECP issue affects only ICMP -> we may cancel LLT-5991, as writing packets directly is pretty much the same level workaround as rebinding.

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -79,21 +79,25 @@ impl SessionKeeper {
...
     }
 
     fn make_builder(proto: ICMP) -> ConfigBuilder {
         let mut config_builder = PingerConfig::builder().kind(proto);
-        if cfg!(any(
-            target_os = "ios",
-            target_os = "macos",
-            target_os = "tvos",
-        )) {
+        if cfg!(any(target_os = "macos", target_os = "tvos",)) {
```
Jauler, line 90: IMHO it is much more likely that `tvOS` will behave the same as `iOS`. Are you sure we should include tvos here? Maybe it should use Type::DGRAM?

crates/telio-pinger/src/lib.rs
```diff
@@ -140,8 +159,22 @@ impl Pinger {
         let mut sum = Duration::default();
 
         for i in 0..self.no_of_tries {
+            // This is a temporary solution for iOS due to NECP re-binding the socket to
+            // the main interface after every write, refer to LLT-5886
+            //
+            // TODO: LLT-5991: Refactor writing ICMP packets directly into the tunnel
```
Jauler, line 165: If this NECP issue affects only ICMP -> imho we can cancel LLT-5991, as it is similar level workaround as rebinding before writes.

crates/telio-traversal/src/session_keeper.rs
tomasz-grz: I suspect so too, but both would be guesses until we test this 🤷 so I only moved iOS since we got the bug reports only from them.

crates/telio-traversal/src/session_keeper.rs
tomasz-grz: Only ICMP was tested from the Network layer, so other protocols could still be affected.. Are we writing packets like this any other protocols? Transport layer protocols and above don't seem to be affected.

Possibly, also might not be worth the effort if it affects only one platform.. :D But I just make new tickets and let PO's decide on the priority.. The question is, if the binding workaround stops working at any point when Apple decides to update something.

crates/telio-traversal/src/session_keeper.rs
tomasz-grz: Every temporary solution is quite likely to become a permanent solution :D

crates/telio-traversal/src/session_keeper.rs
```diff
@@ -195,6 +210,23 @@ impl SessionKeeperTrait for SessionKeeper {
                 interval,
                 Arc::new(move |c| {
                     Box::pin(async move {
+                        // This is a temporary solution for iOS due to NECP re-binding the socket to
+                        // the main interface after every write, refer to LLT-5886
+                        //
+                        // TODO: LLT-5991: Refactor writing ICMP packets directly into the tunnel
+                        if cfg!(target_os = "ios") {
```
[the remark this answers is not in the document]
tomasz-grz: I guess we could assume that "re-binding" just in case wouldn't hurt, but I would rather be conservative with the amount of changes

crates/telio-pinger/src/lib.rs
Jauler: "Cleaner" is probably subjective thing... We may leave the ticket open if you think it may be useful.

PR #1077, opened 2025-01-24
Lukasp/more prints in natlab tests

LukasPukenis:
### Problem

### Solution


nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -12,9 +12,8 @@
 from threading import Lock
 from typing import List
 
-REMOTE_LOG = "remote.log"
-TCLI_LOG = "tcli.log"
-MOOSE_LOGS_DIR = "/moose_logs"
+# type: ignore # pylint: disable=import-error
+from uniffi.constants import TCLI_LOG, NATLAB_EVENTS_LOG, REMOTE_LOG, MOOSE_LOGS_DIR
```
gytsto, line 16: This creates circular dependency and you are just hiding by disabling pylint error. I think it should be something like this, but not 100% sure:
```python
from .constants import TCLI_LOG, ...
```

LukasPukenis: Actually it was a leftover but good catch. I removed it and pylint is still happy

mathiaspeters (approved):
+2

mathiaspeters (dismissed):

nat-lab/tests/telio.py
```diff
@@ -411,6 +411,7 @@ async def on_stderr(stderr: str) -> None:
             await exit_stack.enter_async_context(
                 self._process.run(stdout_callback=on_stdout, stderr_callback=on_stderr)
             )
+            name = self._connection.target_name()
```
mathiaspeters, line 414: Feel free to disagree with me but I'm wondering if it would be better to use the node name here, since most of the prints we do during the test use that, which could make it easier to correlate which logs belong together

nat-lab/tests/telio.py
```diff
@@ -876,11 +873,16 @@ async def _event_request_loop(self) -> None:
                 event = await self.get_proxy().next_event()
                 while event:
                     if self._runtime:
-                        print(f"[{self._node.name}]: event [{datetime.now()}]: {event}")
+                        print(
+                            f"[{self._node.name}]:[{datetime.now()}] event loop procesing: {event}"
```
mathiaspeters, line 877: Super small nit:
```suggestion
                            f"[{self._node.name}]:[{datetime.now()}] event loop processing: {event}"
```

nat-lab/tests/uniffi/libtelio_proxy.py
```diff
@@ -87,7 +88,14 @@ def create(self):
 
     @move_to_async_thread
     def next_event(self) -> libtelio.Event:
-        return self._handle_remote_error(lambda r: r.next_event())
+        try:
+            ev = self._handle_remote_error(lambda r: r.next_event())
+            print(f"[{self._name}]:{datetime.now()} pyro::next_event returned: {ev}")
+        except Exception as e:
+            print(f"[{self._name}]:{datetime.now()} pyro::next_event: exception {e}")
```
mathiaspeters, line 95: I would call these `proxy` instead of `pyro`. It's true that they go over the pyro connection but it's our own proxy code being called

nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -45,6 +45,7 @@ def __init__(self):
         self._events: list[libtelio.Event] = []
 
     def event(self, payload):
+        print(f"{datetime.datetime.now()} enqueue event: {payload}")
```
mathiaspeters, line 48: It would be nice to get a name in here as well. You could pass it to the `create` method here in the remote and then include that in the prints

nat-lab/tests/telio.py
LukasPukenis: I agree, replaced it, having `alpha` or `beta` makes it nicer, thanks for the suggestion :)

nat-lab/tests/uniffi/libtelio_proxy.py
LukasPukenis: Renamed it <3

nat-lab/tests/telio.py
LukasPukenis: Nice catch!

nat-lab/tests/uniffi/libtelio_remote.py
LukasPukenis: This one is weird in that it already prints the `[name]` so I deliberately skipped it in here :)

mathiaspeters: Oh right, I forgot about this one. There are stdout and stderr callbacks that make sure these get printed and those include the node name

mathiaspeters (approved):
+2

PR #1078, opened 2025-01-27
Enable merge queue in our workflows

jjanowsk:
We need to enable merge queues in this repo, because manual rebasing is annoying. This is the first step to do that.


Jauler (approved):
+2.0

PR #1081, opened 2025-01-28
Lukasp/battery doc

LukasPukenis:
### Problem
Battery optimisation ideas are all around the internal docs while it would be nicer to have that info closer to the source code.

### Solution
add `BATTERY.MD` file describing toggles and knobs for better battery performance

- [x] README.md is updated


lcruz99 (approved):
+2

PR #1082, opened 2025-01-28
Convert nat-lab related python usage to uv

tomaszklak:
### Problem
Currently we are using pip(env) to manage python dependencies. It's more complicated than it should be:
- python dependencies are split across different files
- configuration is in yet another file
- preparing environment requires multiple commands
- requires host system to have correct python version

### Solution
[uv](https://docs.astral.sh/uv/) solves most above issue by combining project configuration into one file (`pyproject.toml`), making it easy to run arbitrary commands in correct environment (it's enough to run it using `uv run` which will construct required environment on the fly) and being able to install any python version.

- [x] README.md is updated


.gitlab-ci.yml
```diff
@@ -15,4 +15,5 @@ libtelio-build-pipeline:
 
   trigger:
     project: $LIBTELIO_BUILD_PROJECT_PATH
-    branch: v2.8.2 # REMEMBER to also update in .github/workflows/gitlab.yml
+    branch: convert-to-uv # REMEMBER to also update in .github/workflows/gitlab.yml
```
tomaszklak, line 18: REMINDER: change before merge

Justfile
```diff
@@ -55,23 +55,41 @@ deny: _deny-install
...
+
+# Run mypy type checker
+[working-directory: 'nat-lab']
+mypy:
+    uv run --isolated mypy .
+
+# Run the autoflake linter
+[working-directory: 'nat-lab']
+autoflake:
+    uv run --isolated autoflake --quiet --check .
```
LukasPukenis, line 91: Offtopic: while looking up about `uv` I found https://github.com/astral-sh/ruff - we could replace a bunch of different tools with just that one as a follow-up.

What's your opinion on that one?

nat-lab/README.md
```diff
@@ -5,34 +5,38 @@ We are working on making it locally runnable for everyone outside of
...
 ```
 
 ## Linter and typecheck
 
 Python code is checked using https://github.com/python/mypy and https://github.com/pylint-dev/pylint.
 Any faulty code will automatically be declined by the CI.
 ```
-mypy . && pylint .
+just mypy && just pylint
 ```
```
LukasPukenis, line 27: Possible improvement: I have an idea to propose but haven't yet though out it fully -  combining all of these jobs(3 formatters, 2 linters) into one `just python` and all github jobs into one.

In that case when developing, at the end we can `just python` and it's ready for pushing. If it fails due to linting for example, we fix it until `just python` passes, then push. In the pipeline if we see it fail, we know we need `just python` locally and see what we missed. What do you think about this one? :)

LukasPukenis (commented):
I've experimented with it locally, it runs fine and looks cleaner which is a nice win. Not having multiple Pipfiles is very nice.

I left two possible suggestions to potentially simplify things even further but those are QOL improvements

+1

Justfile
tomaszklak: I've also seen it but to be honest I'm not knowledgeable enough about python ecosystem to have much of an opinion. It's probably fine and I would much prefer to have just one linter, but I have no idea why we have some many different kinds of linters in the first place - it's a Chesterton's fence for me :)

nat-lab/README.md
tomaszklak: I've added new just action `python_checks` to run all python related linters.

Justfile
LukasPukenis: Ah okay, thanks for elaborating :)

.github/workflows/linters.yml
```diff
@@ -87,24 +87,39 @@ jobs:
...
         - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
-        - run: pip3 install --no-deps -r requirements.txt
-        - run: pipenv install --system
-        - run: isort --check-only --diff .
-          working-directory: nat-lab
+        - uses: extractions/setup-just@dd310ad5a97d8e7b41793f8ef055398d51ad4de6 # v2.0.0
+          with:
+            just-version: 1.38.0
+        - name: Install uv
+          uses: astral-sh/setup-uv@v4
```
Jauler, line 105: Could you add hash here 🙏 ?

.github/workflows/linters.yml
```diff
@@ -87,24 +87,39 @@ jobs:
...
         - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
-        - run: pip3 install --no-deps -r requirements.txt
-        - run: pipenv install --system
-        - run: autoflake --quiet --check .
-          working-directory: nat-lab
+        - uses: extractions/setup-just@dd310ad5a97d8e7b41793f8ef055398d51ad4de6 # v2.0.0
+          with:
+            just-version: 1.38.0
+        - name: Install uv
+          uses: astral-sh/setup-uv@v4
```
Jauler, line 118: Could you add hash here 🙏 ?

.github/workflows/linters.yml
```diff
@@ -87,24 +87,39 @@ jobs:
         - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
         - uses: extractions/setup-just@dd310ad5a97d8e7b41793f8ef055398d51ad4de6 # v2.0.0
           with:
-            just-version: 1.37.0
+            just-version: 1.38.0
+        - name: Install uv
+          uses: astral-sh/setup-uv@v4
```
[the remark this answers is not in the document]
tomaszklak: Done 👍

PR #1085, opened 2025-01-28
[LLT-5960] Add spans to Derp

stalowyjez:
### Problem
Maybe it would be better to add spans to track which logs are from which module.

### Solution
Soo, this is not a full solution, rather some small sample for Derp and I'm not really sure if we want to add this in the whole code base, as this looks kinda ugly imho (i.e. you have everywhere this `instrument(tracing::Span::current())` polluting the code) - maybe I understood something wrong and there is some nicer way of doing that.

Additionally we get a bunch of auto-generated tracing Events about spans (i.e. when they are created, or entered, or closed), which can be useful (probably especially when we would use some smarter Subscriber):
```TelioLogLevel.INFO 2025-01-28T16:02:57.700119637Z ThreadId(15) "telio_relay::derp":392 (DerpRelay::get_remote_peer_states) close time.busy=6.38µs time.idle=37.0µs```


packgron (commented):
I would recommend to actually start from telio-task, to give a general idea of in witch task something is happening, then move with concrete components like derp.

Other nice thing you could do is probably imporve task_exec macro to create originating span, with info about what triggered so action in Task.
There is this thing: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.follows_from.

I image then we not only see that some log happened in TaskA, but what caused TaskA to do something.

stalowyjez:
>I would recommend to actually start from telio-task, to give a general idea of in witch task something is happening, then move with concrete components like derp.

I'm not sure what do you mean by that - i.e. I added it in derp to have a general idea that stuff happens in derp task, but I don't understand what would be benefit of placing them in `telio-task`? I thought in the beginning that I can add some 

>Other nice thing you could do is probably imporve task_exec macro to create originating span, with info about what triggered so action in Task.
There is this thing: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.follows_from.

Soo, do you think `task_exec` should have an additional field, sth like `span_name` to create this new spans?

Jauler:
I had slightly different understanding on "expected" usage of spans. I am totally not sure if my understanding is correct, so take this with a grain of salt. But we can discuss this anyways.

My understanding is that span is something that is roughly designed to track lifetimes of things. Like a lifetime of an object (struct) or entity etc. It can also be used to track lifetime (e.g. when it started, when it ended and what has happened in the "middle") of semantic things as well (e.g. like relay connection), and here is the place where the most benefit could be reaped out of it. 

In relays code case we have generally a few such thingies, which to me looks "interesting" to spannize :laughing::
* The `derp.rs` task itself (e.g. enter span on start, exit span on stop)
* Then there is this "State.connecting" task, which handles connection inititiation, it may have its own span. Which somehow should be related to relay task span. 
* and then we have `State.conn` object, which should live as long as relay connection. Spannizing (tracking the lifetime of) this connection object should be of interest for us as well. 

I am hoping that you can attach some chain or tree of spans, such that relay task is kind of top-level, triggering two child ones, either `connecting` or `State.conn` one. These spans lifetime (enter/exit) should match creation or deletion of corresponding object (in other words span should be a member of corresponding struct, such that their lifetimes matches exactly), and then whenever we are logging, we should be using them somehow.

This way later, when analyzing logs, we can easily separate out various logs for specific connection to see what happened to particular connection. We can see when specific connection was created and how many in parallel we had at a time etc. This way of adding spans would support concurrent connections and so on.

But again, this understanding of mine comes with quite a few assumptions, which may be incorrect, so take it with a grain of salt.

PR #1086, opened 2025-01-29
Move dns request handler to separate thread

Hasan6979:
### Problem
Sometimes, dns tests fail in natlab due to some problem in the [hickory-server ](https://github.com/NordSecurity/trust-dns) (the root cause isn't clear yet). This causes the blocking in the runtime.

### Solution
Handle DNS request in a separate task, which wouldn't block the rest of the code even if that task is blocked due to unresponsive server. Ideally even if there is a problem, hickory-server should respond with an error.

Note: This PR doesn't solve dns tests failing, rather just makes sure we don't block if the hickory-server doesn't respond.

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/src/nameserver.rs
+++ b/crates/telio-dns/src/nameserver.rs
@@ -21,9 +21,13 @@ use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     str::FromStr,
     sync::Arc,
+    time::Duration,
 };
 use telio_model::features::TtlValue;
-use tokio::sync::{RwLock, RwLockMappedWriteGuard, RwLockWriteGuard, Semaphore};
+use tokio::sync::{
+    mpsc::{Receiver, Sender},
+    RwLock, RwLockMappedWriteGuard, RwLockWriteGuard, Semaphore,
+};
 use tokio::task::JoinHandle;
 use tokio::{net::UdpSocket, sync::Mutex};
 
@@ -62,6 +66,7 @@ pub trait NameServer {
 pub struct LocalNameServer {
     zones: Arc<ClonableZones>,
     task_handle: Option<JoinHandle<()>>,
+    shutdown_request_sender: Option<Sender<()>>,
 }
 
 impl LocalNameServer {
@@ -71,6 +76,7 @@ impl LocalNameServer {
         let ns = Arc::new(RwLock::new(LocalNameServer {
             zones: Arc::new(ClonableZones::new()),
             task_handle: None,
+            shutdown_request_sender: None,
         }));
         ns.forward(forward_ips).await?;
         Ok(ns)
@@ -78,9 +84,9 @@ impl LocalNameServer {
 
     async fn dns_service(
         peer: Arc<Mutex<Tunn>>,
-        nameserver: Arc<RwLock<LocalNameServer>>,
         socket: Arc<UdpSocket>,
         dst_address: SocketAddr,
+        tx: tokio::sync::mpsc::Sender<Vec<u8>>,
     ) {
         let mut receiving_buffer = vec![0u8; MAX_PACKET];
         let mut sending_buffer = vec![0u8; MAX_PACKET];
@@ -112,10 +118,10 @@ impl LocalNameServer {
 
             let peer = peer.clone();
             let socket = socket.clone();
-            let nameserver = nameserver.clone();
             let mut receiving_buffer = receiving_buffer.clone();
             let mut sending_buffer = sending_buffer.clone();
             let semaphore = semaphore.clone();
+            let tx = tx.clone();
             tokio::spawn(async move {
                 let res = peer.lock().await.decapsulate(
                     None,
@@ -153,51 +159,74 @@ impl LocalNameServer {
                                 return;
                             }
                         };
-
-                        let nameserver = nameserver.clone();
-                        let length = match LocalNameServer::process_packet(
-                            nameserver,
-                            packet,
-                            &mut receiving_buffer,
-                        )
-                        .await
-                        {
-                            Ok(length) => length,
-                            Err(e) => {
-                                telio_log_error!(
-                                    "[DNS] {}. Offending request packet: {:?}",
-                                    e,
-                                    packet
-                                );
-                                return;
-                            }
-                        };
-
-                        let tunn_res = peer.lock().await.encapsulate(
-                            receiving_buffer.get(..length).unwrap_or(&[]),
-                            &mut sending_buffer,
-                        );
-                        match tunn_res {
-                            TunnResult::WriteToNetwork(dns) => {
-                                if let Err(e) = socket.send_to(dns, dst_address).await {
-                                    telio_log_warn!(
-                                        "[DNS] Failed to send DNS query response  {:?}",
-                                        e
-                                    )
-                                };
-                            }
-                            TunnResult::Err(e) => {
-                                telio_log_warn!(
-                                    "[DNS] Failed to encapsulate DNS query response  {:?}",
-                                    e
-                                )
-                            }
-                            _ => {}
+                        #[allow(index_access_check)]
+                        receiving_buffer[..packet.len()].copy_from_slice(packet);
+                        if let Err(e) = tx.send(receiving_buffer).await {
+                            telio_log_error!("Unable to forward packet to dns request handler {e}");
                         }
                     }
                     _ => {}
                 }
             });
+        }
+    }
+
+    /// Handle processing dns request and sending to
+    /// hickory-server to avoid getting stuck due to
+    /// unresponsivenes from hickory
+    fn handle_dns_request(
+        socket: Arc<UdpSocket>,
+        nameserver: Arc<RwLock<LocalNameServer>>,
+        peer: Arc<Mutex<Tunn>>,
+        dst_address: SocketAddr,
+        mut rx: Receiver<Vec<u8>>,
+        mut awaiting_shutdown: Receiver<()>,
+    ) {
+        tokio::task::spawn(async move {
+            while let Some(packet) = rx.recv().await {
+                telio_log_debug!("Handling dns request");
+                let mut receiving_buffer = vec![0u8; MAX_PACKET];
+                let mut sending_buffer = vec![0u8; MAX_PACKET];
+                let nameserver = nameserver.clone();
+                let length = match LocalNameServer::process_packet(
+                    nameserver,
+                    &packet,
+                    &mut receiving_buffer,
+                )
+                .await
+                {
+                    Ok(length) => length,
+                    Err(e) => {
+                        telio_log_error!("[DNS] {}. Offending request packet: {:?}", e, packet);
+                        MAX_PACKET + 1
+                    }
+                };
+
+                if length < MAX_PACKET {
+                    let tunn_res = peer.lock().await.encapsulate(
+                        receiving_buffer.get(..length).unwrap_or(&[]),
+                        &mut sending_buffer,
+                    );
+                    match tunn_res {
+                        TunnResult::WriteToNetwork(dns) => {
+                            if let Err(e) = socket.send_to(dns, dst_address).await {
+                                telio_log_warn!("[DNS] Failed to send DNS query response  {:?}", e)
+                            };
+                        }
+                        TunnResult::Err(e) => {
+                            telio_log_warn!(
+                                "[DNS] Failed to encapsulate DNS query response  {:?}",
+                                e
+                            )
+                        }
+                        _ => {}
+                    }
+                }
+            }
+        });
+
+        if awaiting_shutdown.blocking_recv().is_some() {
+            telio_log_debug!("Closing dns request handler");
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+                .await
+                {
+                    Ok(length) => length,
+                    Err(e) => {
+                        telio_log_error!("[DNS] {}. Offending request packet: {:?}", e, packet);
+                        MAX_PACKET + 1
+                    }
+                };
+
+                if length < MAX_PACKET {
```
stalowyjez, line 205: I'd add some else to that, at least that some malformed packet arrived

stalowyjez: Or, if this is just a way of passing info that sth failed, maybe some error would be better? Or just moving this to that match arm?

crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+    /// Handle processing dns request and sending to
+    /// hickory-server to avoid getting stuck due to
+    /// unresponsivenes from hickory
+    fn handle_dns_request(
+        socket: Arc<UdpSocket>,
+        nameserver: Arc<RwLock<LocalNameServer>>,
+        peer: Arc<Mutex<Tunn>>,
+        dst_address: SocketAddr,
+        mut rx: Receiver<Vec<u8>>,
+        mut awaiting_shutdown: Receiver<()>,
```
stalowyjez, line 183: Maaybe we could merge this channels with some `Shutdown / Request {request}` enum?

crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+                let length = match LocalNameServer::process_packet(
+                    nameserver,
+                    &packet,
+                    &mut receiving_buffer,
+                )
+                .await
+                {
+                    Ok(length) => length,
+                    Err(e) => {
+                        telio_log_error!("[DNS] {}. Offending request packet: {:?}", e, packet);
```
stalowyjez, line 200: Nit: Maybe this log could be a bit more consistent with others? Like "DNS request failed: {error} "?

crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+        mut rx: Receiver<Vec<u8>>,
+        mut awaiting_shutdown: Receiver<()>,
+    ) {
+        tokio::task::spawn(async move {
+            while let Some(packet) = rx.recv().await {
+                telio_log_debug!("Handling dns request");
+                let mut receiving_buffer = vec![0u8; MAX_PACKET];
+                let mut sending_buffer = vec![0u8; MAX_PACKET];
+                let nameserver = nameserver.clone();
+                let length = match LocalNameServer::process_packet(
```
stalowyjez, line 191: We could rename it, imho `process_packet` doesn't make it clear that it is about handling DNS request

crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+
+                if length < MAX_PACKET {
+                    let tunn_res = peer.lock().await.encapsulate(
+                        receiving_buffer.get(..length).unwrap_or(&[]),
+                        &mut sending_buffer,
+                    );
+                    match tunn_res {
+                        TunnResult::WriteToNetwork(dns) => {
+                            if let Err(e) = socket.send_to(dns, dst_address).await {
+                                telio_log_warn!("[DNS] Failed to send DNS query response  {:?}", e)
```
stalowyjez, line 213: Same, it's kinda weird that some logs has this `[DNS]` prefix and some not

crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+    fn handle_dns_request(
+        socket: Arc<UdpSocket>,
+        nameserver: Arc<RwLock<LocalNameServer>>,
+        peer: Arc<Mutex<Tunn>>,
+        dst_address: SocketAddr,
+        mut rx: Receiver<Vec<u8>>,
+        mut awaiting_shutdown: Receiver<()>,
+    ) {
+        tokio::task::spawn(async move {
+            while let Some(packet) = rx.recv().await {
```
stalowyjez, line 186: Or maybe it is enough to just close sender and this loop will just stop and this shutdown channel is just unnecesary?

crates/telio-dns/src/nameserver.rs
Hasan6979: This comes from legacy, I guess back when we didn't have tracing. But it can be cleaned up 👍

crates/telio-dns/src/nameserver.rs
Hasan6979: I would prefer it to be separate channels. The channels would have to be mpmc, that means that at every command both of them would be woken up and will check the data on the channel. DNS requests come often, so it would be a bit wasteful

crates/telio-dns/src/nameserver.rs
Hasan6979: I think this should work.

crates/telio-dns/src/nameserver.rs
mathiaspeters: I think I disagree with this. If there is a `process_packet` being called on a `NameServer` object in a function called `handle_dns_request`, I think it's clear enough that it handles DNS requests. But either way, I think this is out of scope of this PR

crates/telio-dns/src/nameserver.rs
```diff
@@ -666,20 +695,39 @@ impl NameServer for Arc<RwLock<LocalNameServer>> {
...
+        tokio::task::spawn_blocking(move || {
+            LocalNameServer::handle_dns_request(
+                socket,
+                nameserver,
+                peer,
+                dst_address,
+                rx,
+                shutdown_rx,
+            )
+        });
```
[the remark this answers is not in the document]
mathiaspeters: It looks to me like they're separate threads

crates/telio-dns/src/nameserver.rs
```diff
@@ -666,20 +695,39 @@ impl NameServer for Arc<RwLock<LocalNameServer>> {
...
 
-    #[allow(unwrap_check)]
     async fn start(&self, peer: Arc<Mutex<Tunn>>, socket: Arc<UdpSocket>, dst_address: SocketAddr) {
         let nameserver = self.clone();
-        self.write().await.task_handle = Some(tokio::spawn(LocalNameServer::dns_service(
-            peer,
-            nameserver,
-            socket,
+        let (tx, rx) = telio_utils::sync::mpsc::channel(5);
+        let (shutdown_tx, shutdown_rx) = telio_utils::sync::mpsc::channel(1);
```
mathiaspeters, line 712: Looks like this could be a `oneshot` channel

Hasan6979: That is what I had tried earlier. Borrow checker complains (here) [https://github.com/NordSecurity/libtelio/pull/1086/files#diff-b0e6219b0aeb20e202364d2c93a242f125ea2023535246e86174975122963205R703] because can't move out of reference as OneShot send takes ownership of channel.

crates/telio-dns/src/nameserver.rs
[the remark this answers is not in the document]
stalowyjez: Ok, could you add some comment then, for me and the others who weren't on that natlab meeting and may feel confused with this task matryoshka? ^^'

crates/telio-dns/src/nameserver.rs
stalowyjez: The link doesn't work for me unfortunately :/

crates/telio-dns/src/nameserver.rs
```diff
@@ -201,6 +171,65 @@ impl LocalNameServer {
...
+                let mut sending_buffer = vec![0u8; MAX_PACKET];
+                let nameserver = nameserver.clone();
+                let length = match LocalNameServer::process_packet(
+                    nameserver,
+                    &packet,
+                    &mut receiving_buffer,
+                )
+                .await
+                {
+                    Ok(length) => length,
```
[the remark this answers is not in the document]
Hasan6979: Done ✅

crates/telio-dns/src/nameserver.rs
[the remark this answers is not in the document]
mathiaspeters: A comment is probably a good idea but I wouldn't call this nested tasks. First one task is created for running the dns service and then a separate task is started for processing DNS requests

crates/telio-dns/src/nameserver.rs
mathiaspeters: Link doesn't work for me either, but I don't think it's too important. This PR, if I understand correctly, is mostly exploration and will be cleaned up or changed depending on the results, so for me it's good like it is for now

mathiaspeters:
+1

crates/telio-dns/src/nameserver.rs
Hasan6979: Done ✅

crates/telio-dns/src/nameserver.rs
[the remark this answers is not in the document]
Hasan6979: I had added description about `handle_dns_request` on line 174

crates/telio-dns/src/nameserver.rs
Hasan6979: That might be because code was changed after that. I was pointing to https://github.com/NordSecurity/libtelio/pull/1086/files#diff-b0e6219b0aeb20e202364d2c93a242f125ea2023535246e86174975122963205R697

in case this also doesn't work. It on line 697 of `nameserver.rs`

crates/telio-dns/src/nameserver.rs
```diff
@@ -149,55 +155,72 @@ impl LocalNameServer {
...
+                telio_log_debug!("Handling dns request");
+                let mut receiving_buffer = vec![0u8; MAX_PACKET];
+                let mut sending_buffer = vec![0u8; MAX_PACKET];
+                let nameserver = nameserver.clone();
+                match LocalNameServer::process_dns_request(
+                    nameserver,
+                    &packet,
+                    &mut receiving_buffer,
+                )
+                .await
```
stalowyjez, line 196: Ok, I misunderstood that a bit - so the real problem is possible hang on this await, isn't it? Maybe we could select on this and shutdown channel here? Or maybe the solution should be even different - In this scenario after one hang it won't resolve any requests more - maybe better idea would be to first receive a message from the channel and then spawn a task resolving it? And in such case you won't wait for the server in the loop, so you could just end it receiving `None`?

crates/telio-dns/src/nameserver.rs
Hasan6979: ✅

crates/telio-dns/src/nameserver.rs
stalowyjez: I mean, Tokio async tasks are quite lightweight, so maybe it's not a problem to just spawn them while serializing DNS queries sounds a bit inefficient to me and the code would be probably a bit cleaner, without all of this channels

Hasan6979: Spawning a task everytime a request comes isn't a good idea because DNS requests comes a lot.

stalowyjez: Hmm, so if there are a lot of them handling them all in one loop IS a good idea? This sounds like a heavily I/O bound job, so I guess spawning more tasks for them sounds like a necessity

Hasan6979: Handling them in a single loop itself isn't a problem, i,e, handling them sequentially.

Hasan6979: Spawning more tasks may not be the best idea because firstly we don't know how many more tasks should we spawn. Secondly, this whole change is a bit experimental and might be reverted as it may not solve anything at all.

stalowyjez: Hmm, I cannot say that I'm convinced. Albo, I don't like that idea that emerged recently that `main` branch is a playground to test your ideas 😅

crates/telio-dns/src/nameserver.rs
```diff
@@ -149,55 +155,72 @@ impl LocalNameServer {
...
-                                    e,
-                                    packet
-                                );
+                                telio_log_debug!("Error semaphore.try_acquire()");
                                 return;
                             }
                         };
+                        #[allow(index_access_check)]
+                        receiving_buffer[..packet.len()].copy_from_slice(packet);
+                        if let Err(e) = tx.send(receiving_buffer).await {
```
stalowyjez, line 164: Hmm if loop is blocked on the server won't this send be albo blocking as that thread is not receiving while blocked?

stalowyjez:
Ok, so I'm leaving only these two threads which I find a bit puzzling about this PR - if you find them irrelevant, you can just resolve them, but I think someone else should have a look at this.

crates/telio-dns/src/nameserver.rs
Hasan6979: It would send in the channel and move on, wouldn't it ?

crates/telio-dns/src/nameserver.rs
Hasan6979: It was decided that DNS getting blocked issue is rare and difficult to reproduce and quite time consuming to get to the root cause.
Threads were getting parked, and they shouldn't. Hence for now a hotfix was decided to move this part into a separate non-blocking task so that other things dont get blocked.

Hasan6979:
Closing this PR because after some discussion, this won't really solve anything and we need to get to the root cause of the problem.

PR #1089, opened 2025-01-29
Fix dod status on merge queue

jjanowsk:
Seems that skipped status is not enough, we need success status for check-dod job. So let's execute checkout always and dod step only when we should.

- [x] README.md is updated


LukasPukenis (commented):
+1

PR #1092, opened 2025-01-30
Wait for interface to report Connected state

LukasPukenis:
Windows tests fail sometimes by not reporting any
IP which is set without errors. Experimentation
revealed that the device might not have been
in a `Connected` state, so we enforce that.

### Problem

### Solution

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/router/windows_router.py
+++ b/nat-lab/tests/utils/router/windows_router.py
@@ -23,8 +23,32 @@ class WindowsRouter(Router):
         return self._interface_name
 
     async def setup_interface(self, addresses: List[str]) -> None:
+        assert self._interface_name
+
         for address in addresses:
             addr_proto = self.check_ip_address(address)
+
+            cmd = CommandGrepper(
+                self._connection,
+                [
+                    "netsh",
+                    "interface",
+                    "ipv4" if addr_proto == IPProto.IPv4 else "ipv6",
+                    "show",
+                    "interface",
+                    self._interface_name,
+                ],
+                timeout=self._status_check_timeout,
+            )
+            if not await cmd.check_exists("State", ["connected"]):
+                raise RuntimeError(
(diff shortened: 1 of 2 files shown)


nat-lab/tests/utils/router/windows_router.py
```diff
@@ -23,9 +23,43 @@ def get_interface_name(self) -> str:
...
+                    "ipv4" if addr_proto == IPProto.IPv4 else "ipv6",
+                    "show",
+                    "interface",
+                    self._interface_name,
+                ],
+                timeout=self._status_check_timeout,
+            ).check_exists("State", ["connected"]):
+                raise Exception("Interface didn't provide connected state")
+
+            await self._connection.create_process(
```
mathiaspeters, line 52: Instead of calling the same command twice, which makes this call site pretty crowded and potentially hard to read, I would add a `_last_stderr` field to the `CommandGrepper` and then here you get the last stdout and last stderr from the `CommandGrepper` and print whichever one is not empty

LukasPukenis: This is a good observation and if you feel strong about it I can refactor it :+1: 

My POV is that these two things are separate ones - we want the information to be printed, hence the second execution of the command which would stay regardless if this fixes the issue or not but the `CommandGrepper` instance is for ensuring we reach `connected` state which is not confirmed to be needed yet, so in my POV these just look the same, but have different purposes.

mathiaspeters: If we end up keeping these around I would want the refactoring to happen, but you raise a good point that for now it's mostly an experiment. Could you create a ticket for refactoring it and adding a `# TODO(<ticket number>)` here?

LukasPukenis: I've refactored it a bit - logging stdout everytime

mathiaspeters (approved):
+2

PR #1098, opened 2025-02-03
Add linux native adapter in dns tests

Hasan6979:
### Problem
Dns tests sometimes fail. It might be possible that the issue is not on our end, but on neptun's end.

### Solution
To test out that theory, adding linux native adaptor in the dns tests as well. When the tests fail again, and if they fail only on neptun that confirms the theory. Otherwise back to the drawing board.

- [x] README.md is updated


lcruz99 (dismissed):
+2

lcruz99 (approved):
+2

PR #1099, opened 2025-02-03
LLT-5935: Only restart PQ if there was actually a previous handshake

mathiaspeters:
### Problem
There was a potential race condition in our PQ restart logic that could cause restart to be called erroneously before the initial handshake was performed. The decision on whether to restart PQ was done based on the time since last handshake, as reported by the wireguard interface. The problem was that wireguard will uses None to represent both the case where no handshake has been performed and the case where a connection has been rejected after 180 seconds, so our PQ restart logic would incorrectly trigger if there was a wg consolidation happening between PQ peer being added and a handshake happening

### Solution
Now there's an additional timestamp in the PQ entity to help guide if PQ should be restarted

- [x] README.md is updated

### DIFF
--- a/crates/telio-pq/src/entity.rs
+++ b/crates/telio-pq/src/entity.rs
@@ -1,11 +1,16 @@
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 
 use parking_lot::Mutex;
 
 use telio_model::features::FeaturePostQuantumVPN;
 use telio_task::io::chan;
 use telio_utils::telio_log_debug;
+
+// This constant is based on section 6.1 of the wireguard whitepaper
+// It is the amount of time that has to happen since the last handshake before a connection is abandoned
+const REJECT_AFTER_TIME: Duration = Duration::from_secs(180);
 
 struct Peer {
     pubkey: telio_crypto::PublicKey,
@@ -14,6 +19,7 @@ struct Peer {
     /// This is a key rotation task guard, its `Drop` implementation aborts the task
     _rotation_task: super::conn::ConnKeyRotation,
     keys: Option<super::Keys>,
+    last_handshake_ts: Option<Instant>,
 }
 
 pub struct Entity {
@@ -53,6 +59,7 @@ impl Entity {
                 super::Event::Handshake(addr, keys) => {
                     if peer.addr == addr {
                         peer.keys = Some(keys);
+                        peer.last_handshake_ts = Some(Instant::now());
                     }
                 }
                 super::Event::Rekey(super::Keys {
@@ -106,8 +113,20 @@ impl Entity {
     }
 
     pub async fn restart(&self) {
-        let peer = self.peer.lock().take();
+        let peer = {
+            let mut peer = self.peer.lock();
+            let should_restart = peer
+                .as_ref()
+                .and_then(|peer| peer.last_handshake_ts)
+                .is_some_and(|ts| ts.elapsed() > REJECT_AFTER_TIME);
+            if should_restart {
+                peer.take()
+            } else {
+                None
+            }
+        };
         if let Some(peer) = peer {
+            telio_log_debug!("Restarting postquantum entity");
             self.start_impl(peer.addr, peer.wg_secret, peer.pubkey)
                 .await;
         }
@@ -132,6 +151,7 @@ impl Entity {
                 &self.features,
             ),
             keys: None,
+            last_handshake_ts: None,
         });
 
         #[allow(mpsc_blocking_send)]
--- a/nat-lab/tests/test_pq.py
+++ b/nat-lab/tests/test_pq.py
@@ -1,4 +1,3 @@
-import asyncio
 import config
 import pytest
 from contextlib import AsyncExitStack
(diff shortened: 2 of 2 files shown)


nat-lab/tests/test_pq.py
```diff
@@ -546,6 +546,6 @@ async def test_pq_vpn_handshake_after_nonet() -> None:
         async with client_alpha.get_router().break_udp_conn_to_host(
             str(config.NLX_SERVER["ipv4"])
         ):
-            await asyncio.sleep(195)
+            await asyncio.sleep(210)
```
mathiaspeters, line 549: I had an issue locally with the slightly shorter sleep here so I increased it a bit to make sure it's a good regression test. There is an in-progress ticket to get rid of the sleep but it's an exploration at this point so I don't want to put a `TODO` for it in the code here

crates/telio-pq/src/entity.rs
```diff
@@ -106,10 +109,16 @@ impl Entity {
     }
 
     pub async fn restart(&self) {
-        let peer = self.peer.lock().take();
-        if let Some(peer) = peer {
-            self.start_impl(peer.addr, peer.wg_secret, peer.pubkey)
-                .await;
+        let should_restart = self.peer.lock().as_ref().is_some_and(|peer| {
+            peer.last_handshake_ts
+                .is_some_and(|ts| Instant::now().duration_since(ts).as_secs() > 180)
```
LukasPukenis, line 114: We could use WG constants(or reuse if present in the codebase). WireGuard#6.1 `REJECT-AFTER-TIME` states it's 180s :)

nat-lab/tests/test_pq.py
```diff
@@ -205,7 +205,7 @@ async def test_pq_vpn_connection(
                 connection_tracker_config=generate_connection_tracker_config(
                     ConnectionTag.WINDOWS_VM_1,
                     stun_limits=(1, 1),
-                    nlx_1_limits=(1, 3),
+                    nlx_1_limits=(1, 2),
```
LukasPukenis, line 208: question - why was it 3 and now 2? :thinking:

LukasPukenis (commented):
Pretty elegant fix :+1: 

Looking at it though raised me an idea that for this specific case, a unit test could be more optimal. We can fast-forward time and either expect a new handshake or not and with or without the preshared key.

This however doesn't block the fix itself which looks correct

+1.5

crates/telio-pq/src/entity.rs
tomaszklak: Also, I think that `Instant::now().duration_since(ts)` is just `ts.elapsed()`.

crates/telio-pq/src/entity.rs
```diff
@@ -106,10 +109,16 @@ impl Entity {
     }
 
     pub async fn restart(&self) {
-        let peer = self.peer.lock().take();
-        if let Some(peer) = peer {
-            self.start_impl(peer.addr, peer.wg_secret, peer.pubkey)
-                .await;
+        let should_restart = self.peer.lock().as_ref().is_some_and(|peer| {
```
tomaszklak, line 112: nit: There is another `peer.lock()` call inside the if below. I think it will be simpler to merge them into one `lock` call.

nat-lab/tests/test_pq.py
tomaszklak: I'm not sure what is the in-progress ticket you are referring to here but maybe it would be simpler to add a debug log in the restart case, and wait for it here instead of the sleep?

tomaszklak (approved):
+1

nat-lab/tests/test_pq.py
mathiaspeters: It was increased from 2 to 3 with the initial bug fix so we could merge that and roll it out, now I'm just restoring these to what they were before

LukasPukenis: I just realized these are limits so it makes sense that they were passing with 3 and now with 2 :)

crates/telio-pq/src/entity.rs
mathiaspeters: I added a constant for this, I'm not sure if it's what you were suggesting :)

mathiaspeters:
> Pretty elegant fix 👍
> 
> Looking at it though raised me an idea that for this specific case, a unit test could be more optimal. We can fast-forward time and either expect a new handshake or not and with or without the preshared key.
> 
> This however doesn't block the fix itself which looks correct
> 
> +1.5

Unit test is a great idea and I will take it into consideration for the follow-up PR trying to remove the sleep from this test

mathiaspeters:
@tomaszklak For some reason I can't respond to two of your comments directly, but I have addressed all the points you brought up

nat-lab/tests/test_pq.py
mathiaspeters: Good idea! Added

crates/telio-pq/src/entity.rs
mathiaspeters: And used `ts.elapsed()`

LukasPukenis (approved):

PR #1103, opened 2025-02-04
Do not run our CI on push to branch

jjanowsk:
While implementing the merge queue I've discovered that our CI has a bug. This is caused by jobs marked as skipped. When we have a workflow that is triggerd on both push and pull_reqeust events and then the trigger job that is responsible for cancelling duplicated CI runs on our internal pull requests, GH is treating cancelled job as successful in its 'requred checks' logic which protects branches from merging PRs with failing CI. So in the situation where we push a change that fails the skipped jobs are reported first and the actual runs are reported few minutes later when they actually finish. This creates the window during which the checks are considered successful.
There is also a second bug. check-all-green job was not marked to run 'always' This is a problem because in case any of its needed job fails this job is skipped. And as meantioned earlier skipped job is considered successful.
The first bug could be fixed by setting a commit status in the check-all-green job and then using it as the 'required check' similarly as we use GitLab commit status to signal GitLab pipeline status. But that complicates the CI even further and I think it is just easier and more error prone to not run the CI on the pushes to developers branches. This has a drawback because now you need to open a PR to run a CI but since most branches anyway end up in the PR this is a small cost to pay for a more robust CI IMO.
Funnily enough skipped jobs are NOT considered successful when run in the merge_group context so thats why dod and changelog jobs must actually run checkout even though they do not do anything useful at this stage.

- [x] README.md is updated

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -2,6 +2,8 @@ name: Definition of Done
 on:
   pull_request:
     types: [opened, synchronize]
+  merge_group:
+    types: checks_requested
 permissions: {}
 
 jobs:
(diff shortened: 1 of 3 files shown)


tomaszklak (commented):
+1

stalowyjez (commented):
Looks good, +1

PR #1104, opened 2025-02-04
More Pyro logs around event logging

LukasPukenis:
We observe that sometimes Pyro delays events reported but still reports them and we suspect something around communication between natlab and the node itself
which is done with Pyro.

This commit adds more information to aid debugging

### Problem

### Solution

- [x] README.md is updated


Hasan6979 (approved):
+2

PR #1107, opened 2025-02-05
nat-lab investigation

jjanowsk:
1. Print nslookup output in nat-lab
2. Increase timeout of nslookups that fail so we can see if at some point
they get unstuck but fail the test manually if the lookup took more then
the second so we can still see those failures in the nightly runs.

- [x] README.md is updated


LukasPukenis:
+1

Hasan6979 (approved):
+2

PR #1110, opened 2025-02-06
Remove timeout when waiting for disconnected event

LukasPukenis:
Timeouts can be undeterministic, especially when we use VMs, since they are not as lightweight and might cause delays.

In test_event_content_meshnet case we observed constantly that Windows peer fails to receive the event in a timely manner. It is delayed by several seconds sometimes.

Why is that it's not entirely clear, and packet capture is not present on Windows at this point. Maybe it's due to same interface being used for libtelio and natlab and TCP retransmissions are to blame for retries that might cause delays.

As an easy solution, the timeout is being removed to see if it helps.

### Problem

### Solution

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_events.py
+++ b/nat-lab/tests/test_events.py
@@ -186,12 +186,8 @@ async def test_event_content_meshnet(
 
         await client_alpha.set_meshnet_config(api.get_meshnet_config(alpha.id))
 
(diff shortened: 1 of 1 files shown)


mathiaspeters (approved):
+2

PR #1111, opened 2025-02-06
Add logs around DNS zone locks to pinpoint possible contention

mathiaspeters:
### Problem
We have found that lock contention around DNS zones might be the culprit of some failing DNS tests but our current logs aren't providing enough information to be sure that is the case

### Solution
Add logs to help pinpoint said contention


crates/telio-dns/src/nameserver.rs
```diff
@@ -640,7 +642,14 @@ trait WithZones {
 #[async_trait]
 impl WithZones for Arc<RwLock<LocalNameServer>> {
     async fn zones(&self) -> Arc<ClonableZones> {
-        self.read().await.zones.clone()
+        let start = std::time::Instant::now();
+        telio_log_debug!("Aquiring read lock for DNS zones at");
+        let zones = { self.read().await.zones.clone() };
```
LukasPukenis, line 647: What are the benefits of moving to a closure?

crates/telio-dns/src/nameserver.rs
```diff
@@ -640,7 +642,14 @@ trait WithZones {
 #[async_trait]
 impl WithZones for Arc<RwLock<LocalNameServer>> {
     async fn zones(&self) -> Arc<ClonableZones> {
-        self.read().await.zones.clone()
+        let start = std::time::Instant::now();
+        telio_log_debug!("Aquiring read lock for DNS zones at");
```
LukasPukenis, line 646: According to the message it probably was meant to embed `start` :)

LukasPukenis (commented):
+1 more logs <3

crates/telio-dns/src/nameserver.rs
mathiaspeters: Not a closure but a scoped block, and it's there so that the `drop()` of the lock is called before printing `Released read lock...`

crates/telio-dns/src/nameserver.rs
mathiaspeters: No I just forgot to remove that bit of the string :sweat_smile: I thought first of adding timestamps to each log here but that is done automatically so instead I just add total time taken to the `Released` log. but I'll cleanup the strings here to remove confusion :)

crates/telio-dns/src/nameserver.rs
LukasPukenis: :thinking: IMHO it will be dropped at the same time - once `zones` is assigned the value, in both cases. 

However this is totally a not blocker to my eyes

LukasPukenis (approved):

tomasz-grz (approved):
+1

PR #1112, opened 2025-02-06
LLT-5989: Fix PQ restart for other adapter types

mathiaspeters:
### Problem
NepTUN is behaving differently from the other wireguard implementations in that NepTUN will report time since last handshake as `None` after 180 seconds (wireguard `REJECT-AFTER-TIME` value) whereas the other implementations keep reporting an actual time. The fix we added before assumes that the time since last handshake is `None` after 180s and therefore doesn't work with other adapter types 

### Solution
1. Parameterize the regression test so we know it works with all adapter types
2. We added a timestamp since last handshake to the PQ entity, let's base the restart decision purely on that timestamp instead

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_pq.py
+++ b/nat-lab/tests/test_pq.py
@@ -514,19 +514,75 @@ async def test_pq_vpn_upgrade_from_non_pq(
 
 # Regression test for LLT-5884
 @pytest.mark.timeout(240)
-async def test_pq_vpn_handshake_after_nonet() -> None:
-    public_ip = "10.0.254.1"
+@pytest.mark.parametrize(
+    "setup_params, public_ip",
+    [
+        pytest.param(
+            SetupParameters(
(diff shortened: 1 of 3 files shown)


nat-lab/tests/test_pq.py
```diff
@@ -514,19 +514,79 @@ async def test_pq_vpn_upgrade_from_non_pq(
...
+            ),
+            "10.0.254.7",
+            marks=[
+                pytest.mark.windows,
+            ],
+        ),
+    ],
+)
+async def test_pq_vpn_handshake_after_nonet(
+    alpha_setup_params: SetupParameters,
```
LukasPukenis, line 582: Maybe just `setup_params` since there's only one peer :)

LukasPukenis (dismissed):
+2 This looks nice. Expanded test suite <3

LukasPukenis (approved):

PR #1113, opened 2025-02-06
Catch signals a bit earlier

stalowyjez:
### Problem
Currently when Teliod is interrupted by a signal while waiting for an identity or NC to start signal handlers are not registered yet, so they kill Teliod brutaly and it doesn't make a proper cleanup (I'm talking mainly about socket removal tbh).

### Solution
Let's start catching signals before we start client setup

### DIFF
--- a/clis/teliod/src/daemon.rs
+++ b/clis/teliod/src/daemon.rs
@@ -12,7 +12,7 @@ use telio::{
     telio_wg::AdapterType,
 };
 use tokio::{sync::mpsc, sync::oneshot, time::Duration};
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info, trace, warn};
 
 use crate::core_api::{get_meshmap as get_meshmap_from_server, init_with_api};
 use crate::ClientCmd;
@@ -192,6 +192,8 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
 
     debug!("started with config: {config:?}");
 
+    let mut signals = Signals::new([SIGHUP, SIGTERM, SIGINT, SIGQUIT])?;
+
     let socket = DaemonSocket::new(&DaemonSocket::get_ipc_socket_path()?)?;
 
     // Tx is unused here, but this channel can be used to communicate with the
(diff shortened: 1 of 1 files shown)


clis/teliod/src/daemon.rs
```diff
@@ -205,30 +207,43 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
...
-    let nc = NotificationCenter::new(&config, &identity.hw_identifier).await?;
-
-    let tx_clone = tx.clone();
+    let nc = select! {
+        nc = NotificationCenter::new(&config, &identity.hw_identifier) => nc?,
+        _ = signals.next() => {
+            warn!("Interrupted while obtaining identity - stopping");
+            return Ok(());
+        }
+    };
```
LukasPukenis, line 225: :thinking: Will it retry after the interruption?

stalowyjez: No, why? I guess we consider signals a method for stopping the daemon, so it probably shouldn't?

LukasPukenis: Nevermind, my comment doesn't make sense, got a bit confused :)

LukasPukenis (commented):
+1

clis/teliod/src/daemon.rs
```diff
@@ -205,30 +207,43 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
     // are dummy and program will not run as it expects real tokens.
     let mut identity = DeviceIdentity::default();
     if !config.authentication_token.eq(EMPTY_TOKEN) {
-        identity = init_with_api(&config.authentication_token).await?;
+        identity = select! {
```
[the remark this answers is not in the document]
stalowyjez: Signal handlers are registered on `Signals::new` - so they will be caught anyway. Then we need to be responsive to  them and it seems that there are two places which may block it, i.e. in which we wait for some packets which can take a lot of time, so I guess we should wait there for these signals too.

Maybe a bit better option would be to make some async init there, so we can do the select on this and signals? This would require some redesign, though.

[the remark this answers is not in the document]
matislovas: Yeah, maybe `spawn_blocking!` would work. But anyway, if the signal is caught after all initialization, where is handled?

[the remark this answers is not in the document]
stalowyjez: In the main loop: https://github.com/NordSecurity/libtelio/blob/519763f79e196b1e02fb1d14428f1786f5451555/clis/teliod/src/daemon.rs#L269

Tbh idk how blocking thread would help here - it seems to me that to handle it gracefully we would need to do selects inside anyway?

clis/teliod/src/daemon.rs
```diff
@@ -205,30 +207,43 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
...
+            identity = init_with_api(&config.authentication_token) => identity?,
+            _ = signals.next() => {
```
Jauler, line 212: A small nitpick, maybe there would be a good way of structuring a code to avoid duplication of `signals.next()` in select block? 

If it is achievable in some easy way - I would say it is worth it, but if there is no way without requiring significant refactoring - just mark this comment as resolved and carry on ;).

Jauler (approved):
+1.0. IMHO the logic looks good. And it is up to you to decide whether it makes sense to spend some effort on limiting dublication of singals receival or whether it is not worth it.

clis/teliod/src/daemon.rs
stalowyjez: Does it look good now?

clis/teliod/src/daemon.rs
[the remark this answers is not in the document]
stalowyjez: @matislovas what do you think about this one?

Jauler (approved):
+1.0

PR #1114, opened 2025-02-06
Parse PQ v1 errors from the server

LukasPukenis:
PQ upgrader returns errors encoded in the 5th byte of the response.
The response in that case is exactly 5 bytes long.

Server errors are mapped to an enum and printed with an error message.

This should aid in debugging PQ issues a tiny bit better

### Problem

### Solution

- [x] README.md is updated


crates/telio-pq/src/proto.rs
```diff
@@ -288,28 +321,40 @@ fn validate_get_response_udp(
...
-        return Err(io::Error::new(
+        return Err(super::Error::Io(io::Error::new(
             io::ErrorKind::InvalidData,
             "Server responded with invalid PQ handshake version",
-        ));
+        )));
+    }
+
+    // Error handling for v1
+    if PQ_PROTO_VERSION == 1 && payload.len() == 5 {
```
mathiaspeters, line 338: Can errors only happen if payload is "truncated"?

crates/telio-pq/src/entity.rs
```diff
@@ -106,8 +113,20 @@ impl Entity {
     }
 
     pub async fn restart(&self) {
-        let peer = self.peer.lock().take();
+        let peer = {
```
matislovas, line 116: Do we have a test for this?

mathiaspeters: https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/test_pq.py#L577 

he rebased onto my branch before I merged it to main, but this code is now on main

LukasPukenis: Yep, sorry for the confusion

crates/telio-pq/src/proto.rs
LukasPukenis: I've added a guard and a comment explaining the error format

mathiaspeters (commented):
+1

matislovas (dismissed):
+1

mathiaspeters (approved):

PR #1116, opened 2025-02-10
LLT-5996: Reset PQ handshake ts on successful rekey

mathiaspeters:
### Problem
PQ needs to be restarted if more than 180s pass since last handshake, but for this purpose a successful rekey is also considered a handshake, which was not taking into consideration, which meant that 180s after a connection was established, PQ would erroneously be restarted

### Solution
Reset PQ handshake timestamp on successful rekey

- [x] README.md is updated

### DIFF
--- a/crates/telio-pq/src/entity.rs
+++ b/crates/telio-pq/src/entity.rs
@@ -72,6 +72,7 @@ impl Entity {
                             // and only then update the preshared key,
                             // otherwise we're connecting to different node already
                             keys.pq_shared = pq_shared;
+                            peer.last_handshake_ts = Some(Instant::now());
                         } else {
                             telio_log_debug!(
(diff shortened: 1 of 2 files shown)


crates/telio-pq/src/entity.rs
```diff
@@ -72,6 +72,7 @@ impl Entity {
                             // and only then update the preshared key,
                             // otherwise we're connecting to different node already
                             keys.pq_shared = pq_shared;
+                            peer.last_handshake_ts = Some(Instant::now());
```
tomaszklak, line 75: Can you add a (nat-lab?) test to verify that PQ is not restarted after 180s?

mathiaspeters: Done

LukasPukenis:
+1

tomaszklak (approved):
+1

PR #1119, opened 2025-02-11
Revert back moose db timeout in natlab to 60s

LukasPukenis:
Revert back moose db timeout in natlab to 60s
Additionally, measure how long does it take to fetch moose db

### Problem

### Solution

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_lana.py
+++ b/nat-lab/tests/test_lana.py
@@ -13,6 +13,7 @@ from config import (
     DERP_TERTIARY,
 )
 from contextlib import AsyncExitStack
+from datetime import datetime
 from helpers import connectivity_stack
 from mesh_api import API, Node
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


mathiaspeters (approved):
+2

tomaszklak:
nit: I would make a separate commit for the revert to 60s

nat-lab/tests/test_lana.py
```diff
@@ -170,15 +171,21 @@ async def get_moose_db_file(
     container_backup_path: str,
     local_path: str,
 ) -> None:
+    print(f"[{datetime.now()}]: `get_moose_db_file::start")
```
tomaszklak, line 174: nit: If this kind of tracing is becoming common (I don't know!) than maybe it would be good idea to package it up as a decorator, eg:

```python
def log_function_call(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        start = datetime.now()
        print(f"[{start}] Entering {func.__name__}...")
        result = func(*args, **kwargs)
        end = datetime.now()
        print(f"[{end}] Exiting {func.__name__} with return = {result} ({end-start})")
        return result
    return wrapper
# ...

@log_function_call
async def get_moose_db_file(
# ...

```

tomaszklak (approved):
+1

nat-lab/tests/test_lana.py
LukasPukenis: Thanks for the advice, I think it might be useful in some other context as currently we're just on a hunt why once it has failed(apparently either backing up or downloading took too long).

Using the decorator if I am not mistaken would require to put every inner call to a lambda or something like that :)

PR #1120, opened 2025-02-12
Fix wait_for_log case sensitivity

mathiaspeters:
### Problem
`wait_for_log` has an option to be case insensitive (which is the default) in which case it will convert your search string to lowercase, but it always converts the log to lower case. So if you were to pass `case_insensitive=False` and a search string that's not just lowercase letters, you would never get a hit. The reason we haven't seen it happen so far is that no callsite is passing `case_insensitive=False`

### Solution
Convert log to lowercase only if  `case_insensitive` is `True`

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -927,7 +927,10 @@ class Client:
         if case_insensitive:
             what = what.lower()
         while True:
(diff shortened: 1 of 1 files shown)


lcruz99 (approved):
+2

PR #1126, opened 2025-02-14
Truncate only long nat-lab path names

tomasz-grz:
### Problem
Some nat-lab test logs get overwritten if they have similar parameter names, because long parameter names get truncated.

### Solution
If the path is too long, truncate it and append a short hash to differentiate them.


Hasan6979:
+1

tomaszklak (approved):
+1

lcruz99 (approved):
+1

PR #1128, opened 2025-02-17
Temporarily disable logs to see if dns tests stop failing

mathiaspeters:
### Problem
We have a suspicion that writing to logs is blocking a thread that causes DNS queries to sometimes be delayed. 

### Solution
Temporarily disable logs completely to see if the errors go away.

Note: This will be reverted in 2-3 days, but until then we're not going to be able to debug other test issues 

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/uniffi/libtelio_remote.py
+++ b/nat-lab/tests/uniffi/libtelio_remote.py
@@ -85,8 +85,9 @@ class LibtelioWrapper:
         self._libtelio = None
         self._event_cb = TelioEventCbImpl()
         self._logger_cb = TelioLoggerCbImpl(logfile)
-        libtelio.add_timestamps_to_logs()
-        libtelio.set_global_logger(libtelio.TelioLogLevel.DEBUG, self._logger_cb)
+        # TODO(mathiaspeters): reenable logs
+        # libtelio.add_timestamps_to_logs()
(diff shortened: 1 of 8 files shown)


nat-lab/tests/uniffi/libtelio_remote.py
```diff
@@ -85,8 +85,9 @@ def __init__(self, daemon, logfile):
         self._libtelio = None
         self._event_cb = TelioEventCbImpl()
         self._logger_cb = TelioLoggerCbImpl(logfile)
-        libtelio.add_timestamps_to_logs()
-        libtelio.set_global_logger(libtelio.TelioLogLevel.DEBUG, self._logger_cb)
+        # TODO(mathiaspeters): reenable logs
+        # libtelio.add_timestamps_to_logs()
+        # libtelio.set_global_logger(libtelio.TelioLogLevel.DEBUG, self._logger_cb)
```
LukasPukenis, line 90: :thinking: does not setting the logger bypass the logging code entirely? Just making sure :)

LukasPukenis: Not setting global logger doesn't build subscriber so our filtering code is not called it seems, thanks for explaining :)

LukasPukenis (dismissed):
+2

Hasan6979 (approved):
+2

PR #1131, opened 2025-02-20
Track slow nslookup performance

LukasPukenis:
### Problem
Please see commit message

This is meant to be present temporarily for a nightly or two

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/dns.py
+++ b/nat-lab/tests/utils/dns.py
@@ -2,7 +2,7 @@ import re
 from config import LIBTELIO_DNS_IPV4
 from datetime import datetime
 from typing import List, Optional
-from utils.connection import Connection
+from utils.connection import Connection, TargetOS
 
 
 async def query_dns(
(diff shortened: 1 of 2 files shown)


nat-lab/tests/utils/dns.py
```diff
@@ -12,7 +12,14 @@ async def query_dns(
     dns_server: Optional[str] = None,
     options: Optional[List[str]] = None,
 ) -> None:
-    args = ["nslookup"]
+    # TODO(Lukas): nslookup sometimes is slow in test_dns natlab testcases.
+    # It results in packets being emitted after a 1-3sec delay. To see why
+    # that is happening strace is used but after investigation this should be removed.
+    if connection.target_os == TargetOS.Linux:
+        args = ["strace", "-tt", "nslookup"]
```
Jauler, line 19: This will spam `report.html` a bit too much, could you redirect it to separate file?

LukasPukenis: I consider this a temporary for one or two nightlies so don't consider making it proper worth it..

Jauler:
Keeping in mind our recent discoveries with IO saturation, do we need this merged still?

LukasPukenis:
@Jauler I do not have capacity to fix this PR according to the comment but I think it would provide more information about the startup of the process(for example would expect a delay after `execve()` but of course, not necessarily.

If you do not see gain in this one I will just close it then.

PR #1133, opened 2025-02-20
Update trust-dns to a temporary version

jjanowsk:
We need even more logs from trust-dns
Here is the commit with those extra logs: https://github.com/NordSecurity/trust-dns/commit/cc95506ae1a63785d472a9a0a654191fc73aff0b


LukasPukenis (approved):
+2

PR #1137, opened 2025-02-24
[LLT-5972] Control Up/Down state of WG-NT adapter according to presence of peers.

Jauler:
### Problem
It may be beneficial to show the adapter as "disconnected" or "down" in case it is idling in libtelio. This is especially useful if one keeps it preconfigured for potential use in the future, as adapter configuration can sometimes take seconds. For this reason this PR adds a logic which makes sure that adapter is down if there are no peers requested for WG and brings it up when peers are added. 

There is RFC LLT-0089 for more details.

Now in order to finalize the PR there had to be a few changes to other places.

### Windows IP address setup
Previously we have been configuring windows adapter IP address with standard `netsh` command and later waiting until the IP address is actually set, Unfortunately due to a fact that this operation now is operating on "disconnected" adapter before any feature is turned on - the waiting would be indefinite, as until any feature is enabled, the adapter will not get its IP. This waiting was added due to a fact that we have observed that it may take seconds for the IP address to be actually set. The delay was coming from windows duplicate address detection, but duplicate address detection on tunnel interface is not useful. For this reason in this PR, the DAD was disabled and waiting for IPs removed.

### Consolidating WG port for proxy and starcast virtual peer
Previously assumption was made that WireGuard listen port never changes during runtime. While this assumption is mostly true, it is a bit dangerous, and out right broken in case we are starting with disabled adapter. For this reason in this PR a runtime consolidation step to align listen port used by `telio-starcast` and `telio-proxy` is added on wg_controller consolidation cycle.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/adapter/windows_native_wg.rs
+++ b/crates/telio-wg/src/adapter/windows_native_wg.rs
@@ -10,14 +10,14 @@ use sha2::{Digest, Sha256};
 use std::io::Error as IOError;
 use std::slice::Windows;
 use std::sync::{Arc, Mutex, MutexGuard};
-use std::thread::sleep;
 use std::time::Duration;
 use std::{mem, option, result};
 use telio_utils::{
     telio_log_debug, telio_log_error, telio_log_info, telio_log_trace, telio_log_warn,
 };
+use tokio::time::sleep;
 #[cfg(windows)]
-use wireguard_nt::{self, set_logger, SetInterface, SetPeer};
+use wireguard_nt::{self, set_logger, SetInterface, SetPeer, WIREGUARD_STATE_UP};
 use wireguard_uapi::xplatform;
 
 /// Telio wrapper around wireguard-nt
@@ -42,6 +42,12 @@ pub struct WindowsNativeWg {
 pub enum Error {
     #[error("wireguard-nt: {0}")]
     Fail(String),
+}
+
+#[derive(Debug, PartialEq)]
+enum AdapterState {
+    Up,
+    Down,
 }
 
 // Windows native associated functions
@@ -159,19 +165,7 @@ impl WindowsNativeWg {
             ))));
         }
 
-        let mut os_error = IOError::from_raw_os_error(0);
-        for _ in 0..5 {
-            if wg_dev.adapter.up() {
-                telio_log_debug!("Adapter state set to up");
-                return Ok(wg_dev);
-            }
-            os_error = IOError::last_os_error();
-            telio_log_warn!("Failed to set adapter state to up, last error: {os_error:?}");
-            sleep(Duration::from_millis(200));
-        }
-        Err(AdapterError::WindowsNativeWg(Error::Fail(format!(
-            "Failed to set adapter's state to up, last error: {os_error:?}",
-        ))))
+        Ok(wg_dev)
     }
 
     fn get_config_uapi(&self) -> Response {
@@ -185,6 +179,72 @@ impl WindowsNativeWg {
                 interface: None,
             },
         }
+    }
+
+    /// Ensure adapter state (up/down)
+    ///
+    /// `ensure_adapter_state` attempts to bring interface up (requested_state == true) if
+    /// it is requested to be up and is not already up. Or bring it down if it is requested
+    /// to be down, but adapter is not yet down.
+    ///
+    /// As one may expect - this function is idempotent.
+    async fn ensure_adapter_state(
+        &self,
+        want_adapter_state: AdapterState,
+    ) -> Result<(), AdapterError> {
+        // We have seen a few cases where single attempt to bring
+        // The adapter up caused some issues. Especially if performed
+        // Early after driver installation.
+        // Exact root cause is not yet found, therefore lets attempt to bring
+        // The interface up for slightly longer before declaring error.
+        //
+        // See LLT-5443 for more details
+        let mut os_error = IOError::from_raw_os_error(0);
+        for _ in 0..5 {
+            // Retrieve up/down state of the adapter
+            let have_adapter_state = match self.adapter.get_adapter_state() {
+                Ok(state) => {
+                    if state == WIREGUARD_STATE_UP {
+                        AdapterState::Up
+                    } else {
+                        AdapterState::Down
+                    }
+                }
+                Err(_) => {
+                    os_error = IOError::last_os_error();
+                    telio_log_warn!("Failed to get adapter state, last error: {os_error:?}");
+                    continue;
+                }
+            };
+
+            // Make this function idempotent
+            if want_adapter_state == have_adapter_state {
+                return Ok(());
+            }
+
+            // The wireguard-nt-rust-wrapper here indicates success using
+            // bool for `.up()` or `.down()` functions
+            let success = match want_adapter_state {
+                AdapterState::Up => self.adapter.up(),
+                AdapterState::Down => self.adapter.down(),
+            };
+
+            // Terminate if we succeeded
+            if success {
+                return Ok(());
+            }
+
+            // Continue if we failed
+            os_error = IOError::last_os_error();
+            telio_log_warn!(
+                "Failed to set adapter state to {want_adapter_state:?}, last error: {os_error:?}"
+            );
+            sleep(Duration::from_millis(200)).await;
+        }
+        telio_log_error!("Failed to set adapter state for 5 times. Giving up!");
+        Err(AdapterError::WindowsNativeWg(Error::Fail(format!(
+            "Failed to set adapter's state to {want_adapter_state:?}, last error: {os_error:?}",
+        ))))
     }
 
     fn cleanup(&self) {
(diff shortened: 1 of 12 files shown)


ThrasherLT:
+1.0

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -187,6 +175,60 @@ impl WindowsNativeWg {
...
+    /// `ensure_adapter_state` attempts to bring interface up (requested_state == true) if
+    /// it is requested to be up and is not already up. Or bring it down if it is requested
+    /// to be down, but adapter is not yet down.
+    ///
+    /// As one may expect - this function is idempotent.
+    async fn ensure_adapter_state(&self, requested_state: bool) -> Result<(), AdapterError> {
+        // We have seen a few cases where single attempt to bring
+        // The adapter up caused some issues. Especially if performed
+        // Early after driver installation.
+        // Exact root cause is not yet found, therefore lets attempt to bring
```
lcruz99, line 189: Is there any ticket for this investigation/issue? If so, please add the ticket # here

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -187,6 +175,60 @@ impl WindowsNativeWg {
...
+        // The interface up for slightly longer before declaring error.
+        let state_str = if requested_state { "up" } else { "down" };
+        let mut os_error = IOError::from_raw_os_error(0);
+        for _ in 0..5 {
+            // Retreive up/down state of the adapter
```
lcruz99, line 194: nit: Retreive -> Retrieve

src/device/wg_controller.rs
```diff
@@ -183,6 +190,27 @@ async fn consolidate_wg_private_key<W: WireGuard>(
     Ok(())
 }
 
+async fn consolidate_wg_listen_port<W: WireGuard, P: Proxy>(
+    _: &RequestedState,
```
mathiaspeters, line 194: Is there are reason this is here when it's unused?

lcruz99:
Should we test the interface state/config when there's no peers?

src/device/wg_controller.rs
```diff
@@ -183,6 +190,27 @@ async fn consolidate_wg_private_key<W: WireGuard>(
...
+    wireguard_interface: &W,
+    proxy: Option<&P>,
+    starcast_vpeer: Option<&Arc<StarcastPeer>>,
+) -> Result {
+    if let Some(proxy) = proxy {
+        let listen_port = wireguard_interface.get_interface().await?.listen_port;
+        let wg_addr = listen_port.map(|p| (Ipv4Addr::LOCALHOST, p).into());
+        proxy.set_wg_address(wg_addr).await?;
```
LukasPukenis, line 202: Question - if the adapter is `Down` does it report any `Listen Port` via UAPI? I assume it's not there but want to be sure :thinking:

src/device/wg_controller.rs
```diff
@@ -183,6 +190,27 @@ async fn consolidate_wg_private_key<W: WireGuard>(
...
+    }
+
+    if let Some(starcast_vpeer) = starcast_vpeer {
+        let listen_port = wireguard_interface.get_interface().await?.listen_port;
+        let wg_addr = listen_port.map(|p| (Ipv4Addr::LOCALHOST, p).into());
+        starcast_vpeer.set_wg_address(wg_addr).await?;
+    }
```
LukasPukenis, line 209: This also would be great to have unit-tested and that shouldn't be too difficult given that all relevant parts are already interfaces. If possible, I would like to have this unit tested or at least natlab-tested (whichever is easier)

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -208,22 +250,41 @@ impl Adapter for WindowsNativeWg {
...
+                            interface: None,
+                        }),
+                        None,
+                    ),
+                };
 
-                    Ok(self.get_config_uapi())
+                // If all of the peers has been removed -> bring the adapter down
+                if peer_cnt.map(|p| p == 0).unwrap_or(false) {
+                    self.ensure_adapter_state(false).await?;
```
LukasPukenis, line 283: Opinion: it makes sense that if the source of the truth - peer_cnt is present we use that to set the state and if not we're left with `set_cfg.peers` however I think it would be clearer to do that once, at the end of the function which would clearly show the state machine and reasoning as it might be easy to miss one of the two `ensure_adapter_state()`.

Nitpick: Could we add enum(Down, Up) instead of bools :bow:

src/device/wg_controller.rs
Jauler: The WG-NT behavior is rather strange, it reports a listen port, but it changes after going `Up`.

This is why I decided to add WG listen port consolidation in `wg_controller` such that we would not just assume that WG listen port never changes.

src/device/wg_controller.rs
Jauler: This was in-fact natlab finding :stuck_out_tongue: So it is being tested. But I missed direct testcase for the interface control itself, although I remember writing it.

I will add that one.

crates/telio-wg/src/adapter/windows_native_wg.rs
Jauler: >  however I think it would be clearer to do that once, at the end of the function which would clearly show the state machine and reasoning as it might be easy to miss one of the two ensure_adapter_state()

One must set the interface up before configuring peers. And in a mirroring fashion - it is considered "clean" to remove all of them before bringing interface down (in other words, any peer lifetime should be fully covered with interface being up). This is why `Down->Up` transition is handled before executing `Set`, and `Up->Down` after executing `Set` command

Jauler: > Nitpick: Could we add enum(Down, Up) instead of bools 🙇

That is a good one, Will do.

crates/telio-wg/src/adapter/windows_native_wg.rs
Jauler: Fixed

crates/telio-wg/src/adapter/windows_native_wg.rs
Jauler: Fixed

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -187,6 +175,60 @@ impl WindowsNativeWg {
...
     }
 
+    /// Ensure adapter state (up/down)
+    ///
+    /// `ensure_adapter_state` attempts to bring interface up (requested_state == true) if
+    /// it is requested to be up and is not already up. Or bring it down if it is requested
+    /// to be down, but adapter is not yet down.
+    ///
+    /// As one may expect - this function is idempotent.
+    async fn ensure_adapter_state(&self, requested_state: bool) -> Result<(), AdapterError> {
```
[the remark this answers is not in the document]
Jauler: I have introduced `AdapterState` enum, which have members `Up` and `Down` instead of a `bool`. Also changed `requested/actual` to `want/have`.

Jauler:
> Should we test the interface state/config when there's no peers?

I will add a few nat-lab tests. It make sense.

src/device/wg_controller.rs
LukasPukenis: Thanks for explaining. It's worth mentioning that on a comment in case other people have the same question in the future :bow:

crates/telio-wg/src/adapter/windows_native_wg.rs
LukasPukenis: > One must set the interface up before configuring peers. And in a mirroring fashion - it is considered "clean" to remove all of them before bringing interface down (in other words, any peer lifetime should be fully covered with interface being up). This is why Down->Up transition is handled before executing Set, and Up->Down after executing Set command

Could it become a comment so it wouldn't be accidentally _refactored_ into what I suggested? :D

Jauler: TBH this comes to a point where I think it is too cluttery to comment such thing. Eg. its like commenting every `.start()` function in our crates saying that you must first start the module before you can configure it...

src/device/wg_controller.rs
```diff
@@ -183,6 +189,45 @@ async fn consolidate_wg_private_key<W: WireGuard>(
...
+    // down->up transition, listen-port may change. Therefore this consolidation has been added to
+    // keep it in sync all the time.
+
+    if let (None, None) = (proxy, starcast_vpeer) {
+        return Ok(());
+    }
+
+    // Determine WireGuard port. Prefer listen_proxy_port if it is set, otherwise - use
+    // listen_port.
+    // This is only required due to pecuiarities of WireGuard-go bind. Other WireGuard
```
mathiaspeters, line 208: Nit: peculiarities

mathiaspeters: And a question: when we eventually get rid of wireguard-go as a supported adapter type, can we simplify this to just use `listen_port`?

nat-lab/tests/test_adapter.py
```diff
@@ -139,3 +169,121 @@ async def test_adapter_service_loading(
 
     async with AsyncExitStack() as exit_stack:
         _ = await setup_mesh_nodes(exit_stack, [alpha_setup_params, beta_setup_params])
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
Jauler: I was thinking, that it would be nice to add at least one test which would go though two cycles, `down->up->down->up`, but did not want to add empty cycling (e.g. on the same scenario or so), this is why I have combined vpn and dns. Meshnet requires slightly different initial setup -> that is why it is separated.

src/device/wg_controller.rs
Jauler: yes. We can then eliminate `proxy_listen_port`.

nat-lab/tests/test_adapter.py
```diff
@@ -138,3 +169,121 @@ async def test_adapter_service_loading(
...
+        first_node_id = next(iter(api.nodes))
+        await client_alpha.set_meshnet_config(
+            api.get_meshnet_config(first_node_id, derp_servers=[config.DERP_PRIMARY])
+        )
+
+        # Verify that interface is connected after we have enabled meshnet with peers
+        state = await get_interface_state(client_conn, client_alpha)
+        assert state == AdapterState.UP
+
+        await client_alpha.set_mesh_off()
```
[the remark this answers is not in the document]
lcruz99: If I understood correctly, if the meshnet gets empty and libtelio goes idle, the interface should go down.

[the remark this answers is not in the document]
Jauler: Generally it is equivalent

lcruz99 (approved):
+1

PR #1139, opened 2025-02-25
Bump cargo dependency versions

tomasz-grz:
### Problem
Cargo dependencies are out of date

### Solution
Updated the dependencies and refactored some of the code that had braking changes.
- Migrated `clap` from version 3.x to 4.x.
- Bumped minor versions, that didn't have braking changes.

- [x] README.md is updated

### DIFF
--- a/clis/interderpcli/Cargo.toml
+++ b/clis/interderpcli/Cargo.toml
@@ -7,13 +7,10 @@ license = "GPL-3.0-only"
 repository = "https://github.com/NordSecurity/libtelio"
 
 [dependencies]
-clap = { version = "4.5.26", features = ["std", "derive"], default-features = false }
+clap.workspace = true
 ring = { default-features = false, version = "0.17.5" }
 
 anyhow.workspace = true
-base64.workspace = true
-crypto_box.workspace = true
-futures.workspace = true
 itertools.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
--- a/crates/telio-network-monitors/Cargo.toml
+++ b/crates/telio-network-monitors/Cargo.toml
@@ -26,10 +26,7 @@ neli = { version = "0.6.3", features = ["async"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { workspace = true, features = ["iphlpapi", "netioapi", "winnt", "ws2def"] }
-windows = { version = "0.59.0", features = [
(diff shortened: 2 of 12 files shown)


clis/interderpcli/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "GPL-3.0-only"
 repository = "https://github.com/NordSecurity/libtelio"
 
 [dependencies]
-clap = { version = "4.5.26", features = ["std", "derive"], default-features = false }
+clap = { version = "4.5", features = ["std", "derive"], default-features = false }
```
mathiaspeters, line 10: This might be out of scope for this PR, but is there a reason we couldn't make the workspace version of `clap` match this and then use that here? Instead of manually specifying clap again

crates/telio-network-monitors/Cargo.toml
```diff
@@ -26,7 +26,7 @@ neli = { version = "0.6.3", features = ["async"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { workspace = true, features = ["iphlpapi", "netioapi", "winnt", "ws2def"] }
-windows = { version = "0.59.0", features = [
+windows = { version = "0.60.0", features = [
```
mathiaspeters, line 29: Looks like this could use a workspace dependency instead

lcruz99:
lgtm +1

LukasPukenis (commented):
+1

crates/telio-network-monitors/Cargo.toml
tomasz-grz: Good point!

clis/interderpcli/Cargo.toml
tomasz-grz: Unfortunately we are specifying `default-features = false` here which conflicts with the workspace definition. It's just a warning now but "this could become a hard error in the future".. ⚠️ 

But I also don't know, maybe we could include default features here and then use the workspace.

crates/telio-network-monitors/Cargo.toml
tomasz-grz: Updated

clis/interderpcli/Cargo.toml
tomasz-grz: Updated

mathiaspeters (approved):
+1

PR #1140, opened 2025-02-25
Fix panic if empty direct config is passed

ThrasherLT:
### Problem
libtelio panic's if an empty `Direct` config is passed, since then it passes `0` as the period for `interval_at`.

### Solution
Added a check for empty config to prevent 0 as interval and return a proper error instead.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -185,6 +185,8 @@ pub enum Error {
     TransportError(#[from] telio_starcast::transport::Error),
     #[error("Events processing thread failed to start: {0}")]
     EventsProcessingThreadStartError(std::io::Error),
+    #[error("Polling period cannot be zero")]
+    PollingPeriodZero,
 }
 
 pub type Result<T = ()> = std::result::Result<T, Error>;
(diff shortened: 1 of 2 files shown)


src/device.rs
```diff
@@ -1370,7 +1372,9 @@ impl Runtime {
                 self.requested_state.device_config.private_key.clone(),
             )));
             let mut endpoint_providers: Vec<Arc<dyn EndpointProvider>> = Vec::new();
-
+            if direct.endpoint_interval_secs == 0 {
+                return Err(Error::PollingPeriodZero);
```
tomasz-grz, line 1376: Is there is some sane default value that can be used instead of failing here?

I remember for example `"{"direct": {}}"` used to be a valid config.

tomasz-grz: Hmm in [fact it does look like the default ](https://github.com/NordSecurity/libtelio/blob/47937157b2ae9131ded960934fdb01f6b82770c4/crates/telio-model/src/features.rs#L256)is 25. In which cases would this be 0?

mathiaspeters: I think failing is the right approach here. The problem isn't really that "empty direct config" was passed, but rather that an explicit attempt was made at setting the interval to 0, which is an error.

ThrasherLT: Sure, I'll fix the comments with this PR

gytsto (commented):
+1

mathiaspeters (approved):
+1

tomasz-grz (approved):
+1

PR #1146, opened 2025-03-03
Bump libtelio-build to v2.8.7

LukasPukenis:
v2.8.7 contains audit logs disabled as they were
found to be causing a lot of journal entries thus
possibly causing high IO and as a result
flakyness in NatLab

### Problem

### Solution


tomaszklak (approved):
+2

PR #1143, opened 2025-02-27
LLT-6027: Cleanup natlab logs

lcruz99:
### Problem
During the recent debugging of several natlab issues extra logs were added, now that they are no longer needed cleanup should take place.

### Solution
Remove added logs.

### :ballot_box_with_check: Definition of Done checklist
- [x] Commit history is clean ([requirements](../blob/main/docs/git_commit_messages_requirements.md))
- [x] README.md is updated
- [x] Functionality is covered by unit or integration tests

PR #1149, opened 2025-03-04
Improve natlab logs

lcruz99:
### Problem
With https://github.com/NordSecurity/libtelio/pull/1143 the nat-lab logs were cleaned, during that task execution it was realized that the logs were still too verbose, many adding trivial information.

Nat-lab could also make use of a logger to standardize logs format and providing the user the option to choose the amount of verbosity.

### Solution
- Add a logger.
- Remove trivial logs or set them to `DEBUG` log level.
- Default log level is `INFO` (locally and CI), `DEBUG` logs are still shown on report but on a different section and saved  when `NATLAB_SAVE_LOGS` is set.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection/docker_connection.py
+++ b/nat-lab/tests/utils/connection/docker_connection.py
@@ -98,7 +98,6 @@ class DockerConnection(Connection):
     async def __aexit__(self, *exc_details):
         await self.restore_ip_tables()
         await self.clean_interface()
-        return self
 
     @classmethod
     @asynccontextmanager


nat-lab/tests/utils/command_grepper.py
```diff
@@ -89,14 +93,16 @@ async def check_not_exists(
...
         except TimeoutError as e:
-            print(f"Timeout error: {e}, last stdout: {self._last_stdout}")
+            log.error("Timeout error: %s, last stdout: %s", e, self._last_stdout)
             return False
         except Exception as e:
-            print(
-                f"Some other exception happened: {e}, last stdout: {self._last_stdout}"
+            log.error(
+                "Some other exception happened: %s, last stdout: %s",
+                e,
```
stalowyjez, line 104: Hmm, aren't python f-strings generally considered prettier than such printf-like syntax?

nat-lab/tests/conftest.py
```diff
@@ -481,6 +457,46 @@ async def collect_mac_diagnostic_reports():
...
+
+
+@pytest.fixture(autouse=True)
+def setup_logger(tmp_path, request):
+    # if os.environ.get("NATLAB_SAVE_LOGS"):
+    log_dir = get_current_test_log_path()
+    os.makedirs(log_dir, exist_ok=True)
+    log_file = os.path.join(log_dir, "debug.log")
+    # else:
+    #     log_file = os.path.join(tmp_path, f"{request.node.nodeid}.log")
```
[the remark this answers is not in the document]
lcruz99: Removed

nat-lab/tests/utils/logger.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def addFileHandler(self, handler: logging.FileHandler) -> None:
+        handler.setLevel(logging.DEBUG)
+        handler.setFormatter(
+            logging.Formatter("[%(asctime)s] %(levelname)s: %(message)s")
+        )
+        self._logger.addHandler(handler)
+
+    def removeHandler(self, handler: logging.Handler) -> None:
+        self._logger.removeHandler(handler)
```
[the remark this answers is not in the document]
lcruz99: Method removed

nat-lab/tests/utils/logger.py
```diff
@@ -0,0 +1,57 @@
...
+            console_handler = logging.StreamHandler()
+            console_handler.setLevel(level)
+            console_handler.setFormatter(
+                logging.Formatter("[%(asctime)s]: %(message)s")
+            )
+            self._logger.addHandler(console_handler)
+
+            if log_file:
+                file_handler = logging.FileHandler(log_file, mode="w")
+                self.addFileHandler(file_handler)
```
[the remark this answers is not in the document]
lcruz99: agreed, removed

nat-lab/tests/utils/logger.py
```diff
@@ -0,0 +1,57 @@
+import logging
+import threading
+from typing import Optional
+
+LOG_LEVEL = logging.INFO
+
+
+class Logger:
```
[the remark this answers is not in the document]
lcruz99: We don't have to import and call `logging.getLogger()` on every module, and having this Logger class we have a centralized definition of our Logger.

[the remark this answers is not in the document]
lcruz99: I guess an alternative could be configuring the `logging.Logger` instance on one of the pytest session hooks and then export it to other modules, but IMO this looks more tidy.

stalowyjez:
Looks good, +1

nat-lab/tests/utils/router/mac_router.py
```diff
@@ -73,6 +76,7 @@ async def deconfigure_interface(self, addresses: List[str]) -> None:
                         "prefixlen",
                         "64",
                     ],
+                    quiet=True,
```
Hasan6979, line 79: Does it make sense to have the default as quiet=True? I am not familiar, but are there more cases or quiet=False which well aren't being shown in the diff ?

nat-lab/tests/utils/traffic.py
```diff
@@ -96,7 +97,7 @@ def generate_packet_delay_histogram(
 ) -> typing.List[int]:
     """Generate histogram based on the relative packet(and packet before) timestamp differences. Good for observing bursts"""
 
-    print("Looking for a pcap at", pcap_path)
+    log.info("Looking for a pcap at %s", pcap_path)
```
Hasan6979, line 100: nit: I think this can be debug

nat-lab/tests/utils/traffic.py
```diff
@@ -134,7 +135,7 @@ def generate_packet_distribution_histogram(
 ) -> typing.List[int]:
     """Generate histogram based on absolute packet timestamps. Good for observing trends and patterns"""
 
-    print("Looking for a pcap at", pcap_path)
+    log.info("Looking for a pcap at %s", pcap_path)
```
Hasan6979, line 138: Same for this.

nat-lab/tests/utils/testing.py
```diff
@@ -31,7 +30,6 @@ def get_current_test_case_and_parameters() -> Tuple[Optional[str], Optional[str]
                 _format_path_string(test_parts[1]),
             )
         return (_format_path_string(test_parts[0]), None)
-    print(datetime.now(), "PYTEST_CURRENT_TEST is None")
```
Hasan6979, line 34: Why is the print removed from here ?

nat-lab/tests/utils/tcpdump.py
```diff
@@ -75,20 +77,18 @@ def get_stderr(self) -> str:
         return self.stderr
 
     async def on_stdout(self, output: str) -> None:
-        print(f"tcpdump: {output}")
```
Hasan6979, line 78: Why is this print removed ?

nat-lab/tests/utils/tcpdump.py
```diff
@@ -75,20 +77,18 @@ def get_stderr(self) -> str:
...
         self.stdout += output
         await self.output_notifier.handle_output(output)
 
     async def on_stderr(self, output: str) -> None:
-        print(f"tcpdump err: {output}")
```
Hasan6979, line 83: This as well ?

nat-lab/tests/utils/process/ssh_process.py
```diff
@@ -47,7 +48,7 @@ async def execute(
         privileged: bool = False,
     ) -> "SshProcess":
         if privileged:
-            print("'privileged' does nothing for ssh processes")
+            log.warning("'privileged' does nothing for ssh processes")
```
Hasan6979, line 51: nit: This can be reduced to info or even debug.

lcruz99: I disagree because the user should be warned that a functionality doesn't work as he's expecting, in this case regarding process privilege access

nat-lab/tests/utils/router/mac_router.py
lcruz99: Not sure if I understood, do you mean to make `quiet=True` as default in the create_process() functions?

I haven't thought much about it but wouldn't say there are many more `True` cases than `False`.

nat-lab/tests/utils/tcpdump.py
lcruz99: It was printing trivial logs, and the caller should decide what to do with the standard output.

nat-lab/tests/utils/tcpdump.py
lcruz99: It was printing trivial logs, in case there's an error an exception is thrown and the stderr is then printed. We shouldn't print it by default.

nat-lab/tests/utils/traffic.py
lcruz99: Agree, changed.

nat-lab/tests/utils/testing.py
lcruz99: I thought there's no reason to print it. wdyt?

nat-lab/tests/utils/traffic.py
lcruz99: Agree, changed.

nat-lab/tests/utils/router/mac_router.py
Hasan6979: Yes. If there are more places that we need quiet=True -> makes sense to make that default. Less code changes in the current PR 🤷

nat-lab/tests/utils/tcpdump.py
Hasan6979: I would recommend keeping the logs, albeit with trace.

Hasan6979 (approved):
+1

PR #1151, opened 2025-03-04
natlab: Setup ephemeral ports only when needed

lcruz99:
### Problem
Ephemeral ports setup takes about 5 seconds, it is being done on every test for every node and that's not required.

### Solution
Do the ephemeral ports setup only when a new connection is created instead.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection_util.py
+++ b/nat-lab/tests/utils/connection_util.py
@@ -1,9 +1,11 @@
 import config
+import random
 from aiodocker import Docker
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
+from datetime import datetime
 from enum import Enum, auto
-from typing import AsyncIterator, Dict, Tuple, Optional, List, Union
+from typing import AsyncIterator, Dict, Tuple, Optional, List, Union, Set
 from utils.connection import Connection, TargetOS, DockerConnection
 from utils.connection_tracker import (
     ConnTrackerEventsValidator,
@@ -175,6 +177,8 @@ LAN_ADDR_MAP_V6: Dict[ConnectionTag, str] = {
     ),
 }
 
+EPHEMERAL_SETUP_SET: Set[ConnectionTag] = set()
+
 
 @dataclass
 class ConnectionManager:
@@ -210,6 +214,19 @@ def get_uniffi_path(connection: Connection) -> str:
 
 
 @asynccontextmanager
+async def connection_setup(
+    connection: Connection, tag: ConnectionTag
+) -> AsyncIterator[Connection]:
+    await setup_ephemeral_ports(connection, tag)
+    try:
+        yield connection
+    finally:
+        if isinstance(connection, DockerConnection):
+            await connection.restore_ip_tables()
+            await connection.clean_interface()
+
+
+@asynccontextmanager
 async def new_connection_raw(
     tag: ConnectionTag,
 ) -> AsyncIterator[Connection]:
@@ -218,19 +235,18 @@ async def new_connection_raw(
             connection: Connection = await DockerConnection.new(
                 docker, container_id(tag)
             )
-            try:
-                yield connection
-            finally:
-                await connection.restore_ip_tables()
-                await connection.clean_interface()
+            async with connection_setup(connection, tag) as conn:
+                yield conn
 
     elif tag in [ConnectionTag.WINDOWS_VM_1, ConnectionTag.WINDOWS_VM_2]:
         async with windows_vm_util.new_connection(LAN_ADDR_MAP[tag]) as connection:
-            yield connection
+            async with connection_setup(connection, tag) as conn:
+                yield conn
 
     elif tag == ConnectionTag.MAC_VM:
         async with mac_vm_util.new_connection() as connection:
-            yield connection
+            async with connection_setup(connection, tag) as conn:
+                yield conn
 
     else:
         assert False, f"tag {tag} not supported"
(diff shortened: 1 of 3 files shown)


nat-lab/tests/utils/connection_util.py
```diff
@@ -219,17 +221,20 @@ async def new_connection_raw(
                 docker, container_id(tag)
             )
             try:
+                await setup_ephemeral_ports(connection, tag)
```
tomaszklak, line 224: As it is right now, if I call the `new_connection_raw` twice in the same test for the same connection, I will get two different ranges which I think is incorrect.

nat-lab/tests/utils/connection_util.py
```diff
@@ -492,3 +497,40 @@ async def remove_traffic_control_rules(connection):
...
+            f"net.inet.ip.portrange.last={start_port + num_ports}",
+        ]
+    elif connection_tag.name.startswith("DOCKER_") and "CLIENT" in connection_tag.name:
+        cmd = [
+            "sysctl",
+            "-w",
+            f"net.ipv4.ip_local_port_range={start_port} {start_port + num_ports}",
+        ]
+    else:
+        return
```
tomaszklak, line 534: this should throw

nat-lab/tests/utils/connection_util.py
lcruz99: You are right, however is really that incorrect?

nat-lab/tests/utils/connection_util.py
lcruz99: It shouldn't. Eg: if you call `new_connection_raw()` on a gw container it shouldn't setup the ephemeral ports, in that case it should only return/pass.

lcruz99: We only want to setup the port ranges for those 3 cases, the others we can ignore.

tomaszklak: I would say we don't expect that there is any other case, right? And if that is the case, it's better to loudly signal that something unexpected happened instead of silently ignoring it.

tomaszklak: Unless this now happens for all docker containers, including those which aren't clients?

nat-lab/tests/utils/connection_util.py
lcruz99: I understand.. so do we have reasons to maintain the port range between connections to the same container on each test?

I think I can make a change so that it only runs once for each test.

nat-lab/tests/utils/connection_util.py
lcruz99: > I would say we don't expect that there is any other case, right? And if that is the case, it's better to loudly signal that something unexpected happened instead of silently ignoring it.

I agree with that.

> Unless this now happens for all docker containers, including those which aren't clients?

It can happen that someone or some test creates a connection to a container that's not a win/mac VM nor a docker client, like `DOCKER_VPN_1` [here](https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/test_pmtu.py#L199).

nat-lab/tests/utils/connection_util.py
```diff
@@ -492,3 +497,40 @@ async def remove_traffic_control_rules(connection):
...
     except:
         pass
+
+
+async def setup_ephemeral_ports(connection: Connection, connection_tag: ConnectionTag):
+    async def on_output(output: str) -> None:
+        print(datetime.now(), f"[{connection_tag.name}]: {output}")
+
+    start_port = random.randint(5000, 55000)
+    num_ports = random.randint(2000, 5000)
```
LukasPukenis, line 507: Could we add the reason while we're at it for this port range and what it achieves?

nat-lab/tests/utils/connection_util.py
```diff
@@ -219,17 +221,20 @@ async def new_connection_raw(
...
                 await connection.clean_interface()
 
     elif tag in [ConnectionTag.WINDOWS_VM_1, ConnectionTag.WINDOWS_VM_2]:
         async with windows_vm_util.new_connection(LAN_ADDR_MAP[tag]) as connection:
+            await setup_ephemeral_ports(connection, tag)
             yield connection
 
     elif tag == ConnectionTag.MAC_VM:
         async with mac_vm_util.new_connection() as connection:
+            await setup_ephemeral_ports(connection, tag)
```
LukasPukenis, line 237: nitpick: the code now follows a pattern of
```
setup_ephemeral_ports()
yield
```

It's good enough but is a bit error prone, easy to remove accidentally one of the lines. Could we DRY it a bit more so those two lines would be executed for all the connections and tags before yielding?

nat-lab/tests/utils/connection_util.py
lcruz99: Honestly I don't know, maybe @matislovas can help here.

nat-lab/tests/utils/connection_util.py
lcruz99: changed :)

tomaszklak (approved):
+1

nat-lab/tests/utils/connection_util.py
LukasPukenis: I have a suspicion is below the ports Pyro uses

LukasPukenis (dismissed):

nat-lab/tests/utils/connection_util.py
matislovas: The original problem is stated in https://github.com/NordSecurity/libtelio/pull/980 . DERP is reusing ports and it sometimes clashed with assigned ssh ephemeral ports IIRC.

tomaszklak (approved):

LukasPukenis (approved):

PR #1154, opened 2025-03-10
Add VERSIONS file

LukasPukenis:
Since moose v15 and v16 are not compatible
it is crucial to communicate it properly about
release branches and their used moose version.

### Problem

### Solution

- [x] README.md is updated

### DIFF
--- a/VERSIONS
+++ b/VERSIONS
@@ -1,0 +1,4 @@
+| Release Branch    | Moose         |
+| ----------------- | ------------- |
+| release/v5.1      | v15.0.1       |
+| release/v5.2      | v16.0.1       |


matislovas:
DOD is failing

+2

matislovas (approved):

VERSIONS
```diff
@@ -0,0 +1,4 @@
+| Release Branch    | Moose         |
```
stalowyjez, line 1: As I said on Slack, I think it could be a README section, this table looks somehow lonely in this file. If it must stay in the separate file, I would add md extension, so github would render this table nicely

tomaszklak:
This information is already in `ci/env.py` and in the changelog. Do we really need 3rd place and if so, how are we going to make sure that it's correct/up-to-date?

tomaszklak:
If we want to put it in a place that the integrators of libtelio are expected to look, we can use rust-doc for this - I've prepared an example here: https://github.com/NordSecurity/libtelio/compare/main...mooose-ver-doc

LukasPukenis:
Thanks for the comments. Actually a better place for this would be confluence doc about releases. I am closing this one.

PR #1158, opened 2025-03-10
Fix assign_ip function

jjanowsk:
Instead of failing the test in case of duplicate IP just regenerate it

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/mesh_api.py
+++ b/nat-lab/tests/mesh_api.py
@@ -67,10 +67,6 @@ class MissingNodeError(NodeError):
     pass
 
 
-class AddressCollisionError(NodeError):
-    pass
-
-
 class NicknameInvalidError(NicknameError):
     pass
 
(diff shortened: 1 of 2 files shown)


mathiaspeters (commented):
+1

nat-lab/tests/mesh_api.py
```diff
@@ -244,10 +240,15 @@ def register(
...
         del self.nodes[node_id]
 
-    def assign_ip(self, node_id: str, address: str) -> None:
-        for _, node in self.nodes.items():
-            if address in node.ip_addresses:
-                raise AddressCollisionError(node.id)
+    def assign_random_ip(self, node_id: str, proto: IPProto) -> None:
+        def get_random_ip(proto: IPProto) -> str:
+            if proto == IPProto.IPv4:
+                return f"100.{random.randint(64, 127)}.{random.randint(0, 255)}.{random.randint(8, 254)}"
```
LukasPukenis, line 246: :thinking: What I didn't think of initially - what's your opinion on instead of random, making the IPs all sequential starting from 10.x.y.z and just iterating the `z` or maybe other component?

Jauler (approved):
+1.0 once linters are fixed :P

nat-lab/tests/mesh_api.py
jjanowsk: TBH I would leave the randomization in place.

PR #1159, opened 2025-03-10
Use SystemTime instead of Instant in PQ restart

jjanowsk:
Instant does not give any guarantees when it comes to whether it counts time that elapsed during the system sleep. Because we need to measure real passage of time in PQ restart logic let's use SystemTime instead.

- [x] README.md is updated

### DIFF
--- a/crates/telio-pq/src/entity.rs
+++ b/crates/telio-pq/src/entity.rs
@@ -1,6 +1,6 @@
 use std::net::SocketAddr;
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::{Duration, SystemTime};
 
 use parking_lot::Mutex;
 
@@ -19,7 +19,7 @@ struct Peer {
     /// This is a key rotation task guard, its `Drop` implementation aborts the task
     _rotation_task: super::conn::ConnKeyRotation,
     keys: Option<super::Keys>,
-    last_handshake_ts: Option<Instant>,
+    last_handshake_ts: Option<SystemTime>,
 }
 
 pub struct Entity {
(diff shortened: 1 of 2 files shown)


mathiaspeters (commented):
Were you able to reproduce the issue before adding the fix? I'm confident this will work but it would be nice to have a manual test to check it

+1

crates/telio-pq/src/entity.rs
```diff
@@ -127,7 +127,9 @@ impl Entity {
             let should_restart = peer
                 .as_ref()
                 .and_then(|peer| peer.last_handshake_ts)
-                .is_some_and(|ts| ts.elapsed() > REJECT_AFTER_TIME);
+                .and_then(|ts| ts.elapsed().ok())
```
LukasPukenis, line 130: According to https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.elapsed
```
Returns an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err) if self is later than the current system time, and the error contains how far from the current system time self is.
```

It would be very helpful to catch the error and emit a warning in case we would chase a PQ restart failure and could attribute to this :)

gytsto (approved):
+1

LukasPukenis:
I took the liberty of clicking `upload-xcframework` so I could ask the apps for quick in branch test :)

LukasPukenis:
iOS responded that they do not reproduce the PQ sleep issue with this branch :balloon:

crates/telio-pq/src/entity.rs
LukasPukenis: I just realized - my comment above is about just printing but rather it should be handling the error:
- for clock going forwards(either naturally or due to adjustments) nothing needs to be done
- for clock going backwards we need to compensate for this one here

This might happen and should be expected. This of course also means that any other `SystemTime` is suspectable to this but because this is a specific fix for PQ, lets just handle it here if we can

jjanowsk: Yes, good catch! I've also noticed it after I posted the review. As well as @tomaszklak. New version addresses this issue.

gytsto (approved):

mathiaspeters (approved):

PR #1161, opened 2025-03-11
Update rust_build_utils to 4.2.0

tomaszklak:
This brings in the change to the way post actions are run - they will now run only if at least one build artifact has changed.

### Problem

### Solution


stalowyjez (dismissed):
+2

stalowyjez (approved):

PR #1163, opened 2025-03-12
Add daily log rotation for the teliod

tomaszklak:
### Problem
Logs generated by teliod can, over time, fill up the filesystem.

### Solution
Keep the logs from last `log_file_count` (default 7) days.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/example_teliod_config.json
+++ b/clis/teliod/example_teliod_config.json
@@ -1,6 +1,6 @@
 {
     "log_level": "trace",
-    "log_file_path": "example_log_file.log",
+    "log_file_path": "./example_log_file.log",
     "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
     "interface": {
       "name": "utun10",
--- a/clis/teliod/src/cgi/api.rs
+++ b/clis/teliod/src/cgi/api.rs
@@ -278,6 +278,7 @@ mod tests {
         let mut expected_config = TeliodDaemonConfig {
             log_level: LevelFilter::DEBUG,
             log_file_path: "/path/to/log".to_owned(),
+            log_file_count: 7,
             interface: InterfaceConfig {
                 name: "eth0".to_owned(),
                 config_provider: InterfaceConfigurationProvider::Manual,
@@ -319,6 +320,7 @@ mod tests {
 
         expected_config.log_level = LevelFilter::INFO;
         expected_config.log_file_path = "/new/path/to/log".to_owned();
+        expected_config.log_file_count = 8;
         expected_config.authentication_token =
             "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb".to_owned();
         expected_config.interface = InterfaceConfig {
@@ -335,6 +337,7 @@ mod tests {
         {
             "log_level": "info",
             "log_file_path": "/new/path/to/log",
+            "log_file_count": 8,
             "authentication_token": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
             "interface": {
                 "name": "eth1",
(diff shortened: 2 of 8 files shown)


clis/teliod/src/logging.rs
```diff
@@ -0,0 +1,58 @@
...
+            "log_file_path".to_owned(),
+            e.to_string(),
+            log_file_path.to_string(),
+        )
+    })?;
+
+    let empty: &Path = Path::new("");
+
+    let (log_dir, log_file) = match (log_file_pathbuf.parent(), log_file_pathbuf.file_name()) {
+        (Some(log_dir), Some(file_name)) if log_dir != empty && file_name != empty => {
```
stalowyjez, line 30: Hmm, imho forcing user to use `./teliod.log` instead of simpler (and usually equivalent) `teliod.log` is also not the best ux.

clis/teliod/src/logging.rs
```diff
@@ -0,0 +1,58 @@
...
+                "log_file_path".to_owned(),
+                "log_file_path needs to contain both directory and file name".to_owned(),
+                log_file_path.to_owned(),
+            ))
+        }
+    };
+
+    let log_appender = Builder::new()
+        .rotation(Rotation::DAILY)
+        .filename_prefix(log_file.to_string_lossy())
```
stalowyjez, line 44: Maybe it would be a bit better ux to use suffix (i.e. ) as well? I.e. when user wants logs in `logs.txt` then maybe `logs.2025-03-13.txt`, `logs2025-03-14.txt` etc. are more welcome than `logs.txt.2025-03-13`, `logs.txt.2025-03-14` ... ? ID, it would feel just more intuitive to me.

tomaszklak: You meant infix, not suffix right? Suffix is already used now, and it's a well established standard, eg: https://www.redhat.com/en/blog/setting-logrotate

clis/teliod/src/logging.rs
tomaszklak: I the real world, you wouldn't want to use `./teliod.log` since it implicitly depends on the current working directory. In reality the user will set it to absolute path.

clis/teliod/src/cgi/api.rs
```diff
@@ -278,6 +278,7 @@ mod tests {
         let mut expected_config = TeliodDaemonConfig {
             log_level: LevelFilter::DEBUG,
             log_file_path: "/path/to/log".to_owned(),
+            log_file_count: 7,
```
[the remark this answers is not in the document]
tomaszklak: Good point! Done :)

nat-lab/data/teliod/config.json
```diff
@@ -1,6 +1,6 @@
 {
     "log_level": "trace",
-    "log_file_path": "teliod_natlab.log",
+    "log_file_path": "./teliod_natlab.log",
```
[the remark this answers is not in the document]
tomaszklak: Yes, it now contains both the directory and the path.

clis/teliod/example_teliod_config.json
```diff
@@ -1,6 +1,6 @@
 {
     "log_level": "trace",
-    "log_file_path": "example_log_file.log",
+    "log_file_path": "./example_log_file.log",
```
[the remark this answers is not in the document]
tomaszklak: Yes, it now contains both the directory and the path.

clis/teliod/src/logging.rs
stalowyjez: Hmm, I meant suffix for the extension ^^'
A date in the end of file name just looks a bit strange to me, but maybe I'm just not used to

clis/teliod/src/logging.rs
stalowyjez: That's interesting, I'm quite convinced it was relative to teliod working direcotry (i.e. `/run` or sth like this), not cwd - but in this case it doesn't make a lot of sense probably

clis/teliod/src/logging.rs
tomaszklak: I personally don't care, but I think it's better to adhere to the existing practice - so I would prefer to keep it as is.

Jauler (approved):
+1.0

PR #1175, opened 2025-03-26
Fix nat-lab test exception handling

Jauler:
### Problem
Async contexts must define `__aexit__` method, but it matters what we return out of it. According to [PEP 492](https://peps.python.org/pep-0492/#new-syntax) the `__atexit__` usage is equivalent to this

```python
mgr = (EXPR)
aexit = type(mgr).__aexit__
aenter = type(mgr).__aenter__

VAR = await aenter(mgr)
try:
    BLOCK
except:
    if not await aexit(mgr, *sys.exc_info()):
        raise
else:
    await aexit(mgr, None, None, None)
```

In case of exception -> we need to return something that would evaluate to `False`, or otherwise we will loose exceptions. Omiting `return` statement altogether - is equivalent to `return None` which evaluates to `False`.

This PR fixes `__aexit__` implementation according to that.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/adapter/windows_native_wg.rs
+++ b/crates/telio-wg/src/adapter/windows_native_wg.rs
@@ -222,6 +222,12 @@ impl WindowsNativeWg {
                 return Ok(());
             }
 
+            telio_log_debug!(
+                "Attempting to bring interface {:?}, currently it is: {:?}",
+                want_adapter_state,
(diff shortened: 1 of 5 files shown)


lcruz99 (approved):
+2

LukasPukenis (approved):

PR #1176, opened 2025-03-26
Send logs to opensearch

Hasan6979:
### Problem

### Solution

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -4,6 +4,7 @@ import glob
 import os
 import platform
 import re
+import socket
 import uuid
 import warnings
 from collections import Counter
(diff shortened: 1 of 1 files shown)


mathiaspeters:
I'm very much in favor of having the logs in opensearch, but I'm wondering if this is the right place to send the logs. Maybe to start a small discussion, and I hope others chime in as well, was it considered to do this in the CI scripts instead? At least the way I envision it, nat-lab should be as independent as possible to any of our internal infrastructure and this, while not explicitly referencing internal infrastructure, creates a dependency that I think we should avoid. Thoughts?

tomaszklak:
> I'm very much in favor of having the logs in opensearch, but I'm wondering if this is the right place to send the logs. Maybe to start a small discussion, and I hope others chime in as well, was it considered to do this in the CI scripts instead?

Like a post-processing step after nat-lab step has completed? That sound cleaner than processing and sending logs in the nat-lab itself.

mathiaspeters:
> Like a post-processing step after nat-lab step has completed?

Exactly! Nat-lab produces logs in the same directory structure as it currently does, and a CI job uploads it to opensearch

PR #1189, opened 2025-04-03
Print the time it take for tcpdump to become ready

tomaszklak:
### Problem
During today's nat-lab meeting, we analyzed a failure of a tests due to `tcpdump` never becoming ready. 
### Solution
To make it easier to analyze in future, I'm adding tracking of the time it takes `tcpdump` to print relevant message.

We also discussed increasing the time (10s) the tests wait for `tcpdump` to become ready. But analyzing the times from test CI run, it seems that 10 seconds is more than enough:

```
 Count: 1038
   Min: 0:00:00.013659
   Max: 0:00:01.110213
Median: 0:00:00.063643
  Mean: 0:00:00.065239
```


nat-lab/tests/utils/tcpdump.py
```diff
@@ -77,10 +79,12 @@ def get_stderr(self) -> str:
         return self.stderr
 
     async def on_stdout(self, output: str) -> None:
+        log.info("tcpdump: %s", output)
         self.stdout += output
         await self.output_notifier.handle_output(output)
 
     async def on_stderr(self, output: str) -> None:
+        log.debug("tcpdump err: %s", output)
```
lcruz99, line 87: Why debug and not error level?

tomaszklak: `tcpdupm` is printing various non-error things to stderr, like for example:
```
[2025-04-03 13:51:11,079] DEBUG: tcpdump err: tcpdump: data link type LINUX_SLL2
[2025-04-03 13:51:11,079] DEBUG: tcpdump err: tcpdump: listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
```
So I decided to put it to stderr.

nat-lab/tests/utils/tcpdump.py
```diff
@@ -77,10 +79,12 @@ def get_stderr(self) -> str:
         return self.stderr
 
     async def on_stdout(self, output: str) -> None:
+        log.info("tcpdump: %s", output)
```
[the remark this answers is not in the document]
tomaszklak: since we are waiting for the `listening on` print, I think it makes it important enough to capture the stdout.

lcruz99 (approved):
+2

PR #1197, opened 2025-04-08
Link state detection should be based on the oldest tx packet after latest rx packet

tomaszklak:
### Problem
The link state detection was based on the incoming (rx) packet timestamp. This is incorrect, since WireGuard start the timer with the first outgoing (tx) packet that is sent after some packet was received.

### Solution
Align the link state detection with the passive keep-alive logic in WireGuard. Additionally, `test_event_link_state_peer_goes_offline` is fixed so that the time it takes to stop the device, doesn't impact the test.


crates/telio-wg/src/link_detection.rs
```diff
@@ -348,97 +355,120 @@ mod tests {
         assert!(matches!(down.variant, StateVariant::Down));
     }
 
-    #[tokio::test(start_paused = true)]
+    #[test_log::test(tokio::test(start_paused = true))]
```
lcruz99, line 358: Can you describe why are is `test_log` crate used here and only at this test?

nat-lab/tests/test_events_link_state.py
```diff
@@ -331,9 +336,9 @@ async def test_event_link_state_peer_goes_offline(
             return_exceptions=True,
```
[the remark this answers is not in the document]
tomaszklak: This is actually debug code, that I forgot to remove 😓

crates/telio-wg/src/link_detection.rs
tomaszklak: It's used here because I needed the telio logs for debugging why it was initially failing. I didn't add it everywhere because, it's only here and in the next test that I needed logs.

dfetti (commented):
+1.0 Looks good

crates/telio-wg/src/link_detection.rs
lcruz99: should we merge it?

tomaszklak: I would merge it :)

PR #1201, opened 2025-04-10
Kani

tomaszklak:
This PR adds a automated formal proof for some of the properties of `BytesAndTimestamps`. It uses [Kani](https://github.com/model-checking/kani) model checker, which transforms rust's [MIR](https://rustc-dev-guide.rust-lang.org/mir/index.html) into [SAT/SMT](https://en.wikipedia.org/wiki/Satisfiability_modulo_theories) and does **exhaustive** search (proof).

The `BytesAndTimestamps` are extracted from `telio-wg`, because Kani doesn't support SSE(2) but `telio-wg` depends on `ring`, which assumes SSE(2) for intel targets.

Custom `Instant` is used in the Kani build, because Kani doesn't support required ffi, eg:

```
 ** 1 of 7685 failed (7684 undetermined)
Failed Checks: call to foreign "C" function `clock_gettime` is not currently supported by Kani. Please post your example at https://github.com/model-checking/kani/issues/2423
 File: "/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.171/src/unix/linux_like/mod.rs", line 1775, in std::sys::pal::unix::time::Timespec::now

VERIFICATION:- FAILED
** WARNING: A Rust construct that is not currently supported by Kani was found to be reachable. Check the results for more details.
Verification Time: 1.79865s
```

For short explanation of how this relates to fuzzing and property testing see: https://model-checking.github.io/kani/tool-comparison.html

Kani is already used in production, for example by Amazan's quic implementation: https://model-checking.github.io/kani-verifier-blog/2023/05/30/how-s2n-quic-uses-kani-to-inspire-confidence.html

- [x] README.md is updated


mathiaspeters:
+1

LukasPukenis (approved):
Very very interesting. Nice demo +1

crates/telio-utils/src/bytes_and_timestamps.rs
```diff
@@ -230,7 +230,16 @@ mod proptests {
...
+        let max_delta = Duration::from_secs(
+            rx_sec_ago
+                .into_iter()
+                .chain(tx_sec_ago.into_iter())
+                .max()
+                .unwrap_or_default(),
+        );
+        // The Instant might be a time since boot, and if this test is run right after boot
+        // it can result in a overflow while subtracting few lines below.
+        let start = Instant::now() + max_delta;
```
LukasPukenis, line 242: Minor: I had some trouble understanding this at first but basically this is to set `start` big enough that subtracting timestamp from it would not underflow If so could we make the intention clearer by just  `start = max(tx_sec_ago, rx_sec_ago)`?

LukasPukenis (approved):

PR #1202, opened 2025-04-10
teliod log filename format

tomasz-grz:
### Problem
teliod log files use the current date as filename suffix (`log_file.log.2025-04-09`).
Not complying with existing `.gitignore` rule and any file extension based log viewer.

### Solution
Use the date as filename prefix instead (`2025-04-09.log_file.log`).


tomaszklak:
I'm not strongly against this, but I believe that current format is the "industry standard": https://github.com/NordSecurity/libtelio/pull/1163#discussion_r1993580353

tomasz-grz:
> I'm not strongly against this, but I believe that current format is somewhat of a "industry standard": [#1163 (comment)](https://github.com/NordSecurity/libtelio/pull/1163#discussion_r1993580353)

@tomaszklak that's a good point, I'll update `.gitignore` then at least :D

matislovas (approved):
+2

Jauler (approved):
+2.0

PR #1205, opened 2025-04-11
Improve socket rebinding on Apple platforms

twardakm:
### Problem
When [includeAllNetworks](https://developer.apple.com/documentation/networkextension/nevpnprotocol/includeallnetworks) parameter is set, and the user changes the network interface (e.g. connects/disconnects Ethernet cable to a mac) we end up in a no-net situation. When `includeAllNetworks` flag is set we cannot call the `bind` method too soon, as the system will block the traffic

### Solution
The solution to this problem is to call `bind` method as soon as the new network interface is fully reconfigured. This brings the challenge of knowing when this has happened. I tried to find some indication in `nw_path_monitor` or in the DynamicStore, unfortunately, I didn't find a single indicator that would say it is 100% safe to rebind the sockets.

I settled on a solution to observe 2 keys in the DynamicStore:
1. `State:/Network/Interface/.*/IPConfigurationBusy`
2. `State:/Network/Global/IPv4`

The 1st one (`IPConfigurationBusy`) appears on a network interface when it is being reconfigured and disappears once the reconfiguration is finished. It is the most reliable one on my real device (macOS 15.4), but I found it problematic in a VM for macOS 14.0, where it tends to be advertised much quicker, and we end up rebinding too soon. To solve that problem I also added an observer for `State:/Network/Global/IPv4` and this combination seems to work perfectly fine.

As a drawback, we ended up rebinding sockets around 4 - 5 times per network change event.

- [x] README.md is updated

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -32,8 +32,9 @@ use tokio::{
 use crate::native::{interface_index_from_name, NativeSocket};
 use crate::Protector;
 use network_framework_sys::{
-    nw_interface_get_name, nw_interface_t, nw_path_monitor_create, nw_path_monitor_set_queue,
-    nw_path_monitor_start, nw_path_monitor_t, nw_path_t,
+    nw_interface_get_name, nw_interface_t, nw_path_get_status, nw_path_monitor_create,
+    nw_path_monitor_set_queue, nw_path_monitor_start, nw_path_monitor_t, nw_path_status_satisfied,
+    nw_path_t,
 };
 use nix::sys::socket::{getsockname, AddressFamily, SockaddrLike, SockaddrStorage};
 use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn};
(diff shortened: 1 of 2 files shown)


crates/telio-sockets/src/protector/apple.rs
```diff
@@ -444,7 +445,12 @@ fn dynamic_store_callback(
     context: &mut Context,
 ) {
     if let Some(sockets) = context.sockets.upgrade() {
-        sockets.lock().rebind_all(false);
+        // NMACOS-8047 The callback is executed when specified keys in the DynamicStore change.
+        // In such case we should force rebind the sockets to avoid no-net when `includeAllNetworks` is set.
+        telio_log_info!("Force rebinding the sockets because of DynamicStore key change");
+        telio_log_debug!("DynamicStore key changed: {:?}", _changed_keys);
+
+        sockets.lock().rebind_all(true);
```
tomaszklak, line 453: I think that after this change, all callsites of `rebind_all` pass `true`. If that's the case, please remove this parameter and hardcod the `true` behaviour.

tomaszklak (approved):
+1

crates/telio-sockets/src/protector/apple.rs
twardakm: true that! I removed the unnecessary parameter

tomaszklak:
Please remove the [Apply cargo formatter](https://github.com/NordSecurity/libtelio/pull/1205/commits/6dbdf0816719a00df351d1380a99c08298f08a07) commit, and apply it's changes to other existing commits so that the history is clean.

twardakm:
> Please remove the [Apply cargo formatter](https://github.com/NordSecurity/libtelio/pull/1205/commits/6dbdf0816719a00df351d1380a99c08298f08a07) commit, and apply it's changes to other existing commits so that the history is clean.

sure, I have done that, but aren't those commits squashed together while merging to the main branch?

tomaszklak:
> > Please remove the [Apply cargo formatter](https://github.com/NordSecurity/libtelio/pull/1205/commits/6dbdf0816719a00df351d1380a99c08298f08a07) commit, and apply it's changes to other existing commits so that the history is clean.
> 
> sure, I have done that, but aren't those commits squashed together while merging to the main branch?

That's up to you to decide how this PR should be merged :)

tomaszklak (approved):

stalowyjez (approved):
Lgtm, +1

PR #1206, opened 2025-04-14
Link can't be up before the first received packet

tomaszklak:
In the last week's change to link state detection there was a bug for the case before receiving first ever packet. In that case, we should always report that the link is down.

- [x] README.md is updated

### DIFF
--- a/.unreleased/is_link_up_is_always_false_before_first_received_packet
+++ b/.unreleased/is_link_up_is_always_false_before_first_received_packet
@@ -1,0 +1,1 @@
+Never report a link state UP before the first packet is received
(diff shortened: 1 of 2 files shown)


tomasz-grz (commented):
+1

tomaszklak:
> I think we may have another edge case. Although I have not double-checked it.
> 
> In case we have received something and then we are sitting idle. Would'nt it be possible to report link as UP even though WireGuard session has expired (180s+)

I think that you are correct, but it only affects responders, since initiators will send out a packet after 120s. Do we have a way to know if we are responder or initiator?

PR #1208, opened 2025-04-14
LLT-6123: Teliod daemonize

tomasz-grz:
### Problem
According to the [daemon](https://linux.die.net/man/1/daemonize) convention, a daemon should detach itself from the terminal, disassociate from its process group and run in the background.

### Solution
Use `daemonize` crate to detach the daemon process, set umask, change process group, and redirect stdout and stderr to a file.
The daemon process needs to be forked before starting the `tokio` runtime, otherwise it panics.

`teliod daemon` command was replaced with `teliod start` to make it more semantically correct.
`--no-detach` flag was added to run teliod attached to the terminal as a regular process.
Added a `--stdout-path` parameter to specify where to redirect stdout and stderr.
Added a `--working-directory` parameter to specify which working directory the daemon should use.

The unix socket mode is set to `600` so it's not accessible to the non-root user.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/README.md
+++ b/clis/teliod/README.md
@@ -1,42 +1,68 @@
-### Building Teliod
+# Teliod
+
+Teliod is small Meshnet and VPN application, designed to be used in standalone
+and embedded environments.
+
+## Building Teliod
 
 For typical Linux environment it might be built using simply:
 
 ```cargo build```
 
-For OpenWRT you might need a bit more complex command, including your router architecture and the fact the OpenWRT is MUSLE-based, for example:
+For OpenWRT you might need a bit more complex command, including your router architecture
+and the fact the OpenWRT is MUSLE-based, for example:
 
 ```CARGO_TARGET_ARMV7_UNKNOWN_LINUX_MUSLEABIHF_LINKER=rust-lld CC=/path/to/arm-linux-gnueabi-gcc cargo build --package teliod --target armv7-unknown-linux-musleabihf```
 
 You may need to download some sufficient MUSLE toolchains from `musle.cc`.
 
-### Using Teliod daemon
+## Using Teliod
 
-Teliod runs Telio library in the background and provides a simple CLI tool to manage it.
+Teliod runs Telio library in the background and provides a simple CLI tool to
+manage it.
+
 There is a command for running the daemon:
 
-- `teliod daemon <path_to_config_file>` - starts the daemon. The config file should be provided in a JSON format (see `example_teliod_config.json` file). Currently supported configuration variables:
-  - `log_level` - filters the logged messages by priority, possible levels:
-    - `error`
-    - `warn`
-    - `info`
-    - `debug`
-    - `trace`
-    - `off`
-  - `log_file_path` - a path to the daemon's logging file, needs to specify both the directory and the file name
-  - `log_file_count` - number of recent log files (log files are rotated daily)
-  - `authentication_token` - Token from Nord VPN account to authenticate API calls
-  - `app_user_uid` - A unique number for each user of the application
-  - `adapter_type` - Wireguard adapter to use, possible options:
-    - `neptun` - User space implementation, available on multiple platforms
-    - `linux-native` - Linux native implementation
-  - `interface`
-    - `name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
-    - `config_provider` - Provider for configuring the interface address, possible options:
-      - `manual` - do not configure interfaces automatically
-      - `ifconfig` - systems using ifconfig command
-      - `iproute` - systems using iproute2 command
+- `teliod start <path_to_config_file>` - starts the daemon.
+  - `--no-detach` - run the teliod in the foreground as a regular process,
+  without detaching from the terminal.
+  - `--stdout-path` - Redirect daemon standard output to the specified file,
+  Defaults to `/var/log/teliod.log`, ignored when used with `--no-detach`
+  Some early logs may still be printed to stdout before redirection.
+  - `--working-directory` - Specifies the daemons working directory,
+  Defaults to `/`, ignored when used with `--no-detach`
+
+The config file should be provided in a JSON format
+(see `example_teliod_config.json` file).
+
+Currently supported configuration variables:
+
+- `log_level` - filters the logged messages by priority, possible levels:
+  - `error`
+  - `warn`
+  - `info`
+  - `debug`
+  - `trace`
+  - `off`
+- `log_file_path` - a path to store the daemon's logs,
+needs be absolute, otherwise will be relative to `working-directory` when daemonized
+- `log_file_count` - number of recent log files (log files are rotated daily)
+- `authentication_token` - Token from Nord VPN account to authenticate API calls
+- `app_user_uid` - A unique number for each user of the application
+- `adapter_type` - Wireguard adapter to use, possible options:
+  - `neptun` - User space implementation, available on multiple platforms
+  - `linux-native` - Linux native implementation
+- `interface`
+  - `name` - Name of tunnel interface to connect to. Note that for macOS
+  the name has to be in form `tun#` where `#` can be any integer number
+  - `config_provider` - Provider for configuring the interface address,
+  possible options:
+    - `manual` - do not configure interfaces automatically
+    - `ifconfig` - systems using ifconfig command
+    - `iproute` - systems using iproute2 command
 
 And following cli commands:
 
 - `teliod get-status` - returns the status of teliod and the meshnet peer list
+- `teliod is-alive` - query if the daemon is running
+- `teliod quit-daemon` - stop daemon execution
--- a/clis/teliod/example_teliod_config.json
+++ b/clis/teliod/example_teliod_config.json
@@ -1,6 +1,6 @@
 {
     "log_level": "trace",
-    "log_file_path": "./example_log_file.log",
+    "log_file_path": "/var/log/example_log_file.log",
     "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
     "adapter_type": "neptun",
     "interface": {
--- a/clis/teliod/src/comms.rs
+++ b/clis/teliod/src/comms.rs
@@ -14,7 +14,7 @@ use interprocess::{
         traits::tokio::{Listener, Stream},
         ListenerOptions, ToFsName,
     },
-    os::unix::local_socket::FilesystemUdSocket,
+    os::unix::local_socket::{FilesystemUdSocket, ListenerOptionsExt},
 };
 
 /// Struct for handling connections of the daemon's side of the IPC communication with the API.
@@ -56,10 +56,19 @@ impl DaemonSocket {
     pub fn new(ipc_socket_path: &Path) -> Result<Self> {
         // Delete the socket file if it already exists
         let _ = fs::remove_file(ipc_socket_path);
-        let socket = ListenerOptions::new()
+
+        let options = ListenerOptions::new()
             .name(ipc_socket_path.to_fs_name::<FilesystemUdSocket>()?)
-            .reclaim_name(true)
-            .create_tokio()?;
+            .reclaim_name(true);
+
+        // Disable setting umask for tests.
+        // `umask()` C API call is operating on a process level, and therfore is not thread-safe.
+        // As multiple cargo tests are executing in parallel this causes a panic.
+        #[cfg(not(test))]
+        // Mode for unix socket rw-------
+        let options = options.mode(0o600);
+
+        let socket = options.create_tokio()?;
 
         Ok(Self { socket })
     }
--- a/clis/teliod/src/daemon.rs
+++ b/clis/teliod/src/daemon.rs
@@ -11,21 +11,18 @@ use telio::{
     telio_utils::select,
     telio_wg::AdapterType,
 };
-use tokio::sync::mpsc::Sender;
-use tokio::{sync::mpsc, sync::oneshot, time::Duration};
+use tokio::{sync::mpsc, sync::mpsc::Sender, sync::oneshot, time::Duration};
 use tracing::{debug, error, info, trace, warn};
 
-use crate::core_api::{get_meshmap as get_meshmap_from_server, init_with_api};
-use crate::logging::setup_logging;
-use crate::ClientCmd;
 use crate::Hidden;
 use crate::{
     command_listener::CommandListener,
     comms::DaemonSocket,
     config::DeviceIdentity,
     config::{InterfaceConfig, TeliodDaemonConfig},
+    core_api::{get_meshmap as get_meshmap_from_server, init_with_api},
     nc::NotificationCenter,
-    TelioStatusReport, TeliodError,
+    ClientCmd, TelioStatusReport, TeliodError,
 };
 
 #[derive(Debug)]
@@ -209,13 +206,6 @@ async fn daemon_init(
 }
 
 pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodError> {
-    let _tracing_worker_guard = setup_logging(
-        &config.log_file_path,
-        config.log_level,
-        config.log_file_count,
-    )
-    .await?;
-
     debug!("started with config: {config:?}");
 
     let mut signals = Signals::new([SIGHUP, SIGTERM, SIGINT, SIGQUIT])?;
@@ -248,8 +238,6 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
     });
 
     info!("Entering event loop");
-    eprintln!("Daemon started");
-
     loop {
         select! {
             // Check if telio_task completes and exit if it fails
--- a/clis/teliod/src/main.rs
+++ b/clis/teliod/src/main.rs
@@ -1,16 +1,16 @@
 //! Main and implementation of config and commands for Teliod - simple telio daemon for Linux and OpenWRT
 
 use clap::Parser;
+use daemonize::{Daemonize, Outcome};
 use serde::{Deserialize, Serialize};
 use serde_json::error::Error as SerdeJsonError;
-use std::{fs, net::IpAddr};
+use std::{fs::OpenOptions, net::IpAddr};
 use telio::{device::Error as DeviceError, telio_model::mesh::Node, telio_utils::Hidden};
 use thiserror::Error as ThisError;
 use tokio::{
     task::JoinError,
     time::{timeout, Duration},
 };
-use tracing::{debug, error};
 use tracing_appender::rolling::InitError;
 
 #[cfg(feature = "cgi")]
@@ -33,6 +33,9 @@ use crate::{
 
 const TIMEOUT_SEC: u64 = 1;
 
+/// Umask allows only rw-rw-r--
+const DEFAULT_UMASK: u32 = 0o113;
+
 #[derive(Parser, Debug, PartialEq)]
 #[clap()]
 #[derive(Serialize, Deserialize)]
@@ -49,12 +52,36 @@ enum ClientCmd {
 #[clap()]
 enum Cmd {
     #[clap(about = "Runs the teliod event loop")]
-    Daemon { config_path: String },
+    Start(DaemonOpts),
     #[clap(flatten)]
     Client(ClientCmd),
     #[cfg(feature = "cgi")]
     #[clap(about = "Receive and parse http requests")]
     Cgi,
+}
+
+#[derive(Parser, Debug)]
+struct DaemonOpts {
+    /// Path to the config file
+    config_path: String,
+
+    /// Do not detach the teliod process from the terminal
+    #[clap(long = "no-detach")]
+    no_detach: bool,
+
+    /// Specifies the daemons working directory.
+    ///
+    /// Defaults to "/".
+    /// Ignored with no_detach flag.
+    #[clap(long = "working-directory", default_value = "/")]
+    working_directory: String,
+
+    /// Redirect standard output to the specified file
+    ///
+    /// Defaults to "/var/log/teliod.log".
+    /// Ignored with no-detach flag.
+    #[clap(long = "stdout-path", default_value = "/var/log/teliod.log")]
+    stdout_path: String,
 }
 
 #[derive(Debug, ThisError)]
@@ -91,6 +118,8 @@ enum TeliodError {
     InvalidConfigOption(String, String, String),
     #[error(transparent)]
     LogAppenderError(#[from] InitError),
+    #[error(transparent)]
+    DaemonizeError(#[from] daemonize::Error),
 }
 
 /// Libtelio and meshnet status report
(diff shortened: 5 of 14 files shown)


clis/teliod/src/comms.rs
```diff
@@ -159,3 +157,24 @@ impl DaemonConnection {
...
+    } else if Path::new("/var/run/").exists() {
+        PathBuf::from("/var/run/")
+    } else {
+        return Err(Error::new(
+            ErrorKind::NotFound,
+            "Neither /run/ nor /var/run/ exists",
+        ));
+    }
+    .join("teliod"))
+}
```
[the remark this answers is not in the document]
LukasPukenis: Oh, it's already handled in `main.rs` which makes sense so this rather returns just a path. Because this function is used only in one place, maybe it could also handle the creation part of it?

nat-lab/tests/test_teliod.py
```diff
@@ -54,17 +68,63 @@ async def test_teliod() -> None:
...
+                await connection.create_process(TELIOD_QUIT_DAEMON_PARAMS).execute()
+            ).get_stdout()
+        )
+
+        assert not await is_teliod_running(connection)
+
+        # Run the is-alive command again - this time it should fail
+        with pytest.raises(ProcessExecError) as err:
+            await connection.create_process(TELIOD_IS_ALIVE_PARAMS).execute()
+        assert err.value.stderr == "Error: DaemonIsNotRunning"
```
LukasPukenis, line 130: Could I ask to also add the the logger checks? Basically just observe that log files that are expected are created and contain such and such string. Doesn't need to be very precise, just to be make it sure that the actual logs go to where they should go

clis/teliod/src/daemon.rs
```diff
@@ -247,7 +245,7 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
     });
 
     info!("Entering event loop");
-    eprintln!("Daemon started");
+    println!("Daemon started");
```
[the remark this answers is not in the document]
tomasz-grz: Actually, I added this print line myself some time ago, it was mostly useful during back-end API calls, to see if it was running properly directly from the terminal, without having to open the log files and filter out through things..

`println!` would indeed have it's output re-directed after being daemonized, otherwise it will be print to the terminal.
I wanted to keep STDERR file clean of any other "warnings" besides fatal crashes, that's why I change this from `eprintln` to `println`

clis/teliod/src/main.rs
```diff
@@ -104,28 +124,81 @@ pub struct TelioStatusReport {
...
+                Outcome::Child(Ok(_)) => {}
+                // Errors
+                Outcome::Parent(Err(err)) => {
+                    eprintln!("Fork parent error: {}", err);
+                    return Err(err.into());
+                }
+                Outcome::Child(Err(err)) => {
+                    eprintln!("Child error {}", err);
+                    return Err(err.into());
                 }
```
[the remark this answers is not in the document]
tomasz-grz: Unfortunately we can not setup logger here as it relies on `async`, but we need to deamonize before starting Tokio runtime. Unless we setup an "intermediate" non async logger.. but that might be an overkill just for a few prints..

nat-lab/tests/test_teliod.py
tomasz-grz: That's a good idea, I'll try to add some tests for logging.

In this case though the message does not go to a log file, but is displayed in the terminal as the `TELIOD_IS_ALIVE` is a "client" command.

clis/teliod/src/main.rs
```diff
@@ -104,28 +124,81 @@ pub struct TelioStatusReport {
...
-                        config.authentication_token = token;
-                    } else {
-                        error!("Token from env not valid")
-                    }
+        // Check if daemon already is running before forking
+        if DaemonSocket::get_ipc_socket_path()?.exists() {
+            return Err(TeliodError::DaemonIsRunning);
+        }
+
+        // Fork the process before starting tokio runtime
```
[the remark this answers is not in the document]
tomasz-grz: Good point, I'll add a better description in the comments :D

clis/teliod/README.md
```diff
@@ -1,42 +1,59 @@
...
+Currently supported configuration variables:
+
+- `log_level` - filters the logged messages by priority, possible levels:
+  - `error`
+  - `warn`
+  - `info`
+  - `debug`
+  - `trace`
+  - `off`
+- `log_file_path` - a path to store the daemon's logs,
```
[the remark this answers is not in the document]
tomasz-grz: `log_file_path` is only for the "libtelio" aka "meshnet" logs...

clis/teliod/README.md
```diff
@@ -1,42 +1,59 @@
...
-    - `neptun` - User space implementation, available on multiple platforms
-    - `linux-native` - Linux native implementation
-  - `interface`
-    - `name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
-    - `config_provider` - Provider for configuring the interface address, possible options:
-      - `manual` - do not configure interfaces automatically
-      - `ifconfig` - systems using ifconfig command
-      - `iproute` - systems using iproute2 command
+- `teliod daemon <path_to_config_file>` - starts the daemon.
+  - `--no-detach` - run the daemon in the foreground as a regular process.
```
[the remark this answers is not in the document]
tomasz-grz: So you would rather have it by default run as a process, and only daemonized with an optional `-d` flag? That part shouldn't be big to change

clis/teliod/src/main.rs
```diff
@@ -104,28 +124,81 @@ pub struct TelioStatusReport {
...
+        if !opts.no_detach {
+            // Fix relative path before changing working directory
+            let config_path = PathBuf::from(&opts.config_path);
+            if config_path.is_relative() {
+                opts.config_path = config_path.canonicalize()?.to_string_lossy().to_string();
+            }
+
+            // redirect stdout and stderr to files
+            let stdout = File::create(wd_path.join("stdout.log"))?;
+            let stderr = File::create(wd_path.join("stderr.log"))?;
```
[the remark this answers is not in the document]
tomasz-grz: I actually like this idea, we change `log_file_path` to `log_dir_path` and then all the logs live there?

nat-lab/tests/test_teliod.py
tomasz-grz: Added some tests for checking if log files exist

clis/teliod/src/main.rs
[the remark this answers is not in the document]
tomasz-grz: I moved the initialization of the daemon logger to the tokio_main function.

clis/teliod/src/main.rs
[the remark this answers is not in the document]
tomasz-grz: Added better comment

clis/teliod/src/comms.rs
```diff
@@ -56,10 +46,18 @@ impl DaemonSocket {
...
-        let socket = ListenerOptions::new()
+
+        let options = ListenerOptions::new()
             .name(ipc_socket_path.to_fs_name::<FilesystemUdSocket>()?)
-            .reclaim_name(true)
-            .create_tokio()?;
+            .reclaim_name(true);
+
+        // Disable setting umask for tests,
+        // they are running in parallel and this panics because of a concurent umask call
```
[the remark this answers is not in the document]
tomasz-grz: Added a better description

clis/teliod/README.md
```diff
@@ -1,42 +1,59 @@
...
-  - `adapter_type` - Wireguard adapter to use, possible options:
-    - `neptun` - User space implementation, available on multiple platforms
-    - `linux-native` - Linux native implementation
-  - `interface`
-    - `name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
-    - `config_provider` - Provider for configuring the interface address, possible options:
-      - `manual` - do not configure interfaces automatically
-      - `ifconfig` - systems using ifconfig command
-      - `iproute` - systems using iproute2 command
+- `teliod daemon <path_to_config_file>` - starts the daemon.
```
lcruz99, line 23: Isn't it confusing for the user to decide if he wants to run `teliod daemon` with or without `--daemonize`?

At this point I would change `teliod daemon` to something else.

clis/teliod/example_teliod_config.json
```diff
@@ -1,6 +1,6 @@
 {
     "log_level": "trace",
-    "log_file_path": "./example_log_file.log",
+    "log_file_path": "/var/log/example_log_file.log",
```
lcruz99, line 3: Not to be mistaken with `/var/log/teliod.log`. Maybe change the name to avoid confusion with which logs are each.

clis/teliod/README.md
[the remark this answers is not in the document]
tomasz-grz: Actually now after having a chat with Luis about it.. 
`teliod daemon -d` aka `teliodaemon daemon --daemonize` 👿 
That's why I initially chose the `--no-detach` and it would by default behave as a daemon..

Otherwise maybe we should change it to `teliod server -d`, but this would be a pretty big change so I would make it a separate PR..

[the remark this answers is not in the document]
LukasPukenis: This is a minor comment, I marked it as such but my reasoning is that daemon is basically just how the server behaves and I like that daemonizing stuff is additional flag the way I indicated in the docker example. Basically a thing can run on it's own but optionally it can be a daemon.

I also mentioned `teliod server` instead of `teliod daemon` since I think server indicates the purpose better and `-d` just makes it daemonized like `teliod server -d`

clis/teliod/README.md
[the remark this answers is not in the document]
LukasPukenis: TLDR: my bad.

I think I see where my confusion came from.. So `teliod` is a wrapper for libtelio and both live in the same process, and I thought that `teliod` also uses `telio_log_*` macros, thus the logs would go to the same logfile with no clear separation. **I was wrong here** since `clis/teliod` doesn't contain any `telio_log_*` invocation just few `println!` and `eprintln!()` calls that can be redirected where we want :)

To me this separation of the logging macros was not clear initially :bow:

LukasPukenis:
Upon another round of review I marked these questions/points:

- `teliod` already implies being a daemon by the `d` suffix. I missed it initially. It makes the `daemon` part in `teliod daemon` redundant. By the naming convention it should be that `teliod` is already launching a daemon and it isn't interactive for control. A possible improvement for it would be to rename `teliod` to `telio` and it would have the ability to daemonize, like `telio server --daemonize`. It however, might not be a pleasant change(we already use `teliod` in our vocabulary a lot). Another idea is to rename the `daemon` argument to `start` which would be very cheap to do.

- Do we need not to daemonize teliod ever or is it just for the convenience when testing? If we do not strictly need it, I would suggest just omitting it and not caring. It adds another potential control flow we could avoid.

- Should we do something about `println!` vs tracing macros in the `teliod` code? This means that parts of `teliod` logs are going to one place, and other logs go together with meshnet logs. One way to "normalize" it would be to replace tracing macros with `prinltn!` and `eprintln!` in `teliod` code and not use tracing macros there at all except for setting up the logger for libtelio. This way the output would be redirected to where the daemon logs are redirected at the moment(/var/log/) and `log_file_path` would be solely for libtelio operations(unrelated to daemon itself).

clis/teliod/example_teliod_config.json
tomasz-grz: Isn't `/var/log/teliod.log` just the stdout?

clis/teliod/README.md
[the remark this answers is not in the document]
tomasz-grz: Yes all `debug!`, `info!` and `warn!` logs go to `log_file_path` file

clis/teliod/README.md
tomasz-grz: Renamed to `teliod start`

clis/teliod/example_teliod_config.json
tomasz-grz: I would like to address all the interactions with QNAP and CGI in a separate PR, and keep those changes minimal here, just the addition of `--no-detach` flag to keep it working exactly as previously.

lcruz99: > Isn't /var/log/teliod.log just the stdout?

stdout and stderr ([see](https://github.com/NordSecurity/libtelio/blob/teliod_daemonize/qnap/shared/NordSecurityMeshnet.sh#L52))

nat-lab/tests/test_teliod.py
```diff
@@ -54,17 +73,110 @@ async def test_teliod() -> None:
...
+
+
+async def test_teliod_logs() -> None:
```
LukasPukenis, line 138: I have two small suggestions for improving the testcases on otherwise already solid setup:
1. Could we parametrize this testcase or all of the teliod testcases with daemon and nodaemon modes? I feel this could be very beneficial since it might behave slightly different depending on the mode especially in the logging case since it will produce different files for example, would be nice to have guards on those things.

2. Additionally, if the log files, at this point, have something in them already written, maybe we could check for some strings that are there? This would secure us against an accidental move of the logger setup code to a later stage resulting in missing logs before that.

nat-lab/tests/test_teliod.py
```diff
@@ -2,21 +2,35 @@
...
 from utils.connection import ConnectionTag
 from utils.process.process import ProcessExecError
 
 TELIOD_EXEC_PATH = f"{LIBTELIO_BINARY_PATH_DOCKER}/teliod"
 CONFIG_FILE_PATH = "/etc/teliod/config.json"
 SOCKET_FILE_PATH = "/run/teliod.sock"
+STDOUT_FILE_PATH = "/var/log/teliod_stdout.log"
+STDERR_FILE_PATH = "/var/log/teliod_error.log"
+# Build a dated log filename
+LOG_FILE_PATH = f"/var/log/teliod_natlab.log.{datetime.today().strftime('%Y-%m-%d')}"
```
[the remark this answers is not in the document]
tomasz-grz: Because we have log rotation..
Buut you might be absolutely right, possibly we can change `log_file_count` in the config, I wonder if changing that to 1, disables log rotation

clis/teliod/README.md
```diff
@@ -1,42 +1,63 @@
...
+
+- `log_level` - filters the logged messages by priority, possible levels:
+  - `error`
+  - `warn`
+  - `info`
+  - `debug`
+  - `trace`
+  - `off`
+- `log_file_path` - a path to store the daemon's logs,
+needs be absolute, otherwise `/var/run/` will be used when daemonized
```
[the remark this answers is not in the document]
tomasz-grz: This is the daemons running directory after it detaches from the terminal ( it does `chdir()`). By convention it's `/`, but I thought `/var/run/` is probably more suitable since it's socket also lives there..
In theory the process can also `chdir()` to `/var/log/` 🤔

nat-lab/tests/test_teliod.py
tomasz-grz: It's a good idea, though there are some caveats. 
Non-daemon modes of course will not have the `stdout` and `stderror` longs as those outputs are directed to the terminal.
And `stderror` would be empty in most cases. But it's good idea maybe to deliberately introduce an error and check for a message in the error logs.

Regarding `stdout`, what message should we look for? Currently it only has messages from "Telio tasks", and if somebody changes that message, should it really be considered an error to fail the pipeline? Or would it be their responsibility to also change the failing test?
with `teliod_natlab.log` I guess it's easier since it's in the "our code base" so we can "deliberately" add a marker to look for.. 🤔 What do you think?

We could also simply check that the files are not empty?

lcruz99 (changes requested):
Reminder to clean the commit history before merging.

clis/teliod/src/main.rs
```diff
@@ -104,28 +126,84 @@ pub struct TelioStatusReport {
...
+                }
+            }
+        }
+    }
+    tokio_main(cmd)
+}
 
-                Box::pin(daemon::daemon_event_loop(config)).await
+#[tokio::main]
+async fn tokio_main(cmd: Cmd) -> Result<(), TeliodError> {
```
lcruz99, line 185: nitpick: This is teliod's main that is decorated by `tokio::main` macro, so I would give it another name, eg: async_main, main_multi_thread, etc.

clis/teliod/src/main.rs
```diff
@@ -104,28 +126,84 @@ pub struct TelioStatusReport {
...
+        // Tokio creates a multi-threaded asynchronous runtime,
+        // but during forking only a single thread survives,
+        // leaving tokio runtime in an undefined state and resulting in a panic.
+        // https://github.com/tokio-rs/tokio/issues/4301
+        if !opts.no_detach {
+            // Fix relative config path before changing daemons working directory
+            let config_path = PathBuf::from(&opts.config_path);
+            if config_path.is_relative() {
+                opts.config_path = config_path.canonicalize()?.to_string_lossy().to_string();
+            }
```
lcruz99, line 149: What are you trying to do here? 🤔

tomasz-grz: As mentioned in the comment, convert relative path to absolute.

If a relative path is given to the config file. The daemonized process can not access it since it was relative to the current path in the terminal.

tomasz-grz: Of-course it's not strictly necessary, but a quality of life improvement. I can remove this if it's intuitive that only absolute paths can be used when daemonized, but also relative paths if it's attached to the terminal.

lcruz99 (changes requested):
Logs originating from the teliod process - daemonized or not - should only use println!/eprintln! macros and leave the logger for libtelio logs. Eg: https://github.com/NordSecurity/libtelio/blob/6cb8be9f4025caa84254da4393fd49ae78608766/clis/teliod/src/main.rs#L199, https://github.com/NordSecurity/libtelio/blob/6cb8be9f4025caa84254da4393fd49ae78608766/clis/teliod/src/main.rs#L203

clis/teliod/src/main.rs
```diff
@@ -104,28 +126,84 @@ pub struct TelioStatusReport {
...
-                    debug!("Overriding token from env");
-                    if token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit()) {
-                        config.authentication_token = token;
-                    } else {
-                        error!("Token from env not valid")
-                    }
+            // Redirect stdout and stderr to files in /var/log
+            let log_path = PathBuf::from("/var/log");
+            let stdout = File::create(log_path.join("teliod_stdout.log"))?;
+            let stderr = File::create(log_path.join("teliod_error.log"))?;
```
lcruz99, line 154: Isn't it possible for both handles to share the same file?

lcruz99:
```suggestion
            let daemon = Daemonize::new()
                .umask(DEFAULT_UMASK)
                .working_directory(get_wd_path()?)
                .stdout(stdout)
                .stderr(stderr);
```

tomasz-grz: It might be possible.
The benefit of having a separate `stderr` is that if it's empty, you know that no errors happened.

lcruz99: It's a good point but I wouldn't do it to minimize the app footprint as possible.

lcruz99: Error logs should be easily recognized by default.

clis/teliod/src/main.rs
tomasz-grz: Removed,
Now running `sudo ./teliod start teliod_config.json` will make the daemon to fail even if `teliod_config.json` is located in the current directory.

Instead it will have to be run like this `sudo ./teliod start /Users/tomasz/Developer/Nord/LLT/libtelio-build/libtelio/clis/teliod/teliod_config.json`

clis/teliod/src/main.rs
tomasz-grz: Unified the two logs into a single file

nat-lab/tests/test_teliod.py
tomasz-grz: Added tests for content

nat-lab/tests/test_teliod.py
```diff
@@ -54,17 +73,110 @@ async def test_teliod() -> None:
...
+        assert err.value.stderr == "Error: DaemonIsNotRunning"
+
+        # Run teliod
+        await exit_stack.enter_async_context(
+            connection.create_process(start_daemon_params).run()
+        )
+
+        # Let the daemon start
+        while not await is_teliod_running(connection):
+            await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
tomasz-grz: Added timeout

nat-lab/tests/test_teliod.py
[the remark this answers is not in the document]
tomasz-grz: Setting `log_file_count` to 1 or even 0 does not disable log rotation.. i think this is more on how many old files we keep around.

clis/teliod/src/main.rs
```diff
@@ -141,21 +138,19 @@ fn main() -> Result<(), TeliodError> {
...
-            // Redirect stdout and stderr to files in /var/log
-            let log_path = PathBuf::from("/var/log");
-            let stdout = File::create(log_path.join("teliod_stdout.log"))?;
-            let stderr = File::create(log_path.join("teliod_error.log"))?;
+            // Redirect stdout and stderr to file in /var/log/teliod_stdout.log
+            let log_file = OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(true)
+                .open("/var/log/teliod_stdout.log")?;
```
lcruz99, line 146: nit: teliod_stdout.log -> teliod.log

nat-lab/tests/test_teliod.py
```diff
@@ -11,7 +12,6 @@
 CONFIG_FILE_PATH = "/etc/teliod/config.json"
 SOCKET_FILE_PATH = "/run/teliod.sock"
```
lcruz99, line 13: In the future we can make this also configurable through command line flag

clis/teliod/src/main.rs
tomasz-grz: I wanted to make this more explicit since it's the redirected stdout & stderr,to be more distinguishable from `/var/log/teliod.log.2025-05-13` for example if it's set in `log_file_path`

What would you recommend calling `log_file_path`? `libtelio.log` or `teliod_tunnel.log` maybe?

nat-lab/tests/test_teliod.py
tomasz-grz: Possibly, but in that case the user needs to know where the socket is located all the time, which I don't feel is very user friendly..
for example it would look something like `teliod get-status /run/teliod.sock`?

clis/teliod/src/main.rs
```diff
@@ -104,28 +122,75 @@ pub struct TelioStatusReport {
...
+            let daemon = Daemonize::new()
+                .umask(DEFAULT_UMASK)
+                .stdout(log_file.try_clone()?)
+                .stderr(log_file);
+
+            // Daemonize the process
+            match daemon.execute() {
+                // Quit parent process
+                Outcome::Parent(Ok(_)) => {
+                    return Ok(());
```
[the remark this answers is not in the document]
tomasz-grz: Yes those prints are right below:

```
                // Errors
                Outcome::Parent(Err(err)) => {
                    eprintln!("Fork parent error: {}", err);
                    return Err(err.into());
                }
                Outcome::Child(Err(err)) => {
                    eprintln!("Child error {}", err);
                    return Err(err.into());
                }
```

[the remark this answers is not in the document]
tomasz-grz: But of course the child error would not be printed here bout would go into the `stderr` file.

nat-lab/tests/test_teliod.py
tomasz-grz: I guess as an optional it could make more sense :D please submit a feature request

clis/teliod/src/main.rs
lcruz99: It's just a nit, I'd call it just `teliod.log` and `log_file_path` libtelio.log/tunnel.log/meshnet.log. QNAP will have a different preconfigured name since it won't have VPN functionalities

tomasz-grz: Renamed

clis/teliod/README.md
```diff
@@ -26,6 +26,10 @@ There is a command for running the daemon:
 - `teliod start <path_to_config_file>` - starts the daemon.
   - `--no-detach` - run the teliod in the foreground as a regular process,
   without detaching from the terminal.
+  - `--stdout_path` - Redirect standard output to the specified file,
+  Defaults to "/var/log/teliod.log", ignored when used with `--no-detach`
+  - `--working_directory` - Specifies the daemons working directory,
+  Defaults to "/", ignored when used with `--no-detach`
```
LukasPukenis, line 32: Could we add a little disclaimer here, something along the lines:
```
teliod daemonizes early; some logs may still appear on stdout/stderr before that.
```

Alternatively, in a follow-up PR we could use `freopen` and force redirect `stdout` and `stderr` before the daemon to append to the specified/default logfile if needed :)

clis/teliod/src/main.rs
```diff
@@ -72,6 +68,19 @@ struct DaemonOpts {
...
+    /// Defaults to "/".
+    /// Ignored with no_detach flag.
+    #[clap(long = "working_directory")]
+    working_directory: Option<String>,
+    /// Redirect standard output to the specified file
+    ///
+    /// Defaults to "/var/log/teliod.log".
+    /// Ignored with no_detach flag.
+    #[clap(long = "stdout_path")]
+    stdout_path: Option<String>,
```
LukasPukenis, line 83: Could we refactor this a bit and instead of comments about the default values, inline them with default values via clap decorator? That would require changing `Option<String>` to `String` but at the call sites we don't need to know if the value was provided or default, we just need a value. The current way, the comment is not tied to the value :)

clis/teliod/README.md
```diff
@@ -1,42 +1,67 @@
...
-  - `interface`
-    - `name` - Name of tunnel interface to connect to. Note that for macOS the name has to be in form `tun#` where `#` can be any integer number
-    - `config_provider` - Provider for configuring the interface address, possible options:
-      - `manual` - do not configure interfaces automatically
-      - `ifconfig` - systems using ifconfig command
-      - `iproute` - systems using iproute2 command
+- `teliod start <path_to_config_file>` - starts the daemon.
+  - `--no-detach` - run the teliod in the foreground as a regular process,
+  without detaching from the terminal.
+  - `--stdout_path` - Redirect standard output to the specified file,
```
lcruz99, line 29: underscore should not be used as some terminals don't show it clearly, a dash should be used instead:  --stdout-path

https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

clis/teliod/README.md
```diff
@@ -1,42 +1,67 @@
...
+  Defaults to "/var/log/teliod.log", ignored when used with `--no-detach`
+  - `--working_directory` - Specifies the daemons working directory,
```
lcruz99, line 31: same comment as above

clis/teliod/README.md
tomasz-grz: Good point! also consistency with no-detach

clis/teliod/src/main.rs
tomasz-grz: Very good idea, the default values come from the `daemonize` crate, but that's not so obvious. This should make it a bit more "self documenting"

clis/teliod/README.md
tomasz-grz: Renamed

clis/teliod/README.md
tomasz-grz: Renamed

clis/teliod/README.md
tomasz-grz: Added a note

clis/teliod/src/main.rs
tomasz-grz: Converted to `clap` defaults

clis/teliod/src/main.rs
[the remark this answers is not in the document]
LukasPukenis: I see there's confusion about it, I added it as a `nit` hoping it's a super cheap add but if it isn't I am fine with it, we will fix it once we ecounter an issue later on(if so) :)

lcruz99 (approved):
+1

PR #1209, opened 2025-04-16
Add feature to modify SO buffer size in neptun

Hasan6979:
### Problem
Tests done by Android team showed that modifying SO_SND and SO_RECV gives a big boost in speed, hence a need to be able to modify buffer sizes.

### Solution
Add socket buffer size as a feature to be configurable by upstream Apps.

Tests regarding socket buffer sizes have been added in NepTUN. Corresponding NepTUN merged PR
https://github.com/NordSecurity/NepTUN/pull/29. for reference.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-6095
+++ b/.unreleased/LLT-6095
@@ -1,0 +1,1 @@
+skt_buffer_size parameter added in FeaturesWirguard to configure NepTUN socket buffer size
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "telio"
-version = "5.3.0"
+version = "5.4.0"
 authors = ["info@nordvpn.com"]
 edition = "2018"
 license = "GPL-3.0-only"
@@ -174,7 +174,7 @@ windows = { version = "0.60", features = [
     "Win32_NetworkManagement_IpHelper",
 ] }
 
-neptun = { git = "https://github.com/NordSecurity/neptun.git", tag = "v1.0.6", features = ["device"] }
+neptun = { git = "https://github.com/NordSecurity/neptun.git", tag = "v1.0.7", features = ["device"] }
 x25519-dalek = { version = "2.0.1", features = ["reusable_secrets", "static_secrets"] }
 
 
--- a/crates/telio-wg/src/adapter.rs
+++ b/crates/telio-wg/src/adapter.rs
@@ -172,7 +172,7 @@ pub enum Error {
 }
 
 /// Enumeration of types for `Adapter` struct
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum AdapterType {
     /// NepTUN
     NepTUN,
@@ -234,6 +234,7 @@ pub(crate) async fn start(cfg: &Config) -> Result<Box<dyn Adapter>, Error> {
                 cfg.firewall_process_inbound_callback.clone(),
                 cfg.firewall_process_outbound_callback.clone(),
                 cfg.firewall_reset_connections.clone(),
+                cfg.skt_buffer_size,
             )?))
         }
         AdapterType::LinuxNativeWg => {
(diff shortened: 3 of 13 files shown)


.unreleased/LLT-6095
Jauler, line 1: Please add some message for the changelog :pray: This is very well integrators-facing change.

tomaszklak (commented):
Please add a nat-lab test that uses custom buffer size

Cargo.toml
```diff
@@ -174,7 +174,7 @@ windows = { version = "0.60", features = [
     "Win32_NetworkManagement_IpHelper",
 ] }
 
-neptun = { git = "https://github.com/NordSecurity/neptun.git", tag = "v1.0.6", features = ["device"] }
+neptun = { git = "https://github.com/NordSecurity/neptun.git", tag = "v1.0.7", features = ["device"] }
```
Jauler, line 177: I think you should not need any NepTUN changes, as NepTUN sockets are created by `telio-sockets`. 

Is there a good reason why we are setting buffers in neptun vs libtelio?

Hasan6979: As far as my understanding of the ticket, the issue ticket describes setting it in NepTUN, this was discussed in refinement as well and NepTUN was supposed to offer an API to set the custom value of the sockets.

src/libtelio.udl
```diff
@@ -526,6 +526,10 @@ interface FeaturesDefaultsBuilder {
     /// Enable dynamic WireGuard-NT control as per RFC LLT-0089
     [Self=ByArc]
     FeaturesDefaultsBuilder enable_dynamic_wg_nt_control();
+
+    /// Enable custom socket buffer sizes for NepTUN
+    [Self=ByArc]
+    FeaturesDefaultsBuilder enable_custom_skt_buffer_size(u32 skt_buffer_size);
```
tomaszklak, line 532: nit: I would call it `set_skt_buffer_size`

Cargo.toml
Jauler: After slack discussion, yeah, probably having it set in NepTUN gives us a bit more fine-grained control instead of using telio-sockets.

nat-lab/tests/test_neptun.py
```diff
@@ -0,0 +1,47 @@
...
+            exit_stack, [alpha_setup_params, beta_setup_params]
+        )
+        alpha, beta = env.nodes
+        connection_alpha, connection_beta = [
+            conn.connection for conn in env.connections
+        ]
+
+        # Ping to check connection works both ways
+        await ping(connection_alpha, beta.ip_addresses[0])
+        await ping(connection_beta, alpha.ip_addresses[0])
```
Jauler, line 47: Would it be possible to somehow verify whether socket buffers has actually been set :thinking:.

Even something simple, like logging currently used bufffer size at startup in NepTUN and verifying whether it matches the feature flag. As I am not sure we will be able to observe the desired effect directly.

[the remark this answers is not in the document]
Hasan6979: ✅

nat-lab/tests/test_neptun.py
```diff
@@ -0,0 +1,47 @@
...
+@pytest.mark.parametrize(
+    "beta_setup_params",
+    [
+        pytest.param(
+            SetupParameters(
+                connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_2,
+            )
+        )
+    ],
+)
```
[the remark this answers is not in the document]
Hasan6979: Removed

crates/telio-wg/src/adapter.rs
```diff
@@ -234,6 +234,7 @@ pub(crate) async fn start(cfg: &Config) -> Result<Box<dyn Adapter>, Error> {
                 cfg.firewall_process_inbound_callback.clone(),
                 cfg.firewall_process_outbound_callback.clone(),
                 cfg.firewall_reset_connections.clone(),
+                cfg.skt_buffer_size,
```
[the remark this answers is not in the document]
Hasan6979: Added a function to sanitize its `skt_buffer_size`

nat-lab/tests/test_neptun.py
Hasan6979: Added verification in logs when sockets are set.

PR #1215, opened 2025-04-23
Expose an api to change TUN interface at runtime

tomaszklak:
This new ffi api works only for the NepTUN adapter.

- [x] README.md is updated

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -644,6 +644,7 @@ impl Device {
     pub fn set_tun(&self, tun: i32) -> Result {
         self.async_runtime()?.block_on(async {
             task_exec!(self.rt()?, async move |rt| {
+                let _guard = rt.entities.network_monitor.pause();
                 Ok(rt.set_tun(tun).boxed().await)
             })
             .await?
--- a/src/device.rs
+++ b/src/device.rs
@@ -640,6 +640,16 @@ impl Device {
         })
     }
 
+    /// Set the (u)tun file descriptor to be used by the adapter
+    pub fn set_tun(&self, tun: i32) -> Result {
+        self.async_runtime()?.block_on(async {
+            task_exec!(self.rt()?, async move |rt| {
+                Ok(rt.set_tun(tun).boxed().await)
+            })
+            .await?
+        })
+    }
+
     /// Retrieves currently configured private key for the interface
     pub fn get_private_key(&self) -> Result<SecretKey> {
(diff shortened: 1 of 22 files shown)


src/device.rs
```diff
@@ -640,6 +640,17 @@ impl Device {
         })
     }
 
+    /// Set the (u)tun file descriptor to be used by the adapter
+    pub fn set_tun(&self, tun: i32) -> Result {
+        self.async_runtime()?.block_on(async {
+            task_exec!(self.rt()?, async move |rt| {
+                let _guard = rt.entities.network_monitor.pause();
```
Jauler, line 647: Is network monitor "unpaused" anywhere? I missed this part :thinking:

Jauler: Oh its handled at the `_guard` drop. Nice. I missed this inittialy

Jauler:
I do not see anything out-of-order. But I wonder, maybe it would make sense to ask for an in-branch test from android team? 

@tomaszklak what do you think?

src/device.rs
```diff
@@ -1642,6 +1653,22 @@ impl Runtime {
             .await?;
         Ok(())
     }
+    async fn set_tun(&mut self, tun: i32) -> Result {
+        self.requested_state.device_config.name = None;
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
```
Jauler, line 1658: Should'nt android be in this list as well?

Also maybe it would make sense to check if neptun is our current WG adapter of choice and return something like "unsupported" in other cases in order to minimize silent failures?

tomaszklak: Yup, it should - fixed.

The check for adapter doesn't need happen here - the error will be propagated from the non NepTUN adapter in this call:

```
self.entities.wireguard_interface.set_tun(tun).await?;
```

Jauler (approved):
+1.0 once android in-branch test goes through :)

Jauler (changes requested):
One small request.

As this is going to be released separatly from Socket Buffer changes. And this introduces new features in API. 

Could you please update `cargo.toml` version to v5.5.0

matislovas:
+1

Jauler (dismissed):
+1.0. Once we have results from in-branch test, if everything is ok - lets merge :)

matislovas (approved):

PR #1216, opened 2025-04-25
Fix race in test_vpn_plus_mesh

stalowyjez:
### Problem
There is a race between libtelio `Disconnected` event and it's assert

### Solution
Run the check in async task before the disconnecting code is actually run

- [x] README.md is updated


tomasz-grz (commented):
+1

mathiaspeters (approved):
+1

PR #1217, opened 2025-04-28
Fix derp restart test race condition

gytsto:
### Problem
When we restart derp server, race condition appears. Server can be started faster than we actually connect to any other server, therefore, connecting to same derp server again.

### Solution
Shutdown derp server, connect to other derp server and then bring back previous derp server

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_derp_connect.py
+++ b/nat-lab/tests/test_derp_connect.py
@@ -257,7 +257,7 @@ async def test_derp_reconnect_3clients(setup_params: List[SetupParameters]) -> N
 
 
 @asynccontextmanager
-async def restart_container(container_name: str):
+async def stop_container(container_name: str):
     try:
         os.system(f"docker stop {container_name}")
         yield
@@ -366,7 +366,7 @@ async def test_derp_restart(setup_params: List[SetupParameters]) -> None:
         #  [GW1]    [GW2]   [Symmetric-GW]
         #    |        |        |
         # [ALPHA]   [BETA]  [GAMMA]
-        async with restart_container("nat-lab-derp-01-1"):
+        async with stop_container("nat-lab-derp-01-1"):
             await asyncio.gather(
                 alpha_client.wait_for_state_derp(_DERP2_IP, [RelayState.CONNECTED]),
                 beta_client.wait_for_state_derp(_DERP2_IP, [RelayState.CONNECTED]),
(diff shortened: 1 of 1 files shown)


nat-lab/tests/test_derp_connect.py
```diff
@@ -358,14 +358,16 @@ async def test_derp_restart(setup_params: List[SetupParameters]) -> None:
...
-        os.system("docker restart nat-lab-derp-01-1")
+        os.system("docker stop nat-lab-derp-01-1")
 
         await asyncio.gather(
             alpha_client.wait_for_state_derp(_DERP2_IP, [RelayState.CONNECTED]),
             beta_client.wait_for_state_derp(_DERP2_IP, [RelayState.CONNECTED]),
             gamma_client.wait_for_state_derp(_DERP3_IP, [RelayState.CONNECTED]),
         )
 
+        os.system("docker start nat-lab-derp-01-1")
```
tomaszklak, line 369: Could you use context managers or try/finally to restore the docker state? Without it, any failure during the test will also break future tests that depend on the derp servers.

gytsto: :+1: good idea, converted to context manager

nat-lab/tests/test_derp_connect.py
```diff
@@ -256,6 +256,15 @@ async def test_derp_reconnect_3clients(setup_params: List[SetupParameters]) -> N
         await ping(gamma_connection, beta.ip_addresses[0])
 
 
+@asynccontextmanager
+async def restart_container(container_name: str):
```
tomaszklak, line 260: nit: I would call it `stop_container` since I think that contents of `__enter__` is what is used to name the context manager.

tomaszklak (dismissed):
+2

nat-lab/tests/test_derp_connect.py
```diff
@@ -256,6 +256,15 @@ async def test_derp_reconnect_3clients(setup_params: List[SetupParameters]) -> N
...
+    try:
+        os.system(f"docker stop {container_name}")
+        yield
+    finally:
+        os.system(f"docker start {container_name}")
```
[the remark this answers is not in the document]
gytsto: As written in PR description. Restarting container creates race condition, because server might get up sooner then we expected. And we expect for it to be up, once we connect to other derp server. That's I did with stop and start.
```
stop
wait for event on diff server
start
```
while previously
```
restart (and hope it doesn't start before we actually connect to other server)
wait for event on diff server
```

nat-lab/tests/test_derp_connect.py
gytsto: updated

nat-lab/tests/test_derp_connect.py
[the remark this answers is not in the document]
LukasPukenis: My bad, the `start` here will only happen once we leave the scope thus I can see how it fixes the race. Thanks.

LukasPukenis (commented):
+1

PR #1220, opened 2025-04-29
natlab: Misc improvements

lcruz99:
Fixes:
- Libvirt VMs routing was being configured only when binaries were pushed, this PR configures it as default for every new ssh connection.
- Exceptions on docker/ssh processes were not being raised against the context manager of the cleanup stage, making  pytest hang for the whole test timeout duration. This PR fixes it by running the processes inside a try/except/finally block.
- If telio fails to launch won't set the client proxy port, creating an infinite loop waiting for the port. This PR fixes it by additionally checking if the telio process is running.
- Other logging and duplicated code minor fixes.

Features:
- This PR adds link state check as part of the setup_mesh_nodes() helper function.
- Add enable/disable interface methods to router classes.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/helpers.py
+++ b/nat-lab/tests/helpers.py
@@ -13,6 +13,7 @@ from utils.bindings import (
     Features,
     Server,
     Config,
+    LinkState,
     RelayState,
     NodeState,
     PathType,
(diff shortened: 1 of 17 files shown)


nat-lab/tests/helpers.py
```diff
@@ -368,6 +369,8 @@ async def setup_mesh_nodes(
     env = await exit_stack.enter_async_context(
         setup_environment(exit_stack, instances, provided_api, prepare_vpn)
     )
+    if len(env.nodes) != len(env.clients):
```
tomaszklak, line 372: Shouldn't this check be done inside of `setup_environment`?

tomaszklak:
In e438f90f817eff3bcf84d19d5c9a5a5d908da399 I think "swollen" should be replaced by "swallowed" 🤔

nat-lab/tests/utils/connection/ssh_connection.py
```diff
@@ -46,7 +46,7 @@ async def new_connection(
         ip: str,
         tag: ConnectionTag,
         copy_binaries: bool = False,
-        reenable_nat=False,
+        reenable_nat=True,
```
tomaszklak, line 49: Is there any place that sets `reenable_nat` to `False`? If not, than please remove this parameter.

nat-lab/tests/helpers.py
lcruz99: yes, done

nat-lab/tests/utils/connection/ssh_connection.py
lcruz99: Agree, removed

nat-lab/tests/helpers.py
lcruz99: actually I've removed that check. I've added it previously to satisfy lints but somehow not needed anymore

tomaszklak (commented):
+1

Hasan6979 (commented):
looks good generally. 
Small change though, can you add description in the changelog file what is changed ?

Hasan6979 (approved):
+1

PR #1229, opened 2025-05-07
Official QNAP design

LukasPukenis:
Official QNAP. 

**THE PR looks huge but it's just `web.rs` which is of importance to view.** Other than that it's a cleanup and/or copy-paste job.

I shared the screenshots with the designers and they told it looks good so at least I have some confidence. Functionally it hasn't changed, it still functions, toggling reveals the bottom panel which is refreshed every 2seconds and shows peers.

Notable things missing(ticketed):
- Device icons
- Loader animation when toggling(I removed it since it requires some effort to make it look good in new design).

Left is design reference, Right is QNAP:
image

- [x] README.md is updated


clis/teliod/src/cgi/web.rs
```diff
@@ -32,27 +33,29 @@ lazy_static! {
     static ref ASSETS: HashMap<&'static str, (&'static str, &'static [u8])> = {
         [
             ("/static/telio.js", ("text/javascript", asset!("telio.js"))),
+            ("/static/style.css", ("text/css", asset!("style.css"))),
+            ("/static/output.css", ("text/css", asset!("output.css"))),
             (
-                "/static/spinner.svg",
-                ("image/svg+xml", asset!("spinner.svg")),
+                "/static/tailwind.config.js",
```
LukasPukenis, line 39: For now spinner is removed since it clashes with the toggle button introduced but it is ticketed as #6237

clis/teliod/src/cgi/web.rs
```diff
@@ -110,55 +110,25 @@ fn index(base_ref: String) -> Markup {
...
-                    "extend: {"
-                    "colors: {"
-                    "'nord-blue': '#5E81AC',"
-                    "'nord-green': '#A3BE8C',"
-                    "'nord-orange': '#D08770'"
-                    "}" "}" "}" "}";
-                }
+                script src="static/tailwind.config.js" {}
+                link rel="stylesheet" href="static/output.css" {}
+                link rel="stylesheet" href="static/style.css" {}
```
LukasPukenis, line 115: We could potentially merge these two together but this was provided by the designers so I feel like it's OK to leave them separated for now while it's WIP.

clis/teliod/assets/style.css
```diff
@@ -18,17 +750,72 @@ body {
...
     padding: 24px;
     margin: 40px 24px 0px;
     border-radius: 12px;
     overflow: hidden;
     text-align: left;
+}
+
+/* inter-100 - latin */
+@font-face {
+    font-display: swap;
```
tomaszklak, line 770: If anyone is reading this and wants to know why not `auto` - we want to avoid FOIT and prefer FOUT, see for example here: https://www.zachleat.com/web-fonts/foitfout/#4000,4000,4000,4000

clis/teliod/src/cgi/web.rs
```diff
@@ -222,114 +197,76 @@ fn update_config(app: &mut AppState, request: &CgiRequest) {
...
...066 20.7107 16.4066 20.1301C16.4066 19.4895 16.1964 18.899 15.8561 18.4085L18.3484 15.9161C18.8189 16.2264 19.3894 16.4066 19.99 16.4066C20.5906 16.4066 21.1711 16.2164 21.6516 15.9061L24.1439 18.3985C23.7936 18.8889 23.5934 19.4795 23.5934 20.1301C23.5934 20.7107 23.7636 21.2512 24.0539 21.7116L21.6616 24.1039C21.1811 23.7836 20.6106 23.6034 20 23.6034V23.5934ZM20 11.602C20.9909 11.602 21.8017 12.4128 21.8017 13.4037C21.8017 14.3947 20.9909 15.2054 20 15.2054C19.0091 15.2054 18.1983 14.3947 18.1983 13.4037C18.1983 12.4128 19.0091 11.602 20 11.602ZM11.602 20.1301C11.602 19.1392 12.4128 18.3284 13.4037 18.3284C14.3947 18.3284 15.2054 19.1392 15.2054 20.1301C15.2054 21.1211 14.3947 21.9318 13.4037 21.9318C12.4128 21.9318 11.602 21.1211 11.602 20.1301ZM20 28.398C19.0091 28.398 18.1983 27.5872 18.1983 26.5963C18.1983 25.6053 19.0091 24.7946 20 24.7946C20.9909 24.7946 21.8017 25.6053 21.8017 26.5963C21.8017 27.5872 20.9909 28.398 20 28.398ZM26.5963 21.9318C25.6053 21.9318 24.7946 21.1211 24.7946 20.1301C24.7946 19.1392 25.6053 18.3284 26.5963 18.3284C27.5872 18.3284 28.398 19.1392 28.398 20.1301C28.398 21.1211 27.5872 21.9318 26.5963 21.9318Z" class="dark:fill-blue-900 fill-blue-600" {}
```
tomaszklak, line 222: why is this not in a separate `.svg` file and `include`ed?

tomaszklak:
How can I run it locally to test the ui?

clis/teliod/assets/style.css
LukasPukenis: I went to https://gwfh.mranftl.com/fonts/inter?subsets=latin and selected appropriate font with appropriate weights and selected Modern browsers and pasted the resulting font related styles in here. To be honest I am not even sure what FOIT and FOUT are and how relevant are they here. If you could explain that would be great!

clis/teliod/src/cgi/web.rs
LukasPukenis: You're definitely not wrong :) However it's supposed to be temporary icon since we already have a task to embed `per-os`+`per-device-type` icons(and we have svg icon files on the ticket) so I didn't bother extracting it in here..

clis/teliod/assets/style.css
tomaszklak: Oh, I wasn't aware of those terms also just a few moments ago! Basically as far as I understand the auto behvaiour means that for a split second there will be no text displayed at all and once the font is loaded up, correctly styled text will appear - **F**lash **O**f **I**nvisible **T**ext.
On the other hand, the `swap` means in practice, browser will first render the text with a best matching already available, and once the correct font is loaded up, it will change the styling of already rendered text - **F**lash **O**f **U**nstyled **T**ext.
The expectation is that in the case of FOIT the change is much bigger and more jarring, and in the case of FOUT it is much less so.
The link I posted simulates this behaviour with configurable much bigger times, so that it's easier to understand :)

tomaszklak:
> @tomaszklak re: running locally - Sure thing! `just web` and then head to the page. Just remove the rightmost `/` from the URL and you should see the page.
> 
> Just note that in the development version, for every request it does `cargo build` in the CGI config, which means every request is very slow

It doesn't work on macOS so I will have to trust you screeshots.

+0.5

clis/teliod/assets/style.css
LukasPukenis: Oh, thanks for explaining <3 TIL! :) The link is pretty good by the way to simulate the behaviour

clis/teliod/src/cgi/web.rs
```diff
@@ -32,27 +33,33 @@ lazy_static! {
     static ref ASSETS: HashMap<&'static str, (&'static str, &'static [u8])> = {
```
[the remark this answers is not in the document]
LukasPukenis: Current HashMap is from stdlib so less dependencies(I think in this case where we have just few entries and this would fall into microoptimisation it's enough of a counter argument) :)

clis/teliod/src/cgi/web.rs
```diff
@@ -72,13 +79,11 @@ pub fn handle_web_ui(request: &CgiRequest) -> Option<Response> {
             let base_ref = format!("{}/teliod.cgi", request.uri().path());
```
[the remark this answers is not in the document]
LukasPukenis: `handle_web_ui()` hides any failures it found as `5xx` http status code with an error message, all encoded into `Response`. `Option` part is just so if it doesn't handle the path at all. I think this design makes sense and is easy to follow :)

clis/teliod/src/cgi/web.rs
```diff
@@ -222,114 +197,76 @@ fn update_config(app: &mut AppState, request: &CgiRequest) {
...
-                @for option in log_options {
-                    @if option == level.as_str() {
-                        option value=(option) selected = "true" {
-                            (option)
+            // Inputs
+            form id="config" hx-post="meshnet" hx-target="body" class="mb-0" {
+                div class="flex flex-col gap-4" {
+                    div class="space-y-2" {
+                        label class="block text-primary body-xs-medium" { "Access Token" }
+                        input name=(ACCESS_TOKEN) type="password" class="w-full px-4 py-3 bg-transparent text-primary border border-input rounded-sm focus-visible:outline-none focus-visible:shadow-focus" value=(app.config.authentication_token) "hx-on:htmx:validation:validate"="telio.validateToken(this)";
```
tomasz-grz, line 245: Is there some result `validateToken` function?
If I give it some garbage for token and try to set the toggle to On, it doesn't move, would be nice to have some error message that the token is invalid, is this covered by the [other PR](https://github.com/NordSecurity/libtelio/pull/1233)?

tomasz-grz (commented):
Other wise seems to work, and looks good 🚀

lcruz99 (approved):
+1

PR #1230, opened 2025-05-07
Add retries for fetching external resources

olekoliinyk:
### Problem

### Solution


ci/fetch_artifacts.py
```diff
@@ -47,8 +48,13 @@ def _extract_date(self, tag):
 
     def _get_latest_tag(self):
         """Find the latest tag with the given prefix."""
-        subprocess.run(
-            ["git", "-C", self.repo_dir, "fetch", "--tags", "--quiet"], check=True
+        run_with_retry(
+            lambda: subprocess.run(
+                ["git", "-C", self.repo_dir, "fetch", "--tags", "--quiet"],
+                stderr=subprocess.STDOUT,
```
gytsto, line 54: why pipe stderr to stdout? if this function throws an exception, will there be anything in exception stderr, that is printed in `run_with_retry` ?

olekoliinyk: Agree, nothing will be there (checked also locally).
Pushed the fix.

ci/moose_utils.py
```diff
@@ -44,8 +44,12 @@ def _download_moose_file(target_os: str, arch: str, file_name: str):
 
     url = f"{nexus_url}/repository/ll-gitlab-release/{MOOSE_PROJECT_ID}/{LIBTELIO_ENV_MOOSE_RELEASE_TAG}/bin/common/{target_os}/{arch}/{file_name}"
 
-    subprocess.check_call(
-        ["curl", "-f", "-u", nexus_credentials, url, "-o", output_path]
+    run_with_retry(
+        lambda: subprocess.check_output(
+            ["curl", "-f", "-u", nexus_credentials, url, "-o", output_path],
+            stderr=subprocess.STDOUT,
```
gytsto, line 50: same applies here

gytsto (dismissed):
+2

gytsto (approved):

PR #1231, opened 2025-05-08
Implement suspend-aware replacement for `tokio::time::Instant`

tomaszklak:
### Problem
Both `std::time::Instant` and `tokio::time::Instant` do not track the time spent by the operating system in suspend mode. This can lead to telio not reacting in time to various events or timeouts.

### Solution
Add a custom suspend-aware replacement `Instant` implementation. The key points:
- most of the work is done by already existing create `boot_time` which is used in release builds
- in debug builds, this new `Instant` is using `tokio::time::Instant` which means that time suspending used in rust tests is still working and can be used
- `boot_time` crate doesn't compile for Linux 32bit systems, which is why for now we use our own fork with a fix for those platform
- this PR only tackles time measurement, notifications via `tokio::time::Interval` are left unchanged - this might be fixed in a future PR

- [x] README.md is updated

### DIFF
--- a/.github/workflows/nagger.yml
+++ b/.github/workflows/nagger.yml
@@ -5,7 +5,7 @@ permissions: {}
 jobs:
   nagger:
     runs-on: ubuntu-22.04
-    container: ghcr.io/nordsecurity/nagger:v4.0.0
+    container: ghcr.io/nordsecurity/nagger:v4.1.0
     steps:
       - name: Checkout
         uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 #v4.1.6
--- a/crates/telio-utils/src/instant.rs
+++ b/crates/telio-utils/src/instant.rs
@@ -16,6 +16,7 @@ pub struct Instant(InnerInstant);
 impl Instant {
     /// Returns an instant corresponding to “now”.
     pub fn now() -> Self {
+        #[allow(instant)]
         Self(InnerInstant::now())
     }
 
--- a/crates/telio-utils/src/instant.rs
+++ b/crates/telio-utils/src/instant.rs
@@ -1,0 +1,95 @@
+use std::time::Duration;
+
+const CHECK_PERIOD: Duration = Duration::from_secs(60);
+
+/// Wrapper for the `tokio::time::Instant`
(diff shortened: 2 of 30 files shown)


.github/workflows/nagger.yml
```diff
@@ -5,7 +5,7 @@ permissions: {}
 jobs:
   nagger:
     runs-on: ubuntu-22.04
-    container: ghcr.io/nordsecurity/nagger:v4.0.0
+    container: ghcr.io/nordsecurity/nagger:instant
```
tomaszklak, line 8: 🚨 THIS NEEDS TO BE UPDATED BEFORE MERGE 🚨

mathiaspeters:
+1

crates/telio-utils/src/instant.rs
```diff
@@ -0,0 +1,103 @@
+use std::time::Duration;
+
+const CHECK_PERIOD: Duration = Duration::from_secs(60);
+
+#[cfg(debug_assertions)]
+type InnerInstant = tokio::time::Instant;
```
Jauler, line 6: Just to double check - Different `Instant`'s are used in order to enable time jumps in unit tests?

crates/telio-utils/src/instant.rs
```diff
@@ -0,0 +1,103 @@
...
+    fn sub(self, rhs: Instant) -> Self::Output {
+        self.0 - rhs.0
+    }
+}
+
+/// Suspend aware replacement for tokio::time::sleep_until. In case
+/// operating system was suspended long enough, so that `instant` is
+/// in the past when operating system resumes - in that case `sleep_until`
+/// will complete within 60 seconds after the resume.
+pub async fn sleep_until(instant: Instant) {
```
Jauler, line 83: I wonder, would there be any benefit of allowing `CHECK_PERIOD` to be specified by the caller of `sleep_until()`? 

I guess the callers would have more information on how "sensitive" they are with respect to incorrect sleeping time. But for sure we need some `default` where callers do not care about strict accuracy.

crates/telio-utils/src/instant.rs
tomaszklak: Yes, this is exactly why it's done this way.

PR #1232, opened 2025-05-08
telio-wg: Remove WireguardGo support

lcruz99:
Windows is dropping `wireguard-go` support thus we decide to also drop and stop maintaning it.

This PR removes the `wireguard-go` adapter from telio-wg, the nat-lab tests and the respective wrapper crate.

- [x] README.md is updated


crates/telio-utils/src/utils.rs
```diff
@@ -69,6 +69,21 @@ pub fn format_hex(bytes: &[u8]) -> String {
         .join(" ")
 }
 
+/// Allow dead code if the target_os is Windows
+#[macro_export]
+macro_rules! allow_dead_code_windows {
```
jjanowsk, line 74: If this is a deadcode why do we want to allow it? Why not just remove it on windows and leave it on other platforms with simple `#[cfg(not(target_os = "windows"))]`?

jjanowsk (changes requested):
Please also remove references to wireguard-go from Cargo.toml, .gitignore and cli.rs

crates/telio-utils/src/utils.rs
lcruz99: This function is used in mod tests and ran in Windows.

jjanowsk: So: `#[cfg(any(not(target_os = "windows"), test))]`

jjanowsk: Or even shorter: `#[cfg(any(not(windows), test))]`

lcruz99: oh you're right! will change

jjanowsk (approved):
+2

Justfile
```diff
@@ -116,3 +116,6 @@ _rust1_85-install:
 
 _rust-from-toolchain-file-install:
     rustup toolchain install
+
+_rust1_85-clippy-install:
+    rustup component add clippy --toolchain {{ rust1_85}}
```
jjanowsk, line 121: I think it should be just `clippy-install` without `--toolchain` parameter. When we run clippy we don not specify the version. We rely on toolchain file, and the same machanism should be used here IMO.

jjanowsk (changes requested):

Justfile
lcruz99: Agreed! changes

jjanowsk (approved):
+2

PR #1233, opened 2025-05-08
Introduce failure feedback in QNAP UI

LukasPukenis:
Upon receiving POST request, after toggling
the daemon, now the result of it is transferred
back to the page so the user would get the feedback.

It's not very granular at the moment but at least it is
explicit about telling that something has failed instead
of just staying silent(refreshes the page but meshnet doesn't
turn on).

Below is the screenshot when incorrect token was entered and
meshnet was toggled:
image

### Problem

### Solution

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/cgi/api.rs
+++ b/clis/teliod/src/cgi/api.rs
@@ -37,8 +37,15 @@ macro_rules! teliod_blocking_query {
 
 pub(crate) fn handle_api(request: &CgiRequest) -> Option<Response> {
     match (request.method(), request.route()) {
-        (&Method::POST, "" | "/") => Some(start_daemon()),
-        (&Method::DELETE, "" | "/") => Some(stop_daemon()),
+        (&Method::POST, "" | "/") => {
+            let (status_code, body) = start_daemon();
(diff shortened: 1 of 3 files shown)


tomaszklak (commented):
Please add some tests for new functionality.

LukasPukenis:
@tomaszklak re: tests. Do you have anything specific in your mind you could share? One way to add those is to spin a CGI in NatLab obviously and at least check the responses when running teliod but I fear this is gonna be much larger task than all of the QNAP related effort combined. Though I thought about this one, just decided not to pursue it.

As for unit tests, that's not trivial either since the current architecture checks for actual `teliod` process being in action to check for failures.

The only possibility without exploding the complexity of this task I see is to do a unit test for comparing the emitted HTML based on the `AppState` and `error` argument but I doubt about it's usefullness.

tomaszklak (approved):
+1 given that the test are scheduled in jira to be added sometime later

tomasz-grz (dismissed):
+1

Hasan6979 (approved):

PR #1238, opened 2025-05-13
Increase the start of ephemeral port range in nat-lab

tomaszklak:
### Problem
When running nat-lab, sometimes wireguard got assigned a random port from the ephemeral port range that clashed with the port hardcoded for the starcast (13569). This is because we are overriding ephemeral port range in tests to avoid clashes between tests run on the same machine.

### Solution
Increase minimal ephemeral port range start so that it's greater than the starcast port.

- [x] README.md is updated


gytsto (approved):
+2

jjanowsk (approved):
+2

PR #1240, opened 2025-05-13
[LLT-5073] Link detection failure on Windows

lcruz99:
### Problem

### Solution


crates/telio-model/src/features.rs
```diff
@@ -112,7 +112,7 @@ impl FeatureWireguard {
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, SmartDefault)]
 #[serde(default)]
 pub struct FeaturePersistentKeepalive {
-    /// Persistent keepalive period given for VPN peers (in seconds) [default 15s]
+    /// Persistent keepalive period given for VPN peers (in seconds) [default 25s]
```
LukasPukenis, line 133: I would argue that mentioning the default values is not necessary in the comments at all since they are always prone to be outdated and are also redundant(the code is just one line below) :)

nat-lab/tests/utils/connection_util.py
```diff
@@ -339,3 +340,48 @@ def is_tag_valid_for_ssh_connection(tag: ConnectionTag) -> bool:
         ConnectionTag.VM_WINDOWS_2,
         ConnectionTag.VM_MAC,
     ]
+
+
+async def set_secondary_ifc_state(
```
LukasPukenis, line 347: There must be a function which is the same just for primary interface and it probably is different only in `eth1` versus `eth0`, if so it would be nice to DRY it :)

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -9,93 +9,240 @@
...
+        # Look for existing enabled/disabled interfaces
+        process = await connection.create_process(
+            ["netsh", "interface", "show", "interface"], quiet=True
+        ).execute()
+
+        stdout = process.get_stdout()
+        log.debug("[%s]: %s", connection.tag.name, stdout)
+
+        matches = re.finditer(
+            r"^(.+[a][b][l][e][d])[\s]+[\w]+[\s]+[\w]+[\s]+(.*$)",
```
LukasPukenis: An examplary output in the comments of the output would be nice. Then it's much easier to validate that the regex is valid and also helps to find differences when the output changes and it fails.

An even better way - I suggest writing few unit tests that parse the output and embed it there, this way we are very sure that the parsing works as expected at the cost of a few more lines of code :)

In addition I think the regex can be refactored since it looks for the word `abled`. Maybe it can be just
```
r".*\b(?:disabled|enabled)\b.*"
```

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -9,93 +9,240 @@
...
+        # Find their addresses for enabled interfaces (except loopback)
         process = await connection.create_process(
-            ["netsh", "interface", "ipv4", "show", "addresses"], quiet=True
+            ["netsh", "interface", "ipv4", "show", "addresses"],
+            quiet=True,
         ).execute()
 
         stdout = process.get_stdout()
-        log.debug(stdout)
+        log.debug("[%s]: %s", connection.tag.name, stdout)
```
LukasPukenis: minor nitpick: double indirection could be eliminated by just printing `connection.tag` or a `repr()` of it. My :two: cents

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -9,93 +9,240 @@
...
+                [
+                    "netsh",
+                    "interface",
+                    "ipv4",
+                    "delete",
+                    "route",
+                    "0.0.0.0/0",
+                    self.name,
+                ],
+                quiet=True,
```
LukasPukenis, line 162: Outside of scope for this PR but maybe a good place to start discussion:

I've tried to follow `quiet` argument to see what does it mean, expecting that it suppresses the output but rather in `docker_connection.py` it is
```
 if not quiet:
    log_level = INFO
 else:
    log_level = DEBUG
```

It isn't quite, it's maybe `quiter` and in addition, `Debug` will generate more logs since it's more verbose. If you do not have opinion on this, you can totally ignore the comment but if we can confirm that it's a bit confusing, we can mark it as `TODO` comment to be fixed in the future :)

nat-lab/tests/test_events_link_state.py
```diff
@@ -1,27 +1,40 @@
...
-# The time by which we otherwise would expect to emit at least one link-state event.
-IDLE_TIMEOUT = 2 * LINK_STATE_INTERVAL
+WG_POLLING_PERIOD_S = 1
+WG_PASSIVE_KEEPALIVE_S = 10
+MAX_RTT_ALLOWED_S = 1
+# The duration which after a link is considered not to be up. (ie: PossibleDown->Down)
+LINK_STATE_TIMEOUT_S = WG_PASSIVE_KEEPALIVE_S + MAX_RTT_ALLOWED_S
+POSSIBLE_DOWN_DELAY = 3
+# Enhanced detection (ED) issues a ping when there's the Up->PossibleDown link state transition,
+# if no ICMP reply is received for LINK_STATE_TIMEOUT_S the link state will transit to Down.
```
LukasPukenis: Any reply works, the point is to trigger the peer to send back any, or if he doesn't then peer's WG will generate passive-keepalive and send it back :)

nat-lab/tests/test_events_link_state.py
```diff
@@ -1,27 +1,40 @@
...
+POSSIBLE_DOWN_DELAY_ED = LINK_STATE_TIMEOUT_S
+# The time by which we would expect to emit at least one link-state event, when tx_ts > rx_ts and none rx packet for the same period.
+TOLERANCE = 1.5
+IDLE_TIMEOUT_S = round((LINK_STATE_TIMEOUT_S + POSSIBLE_DOWN_DELAY) * TOLERANCE)
+IDLE_TIMEOUT_ED_S = round((LINK_STATE_TIMEOUT_S + POSSIBLE_DOWN_DELAY_ED) * TOLERANCE)
```
LukasPukenis, line 40: I am not entirely sure if we need a `1.5` multiplier - link detection is a finite state machine with very clear path to transition between the states, the only wiggle-room there is is due to:
- VM<->VM comms
- NatLab sampling rate
- Libtelio sampling rate via UAPI

I would suggest to mention more explicitly why `1.5` is needed since I can sense that if these test are gonna be "flaky", a very natural thing to consider will be to just bump this number to a higher one. Maybe just add a constant of say `5` or `10` seconds and mention that it should be enough to include all the delays?

nat-lab/tests/test_events_link_state.py
```diff
@@ -41,93 +54,85 @@ def long_persistent_keepalive_periods() -> FeatureWireguard:
...
+    else:
+        features.wireguard = long_persistent_keepalive_periods()
 
     return [
         SetupParameters(
             connection_tag=tag,
             adapter_type_override=adapter,
             features=features,
+            # TODO: Remove this if you're adding a "vpn with mesh" test
+            is_meshnet=not vpn,
```
LukasPukenis, line 97: :thinking:

nat-lab/tests/test_events_link_state.py
```diff
@@ -299,24 +320,28 @@ async def test_event_link_state_peer_goes_offline(
         await ping(connection_alpha, beta.ip_addresses[0])
 
         await client_beta.stop_device()
+        # Stopping beta generates rx/tx traffic with alpha, so wait 1 second before the next ping to skip the current
+        # polling interval and therefore guarantee that only tx increases
+        await asyncio.sleep(WG_POLLING_PERIOD_S)
```
LukasPukenis, line 330: Could you explain the `generates rx/tx traffic with alpha` comment a bit as I find it confusing? Additionally, why do we need to pings alpha->beta?

nat-lab/tests/test_events_link_state.py
```diff
@@ -422,5 +450,429 @@ async def test_event_link_state_peer_doesnt_respond(
...
+        delay_s = resolve_idle_timeout(setup_params[0])
+        await exit_stack.enter_async_context(
+            add_outgoing_packets_delay(connection_beta, f"{delay_s}s")
+        )
+
+        # Waiting for the finish of the previous ping polling interval.
+        await asyncio.sleep(WG_POLLING_PERIOD_S)
+
+        # alpha will only receive the ICMP reply 20s after and
+        # in the meantime beta's link state will go DOWN.
```
LukasPukenis: I noticed that I was using the terms not entirely correct myself in LinkDetection context but LinkDetection makes sense only when talking about two peers and the perspective matters, so for example _alpha->beta_ might be **DOWN** but _beta->alpha_ might be **UP**.

Could we rephrase this comment to `in the meantime beta->alpha link state will go DOWN` to be explicit that this is what we mean? It would be nice to fix all the occurrences of this, if possible in this PR while we're at it.

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -150,111 +297,11 @@ async def switch_to_secondary_network(self) -> None:
...
-            while bool([
-                iface
-                for iface in await Interface.get_network_interfaces(self._connection)
-                if Interface(self._interfaces.default, "").ipv4 == iface.ipv4
-            ]):
-                await asyncio.sleep(0.1)
+        if self._mgmt_interface:
+            await self._mgmt_interface.disable(self._connection)
+        await self._primary_interface.delete_route(self._connection)
+        await self._secondary_interface.delete_route(self._connection)
```
LukasPukenis, line 362: *network_switcher_windows.py* has a number of changes and proving that they work correctly via separate tests would be very nice.

Having the utilities exercised in libtelio-based testcases also works but it also increases the surface area when inspecting problems so separate tests just for this component would be nice(both unit and integration tests). 

It would also make it easy to validate that the parsing is done right, regexes are correct, etc :)

crates/telio-model/src/features.rs
lcruz99: This is part of the documentation and IMO should be accurate. eg: in any other 3rd party client that use our bindings, when you hover your mouse over the feature this will be what's printed..

lcruz99: .. it actually led me to check these values because I knew that it was wrong

LukasPukenis: Hmmm, good point on documentation generation... So it's in the end a tradeoff between documentation and outdated values. Clear, thanks! :)

nat-lab/tests/utils/connection_util.py
lcruz99: This function stems from the fact that pytest connection with VMs and libtelio instances - through FFI (remote/proxy.py) - always rely on the primary interface so disabling is non practical and I couldn't think in a use case for it.

Nevertheless, I fully agree with DRYing it and [my intention was](https://github.com/NordSecurity/libtelio/blob/1bd4ca3d2ead3261ff80074b83a39bc97d7acc4a/nat-lab/tests/utils/connection_util.py#L376) to create a virtual interface class for Linux and macOS targets [as done for Windows](https://github.com/NordSecurity/libtelio/blob/1bd4ca3d2ead3261ff80074b83a39bc97d7acc4a/nat-lab/tests/utils/network_switcher/network_switcher_windows.py#L16).

nat-lab/tests/test_events_link_state.py
lcruz99: agreed, will change

nat-lab/tests/test_events_link_state.py
lcruz99: telio.stop_device() is graceful which means it'll notify the connected peers about it, if the next ICMP request is sent in the same polling interval as the stop_device() rx packet from beta, then ts_tx == ts_rx. This sleep avoids that.

> why do we need to pings alpha->beta?

you're referring to the ping before stop_device()?

nat-lab/tests/test_events_link_state.py
lcruz99: I just didn't bother to tweak this function further, I think having tests with external meshent peers might be trivial for link detection.

nat-lab/tests/test_events_link_state.py
lcruz99: > I can sense that if these test are gonna be "flaky", a very natural thing to consider will be to just bump this number to a higher one. 


You're right, I can imagine that happening.

lcruz99: However is having a constant instead any different from the multiplier?

lcruz99: In my experience, a DOWN link detection that should happen in theory in 14s without Enhanced Detection (keepalive timeout + rtt + delay) was taking around 18s (+33%) and on CI I'm finding the same values. With ED, theorical value is 21s and in practice is 26s (+23%)

Given this, having a 10s overtime without ED might be of concern, while with ED it might not. Hence, I chose to use a multiplier.

nat-lab/tests/test_events_link_state.py
lcruz99: Right, corrected!

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: > An even better way - I suggest writing few unit tests that parse the output and embed it there, this way we are very sure that the parsing works as expected at the cost of a few more lines of code :)

Agree, I'll add some tests.

> In addition I think the regex can be refactored since it looks for the word abled. Maybe it can be just
r".*\b(?:disabled|enabled)\b.*" 

That would capture just the state of the interface, the code is looking also for its name on the second capture group. It's a good suggestion though, I'll use the full word matching instead of spelling "abled"

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: sorry didn't get it, what do you mean?

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: I think I didn't get it. If you're running the tests with INFO level, this specific command won't be logged (quiet==true), however it will be printed if you're running with DEBUG level set in the logger.

It's quiet in the INFO level lense, but when you run the tests with DEBUG it will naturally be more verbose and print every log marked with quiet==true

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: Will add some testing

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: I've looked twice and understood now :D

Agreed will implement repr() or similar

nat-lab/tests/test_events_link_state.py
```diff
@@ -424,5 +454,430 @@ async def test_event_link_state_peer_doesnt_respond(
...
+            ),
+            marks=pytest.mark.windows,
+        ),
+    ],
+)
+async def test_event_link_detection_after_disabling_ethernet_adapter_direct_path(
+    setup_params: List[SetupParameters],
+) -> None:
+    if setup_params[0].connection_tag is ConnectionTag.VM_WINDOWS_1:
+        pytest.xfail("Expected failure on Windows, see LLT-5073")
```
lcruz99, line 699: xfail added here

nat-lab/tests/test_events_link_state.py
```diff
@@ -424,5 +454,430 @@ async def test_event_link_state_peer_doesnt_respond(
...
+            ),
+            marks=pytest.mark.windows,
+        ),
+    ],
+)
+async def test_event_link_detection_after_disabling_ethernet_adapter_with_vpn(
+    alpha_setup_params: List[SetupParameters],
+) -> None:
+    if alpha_setup_params[0].connection_tag is ConnectionTag.VM_WINDOWS_1:
+        pytest.xfail("Expected failure on Windows, see LLT-5073")
```
lcruz99, line 827: xfail added here

nat-lab/tests/test_network_switch.py
```diff
@@ -301,3 +307,308 @@ async def test_mesh_network_switch_direct(
...
+        mock_show_address_process.execute = AsyncMock(
+            return_value=mock_show_address_process
+        )
+
+        mock_connection.create_process.side_effect = [
+            mock_show_interface_process,
+            mock_show_address_process,
+        ]
+
+        interfaces = await Interface.get_network_interfaces(mock_connection)
```
LukasPukenis: nit: could we `import Interface as WinInterface`?

nat-lab/tests/test_network_switch.py
```diff
@@ -301,3 +307,308 @@ async def test_mesh_network_switch_direct(
         beta_client.allow_errors([
             "telio_traversal::endpoint_providers::stun.*Starting session failed.*A socket operation was attempted to an unreachable network"
         ])
+
+
+class TestInterfaceWindows:
```
tomaszklak, line 313: nit: Imo, there is no need for this commit. For example, those windows tests should be part of the commit that makes changes to the `network_switcher_windows.py`.

nat-lab/tests/test_events_link_state.py
tomaszklak: nit: Imo, putting xfail inside the test makes it easy to miss - I would expect that the initial few lines of many nat-lab tests are much less interesting than the rest of the test and so are often glanced over. It would be more readable if it was part of the `setup_params`

tomaszklak (commented):
+1 (really nice split into commits! 👍 )

nat-lab/tests/test_network_switch.py
lcruz99: Agreed, changed

nat-lab/tests/test_events_link_state.py
lcruz99: I agree, changed

nat-lab/tests/utils/connection/connection.py
```diff
@@ -128,7 +131,7 @@ async def setup_ephemeral_ports(connection: Connection):
         return
 
     async def on_output(output: str) -> None:
-        log.debug("[%s]: %s", connection.tag.name, output)
+        log.debug("[%s]: %s", connection.tag, output)
```
LukasPukenis, line 134: nit: could `connection.tag` itself have `__repr__` to print it's `name`?

nat-lab/tests/test_events_link_state.py
```diff
@@ -29,10 +29,12 @@
...
 POSSIBLE_DOWN_DELAY = 3
 # Enhanced detection (ED) issues a ping when there's the Up->PossibleDown link state transition,
-# if no ICMP reply is received for LINK_STATE_TIMEOUT_S the link state will transit to Down.
+# if no reply is received for LINK_STATE_TIMEOUT_S the link state will transit to Down.
 POSSIBLE_DOWN_DELAY_ED = LINK_STATE_TIMEOUT_S
-# The time by which we would expect to emit at least one link-state event, when tx_ts > rx_ts and none rx packet for the same period.
+# As observed, a 50% tolerance is more than enough to detect down link state, with or without ED, whether on personal setup or on CI.
+# If the tests happen to be flaky, this value shouldn't be increased.
 TOLERANCE = 1.5
+# The time by which we would expect to emit at least one link-state event, when tx_ts > rx_ts and none rx packet for the same period.
```
LukasPukenis: question: link detection itself is very deterministic - once data goes out, the timer starts and should emit link down once we reach 10seconds + rtt. This rtt is configurable since the wireguard protocol states that after 10seconds pass, wireguad will generate an empty packet to send, so basically `rtt` is time we give for this packet to reach us(though it's a UDP packet, it might not ever reach us).

So my question is - if we know pretty well when we expect the packet, why do we need a `TOLERANCE` multiplier? :pray:

nat-lab/tests/test_events_link_state.py
```diff
@@ -429,14 +431,14 @@ async def test_event_link_state_peer_doesnt_respond(
             conn.connection for conn in env.connections
         ]
 
-        # After this ping beta will have ts_tx > ts_rx (ICMP reply)
+        # After this ping beta's wireguard drivers will have ts_tx > ts_rx (link state feature will probably have ts_tx == ts_rx)
```
LukasPukenis, line 435: To make things more deterministic - could we eliminate the `ts_tx == ts_rx` case by sleeping for poll period(1s)?

nat-lab/tests/test_events_link_state.py
```diff
@@ -631,7 +633,7 @@ async def test_event_link_detection_after_disabling_ethernet_adapter(
             LinkState.UP,
             LinkState.DOWN,
         ]
-        # (*) At this stage, we cannot ensure the link state on Beta due to the counters update rate (1s == polling period).
+        # (*) At this stage, we cannot avail beta->alpha link state due to the counters update rate (polling period is 1s).
```
LukasPukenis, line 637: Nice refinement, it clearly shows that the link state is per connection. :+1:

nat-lab/tests/utils/connection_util.py
LukasPukenis: Thank you, very good explanation. Could we have it as a comment on top then that we cannot disable primary interface(as it doesn't make sense in our setup). Just to help the future readers?(such as me) :)

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
LukasPukenis: My point could be simplified down to - we have log levels already and I do not really see a need to have them dynamic based on a boolean flag. However to reduce the noise I will just resolve the comment :)

nat-lab/tests/utils/connection_util.py
```diff
@@ -339,3 +340,48 @@ def is_tag_valid_for_ssh_connection(tag: ConnectionTag) -> bool:
...
+                        secondary_ifc = interface
+        assert secondary_ifc, LookupError("Couldn't find secondary interface")
+        if enable:
+            await secondary_ifc.enable(connection)
+        else:
+            await secondary_ifc.disable(connection)
+        return secondary_ifc
+
+    # TODO: Create Interface class for Linux/Macos
+    return None
```
LukasPukenis, line 379: The function is marked to return `Optional[Interface]` yet it only returns when `target_os == Windows`. I think(guess) it is returned for chaining support, but in that case we should return it regardless and raise exceptions when unexpected happens :)

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
```diff
@@ -9,93 +9,241 @@
...
+STATUS_CHECK_TIMEOUT_S: float = 20.0
+
 
 @dataclass
 class Interface:
-    def __init__(self, name: str, ipv4: str) -> None:
-        self.name = name
-        self.ipv4 = ipv4
+    name: str
+    enabled: bool
```
LukasPukenis: I am a bit suspicious about this parameter - in `disable()` it is set to `false` but nothing is done in `enable()` which makes it a bit weird :).

Furthermore the passed value is simply stored but it doesn't represent real state of the `Interface`. The correct way would be to extract the state when we need it(since it can also change without our influence). Could you explain the reasoning for this parameter/argument?

nat-lab/tests/utils/connection/connection.py
LukasPukenis: Oh it already is, nevermind

nat-lab/tests/utils/connection_util.py
lcruz99: added comment

nat-lab/tests/test_network_switch.py
lcruz99: sure, changed

nat-lab/tests/test_events_link_state.py
lcruz99: you're right, it should be deterministic, in theory.. In practice I've observed a small variance in the detection times: 14-21s.

I've commented it [here](https://github.com/NordSecurity/libtelio/blob/32dda575db4affe653c93b7f7630f80e5a102435/nat-lab/tests/test_events_link_state.py#L441), maybe this should be included in this comments?

nat-lab/tests/test_events_link_state.py
lcruz99: No, that would simply move the polling window 1s forward and not guarantee that `ts_tx != tx_rs`. That would work if we had full control on which packets the node receives, in reality we can't know when will a packet arrive and thus `tx_rs` increase.

nat-lab/tests/utils/connection_util.py
lcruz99: nice catch, will do

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: In get_network_interfaces() we fetch the list of the interfaces and their state with the `netsh interface show interface` windows command.

> Furthermore the passed value is simply stored but it doesn't represent real state of the Interface. The correct way would be to extract the state when we need it(since it can also change without our influence).

Completely agree, I guess I've missed the fact that interfaces can change on the backstage. Will implement a getter/setter for this variable with proper real state tracking.

Nevertheless, `Interface::enable()`  should indeed set the `enabled` attribute to `True` 🤔  Will change.

lcruz99:
> @[lcruz99](https://github.com/lcruz99) lcruz99 deleted a comment from [LukasPukenis](https://github.com/LukasPukenis) [20 hours ago](https://github.com/NordSecurity/libtelio/pull/1240#event-19916768066)

it was dplicated 😖

nat-lab/tests/test_events_link_state.py
LukasPukenis: Sure why not :)

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
lcruz99: @LukasPukenis changed interface state tracking so that everytime the interface state is read an up-to-date value is given.

lcruz99:
@LukasPukenis updated Jira references and ticket tracking file. It wasn't possible to update the branch though.

nat-lab/tests/utils/network_switcher/network_switcher_windows.py
LukasPukenis: Looks nice :)

nat-lab/tests/utils/connection_util.py
lcruz99: actually it should return None while Linux/MacOS don't have the Interface class defined, that's why they have the secondary itnerface hardcoded ("eth1").

either way I've added this explanation to the comment.

LukasPukenis: nonblocking:

While it reads a bit unexpectedly now I see how it works, thanks. I still think that it would benefit from a small refctor and the easiest one would be to simply introduce interfaces for other OSes that would wrap `eth1` as a name and probably that would be enough. This would make it very straight forward.

By the way, why can we harcode for linux and macos but not for windows? I couldn't find a comment.

lcruz99: > still think that it would benefit from a small refctor and the easiest one would be to simply introduce interfaces for other OSes that would wrap eth1 as a name and probably that would be enough.

I agree and thought about that during development, left a TODO as it would require some more effort and deserves a separate PR.

> By the way, why can we harcode for linux and macos but not for windows? I couldn't find a comment.

Secondary interface have deterministic names for linux/macos but not Windows 🤷

PR #1241, opened 2025-05-14
Fix _get_artifacts usage

olekoliinyk:
### Problem
`unzip=False` was removed from `_get_artifacts ` hence `self._get_artifacts(job, unzip=True)` usage causing an error.

### Solution

- [x] README.md is updated


gytsto (approved):
+2

PR #1244, opened 2025-05-16
Lukasp/qnap small improvements

LukasPukenis:
### Problem
    Minor QNAP fixes
    
    * validation fixed, previously after initial bad input
    it would not trigger when clicked even if correct token
    was entered.
    
    * Inputs are disabled when meshnet running.
    
    * Release mode has INFO, WARN and ERROR severities
    selectable only.

### Solution

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/cgi/web.rs
+++ b/clis/teliod/src/cgi/web.rs
@@ -9,7 +9,7 @@ use rust_cgi::{
     Response,
 };
 use telio::telio_model::mesh::{Node, NodeState};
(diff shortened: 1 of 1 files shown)


Hasan6979 (commented):
It looks okay to me. But needs 👀  more familiar with this code hence +0.5

lcruz99 (approved):
+1.5

jjanowsk (approved):

PR #1254, opened 2025-05-21
LLT-6139: Implement VPN functionality for teliod

mathiaspeters:
`teliod` now needs to be able to connect to VPN and this is a first version.

The config is updated to allow passing the ip and public key of a server to connect to. Future updates would likely include more user-friendly options, but for this PR, this is how you connect to VPN.

Routing information is updated but only for the existing configuration options (`ip` and `ifconfig`). Basic routing is set up, along with disabling IPv6.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/README.md
+++ b/clis/teliod/README.md
@@ -60,6 +60,9 @@ needs be absolute, otherwise will be relative to `working-directory` when daemon
     - `manual` - do not configure interfaces automatically
     - `ifconfig` - systems using ifconfig command
     - `iproute` - systems using iproute2 command
+  - `vpn` - optional vpn config. If omitted, a VPN connection will not be established
+    - `server_endpoint` - The endpoint (ip:port) of the server/endpoint to connect to 
+    - `server_pubkey` - The public key of the server/peer to connect to
 
 And following cli commands:
 
--- a/clis/teliod/src/config.rs
+++ b/clis/teliod/src/config.rs
@@ -1,7 +1,9 @@
 use std::{
+    net::SocketAddr,
     num::NonZeroU64,
     path::{Path, PathBuf},
     str::FromStr,
+    sync::Arc,
 };
 
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
@@ -10,7 +12,11 @@ use std::fs;
 use tracing::{debug, info, level_filters::LevelFilter, warn, Level};
 use uuid::Uuid;
 
-use telio::{crypto::SecretKey, device::AdapterType, telio_utils::Hidden};
+use telio::telio_utils::Hidden;
+use telio::{
+    crypto::{PublicKey, SecretKey},
+    device::AdapterType,
+};
 
 use crate::{configure_interface::InterfaceConfigurationProvider, TeliodError};
 
@@ -106,15 +112,19 @@ pub struct TeliodDaemonConfig {
     pub log_file_count: usize,
     pub adapter_type: AdapterType,
     pub interface: InterfaceConfig,
+    pub vpn: Option<VpnConfig>,
 
     #[serde(
         deserialize_with = "deserialize_authentication_token",
         serialize_with = "serialize_authentication_token"
     )]
-    pub authentication_token: Hidden<String>,
+    pub authentication_token: Arc<Hidden<String>>,
 
     /// Path to a http pem certificate to be used when connecting to CoreApi
     pub http_certificate_file_path: Option<PathBuf>,
+
+    /// Path to a device identity file, should only be used for testing
+    pub device_identity_file_path: Option<PathBuf>,
 
     #[serde(default)]
     pub mqtt: MqttConfig,
@@ -133,7 +143,7 @@ impl TeliodDaemonConfig {
             self.log_file_count = log_file_count;
         }
         if let Some(authentication_token) = update.authentication_token {
-            self.authentication_token = authentication_token;
+            self.authentication_token = Arc::new(authentication_token);
         }
         if let Some(adapter) = update.adapter_type {
             self.adapter_type = adapter;
@@ -141,8 +151,15 @@ impl TeliodDaemonConfig {
         if let Some(interface) = update.interface {
             self.interface = interface;
         }
+        if let Some(vpn) = update.vpn {
+            self.vpn = Some(vpn);
+        }
         if let Some(http_certificate_file_path) = update.http_certificate_file_path {
             self.http_certificate_file_path = http_certificate_file_path;
+        }
+
+        if let Some(device_identity_file_path) = update.device_identity_file_path {
+            self.device_identity_file_path = device_identity_file_path
         }
         if let Some(mqtt) = update.mqtt {
             self.mqtt = mqtt;
@@ -184,7 +201,7 @@ impl TeliodDaemonConfig {
         if let Ok(token) = std::env::var("NORD_TOKEN") {
             println!("Overriding token from env");
             if token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit()) {
-                config.authentication_token = Hidden::<String>(token);
+                config.authentication_token = Arc::new(Hidden::<String>(token));
             } else {
                 eprintln!("Token from env not valid")
             }
@@ -219,8 +236,10 @@ impl Default for TeliodDaemonConfig {
                 name: "nlx".to_string(),
                 config_provider: Default::default(),
             },
-            authentication_token: "".to_string().into(),
+            vpn: None,
+            authentication_token: Arc::new(Hidden("".to_string())),
             http_certificate_file_path: None,
+            device_identity_file_path: None,
             mqtt: MqttConfig::default(),
         }
     }
@@ -266,11 +285,11 @@ where
 
 fn deserialize_authentication_token<'de, D: Deserializer<'de>>(
     deserializer: D,
-) -> Result<Hidden<String>, D::Error> {
+) -> Result<Arc<Hidden<String>>, D::Error> {
     let raw_string: Hidden<String> = de::Deserialize::deserialize(deserializer)?;
     let re = regex::Regex::new("[0-9a-f]{64}").map_err(de::Error::custom)?;
     if raw_string.is_empty() || re.is_match(&raw_string) {
-        Ok(raw_string)
+        Ok(Arc::new(raw_string))
     } else {
         Err(de::Error::custom("Incorrect authentication token"))
     }
@@ -301,6 +320,12 @@ pub struct InterfaceConfig {
     pub config_provider: InterfaceConfigurationProvider,
 }
 
+#[derive(PartialEq, Eq, Deserialize, Serialize, Debug, Copy, Clone)]
+pub struct VpnConfig {
+    pub server_endpoint: SocketAddr,
+    pub server_pubkey: PublicKey,
+}
+
 #[allow(dead_code)]
 #[derive(Deserialize, Debug, Default)]
 pub struct TeliodDaemonConfigPartial {
@@ -310,10 +335,12 @@ pub struct TeliodDaemonConfigPartial {
     pub log_file_count: Option<usize>,
     pub adapter_type: Option<AdapterType>,
     pub interface: Option<InterfaceConfig>,
+    pub vpn: Option<VpnConfig>,
     pub app_user_uid: Option<Uuid>,
     #[serde(default, deserialize_with = "deserialize_partial_authentication_token")]
     pub authentication_token: Option<Hidden<String>>,
     pub http_certificate_file_path: Option<Option<PathBuf>>,
+    pub device_identity_file_path: Option<Option<PathBuf>>,
     pub mqtt: Option<MqttConfig>,
 }
 
@@ -400,11 +427,14 @@ mod tests {
                 name: "utun10".to_owned(),
                 config_provider: InterfaceConfigurationProvider::Manual,
             },
-            authentication_token:
+            vpn: None,
+            authentication_token: Arc::new(
                 "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                     .to_owned()
                     .into(),
+            ),
             http_certificate_file_path: None,
+            device_identity_file_path: None,
             mqtt: MqttConfig {
                 backoff_initial: NonZeroU64::new(1).unwrap(),
                 backoff_maximal: NonZeroU64::new(300).unwrap(),
--- a/clis/teliod/src/configure_interface.rs
+++ b/clis/teliod/src/configure_interface.rs
@@ -1,22 +1,47 @@
 use crate::TeliodError;
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::net::IpAddr;
 use std::process::Command;
-use tracing::{error, info};
+use tracing::{debug, error, info};
 
 pub trait ConfigureInterface {
+    /// Initialize the interface
+    fn initialize(&mut self) -> Result<(), TeliodError>;
     /// Configure the IP address and routes for a given interface
-    fn set_ip(&self, _adapter_name: &str, _ip_address: &IpAddr) -> Result<(), TeliodError>;
+    fn set_ip(&mut self, ip_address: &IpAddr) -> Result<(), TeliodError>;
+    /// Configure routes for exit routing
+    fn set_exit_routes(&mut self, exit_node: &IpAddr) -> Result<(), TeliodError>;
+    /// Some of the configured routes are not cleared when the adapter is removed and must be removed manually
+    fn cleanup_exit_routes(&mut self) -> Result<(), TeliodError>;
 }
 
 /// Helper function to execute a system command
 fn execute(command: &mut Command) -> Result<(), TeliodError> {
+    debug!("Executing command '{command:?}'");
     let output = command
         .output()
         .map_err(|e| TeliodError::SystemCommandFailed(e.to_string()))?;
 
     if output.status.success() {
         Ok(())
+    } else {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        error!("Error executing command {:?}: {:?}", command, stderr);
+        Err(TeliodError::SystemCommandFailed(stderr.into()))
+    }
+}
+
+/// Helper function to execute a system command with output
+fn execute_with_output(command: &mut Command) -> Result<String, TeliodError> {
+    debug!("Executing command with output '{command:?}'");
+    let output = command
+        .output()
+        .map_err(|e| TeliodError::SystemCommandFailed(e.to_string()))?;
+
+    if output.status.success() {
+        String::from_utf8(output.stdout)
+            .map_err(|e| TeliodError::SystemCommandFailed(e.to_string()))
     } else {
         let stderr = String::from_utf8_lossy(&output.stderr);
         error!("Error executing command {:?}: {:?}", command, stderr);
@@ -36,12 +61,12 @@ pub enum InterfaceConfigurationProvider {
 // TODO(tomasz-grz): Try using enum_dispatch instead of dynamic dyspatch
 impl InterfaceConfigurationProvider {
     /// Create a dynamic instance of a configuration provider
-    pub fn create(&self) -> Box<dyn ConfigureInterface> {
+    pub fn create(&self, adapter_name: String) -> Box<dyn ConfigureInterface> {
         info!("Creating interface config provider for {:?}", self);
         match self {
             Self::Manual => Box::new(Manual),
-            Self::Ifconfig => Box::new(Ifconfig),
-            Self::Iproute => Box::new(Iproute),
+            Self::Ifconfig => Box::new(Ifconfig::new(adapter_name)),
+            Self::Iproute => Box::new(Iproute::new(adapter_name)),
         }
     }
 }
@@ -51,17 +76,42 @@ impl InterfaceConfigurationProvider {
 pub struct Manual;
 
 impl ConfigureInterface for Manual {
-    fn set_ip(&self, _adapter_name: &str, _ip_address: &IpAddr) -> Result<(), TeliodError> {
+    fn initialize(&mut self) -> Result<(), TeliodError> {
+        Ok(()) // No-op implementaiton
+    }
+
+    fn set_ip(&mut self, _ip_address: &IpAddr) -> Result<(), TeliodError> {
+        Ok(()) // No-op implementation
+    }
+
+    fn set_exit_routes(&mut self, _exit_node: &IpAddr) -> Result<(), TeliodError> {
+        Ok(()) // No-op implementation
+    }
+
+    fn cleanup_exit_routes(&mut self) -> Result<(), TeliodError> {
         Ok(()) // No-op implementation
     }
 }
 
 /// Implementation using `ifconfig`
-#[derive(Debug, PartialEq, Eq)]
-pub struct Ifconfig;
+#[derive(Debug)]
+pub struct Ifconfig {
+    adapter_name: String,
+}
+
+impl Ifconfig {
+    fn new(adapter_name: String) -> Self {
+        Self { adapter_name }
+    }
+}
 
 impl ConfigureInterface for Ifconfig {
-    fn set_ip(&self, adapter_name: &str, ip_address: &IpAddr) -> Result<(), TeliodError> {
+    fn initialize(&mut self) -> Result<(), TeliodError> {
+        execute(Command::new("ifconfig").args([&self.adapter_name, "mtu", "1420"]))?;
+        execute(Command::new("ifconfig").args([&self.adapter_name, "up"]))
+    }
+
+    fn set_ip(&mut self, ip_address: &IpAddr) -> Result<(), TeliodError> {
         let ip_string = ip_address.to_string();
         let cidr_suffix = if ip_address.is_ipv4() { "/10" } else { "/64" };
         let cidr_string = format!("{}{}", ip_string, cidr_suffix);
@@ -73,23 +123,24 @@ impl ConfigureInterface for Ifconfig {
 
         info!(
             "Assigning IP address for {} to {}",
-            adapter_name, cidr_string
+            self.adapter_name, cidr_string
         );
 
         match std::env::consts::OS {
             "macos" => {
                 execute(Command::new("ifconfig").args([
-                    adapter_name,
+                    &self.adapter_name,
                     ip_type,
                     &cidr_string,
                     &ip_string,
                 ]))?;
 
                 if ip_address.is_ipv4() {
-                    execute(Command::new("route").args(["add", "100.64/10", &ip_string]))?;
+                    execute(Command::new("route").args(["-n", "add", "100.64/10", &ip_string]))?;
                 } else {
                     execute(Command::new("route").args([
                         "add",
+                        "-n",
                         "-inet6",
                         "fd74:656c:696f::/64",
                         &ip_string,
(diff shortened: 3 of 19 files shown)


clis/teliod/src/config.rs
```diff
@@ -251,6 +259,12 @@ pub struct InterfaceConfig {
     pub config_provider: InterfaceConfigurationProvider,
 }
 
+#[derive(PartialEq, Eq, Deserialize, Serialize, Debug, Clone)]
+pub struct VpnConfig {
+    pub server_ip: SocketAddr,
```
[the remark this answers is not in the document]
mathiaspeters: Fixed and fixed!

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
+    fn set_exit_routes(&self, adapter_name: &str, exit_node: &IpAddr) -> Result<(), TeliodError> {
+        if exit_node.is_ipv4() {
+            match std::env::consts::OS {
+                "macos" => {
+                    execute(Command::new("ifconfig").args([adapter_name, "add", "10.5.0.1/32"]))?;
+                    execute(Command::new("route").args([
+                        "-n",
+                        "add",
+                        "-inet",
+                        "0.0.0.0/1",
```
tomasz-grz, line 202:
```
                        "-n",
                        "add",
                        "-inet",
                        "0.0.0.0/1",
                        "-interface",
                        adapter_name,
```

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
+                        adapter_name,
+                    ]))?;
+                    execute(Command::new("route").args([
+                        "-n",
+                        "add",
+                        "-inet",
+                        "128.0.0.0/1",
+                        adapter_name,
```
tomasz-grz, line 210: missing `"-interface",` flag for macOS

clis/teliod/src/daemon.rs
```diff
@@ -247,6 +290,21 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
...
     });
 
+    #[allow(mpsc_blocking_send)]
+    if let Some(VpnConfig {
+        server_endpoint,
+        server_pubkey,
+    }) = &config.vpn
+    {
+        _ = telio_tx
+            .send(TelioTaskCmd::ConnectToExitNode(
```
tomasz-grz, line 300: I wonder if this could potentially be racey 🚗 🏁 

From the logs look like we are sending `ConnectToExitNode` command (which calls `set_exit_routes()`) before sending the`SetIp` command which calls `set_ip()` and bring the interface up.

Also in the middle of API calls to `api.nordvpn.com` which might be breaking it.

```
hyper_util::client::legacy::pool: 671: checkout waiting for idle connection: ("https", api.nordvpn.com)
reqwest::connect: 625: starting new connection: https://api.nordvpn.com/    
hyper_util::client::legacy::connect::http: 481: Http::connect; scheme=Some("https"), host=Some("api.nordvpn.com"), port=None
hyper_util::client::legacy::connect::http: 769: connecting to 104.19.159.190:443
hyper_util::client::legacy::connect::http: 776: connect error for 104.19.159.190:443: ConnectError("tcp connect error", Os { code: 49, kind: AddrNotAvailable, message: "Can't assign requested address" })
hyper_util::client::legacy::connect::http: 769: connecting to 104.16.208.203:443
hyper_util::client::legacy::connect::http: 776: connect error for 104.16.208.203:443: ConnectError("tcp connect error", Os { code: 49, kind: AddrNotAvailable, message: "Can't assign requested address" })
hyper_util::client::legacy::pool: 713: checkout dropped for ("https", api.nordvpn.com)
```

As a matter of fact, I'm getting no-net, will investigate it a bit better tomorrow.

mathiaspeters: `set_ip` is no longer bringing the interface up, there is a separate `initialize` call responsible for that. But you might still be right that it's a bit racy

clis/teliod/src/configure_interface.rs
```diff
@@ -51,16 +75,71 @@ impl InterfaceConfigurationProvider {
...
+            default_gateway_ipv4: Self::get_default_gateway("inet"),
+            exit_route_ip: Cell::new(None),
+            ipv6_disabler: Ipv6Disabler::default(),
+        }
+    }
+
+    #[allow(index_access_check)]
+    fn get_default_gateway(family: &str) -> Option<IpAddr> {
+        let stdout =
+            execute_with_output(Command::new("netstat").args(["-nr", "-f", family])).ok()?;
```
tomaszklak, line 115: please use full paramater names (`--something`) instead of short versions (`-s`). This applies to all places in this review.

clis/teliod/src/configure_interface.rs
```diff
@@ -51,16 +75,71 @@ impl InterfaceConfigurationProvider {
...
+        for line in stdout.lines() {
+            let fields: Vec<&str> = line.split_whitespace().collect();
+            if fields.len() < 2 {
```
tomaszklak, line 118: What if there are more than two lines? This looks like it can be very fragile and easily **not** fail if run with a different version of `netstat` with slightly different, but incompatible, output.

clis/teliod/src/configure_interface.rs
```diff
@@ -51,16 +75,71 @@ impl InterfaceConfigurationProvider {
...
+                continue;
+            }
+
+            let destination = fields[0];
+            let gateway = fields[1];
```
tomaszklak, line 123: Use something like [collect_touple](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.collect_tuple) and remove the `#[allow(index_access_check)]`

clis/teliod/src/configure_interface.rs
```diff
@@ -51,16 +75,71 @@ impl InterfaceConfigurationProvider {
...
+    }
 }
 
 /// Implementation using `ifconfig`
-#[derive(Debug, PartialEq, Eq)]
-pub struct Ifconfig;
+#[derive(Debug)]
+pub struct Ifconfig {
+    default_gateway_ipv4: Option<IpAddr>,
+    exit_route_ip: Cell<Option<IpAddr>>,
```
tomaszklak, line 99: Could you remove the `Cell` and mark the mutating functions with the `mut` attribute for the `Ifconfig` reference?

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
     }
 }
 
 /// Implementation using `iproute2`
-#[derive(Debug, PartialEq, Eq)]
-pub struct Iproute;
+#[derive(Debug)]
+pub struct Iproute {
+    default_gateway_ipv4: Option<String>,
+    exit_route_ip: Cell<Option<IpAddr>>,
```
tomaszklak, line 303: Same here, remove the `Cell` and mark references with `mut`.

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
+            exit_route_ip: Cell::new(None),
+            ipv6_disabler: Ipv6Disabler::default(),
+        }
+    }
+
+    #[allow(clippy::expect_used)]
+    fn get_default_gateway(family: &str) -> Option<String> {
+        let metric_pattern = Regex::new("metric ([0-9]+)").expect("Failed to compile metric regex");
+        let via_pattern = Regex::new("via ([0-9\\.\\/]+)").expect("Failed to compile via regex");
+        let dev_pattern = Regex::new("dev ([A-Za-z0-9]+)").expect("Failed to compile dev regex");
```
tomaszklak, line 320: You can use capture groups to extract submatches, this will allow for removal o the `split_whitespace` and `nth` re-parsing and make it clear which part of string (just by looking at regex) is being used to produce a value.

clis/teliod/src/configure_interface.rs
```diff
@@ -1,12 +1,20 @@
...
 pub trait ConfigureInterface {
+    /// Initialize the interface
+    fn initialize(&self, adapter_name: &str) -> Result<(), TeliodError>;
     /// Configure the IP address and routes for a given interface
-    fn set_ip(&self, _adapter_name: &str, _ip_address: &IpAddr) -> Result<(), TeliodError>;
+    fn set_ip(&self, adapter_name: &str, ip_address: &IpAddr) -> Result<(), TeliodError>;
+    /// Configure routes for exit routing
+    fn set_exit_routes(&self, adapter_name: &str, exit_node: &IpAddr) -> Result<(), TeliodError>;
+    /// Some of the configured routes are not cleared when the adapter is removed and must be removed manually
+    fn cleanup_exit_routes(&self) -> Result<(), TeliodError>;
```
tomaszklak, line 17: I would expect all of those will use `&mut self` - aren't they mutating the `self`?

clis/teliod/src/daemon.rs
```diff
@@ -85,6 +91,7 @@ fn telio_task(
             &interface_config.name,
             adapter.to_owned(),
         )?;
+        _ = sys_config.initialize(&interface_config.name);
```
tomaszklak, line 94: Why do we ignore the result?

clis/teliod/src/daemon.rs
```diff
@@ -126,8 +133,44 @@ fn telio_task(
...
+                TelioTaskCmd::ConnectToExitNode(ip, pubkey) => {
+                    let node = ExitNode {
+                        identifier: uuid::Uuid::new_v4().to_string(),
+                        public_key: pubkey,
+                        allowed_ips: None,
+                        endpoint: Some(ip),
+                    };
+                    match telio.connect_exit_node(&node) {
+                        Ok(_) => {
+                            _ = sys_config.set_exit_routes(&interface_config.name, &ip.ip());
```
tomaszklak, line 145: Same, why do we ignore the result?

clis/teliod/src/configure_interface.rs
LukasPukenis: If the methods delegate the commands to backends such as `ifconfig` or `ip` then they do not need to modify `self`(if they do not store state). I wonder if an access to `self` is needed at all in this case and it can be `self`-less :)

clis/teliod/src/configure_interface.rs
```diff
@@ -51,16 +75,71 @@ impl InterfaceConfigurationProvider {
...
+
+            let destination = fields[0];
+            let gateway = fields[1];
+
+            if (destination == "default" || destination == "0.0.0.0")
+                && !gateway.starts_with("link#")
+            {
+                debug!("default gateway for {}: {}", family, gateway);
+                return IpAddr::from_str(gateway).ok();
+            }
```
LukasPukenis, line 130: minor: in such cases I am starting to like the idea to paste an examplary output in the comments. I think it allows for better understanding of what this code parses and when it fails, it's very easy to compare the outputs

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
+                            "add",
+                            exit_node.to_string().as_str(),
+                            "gw",
+                            gateway.to_string().as_str(),
+                        ]))?;
+                    }
+                    // We later disable IPv6 on all interfaces (except the tunnel interface)
+                    // but interfaces that get added later could still have IPv6 enabled.
+                    // As a backup solution we route all IPv6 packets into the tunnel
+                    execute(Command::new("route").args([
```
LukasPukenis, line 263: How does it work together with `Ipv6Disabler`? :thinking:

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
         }
+        Ok(())
+    }
 
-        execute(Command::new("ifconfig").args([adapter_name, "mtu", "1420"]))?;
-        execute(Command::new("ifconfig").args([adapter_name, "up"]))
+    fn set_exit_routes(&self, adapter_name: &str, exit_node: &IpAddr) -> Result<(), TeliodError> {
+        if exit_node.is_ipv4() {
+            match std::env::consts::OS {
+                "macos" => {
```
LukasPukenis, line 196: Discussion wanted: I'm a bit concerned that while we allow selecting a config backend (ifconfig, ip, manual), we still tweak the command syntax at runtime for `ifconfig` and for `route` based on OS just to give two examples.

I am entertaining an idea to offload this logic into the config file entirely:
```
"interface": {
    "name": "eth1",
    "config_provider": "ifconfig_linux",
    "route_backend": "route_linux"
},
```

and the code would just be trait implementations with no extra logic, straight forward commands and that's it. This might potentially counter situation where new MacOS release for example has different syntax for `ifconfig`. I am stretching a bit here though.

clis/teliod/src/daemon.rs
tomasz-grz: Right, also `ifconfig utun10 inet 10.5.0.2/32 10.5.0.2 alias` brings the interface already up.

tomasz-grz: I would be in favor of having a state machine, at-least in diagram form.
I'm not sure what should come first setting up the VPN, or setting up Meshnet, but I have a feeling each of those tasks should be completed in sequence. What do you think?

clis/teliod/src/configure_interface.rs
```diff
@@ -119,9 +355,160 @@ impl ConfigureInterface for Iproute {
...
         );
+        execute(Command::new("ip").args(["addr", "add", &cidr_string, "dev", adapter_name]))
+    }
 
-        execute(Command::new("ip").args(["addr", "add", &cidr_string, "dev", adapter_name]))?;
-        execute(Command::new("ip").args(["link", "set", "dev", adapter_name, "mtu", "1420"]))?;
-        execute(Command::new("ip").args(["link", "set", "dev", adapter_name, "up"]))
+    fn set_exit_routes(&self, adapter_name: &str, exit_node: &IpAddr) -> Result<(), TeliodError> {
+        if exit_node.is_ipv4() {
+            execute(Command::new("ip").args(["addr", "add", "10.5.0.1", "dev", adapter_name]))?;
```
tomasz-grz, line 363: Same should this be `10.5.0.2`?

clis/teliod/src/daemon.rs
```diff
@@ -126,8 +133,44 @@ fn telio_task(
...
+
+                            // Wait for VPN to actually be connected before handing control back to the event loop
```
tomasz-grz: I am little bit skeptical about this point...
I personally believe nothing should block `event loop`.

From user perspective it shouldn't look "stuck", the user should still be able to do `get-status` and `quit-daemon` no matter what state the daemeon is.

What would happen if the network momentarily goes down while we are in this loop, or we never receive `NodeState::Connected` due to whatever reason (like try to connect to meshnet peer that is offline)?

Would be nice to at least have a timeout here, what do you think?

clis/teliod/src/configure_interface.rs
LukasPukenis: ould the reason for choosing `0.0.0.0/1` instead of `0.0.0.0/0` be mentioned in a comment above? I googled for it and found that it might be used to take over precedence but I wonder about this exact situation why it was used.

https://serverfault.com/questions/1100250/what-is-the-difference-between-0-0-0-0-0-and-0-0-0-0-1

clis/teliod/src/configure_interface.rs
```diff
@@ -86,31 +165,188 @@ impl ConfigureInterface for Ifconfig {
...
+        Ok(())
+    }
 
-        execute(Command::new("ifconfig").args([adapter_name, "mtu", "1420"]))?;
-        execute(Command::new("ifconfig").args([adapter_name, "up"]))
+    fn set_exit_routes(&self, adapter_name: &str, exit_node: &IpAddr) -> Result<(), TeliodError> {
+        if exit_node.is_ipv4() {
+            match std::env::consts::OS {
+                "macos" => {
+                    execute(Command::new("ifconfig").args([adapter_name, "add", "10.5.0.1/32"]))?;
```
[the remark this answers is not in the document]
mathiaspeters: This was changed to `10.5.0.2` but macos support has not been fixed as it doesn't align with the OKR. We should fix it, but it will be a separate piece of work

clis/teliod/src/configure_interface.rs
mathiaspeters: The ifconfig additions were removed  so this wasn't fixed. As for Lukas' comment, on macos you're not allowed to set a route for `0.0.0.0/0` so instead you can set routes for `0.0.0.0/1` and `128.0.0.0/1` (thank you @tomasz-grz for teaching me that)

clis/teliod/src/configure_interface.rs
mathiaspeters: As mentioned, ifconfig additions were removed

clis/teliod/src/daemon.rs
mathiaspeters: I created a ticket for turning it into more of a state machine. This setup wasn't racey when running locally but it was racey when running as part of natlab so nwo the call to connect to VPN (and the associated routing) is happening only after the meshnet connection has been set up

clis/teliod/src/configure_interface.rs
mathiaspeters: Done, with the exception of some that don't have full names (but please correct me if I'm wrong about that)

clis/teliod/src/configure_interface.rs
mathiaspeters: The ifconfig additions were removed so this is no longer relevant. Only linux support is planned for now in our OKRs. We should eventually add macos support  but then with `networksetup` instead of `ifconfig` as that is what my limited research told was default on macos these days

clis/teliod/src/configure_interface.rs
mathiaspeters: See previous response

clis/teliod/src/configure_interface.rs
mathiaspeters: Yup, done

clis/teliod/src/configure_interface.rs
mathiaspeters: Done as well

clis/teliod/src/configure_interface.rs
mathiaspeters: Done

clis/teliod/src/configure_interface.rs
mathiaspeters: All methods take `&mut self` now

clis/teliod/src/daemon.rs
mathiaspeters: It was an oversight. Some results we don't really care about, and I was too quick to copy-paste

clis/teliod/src/daemon.rs
mathiaspeters: Same answer

clis/teliod/src/daemon.rs
```diff
@@ -126,8 +133,44 @@ fn telio_task(
...
+                                    .ok()
+                                    .and_then(|nodes| {
+                                        nodes
+                                            .iter()
+                                            .find(|n| n.public_key == node.public_key)
+                                            .map(|n| n.state == NodeState::Connected)
+                                    })
+                                    .unwrap_or(false);
+                                if is_connected {
+                                    debug!("Successfully connected to VPN");
```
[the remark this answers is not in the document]
mathiaspeters: Good point, changed

clis/teliod/src/configure_interface.rs
```diff
@@ -24,6 +32,22 @@ fn execute(command: &mut Command) -> Result<(), TeliodError> {
     }
 }
 
+/// Helper function to execute a system command with output
+fn execute_with_output(command: &mut Command) -> Result<String, TeliodError> {
+    let output = command
```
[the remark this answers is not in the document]
mathiaspeters: Added

clis/teliod/src/configure_interface.rs
```diff
@@ -51,16 +75,71 @@ impl InterfaceConfigurationProvider {
...
 }
 
 /// Implementation using `ifconfig`
-#[derive(Debug, PartialEq, Eq)]
-pub struct Ifconfig;
+#[derive(Debug)]
+pub struct Ifconfig {
+    default_gateway_ipv4: Option<IpAddr>,
+    exit_route_ip: Cell<Option<IpAddr>>,
+    ipv6_disabler: Ipv6Disabler,
```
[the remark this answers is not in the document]
mathiaspeters: I renamed it to `Ipv6SupportManager` and added some bigger comments to explain their purpose

clis/teliod/src/configure_interface.rs
mathiaspeters: The ifconfig additions were removed so no longer relevant

clis/teliod/src/configure_interface.rs
mathiaspeters: I feel like that was already in the comment just above the line :smile: but there are some more comments now on the `Ipv6SupportManager` that hopefully clear up their existence even more

clis/teliod/src/configure_interface.rs
mathiaspeters: Yup, changed

clis/teliod/src/daemon.rs
mathiaspeters: Good point. I made it into a sort of callback loop :+1:

mathiaspeters:
> Please add tests in nat-lab

There is one test added for now to just test if the VPN functionality is working. There will be a separate test for doing the network setup with the `iproute` config option but in a separate PR. There have been persistent issues getting that to work so it would be good to finish this to unblock other work and then have a follow-up PR for it. Does that sound good?

nat-lab/docker-compose.yml.bak
```diff
@@ -0,0 +1,823 @@
+version: '3.2'
```
LukasPukenis: This should be ignored by git due to :thinking: 
```
.gitignore:
/nat-lab/docker-compose.yml.bak
```

clis/teliod/src/configure_interface.rs
```diff
@@ -86,42 +136,304 @@ impl ConfigureInterface for Ifconfig {
...
+            exit_route_ip: None,
+            ipv6_support_manager: Ipv6SupportManager::default(),
+        }
+    }
+
+    #[allow(clippy::expect_used)]
+    fn get_default_gateway(family: &str) -> Option<String> {
+        let metric_pattern = Regex::new("metric ([0-9]+)").expect("Failed to compile metric regex");
+        let via_pattern = Regex::new("via ([0-9\\.\\/]+)").expect("Failed to compile via regex");
+        let dev_pattern = Regex::new("dev ([A-Za-z0-9]+)").expect("Failed to compile dev regex");
```
tomaszklak, line 196: tiny nitpick: those regexes can be made global using https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html

clis/teliod/src/daemon.rs
```diff
@@ -174,6 +268,7 @@ fn start_telio(
         adapter,
         ..Default::default()
     })?;
+    telio.set_fwmark(11673110)?;
```
tomaszklak: This is a third place where this value will be hardcoded:

```
$ rg 11673110 -t rust
crates/telio-sockets/src/protector/linux.rs
82:        protector.set_fwmark(11673110);

clis/tcli/src/cli.rs
32:const FWMARK_VALUE: u32 = 11673110;
```

Can you try to move it to one common place?

nat-lab/bin/core-api.py
```diff
@@ -83,6 +84,17 @@ def __init__(self, server_address, RequestHandlerClass, mqttc: mqtt.Client) -> N
...
+        """Override to ensure proper SSL shutdown"""
+        # Handle the request normally
+        super().finish_request(request, client_address)
+
+        # Ensure proper SSL shutdown
+        try:
+            request.unwrap()  # This sends close_notify
+        except:
+            pass  # Ignore errors during unwrap
+
```
tomaszklak: Could you add a comment explaining why this function is needed? It's not clear to me what kind of a problem it's preventing.

nat-lab/docker-compose.yml.bak
```diff
@@ -0,0 +1,823 @@
...
+
+services:
```
tomaszklak: This file shouldn't be committed.

nat-lab/tests/config.py
```diff
@@ -246,3 +246,11 @@ def get_root_path(path: str) -> str:
 MQTT_BROKER_HOST = "mqtt.nordvpn.com"
 MQTT_BROKER_IP = "10.0.80.85"
 CORE_API_CA_CERTIFICATE_PATH = "/etc/ssl/server_certificate/test.pem"
+CORE_API_CREDENTIALS = {
+    "username": "token",
+    "password": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
```
tomaszklak, line 253: I think it would be good idea to add a comment explaining that this is a test only password, not used in production. otherwise we might get bogus security reports about this from external 'researchers'.

clis/teliod/src/nc.rs
```diff
@@ -299,6 +299,10 @@ async fn request_nc_credentials(
         for cert in Certificate::from_pem_bundle(&cert)? {
             builder = builder.add_root_certificate(cert);
         }
+        // Due to a bug in reqwest, the rustls backend has to be specified
+        // Additionally, reqwest considers some self-signed certs, like the natlab one,
+        // as invalid and we therefore need to call the danger_accept_invalid_certs
+        builder = builder.use_rustls_tls().danger_accept_invalid_certs(true);
```
LukasPukenis, line 306: This looks a bit dangerous, could we compile it in only for Debug builds(since those are used for NatLab)?

clis/teliod/src/daemon.rs
```diff
@@ -120,11 +133,55 @@ fn telio_task(
...
-                        _ = sys_config.set_ip(&interface_config.name, &new_ip_address);
+                        _ = sys_config
+                            .set_ip(&interface_config.name, &new_ip_address)
+                            .inspect_err(|e| error!("Failed to set IP with error '{e:?}'"));
+                        task_connect_to_exit_node(vpn_config, tx_channel.clone());
+                    }
+                }
+                TelioTaskCmd::ConnectToExitNode(ip, pubkey) => {
+                    let node = ExitNode {
+                        identifier: uuid::Uuid::new_v4().to_string(),
```
LukasPukenis, line 149: Is this okay to be a random UUID? Looks a bit :fish: :thinking:

clis/teliod/src/daemon.rs
```diff
@@ -120,11 +133,55 @@ fn telio_task(
...
+                        .unwrap_or(false);
+                    if is_connected {
+                        info!("Successfully connected to VPN");
+                        _ = sys_config
+                            .set_exit_routes(&interface_config.name, &ip)
+                            .inspect_err(|e| {
+                                error!("Failed to set routes for exit routing with error '{e:?}'")
+                            });
+                    } else {
+                        task_maybe_setup_vpn_routes(tx_channel.clone(), ip, pubkey);
```
LukasPukenis, line 187: I have a bit of trouble understanding how `MaybeSetupVPNRoutes` and `ConnectToExitNode` interact. It looks like it is first `ConnectToExitNode` and then `MaybeSetupVPNRoutes`, however looking at the code, i see `task_maybe_setup_vpn_routes` invocation in both commands where in the `ConnectToExitNode` it is put under the guard while in the other command it is not. This makes me a bit confused if I missed something :thinking:

clis/teliod/src/daemon.rs
mathiaspeters: We do the same in production apps. If no ID is provided to `connect_to_exit_node` by the apps, we assign a random UUID

clis/teliod/src/nc.rs
mathiaspeters: Natlab builds are release builds, however, this will only be used if the user specifies a self-signed certificate. We could log a warning that it's unsafe and should be avoided

clis/teliod/src/configure_interface.rs
mathiaspeters: I'm not sure this is necessary or even useful at this point. It's only used in this one place that's called once per program execution. If we need to same functionality in the future I would be very much in favor of doing this, but for now, I don't see it as something worth doing

clis/teliod/src/daemon.rs
mathiaspeters: I added some comments to clarify this

nat-lab/tests/config.py
mathiaspeters: Good point. Added

nat-lab/docker-compose.yml.bak
mathiaspeters: I must have missed that one. Removed

clis/teliod/src/daemon.rs
mathiaspeters: I moved it to `telio-utils/lib.rs`. It won't cover the usage in natlab, but there's already a constant for that there

nat-lab/bin/core-api.py
mathiaspeters: Yup, added

nat-lab/docker-compose.yml.bak
mathiaspeters: Removed

clis/teliod/README.md
```diff
@@ -60,6 +60,9 @@ needs be absolute, otherwise will be relative to `working-directory` when daemon
     - `manual` - do not configure interfaces automatically
     - `ifconfig` - systems using ifconfig command
     - `iproute` - systems using iproute2 command
+  - `vpn`
```
tomasz-grz: Maybe worth noting that this field is optional, and if not provided, the VPN connection will not be established, what do you think?

clis/teliod/src/configure_interface.rs
```diff
@@ -1,16 +1,28 @@
...
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::net::IpAddr;
 use std::process::Command;
-use tracing::{error, info};
+use tracing::{debug, error, info};
 
 pub trait ConfigureInterface {
+    /// Initialize the interface
+    fn initialize(&mut self, adapter_name: &str) -> Result<(), TeliodError>;
```
tomasz-grz: I wonder, since we have `initialize()` and pass `adapter_name` every time.. Should we make the trait a bit more "object oriented", what do you think?

```suggestion
    fn initialize(adapter_name: &str) -> Result<Self, TeliodError>;
    fn set_ip(&mut self, ip_address: &IpAddr) -> Result<(), TeliodError>;
    fn set_exit_routes(&mut self, exit_node: &IpAddr) -> Result<(), TeliodError>;
```

clis/teliod/src/configure_interface.rs
```diff
@@ -86,42 +136,304 @@ impl ConfigureInterface for Ifconfig {
...
             }
             _ => {
-                execute(Command::new("ifconfig").args([adapter_name, ip_type, &cidr_string]))?;
+                execute(Command::new("ifconfig").args([
+                    adapter_name,
+                    "inet",
+                    "add",
+                    ip_address.to_string().as_str(),
+                    "netmask",
+                    "255.255.0.0",
```
tomasz-grz: If I'm not mistaken`100.64/10`should have a 10 bit mask `255.192.0.0`, not 16 bit

clis/teliod/src/nc.rs
```diff
@@ -10,11 +10,14 @@ use rumqttc::{
 };
 use serde::{Deserialize, Serialize};
 use std::{path::PathBuf, sync::Arc, time::Duration};
-use telio::telio_utils::{
-    exponential_backoff::{
-        Backoff, Error as BackoffError, ExponentialBackoff, ExponentialBackoffBounds,
+use telio::{
+    telio_lana::telio_log_warn,
```
tomasz-grz: nit: do we need to include `telio_lana` here? why not use `tracing::warn`?

clis/teliod/src/daemon.rs
```diff
@@ -189,16 +299,27 @@ async fn daemon_init(
     // This is to not look for tokens in a test environment right now as the values
     // are dummy and program will not run as it expects real tokens.
     let identity = Arc::new(if *config.authentication_token != EMPTY_TOKEN {
-        Box::pin(init_with_api(&config.authentication_token)).await?
+        Box::pin(init_with_api(
```
tomasz-grz: Nit: could we just pass it `&config`?

clis/teliod/src/daemon.rs
```diff
@@ -230,10 +353,12 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
         telio_task(
             identity_ptr,
             token_ptr,
+            cert_path,
             telio_rx,
             tx_clone,
             &config.interface,
             &config.adapter_type,
+            config.vpn,
```
tomasz-grz: Nit: should we just pass the whole config?

clis/teliod/src/configure_interface.rs
```diff
@@ -86,42 +136,304 @@ impl ConfigureInterface for Ifconfig {
...
+    fn set_exit_routes(
+        &mut self,
+        adapter_name: &str,
+        exit_node: &IpAddr,
+    ) -> Result<(), TeliodError> {
+        if exit_node.is_ipv4() {
+            Self::ignore_file_exists_error(execute(Command::new("ip").args([
+                "addr",
+                "add",
+                "10.5.0.2",
```
tomasz-grz: Since we are calling `set_exit_routes()` after `set_ip()` , why is this needed?

clis/teliod/README.md
mathiaspeters: Good idea. Added

clis/teliod/src/configure_interface.rs
mathiaspeters: This specific suggestion is not possible because all trait functions must take some form of self. But I added the adapter name to the `create` function of the provider so that any configurator that has state can get it through it's own constructor

clis/teliod/src/configure_interface.rs
mathiaspeters: I think you're right

clis/teliod/src/nc.rs
mathiaspeters: It was due to mindless copy-pasting

clis/teliod/src/daemon.rs
mathiaspeters: Yup, done

clis/teliod/src/daemon.rs
mathiaspeters: Done

clis/teliod/src/configure_interface.rs
mathiaspeters: It's not, it was just a leftover that wasn't removed. But it's removed now

clis/teliod/src/nc.rs
LukasPukenis: Hmmm... It would be nicer to isolate it as much as possible - logging will help a bit but will not isolate. If natlab would run debug builds, then it would be very easy to solve.

I am afraid of leaving this in the code since it enables any MITM crafted `nc` message to be passed and validated very easily. Maybe there's a function to only include this one NatLab certificate we have instead of enabling invalid certs?

mathiaspeters: You have to actually specify the custom cert location in the teliod config file for this code path to be taken so it's not just about the certificate being present. And the production core api would not accept any self-signed cert so I don't really see a security risk. Building natlab in debug mode would for sure be nice but it has been tried in the past and was reverted after some number of failures and then not revisited

LukasPukenis: Hmm... I now see that this is called only when the config has the path set, however now I wonder - what is the use of that feature - is it only for the NatLab?

This leaves us in a funny situation where we will definitely get a security report with a simple repro:
- Have user path set for the certificate in the config file(for example by accident we pushed one)
- Self-sign and MITM `nc` responses back to the user.

It does sound like a stretch since we control the configs, but it would be a bit safer to:
- include just this one certificate pointed explicitly into the certificate chain
- or introduce a feature flag for natlab builds and wrap it under one

:thinking:

mathiaspeters: We can do that, but I would prefer to do it in a separate PR

clis/teliod/src/configure_interface.rs
tomasz-grz: Nice! thanks!

clis/teliod/src/nc.rs
mathiaspeters: I created LLT-6428 for it

tomasz-grz (approved):
Nice work! +1

PR #1255, opened 2025-05-21
Fix token being printed as `****` in QNAP web ui

LukasPukenis:
At the moment, due to the previous introduction of Hidde<T> use around the token, it is literally printed as fource asterisk symbols, meaning trying to turn it on breaks as the token is sent as "****" literally.

### Problem

### Solution


tomasz-grz (approved):
+2

tomasz-grz (approved):

PR #1259, opened 2025-05-22
Fix enabling static runtime for windows

matislovas:
### Problem
The C static runtime for windows was actually turned off for libtelio.

### Solution
Turn it on

- [x] README.md is updated

### DIFF
--- a/ci/build_libtelio.py
+++ b/ci/build_libtelio.py
@@ -16,7 +16,12 @@ WORKING_DIR = f"{PROJECT_ROOT}"
 sys.path += [f"{PROJECT_ROOT}/3rd-party/rust_build_utils"]
 
 import rust_build_utils.rust_utils as rutils
-from rust_build_utils.rust_utils_config import GLOBAL_CONFIG
+import rust_build_utils.msvc as msvc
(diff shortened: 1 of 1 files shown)


ci/build_libtelio.py
```diff
@@ -523,7 +523,7 @@ def exec_build(args):
                 GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"] = ()
 
             GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"] += (
-                [" -C target-feature=-crt-static "],
```
tomaszklak, line 526: Can you add a check (maybe in nat-lab?) that would verify that the resulting library has the c runtime linked statically?

tomasz-grz (commented):
+1

tomaszklak:
I'm completely unfamiliar with windows - how do we know that the current state is incorrect and that we want the change it?

ci/build_libtelio.py
matislovas: Added!

tomaszklak (approved):
+1

PR #1261, opened 2025-05-26
Lukasp/fix auth token in teliod being hidden

LukasPukenis:
### Problem

### Solution


tomasz-grz (commented):
+1

PR #1265, opened 2025-05-26
Just web httpd accept any address

tomasz-grz:
### Problem
When developing on macOS, CGI server can only be run through a VM, but it only accepts connections from localhost. 

### Solution
CGI Server accepts connections from any address.

- [x] README.md is updated


Jauler (approved):
+1.0

PR #1270, opened 2025-05-28
Add a method to use event generation timestamps to decide if we receive them on time

tomaszklak:
This is in contrast to event reception time, which is used in other places. This is meant to work around the delays in event reception times cause by transient network problems caused by changing the network configuration.


jjanowsk (commented):
+1

PR #1275, opened 2025-06-02
NMACOS-8047 - Add workarounds for no-nets when AKS is enabled on macOS

matislovas:
### Problem
Apple s having no-nets after re-enabling meshnet, when AKS (Advanced Kill Switch) is enabled.

### Problem
Solution author: Marcin Twardak. Add a slight delay before socket rebind. Write empty packet after re-binding.

- [x] README.md is updated

### DIFF
--- a/crates/telio-sockets/src/protector/apple.rs
+++ b/crates/telio-sockets/src/protector/apple.rs
@@ -29,6 +29,7 @@ use tokio::{
     self,
     sync::{broadcast::Sender, Notify},
     task::JoinHandle,
+    time::Duration,
 };
 
 use crate::native::{interface_index_from_name, NativeSocket};
@@ -284,6 +285,13 @@ impl Sockets {
         if let Some(&interface) = self.default_interface.as_ref() {
             for socket in &self.sockets {
                 Sockets::bind(interface as u32, socket);
+
+                // NMACOS-8047 Writing an empty packet to the socket unblocks the sockets when
+                // `includeAllNetworks` flag is enabled.
+                #[cfg(target_os = "macos")]
+                unsafe {
+                    _ = libc::write(*socket, std::ptr::null(), 0);
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


crates/telio-sockets/src/protector/apple.rs
```diff
@@ -284,6 +293,13 @@ impl Sockets {
         if let Some(&interface) = self.default_interface.as_ref() {
             for socket in &self.sockets {
                 Sockets::bind(interface as u32, socket);
+
+                // NMACOS-8047 Writing an empty packet to the socket unblocks the sockets when
+                // `includeAllNetworks` flag is enabled.
+                #[cfg(target_os = "macos")]
```
stalowyjez, line 291: Why do you need this `target_os = "macos"`? Isn't whole this file for this OS?

twardakm:
@jjanowsk 

> If I understand correctly the whole point of this PR is to rebind again after 5 seconds

It's not exactly the purpose of this PR. We receive the network change notification in bursts. We might receive 1, 5 or even 10 notification that something has changed in the network. We do the rebindings after each of those notifications. Those notifications usually come in a very short period of time, e.g. 10 notifications might come within a second.

What I wanted to achieve with this code is to rebind **again 5 seconds after the last rebind happened**. From the proposed snippet, I understand the rebind will happen 5 seconds **after each rebind**, which significantly increases the number of rebindings we are performing.

twardakm:
Thank you for the explanation; you are right. I misunderstood the code, but it works as expected.

twardakm:
I simplified the implementation as per @jjanowsk suggestions.

crates/telio-sockets/src/protector/apple.rs
```diff
@@ -455,7 +481,12 @@ fn dynamic_store_callback(
         telio_log_info!("Force rebinding the sockets because of DynamicStore key change");
         telio_log_debug!("DynamicStore key changed: {:?}", _changed_keys);
 
-        sockets.lock().rebind_all();
+        if let Err(_e) = PROTECTOR_PATH_CHANGE_BROADCAST.send(()) {
+            telio_log_warn!("Failed to notify about Dynamic Store change");
+        } else {
+            telio_log_info!("Notfiy about DynamicStore key change");
+        }
+        telio_log_debug!("DynamicStore key changed: {:?}", _changed_keys);
```
stalowyjez: This line seems to be duplicated now

crates/telio-sockets/src/protector/apple.rs
```diff
@@ -455,7 +481,12 @@ fn dynamic_store_callback(
         telio_log_info!("Force rebinding the sockets because of DynamicStore key change");
         telio_log_debug!("DynamicStore key changed: {:?}", _changed_keys);
 
-        sockets.lock().rebind_all();
+        if let Err(_e) = PROTECTOR_PATH_CHANGE_BROADCAST.send(()) {
+            telio_log_warn!("Failed to notify about Dynamic Store change");
+        } else {
+            telio_log_info!("Notfiy about DynamicStore key change");
```
stalowyjez: Nit: A typo?

stalowyjez (commented):
Looks good to me, just a few nits, +1

crates/telio-sockets/src/protector/apple.rs
twardakm: no, since iOS 15.7.1, it is also used on iOS. Look at this fragment:

```rust
        #[cfg(target_os = "macos")]
        let should_create_socket_watcher = !is_test_env;

        #[cfg(any(target_os = "ios", target_os = "tvos"))]
        let should_create_socket_watcher = NativeProtector::should_create_socket_watcher();
```

```rust
const SOCKET_WATCHER_MIN_IOS: &str = "15.7.1";

    #[cfg(any(target_os = "ios", target_os = "tvos"))]
    fn should_create_socket_watcher() -> bool {
        matches!(
            compare_to(
                NativeProtector::get_ios_version().as_str(),
                SOCKET_WATCHER_MIN_IOS,
                Cmp::Gt,
            ),
            Ok(true)
        )
    }
```

The workaround is applied only for macOS, because we don't use (and don't plan to use in the near future) `includeAllNetworks` parameter.

crates/telio-sockets/src/protector/apple.rs
twardakm: Resolved in new commit

crates/telio-sockets/src/protector/apple.rs
twardakm: Resolved in new commit

jjanowsk (commented):
+1, just remove line 478 in apple.rs since it's unused now.

PR #1277, opened 2025-06-02
Run telio as prerequisite on windows vm to pre-load the driver

matislovas:
### Problem
We do receive sometimes an error (in production and on nightly tests):
```
WindowsNativeWg(Fail("Failed to create adapter: \\"Failed to create adapter, last os error: 2\\")\\")
```
We do suspect that this is due to race condition in `wireguard-nt` driver. And this is always seen on a first windows test in pipeline.

### Solution
Preload the driver as a pre-requisite in pipeline.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -6,7 +6,7 @@ import shutil
 import subprocess
 from config import DERP_PRIMARY, LAN_ADDR_MAP
 from contextlib import AsyncExitStack
-from helpers import SetupParameters
+from helpers import SetupParameters, setup_mesh_nodes
 from interderp_cli import InterDerpClient
 from itertools import combinations
 from typing import Dict, List, Tuple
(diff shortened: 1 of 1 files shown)


nat-lab/tests/conftest.py
```diff
@@ -300,6 +317,14 @@ async def _copy_vm_binaries_if_needed(items):
                 return
 
 
+async def _setup_windows_adapters_if_needed(items):
```
LukasPukenis, line 320: `if_needed` part isn't exercised, there's no conditional that's visible in this PR that would make that part, maybe rename it to just `_setup_windows_adapters`?

LukasPukenis (commented):
+1 nit: the description of PR could be the commit message

nat-lab/tests/conftest.py
matislovas: There can be pipelines, that does not have windows tests, and that function searches for `windows` marks in testing setup.

PR #1280, opened 2025-06-03
Added a new ffi function do serialize `Feature` to a json string

tomaszklak:
Also, property tests are added to make sure we can deserialize what we produce.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -1,9 +1,8 @@
 //! Object descriptions of various
 //! telio configurable features via API
 
-use std::{collections::HashSet, fmt, net::IpAddr};
+use std::{collections::HashSet, fmt, net::IpAddr, str::FromStr};
 
-use ipnet::Ipv4Net;
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use serde::{de::IntoDeserializer, Deserialize, Deserializer, Serialize};
 use smart_default::SmartDefault;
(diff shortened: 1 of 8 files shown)


crates/telio-model/src/features.rs
```diff
@@ -694,7 +765,9 @@ mod tests {
                     firewall: FeatureFirewall {
                         neptun_reset_conns: true,
                         boringtun_reset_conns: true,
-                        exclude_private_ip_range: None,
+                        exclude_private_ip_range: Some(Ipv4Net(
+                            "74.155.187.240/25".parse().unwrap()
```
matislovas, line 769: Maybe this can be put into some constant ..?

matislovas:
+1

crates/telio-model/src/features.rs
tomaszklak: If I put it in a constant and try to use it in the giant string above by wrapping it in `&format!(...)` it's not going to work since the json syntax in the string "collides" with rust string interpolation syntax. Which means I would have to concatenate a prefix of the huge string with the constant and the suffix of the string. This will greatly reduce the readability of this so I will keep it as is. Unless there is some other nicer way :)

matislovas: Yeah, makes sense :+1:

tomasz-grz (approved):
Nice! +1

PR #1281, opened 2025-06-03
more qnap/cgi tests

LukasPukenis:
Ideally we would have end-to-end tests for CGI and QNAP but that's more involved work at this point.

As an attempt to cover some middle ground, here I introduce unit tests for web endpoint handling and rendering. They validate certain paths in the CGI and QNAP handling but obviously miss some as well.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/cgi/api.rs
+++ b/clis/teliod/src/cgi/api.rs
@@ -56,7 +56,7 @@ pub(crate) fn handle_api(request: &CgiRequest) -> Option<Response> {
                     ))
                 }
             };
-            Some(update_config(body))
+            Some(update_config_file(body))
         }
         (&Method::GET, "/get-status") => Some(get_status()),
         (&Method::GET, "/get-teliod-logs") => Some(get_teliod_logs()),
@@ -129,7 +129,7 @@ pub(crate) fn start_daemon() -> (StatusCode, String) {
     // Logs must be checked manually for errors.
     // As a quick QoL improvement, we validate the config and auth tokens early
     // to provide immediate feedback instead of waiting for `teliod` to show up in the process list.
-    match get_config() {
+    match get_config_from_file() {
         Err(e) => {
             return (
                 StatusCode::INTERNAL_SERVER_ERROR,
@@ -192,13 +192,13 @@ pub(crate) fn stop_daemon() -> (StatusCode, String) {
     }
 }
 
-pub(crate) fn get_config() -> io::Result<TeliodDaemonConfig> {
+pub(crate) fn get_config_from_file() -> io::Result<TeliodDaemonConfig> {
     fs::read_to_string(TELIOD_CFG)
         .and_then(|content| serde_json::from_str(&content).map_err(|e| e.into()))
 }
 
-pub(crate) fn update_config(body: &str) -> Response {
-    let mut config: TeliodDaemonConfig = match get_config() {
+pub(crate) fn update_config_file(body: &str) -> Response {
+    let mut config: TeliodDaemonConfig = match get_config_from_file() {
         Ok(config) => config,
         Err(e) => {
             eprintln!("Error reading config file: {}", e);
@@ -298,23 +298,20 @@ fn get_meshnet_logs() -> Response {
 
 #[cfg(test)]
 mod tests {
-    use std::{fs, num::NonZeroU64, path::PathBuf};
+    use std::{num::NonZeroU64, path::PathBuf};
 
     use reqwest::StatusCode;
-    use serial_test::serial;
     use telio::device::AdapterType;
     use tracing::level_filters::LevelFilter;
 
-    use super::{update_config, TeliodDaemonConfig};
+    use super::{update_config_file, TeliodDaemonConfig, TeliodDaemonConfigPartial};
     use crate::{
-        cgi::constants::TELIOD_CFG,
         config::{InterfaceConfig, MqttConfig, Percentage},
         configure_interface::InterfaceConfigurationProvider,
     };
 
     #[test]
-    #[serial]
-    fn test_update_config() {
+    fn test_config_serialize_deserialize_and_update() {
         let mut expected_config = TeliodDaemonConfig {
             log_level: LevelFilter::DEBUG,
             log_file_path: "/path/to/log".to_owned(),
@@ -355,13 +352,17 @@ mod tests {
             }
         }
         "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
 
-        let read_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
-        assert_eq!(read_config, expected_config);
+        // Test deserialization
+        let mut config: TeliodDaemonConfig = serde_json::from_str(initial_config).unwrap();
+        assert_eq!(config, expected_config);
 
+        // Test serialization
+        let serialized = serde_json::to_string_pretty(&config).unwrap();
+        let deserialized: TeliodDaemonConfig = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(deserialized, config);
+
+        // Test update logic
         expected_config.log_level = LevelFilter::INFO;
         expected_config.log_file_path = "/new/path/to/log".to_owned();
         expected_config.log_file_count = 8;
@@ -399,89 +400,9 @@ mod tests {
             }
         }
         "#;
-        assert_eq!(update_config(update_body).status(), StatusCode::OK);
 
-        let updated_config: TeliodDaemonConfig =
-            serde_json::from_str(&fs::read_to_string(TELIOD_CFG).unwrap()).unwrap();
-        assert_eq!(updated_config, expected_config);
-    }
-
-    #[test]
-    #[serial]
-    fn test_update_partial_config() {
-        let mut expected_config = TeliodDaemonConfig {
-            log_level: LevelFilter::DEBUG,
-            log_file_path: "/path/to/log".to_owned(),
-            log_file_count: 7,
-            adapter_type: AdapterType::NepTUN,
-            interface: InterfaceConfig {
-                name: "eth0".to_owned(),
-                config_provider: InterfaceConfigurationProvider::Manual,
-            },
-            authentication_token:
-                "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
-                    .to_owned()
-                    .into(),
-            http_certificate_file_path: Some(PathBuf::from("/http/certificate/path/")),
-            mqtt: MqttConfig::default(),
-        };
-        let initial_config = r#"
-        {
-            "log_level": "debug",
-            "log_file_path": "/path/to/log",
-            "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
-            "adapter_type": "neptun",
-            "interface": {
-                "name": "eth0",
-                "config_provider": "manual"
-            },
-            "http_certificate_file_path": "/http/certificate/path/",
-            "mqtt": {
-                "backoff_initial": 1,
-                "backoff_maximal": 300,
-                "reconnect_after_expiry": 90,
-                "certificate_file_path": null
-            }
-        }
-        "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
-
-        let read_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
-        assert_eq!(read_config, expected_config);
-
-        expected_config.interface.name = "eth1".to_owned();
-        expected_config.interface.config_provider = InterfaceConfigurationProvider::Ifconfig;
-        let update_body = r#"
-        {
-            "interface": {
-                "name": "eth1",
-                "config_provider": "ifconfig"
-            }
-        }
-        "#;
-        assert_eq!(update_config(update_body).status(), StatusCode::OK);
-
-        let updated_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
-        assert_eq!(updated_config, expected_config);
-
-        expected_config.authentication_token =
-            "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
-                .to_owned()
-                .into();
-        let update_body = r#"
-        {
-            "authentication_token": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
-        }
-        "#;
-        assert_eq!(update_config(update_body).status(), StatusCode::OK);
-
-        let updated_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
-        assert_eq!(updated_config, expected_config);
+        let updates: TeliodDaemonConfigPartial = serde_json::from_str(update_body).unwrap();
+        config.update(updates);
+        assert_eq!(config, expected_config);
     }
 }
--- a/clis/teliod/src/cgi/mod.rs
+++ b/clis/teliod/src/cgi/mod.rs
@@ -1,13 +1,10 @@
-use std::{env::var, ops::Deref};
+use std::ops::Deref;
 
 use rust_cgi::{http::StatusCode, text_response, Request, Response};
 
 use crate::TIMEOUT_SEC;
 use telio::telio_utils::Hidden;
 use tracing::trace;
-
-#[cfg(feature = "qnap")]
-use qnap::QnapUserAuthorization;
 
 mod api;
 mod app;
@@ -42,10 +39,10 @@ pub struct CgiRequest {
 }
 
 impl CgiRequest {
-    fn new(inner: Request) -> Self {
+    fn new(inner: Request, route: &str) -> Self {
         Self {
             inner,
-            route: var("PATH_INFO").unwrap_or_default(),
+            route: route.to_string(),
         }
     }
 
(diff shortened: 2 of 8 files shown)


clis/teliod/src/cgi/mod.rs
```diff
@@ -76,44 +73,89 @@ enum AdminGroupStatus {
...
+pub fn handle_request_env(request: Request) -> Response {
+    static PATH_INFO_KEY: &str = "PATH_INFO";
+    let path = std::env::var(PATH_INFO_KEY).unwrap_or_default();
     #[cfg(feature = "qnap")]
-    if let Err(error) = authorize::<QnapUserAuthorization>(&request) {
-        return text_response(StatusCode::UNAUTHORIZED, format!("Unauthorized: {}", error));
+    {
+        handle_request::<qnap::QnapUserAuthorization>(request, &path)
+    }
+    #[cfg(all(not(feature = "qnap"), feature = "cgi"))]
```
tomasz-grz: Is `feature = "cgi"` actually needed? This whole module should not be compiled if `cgi` is disabled.

clis/teliod/src/cgi/api.rs
```diff
@@ -355,11 +357,12 @@ mod tests {
             }
         }
         "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
+        fs::write(TEST_CFG_PATH, initial_config).unwrap();
 
-        let read_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
+        let read_config = serde_json::from_str::<TeliodDaemonConfig>(
```
tomasz-grz: What do you think about having a helper function that simply returns a "test" `TeliodDaemonConfig`?

Also, do we really want need to test `fs::read_to_string` and `serde_json::from_str` repeatedly? Maybe we could mock it?

clis/teliod/src/cgi/api.rs
```diff
@@ -444,11 +451,12 @@ mod tests {
             }
         }
         "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
+        fs::write(TEST_CFG_PATH, initial_config).unwrap();
```
tomasz-grz: I'm introducing `temp-dir`dependency for creating temporary test files, would it be suitable to use here as well?
Could we then remove the `#[serial]` attribute?

clis/teliod/src/cgi/mod.rs
LukasPukenis: You raise a very good point :thinking: I will look into it <3

clis/teliod/src/cgi/api.rs
LukasPukenis: Good point, it's essentially just testing serde declarations here, I will try to refactor it to reduce the file writes

clis/teliod/src/cgi/api.rs
LukasPukenis: Actualy, initially I introduced another crate but in the end opted for a simpler approach - hardcoding a path which to me looked already like an improvement since existing one uses a constant which was fragile. However by addressing the previous comment about file writes and reads this might be eliminated entirely except for maybe one case, I will consider that

clis/teliod/src/cgi/api.rs
```diff
@@ -83,7 +83,7 @@ fn shutdown_teliod() -> Result<(), TeliodError> {
     Err(TeliodError::ClientTimeoutError)
 }
 
-pub(crate) fn start_daemon() -> (StatusCode, String) {
+pub(crate) fn start_daemon(path: &str) -> (StatusCode, String) {
```
lcruz99: the name of the argument "path", path for the config? Change it to a more descriptive name pls

clis/teliod/src/cgi/api.rs
```diff
@@ -192,13 +192,13 @@ pub(crate) fn stop_daemon() -> (StatusCode, String) {
     }
 }
 
-pub(crate) fn get_config() -> io::Result<TeliodDaemonConfig> {
-    fs::read_to_string(TELIOD_CFG)
+pub(crate) fn get_config(path: &str) -> io::Result<TeliodDaemonConfig> {
```
lcruz99: same as https://github.com/NordSecurity/libtelio/pull/1281/files#r2126727755

clis/teliod/src/cgi/api.rs
```diff
@@ -355,11 +357,12 @@ mod tests {
...
         }
         "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
+        fs::write(TEST_CFG_PATH, initial_config).unwrap();
 
-        let read_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
+        let read_config = serde_json::from_str::<TeliodDaemonConfig>(
+            &std::fs::read_to_string(TEST_CFG_PATH).unwrap(),
```
lcruz99: what's the reason behind this change?

clis/teliod/src/cgi/api.rs
```diff
@@ -399,10 +402,14 @@ mod tests {
...
-        assert_eq!(update_config(update_body).status(), StatusCode::OK);
+
+        assert_eq!(
+            update_config_file(update_body, TEST_CFG_PATH).status(),
+            StatusCode::OK
+        );
 
         let updated_config: TeliodDaemonConfig =
-            serde_json::from_str(&fs::read_to_string(TELIOD_CFG).unwrap()).unwrap();
+            serde_json::from_str(&std::fs::read_to_string(TEST_CFG_PATH).unwrap()).unwrap();
```
lcruz99: Same as https://github.com/NordSecurity/libtelio/pull/1281/files#r2126742952

clis/teliod/src/cgi/api.rs
```diff
@@ -444,11 +451,12 @@ mod tests {
...
         }
         "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
+        fs::write(TEST_CFG_PATH, initial_config).unwrap();
 
-        let read_config =
-            serde_json::from_str::<TeliodDaemonConfig>(&fs::read_to_string(TELIOD_CFG).unwrap())
-                .unwrap();
+        let read_config = serde_json::from_str::<TeliodDaemonConfig>(
+            &std::fs::read_to_string(TEST_CFG_PATH).unwrap(),
```
lcruz99: same as https://github.com/NordSecurity/libtelio/pull/1281/files#r2126742952

lcruz99: and in other places

clis/teliod/src/cgi/api.rs
lcruz99: Or I would change to `get_config_from_file(path: &str)`

clis/teliod/src/cgi/api.rs
lcruz99: > existing one uses a constant which was fragile

can you help me understand this?

clis/teliod/src/cgi/api.rs
```diff
@@ -355,11 +357,12 @@ mod tests {
             }
         }
         "#;
-        fs::write(TELIOD_CFG, initial_config).unwrap();
+        fs::write(TEST_CFG_PATH, initial_config).unwrap();
```
lcruz99: What's the reason behind this change? IIRC this TELIOD_CFG point to the cargo temporary files in the same directory as the test binary

clis/teliod/src/cgi/mod.rs
LukasPukenis: I eliminated cgi feature conditionals in `cgi` directory, now `libtelio/clis/teliod/src/cgi/ [lukasp/first-qnap-cgi-tests] rg cgi | rg feature` returns 0 results

clis/teliod/src/cgi/api.rs
LukasPukenis: Not a required change but in my eyes - a minor improvement. I didn’t like having to dig into **constants.rs** and trace conditional logic to figure out which config path is being used, especially in tests. For example having a testcase put under `qnap` feature would not work.

If the path was condition-less I wouldn't have done this. But in the end that's just my POV, and I hoped it will bring slightly more clarity to it.

clis/teliod/src/cgi/api.rs
LukasPukenis: Renamed to `config_path` :)

clis/teliod/src/cgi/api.rs
LukasPukenis: I followed the suggestion and renamed it, now it looks like this:
```
pub(crate) fn get_config_from_file(path: &str) -> io::Result<TeliodDaemonConfig>
```

clis/teliod/src/cgi/api.rs
LukasPukenis: Just an artifact of moving code around a lot, I restored the initial 1-liner here in all the other places since the testcases were partially a copy-paste job :)

clis/teliod/src/cgi/api.rs
LukasPukenis: @tomasz-grz I removed the `serial` dependency and it's usage and opted for `tempdir` as per your suggestion. This led to eliminate the `TEST_CFG_PATH` entirely as well

clis/teliod/src/cgi/mod.rs
```diff
@@ -76,44 +73,81 @@ enum AdminGroupStatus {
...
+            "0000000000000000000000000000000000000000000000000000000000000000".to_string(),
+        ))
+    }
+    async fn is_token_valid(_sid: &str) -> Result<Self, Error> {
+        Ok(ByPassValidator)
+    }
+    fn validate(&self) -> Result<(), Error> {
+        Ok(())
+    }
+}
```
lcruz99, line 98: I don't think there's a need to add `ByPassValidator`, what was your rationale?

clis/teliod/src/cgi/api.rs
LukasPukenis: I've reduced both testcases to just one and don't use any file I/O anymore there. We miss the test of `update_config_file()` but that is rather very simple function which essentially just uses serde to serialize into file

clis/teliod/src/cgi/api.rs
```diff
@@ -192,13 +192,13 @@ pub(crate) fn stop_daemon() -> (StatusCode, String) {
     }
 }
 
-pub(crate) fn get_config() -> io::Result<TeliodDaemonConfig> {
-    fs::read_to_string(TELIOD_CFG)
+pub(crate) fn get_config_from_file(path: &str) -> io::Result<TeliodDaemonConfig> {
+    fs::read_to_string(path)
```
lcruz99: I don't think this change is needed, `TELIOD_CFG` is conditionally set at compile time in constants.rs where you can add easily add an additional condition (eg: for the openwrt usecase or [for #cfg(tests) scenario](https://github.com/NordSecurity/libtelio/blob/fc765eae56ad0229d92f46863ea6c23c8c979f62/clis/teliod/src/cgi/api.rs))
This way you'll avoid most of the changes in this module (and checks during runtime). Besides that if teliod possibly runs with sudo permissions, this can create another attack vector.

lcruz99: IMHO this change and related ones should be mentioned in its own separate commit/PR

clis/teliod/src/cgi/api.rs
lcruz99: I kindly disagree and commented about it here: https://github.com/NordSecurity/libtelio/pull/1281#discussion_r2128630564

clis/teliod/src/cgi/mod.rs
```diff
@@ -76,44 +73,81 @@ enum AdminGroupStatus {
...
+
+// CGI entrypoint
+pub fn handle_request_env(request: Request) -> Response {
+    static PATH_INFO_KEY: &str = "PATH_INFO";
```
lcruz99, line 102: nice touch :)

clis/teliod/src/cgi/mod.rs
```diff
@@ -76,44 +73,81 @@ enum AdminGroupStatus {
...
+        return text_response(StatusCode::UNAUTHORIZED, format!("Unauthorized: {}", error));
+    }
     if let Some(response) = web::handle_web_ui(&request) {
         trace!(
             "Returning response..: {:?}",
             std::str::from_utf8(response.body()).ok()
         );
 
+        // Enabling tracing during tests obviously breaks it since the debug information messes up the output
+        #[cfg(all(debug_assertions, not(test)))]
```
lcruz99, line 131: How will a response look like with this addition? Will the web browser still be able to parse the html?

clis/teliod/src/cgi/mod.rs
```diff
@@ -155,3 +189,66 @@ pub fn trace_request(request: &CgiRequest, response: &Response) -> Option<Respon
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rust_cgi::http::Method;
+    use rust_cgi::Request;
+
+    #[tokio::test]
+    async fn test_handle_request_invalid_route() {
+        tokio::task::spawn_blocking(|| {
+            let resp = handle_request::<ByPassValidator>(
```
lcruz99, line 202: to have an increased test coverage, shouldn't handle_request_env() be called instead?

lcruz99: (also for other tests)

clis/teliod/src/cgi/mod.rs
```diff
@@ -76,44 +73,81 @@ enum AdminGroupStatus {
...
+    async fn is_token_valid(_sid: &str) -> Result<Self, Error> {
+        Ok(ByPassValidator)
+    }
+    fn validate(&self) -> Result<(), Error> {
+        Ok(())
+    }
+}
+
+// CGI entrypoint
+pub fn handle_request_env(request: Request) -> Response {
```
lcruz99, line 101: I'd say this function is not objective, it's not handling the request and it simply reads `PATH_INFO` variable which actually was already being done on CgiRequest::new().

clis/teliod/src/cgi/mod.rs
```diff
@@ -155,3 +189,66 @@ pub fn trace_request(request: &CgiRequest, response: &Response) -> Option<Respon
...
+
+    #[tokio::test]
+    async fn test_handle_request_api() {
+        tokio::task::spawn_blocking(|| {
+            let resp = handle_request::<ByPassValidator>(Request::default(), "/meshnet");
+            assert!(resp.status() == StatusCode::OK);
+        })
+        .await
+        .unwrap();
+    }
```
lcruz99, line 230: In this 3 tests you're testing the API endpoints which I think it should be done atmost in their respective modules web.rs and api.rs.

Instead it should be tested the handle_request behaviour for valid and invalid types of requests.

clis/teliod/src/cgi/mod.rs
```diff
@@ -155,3 +189,66 @@ pub fn trace_request(request: &CgiRequest, response: &Response) -> Option<Respon
...
+        }
+
+        tokio::task::spawn_blocking(|| {
+            let resp = handle_request::<DenyValidator>(Request::default(), "/meshnet");
+            assert_eq!(resp.status(), StatusCode::UNAUTHORIZED);
+        })
+        .await
+        .unwrap();
+    }
+}
```
lcruz99, line 254: BypassValidator is trivial which makes this test not relevant.

authorize() is tightly related with QNAP validator, we should mock QNAP's auth validator to test authorize() behaviours.

clis/teliod/src/cgi/api.rs
LukasPukenis: In the end I've reduced the tests to just test how serde serializes/deserializes the config, eliminating IO testing entirely

clis/teliod/src/cgi/api.rs
LukasPukenis: @lcruz99 @tomasz-grz I will respond here since there are a lot of comments about the same thing - I reverted these changes back. 

The reasoning was that for function it would be nice to be flexible and not depending on the global constant, however I didn't convert all of them and while converting the remaining ones it felt that it could be done as a separate effort if at all.

clis/teliod/src/cgi/api.rs
LukasPukenis: I've responded here - https://github.com/NordSecurity/libtelio/pull/1281#discussion_r2128819554 TLDR: reverted

clis/teliod/src/cgi/web.rs
```diff
@@ -416,3 +416,149 @@ fn meshnet(app: &AppState) -> Markup {
...
     }
 }
+
+mod tests {
+    use super::*;
+    use rust_cgi::http::Method;
+    use rust_cgi::Request;
+
+    // Minimal helper to create a CgiRequest for tests
+    #[allow(dead_code)]
```
lcruz99, line 426: is make_cgi_request() considered dead code?

clis/teliod/src/cgi/web.rs
```diff
@@ -416,3 +416,149 @@ fn meshnet(app: &AppState) -> Markup {
         }
     }
 }
+
```
lcruz99, line 419: #[cfg(test)] missing?

clis/teliod/src/cgi/web.rs
lcruz99: see https://github.com/NordSecurity/libtelio/pull/1281#discussion_r2128978089

clis/teliod/src/cgi/web.rs
```diff
@@ -416,3 +416,149 @@ fn meshnet(app: &AppState) -> Markup {
...
+    #[tokio::test]
+    async fn test_config_view_toggle_class() {
+        let mut app = AppState::collect();
+        app.running = false;
+        let html_off = config_view(&app, None).into_string();
+        assert!(!html_off.contains("peer-checked"));
+        app.running = true;
+        let html_on = config_view(&app, None).into_string();
+        assert!(html_on.contains("peer-checked"));
```
lcruz99, line 441: what does it mean to have "peer-checked"?

lcruz99: if it means that meshnet is running, create a helper function like is_meshnet_running() or similar

clis/teliod/src/cgi/web.rs
```diff
@@ -416,3 +416,149 @@ fn meshnet(app: &AppState) -> Markup {
         }
     }
 }
+
+mod tests {
```
lcruz99, line 420: I think it would make sense to have api endpoint tests in this mod.

clis/teliod/src/cgi/api.rs
```diff
@@ -298,23 +298,20 @@ fn get_meshnet_logs() -> Response {
...
     use crate::{
-        cgi::constants::TELIOD_CFG,
         config::{InterfaceConfig, MqttConfig, Percentage},
         configure_interface::InterfaceConfigurationProvider,
     };
 
     #[test]
-    #[serial]
-    fn test_update_config() {
+    fn test_config_serialize_deserialize_and_update() {
```
lcruz99, line 314: each test should be unitary and have only one single test subject, please separate them. ie: test_serialize, test_deserialize, test_update, etc.

PR #1282, opened 2025-06-04
Update rust dependecies

stalowyjez:
- [x] README.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ pretend_to_be_macos = ["telio-model/pretend_to_be_macos"]
 [dependencies]
 cfg-if = "1.0.0"
 ffi_helpers = "0.3.0"
+# We need this specific version for libmoose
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -15,6 +15,7 @@ pretend_to_be_macos = ["telio-model/pretend_to_be_macos"]
 [dependencies]
 cfg-if = "1.0.0"
 ffi_helpers = "0.3.0"
+zerofrom = "=0.1.5"
```
mathiaspeters, line 19: I'm guessing this is just here to pin the version of a transitive dependency, if so it would be good with a comment to clarify that

stalowyjez: @mathiaspeters  Hmm, like this?

gytsto (approved):
+1

mathiaspeters (approved):
+1

PR #1283, opened 2025-06-04
LLT-6080 Windows STUN interface bind

matislovas:
### Problem

### Solution


nat-lab/tests/test_telio_start_methods.py
```diff
@@ -0,0 +1,169 @@
...
+
+        tun12 = await alpha_client.create_tun(12)
+        await alpha_client.get_proxy().set_tun(tun12)
+        await alpha_client.restart_interface(new_name=tun_name_prefix + "12")
+        await alpha_client.get_router().delete_interface(tun_name_prefix + "11")
+        await ping_between_all_nodes(env)
+
+
+@pytest.mark.windows
+async def test_start_named_ext_if_filter() -> None:
```
gytsto, line 91: Can you also add test, where it would show how it actually takes the wrong interface when getting default interface and then it doesn't when started with `ext_if_filter`.

So basically:
- `start_normally` -> default iface returns some random `fake_interface`
- `start_named_ext_if_filter` -> default iface returns actual default interface ignoring `fake_interface`

src/device.rs
```diff
@@ -680,6 +681,20 @@ impl Device {
         })
     }
 
+    /// Configure the private key of the WireGuard interface
```
gytsto: Description probably wrong?

src/ffi.rs
```diff
@@ -410,12 +411,65 @@ impl Telio {
...
                     name: Some(name.clone()),
                     tun: None,
+                    ext_if_filter: None,
                 })
                 .log_result("Telio::start_named")
             })
         })
     }
 
+    /// Start telio with specified adapter, name
```
gytsto:
```suggestion
    /// Start telio with specified adapter type, adapter name
```

src/libtelio.udl
```diff
@@ -238,6 +238,12 @@ interface Telio {
     [Throws=TelioError]
     void start_named(SecretKey secret_key, TelioAdapterType adapter, string name);
 
+    /// Start telio with specified adapter, name and filtered NICs list.
```
gytsto:
```suggestion
    /// Start telio with specified adapter type, adapter name and filtered NICs list.
```

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -1080,6 +1080,8 @@ mod tests {
         fn set_fwmark(&self, _fwmark: u32) {}
 
         fn set_tunnel_interface(&self, _interface: u64) {}
+
+        fn set_ext_if_filter(&self, _list: &[String]) {}
```
gytsto, line 1084: Nit: What does `ext` here mean? Is it external or did you mean exclude? Also maybe to align with the code, function should refer to `nics_fitler` instead of `if_filter`

matislovas: It is external interface filter. I renamed all "NICs" to "interface/ifs", to avoid confusion.

src/device.rs
matislovas: Yeah, fixed

nat-lab/tests/test_telio_start_methods.py
gytsto: nvm, not needed

PR #1284, opened 2025-06-04
LLT-6312 Add teliod dockerfile

asledzinskiy:
Add base Dockerfile that contains teliod binary

### Problem
Need to have a base image with teliod

### Solution
Create Dockerfile with copied teliod daemon 

- [x] README.md is updated

### DIFF
--- a/clis/teliod/Dockerfile
+++ b/clis/teliod/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:24.04
+
+COPY target/release/teliod /usr/local/bin/teliod
+
+# Ensure the binary was copied
+RUN test -f /usr/local/bin/teliod && echo "Binary teliod exists" || (echo "ERROR: teliod binary is missing" && exit 1)
+
+RUN chmod +x /usr/local/bin/teliod


clis/teliod/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:24.04
+
+COPY target/debug/teliod /usr/local/bin/teliod
```
gytsto: should this be debug or release?

asledzinskiy: Release makes more sense, thanks. Will update

clis/teliod/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:24.04
+
+COPY target/release/teliod /usr/local/bin/teliod
+
+RUN chmod +x /usr/local/bin/teliod
+
+# Ensure the binary was copied
+RUN test -f /usr/local/bin/teliod && echo "Binary teliod exists"
```
Hasan6979: I would move this check before the `chmod` and exit with error message if binary wasn't copied.

asledzinskiy: Makes sense, changing

PR #1287, opened 2025-06-05
Remove code that causes test_start_with_tun to fail

tomaszklak:
The removes parts of #1273 which cause`test_start_with_tun` to fail occasionally with the bad file descriptor error - it's not a fix since the root cause is not yet known.

- [x] README.md is updated


lcruz99 (approved):
fair enough +2

PR #1290, opened 2025-06-05
Reduce log level of teliod client commands

tomasz-grz:
### Problem
Client commands like `get-status` and `is-alive` get constantly polled by the CGI, resulting in repeated log entries, potentially filling up space.

### Solution
Reduce the log level from `info` to `debug`

- [x] README.md is updated


LukasPukenis (approved):
+2

PR #1291, opened 2025-06-09
Lukasp/few cgi unit tests

LukasPukenis:
### Problem
CGI(web.rs rendering) doesn't have unit tests for rendering. End-to-end tests are not present at all.

### Solution
Added few unit tests in web.rs to check for rendered HTML markup based on inputs.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/cgi/api.rs
+++ b/clis/teliod/src/cgi/api.rs
@@ -577,6 +577,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_get_logs_valid_files() {
         let dir = TempDir::new().unwrap();
         let log_dir_path = dir.path().to_path_buf();
@@ -604,6 +605,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_get_logs_missing_files() {
         let dir = TempDir::new().unwrap();
         let log_dir_path = dir.path().to_path_buf();
@@ -621,6 +623,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_get_logs_no_files() {
         let dir = TempDir::new().unwrap();
         let log_dir_path = dir.path().to_path_buf();
@@ -648,6 +651,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_missing_init_log() {
         let dir = TempDir::new().unwrap();
         let log_dir_path = dir.path().to_path_buf();
@@ -664,6 +668,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_missing_log_dir() {
         let dir = TempDir::new().unwrap();
         let log_dir_path = dir.path().to_path_buf();
@@ -683,6 +688,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_missing_all_logs() {
         let fake_init_log = Path::new("/made/up/init");
         let fake_stdout_log = Path::new("/made/up/stdout");
--- a/clis/teliod/src/cgi/web.rs
+++ b/clis/teliod/src/cgi/web.rs
@@ -443,3 +443,183 @@ fn meshnet(app: &AppState) -> Markup {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rust_cgi::http::Method;
+    use rust_cgi::Request;
+    use serial_test::serial;
+
+    fn make_cgi_request(method: Method, route: &str) -> CgiRequest {
+        let mut req = Request::default();
+        *req.method_mut() = method;
+
+        CgiRequest::new(req, route.to_string())
+    }
+
+    #[tokio::test]
+    #[serial]
+    async fn test_config_view_toggle_class() {
+        let mut app = AppState::collect();
+        app.running = false;
+        let html_off = config_view(&app, None).into_string();
+        assert!(!html_off.contains("peer-checked"));
+        app.running = true;
+        let html_on = config_view(&app, None).into_string();
+        assert!(html_on.contains("peer-checked"));
+    }
+
+    #[tokio::test]
+    #[serial]
+    async fn test_meshnet_render_all_node_states() {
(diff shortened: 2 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -47,6 +47,9 @@ jobs:
       - run: cargo test --package=teliod --features=cgi -- --nocapture
         env:
           RUST_BACKTRACE: full
+      - run: cargo test --package=teliod --features=qnap -- --nocapture
```
lcruz99: any reason for splitting qnap and cgi tests?

clis/teliod/src/cgi/web.rs
```diff
@@ -421,3 +421,150 @@ fn meshnet(app: &AppState) -> Markup {
         }
     }
 }
+
```
lcruz99, line 446: `#[cfg(test)]` missing?

clis/teliod/src/cgi/web.rs
```diff
@@ -421,3 +421,150 @@ fn meshnet(app: &AppState) -> Markup {
...
+mod tests {
+    use super::*;
+    use rust_cgi::http::Method;
+    use rust_cgi::Request;
+
+    #[allow(dead_code)]
```
lcruz99: Isn't this necessary due to the missing #[cfg(test)] on line 424?

clis/teliod/src/cgi/web.rs
```diff
@@ -421,3 +421,150 @@ fn meshnet(app: &AppState) -> Markup {
...
+    fn make_cgi_request(method: Method, route: &str) -> CgiRequest {
+        let mut req = Request::default();
+        *req.method_mut() = method;
+
+        std::env::set_var("PATH_INFO", route);
```
lcruz99: Tests are running concurrently, this var writing might cause flakyness.

clis/teliod/src/cgi/web.rs
```diff
@@ -421,3 +421,150 @@ fn meshnet(app: &AppState) -> Markup {
...
+        CgiRequest::new(req)
+    }
+
+    #[tokio::test]
+    async fn test_config_view_toggle_class() {
+        let mut app = AppState::collect();
```
lcruz99, line 464: As in the previous comment, this call may return some shared state due to tests running in parallel. Maybe it makes sense to run this tests sequentially instead? 🤔

clis/teliod/src/cgi/web.rs
```diff
@@ -421,3 +421,150 @@ fn meshnet(app: &AppState) -> Markup {
...
+        // Check for the log level select
+        assert!(body.contains("Log Level"), "Missing log level select");
+        // Check for the docs link
+        assert!(
+            body.contains("https://meshnet.nordvpn.com"),
+            "Missing docs link"
+        );
+        assert_eq!(resp.status(), 200, "Status mismatch for index");
+    }
+}
```
lcruz99, line 625: Can you add tests for invalid/unsupported scenarios? (routes/http methods)

clis/teliod/src/cgi/api.rs
```diff
@@ -385,7 +385,7 @@ fn get_logs(
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, not(feature = "qnap")))]
```
tomasz-grz: Why are these disabled for `qnap`? Is it because of the hard-coded app path?

Is there a better way to work around it instead of just disabling the tests? Or at least have an equivalent `#[cfg(all(test, feature = "qnap"))]`

clis/teliod/src/cgi/web.rs
tomasz-grz: Very good point

.github/workflows/tests.yml
LukasPukenis: It felt correct to have each feature tested since it exposes various paths in the code, however a this point cgi and qnap basically mean the same thing and testing `qnap` is more difficult due to it's qnap dependencies.

So long story short - I removed this change

clis/teliod/src/cgi/web.rs
LukasPukenis: You're right! thanks!

clis/teliod/src/cgi/web.rs
LukasPukenis: yep, added the missing `cfg`

clis/teliod/src/cgi/web.rs
LukasPukenis: Fixed this one by making CgiRequest explicitly accept the `route` argument.

clis/teliod/src/cgi/api.rs
LukasPukenis: Removed this one, was unecessary, thanks for shedding light on it.

clis/teliod/src/cgi/web.rs
LukasPukenis: Added an extra unit test for invalid paths with `GET` and `POST`. Not adding any valid `POST` tests since those would trigger `teliod` to be executed.

It also highlights a bit the limitation of the current design - the same endpoint is used for both config updates and starting/stopping daemon. Would they be split, we could test the endpoint code better.

It can be done with a future refactor out of scope for this PR.

clis/teliod/src/cgi/web.rs
LukasPukenis: That is a good point, though I have a slight wonder why I didn't ever trip on it while running. Guess just lucky. I checked `api.rs` and saw that it has more unit tests that are doing I/O on the same files so I marked them as `serial` in addition to marking unit tests that depend on `AppState` in `web.rs`.

To be perfectly honest I do not like such `serial` sprinkling, it's easy to forget and a nicer approach would be to just generate temporary paths instead but that would be much bigger effort to convert code and testcases so for now I think it does the job :).

lcruz99:
+1

LukasPukenis:
Closing since we do not do QNAP anymore :) :(

PR #1292, opened 2025-06-09
Ignore and print exception on windows setup adapters

gytsto:
Initially it was meant to just setup adapters to
extract wintun adapters, because it was causing issues during tests. But if the same issue occurs, exception should have been ignored. This commit fixes that.

Also meshnet of windows VMs are not needed. It's
enough to just start libtelio client.

### Problem
Exception is not ignored on preparation.

### Solution
Ignore exception.


matislovas:
Thanks! +2

matislovas (approved):

PR #1293, opened 2025-06-09
In QNAP open Meshnet in the same Window

LukasPukenis:
### Problem
NordSecurityMeshnet is opened in another browser tab instead of the built-in QNAP desktop environment.

### Solution
QNAP has ability to open applications in an iframe which can be easily done by tweaking the configuration parameter.
fla
image

It eventually lands in /etc/config/qpkg.conf
with `Desktop = 1` written to it.

- [x] README.md is updated

### DIFF
--- a/qnap/qpkg.cfg
+++ b/qnap/qpkg.cfg
@@ -53,11 +53,11 @@ QPKG_USE_PROXY="1"
 
 #Desktop Application (since 4.1)
 # Set value to 1 means to open the QPKG's Web UI inside QTS desktop instead of new window.
-#QPKG_DESKTOP_APP="1"
+QPKG_DESKTOP_APP="1"
 # Desktop Application Window default inner width (since 4.1) (not over 1178)
 #QPKG_DESKTOP_APP_WIN_WIDTH=""
 # Desktop Application Window default inner height (since 4.1) (not over 600)
-#QPKG_DESKTOP_APP_WIN_HEIGHT=""
+QPKG_DESKTOP_APP_WIN_HEIGHT="600"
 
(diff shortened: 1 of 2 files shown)


tomasz-grz (commented):
+1

qnap/qpkg.cfg
```diff
@@ -53,11 +53,11 @@ QPKG_USE_PROXY="1"
 
 #Desktop Application (since 4.1)
 # Set value to 1 means to open the QPKG's Web UI inside QTS desktop instead of new window.
-#QPKG_DESKTOP_APP="1"
+QPKG_DESKTOP_APP="1"
 # Desktop Application Window default inner width (since 4.1) (not over 1178)
 #QPKG_DESKTOP_APP_WIN_WIDTH=""
 # Desktop Application Window default inner height (since 4.1) (not over 600)
-#QPKG_DESKTOP_APP_WIN_HEIGHT=""
+QPKG_DESKTOP_APP_WIN_HEIGHT="600"
```
lcruz99, line 60: Any reason for defining height?

LukasPukenis: The documentation says that 600 is the maximum value and meshnet is a "tall" application when there are peers and it's enabled which was the reason I chose the value. Having too small height and enabling meshnet might potentially not reveal any peers visually just the scrollbar on the right side which to me looks suboptimal.

In addition to that, please take a look at the screenshots which gave me a visual validation.

lcruz99: was wondering why were we defining height and in that case why not also width.. anw, it's clear to me :)

PR #1294, opened 2025-06-09
Persist teliod daemon state using a marker file

LukasPukenis:
### Problem
Meshnet is being started at the start of QNAP regardless if user has started it previously or not. This means that if user enters a valid token, starts meshnet and then stops it and machine restart happens - meshnet will automatically turn on even though it should stay off.

### Solution
In CGI handler we save user's intent by creating a file and then check for it in
NordSecurityMeshnet.sh when starting.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/cgi/constants.rs
+++ b/clis/teliod/src/cgi/constants.rs
@@ -21,3 +21,6 @@ pub const TELIOD_INIT_LOG: &str = "/var/log/teliod_init.log";
 
 #[cfg(debug_assertions)]
 pub const CGI_LOG: &str = concatcp!(APP_DIR, "/cgi.log");
+
+/// Path to the user intent file used to automatically turn meshnet on if found by NordSecurityMeshnet.sh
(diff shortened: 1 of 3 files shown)


lcruz99:
The user manages the applications state through QDK, he can enable or disable it on the web interface which makes start/stop on the teliod app redundant.

Does the user have any usecase like checking logs or configuration when the app is enabled but not meshnet? If not I think we could stick to the default way of managing the apps through QDK.

LukasPukenis:
> The user manages the applications state through QDK, he can enable or disable it on the web interface which makes start/stop on the teliod app redundant.
Does the user have any usecase like checking logs or configuration when the app is enabled but not meshnet? If not I think we could stick to the default way of managing the apps through QDK.

@lcruz99 Do you mean to remove the toggle from the CGI part altogether and leave the control solely through the "App Center" actions - _Start_ and _Stop_ on _NordSecurityMeshnet_?

PR #1300, opened 2025-06-10
Replace raw fd with ownedfd for tunnel descriptor

tomaszklak:
### Problem
Before (#1273) a bug was introduced which caused nat-lab test start_with_tun to sometimes fail. This was later reverted (in #1287). The source of the problem was that the `OwnedFd` which was part of the `Config` was dropped after the underlying descriptor was passed to NepTUN - which meant the descriptor was closed.

### Solution
Instead of passing config by reference, and dropping it (and contained `tun`) afterwards, this PR changes the `start` (and `set_tun`) so that it consumes the `OwnedFd` using `IntoRawFd` and pass it to NepTUN for management.

- [x] README.md is updated

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -195,13 +195,13 @@ pub type Result<T = ()> = std::result::Result<T, Error>;
 pub trait EventCb: Fn(Box<Event>) + Send + 'static {}
 impl<T> EventCb for T where T: Fn(Box<Event>) + Send + 'static {}
 
-#[derive(Clone, Default)]
+#[derive(Default)]
 pub struct DeviceConfig {
     pub private_key: SecretKey,
     pub adapter: AdapterType,
     pub fwmark: Option<u32>,
     pub name: Option<String>,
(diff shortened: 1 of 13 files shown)


src/device.rs
```diff
@@ -1189,15 +1210,15 @@ impl Runtime {
...
+            device_config: requested_device_config,
             keepalive_periods: features.wireguard.persistent_keepalive.clone(),
             ..Default::default()
         };
 
         let dns = Arc::new(Mutex::new(DNS {
             resolver: None,
             #[cfg(unix)]
-            virtual_host_tun_fd: config.tun,
+            virtual_host_tun_fd, // config.tun.clone(),
```
mathiaspeters: Looks like this comment was supposed to be removed

gytsto (commented):
+1

src/device.rs
tomaszklak: Good catch! Thanks! :)

mathiaspeters (approved):
+1

PR #1301, opened 2025-06-10
Fix teliod log path validation

tomasz-grz:
### Problem
When a relative path is given as `log_file_path`, the validation tries to convert it to an absolute path.

Since most likely the file does not yet exist, `canonicalize()` will fail with error `Error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })`

### Solution
Properly convert the path to absolute, by splitting file name and directory parts, and canonicalize only the parent directory.
Add more descriptive error message.

- [x] README.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5204,6 +5204,7 @@ dependencies = [
  "smart-default",
  "telio",
  "temp-dir",
+ "temp-file",
  "thiserror 2.0.12",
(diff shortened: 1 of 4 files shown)


LukasPukenis (commented):
Could we have:
- The PR message in the commit message
- Unit test(s) exercising the new code

PR #1302, opened 2025-06-11
Make QNAP package display name consistent with UI

tomasz-grz:
### Problem
On QNAP the application is displayed as `NordSecurityMeshnet` but once open the UI shows the title `Nord Security Meshnet`.

### Solution
Make `QPKG_DISPLAY_NAME` consistent with UI.


olekoliinyk:
+2

PR #1304, opened 2025-06-11
LLT-6298: Compute QNAP package install path at runtime

lcruz99:
### Problem
Path was hardcoded and it may not be valid for every installation.

### Solution
Compute the path during runtime.


clis/teliod/src/cgi/constants.rs
```diff
@@ -1,26 +1,158 @@
...
+        self.join(".teliod_start_intent")
+    }
+
+    /// Create a path relative to the root directory
+    #[cfg_attr(test, visibility::make(pub))]
+    fn join(&self, path: &str) -> PathBuf {
+        self.root
+            .as_ref()
+            .map(|root| root.join(path))
+            .unwrap_or_else(|| PathBuf::from(path))
```
tomaszklak: What about the case when `getcfg` fails? Is it fine to return `path` in that case?

clis/teliod/src/cgi/constants.rs
```diff
@@ -1,26 +1,158 @@
...
+}
+
+impl LogPaths {
+    /// Creates a new LogPaths instance
+    fn new() -> Self {
+        Self {
+            stdout: PathBuf::from("/var/log/teliod.log"),
+            init_stdout: PathBuf::from("/var/log/teliod_init.log"),
+            lib_dir: PathBuf::from("/var/log"),
+            lib_prefix: String::from("teliod_lib.log"),
```
tomaszklak: nit: To be honest, I would expect that the names are `log_dir` and `log_prefix`, or even just `dir` and `prefix`.

clis/teliod/src/cgi/constants.rs
```diff
@@ -1,26 +1,158 @@
...
+                .output()
+                .ok()
+                .filter(|output| output.status.success())
+                .and_then(|output| String::from_utf8(output.stdout).ok())
+                .map(|stdout| stdout.trim().to_string())
+                .filter(|s| !s.is_empty())
+                .map(PathBuf::from)
+        } else {
+            Some(
+                Path::new(env!("CARGO_MANIFEST_DIR"))
```
tomasz-grz: Doesn't this hard code the value to the build path? and break after moving the binary somewhere else outside of `../target/debug`?

What do you think of using something like `/var/run` where the `teliod.sock` lives or `HOME/.local/share` where the `data.json` lives?

clis/teliod/src/cgi/constants.rs
```diff
@@ -1,26 +1,158 @@
...
+/// System-wide log paths configuration
+pub static LOG_PATHS: LazyLock<LogPaths> = LazyLock::new(LogPaths::new);
+
+/// Handles all teliod-relative paths
+#[derive(Debug, Clone)]
+pub struct AppPaths {
+    root: Option<PathBuf>,
+}
+
+impl AppPaths {
```
tomasz-grz, line 49: Nice refactoring! 💯

clis/teliod/src/cgi/api.rs
```diff
@@ -241,9 +241,15 @@ pub(crate) fn update_config(body: &str) -> Response {
     };
 
     config.update(updated_config);
+    if let Err(e) = config.validate() {
+        return text_response(
+            StatusCode::BAD_REQUEST,
+            format!("Invalid config value: {e}"),
+        );
+    };
```
LukasPukenis: It looks like this could be improved by not allowing the `config` to reach an invalid state at all. That way the `config` instance would be either valid or not exist.

Could we inline the `validate()` function call into `update()` and make `fn update(&mut self) -> Result<_>`?

clis/teliod/src/cgi/constants.rs
```diff
@@ -1,26 +1,158 @@
...
+                    "-f",
+                    "/etc/config/qpkg.conf",
+                ])
+                .output()
+                .ok()
+                .filter(|output| output.status.success())
+                .and_then(|output| String::from_utf8(output.stdout).ok())
+                .map(|stdout| stdout.trim().to_string())
+                .filter(|s| !s.is_empty())
+                .map(PathBuf::from)
```
LukasPukenis: This looks neat, however very hard to validate if it will work correctly by just looking at the code. Would it be possible if it's not too much output - paste the output(redacted if needed) generated by the command in the comment section above?

clis/teliod/src/config.rs
```diff
@@ -156,41 +156,71 @@ impl TeliodDaemonConfig {
...
 
-            // correct if path is not in ./file_name.log format but file_name.log
-            let dir = match log_path.parent() {
+        Ok(config)
+    }
+
+    pub fn validate(&mut self) -> Result<(), TeliodError> {
+        self.log_file_path = self.resolve_log_path()?;
+        if let Ok(env_token) = self.resolve_env_token() {
+            self.authentication_token = env_token;
```
LukasPukenis: I would expect `validate` to have a signature of `validate(&self)` - not modify anything just tell me if it's valid or not :). Could we move the token override to the call site or into a separate function?

clis/teliod/src/config.rs
```diff
@@ -156,41 +156,71 @@ impl TeliodDaemonConfig {
...
+    fn resolve_log_path(&self) -> Result<String, TeliodError> {
+        let path = PathBuf::from(&self.log_file_path);
+
+        let file_name = path.file_name().ok_or_else(|| {
+            TeliodError::InvalidConfigOption(
+                "log_file_path".to_owned(),
+                "must specify a file name".to_owned(),
+                path.to_string_lossy().into_owned(),
+            )
+        })?;
```
LukasPukenis: This is almost out of scope of this PR but it touches the code so I feel that it's a good place to potentially improve on this.

I took a look into the error and it has 3 fields all being `String`:
```
#[error("Invalid config option {0}: {1} (value '{2}')")]
InvalidConfigOption(String, String, String),
```

Because all member types are the same, it is confusing, and also error prone since it's easy to mix them up. My recommendations to improve on this:
- Merge first two fields into one
- Improve the error struct to look like
```
#[derive(Debug, thiserror::Error)]
pub enum TeliodError {
    ...
    #[error("Invalid config option {key}: {msg} (value '{value}')")]
    InvalidConfigOption {
        key: String,
        msg: String,
        value: String,
    },
}
```

This way the call sites need to explicitly mention the argument name which makes it less error prone and more readabl :)

clis/teliod/src/cgi/api.rs
```diff
@@ -120,13 +117,16 @@ pub(crate) fn start_daemon() -> (StatusCode, String) {
         .write(true)
         .read(true)
         .truncate(true)
-        .open(TELIOD_INIT_LOG)
+        .open(LOG_PATHS.init_stdout())
```
LukasPukenis: It reads as if it initializes something about `stdout` on `LOG_PATHS` instance. The previous name was very explicit about it's purpose. It is also used not only for the `stdout` but for `stderr` as well when daemonizing.

Could we rename it to something along the lines of `LOG_PATHS.daemon_init_log_file()`?

lcruz99: agreed, changed

clis/teliod/src/cgi/constants.rs
lcruz99: agreed, changed

clis/teliod/src/cgi/api.rs
lcruz99: agreed, will do

clis/teliod/src/config.rs
lcruz99: agreed, will do

clis/teliod/src/cgi/constants.rs
lcruz99: If I understood your request correctly:
```
[qnap@NAS-INV1183 ~]$ getcfg NordSecurityMeshnet Install_Path -f /etc/config/qpkg.conf
/share/CACHEDEV1_DATA/.qpkg/NordSecurityMeshnet
```

lcruz99: Those chain calls basically extract the success values and trim any leading/trailing white space from the output.

clis/teliod/src/cgi/constants.rs
lcruz99: > Doesn't this hard code the value to the build path? 

Yes, something like `./target/x86_64-unknown-linux-gnu/debug/` where teliod exists.

> What do you think of using something like /var/run where the teliod.sock lives or HOME/.local/share where the data.json lives?

I'm afraid I didn't understand, the application is not prepared to move and provide a custom path for teliod.

clis/teliod/src/cgi/constants.rs
lcruz99: Yes, it will return `path` (eg: "teliod") and the caller will not find the file and error out.

what else could be done? 🤔 
    1. Stop the application at the first time the `getcfg` command fails (it is critical)
    2. make it return a Result<> to make the caller aware of it (and then stop the application?)
    3. return PathBuf::from("./{}", path)

wdyt?

clis/teliod/src/cgi/constants.rs
tomasz-grz: But `target/x86_64-unknown-linux-gnu/debug/` exists only on the system that the package was build, no?

If I do something like
```sh
cargo install --path clis/teliod
cargo clean
```
or simply `scp` `teliod` to my OpenWRT, or download the build from the CI pipeline, what happens then?

clis/teliod/src/config.rs
lcruz99: Nice, changed :)

clis/teliod/src/cgi/constants.rs
tomaszklak: Hm, the field names are still using the `lib_` prefix.

clis/teliod/src/cgi/constants.rs
tomaszklak: I think it's better to make the root not optional - it's better to fail as early as possible, and as loudly as possible.

clis/teliod/src/cgi/constants.rs
```diff
@@ -1,26 +1,155 @@
...
+                .args([
+                    "NordSecurityMeshnet",
+                    "Install_Path",
+                    "-f",
+                    "/etc/config/qpkg.conf",
+                ])
+                .output()
+                .ok()
+                .filter(|output| output.status.success())
+                .and_then(|output| String::from_utf8(output.stdout).ok())
```
tomaszklak: nit: do we need utf8? I think all we need here is `OsString`.

clis/teliod/src/cgi/constants.rs
LukasPukenis: Yep, that's basically what I ask - seeing this output I can at least tell if the code is somewhat right. Doing a unit test would make it easier as well but not necessarily needed. Thanks :)

clis/teliod/src/cgi/constants.rs
lcruz99: agreed, will change

clis/teliod/src/cgi/constants.rs
lcruz99: missed that, will change

clis/teliod/src/cgi/constants.rs
lcruz99: that's a good question 🤔 

CGI feature was initially thought exclusively for QNAP devices, thus only these two paths are being supported at the moment. If support for more devices is added then the CGI features needs eventually to be redesigned.

For this PR context let's keep as is.

clis/teliod/src/cgi/constants.rs
lcruz99: I think it depends on how much do we want to trust `getcfg` cmd.

Either way, I've changed to OsString but still had to get the &str to call .trim(). wdyt?

clis/teliod/src/cgi/constants.rs
tomasz-grz: Regardless of device, I strongly believe it's very bad practice to leave hard-coded temporary paths..

The following code does not even account for target architecture, like `aarch64-unknown-linux-musl` vs `x86_64-unknown-linux-gnu`, nor for custom build profiles besides `release` and `debug`. 

For example my current binary path is `libtelio/target/aarch64-unknown-linux-musl/release-owrt/teliod` but the path will be hard coded to `libtelio/target/release/teliod` which would not even exist for the build in my case..

What do you think about using the runtime value of [current_exe()](https://doc.rust-lang.org/std/env/fn.current_exe.html) instead?

If you prefer to fix it in a separate PR, please at least create a new ticked and add a `TODO` comment here.

tomasz-grz: Otherwise looks like `OUT_DIR` or `TARGET` and `PROFILE` could be used, but those environment variables are only set for build scripts, and would need to be exposed through `cargo:rustc-env` in `build.rs`... but this still leaves a temporary build path hard coded

clis/teliod/src/cgi/constants.rs
tomaszklak: If either way we need to get String/str to run trim, than there is no point - sorry, missed that.

lcruz99: reverted

tomaszklak (approved):

PR #1309, opened 2025-06-16
LLT-6394: Emit debug log when meshnet or feature config can't be deserialized

mathiaspeters:
### Problem
Deserializing configs return very opaque errors upon failure, which makes it hard for apps to find the cause. 

### Solution
When the deserialization fails, emit debug log

- [x] README.md is updated


Hasan6979 (approved):
+2

PR #1312, opened 2025-06-16
[LLT-6314] Prepare conntrack for chained fw

stalowyjez:
### Problem
Conntracker module needs a few adjustments before it can be used as a part of libfw

### Solution
Modify a bit the conntracker:
 - use a callback instead of a sink 
 - change fixed-length `PublicKey` to variable-length `associated_data`
 - add connection states


crates/telio-firewall/src/conntrack.rs
```diff
@@ -166,37 +179,48 @@ impl Conntracker {
...
         }
     }
 
-    pub fn handle_outbound_udp<'a>(&self, pubkey: PublicKey, ip: &impl IpPacket<'a>) {
+    pub fn handle_outbound_udp<'a>(&self, ip: &impl IpPacket<'a>, associated_data: &[u8]) {
         let link = unwrap_option_or_return!(Self::build_conn_info(ip, false)).0;
-        let key = Connection { link, pubkey };
+        let key = Connection {
+            link,
+            associated_data: associated_data.to_owned(),
```
tomaszklak: I think that this is a first (and only?) memory allocation that will be done per udp packet. Have you run the benchmarks? What is (if any) performance difference after this change?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -232,14 +257,23 @@ impl Conntracker {
                 last_chunk.clear();
                 last_chunk.extend_from_slice(pkg_chunk);
 
+                if value.is_remote_initiated
+                    && value.state == LibfwConnectionState::LibfwConnectionStateNew
+                {
+                    value.state = LibfwConnectionState::LibfwConnectionStateEstablished
```
tomaszklak, line 263: All udp "connections" that are **not** remote initiated, will always stay in the state "new" ?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -277,39 +312,50 @@ impl Conntracker {
...
-                    key,
-                    connection_info
+                    occ.key(),
+                    occ.get()
                 );
 
-                if connection_info.is_remote_initiated && drop {
-                    telio_log_trace!("Removing UDP conntrack entry {:?}", key);
+                if occ.get().is_remote_initiated
+                    && matches!(verdict, LibfwVerdict::LibfwVerdictDrop)
```
tomaszklak: nit: why use `matches!` here for `verdict` but few lines below `==` for `state`?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -114,10 +113,23 @@ impl Debug for IcmpConn {
...
+pub(crate) enum LibfwConnectionState {
+    LibfwConnectionStateEstablished,
+    LibfwConnectionStateNew,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum LibfwVerdict {
+    LibfwVerdictAccept,
+    LibfwVerdictDrop,
+}
```
tomaszklak: In both cases, please add derive `Copy`, see: https://doc.rust-lang.org/stable/std/marker/trait.Copy.html#when-should-my-type-be-copy

crates/telio-firewall/src/conntrack.rs
```diff
@@ -277,39 +312,50 @@ impl Conntracker {
...
                     "Matched UDP conntrack entry {:?} {:?}",
-                    key,
-                    connection_info
+                    occ.key(),
+                    occ.get()
                 );
 
-                if connection_info.is_remote_initiated && drop {
-                    telio_log_trace!("Removing UDP conntrack entry {:?}", key);
+                if occ.get().is_remote_initiated
```
tomaszklak: I think it would be better to convert this `if .. else if ... ` into `match (is_remot_initiated, state) { (true, Drop) => {.. }, (_, New) => {...} }`. This way if we ever add a new connection state, the compiler will mark this place as needing the update. With the current approach, this place will not generate any error/warning.

crates/telio-firewall/src/conntrack.rs
stalowyjez: But isn't it inevitable guess if we plan to have associated data of variable length, as in the API in RFC?

tomaszklak: If we expect that the only associated data will be the public key, we can use https://docs.rs/smallvec/latest/smallvec/

crates/telio-firewall/src/firewall.rs
```diff
@@ -391,7 +391,9 @@ impl StatefullFirewall {
 
                     _ = self.conntracker.send_icmp_port_unreachable_packets(
                         std::iter::once((&link, first_chunk)),
-                        sink,
+                        |packet: &[u8]| {
+                            let _ = sink.write_all(packet);
```
tomaszklak: This is ignoring errors that previously had been handled.

crates/telio-firewall/src/firewall.rs
```diff
@@ -405,7 +407,9 @@ impl StatefullFirewall {
                 if blacklist.contains(&SocketAddr::new(link.remote_addr.into(), link.remote_port)) {
                     _ = self.conntracker.send_tcp_rst_packets(
                         std::iter::once((&link, 0, Some(tcp_packet.get_sequence() + 1))),
-                        sink,
+                        |packet: &[u8]| {
+                            let _ = sink.write_all(packet);
```
tomaszklak: This is ignoring errors that previously had been handled.

crates/telio-firewall/src/firewall.rs
```diff
@@ -801,8 +840,12 @@ impl Firewall for StatefullFirewall {
     fn reset_connections(&self, pubkey: &PublicKey, sink: &mut dyn io::Write) -> io::Result<()> {
         telio_log_debug!("Constructing connetion reset packets");
 
-        self.conntracker.reset_tcp_conns(pubkey, sink)?;
-        self.conntracker.reset_udp_conns(pubkey, sink)?;
+        self.conntracker.reset_tcp_conns(pubkey, |packet: &[u8]| {
+            let _ = sink.write_all(packet);
```
tomaszklak: This is ignoring errors that previously had been handled.

crates/telio-firewall/src/firewall.rs
```diff
@@ -801,8 +840,12 @@ impl Firewall for StatefullFirewall {
...
+        })?;
+        self.conntracker.reset_udp_conns(pubkey, |packet: &[u8]| {
+            let _ = sink.write_all(packet);
```
tomaszklak: This is ignoring errors that previously had been handled.

crates/telio-firewall/src/conntrack.rs
```diff
@@ -90,7 +89,7 @@ pub(crate) struct IcmpConn {
     identifier_sequence: u32,
     // This public key refers to source peer for inbound connections and
     // destination peer for outbound connections
```
tomaszklak: This comment is no longer valid

crates/telio-firewall/src/conntrack.rs
```diff
@@ -79,7 +78,7 @@ pub(crate) struct Connection {
     pub(crate) link: IpConnWithPort,
     // This public key refers to source peer for inbound connections and
     // destination peer for outbound connections
```
tomaszklak: This comment is no longer valid

crates/telio-firewall/src/conntrack.rs
stalowyjez: Hmm, but these are different things - `Drop` is a verdict, while `New` is a connection state - it seems hard to refactor it to such `match` statement in a reasonable way

crates/telio-firewall/src/conntrack.rs
stalowyjez: I don't know - I think that changing this to more flexible API makes sense if we don't want to assume that anymore?

crates/telio-firewall/src/firewall.rs
stalowyjez: "previously had been handled" are quite big words - they were just forwarded and silenced here ^^'
I don't think this handling is that much worse now, but if you have some good idea how to improve it (assuming that this will be provided ), I'm all ears.

We could add some error type for the user callback so it can tell us whether it was successful or not - it is not really compatible with the callback type described in RFC - but we probably need to change it a bit anyway

crates/telio-firewall/src/conntrack.rs
stalowyjez: Maybe something like this?

crates/telio-firewall/src/firewall.rs
tomaszklak: The errors that happen in `reset_connections` are handled in `device.rs` so at least has to be preserved. In other cases if it's not practical to propagate the error, logging it at the trace level is a good alternative.

The `libfw_inject_packet_callback_t` returns `void` so I think it can easily be changed to return an enum indicating if it succeed or not (and why).

crates/telio-firewall/src/conntrack.rs
stalowyjez: Hmm, like this?

tomaszklak: I think you also want to convert associated data in `IcmpConn`. Also, and this is a small nitpick, instead of for example `SmallVec::from_slice(&peer)` you can just use ToSmallVec: `peer.to_smallvec()`.

stalowyjez: Hmm, about `IcmpConn` - unfortunately clippy complains a bit about it:
```
warning: the `Err`-variant returned from this function is very large
   --> crates/telio-firewall/src/conntrack.rs:604:10
    |
105 |     Icmp(Option<IcmpConn>),
    |     ---------------------- the largest variant contains at least 128 bytes
...
604 |     ) -> IcmpKey {
    |          ^^^^^^^
    |
    = help: try reducing the size of `conntrack::IcmpErrorKey`, for example by boxing large elements or replacing it with `Box<conntrack::IcmpErrorKey>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
    = note: `#[warn(clippy::result_large_err)]` on by default
    ```

File: crates/telio-firewall/src/conntrack.rs [hunk 12]
```diff
@@ -676,7 +758,9 @@ impl Conntracker {
         );
 
         let iter = tcp_conn_cache.iter().filter_map(|(k, v)| {
-            if k.pubkey == *pubkey {
+            if k.associated_data
+                == SmallVec::<[u8; size_of::<PublicKey>()]>::from_slice(associated_data)
+            {
```
tomaszklak: There is no need to construct new smallvec just to compare a slice. It should be enough to `&*k.associated_data == associated_data`.

crates/telio-firewall/src/conntrack.rs
```diff
@@ -77,9 +78,10 @@ impl Debug for IpConnWithPort {
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
 pub(crate) struct Connection {
     pub(crate) link: IpConnWithPort,
-    // This public key refers to source peer for inbound connections and
-    // destination peer for outbound connections
-    pub(crate) pubkey: PublicKey,
+    // Data associated with the packets, usually the public keys refering
+    // to source peer for inbound connections and destination peer for outbound
+    // connections, which may be used for tracking and filtering packets
+    pub(crate) associated_data: SmallVec<[u8; size_of::<PublicKey>()]>,
```
tomaszklak: nit: I'm not sure that using `size_of` is a good idea here. I would rather prefer we use the `telio_crypto::KEY_SIZE`. The size of the `PublicKey` might change in the future if we want to add some fields to it, but the size of the wireguard key is far less likely to change.

stalowyjez: changed

crates/telio-firewall/src/conntrack.rs
stalowyjez: Changed that, I'm not that much convinced that dereferencing a vector is intuitive ^^'

crates/telio-firewall/src/firewall.rs
stalowyjez: Added some result to the callbacks

crates/telio-firewall/src/conntrack.rs
stalowyjez: Boxed `IcmpConn` and rewrote it to `SmallVec`s

crates/telio-firewall/src/conntrack.rs
tomaszklak: I think it's pretty common pattern in rust code, but here it also lets us avoid making a copy of associated data just to compare against it and discard it right after.

crates/telio-firewall/src/firewall.rs
```diff
@@ -1486,10 +1524,10 @@ pub mod tests {
                 local_addr: test_input.us_ip(),
                 local_port: test_input.us_port(),
             };
-            let tcp_key = Connection { link , pubkey: PublicKey(peer) };
+            let tcp_key = Connection { link , associated_data: SmallVec::from_slice(&peer) };
```
tomaszklak: nit: here also to_smallvec can be used

tomaszklak:
Oh, I forgot (but I already mentioned it in one of the threads). Please post a comparison of benchmarks (you need to run them on your local machine, when it's 'quiet' nothing more is happening on it. And when it's connected to the outlet, not on a battery). Something like what @Hasan6979 did in #529

crates/telio-firewall/src/conntrack.rs
```diff
@@ -277,39 +315,48 @@ impl Conntracker {
...
+                    occ.key(),
+                    occ.get()
                 );
 
-                if connection_info.is_remote_initiated && drop {
-                    telio_log_trace!("Removing UDP conntrack entry {:?}", key);
+                if occ.get().is_remote_initiated && verdict == LibfwVerdict::LibfwVerdictDrop {
+                    telio_log_trace!("Removing UDP conntrack entry {:?}", occ.key());
                     occ.remove();
+                } else if let LibfwConnectionState::LibfwConnectionStateNew = occ.get().state {
```
jjanowsk: Nit: why not just use `if ... == ... {` for enums instead of `if let`? Feels weird at first

crates/telio-firewall/src/conntrack.rs
```diff
@@ -232,14 +260,23 @@ impl Conntracker {
...
+                    value.state = LibfwConnectionState::LibfwConnectionStateEstablished
+                }
+
                 value.last_out_pkg_chunk = Some(last_chunk);
             }
         }
     }
 
-    pub fn handle_outbound_tcp<'a>(&self, pubkey: PublicKey, ip: &impl IpPacket<'a>) {
+    pub fn handle_outbound_tcp<'a>(&self, ip: &impl IpPacket<'a>, associated_data: &[u8]) {
```
jjanowsk: I'm missing setting `LibfwConnectionStateEstablished` for TCP connections that are remote initiated. It should be done in `handle_outbound_tcp`

jjanowsk (commented):
I see one issue

crates/telio-firewall/src/conntrack.rs
tomaszklak: Agree :)

crates/telio-firewall/src/conntrack.rs
tomaszklak: Hm, remote initiated connection should be marked as established when the handshake (SYN, SYN-ACK, ACK) is completed which is in this case when we receive (inbound) 3rd (in the stream) tcp packet with flags set to ACK.

Did you mean the **not** remote initiated connection?

jjanowsk: I meant remote initiated. Usually from connection tracking perspective the TCP connection is considered established when the packet was seen flowing from both directions. TCP handshake is not considered. Imagine you want to create iptables rules which allow only new outbound tcp connections. Usually you do it by allowing outbound packets for new and established connection state and allow inbound packets only in established state. Now imagine this situation: first SYN packet flowing out, the connection is now in the new state. But if the SYN-ACK flowing back does not change the state to established it will never be let in. Of course conntrack also keeps more detailed states for the TCP like SYN_SENT, SYN_RECV, ESTABLISHED etc. but I don't think that should be the meaning of our LibfwConnectionStateEstablished.

crates/telio-firewall/src/conntrack.rs
```diff
@@ -263,12 +302,16 @@ impl Conntracker {
                 let TcpConnectionInfo { tx_alive, .. } = e.get_mut();
                 *tx_alive = false;
             }
+        } else if flags & syn_ack == syn_ack {
```
jjanowsk: I'm not sure that we need to check flags here, but also I think it does not matter to much at this stage.  It seems that this part will need some refactoring anyway. In the current state we remove connection from the cache when we receive RST packet, however when we refactor our firewall to handle filter rules like "accept incomming connections in established state" we will need some way to pass the information about connection state back to the firewall. And then the connection where a RST packet is returning after a SYN packet was sent should be considered established so that the RST packet can reach the destination. I don't  know if this refactoring should be the part of this PR or not.

jjanowsk: I've now looked into LLT-6314 and it seems that this also should be part of this PR

jjanowsk: Also I just noticed that in the current implementation the FW rules are evaluated before the conntrack which is kind of backwards. For the reasons described above conntrack should be the first module that the packet hits, and after the connection state is updated the FW rules should be applied. At least that's the linux behavior and if we want to more or less follow it as described in the RFC then I think that keeping it backwards will make it much harder if not impossible.

stalowyjez: @jjanowsk Hmm, currently we also depend on information from firewall when deleting conntrack entries - I guess we can separate it, so the flow is like:
 1. Updating conntracker entries
 2. Filtering
 3. Cleaning up tracked connections for certain packets
 
 (so I pushed some commit with such solution here)

crates/telio-firewall/src/conntrack.rs
```diff
@@ -263,103 +262,156 @@ impl Conntracker {
...
+    }
+
+    pub fn handle_inbound_udp<'a>(
+        &self,
+        ip: &impl IpPacket<'a>,
+        associated_data: &[u8],
+    ) -> LibfwConnectionState {
+        let (link, _) = unwrap_option_or_return!(
+            Self::build_conn_info(ip, true),
+            LibfwConnectionState::LibfwConnectionStateNew
```
Jauler: I see there is a bit of a common theme that in case conntrack fails -> silently (or, in some cases, with logs) convert the state to `New`.  And I wonder if it would be better to leave this decision on what to do when connection tracking fails for a packet to the top-level module (in our case it would be firewall.rs) and from conntrack.rs just return `Result<LibfwConnectionState, ....>`  instead of `LibfwConnectionState`. 

This would mean - on the other hand, that various functions, like `build_conn_info()` would have to also return `Result<(IpConnWithPort, Option<TcpPacket>)>`

@stalowyjez what do you think?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -401,87 +460,104 @@ impl Conntracker {
...
             IcmpErrorKey::Tcp(Some(link)) => {
-                let tcp_key = Connection { link, pubkey };
+                let tcp_key = Connection {
+                    link,
+                    associated_data: associated_data.to_smallvec(),
+                };
 
-                let mut tcp_cache = unwrap_lock_or_return!(self.tcp.lock(), false);
+                let mut tcp_cache = unwrap_lock_or_return!(
+                    self.tcp.lock(),
```
Jauler: Referencing: https://github.com/NordSecurity/libtelio/pull/1312/files#r2185431747 - for example here, if we fail to take `self.tcp.lock()` -> it does not mean, that connection state is `New`. Hence indicating that with `Result<..., ...>` sounds like should avoid possible confusions. 

What is more, such a failure would be silent, without indicating any issues in the logs. 

There is of course and example, where logs are being used, but still the decision is not at the top-level component:
- https://github.com/NordSecurity/libtelio/pull/1312/files#diff-eb66c1ba97e4cb145ebe52b4a190ff8ac3251db1011f25715566d2705302aa8fR573
- https://github.com/NordSecurity/libtelio/pull/1312/files#diff-eb66c1ba97e4cb145ebe52b4a190ff8ac3251db1011f25715566d2705302aa8fR583
- https://github.com/NordSecurity/libtelio/pull/1312/files#diff-eb66c1ba97e4cb145ebe52b4a190ff8ac3251db1011f25715566d2705302aa8fR603

Any of these if hit -> produces a hard-to-track-down issue.

What is more, for ICMP this is already done using `Result<>` which would mean, that TCP/UDP is kind of mismatching ICMP, and we could unify all of them :).

stalowyjez: Hmm, in case of ICMP this `Result` is a bit abused, I'm afraid - because you mean `IcmpKey = Result<IcmpConn, IcmpErrorKey>`, don't you?

crates/telio-firewall/src/conntrack.rs
stalowyjez: Yeah, sounds like a good idea - it'll try to do that now

crates/telio-firewall/src/conntrack.rs
```diff
@@ -166,37 +177,12 @@ impl Conntracker {
...
-        let mut tcp_cache = unwrap_lock_or_return!(self.tcp.lock(), None);
-        match tcp_cache.entry(key, false) {
-            Entry::Occupied(occupied_entry) => Some(occupied_entry.get().clone()),
-            Entry::Vacant(_) => None,
-        }
-    }
-
-    pub fn handle_outbound_udp<'a>(&self, pubkey: PublicKey, ip: &impl IpPacket<'a>) {
+    pub fn handle_outbound_udp<'a>(&self, ip: &impl IpPacket<'a>, associated_data: &[u8]) {
         let link = unwrap_option_or_return!(Self::build_conn_info(ip, false)).0;
```
Jauler: Maybe it would be clearer if `build_conn_info` would accept packet direction as an enum (with variants `inbound` and `outbound`) instead of bool :thinking: 

Altgouh this is a nitpick.

crates/telio-firewall/src/conntrack.rs
stalowyjez: @Jauler sth like this?

crates/telio-firewall/src/conntrack.rs
Jauler: Yeah, `IcmpKey` type alias probably should not be `Result<...>`, but I was thinking that `build_conn_info()` and `build_icmp_key()` could return `Result<XXXX_Key>` to avoid assiging state `New` for cases, where we could not determine actual state. And leave the decision on what to do with such packets to top-level components.

crates/telio-firewall/src/conntrack.rs
Jauler: its in the right direction, but now when `build_conn_info()` errors out, it just returns "hardcoded" `NotFound` error (it is done in the `unwrap_option_or_return_err` macro), and IMHO it would be better to make `build_conn_info()` to return an Result and for `process_[out|in]bound_[tcp|udp|icmp]` to propogate that error transparently upwards if it is not `Ok()` (e.g. by using `?` or similar). Although, this will require adjusting return types a bit 

This way we achieve both: the "intermediate" functions like `handle_outbound_tcp` becomes simpler, and we have more information about the failure at the top-level component (e.g. like `firewall.rs` in our case).

I am pretty sure there should be some sort of name for such a pattern, where errors are instantiated in "leaf" functions, but handled at "top-level" functions, as one has the most details about the error itself, and the latter has the most context on how to handle errors. And intermediate functions just pass error transparently.

WDYT?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -41,6 +41,18 @@ macro_rules! unwrap_lock_or_return {
...
+        match $guard {
+            Ok(x) => x,
+            Err(_poisoned) => {
+                telio_log_error!("Poisoned lock");
+                return Err(io::Error::from(ErrorKind::InvalidData));
+            }
+        }
+    };
+}
+
```
tomaszklak: Imo, if all we do with poisoned lock is to print an error (which I've never seen in prod or nat-lab) and return a generic error, it's better to use parking_lot's mutex.

crates/telio-firewall/src/conntrack.rs
```diff
@@ -136,6 +177,8 @@ pub(crate) struct TcpConnectionInfo {
     pub(crate) rx_alive: bool,
     pub(crate) conn_remote_initiated: bool,
     pub(crate) next_seq: Option<u32>,
+    pub(crate) state: LibfwConnectionState,
+    pub(crate) to_delete: bool,
```
tomaszklak: Can you add a comment explaining why `to_delete` field has to be part of this struct?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -144,7 +187,7 @@ pub(crate) struct Conntracker {
     /// Recent tcp connections
     pub(crate) tcp: Mutex<LruCache<Connection, TcpConnectionInfo>>,
     /// Recent icmp connections
-    pub(crate) icmp: Mutex<LruCache<IcmpConn, ()>>,
+    pub(crate) icmp: Mutex<LruCache<IcmpConn, bool>>,
```
tomaszklak: Please update the comment with an explanation what is the meaning of the key type. There are now 3 possible statese (missing, present + true, present + false) and I'm not sure what they mean.

crates/telio-firewall/src/conntrack.rs
Jauler: @jjanowsk as we already moved conntrack processing to before filtering - can this thread be resolved?

crates/telio-firewall/src/conntrack.rs
Jauler: > Usually from connection tracking perspective the TCP connection is considered established when the packet was seen flowing from both directions.

I am wondering, is it really safe to base all connection tracking just on seeing replies or not :thinking:

crates/telio-firewall/src/conntrack.rs
stalowyjez: That was some leftover from an intermediate solution, removed

crates/telio-firewall/src/conntrack.rs
stalowyjez: Same as with `to_delete` ^^'

crates/telio-firewall/src/conntrack.rs
Jauler: I would agree

stalowyjez: Sooo, blacklists and whitelists should be also behind parking-lot `RwLocks`?

stalowyjez: Moved to `parking-lot`

crates/telio-firewall/src/conntrack.rs
stalowyjez: Changed it to some enum

tomaszklak (commented):
still +1, but it seems to me that the commit history got messy (and has at least one typo: `whtlisted`).

crates/telio-firewall/Cargo.toml
```diff
@@ -23,6 +23,8 @@ telio-crypto.workspace = true
 telio-utils.workspace = true
 telio-model.workspace = true
 telio-network-monitors.workspace = true
+smallvec = "1.15.1"
+parking_lot.workspace = true
```
jjanowsk: We keep external dependencies one section above and in an alphabetical order :)

crates/telio-firewall/src/conntrack.rs
```diff
@@ -114,10 +101,50 @@ impl Debug for IcmpConn {
...
+#[allow(clippy::enum_variant_names)]
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LibfwConnectionState {
+    /// Connection established
+    LibfwConnectionStateEstablished,
+    /// Outgoing packets
+    LibfwConnectionStateNew,
+    /// Finished connections
+    LibfwConnectionStateFinished,
```
jjanowsk: Isn't 'closed' a better word in this context?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -252,114 +276,175 @@ impl Conntracker {
...
             telio_log_trace!("Connection {:?} closing", key);
             if let Entry::Occupied(mut e) = tcp_cache.entry(key, false) {
                 let TcpConnectionInfo { tx_alive, .. } = e.get_mut();
                 *tx_alive = false;
+                Ok(LibfwConnectionState::LibfwConnectionStateEstablished)
+            } else {
+                Ok(LibfwConnectionState::LibfwConnectionStateNew)
             }
+        } else if let Some(tcp_connection_info) = tcp_cache.get_mut(&key) {
+            if flags & syn_ack == syn_ack {
```
jjanowsk: I think this shouldn't be here and instead it should just check that the connection is remote_initiated. In this case the outgoing packet is a reply. In the current implementation if you send malicious syn and syn/ack from the host it will switch to the established state.

crates/telio-firewall/src/conntrack.rs
```diff
@@ -252,114 +276,175 @@ impl Conntracker {
...
                 e.insert(());
             }
         }
+
+        Ok(LibfwConnectionState::LibfwConnectionStateNew)
     }
 
-    pub fn handle_inbound_udp<'a>(&self, drop: bool, peer: &PublicKey, ip: &impl IpPacket<'a>) {
-        let (link, _) = unwrap_option_or_return!(Self::build_conn_info(ip, true));
+    pub(crate) fn cleanup_udp_conn_on_drop<'a, P: IpPacket<'a>>(
```
jjanowsk, line 346: Why do we need those cleanups? Can't they just die out?

crates/telio-firewall/src/conntrack.rs
```diff
@@ -388,11 +477,18 @@ impl Conntracker {
                     connection_info.next_seq = Some(next_seq);
                     if is_conn_reset {
                         occ.remove();
+                        return Ok(LibfwConnectionState::LibfwConnectionStateFinished);
+                    } else {
+                        // Connection which is not removed and packets are going in both directions should be established
+                        occ.get_mut().state = LibfwConnectionState::LibfwConnectionStateEstablished;
```
jjanowsk, line 503: Only if this connection was not remote initiated

crates/telio-firewall/src/conntrack.rs
```diff
@@ -545,7 +652,7 @@ impl Conntracker {
             // Request messages get the icmp type of the corresponding reply message
             // Reply messages get their own icmp type, to match that of its corresponding request
             // We only create keys for outbound requests and inbound replies
-            if inbound {
+            if let LibfwDirection::LibfwDirectionInbound = direction {
```
jjanowsk, line 675: In general do you find this: `if let LibfwDirection::LibfwDirectionInbound = direction` more readable then: `if LibfwDirection::LibfwDirectionInbound == direction`? I find it strange

crates/telio-firewall/src/firewall.rs
```diff
@@ -364,58 +345,101 @@ impl StatefullFirewall {
         }
     }
 
+    fn track_outbound_ip_packet<'a, P: IpPacket<'a>>(
```
jjanowsk: Shouldn't those two functions be part of conntracker? And this should be a public interface instead of all handle_outbound_* functions?

crates/telio-firewall/src/conntrack.rs
jjanowsk: Please first add tests for tcp and udp with two packets coming from the same direction (for both directions). They should not update the state to established.

crates/telio-firewall/src/conntrack.rs
Jauler: I do not think it is good to ignore the flags. For example, lets imagine a case where we have:
1. host A sends syn towards host B
2. host B responds with TCP RST
3. later host B sends `syn` as new connection on the same 5-tuple
4. now firewall will incorrectly assume that this `syn` (and any further traffic) is attributed to the connection coming from host A to host B while in reality the direction is the reverse

stalowyjez: I think it's not harmful to check them - although RST itself is handled a bit earlier, so maybe this is not the best example ^^'

Jauler: This discussion relates to this: https://github.com/NordSecurity/libtelio/pull/1312#discussion_r2192119526 as well

crates/telio-firewall/src/conntrack.rs
stalowyjez: They could - the behavior of this part of the code just changed a lot recently, so I feel a bit unsafe about making this change even larger, so I think we can test it a bit and then add it as a next step? WDYT?

crates/telio-firewall/src/conntrack.rs
stalowyjez: I just do it for all enums, I don't have any strong preference, though

crates/telio-firewall/src/conntrack.rs
Jauler: Additionally, I was unsure how Linux conntrack handles this, hence I did experimental test in nat-lab.

I connected to `cone-gw-1` and started monitoring conntrack events. Upon TCP RST -> conntrack entry is immediatelly cleared (no time-wait), allowing next connection to be tracked correctly. 

I have not explored linux kernel code enough yet to figure out details about this though.

jjanowsk: Your test illustrates an exception I've mentioned in our Slack discussion quite some time ago. It is done as an optimization, because it's a common case to send a RST to the SYN to block the connection. It is done here: https://elixir.bootlin.com/linux/v6.15/source/net/netfilter/nf_conntrack_proto_tcp.c#L1304 
However, I don't know how exactly other invalid (from TCP POV) packets are handled. That would require more testing. I think we can leave this here for now. It's safer this way then the other and if we find some exceptions in the future or we find that we should not check it at all then we can modify it accordingly. 

But the second part of my comment stays. We should consider direction in this check. Do we have an agreement on that part?

crates/telio-firewall/src/conntrack.rs
jjanowsk: Sure, this can be optimized/refactored in the future

crates/telio-firewall/src/conntrack.rs
Jauler: > But the second part of my comment stays. We should consider direction in this check. Do we have an agreement on that part?

yup, makes sense

crates/telio-firewall/Cargo.toml
stalowyjez: Fixed

crates/telio-firewall/src/conntrack.rs
stalowyjez: changed

crates/telio-firewall/src/conntrack.rs
stalowyjez: Hmm, so I guess I should leave SYN-ACK check for now and just add is_conn_remote_initialized check?

crates/telio-firewall/src/conntrack.rs
stalowyjez: So leaving them as they are for now

crates/telio-firewall/src/conntrack.rs
stalowyjez: Added the requirement, also the test

crates/telio-firewall/src/firewall.rs
stalowyjez: Moved them

Jauler (approved):
~~+1.0~~

~~But maybe it is worth considering adding some sort of string at startup in logs, like this: https://github.com/NordSecurity/NepTUN/commit/e1620ebb102dad16dfbb81d31ecef67afe3bc549~~

Opps commented on a wrong PR.

jjanowsk (approved):
+1

PR #1314, opened 2025-06-17
add logs to firewall and direct tests

gytsto:
### Problem
Tests are failing and we don't know why

### Solution
Add some logs to make it more clear


tomaszklak (approved):
+2

PR #1315, opened 2025-06-17
LLT-6138: Add UCI support to teliod

tomasz-grz:
### Problem
For the OpenWRT there is a native command [UCI](https://openwrt.org/docs/techref/uci) (Unified Configuration Interface), which should be used to configure networks.

### Solution
Add `InterfaceConfigurationProvider` implementation for `UCI`
Rename `adapter_name` to `interface_name` for consistency.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/configure_interface.rs
+++ b/clis/teliod/src/configure_interface.rs
@@ -62,13 +62,13 @@ pub enum InterfaceConfigurationProvider {
 // TODO(tomasz-grz): Try using enum_dispatch instead of dynamic dyspatch
 impl InterfaceConfigurationProvider {
     /// Create a dynamic instance of a configuration provider
-    pub fn create(&self, adapter_name: String) -> Box<dyn ConfigureInterface> {
+    pub fn create(&self, interface_name: String) -> Box<dyn ConfigureInterface> {
         info!("Creating interface config provider for {:?}", self);
         match self {
             Self::Manual => Box::new(Manual),
-            Self::Ifconfig => Box::new(Ifconfig::new(adapter_name)),
-            Self::Iproute => Box::new(Iproute::new(adapter_name)),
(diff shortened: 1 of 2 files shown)


clis/teliod/src/configure_interface.rs
```diff
@@ -125,3 +127,41 @@ impl ConfigureInterface for Iproute {
...
+            }
+            Err(_) => {
+                debug!("adding new interface");
+                execute(Command::new("uci").args(["add", "network", "interface"]))?;
+                execute(Command::new("uci").args(["rename", "network.@interface[-1]=tun"]))?;
+            }
+        }
+
+        // set options
+        execute(Command::new("uci").args(["set", &format!("network.tun.device={adapter_name}")]))?;
```
Hasan6979: is adapter name name of the tunnel or which wireguard implementation to be used ?

tomasz-grz: It's the name of the tunnel, the one which you would get when listing `ip a`, and set from the `interface.name` in the `teliod_config.json`.

Confusingly, `interface` in OpenWRT UCI is something else..

Hasan6979: nit: Can we have that to something else ? `tunnel_name` or `interface_name` ? Because in libtelio we use adapter to specify Wireguard implementations so it seems slightly confusing. This isn't a blocker, and I leave it to your judgement

tomasz-grz: Making it `interface_name` to be consistent with the config makes sense

tomasz-grz: Renamed

stalowyjez (approved):
Looks good to me, +1

stalowyjez (approved):

PR #1321, opened 2025-06-20
IPv6 compressed form, when nslookup'ing

matislovas:
### Problem
We're storing IPv6 addressees in different form than `nslookup` returns results. Thus, sometimes strings does not match.

### Solution
Use compressed form of IPv6 addresses, when doing `nslookup`.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/helpers.py
+++ b/nat-lab/tests/helpers.py
@@ -4,6 +4,7 @@ import json
 import pytest
 from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass, field
+from ipaddress import AddressValueError, IPv6Address
 from itertools import product, zip_longest
 from mesh_api import Node, API
 from telio import Client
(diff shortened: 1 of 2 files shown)


LukasPukenis (dismissed):
+1

nat-lab/tests/helpers.py
```diff
@@ -507,3 +508,18 @@ def verify_uuid(uuid_to_test, version=4):
     except ValueError:
         assert False, "Not a valid UUID"
     assert str(uuid_obj) == uuid_to_test, "Not a valid UUID"
+
+
+def string_to_compressed_ipv6(ip_str_list: List[str]) -> List[str]:
+    """
+    Pick ipv6 address from a list aand convert them to compressed form
```
tomasz-grz: nit: `and`

nat-lab/tests/test_dns.py
```diff
@@ -444,7 +459,10 @@ async def test_dns_after_mesh_off(alpha_ip_stack: IPStack) -> None:
 
         # If the previous calls didn't fail, we can assume that the resolver is running so no need to wait for the timeout and test the validity of the response
         await query_dns(
-            connection_alpha, "beta.nord", beta.ip_addresses, dns_server_address
+            connection_alpha,
+            "beta.nord",
+            string_to_compressed_ipv6(beta.ip_addresses),
```
tomasz-grz, line 464: Instead of adding `string_to_compressed_ipv6` to every `query_dns()` call, what do you think about doing it once internally in `query_dns()` of `utils/dns.py`?

matislovas: That was the plan, but than I inspected all the `query_dns` calls and sometimes the called is looking for ordinary strings, e.g.:

```
        await query_dns(
            connection_alpha,
            "google.com",
            dns_server=dns_server_address_local,
            expected_output=["Name:.*google.com.*Address"],
        )
```

nat-lab/tests/helpers.py
matislovas: fixed

nat-lab/tests/test_dns.py
tomasz-grz: hmm ok they, makes sense

tomasz-grz (approved):
+1

PR #1322, opened 2025-06-20
LLT-6370: Add identity file cleanup

lcruz99:
### Problem
Teliod works by first checking if device identity file exists and if it does, it assumes that identity by parsing that file. 
It then tries to update the machine info on the API which might fail if the parsed identity doesn’t match the provided auth token, breaking the whole initialization chain.
In addition to this, the file is also preserved after uninstalls, which requires the user to explicitly delete the file usually at ~/admin/.local/share/teliod/data.json in order to fix this issue.

### Solution
Delete device identity file whenever the app is uninstalled or if the cached machine is not found in the API for the given auth token.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -56,8 +56,8 @@ pub enum Error {
     InvalidResponse,
     #[error("Unable to update machine due to Error: {0}")]
     UpdateMachine(StatusCode),
-    #[error("Max retries exceeded for connection: {0}")]
-    ExpBackoffTimeout(String),
+    #[error("Max retries exceeded for connection")]
+    ExpBackoffTimeout(),
     #[error("No local data directory found")]
     NoDataLocalDir,
     #[error("Device not registered with server")]
@@ -295,7 +295,11 @@ async fn update_machine_with_exp_backoff(
                 if let Error::UpdateMachine(_) = e {
                     return Err(e);
                 }
-                wait_with_backoff_delay(&mut backoff, &mut retries, "update machine", e).await?;
+                warn!(
+                    "Failed to update machine due to {e}, will wait for {:?} and retry",
+                    backoff.get_backoff()
+                );
+                wait_with_backoff_delay(&mut backoff, &mut retries).await?;
             }
         }
     }
@@ -312,13 +316,13 @@ async fn fetch_identifier_with_exp_backoff(
     loop {
         match fetch_identifier_from_api(auth_token, cert_path, public_key).await {
             Ok(id) => return Ok(id),
-            Err(e) => match e {
-                Error::FetchingIdentifier(_) | Error::DeviceNotFound => return Err(e),
-                _ => {
-                    wait_with_backoff_delay(&mut backoff, &mut retries, "fetch identifier", e)
-                        .await?;
-                }
-            },
+            Err(e) => {
+                warn!(
+                    "Failed to fetch identifier due to {e:?}, will wait for {:?} and retry",
+                    backoff.get_backoff()
+                );
+                wait_with_backoff_delay(&mut backoff, &mut retries).await?;
+            }
         }
     }
 }
@@ -338,8 +342,11 @@ async fn register_machine_with_exp_backoff(
             Err(e) => match e {
                 Error::PeerRegistering(_) => return Err(e),
                 _ => {
-                    wait_with_backoff_delay(&mut backoff, &mut retries, "register machine", e)
-                        .await?;
+                    warn!(
+                        "Failed to register machine due to {e}, will wait for {:?} and retry",
+                        backoff.get_backoff()
+                    );
+                    wait_with_backoff_delay(&mut backoff, &mut retries).await?;
                 }
             },
         }
@@ -349,16 +356,10 @@ async fn register_machine_with_exp_backoff(
 async fn wait_with_backoff_delay(
     backoff: &mut ExponentialBackoff,
     retries: &mut usize,
-    action: &str,
-    error: Error,
 ) -> Result<(), Error> {
-    warn!(
-        "Failed to {action} due to {error}, will wait for {:?} and retry",
-        backoff.get_backoff()
-    );
     tokio::time::sleep(backoff.get_backoff()).await;
     if *retries > MAX_RETRIES {
-        return Err(Error::ExpBackoffTimeout(action.to_string()));
+        return Err(Error::ExpBackoffTimeout());
     }
     backoff.next_backoff();
     *retries += 1;


clis/teliod/src/core_api.rs
```diff
@@ -64,6 +76,126 @@ struct MeshConfig {
...
+
+impl DeviceIdentity {
+    /// Generate a new device identity.
+    /// A new key and hw id are generated, whereas the machine id is fetched from the
+    /// API with a [1,180]s exponential backoff.
+    pub async fn new(auth_token: &str) -> Result<Self, Error> {
+        let private_key = SecretKey::gen();
+        let hw_identifier = uuid::Uuid::new_v4();
+
+        // LLT-6406: There's likely a bug here since looking for a device with a private key that
```
LukasPukenis, line 98: I am not sure if I understand the wording here - "is trivial and will never be found" :thinking:

qnap/package_routines
```diff
@@ -113,7 +113,8 @@ PKG_POST_REMOVE="{
     $CMD_RM -rf /tmp/nordsecuritymeshnet
     $CMD_RM -f /home/Qhttpd/Web/NordSecurityMeshnet
     $CMD_RM -f /home/httpd/cgi-bin/qpkg/teliod.cgi
-    $CMD_RM -f /var/log/teliod.log
+    $CMD_RM -f /var/log/teliod*
+    $CMD_RM -rf ~/.local/share/teliod/
```
LukasPukenis, line 117: I thought these two lines is gonna cover the whole PR :)

LukasPukenis (commented):
Looks like a nice rework of wrapping `Hidden<String>` into it's own custom type. Several thoughts:

- Empty token shouldn't really be considered a valid token in my opinion. When processing `POST /meshnet` - we don't want to change the token to empty on a form submit(and we don't want to embed the token on the client as it's bad security), so as a cheap way we can interpret empty value as a no-op, but that should be limited just in the context of the REST endpoint handler and should not be considered a valid value. Reading empty token from the config file should not result in a valid `NordToken` but rather in a `None` or something similar. What's your opinion on this? Obviously - a token of the right format can still be invalid but we want to minimize the surface of it.

- There are `teliod` tests in `test_teliod.py`. It would be very nice to add additional testing conditions regarding this PR: `data.json` generation, validation, removal of it(since now there's extra logic to handle that). I found that `core-api.py` has the `NOT_FOUND` defined so we should be able to exercise daemon.rs:198 logic.

What are your opinions on these?

qnap/package_routines
lcruz99: 😄 It's 50% of the PR, it cleans up the identity file but only when uninstalling. Ideally the user should be switchable without having to reinstall the app.

clis/teliod/src/core_api.rs
lcruz99: > looking for a device with a private key that we've just created is trivial and will never be found

The API will obviously never find a private key which was just created a few lines above ([line 91](https://github.com/NordSecurity/libtelio/pull/1322/files/e5b39e882cfc1088b12c535112bc5f2ca058a311#diff-feb6793625ca425fad029689316a4eede8c49ea3bb122669978c13d2ff9ea5c6R91)), making the call to fetch_identifier_with_exp_backoff() trivial because the key will never be found.

Maybe I should reword it 🤔

clis/teliod/src/core_api.rs
```diff
@@ -64,6 +76,126 @@ struct MeshConfig {
...
+                    }
+                    Err(e) => {
+                        debug!("Could not remove directory {}: {}", parent.display(), e);
+                    }
+                };
+            };
+        }
+    }
+
+    fn device_identity_path() -> Result<&'static Path, Error> {
```
tomasz-grz, line 202: Why do we return a `&'static Path` instead of a `PathBuf`?

clis/teliod/src/core_api.rs
```diff
@@ -26,6 +25,17 @@ const OS_NAME: &str = "linux";
 type MachineIdentifier = String;
 const MAX_RETRIES: usize = 10;
 
+static DEVICE_IDENTITY_FILE_PATH: LazyLock<Result<PathBuf, Error>> = LazyLock::new(|| {
```
tomasz-grz, line 32: What is the benefit of declaring a static `LazyLock` instead of just computing
`dirs::data_local_dir()` in `device_identity_path()`?

tomasz-grz (commented):
Just some questions, otherwise nice refactoring! 💪

clis/teliod/src/core_api.rs
lcruz99: Because it's not worth to allocate memory and returned an owned type for a value that is known at compile time and it is not expected to change for the whole program duration.

It's then up to the callers if they want to create a new value or just use a reference to this.

clis/teliod/src/core_api.rs
lcruz99: You don't have to compute and allocate memory for this value everytime you need it because it will always be constant during the program duration thus only one instance of this value is required to exist (static, not const).
It can be accessed from multiple threads asynchrousnly (threadsafe) and its guaranteed to be initialized only once (on the first access aka lazy init).

https://doc.rust-lang.org/std/keyword.static.html
https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html

clis/teliod/src/core_api.rs
lcruz99: It's the same relation as `&'static str` and `String`

clis/teliod/src/core_api.rs
LukasPukenis: Still not clear to me why we should not find it.. 🤔

LukasPukenis (commented):
I expressed my concerns about introduced extra logic regarding data.json removal and creation and missing NatLab tests for those cases, however I do not see problems with the refactoring itself.

+1 given that the tests would be added :)

clis/teliod/src/core_api.rs
tomasz-grz: I see thank you for the clarification. It's a `PathBuf` so would it still be allocated on the heap though?

I was wondering what's the motivation of keeping the memory allocated for this path through out the lifetime of the whole program, instead of only the lifetime of `DeviceIdentity`, if it was stored in `self`? for example like this?

```rust
pub struct DeviceIdentity {
    file_path: PathBuf,
    pub hw_identifier: uuid::Uuid,
    pub private_key: SecretKey,
    pub machine_identifier: String,
}

impl DeviceIdentity {
    pub async fn new<P: AsRef<Path>>(auth_token: &str, file_path: P) -> Result<Self, Error> {...}
    pub fn from_file<P: AsRef<Path>>(file_path: P) -> Result<Self, Error> {...}
}
```

clis/teliod/src/core_api.rs
tomasz-grz: I wonder if this check should actually be in `from_file()` method?

When we create a new `private_key` we expect that `fetch_identifier` would always fail since we didn't register it yet.. But when we read the `private key` `from_file()`, maybe it's good to check whether it's still valid?

clis/teliod/src/core_api.rs
tomasz-grz: Hmm I see now, calling `remove_file(mut self)` without instantiating `DeviceIdentity` first, would not be very ergonomic (or in case `new()` or `from_file()` fails for some reason.)

clis/teliod/src/core_api.rs
lcruz99: As @tomasz-grz said, we won't find any device with that private key because we've just created it and never registered it, therefore there will be no device with that private key, making the fetch_identifier_with_exp_backoff() call trivial.

lcruz99: > I wonder if this check should actually be in from_file() method?

Probably, I just didn't want to touch the logic as it's out of scope for this PR, hence the TODO.

lcruz99:
@LukasPukenis While I was adding test cases I've found that core-api logic is flawed when the identity file is not present, mostly related with [this](https://github.com/NordSecurity/libtelio/pull/1322/files#diff-feb6793625ca425fad029689316a4eede8c49ea3bb122669978c13d2ff9ea5c6R98) TODO.

That issue is tackled on https://github.com/NordSecurity/libtelio/pull/1390 and I'll include the test cases you very well mentioned.

PR #1324, opened 2025-06-21
LLT-6314: Rewrite firewall to chains

stalowyjez:
### Problem
We want to switch from the current firewall logic to chain based libfirewall. The transition will be smoother if we rewrite the firewall logic first to use chains.

### Solution
This PR rewrites the firewall logic to the chains similar to those proposed in the libfirewall RFC

- [x] README.md is updated


crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,558 @@
...
+        ip: &mut MutableIpv4Packet,
+        protocol: IpNextHeaderProtocol,
+        header_length: usize,
+        total_length: usize,
+    ) {
+        ip.set_next_level_protocol(protocol);
+        ip.set_version(4);
+        ip.set_header_length((header_length / 4) as u8);
+        ip.set_total_length(total_length.try_into().unwrap());
+        ip.set_flags(2);
```
mathiaspeters: It would be nice with a comment here to show what this `2` means in the context

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,558 @@
...
+        ip.set_ttl(64);
+        ip.set_source(Ipv4Addr::LOCALHOST);
+        ip.set_destination(Ipv4Addr::new(8, 8, 8, 8));
```
mathiaspeters: I think nord has it's own DNS server, or other static IP that we could use here instead. So we use our own publicly available infra instead of using, in this case, google

crates/telio-firewall/src/firewall.rs
```diff
@@ -2056,7 +2258,7 @@ pub mod tests {
             ]);
 
             assert_eq!(fw.process_outbound_packet(&make_peer(), &make_udp(src, dst)), false);
-            assert_eq!(fw.process_outbound_packet(&make_peer(), &make_tcp(src, dst, 0)), false);
+            // assert_eq!(fw.process_outbound_packet(&make_peer(), &make_tcp(src, dst, 0)), false);
```
mathiaspeters: This looks like it should be removed

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,558 @@
...
+        udp.set_destination(destination.port());
+        udp.set_length((UDP_HEADER + msg_len) as u16);
+        udp.set_checksum(0);
+        udp.payload_mut().copy_from_slice(msg.as_bytes());
+
+        raw
+    }
+
+    #[test]
+    fn network_filter() {
```
mathiaspeters: Nit: Tests in rust are typically named `test_...`

mathiaspeters:
Once the above comments are fixed it's a +1 for me

crates/telio-firewall/src/chain.rs
stalowyjez: Is this comment enough?

crates/telio-firewall/src/chain.rs
stalowyjez: We set these addresses in the different place anyway, so I just removed that

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+                    LibfwNextLevelProtocol::Udp => proto == Udp,
+                    LibfwNextLevelProtocol::Tcp => proto == Tcp,
+                    LibfwNextLevelProtocol::Icmp => proto == Icmp,
+                    LibfwNextLevelProtocol::Icmp6 => proto == Icmpv6,
+                }
+            }
+            FilterData::TcpFlags(flags) if packet.get_next_level_protocol() == Tcp => {
+                TcpPacket::new(packet.payload())
+                    .map(|pkt| pkt.get_flags() & flags != 0)
+                    .unwrap_or_default()
```
tomasz-grz: Will this return default `TcpPacket` if None? Or a bool false?

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+            }
+        } else {
+            None
+        }
+    };
+}
+
+impl NetworkFilterData {
+    fn is_matching<'a>(&self, packet: &impl IpPacket<'a>, filter_type: NetworkFilterType) -> bool {
+        let packet_addr: crate::firewall::IpAddr = if let NetworkFilterType::Src = filter_type {
```
tomasz-grz: What do you think of using `matches!` here and above in `get_inner_packet_port`?

```suggestion
        let packet_addr: crate::firewall::IpAddr = if matches!(filter_type, NetworkFilterType::Src) {
```

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+        }
+
+        match packet.get_next_level_protocol() {
+            IpNextHeaderProtocols::Udp => get_inner_packet_port!(UdpPacket, packet, filter_type)
+                .map(|port| self.ports.contains(&port))
+                .unwrap_or_default(),
+            IpNextHeaderProtocols::Tcp => get_inner_packet_port!(TcpPacket, packet, filter_type)
+                .map(|port| self.ports.contains(&port))
+                .unwrap_or_default(),
+            _ => true,
```
tomasz-grz, line 102: Should this be `false` if it doesn't match any protocol?

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+                    .map(|pkt| pkt.get_flags() & flags != 0)
+                    .unwrap_or_default()
+            }
+            FilterData::TcpFlags(_) => false,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Filter {
```
tomasz-grz: How about adding some doc comments for this and all the exposed `pub struct` and `pub enum` definitions?

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,7 +343,329 @@ impl StatefullFirewall {
...
+                            ports: peer.port()..=peer.port(),
+                        }),
+                        inverted: false,
+                    },
+                ],
+                action: LibfwVerdict::LibfwVerdictReject,
+            });
+        }
+
+        // Drop packets from UDP blacklist
```
tomasz-grz: TCP here?

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,7 +343,329 @@ impl StatefullFirewall {
...
+
+        // Include packets which are going to local IPv4 networks
+        let mut ipv4_local_area_network_filters = ipv4_local_area_networks.map(|network| {
+            vec![Filter {
+                filter_data: FilterData::DstNetwork(NetworkFilterData {
+                    network: IpNet::V4(network),
+                    ports: 0..=65535,
+                }),
+                inverted: false,
+            }]
```
tomasz-grz: Since we repeatedly create a `Filter` for different networks, what do you think of using a helper function like this?
```rust
fn dst_all_ports_filter(ip: IpAddr) -> Filter {
    Filter {
        filter_data: FilterData::DstNetwork(NetworkFilterData {
            network: IpNet::from(ip),
            ports: 0..=65535,
        }),
        inverted: false,
    }
}
```

tomasz-grz (commented):
Nice work! Some small notes from my side.
Since I am not very familiar with this part of code, only +0.5.

crates/telio-firewall/src/chain.rs
jjanowsk: For protocols without notion of ports it should only check IP addresses, so `true` is correct.

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+    fn is_matching<'a>(
+        &self,
+        conn_state: LibfwConnectionState,
+        packet: &impl IpPacket<'a>,
+        assoc_data: Option<&[u8]>,
+        direction: LibfwDirection,
+    ) -> bool {
+        self.filter_data
+            .is_matching(conn_state, packet, assoc_data, direction)
+            != self.inverted
```
jjanowsk, line 214: Unless I'm missing something, this is probably incorrect.

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+        conn_state: LibfwConnectionState,
+        packet: &impl IpPacket<'a>,
+        assoc_data: Option<&[u8]>,
+        direction: LibfwDirection,
+    ) -> LibfwVerdict {
+        self.rules
+            .iter()
+            .find(|rule| rule.is_matching(conn_state, packet, assoc_data, direction))
+            .map(|rule| rule.action)
+            .unwrap_or_else(|| LibfwVerdict::LibfwVerdictDrop)
```
jjanowsk: I think the chain could have configurable default. That's how iptables works as well. E.g. if you want to just block some outgoing connections you add a chain with one rule dropping this connection but you set default Accept verdict for a chain.

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,556 @@
...
+        ));
+        assert!(dst_filter.is_matching(
+            LibfwConnectionState::LibfwConnectionStateNew,
+            &ip,
+            Some(&pk),
+            LibfwDirection::LibfwDirectionOutbound
+        ));
+
+        // Test UDP with proper source port
+        let pkt = make_udp("168.72.0.12:1234", "168.72.0.12:10000");
```
jjanowsk, line 442: "with matching" in the comment would make more sense IMO. 
When it comes to the test scenario itself `// Test UDP with proper source port` tests the same stuff as `// Test UDP with proper source`, doesn't it? Maybe it's worth checking scenario where the ip is matching but the port is not?

crates/telio-firewall/src/chain.rs
jjanowsk: After a thought, I think XOR would work here not !=.

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,7 +343,329 @@ impl StatefullFirewall {
...
+                action: LibfwVerdict::LibfwVerdictAccept,
+            });
+        }
+
+        // Create rules for peers with LocalAreaConnection permissions
+        let ipv4_local_area_networks = [
+            Ipv4Net::new_assert(StdIpv4Addr::new(10, 0, 0, 0), 8),
+            Ipv4Net::new_assert(StdIpv4Addr::new(172, 16, 0, 0), 12),
+            Ipv4Net::new_assert(StdIpv4Addr::new(192, 168, 0, 0), 16),
+        ];
```
jjanowsk: Why local addresses are now hardcoded? We used to query interfaces to collect local addresses.

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,7 +343,329 @@ impl StatefullFirewall {
...
+                        StdIpv6Addr::new(0xfd74, 0x656c, 0x696f, 0, 0, 0, 0, 0),
+                        64,
+                    )),
+                    ports: 0..=65535,
+                }),
+                inverted: true,
+            },
+        ];
+
+        for ip in LOCAL_ADDRS_CACHE.lock().iter().map(|peer| peer.ip()) {
```
jjanowsk: So I feel like this should be a base for the whole` ipv4_local_area_network_filters` contraption. Also it should be monitored and the rules should be reapplied when it changes

crates/telio-firewall/src/firewall.rs
```diff
@@ -2517,86 +2719,6 @@ pub mod tests {
             }
     }
 
-    #[test]
-    fn test_local_network_range() {
```
jjanowsk, line 2536: Why removed?

crates/telio-firewall/src/chain.rs
stalowyjez: But aren't they the same for boolean values? ^^'

jjanowsk: 🤦 True. But it felt strange :D

crates/telio-firewall/src/firewall.rs
stalowyjez: Because `is_private_address_except_local_interface` function itself which is tested here is removed as well, this logic is now encoded in chains.

crates/telio-firewall/src/firewall.rs
stalowyjez: These networks were hardcoded previously too - we used `ipv4.is_private()` in `is_private_address_except_local_interface` function:
```
    #[rustc_const_stable(feature = "const_ip_50", since = "1.50.0")]
    #[stable(since = "1.7.0", feature = "ip_17")]
    #[must_use]
    #[inline]
    pub const fn is_private(&self) -> bool {
        match self.octets() {
            [10, ..] => true,
            [172, b, ..] if b >= 16 && b <= 31 => true,
            [192, 168, ..] => true,
            _ => false,
        }
    }
 ```

File: crates/telio-firewall/src/firewall.rs (hunk 16 above)
Comment (jjanowsk) [reply crates/telio-firewall/src/firewall.rs:2536]: But I think you should keep the assertions. Just modify the test slightly to test your chain. So you know that the rules in the chain map 1to1 to the old implementation.

File: crates/telio-firewall/src/firewall.rs (hunk 14 above)
Comment (jjanowsk) [reply crates/telio-firewall/src/firewall.rs]: Ok, I incorrectly read the old impl.

File: crates/telio-firewall/src/firewall.rs (hunk 15 above)
Comment (jjanowsk) [reply crates/telio-firewall/src/firewall.rs]: Ignore

Comment (jjanowsk) [reply crates/telio-firewall/src/firewall.rs]: Actually do not ignore :) This part I think is still valid: Also it should be monitored and the rules should be reapplied when it changes

File: crates/telio-firewall/src/firewall.rs (hunk 16 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/firewall.rs:2536]: Hmm, I guess testing part of the chain isn't as easy as a function, especially when this function maps to a parts of multiple rules. I guess I can create a function which returns a few vectors of filters, so we could iterate over them and see whether it matches all of the filters from any of the sets or not, but hmm, it sounds like quite complex thing to do for such obvious results

File: crates/telio-firewall/src/firewall.rs (hunk 15 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/firewall.rs]: Well, that makes sense, soo LOCAL_ADDRS_CACHE should be observable, I guess? Or do you have some better ideas?

File: crates/telio-firewall/src/firewall.rs (hunk 16 above)
Comment (jjanowsk) [reply crates/telio-firewall/src/firewall.rs:2536]: I meant more like testing it via permissions and just checking that those ranges work for incomming/lanconnections, but I just checked the whole file and there is `firewall_test_permissions` which does exactly that :)

File: crates/telio-firewall/src/chain.rs (hunk 6 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: Personally I like `if let .. =` more

File: crates/telio-firewall/src/chain.rs (hunk 8 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: The real API will be C FFI API, which will be added in the next step, and there are some docs there already. Maybe I should change these to `pub(crate)`?

File: crates/telio-firewall/src/firewall.rs (hunk 10 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/firewall.rs]: Well, this wouldn't cover a lot of cases, but maybe network instead of ip and maybe inv argument would do the trick?

File: crates/telio-firewall/src/chain.rs (hunk 5 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: bool false, but it should be more explicit imho

File: crates/telio-firewall/src/chain.rs (hunk 12 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: I'm fine with that idea, but that would require some RFC update

Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: On the other hand adding empty accepting rule in the end of chain is almost as simple

File: crates/telio-firewall/src/chain.rs (hunk 13 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs:442]: This test does that in a way - both src and dst ips are matching ones in the but only source port is matching, so checking it against dst filter kinda tests it. (In the previous examples ips are out of range in the wrong case, while ports are fine)

File: crates/telio-firewall/src/chain.rs (hunk 12 above)
Comment (jjanowsk) [reply crates/telio-firewall/src/chain.rs]: Fair point with the empty rule

File: crates/telio-firewall/src/firewall.rs (hunk 15 above)
Comment (jjanowsk) [reply crates/telio-firewall/src/firewall.rs]: I guess it can be changed into a channel sending the updates whenever monitor detects the change.

Comment (stalowyjez) [reply crates/telio-firewall/src/firewall.rs]: But then we need some thread for updating firewall? Maybe observable is easier in the end

Comment (stalowyjez) [reply crates/telio-firewall/src/firewall.rs]: @jjanowsk Maybe sth like this?

File: crates/telio-firewall/src/chain.rs (hunk 5 above)
Comment (tomasz-grz) [reply crates/telio-firewall/src/chain.rs]: `.unwrap_or(false)`?

File: crates/telio-firewall/src/chain.rs (hunk 8 above)
Comment (tomasz-grz) [reply crates/telio-firewall/src/chain.rs]: If I had to work with this code base, I would still like to have a quick reference when hovering over data types in the editor etc.

For example what is `FilterData` and how it's meant to be used.

Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: I think it's quite uncommon to add docs to private structs and functions, but if you insist ^^'

File: crates/telio-firewall/src/chain.rs (hunk 5 above)
Comment (stalowyjez) [reply crates/telio-firewall/src/chain.rs]: I think I replaced all of them

File: crates/telio-firewall/src/chain.rs [hunk 17]
```diff
@@ -0,0 +1,646 @@
...
+///
+enum NetworkFilterType {
+    Src,
+    Dst,
+}
+
+///
+/// Next level protocol (L4)
+///
+#[allow(dead_code)]
```
Jauler, line 39: is this `#[allow(dead_code)]` needed?

crates/telio-firewall/src/firewall.rs
```diff
@@ -296,6 +296,12 @@ pub struct StatefullFirewall {
     outgoing_udp_blacklist: RwLock<Vec<SocketAddr>>,
```
Jauler, line 296: I wonder, is there any chance to make the `chain` to be our source-of-truth instead of replicating the whitelist/ip_addresses/outgoing_tcp_blacklist/outgoing_udp_blacklist? I am not sure yet whether it would make sense, but wanted to discuss such possibility.

Jauler: Although this can definitely be later improvement, rather than this PR.

crates/telio-firewall/src/firewall.rs
```diff
@@ -353,14 +624,24 @@ impl StatefullFirewall {
...
-                let link =
-                    Conntrack::build_conn_info(&ip, LibfwDirection::LibfwDirectionOutbound)?.0;
-                if blacklist.contains(&SocketAddr::new(link.remote_addr.into(), link.remote_port)) {
+        if !ip.check_valid() {
+            telio_log_trace!("Outbound IP packet is not valid, dropping: {:?}", ip);
+            return Ok(false);
+        }
+
+        let verdict = self.chain.read().process_packet(
+            LibfwConnectionState::LibfwConnectionStateNew,
```
Jauler: :suspect: I am not sure whether passing `libfwConnectionStateNew` is correct here. Or am I missing something?

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,9 +349,273 @@ impl StatefullFirewall {
...
+        }
+
+        rules.push(Rule {
+            filters: vec![Filter {
+                filter_data: FilterData::Direction(LibfwDirection::LibfwDirectionOutbound),
+                inverted: false,
+            }],
+            action: LibfwVerdict::LibfwVerdictAccept,
+        });
+
```
Jauler, line 473: I think we have to think about rules order a bit more carefully :thinking: 

As for example if rule at 408 will match - the VPN rule or peer rules will not be executed.

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,9 +349,273 @@ impl StatefullFirewall {
...
+        // Create rules for peers with LocalAreaConnection permissions
+        let ipv4_local_area_networks = [
+            Ipv4Net::new_assert(StdIpv4Addr::new(10, 0, 0, 0), 8),
+            Ipv4Net::new_assert(StdIpv4Addr::new(172, 16, 0, 0), 12),
+            Ipv4Net::new_assert(StdIpv4Addr::new(192, 168, 0, 0), 16),
+        ];
+
+        // Include packets which are going to local IPv4 networks
+        let mut ipv4_local_area_network_filters = ipv4_local_area_networks
```
Jauler: nit: Maybe it would help a little bit to build local_area_network_filters in separate function. To make the main `recreate_chain()` a bit simpler

crates/telio-firewall/src/firewall.rs
Jauler: Oh. 

But it does in-fact look correct.

crates/telio-firewall/src/chain.rs
stalowyjez: Currently we never contruct Icmp cases

crates/telio-firewall/src/firewall.rs
stalowyjez: I think this is a good idea, but I guess it can be done in parallel with the other firewall tasks, so maybe it could be a separate Jira item? What do you think?

crates/telio-firewall/src/firewall.rs
stalowyjez: Oh, that's a good catch - it works for now, as we don't have any conntrack filters for outbound packets now, but needs to be fixed, for sure

crates/telio-firewall/src/firewall.rs
stalowyjez: Sth like this is fine?

crates/telio-firewall/src/firewall.rs
stalowyjez: fixed - added also some hint why that was done like that before, we had some outbound packets which are resulting currently in errors in conntrack, but the logic of the firewall is currently to let them through (we have event tests for that) - after proptest-related conntrack updates the flow here should be more intuitive, I guess

crates/telio-firewall/src/firewall.rs
Jauler: yes, IMHO separate jira item makes sense

Jauler: Created LLT-6557

crates/telio-firewall/src/firewall.rs
Jauler: yup

crates/telio-firewall/src/firewall.rs
tomasz-grz: Nice!

crates/telio-network-monitors/src/monitor.rs
```diff
@@ -23,6 +26,12 @@ struct PausedState {
     is_path_change_enqueued: bool,
 }
 
+/// Trait which should be implemented to receive locall address change notifications
```
mathiaspeters: Very minor nit: `locall` -> `local`

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,646 @@
...
+    ///
+    /// Checks whether the packet matches the network filter.
+    ///
+    /// @param packet       Assumed packet
+    /// @param filter_type  Which packet endpoint should be checked - source or destination
+    ///
+    /// @return             Boolean value, true if packet matches the filter, false otherwise
+    ///
+    fn is_matching<'a>(&self, packet: &impl IpPacket<'a>, filter_type: NetworkFilterType) -> bool {
+        let packet_addr: crate::firewall::IpAddr = if let NetworkFilterType::Src = filter_type {
```
mathiaspeters: Wouldn't it make more sense to flip this, so that there's only one `.into()` call?
I.e.
```
let std_ip_addr: IpAddr= if let NetworkFilterType::Src = filter_type {
    packet.get_source()
} else {
    packet.get_destination()
};
let packet_addr: crate::firewall::IpAddr = std_ip_addr.into()
```

mathiaspeters: How about using a match statement here and in `get_inner_packet_port!`? IMO, 
```
match filter_type {
    NetworkFilterType::Src => // option1
    NetworkFilterType::Dst => // option2
}
```
is generally better than
```
if let NetworkFilterType::Src = filter_type {
    // option 1
} else {
    // option 2
}
```

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,646 @@
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LibfwVerdict {
+    LibfwVerdictAccept,
+    LibfwVerdictDrop,
+    LibfwVerdictReject,
+}
+
+macro_rules! get_inner_packet_port {
+    ($kind: ty, $packet: expr, $filter_type: expr) => {
+        if let Some(p) = <$kind>::new($packet.payload()) {
```
mathiaspeters: And this could use `map` instead of the outer if-else, but that's really getting into `nit` territory

crates/telio-firewall/src/chain.rs
```diff
@@ -0,0 +1,646 @@
...
+    ) -> bool {
+        self.filter_data
+            .is_matching(conn_state, packet, assoc_data, direction)
+            != self.inverted
+    }
+}
+
+///
+/// A definition of firewall rule it consists of filters which determine whether rule applies to packet
+/// being processed and action, which determins what to do with the rule.
```
mathiaspeters: Minor nit: `determins` -> `determines`

crates/telio-firewall/src/firewall.rs
```diff
@@ -342,25 +349,295 @@ impl StatefullFirewall {
...
+        result.push(ipv6_local_area_network_filters);
+        result
+    }
+
+    fn recreate_chain(&self) {
+        let mut rules = vec![];
+
+        // Drop packets from UDP blacklist
+        let udp_blacklist = self.outgoing_udp_blacklist.read();
+        for peer in udp_blacklist.iter() {
```
mathiaspeters: Let's combine these to reduce the scope of the lock.
I.e. `for peer in self.outgoing_udp_blacklist.read() {}`. And same for the tcp blacklist later in the function

crates/telio-firewall/src/chain.rs
stalowyjez: So the trick is, `IpPacket` is a trait, `get_destination()` returns `IpPacket::Addr` which is either IPv4 or IPv6, so it is not simply `IpAddr` buut, I can add `Into<std::net::IpAddr>` trait to it so I'd get rid of the second conversion? Wdyt?

stalowyjez: Yeah, maybe it looks a bit better

mathiaspeters: Oh, in that case my point was invalid. But I think you made it better now

PR #1326, opened 2025-06-23
Natlab fetch windows setup api logs

gytsto:
### Problem
- running libtelio with wireguard native before tests didn't help to prevent problem

### Solution
- remove wireguard native run before tests
- add setup api logs


tomaszklak (approved):
+2

PR #1332, opened 2025-06-27
LLT-6125: Teliod size optimization

tomasz-grz:
### Problem
Binary size of `teliod` debug build is smaller than the release build.

### Solution
Introduce a custom release build profile that optimizes for size.

- [x] README.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -188,7 +188,6 @@ telio-nat-detect = { version = "0.1.0", path = "./crates/telio-nat-detect" }
 telio-network-monitors = { version = "0.1.0", path = "./crates/telio-network-monitors" }
 telio-nurse = { version = "0.1.0", path = "./crates/telio-nurse" }
 telio-proto = { version = "0.1.0", path = "./crates/telio-proto" }
-telio-batcher = { version = "0.1.0", path = "./crates/telio-batcher" }
 telio-proxy = { version = "0.1.0", path = "./crates/telio-proxy" }
 telio-relay = { version = "0.1.0", path = "./crates/telio-relay" }
 telio-sockets = { version = "0.1.0", path = "./crates/telio-sockets" }
@@ -206,6 +205,13 @@ opt-level = "s"
 lto = true
 codegen-units = 1
 debug = "full"
+
+[profile.release-size-optimized]
+inherits = "release"
+opt-level = "z"      # Optimize for size.
+panic = "abort"      # Abort on panic
+strip = true         # Automatically strip symbols from the binary.
+debug = false        # No debug
 
 [profile.dev]
 debug = 0
--- a/clis/teliod/README.md
+++ b/clis/teliod/README.md
@@ -5,16 +5,33 @@ and embedded environments.
 
 ## Building Teliod
 
-For typical Linux environment it might be built using simply:
+For typical Linux environments, Teliod can be built using the following commands:
 
-```cargo build```
+```sh
+# Debug build from teliod directory
+cargo build
 
-For OpenWRT you might need a bit more complex command, including your router architecture
-and the fact the OpenWRT is MUSLE-based, for example:
+# Release build
+cargo build --package=teliod --release 
+```
 
-```CARGO_TARGET_ARMV7_UNKNOWN_LINUX_MUSLEABIHF_LINKER=rust-lld CC=/path/to/arm-linux-gnueabi-gcc cargo build --package teliod --target armv7-unknown-linux-musleabihf```
+For OpenWRT you might need to cross-compile, depending on the router architecture.
 
-You may need to download some sufficient MUSLE toolchains from `musle.cc`.
+OpenWRT does not ship with GNU libc and instead is MUSL based.
+And further needs to be compiled with MUSL tool-chains or statically linked.
+Which can be obtained from from [musle.cc](https://musl.cc/) or the `musl-dev` package
+
+```sh
+# Add MUSL target
+sudo apt install musl-dev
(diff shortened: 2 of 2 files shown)


clis/teliod/README.md
```diff
@@ -5,16 +5,53 @@ and embedded environments.
...
+```sh
+# Add MUSL target
+sudo apt install musl-dev
+rustup target add --toolchain aarch64-unknown-linux-musl
+
+# Regular release build for aarch64
+cargo build --package=teliod --target=aarch64-unknown-linux-musl --release
+
+# Size optimized release build for aarch64
+cargo build --package=teliod --target=aarch64-unknown-linux-musl --profile=release-size-optimized
```
mathiaspeters: Are the size optimizations only relevant for `aarch64`?

lcruz99:
Have you compared the binary size between the different combinations (cargo's opt-level specifically)?

I've came to realize that the extreme values don't always yield the best results.

Cargo.toml
```diff
@@ -207,6 +206,13 @@ lto = true
 codegen-units = 1
 debug = "full"
 
+[profile.release-size-optimized]
+inherits = "release"
+opt-level = "z"      # Optimize for size.
+panic = "abort"      # Abort on panic
```
lcruz99, line 212: Are you aware of the size impact of this option on the output binary? Is it worth to loose stack unwinds, which means no backtrace display nor stack clean up being done?

Cargo.toml
```diff
@@ -207,6 +206,13 @@ lto = true
...
+strip = true         # Automatically strip symbols from the binary.
+debug = false        # No debug
```
lcruz99, line 214: Debug symbols are needed for debugging purposes and should be initially linked against the binary, they can then be removed/splitted afterwards.

I suggest either stripping them using [split-debuginfo](https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo) or add [post build routines from rust-build-utils scripts](https://github.com/NordSecurity/rust_build_utils/blob/main/rust_build_utils/rust_utils_config.py#L82).

Cargo.toml
```diff
@@ -207,6 +206,13 @@ lto = true
 codegen-units = 1
 debug = "full"
 
+[profile.release-size-optimized]
```
lcruz99, line 209: If this is exclusive to OpenWRT please name the profile as so. eg: profile.release-openwrt

clis/teliod/README.md
lcruz99: This should be handled by the build scripts.

tomasz-grz: Nope, this is just example since my target device is `aarch64`, should I make it a bit more generic?

tomasz-grz: I wanted to update the documentation since it vaguely mentioned how to build it for MUSL target

Cargo.toml
tomasz-grz: Maybe we should use debug builds for debugging purposes

Cargo.toml
tomasz-grz: Panic abort 1 MB and `immediate_abort` flag further 100 KB.

Cargo.toml
tomasz-grz: Initially I called it `release-openwrt`, but size optimization could be relevant for any target platform with limited storage, so I wanted to keep it more generic

clis/teliod/README.md
tomasz-grz: Removed additional info about re-building libstd.

tomasz-grz: Updated to be more generic

tomasz-grz:
@lcruz99 Yes I have, but obviously not every possible combination, otherwise I would still be compiling :D

| opt-level | size |
|--------|--------|
| "z" | 4479264 |
| "s" | 5391176 |
| 3 | 7690552 |

mathiaspeters:
+1

Cargo.toml
lcruz99: I'd say we should prefer to unwind the stack as it's valuable when debugging and not skip any clean ups. Please leave panic setting as default ("unwind")

Cargo.toml
lcruz99: We use debug builds for debugging purposes but in reality release builds are distributed.

If a crash happens in production, having separate debug symbols allows us to get meaningful stack traces without shipping huge binaries. 

I suggest to build in release mode but retain symbols initially, then split them in a post-build step as described in my previous comment. This way, we get optimized builds and retain debuggability, we distribute the stripped binaries together with the debug symbols so that anyone can link debug symbols separately if needed.

Cargo.toml
lcruz99: > size optimization could be relevant for any target platform with limited storage, so I wanted to keep it more generic

Having generic profiles limits individual customization, besides that, it's more informative to which system this profile is targetting. 

Maybe we could have a template profile and then [extend it](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles)? Otherwise a generic profile is as the saying goes "Jack of all trades master of none" 😅

clis/teliod/README.md
lcruz99: But do we describe in the documentation how to build using the build scripts? It should be the recommended path.

Cargo.toml
lcruz99: besides that release profile is already size optimized so having a generic "release-size-optimized" profile becomes redundant

PR #1338, opened 2025-07-01
LLT-6139: Use separate table for iproute VPN routing

mathiaspeters:
Teliod with VPN and the iproute interface configuration option would previously add routing rules to the main routing table. It worked when testing locally but didn't work in natlab. Now the rules are written to a separate routing table, which also aligns with the NordVPN Linux app, and it now works in natlab. As such, the natlab test for teliod+VPN now includes a test for using the iproute config option rather than just the manual option

- [x] README.md is updated


clis/teliod/src/configure_interface.rs
```diff
@@ -170,53 +176,96 @@ impl ConfigureInterface for Ifconfig {
...
         Self {
             adapter_name,
-            default_gateway_ipv4: Self::get_default_gateway("-4"),
-            exit_route_ip: None,
+            vpn_ip_info: None,
             ipv6_support_manager: Ipv6SupportManager::default(),
         }
     }
 
-    #[allow(clippy::expect_used)]
```
mathiaspeters, line 194: This is no longer necessary since the routing setup is done differently

clis/teliod/src/configure_interface.rs
```diff
@@ -259,24 +308,33 @@ impl ConfigureInterface for Iproute {
     }
 
     fn set_exit_routes(&mut self, exit_node: &IpAddr) -> Result<(), TeliodError> {
+        #[cfg(target_os = "linux")]
```
mathiaspeters, line 313: The `iproute` configuration option is only really supported on linux so it's fine if this function is a no-op on other platforms

clis/teliod/src/configure_interface.rs
```diff
@@ -170,53 +176,96 @@ impl ConfigureInterface for Ifconfig {
...
+        Ok(Self {
+            table: Self::find_available_table()?,
+            fw_rule_prio: Self::find_available_rule_priority()?,
+        })
+    }
+
+    // Based on the implementation from the NordVPN Linux app
+    #[allow(clippy::expect_used)]
+    fn find_available_table() -> Result<String, TeliodError> {
+        let table_pattern = Regex::new("table ([0-9]+)").expect("Failed to compile ip table regex");
```
tomasz-grz, line 201: Doesn't `expect()` panic? How does it pass Nagger checks?
 On the other hand, it would only panic if regex is wrong?

clis/teliod/src/configure_interface.rs
```diff
@@ -259,24 +308,33 @@ impl ConfigureInterface for Iproute {
...
                 "dev",
                 &self.adapter_name,
+                "table",
+                &vpn_ip_info.table,
             ]))?;
-            if let Some(gateway) = &self.default_gateway_ipv4 {
-                Self::ignore_file_exists_error(execute(Command::new("ip").args([
-                    "route",
-                    "add",
-                    exit_node.to_string().as_str(),
```
tomasz-grz, line 276: How does the exception for packets destined to `exit_node` would work now? They would get routed differently because they have `LIBTELIO_FWMARK`?

clis/teliod/src/configure_interface.rs
```diff
@@ -259,24 +308,33 @@ impl ConfigureInterface for Iproute {
     }
 
     fn set_exit_routes(&mut self, exit_node: &IpAddr) -> Result<(), TeliodError> {
+        #[cfg(target_os = "linux")]
         if exit_node.is_ipv4() {
+            let vpn_ip_info = VpnIpInfo::new()?;
```
tomasz-grz, line 315: With `VpnIpInfo::new()` I personally expect it to create an empty data structure that has to be filled in before being used, like `String::new()`.

What do you think of calling it something like `generate_table()` or `from_available_table()` instead?

tomasz-grz: Actually never mind :D I think I was just confused when I first read the code.

clis/teliod/src/configure_interface.rs
mathiaspeters: It passes nagger because of the `#[allow(...)]` above. And yes, it will only panic if the regex couldn't be compiled which we would notice when building, so not a risk in this case

clis/teliod/src/configure_interface.rs
mathiaspeters: The way it works now, which is the same as the NordVPN Linux app, is that packets with `LIBTELIO_FWMARK` go to the main routing table and any other packet goes through the new routing table, because of the ip rule

clis/teliod/src/configure_interface.rs
mathiaspeters: I don't mind renaming it to `generate`, but I think something like `generate_table` would be slightly misleading since it contains now just the routing table but also the rule priority

clis/teliod/src/configure_interface.rs
tomasz-grz: Hmm I don't expect this to be a compile-time check but rather run-time. So would it still panic during initialization?

What do you think about adding some basic unit test that calls `find_available_table()`? That way at least we will at least see the failure in CI

mathiaspeters: This will be exercised by the natlab test added in this PR that uses the iproute config option for teliod

nat-lab/tests/test_teliod.py
```diff
@@ -310,6 +317,26 @@ async def test_teliod_vpn_connection_with_manual_interface_setup() -> None:
...
+                    is_connected = True
+            if is_connected:
+                break
+
+        router = LinuxRouter(connection, IPStack.IPv4)
+        if use_manual_interface_setup:
+            router.set_interface_name("teliod")
+            await router.setup_interface(response_data["ip_addresses"])
+            await router.create_meshnet_route()
+            await router.create_vpn_route()
```
lcruz99, line 302: This can be abstracted with a generator function which can also include the clean up on [+349 to +352](https://github.com/NordSecurity/libtelio/pull/1338/files#diff-ad0957b95a003082412bb199243caf3ab119b9af35cd68110a91332239101c73R349-R352).

nat-lab/tests/test_teliod.py
```diff
@@ -310,6 +317,26 @@ async def test_teliod_vpn_connection_with_manual_interface_setup() -> None:
             ).get_stdout()
         )
```
lcruz99: The tests on this module are not concise when waiting for the daemon to start, in some places the DaemonIsRunning error is checked and in others the command IsAlive is. I suggest using only the IsAlive command because the DaemonIsRunning error is issued simply when the socket exists where the IsAlive gets an actual response from the daemon.

This can be moved to wait_for_teliod() fn and use it for every test that needs it.

I know this is not related with this PR but should be of little effort.

nat-lab/tests/test_teliod.py
```diff
@@ -310,6 +317,26 @@ async def test_teliod_vpn_connection_with_manual_interface_setup() -> None:
...
+            status = (
+                await connection.create_process(TELIOD_STATUS_PARAMS).execute()
+            ).get_stdout()
+            status_dict = json.loads(status)
+            is_connected = False
+            for node in status_dict["external_nodes"]:
+                if node["is_vpn"] and node["state"] == "connected":
+                    is_connected = True
+            if is_connected:
+                break
```
lcruz99: We should specify a timeout for this while block in order for monitoring vpn connection times, just like we do for telio mod.

If you wish this can also be a TODO for LLT-6404.

clis/teliod/src/configure_interface.rs
tomasz-grz: Wouldn't it be good to have the GitHub CI fail instead to have quicker feedback? Also running `cargo test` locally would reveal issues

nat-lab/tests/test_teliod.py
mathiaspeters: Discussed separately, this will be part of LLT-6404

nat-lab/tests/test_teliod.py
mathiaspeters: Discussed separately, this will not be done as a common pattern in our tests is to not add timeouts to individual parts like this one, but rather rely on the timeout for the test as a whole

clis/teliod/src/configure_interface.rs
tomasz-grz: Although looks like clippy can actually spot some wrong regex syntax.. so up to you

mathiaspeters: Test added. It only checks for panics, nothing else

nat-lab/tests/test_teliod.py
mathiaspeters: Fixed

lcruz99 (commented):
+1

PR #1340, opened 2025-07-02
Move copying of binaries to after connectivity checks

mathiaspeters:
### Problem
We've seen some failures in natlab where binaries could not be copied to some VMs because the connection failed. 

### Solution
We already have precondition checks for connectivity between VMs and those have retries already set up. By moving the copying of binaries to after those connectivity checks have been done, we reduce the risk of the failed connection happening during the copying of binaries

- [x] README.md is updated


gytsto (approved):
+2, just update PR description please with proper format

PR #1342, opened 2025-07-03
[LLT-6212] Use nightlies cron definition in new gitlab scheduled job

olekoliinyk:
### Problem

### Solution

### DIFF
--- a/.github/workflows/gitlab_schedule.yml
+++ b/.github/workflows/gitlab_schedule.yml
@@ -56,5 +56,5 @@ jobs:
     with:
       schedule: true
       cancel-outdated-pipelines: false
-      triggered-ref: v2.11.4
+      triggered-ref: v2.13.1  # REMEMBER to also update in .gitlab-ci.yml and .github/workflows/gitlab.yml
       github-sha-override: ${{ needs.prepare-sha.outputs.recorded_sha }}
--- a/.github/workflows/gitlab_schedule.yml
+++ b/.github/workflows/gitlab_schedule.yml
@@ -1,11 +1,11 @@
 name: Scheduled GitLab CI Pipeline
 on:
   schedule:
-    # Below cron definition only for testing purposes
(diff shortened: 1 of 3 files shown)


.github/workflows/gitlab_schedule.yml
```diff
@@ -56,5 +56,5 @@ jobs:
     with:
       schedule: true
       cancel-outdated-pipelines: false
-      triggered-ref: v2.11.4
+      triggered-ref: v2.13.1  # REMEMBER to also update in .gitlab-ci.yml and .github/workflows/gitlab.yml
```
gytsto, line 59: Damn, one more place to update version for libtelio-build... Maybe it is possible to use some variable here which would be defined in one place and used all over *.yml files?

gytsto (commented):
+1, but please fix PR description with problem/solution template

.github/workflows/gitlab_schedule.yml
olekoliinyk: As far as I know, github does not support cross-file global env variables. 
Possible solutions:
1. Add it to secrets/env variables. But there won't be ability to track the changes and there might be permission issue to update/access it
2. Merge gitlab.yml and gitlab_schedule.yml into single file and defining it in `env` section. On other hand having it split into 2 files give us clear single responsibility principle

asledzinskiy (approved):

.github/workflows/gitlab_schedule.yml
gytsto: Coud we have `variables.yml` with contents like:
```
variables:
  LIBTELIO_BUILD_VERSION: v2.13.1

```

and then just include them into workflows:
```
include:
  - local: "variables.yml"
```

or it doesn't work in github like it does in gitlab?

olekoliinyk: I'll try to address it in separate PR

PR #1343, opened 2025-07-03
LLT-6138: Add VPN support to UCI

tomasz-grz:
### Problem

OpenWRT platforms are using UCI commands to configure routes and firewall zones.

### Solution

Add UCI implementation for setting up VPN functionality. Disabling IPv6 and preserving LAN access between devices.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/configure_interface.rs
+++ b/clis/teliod/src/configure_interface.rs
@@ -182,25 +182,38 @@ impl ConfigureInterface for Ifconfig {
     }
 }
 
-// When setting routing information on linux we set a default route on a custom routing table
-// and an IP rule to make sure that the correct packets go through that table.
-// This struct holds the necessary information to set that up
-#[derive(Debug, Clone)]
-struct VpnIpInfo {
-    table: u32,
-    fw_rule_prio: u32,
+/// Implementation using `iproute2`
+#[derive(Debug)]
+pub struct Iproute {
+    interface_name: String,
+    table: Option<String>,
+    fw_rule_prio: Option<String>,
+    ipv6_support_manager: Ipv6SupportManager,
 }
 
-impl VpnIpInfo {
-    fn new() -> Result<Self, TeliodError> {
-        let (main_prio, existing_prios) = Self::find_main_and_all_rule_priorities()?;
-        Ok(Self {
-            table: Self::find_available_table()?,
-            fw_rule_prio: Self::find_available_lower_rule_priority(main_prio, &existing_prios)?,
-        })
+impl Iproute {
+    pub fn new(interface_name: String) -> Self {
+        Self {
+            interface_name,
+            table: None,
+            fw_rule_prio: None,
+            ipv6_support_manager: Ipv6SupportManager::default(),
+        }
+    }
+
+    // Some ip commands will return "RNETLINK answers: File exists" which means that the command was already executed and we can ignore the error
+    fn ignore_file_exists_error(res: Result<(), TeliodError>) -> Result<(), TeliodError> {
+        match res {
+            Err(TeliodError::SystemCommandFailed(message)) if message.contains("File exists") => {
+                Ok(())
+            }
+            _ => res,
+        }
     }
 
     // Based on the implementation from the NordVPN Linux app
+    // When setting routing information on linux we set a default route on a custom routing table
+    // and an IP rule to make sure that the correct packets go through that table.
     #[allow(clippy::expect_used)]
     pub fn find_available_table() -> Result<u32, TeliodError> {
         let table_pattern = Regex::new("table ([0-9]+)").expect("Failed to compile ip table regex");
@@ -216,21 +229,19 @@ impl VpnIpInfo {
                 .collect::<Vec<_>>();
         existing.sort_unstable();
         existing.dedup();
-        let mut new_table = DEFAULT_ROUTING_TABLE_ID;
-        loop {
-            if !existing.contains(&new_table) {
-                return Ok(new_table);
-            }
-            new_table += 1;
-            if new_table > 60_000 {
-                return Err(TeliodError::IpRoute);
+
+        // Find first available table ID starting from default
+        for table_id in DEFAULT_ROUTING_TABLE_ID..=60_000 {
+            if !existing.contains(&table_id) {
+                return Ok(table_id);
             }
         }
+        Err(TeliodError::IpRoute)
     }
 
     // Based on the implementation from the NordVPN Linux app
-    // Finds the main rule priority and the list of all priorities
-    pub fn find_main_and_all_rule_priorities() -> Result<(u32, Vec<u32>), TeliodError> {
+    // Finds the main rule priority and the list of assigned priorities
+    fn find_main_and_assigned_rule_priorities() -> Result<(u32, Vec<u32>), TeliodError> {
         let mut main_prio = 0;
         let existing_prios = execute_with_output(Command::new("ip").args(["rule", "list"]))?
             .lines()
@@ -250,46 +261,29 @@ impl VpnIpInfo {
     }
 
     // Iterate over existing_prios until we find the next available lower priority
-    pub fn find_available_lower_rule_priority(
-        start_prio: u32,
-        existing_prios: &[u32],
-    ) -> Result<u32, TeliodError> {
-        let mut fw_prio = start_prio;
-        loop {
-            fw_prio = fw_prio.saturating_sub(1);
-            if fw_prio == 0 {
-                return Err(TeliodError::IpRule);
-            } else if !existing_prios.contains(&fw_prio) {
-                return Ok(fw_prio);
+    fn find_available_lower_rule_priority() -> Result<u32, TeliodError> {
+        let (start_prio, existing_prios) = Self::find_main_and_assigned_rule_priorities()?;
+        for prio in (1..start_prio).rev() {
+            if !existing_prios.contains(&prio) {
+                return Ok(prio);
             }
         }
-    }
-}
-
-/// Implementation using `iproute2`
-#[derive(Debug)]
-pub struct Iproute {
-    interface_name: String,
-    vpn_ip_info: Option<VpnIpInfo>,
-    ipv6_support_manager: Ipv6SupportManager,
-}
-
-impl Iproute {
-    pub fn new(interface_name: String) -> Self {
-        Self {
-            interface_name,
-            vpn_ip_info: None,
-            ipv6_support_manager: Ipv6SupportManager::default(),
-        }
+        Err(TeliodError::IpRule)
     }
 
-    // Some ip commands will return "RNETLINK answers: File exists" which means that the command was already executed and we can ignore the error
-    fn ignore_file_exists_error(res: Result<(), TeliodError>) -> Result<(), TeliodError> {
-        match res {
-            Err(TeliodError::SystemCommandFailed(message)) if message.contains("File exists") => {
-                Ok(())
-            }
-            _ => res,
+    // Iterate over existing_prios until we find the number of available lower priorities
+    pub fn find_available_lower_rule_priorities(count: usize) -> Result<Vec<u32>, TeliodError> {
+        let (start_prio, existing_prios) = Self::find_main_and_assigned_rule_priorities()?;
+        let available_prios: Vec<u32> = (1..start_prio)
+            .rev()
+            .filter(|prio| !existing_prios.contains(prio))
+            .take(count)
+            .collect();
+
+        if available_prios.len() == count {
+            Ok(available_prios)
+        } else {
+            Err(TeliodError::IpRule)
         }
     }
 }
@@ -327,7 +321,9 @@ impl ConfigureInterface for Iproute {
     fn set_exit_routes(&mut self, exit_node: &IpAddr) -> Result<(), TeliodError> {
         #[cfg(target_os = "linux")]
         if exit_node.is_ipv4() {
-            let vpn_ip_info = VpnIpInfo::new()?;
+            let table = Self::find_available_table()?.to_string();
+            let fw_rule_prio = Self::find_available_lower_rule_priority()?.to_string();
+
             execute(Command::new("ip").args([
                 "route",
                 "add",
@@ -335,22 +331,23 @@ impl ConfigureInterface for Iproute {
                 "dev",
                 &self.interface_name,
                 "table",
-                &vpn_ip_info.table.to_string(),
+                &table,
             ]))?;
             execute(Command::new("ip").args([
                 "rule",
                 "add",
                 "priority",
-                &vpn_ip_info.fw_rule_prio.to_string(),
+                &fw_rule_prio,
                 "not",
                 "from",
                 "all",
                 "fwmark",
                 &LIBTELIO_FWMARK.to_string(),
                 "lookup",
-                &vpn_ip_info.table.to_string(),
+                &table,
             ]))?;
-            self.vpn_ip_info = Some(vpn_ip_info);
+            self.table = Some(table);
+            self.fw_rule_prio = Some(fw_rule_prio);
 
             self.ipv6_support_manager.disable(&self.interface_name)?;
             // We have already disabled IPv6 on all interfaces (except the tunnel interface)
@@ -369,19 +366,11 @@ impl ConfigureInterface for Iproute {
     }
 
     fn cleanup_exit_routes(&mut self) -> Result<(), TeliodError> {
-        if let Some(vpn_ip_info) = &self.vpn_ip_info {
-            execute(Command::new("ip").args([
-                "rule",
-                "del",
-                "priority",
-                &vpn_ip_info.fw_rule_prio.to_string(),
-            ]))?;
-            execute(Command::new("ip").args([
-                "route",
-                "flush",
-                "table",
-                &vpn_ip_info.table.to_string(),
-            ]))?;
+        if let Some(fw_rule_prio) = &self.fw_rule_prio {
+            execute(Command::new("ip").args(["rule", "del", "priority", fw_rule_prio]))?;
+        }
+        if let Some(table) = &self.table {
+            execute(Command::new("ip").args(["route", "flush", "table", table]))?;
         }
         self.ipv6_support_manager.reenable()
     }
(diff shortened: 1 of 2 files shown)


clis/teliod/src/configure_interface.rs
```diff
@@ -312,30 +307,35 @@ impl ConfigureInterface for Iproute {
     fn set_exit_routes(&mut self, exit_node: &IpAddr) -> Result<(), TeliodError> {
         #[cfg(target_os = "linux")]
         if exit_node.is_ipv4() {
-            let vpn_ip_info = VpnIpInfo::new()?;
+            let (main_prio, existing_prios) = Self::find_main_and_all_rule_priorities()?;
```
mathiaspeters: I think this should be called from inside `find_available_lower_rule_priority`

clis/teliod/src/configure_interface.rs
```diff
@@ -312,30 +307,35 @@ impl ConfigureInterface for Iproute {
...
+
             execute(Command::new("ip").args([
                 "route",
                 "add",
                 "0.0.0.0/0",
                 "dev",
                 &self.interface_name,
                 "table",
-                &vpn_ip_info.table,
+                &table.to_string(),
```
mathiaspeters: We don't have to do any math with these, so before this refactoring I stringified them when generating them to keep the usages cleaner, and I think I prefer that way

clis/teliod/src/configure_interface.rs
```diff
@@ -359,23 +374,119 @@ impl ConfigureInterface for Iproute {
...
+                            Some(true)
+                        } else {
+                            Some(false)
+                        }
+                    }
+                    Err(_) => Some(false),
+                }
+            }
+            Err(_) => None,
+        };
```
mathiaspeters:
``` 
let wan6_disabled = match execute(Command::new("uci").args(["get", "network.wan6"])).ok().and_then(|_| {
    match execute_with_output(
            Command::new("uci").args(["get", "network.wan6.disabled"]),
        ) {
            Ok(res) if res.contains("1") => Some(true),
            _ => Some(false),
        }
});
```

clis/teliod/src/configure_interface.rs
```diff
@@ -359,23 +374,119 @@ impl ConfigureInterface for Iproute {
...
+
+        let ipv6_enabled = match execute(Command::new("uci").args(["get", "network.wan"])) {
```
mathiaspeters: Same general suggestion here

clis/teliod/src/configure_interface.rs
```diff
@@ -359,23 +374,119 @@ impl ConfigureInterface for Iproute {
...
+
+        Self {
+            interface_name,
+            ipv6_enabled,
+            wan6_disabled,
+        }
+    }
+
+    // Helper to disable IPv6
+    fn disable_ipv6(&self) -> Result<(), TeliodError> {
```
mathiaspeters, line 490: Is the IPv6 and wan6 settings global in OpenWRT? In Linux there's both a global setting an a per-interface setting that needs to be managed separately

clis/teliod/src/configure_interface.rs
tomasz-grz: Yes, it's was leftover from my first approach where I was actually doing math on it..
Or storing `existing_prios` as a `Vec<u32>` instead of `Vec<String>` to be a bit more memory efficient.

clis/teliod/src/configure_interface.rs
tomasz-grz: Hmm the issue for UCI we need to find two tables, but I couldn't find a very ergonomic way to do it..

So it's called:
```rust
let (main_prio, existing_prios) = Iproute::find_main_and_all_rule_priorities()?;
let vpn_rule_prio =
    Iproute::find_available_lower_rule_priority(main_prio, &existing_prios)?;
let lan_rule_prio =
    Iproute::find_available_lower_rule_priority(vpn_rule_prio, &existing_prios)?;
```

Maybe something like
```rust
pub fn find_available_lower_rule_priority(
    start_prio: Option<u32>,
)
```

But then we would run `execute_with_output(Command::new("ip").args(["rule", "list"]))?` multiple times for the same output, which is a bit wasteful...

clis/teliod/src/configure_interface.rs
tomasz-grz: Updated

clis/teliod/src/configure_interface.rs
tomasz-grz: Also this would make it not so unit testable..

But I think I found a nicer approach.
Updated

clis/teliod/src/configure_interface.rs
tomasz-grz: Refactored

clis/teliod/src/configure_interface.rs
tomasz-grz: Good suggestion! Refactored 🚜

clis/teliod/src/configure_interface.rs
tomasz-grz: It is per "interface" (interface in UCI is slightly different from what it means in linux, more like a vlan..) and per "device" (this is actually the linux interface).

clis/teliod/src/configure_interface.rs
```diff
@@ -354,28 +366,97 @@ impl ConfigureInterface for Iproute {
...
 }
 
-/// Implementation using `uci` for OpenWRT
-#[derive(Debug, PartialEq, Eq)]
+/// Implementation using `uci` for OpenWRT devices
+#[derive(Debug)]
 pub struct Uci {
     interface_name: String,
+    ipv6_enabled: Option<bool>,
+    wan6_disabled: Option<bool>,
```
mathiaspeters: I think both of these should either be `enabled` or both be `disabled`

clis/teliod/src/configure_interface.rs
```diff
@@ -354,28 +366,97 @@ impl ConfigureInterface for Iproute {
...
 impl Uci {
     fn new(interface_name: String) -> Self {
-        Self { interface_name }
+        let wan6_disabled = execute(Command::new("uci").args(["get", "network.wan6"]))
+            .ok()
+            .and_then(|_| {
+                execute_with_output(Command::new("uci").args(["get", "network.wan6.disabled"]))
+                    .map(|res| res.contains("1"))
+                    .ok()
+            });
```
mathiaspeters: What do you think about moving the `.ok()` to the end so there's just one such call instead of two?

clis/teliod/src/configure_interface.rs
```diff
@@ -354,28 +366,97 @@ impl ConfigureInterface for Iproute {
...
+    fn restore_ipv6(&self) -> Result<(), TeliodError> {
+        if let Err(e) = execute(Command::new("uci").args(["del", "network.teliod_route6"])) {
+            error!("Error removing route6: {e}");
+        }
+
+        if let Some(disabled) = self.wan6_disabled {
+            if !disabled {
+                if let Err(e) = execute(Command::new("uci").args(["del", "network.wan6.disabled"]))
+                {
+                    error!("Error removing wan6 disabled: {e}");
```
mathiaspeters, line 433: I think this error message could also be slightly improved

clis/teliod/src/configure_interface.rs
mathiaspeters: Looks great

clis/teliod/src/configure_interface.rs
```diff
@@ -178,26 +182,40 @@ impl ConfigureInterface for Ifconfig {
...
+            _ => res,
+        }
     }
 
     // Based on the implementation from the NordVPN Linux app
+    // When setting routing information on linux we set a default route on a custom routing table
+    // and an IP rule to make sure that the correct packets go through that table.
     #[allow(clippy::expect_used)]
-    fn find_available_table() -> Result<String, TeliodError> {
+    pub fn find_available_table() -> Result<u32, TeliodError> {
```
LukasPukenis, line 218: Just curious about this one - if `teliod` supports routing through VPN only now for our products, we should just have two routing rules - one for VPN through real physical interface and another rule which routes everything else to a tunnel interface. I wonder if the table introduction is only used so NatLab would work or does it have real consequences in real OpenWRT VPN case?

clis/teliod/src/configure_interface.rs
LukasPukenis: Could we have an explanation on both options in the comment? In addition I see that those are initially extracted from the system and saved for restoration, maybe we can rename them like `original_ipv6_enabled` or `initial_ipv6_enabled`?

clis/teliod/src/configure_interface.rs
```diff
@@ -354,28 +366,97 @@ impl ConfigureInterface for Iproute {
...
+                }
+            }
+        }
+        if let Some(enabled) = self.ipv6_enabled {
+            if enabled {
+                if let Err(e) = execute(Command::new("uci").args(["del", "network.wan.ipv6"])) {
+                    error!("Error removing wan ipv6: {e}");
+                }
+            }
+        }
```
LukasPukenis: I am curious about the logic here, if you could briefly explain it I would appreciate it :bow: - I see that `self.ipv6_enabled` is previously set to a value of `uci get network.wan.ipv6 == 1` then upon restoration the option is deleted if it was previously enabled instead of(what I would expect - setting it back to `1`)

clis/teliod/src/configure_interface.rs
```diff
@@ -354,28 +366,97 @@ impl ConfigureInterface for Iproute {
...
+            ipv6_enabled,
+            wan6_disabled,
+        }
+    }
+
+    // Helper to disable IPv6
+    fn disable_ipv6(&self) -> Result<(), TeliodError> {
+        if self.wan6_disabled.is_some() {
+            execute(Command::new("uci").args(["set", "network.wan6.disabled=1"]))?;
+        }
```
LukasPukenis: I wonder - if we already call `disable_ipv6()` why do we do any checks on existing state instead of just executing `uci set network.wan6.disabled=1`? In that case the initial values for variables could be stored and only read once at the end when restoring things. Unless there's a reason we cannot do that of course :)

clis/teliod/src/configure_interface.rs
mathiaspeters: The way it works is that libtelio will set an fwmark on every packet it sends to the VPN server (the packets that should *not* be routed through the tunnel) and then the separate table and rule says that packets that have that fwmark should go through the normal default route and all other packets should go through the tunnel. So it's not just a way to get natlab to work, it's copying what the NordVPN Linux app does in production, and it happened to also solve some natlab issues

clis/teliod/src/configure_interface.rs
mathiaspeters: At least my idea when I added similar functionality to the iproute configuration option was that when disconnecting we want to restore the state to what it was before and to do that we need to know what it was before we disabled it

clis/teliod/src/configure_interface.rs
LukasPukenis: @mathiaspeters thanks, your explanation led me to the code responsible for that in Linux:
https://github.com/NordSecurity/libtelio/blob/main/crates/telio-sockets/src/protector/linux.rs#L58 which I am very happy to see and it supports the explanation about fwmarks and then tables to some degree :bow:

clis/teliod/src/configure_interface.rs
LukasPukenis: I got confused a bit since it's `disable_ipv6()` called from `set_exit_routes()` where we intentionally disable ipv6, so to me it looks like we can do so regardless the original state, I suppose.

The restoration part makes perfect sense and it is done in the `restore_ipv6()` here:
https://github.com/NordSecurity/libtelio/pull/1343/files/2e4cd8e37c43cf8168778c9b831e962009d2b205#diff-68aeda816ff88fb2a6be49caa48e7cc61506184261ef9069a65d9bf86b8ca20bR431

which is called from `cleanup_exit_routes()` which makes sense - on cleanup we restore to what was saved. So to me there's just slight confusion why do we need to do any state checking if we intend to disable ipv6 :)

clis/teliod/src/configure_interface.rs
tomasz-grz: That would invert the logic for one of them compared to the config, making it potentially confusing..

`network.wan6.disabled`
`network.wan.ipv6`

clis/teliod/src/configure_interface.rs
tomasz-grz: Also using tables and fwarks simplifies handling traffic from LAN devices, in my first implementation it would break access to the router itself as it would try to route that traffic into the tunnel..

clis/teliod/src/configure_interface.rs
tomasz-grz: This checks whether a WAN6 interface exists, and then disables it regardless of the state.
if there is no WAN6 interface, `wan6_disabled` is `None`, otherwise it has the initial state of `network.wan6.disabled`

clis/teliod/src/configure_interface.rs
tomasz-grz: Hmm if `network.wan.ipv6` entry is not present, it assumes an "Auto" value based on the device settings. Deleting it, sets it back to default basically.

According to documentation the `wan.ipv6` option is relevant only for PPPoE and PPPoA protocols

clis/teliod/src/configure_interface.rs
tomasz-grz: Good point!

tomasz-grz: Renamed to `init_wan6_state` etc

clis/teliod/src/configure_interface.rs
tomasz-grz: Updated

clis/teliod/src/configure_interface.rs
tomasz-grz: Added comments and renamed the values

clis/teliod/src/configure_interface.rs
tomasz-grz: Refactored this to restoring the initial values

clis/teliod/src/configure_interface.rs
```diff
@@ -354,28 +366,138 @@ impl ConfigureInterface for Iproute {
...
 pub struct Uci {
     interface_name: String,
+    /// Initial state of IPv6 on WAN interface, used for PPPoE and PPPoA protocols
+    init_wan_ipv6_state: Option<bool>,
+    /// Initial state of WAN6 inteface, if present
+    init_wan6_state: Option<bool>,
+    /// Initial state of Wireless WAN6 inteface, if present
+    init_wwan6_state: Option<bool>,
+    /// Initial state of Smartphone USB tethering6 inteface, if present
+    init_tethering6_state: Option<bool>,
```
mathiaspeters: I think if we want to call these `*_state`, we need to create an enum to represent the options because the sentence `ipv6_state is now true` is very unclear. It could be something like `wan_ipv6_initially_enabled` instead and still be a bool

clis/teliod/src/configure_interface.rs
mathiaspeters: I should have continued this thread but didn't think of it so started a new thread about the names of these. Therefore closing this thread in favor of the new one

clis/teliod/src/configure_interface.rs
tomasz-grz: Good idea, I added the enum states.

mathiaspeters: I think it still has the same issue. `True` is not a state and it's not clear what it means. Either the name needs to have enough context that `True` or `False` is clear what it means, or the type itself needs to have that information. If we go with an enum, the options could be `Enabled` and `Disabled`, for example

LukasPukenis (commented):
+1

clis/teliod/src/configure_interface.rs
tomasz-grz: Hmm I thought about it, but the types represent a sort of "trilean" logic. `1`, `0`, and `default`. Following boolean conventions I thought it would make sense to call them `True` and `False`...

I personally feel like setting the option `network.wan6.disabled` to `true` is much more understandable semantically than `network.wan6.disabled` to `enabled`...

tomasz-grz: Renamed the variables to provide enough context for the settings they represent, despite that contradicting with previous feedback.

mathiaspeters: I still think the current naming will cause confusion over time. A variable called `*_state` should not hold what is essentially a boolean value because you need to understand the full context to have any idea what it means. 

And I do agree that `network.wan6.disabled=true` makes more sense than `network.wan6.disabled=enabled`. The latter doesn't make any sense at all. But I also think `initial_wan6_state=true` is confusing and requires digging through the rest of the code to understand what it means. What to me makes sense to me is either of:
```
let wan6_initially_enabled = true;
network.wan6.disabled = !wan6_initially_enabled;

/** or **/

enum Wan6State {
    Enabled,
    Disabled,
}
let wan6_initial_state = Some(Wan6State::Enabled);
network.wan6.disabled=match wan6_initial_state {
    Some(Enabled) =>true,
    Some(Disabled)=>false,
    None=>//handle default case
}
```
That's of course a bit pseudo-like and naming is just to showcase the point. But this way there is enough information that I can understand what these means just by looking at the name+type of the field, rather than having to dig through the whole interface handling code to understand it.

I see Lukas already approved and this should not be blocked while I'm away on vacation. If my suggestion makes no sense to you then take this message as a +1 and feel free to merge as-is, but I still think the current name+type is confusing and is likely to cause confusion in the future.

tomasz-grz: Since it's bit mixed up in UCI, We would need to remember which options are `enabled = true` and which are `enabled = !true`
```rust
network.wan6.disabled = 0 // enabled
network.wan.ipv6 = 1 // enabled
```

I honestly think `network.wan6.disabled = !wan6_initially_enabled` is just a logic bug waiting to happen...
Pretty much like the example above, should actually be:
```rust
network.wan6.disabled=match wan6_initial_state {
    Some(Enabled) => false,
    Some(Disabled) => true,
    None => //handle default case
}
```

That's why I named them `wan_ipv6_setting_initially_enabled` and `wan6_interface_initially_disabled` to represent exactly the options they are controlling

Also if the state is only binary then the `wan6_initial_state` should actually be 
`Option<Option<Wan6State>>`, since it's not certain that a `network.wan6` interface exists, and then that the `network.wan6.disabled` option is present.

If `network.wan6.disabled` option is not present, it's enabled by default, and to restore it's default state, we delete it.

That's why `UciBoolOption` has three variants: 
```rust
    True,
    False,
    Default,
```

tomasz-grz: I created a new ticket `LLT-6485` to refactor `Ipv6SupportManager` with UCI support, and handle multiple WAN interfaces. 
Please add your recommendations to that ticket, as I feel it's a much better place for them to be implemented.

Hasan6979: It looks good now. With descriptive comments, the confusion should be minimum and looks clear to me. +1

Hasan6979 (approved):

PR #1344, opened 2025-07-03
LLT-6431: Reduce scope of unsafe

mathiaspeters:
### Problem
The function was marked `unsafe` when only a small number of its lines were unsafe

### Solution
Remove the `unsafe` from the function definition and add it to the affected lines

- [x] README.md is updated


tomaszklak (approved):
+2

PR #1346, opened 2025-07-04
teliod: Fix unresponsive teliod while fetching id

lcruz99:
### Problem
Teliod becomes unresponsive for ~10min if the device id is not successfully fetched from the API, the only alternative is to send a termination signal to the process.

This happens because tokio select! macro only has fetch id and signal listener branches.

### Solution
Add a third branch to select! that handles quit command.


gytsto (dismissed):
+1

clis/teliod/src/command_listener.rs
```diff
@@ -92,6 +92,35 @@ impl CommandListener {
...
+
+        match serde_json::from_str::<ClientCmd>(&command_str) {
+            Ok(ClientCmd::QuitDaemon) => {
+                connection.respond(CommandResponse::Ok.serialize()).await?;
+                Ok(())
+            }
+            Ok(_) => {
+                connection
+                    .respond(
+                        CommandResponse::Err("Obtaining identity, ignoring".to_owned()).serialize(),
```
tomasz-grz, line 107: Why should `is-alive` and `get-status` commands be ignored?

clis/teliod/src/command_listener.rs
```diff
@@ -92,6 +92,35 @@ impl CommandListener {
         }
     }
 
+    pub async fn try_recv_quit(&mut self) -> Result<(), TeliodError> {
```
tomasz-grz, line 95: This looks very similar to already existing `handle_client_connection()`, how could we utilize it to reduce code duplication?

clis/teliod/src/command_listener.rs
lcruz99: at this point daemon has not yet been initialized therefore those commands are not accepted.

clis/teliod/src/command_listener.rs
lcruz99: That was my initial approach but realized that the handle_client_connection() was not written with reusability in mind - it depends on the accepted connection and has a fixed command processing where that could have been SRPed - and to avoid refactoring the whole thing I've decided to just create this fn.

tomasz-grz (approved):
+1

PR #1348, opened 2025-07-07
Move pre-test collection of kernel logs to after the setup checks

mathiaspeters:
### Problem
We saw an issue in natlab where kernel logs could not be collected from mac VM because the connection failed

### Solution
Move collecting kernel logs until after setup checks to hopefully avoid the problem since the setup checks do connection checks with retries


gytsto (approved):
+1

PR #1350, opened 2025-07-08
LLT-6386: Censor feature config options from logs

mathiaspeters:
### Problem
We have a couple of feature config options that are only there for us to use during development, like hide_user_info and hide_thread_id. These feature config options that are not active in production builds can give a bad impression to our users if they are spotted in the logs.

### Solution
Hide those options from the logs

- [x] README.md is updated

### DIFF
--- a/crates/telio-utils/Cargo.toml
+++ b/crates/telio-utils/Cargo.toml
@@ -39,6 +39,7 @@ proptest.workspace = true
 proptest-derive.workspace = true
 rand.workspace = true
 rstest.workspace = true
+sn_fake_clock = { workspace = true }
 tokio = { workspace = true, features = ["time", "rt", "macros", "test-util"] }
 
 telio-test.workspace = true
--- a/crates/telio-utils/src/log_censor.rs
+++ b/crates/telio-utils/src/log_censor.rs
@@ -11,11 +11,15 @@ use std::{
 pub struct LogCensor {
     mask_seed: [u8; 32],
     regex: Regex,
+    hide_data_regex: Regex,
     is_enabled: AtomicBool,
 }
 
 impl Default for LogCensor {
     fn default() -> Self {
+        #[allow(clippy::expect_used)]
+        let hide_data_re = Regex::new(r#"(\\?['"])?hide_(user_data|thread_id)(\\?['"])?(\s)*:(\s)*(true|false)(\s)*,(\s)*"#)
+            .expect("Statically known String for filtering hide_user_data and hide_thread_id is a valid regex");
         #[allow(clippy::expect_used)]
         RegexBuilder::new(
             r"
@@ -61,6 +65,7 @@ impl Default for LogCensor {
         .map(|re| LogCensor {
             mask_seed: rand::thread_rng().gen::<[u8; 32]>(),
             regex: re,
+            hide_data_regex: hide_data_re,
             is_enabled: AtomicBool::new(true),
         })
         .expect("Statically known string for LogCensor is a valid regex")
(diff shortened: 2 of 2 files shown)


crates/telio-utils/Cargo.toml
```diff
@@ -39,6 +39,7 @@ proptest.workspace = true
 proptest-derive.workspace = true
 rand.workspace = true
 rstest.workspace = true
+sn_fake_clock = { workspace = true }
```
mathiaspeters, line 42: The tests in `telio-utils` could previously be run from the root of the project but not from the `telio-utils` directory, after adding this it's now possible to run the tests there as well

crates/telio-utils/src/log_censor.rs
```diff
@@ -99,47 +104,58 @@ impl LogCensor {
...
         }
     }
 }
 #[cfg(test)]
 mod test {
     use super::*;
+    use rstest::*;
 
-    const EXAMPLES: [(&'static str, &'static str); 15] = [
+    const EXAMPLES: [(&str, &str); 15] = [
```
mathiaspeters, line 154: Linter complained about this existing code

crates/telio-utils/src/log_censor.rs
```diff
@@ -99,47 +104,58 @@ impl LogCensor {
...
-                if let Ok(ip) = Ipv6Addr::from_str(m.as_str()) {
-                    if Self::incorret_chars_on_bounds(&log, &m) {
-                        return m.as_str().to_owned();
+        let hide_replaced = self
+            .hide_data_regex
+            .replace_all(&log, |captures: &Captures| {
+                println!("Captures: {captures:?}");
+                String::new()
+            });
+        let ips_replaced = if self.should_censor() {
```
mathiaspeters, line 109: This block is the same, other than the fact that it was moved into the if-statement

crates/telio-utils/src/log_censor.rs
```diff
@@ -99,47 +104,58 @@ impl LogCensor {
...
-            }
-            let name = "IP6";
-            if let Some(m) = captures.name(name) {
-                if let Ok(ip) = Ipv6Addr::from_str(m.as_str()) {
-                    if Self::incorret_chars_on_bounds(&log, &m) {
-                        return m.as_str().to_owned();
+        let hide_replaced = self
+            .hide_data_regex
+            .replace_all(&log, |captures: &Captures| {
+                println!("Captures: {captures:?}");
```
tomaszklak: Please remove `println!`

crates/telio-utils/src/log_censor.rs
```diff
@@ -246,4 +262,20 @@ mod test {
             assert_eq!(original_log_ptr, new_log_ptr);
         }
     }
+
+    #[rstest]
+    #[case(true)]
+    #[case(false)]
+    fn test_hide_user_data_and_hide_thread_id_are_filtered_out_when_log_censor_is_enabled_and_disabled(
```
tomaszklak, line 266: Please add a test where both the IPs and hide_user_data or hide_thread_id are present on the same line.

crates/telio-utils/src/log_censor.rs
mathiaspeters: Thanks, I missed that while cleaning up

crates/telio-utils/src/log_censor.rs
mathiaspeters: I added IPv4, IPv6 and domain name to the same test

tomaszklak (approved):
+1

PR #1353, opened 2025-07-09
Remove nat detect crate

Hasan6979:
### Problem
Telio nat detect crate is unused dead code. Its support isn't there in the server and cannot be used, and it is not going to be supported anytime soon.

### Solution
Remove the crate.

- [x] README.md is updated


crates/telio-proto/protos/nurse.proto
```diff
@@ -28,5 +28,5 @@ message Heartbeat {
         Unknown = 8;
     }
 
-    NatType nat_type = 5;
```
jjanowsk: It must stay 5. It's a protobuf's field number, not a default value.

.unreleased/remove_telio_nat_detect_crate
jjanowsk, line 1: We're changing feature config so it should be documented in the changelog

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -281,10 +279,10 @@ impl Runtime for Analytics {
                 ),
 
             // Derp event, only in the `Monitoring` state
-            Ok(event) = derp_event_future, if self.state == RuntimeState::Monitoring =>
+            Ok(_) = derp_event_future, if self.state == RuntimeState::Monitoring =>
```
jjanowsk: I think this whole branch can be removed. As well as `derp_event_future` and `derp_event_channel`. Moreover nat_type field can be removed from `Analytics` and it should be just hardcoded in `crates/telio-proto/src/packet/relayed/nurse.rs` line 48 when the protobuf message is created.

crates/telio-nurse/src/heartbeat.rs
```diff
@@ -519,21 +517,11 @@ impl Analytics {
...
-            .await
-            .unwrap_or(NatData {
-                public_ip: SocketAddr::from(([0, 0, 0, 0], 80)),
-                nat_type: NatType::Unknown,
-            })
-            .nat_type
-        } else {
-            NatType::Unknown
-        };
+    async fn discover_current_nat_type(&mut self) {
```
jjanowsk: Can be removed now

jjanowsk (changes requested):
@Jauler FYI since this PR is changing the API. You probably track this information for releasing purposes .

crates/telio-proto/src/packet/relayed/nurse.rs
```diff
@@ -38,14 +38,14 @@ impl HeartbeatMessage {
         meshnet_id: Vec<u8>,
         node_fingerprint: String,
         statuses: &[heartbeat::Status],
-        nat_type: heartbeat::NatType,
+        _nat_type: heartbeat::NatType, // To be removed when removed from analytics
```
jjanowsk: Why not now?

stalowyjez: Maybe we can have some small JIRA item for that and a corresponding TODO here?

stalowyjez (approved):
lgtm, +1

crates/telio-proto/src/packet/relayed/nurse.rs
Hasan6979: Removed it completely and sending default values now.

jjanowsk: But `Unknown` is not the default value. Why haven't you left it here hardcoded? I don't know if it matters here or not but IMO the number one rule for the PR like this should be to just leave it in the exact state as previously with the feature disabled. Now you're changing it so who knows if getting `UdpBlocked` (This is the default value) in this place will not cause problems on the receivers side. Please hardcode `Unknown` in this place

Hasan6979: Fixed.

jjanowsk (approved):
+2

PR #1357, opened 2025-07-11
Empty

matislovas:
### Problem

### Solution


gytsto (commented):
+1

.unreleased/unsafe_fix
```diff
@@ -0,0 +1 @@
+Increase `unsafe` scope for whole struct operation, not only the derefence its pointer.
```
mathiaspeters: I'm not sure this concerns the apps so the changelog should be left empty. If it does concern the apps, we should create a ticket so it shows up more nicely in the final released changelog and write this in a way that better communicates the apps what changed from their point of view

matislovas: Yeah good point. Fixed the changelog entry. It concerns windows app only, so I am still for including it.

PR #1362, opened 2025-07-16
LLT-6467: Implement grpc client for the ENS service

tomaszklak:
The changes are thoroughly explained in the RFC. This PR focuses on the core logic, and omits Login/authentication and TLS. One non obvious thing is that I've deleted pinger tests - those broke after I updated protobuf and after spending some time on it without success I've decided to delete them (since it's meshnet related, and the pinger is not actively worked on).


nat-lab/bin/ens
```diff
@@ -0,0 +1,97 @@
...
+from grpc_protobuf.ens_pb2_grpc import LoginServicer, ENSServicer, add_LoginServicer_to_server, add_ENSServicer_to_server
+
+# Synchronization between http controll api and grpc stub
+error_event = asyncio.Event()
+error_ready = None
+
+class LoginServicerStub(LoginServicer):
+    async def GetChallenge(self, request, context):
+        print("GetChallenge", request, context)
+        return Challenge(challenge="foo")
```
matislovas, line 26: Why it is hardcoded? Maybe Generate an actual UUID?

crates/telio-model/src/features.rs
```diff
@@ -61,6 +61,8 @@ pub struct Features {
     pub multicast: bool,
     /// Batching feature configuration, disabled by default, used for batching keep-alives
     pub batching: Option<FeatureBatching>,
+    /// TODO
```
matislovas: What "TODO" here?

src/device.rs
```diff
@@ -2479,7 +2506,29 @@ impl TaskRuntime for Runtime {
 
                 Ok(())
             },
-
+            Some((connection_error,vpn_pk)) = self.event_listeners.error_notification_service_subscriber.recv() => {
+                telio_log_debug!("Received (from vpn {vpn_pk}) connection error notification: {connection_error:?}");
+                if let Some(exit_node) = &self.requested_state.exit_node {
+                    if exit_node.public_key == vpn_pk {
+                        if let Some(node) = self.last_transmitted_event.get(&exit_node.public_key) {
```
matislovas: I am a bit lost with a logic here. If there is no last transmitted event, we're not sending one ..?

And also, maybe check if this is not a duplicate event with `is_dublicated_event()` ?

crates/telio-proto/src/ens.rs
```diff
@@ -0,0 +1,325 @@
...
+            watch::channel(false);
+
+        self.quit = Some(quit_tx);
+
+        let vpn_uri = format!("http://{vpn_ip}:{ens_port}");
+
+        let pool = self.socket_pool.clone();
+        let tx = self.tx.clone();
+
+        tokio::spawn(async move {
```
matislovas: Call me paranoid, but I would save the task's handle, and, when closing, would check if it had joined successfully :smile:

crates/telio-proto/src/ens.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    let mut client = grpc::ens_client::EnsClient::new(channel);
+
+    let connection = client
+        .connection_errors(tonic::Request::<Empty>::new(Empty::default()))
+        .await?;
+    let mut stream = connection.into_inner();
+    loop {
+        select! {
+            _ = quit_rx.wait_for(|b| *b)=> {
```
gytsto: nit: missing space before `=>`

crates/telio-proto/src/ens.rs
```diff
@@ -0,0 +1,325 @@
...
+                telio_log_info!("ENS monitor for '{vpn_uri}' ends");
+                break
+            }
+            error_notification = stream.message() => {
+                telio_log_warn!("Received error notification for '{vpn_uri}': {error_notification:?}");
+                match error_notification {
+                    Ok(Some(error_notification)) =>{
```
gytsto: nit: missing space after `=>`

nat-lab/bin/ens
```diff
@@ -0,0 +1,97 @@
...
+# See: https://github.com/grpc/grpc/issues/29459#issuecomment-1652447028
+protobuf_dir = Path(__file__).parent / "grpc_protobuf"
+sys.path.insert(0, str(protobuf_dir))
+
+import grpc  # type: ignore
+import grpc_protobuf.ens_pb2
+from grpc_protobuf.ens_pb2 import Challenge, ConnectionError
+from grpc_protobuf.ens_pb2_grpc import LoginServicer, ENSServicer, add_LoginServicer_to_server, add_ENSServicer_to_server
+
+# Synchronization between http controll api and grpc stub
```
gytsto: nit: control or controller? :smile:

nat-lab/bin/pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[project]
```
gytsto, line 1: Is this really needed? I'd think `nat-lab/pyproject.toml` is enough, but if not, would be nice to move some of the pylint, py-formatter configs here if not inherited from upper directory `.toml`

nat-lab/tests/protobuf/__init__.py
```diff
@@ -1 +1 @@
-from .pinger_pb2 import Pinger
+# type: ignore
```
gytsto: Maybe get rid of  `nat-lab/tests/protobuf/` entirely if not used anywhere. :thinking:

nat-lab/tests/test_vpn.py
```diff
@@ -775,3 +783,248 @@ async def test_vpn_connection_private_key_change(
...
+                url,
+                data=json.dumps(data).encode("utf-8"),
+                headers={"Content-Type": "application/json"},
+                method="POST",
+            )
+            with urllib.request.urlopen(req) as response:
+                return json.loads(response.read().decode("utf-8"))
+
+        with ThreadPoolExecutor() as executor:
+            result = await loop.run_in_executor(executor, blocking_request)
```
gytsto: Wouldn't this be enough `await asyncio.to_thread()`?

nat-lab/tests/test_vpn.py
```diff
@@ -775,3 +783,248 @@ async def test_vpn_connection_private_key_change(
...
+        client_conn, *_ = [conn.connection for conn in env.connections]
+        client_alpha, *_ = env.clients
+
+        if alpha_setup_params.connection_tag == ConnectionTag.VM_MAC:
+            await ensure_interface_router_property_expectations(client_conn)
+
+        ip = await stun.get(client_conn, config.STUN_SERVER)
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        vpn_connection, *_ = await setup_connections(exit_stack, [vpn_conf.conn_tag])
```
gytsto: I don't think vpn server pinging it's peer is needed in this case. Would speed up test abit. :thinking:

nat-lab/tests/test_vpn.py
```diff
@@ -775,3 +783,248 @@ async def test_vpn_connection_private_key_change(
...
+        await exit_stack.enter_async_context(
+            client_alpha.get_router().block_tcp_port(993)
+        )
+
+        await _connect_vpn(
+            client_conn,
+            vpn_connection.connection,
```
gytsto, line 1019: Same here as before, no need to pass vpn_connection here

crates/telio-proto/Cargo.toml
```diff
@@ -17,7 +17,25 @@ thiserror.workspace = true
...
+async-trait.workspace = true
+tokio-stream = "0.1.17"
+tokio = { workspace = true, features = ["full"] }
+telio-task.workspace = true
+http = "1.3.1"
+hyper-util = "0.1.15"
+tower-service = "0.3.3"
+hyper = { version = "1.6.0", features = ["full"] }
+tower = "0.5.2"
+anyhow.workspace = true
```
jjanowsk: Nit: so far we had dependencies ordered in 3 groups: external per crate, external from workspace, internal. Do we want to keep it?

crates/telio-proto/src/ens.rs
```diff
@@ -0,0 +1,325 @@
...
+const ENS_PORT: u16 = 993;
+
+/// ENS errors
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// Malformed VPN uri
+    #[error("Failed to parse the vpn server uri: {0}")]
+    MalformedVpnUri(#[from] http::Error),
+    /// Inner grpc/tonic error
+    #[error("ENS transpert error: {0}")]
```
jjanowsk: transport

crates/telio-proto/src/ens.rs
```diff
@@ -0,0 +1,325 @@
...
+
+            let collected_errors = collect_errors(2, &mut rx).await;
+
+            assert_eq!(errors_to_emit, collected_errors);
+
+            ens.stop_old_monitor();
+        }
+        // Stopping, should eventually be processed by the running task, and it should
+        // drop last remaining instance of the Tx side.
+        while !rx.is_closed() {}
```
jjanowsk: Maybe worth adding `yield_now` in case `is_closed` doesn't yield on it's own (I don't know). Otherwise it can block in case of single-thread executor.

nat-lab/bin/ens
tomaszklak: Since this endpoint is not yet used in this PR, but it will be in the next one, I've decided that a hardcoded value is good enough for now.

nat-lab/bin/pyproject.toml
tomaszklak: This is only added so that it's possible to install `grpcio`, `protobuf` and `aiohttp` in the Dockerfile used by the nat-lab, without it pre-installing those will fail the natlab.py (re)start command:

```
 => ERROR [base 5/7] RUN sh -c ". ~/.bashrc && uv add --directory /opt/bin grpcio==1.73.1 protobuf==6.30.2 aiohttp==3.12.14"                                                                                                     0.1s
------
 > [base 5/7] RUN sh -c ". ~/.bashrc && uv add --directory /opt/bin grpcio==1.73.1 protobuf==6.30.2 aiohttp==3.12.14":
0.132 error: No `pyproject.toml` found in current directory or any parent directory
```

nat-lab/tests/test_vpn.py
tomaszklak: I would keep this one, since here I want to make sure that everything works as before, even if the ENS connection is not working.

crates/telio-proto/src/ens.rs
tomaszklak: Done

src/device.rs
tomaszklak: You are right, we should emit a new event when there was none previously.

PR #1363, opened 2025-07-16
[LLT-6404] natlab: Abstract teliod to its own module

lcruz99:
### Problem:
Teliod natlab tests contains significant boilerplate and duplicated logic for starting, stopping, and interacting with the teliod daemon.

### Solution:
This PR refactors the teliod test suite by introducing a dedicated abstraction (Teliod, Config, and supporting classes) which encapsulates all process management, configuration, daemon lifecycle, log handling, and socket logic.

It centralizes teliod test logic, reducing boilerplate, improving maintainability and preparing the suite for easier future extension.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/teliod.py
+++ b/nat-lab/tests/teliod.py
@@ -8,10 +8,12 @@ from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import AsyncIterator
+from typing import AsyncIterator, List
 from utils.connection import Connection
 from utils.logger import log
 from utils.process import Process, ProcessExecError
+from utils.router import IPStack
+from utils.router.linux_router import LinuxRouter
 
 
 class TeliodObtainingIdentity(Exception):
@@ -267,7 +269,7 @@ class Teliod:
                     return
             except TimeoutError:
                 pass
-            await asyncio.sleep(0.5)
+            await asyncio.sleep(self.SOCKET_CHECK_INTERVAL_S)
         raise TimeoutError("teliod did not start within timeout")
 
     async def wait_for_vpn_connected_state(self):
@@ -277,3 +279,28 @@ class Teliod:
                 if ext_node["is_vpn"] and ext_node["state"] == "connected":
                     return
             await asyncio.sleep(self.TELIOD_CMD_CHECK_INTERVAL_S)
+
+    @asynccontextmanager
+    async def setup_interface(
+        self, ip_addresses: List[str], vpn_routes: bool
+    ) -> AsyncIterator:
+        """
+        Setups interface addresses and routes manually.
+
+        This function should only be used when interface config provider
+        is set to 'manual' on the teliod config.
+        This is not checked by this function. (TODO: LLT-6476)
+        """
+        router = LinuxRouter(self._connection, IPStack.IPv4)
+        try:
+            router.set_interface_name("teliod")
+            await router.setup_interface(ip_addresses)
+            await router.create_meshnet_route()
+            if vpn_routes:
+                await router.create_vpn_route()
+            yield
+        finally:
+            if vpn_routes:
+                await router.delete_vpn_route()
+            await router.delete_exit_node_route()
+            await router.delete_interface()
--- a/nat-lab/tests/teliod.py
+++ b/nat-lab/tests/teliod.py
@@ -1,0 +1,279 @@
+import asyncio
+import json
+import os
+import re
+import time
+from config import LIBTELIO_BINARY_PATH_DOCKER
+from contextlib import AsyncExitStack, asynccontextmanager
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
+from typing import AsyncIterator
+from utils.connection import Connection
+from utils.logger import log
+from utils.process import Process, ProcessExecError
+
+
+class TeliodObtainingIdentity(Exception):
+    pass
+
+
+class IfcConfigType(Enum):
+    DEFAULT = "config.json"
+    VPN_MANUAL = "config_with_vpn_manual_setup.json"
+    VPN_IPROUTE = "config_with_vpn_iproute_setup.json"
+
+
+@dataclass(frozen=True)
+class Paths:
+    exec_path: Path = Path(f"{LIBTELIO_BINARY_PATH_DOCKER}/teliod")
+    config_dir: Path = Path("/etc/teliod")
+    log_dir: Path = Path("/var/log")
+    run_dir: Path = Path("/run")
+
+    def __post_init__(self):
+        if os.environ.get("PYTEST_CURRENT_TEST"):
+            if not self.exec_path.exists():
+                raise FileNotFoundError(
+                    f"Teliod executable not found: {self.exec_path}"
+                )
+
+    @property
+    def socket_file(self) -> Path:
+        return self.run_dir / "teliod.sock"
+
+    @property
+    def daemon_log(self) -> Path:
+        return self.log_dir / "teliod.log"
+
+    @property
+    def lib_log(self) -> Path:
+        return self.log_dir / "teliod_natlab.log"
+
+    def config_path(self, config_type: IfcConfigType = IfcConfigType.DEFAULT) -> Path:
+        return self.config_dir / config_type.value
+
+
+class Command(list):
+    def __str__(self) -> str:
+        return " ".join(str(item) for item in self)
+
+    def __repr__(self) -> str:
+        return f"Command({super().__repr__()})"
+
+    @classmethod
+    def start(cls, config: "Config") -> "Command":
+        cmd = [str(Paths.exec_path), "start"]
+        if config.no_detach:
+            cmd.append("--no-detach")
+        cmd.append(str(config.path()))
+        return cls(cmd)
+
+    @classmethod
+    def is_alive(cls) -> "Command":
+        return cls([str(Paths.exec_path), "is-alive"])
+
+    @classmethod
+    def get_status(cls) -> "Command":
+        return cls([str(Paths.exec_path), "get-status"])
+
+    @classmethod
+    def quit_daemon(cls) -> "Command":
+        return cls([str(Paths.exec_path), "quit-daemon"])
+
+
+class Config:
+    def __init__(
+        self,
+        config_type: IfcConfigType = IfcConfigType.DEFAULT,
+        no_detach: bool = False,
+        paths=Paths(),
+    ):
+        self.paths: Paths = paths
+        self.config_type: IfcConfigType = config_type
+        self.no_detach: bool = no_detach
+
+    def path(self) -> Path:
+        return self.paths.config_path(self.config_type)
+
+    async def assert_match_daemon_start(self, stdout: str):
+        assert (
+            "Starting daemon" in stdout
+        ), f"Could not find 'Starting daemon' in: '{stdout}'"
+
+        config_match = re.search(r"Reading config from:\s*(.+.json)", stdout)
+        assert config_match, f"Could not find config path in: {stdout}"
+        config_path = Path(config_match.group(1).strip())
+        assert (
+            config_path == self.path()
+        ), f"Config path does not match: '{config_path}' != '{self.path()}'"
+
+        log_match = re.search(r"Saving logs to:\s*(.+\.log)", stdout)
+        assert log_match, f"Could not find log path in: {stdout}"
+        log_path = Path(log_match.group(1).strip())
+        assert (
+            log_path == self.paths.lib_log
+        ), f"Log path does not match: '{log_path}' != '{self.paths.lib_log}'"
+
+        return config_path, log_path
+
+
+class Teliod:
+    START_TIMEOUT_S = 3
+    SOCKET_CHECK_INTERVAL_S = 0.5
+    TELIOD_CMD_CHECK_INTERVAL_S = 1
+
+    def __init__(
+        self,
+        connection: Connection,
+        exit_stack: AsyncExitStack,
+        config: Config = Config(),
+    ) -> None:
+        self._connection: Connection = connection
+        self._exit_stack: AsyncExitStack = exit_stack
+        self.config: Config = config
+
+    async def execute_command(
+        self,
+        cmd: Command,
+    ) -> tuple[str, str]:
+        try:
+            proc = await self._connection.create_process(cmd).execute()
+            stdout, stderr = proc.get_stdout(), proc.get_stderr()
+            log.debug("'%s' stdout: '%s', stderr: '%s'", cmd, stdout, stderr)
+            return stdout, stderr
+        except ProcessExecError as exc:
+            log.debug("Exception occured while executing teliod command: %s", exc)
+            raise
+
+    async def run_command(
+        self,
+        cmd: Command,
+    ) -> Process:
+        proc = await self._exit_stack.enter_async_context(
+            self._connection.create_process(cmd).run()
+        )
+        return proc
+
+    @asynccontextmanager
+    async def start(self) -> AsyncIterator["Teliod"]:
+        try:
+            await self.remove_logs()
+
+            async def wait_for_teliod_start():
+                await self.wait_for_teliod_socket()
+                while True:
+                    try:
+                        if not await self.is_alive():
+                            raise RuntimeError(
+                                "socket exists but daemon's not running."
+                            )
+                        break
+                    except TeliodObtainingIdentity:
+                        await asyncio.sleep(self.TELIOD_CMD_CHECK_INTERVAL_S)
+                        continue
+
+            if not self.config.no_detach:
+                stdout, stderr = await self.execute_command(Command.start(self.config))
+                await wait_for_teliod_start()
+            else:
+                proc = await self.run_command(Command.start(self.config))
+                await wait_for_teliod_start()
+                stdout, stderr = proc.get_stdout(), proc.get_stderr()
+
+            assert len(stderr) == 0, f"Stderr is not empty: {stderr}"
+            await self.config.assert_match_daemon_start(stdout)
+            yield self
+        finally:
+            try:
+                await self.quit()
+            except ProcessExecError as exc:
+                if "Error: DaemonIsNotRunning" not in exc.stderr:
+                    log.error(exc)
+                    await self.kill()
+                    await self.remove_socket()
+                else:
+                    log.info("Tried to quit but daemon is already not running")
+
+    async def is_alive(self) -> bool:
+        try:
+            stdout, _ = await self.execute_command(Command.is_alive())
+            return "Command executed successfully" in stdout
+        except ProcessExecError as exc:
+            if "Obtaining identity, ignoring" in exc.stdout:
+                raise TeliodObtainingIdentity() from exc
+            if "Error: DaemonIsNotRunning" in exc.stderr:
+                return False
+            raise exc
+
+    async def get_status(self) -> str:
+        status, _ = await self.execute_command(Command.get_status())
+        return status
+
+    async def quit(self) -> None:
+        stdout, stderr = await self.execute_command(Command.quit_daemon())
+        assert (
+            "Command executed successfully" in stdout
+        ), f"Failed to execute quit-daemon command: {stderr}"
+
+        assert (
+            not await self.is_alive()
+        ), "Quit command was sent successfully but daemon's still running"
+        assert (
+            not await self.socket_exists()
+        ), "Daemon's not running but socket still exists"
+
+    async def kill(self) -> None:
+        try:
+            await self._connection.create_process(
+                ["killall", "-w", "-s", "SIGTERM", "teliod"]
+            ).execute()
+            assert (
+                not await self.is_alive()
+            ), "SIGTERM was sent but daemon's still running"
+        except ProcessExecError as exc:
+            if "teliod: no process found" not in exc.stderr:
+                raise
+
+    async def remove_logs(self) -> None:
+        for path in [self.config.paths.daemon_log, self.config.paths.lib_log]:
+            await self._connection.create_process(["rm", "-f", str(path)]).execute()
+            await self._connection.create_process(
+                ["test", "!", "-f", str(path)]
+            ).execute()
+
+    async def socket_exists(self) -> bool:
+        try:
+            await self._connection.create_process(
+                ["test", "-e", str(self.config.paths.socket_file)]
+            ).execute()
+            return True
+        except ProcessExecError as exc:
+            assert (exc.returncode, exc.stdout, exc.stderr) == (1, "", "")
+            return False
+
+    async def remove_socket(self):
+        await self._connection.create_process(
+            ["rm", "-f", str(self.config.paths.socket_file)]
+        ).execute()
+
+    async def wait_for_teliod_socket(self):
+        start_time = time.monotonic()
+        while time.monotonic() - start_time < self.START_TIMEOUT_S:
+            try:
+                if await asyncio.wait_for(
+                    self.socket_exists(), self.SOCKET_CHECK_INTERVAL_S
+                ):
+                    return
+            except TimeoutError:
+                pass
+            await asyncio.sleep(0.5)
+        raise TimeoutError("teliod did not start within timeout")
+
+    async def wait_for_vpn_connected_state(self):
+        while True:
+            status = json.loads(await self.get_status())
+            for ext_node in status["external_nodes"]:
+                if ext_node["is_vpn"] and ext_node["state"] == "connected":
+                    return
+            await asyncio.sleep(self.TELIOD_CMD_CHECK_INTERVAL_S)
--- a/nat-lab/tests/test_teliod.py
+++ b/nat-lab/tests/test_teliod.py
@@ -22,7 +22,6 @@ from utils.logger import log
 from utils.ping import ping
 from utils.process.process import ProcessExecError
 from utils.router import IPStack
-from utils.router.linux_router import LinuxRouter
 
 if platform.machine() != "x86_64":
     import pure_wg as Key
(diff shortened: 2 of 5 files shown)


nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,312 @@
...
+class Config:
+    def __init__(
+        self,
+        config_type: ConfigType = ConfigType.DEFAULT,
+        detach: bool = True,
+        paths=Paths(),
+    ):
+        self.paths: Paths = paths
+        self.config_type: ConfigType = config_type
+        self.detach_mode: bool = detach
```
mathiaspeters: I find this name confusing. `detach_mode` seems like it should hold an enum rather than a bool, but I think a bool is good to represent the _intent_. Let's rename it to `detach`, or maybe better, invert it and call it `no_detach` to match the CLI option

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def path(self) -> Path:
+        return self.paths.config_path(self.config_type)
+
+    async def assert_match_daemon_start(self, stdout: str):
+        try:
+            assert (
+                "Starting daemon" in stdout
+            ), f"Expected daemon to print to the stdout 'Starting daemon', got instead: '{stdout}'"
```
mathiaspeters: This string seems overly verbose. Let's align with the other logs `Could not find 'Starting daemon' in: '{stdout}'`

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,312 @@
...
+            ), f"Log path does not match: '{log_path}' != '{self.paths.lib_log}'"
+
+        except AssertionError as e:
+            log.error("Expected output from the daemon:")
+            log.error("------")
+            log.error("Reading config from: %s", self.path())
+            log.error("Saving logs to: %s", self.paths.lib_log)
+            log.error("Starting daemon")
+            log.error("------")
+            log.error("Got instead: %s", stdout)
```
mathiaspeters: I'm not sure these are needed when all the assertions have proper error messages. These logs kinda becomes noise since these print all the expected output, not just what failed, and what failed is already printed

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,312 @@
...
+        self.config: Config = config
+
+    async def execute_command(
+        self,
+        cmd: Command,
+    ) -> tuple[str, str]:
+        try:
+            proc = await self._connection.create_process(cmd).execute()
+            stdout, stderr = proc.get_stdout(), proc.get_stderr()
+            log.debug("'%s' stdout: '%s', stderr: '%s'", cmd, stdout, stderr)
```
mathiaspeters, line 145: Let's use f-strings instead of this kind of formatting. There are more places in the file I think we should do the same so it's not just this line

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,312 @@
...
+    @asynccontextmanager
+    async def start(self) -> AsyncIterator["Teliod"]:
+        try:
+            await self.remove_logs()
+
+            async def wait_for_teliod_start():
+                await self.wait_for_teliod_socket()
+                while True:
+                    try:
+                        assert (
```
mathiaspeters: We have seen an issue in CI where some teliod tests fail because `is_alive` is executed before it actually becomes alive. To solve it we would need to loop for a bit to give teliod time to become alive. This seems like it would also just exit if teliod is not immediately alive since a failed assert would just break the loop

nat-lab/tests/test_teliod.py
```diff
@@ -16,155 +13,45 @@
...
-    with pytest.raises(ProcessExecError) as err:
-        await connection.create_process(TELIOD_STATUS_PARAMS).execute()
-    assert err.value.stderr == "Error: DaemonIsNotRunning"
-
 
 @pytest.mark.parametrize(
-    "start_daemon_params",
-    [(TELIOD_START_PARAMS), (TELIOD_START_NODETACH_PARAMS)],
-    ids=["daemonized_mode", "no_detach_mode"],
+    "daemonized_mode",
```
mathiaspeters: This should also be called `detach` or `no_detach`, and I would prefer `no_detach`. Just keep in mind the actual params need to be flipped

nat-lab/tests/test_teliod.py
```diff
@@ -16,155 +13,45 @@
...
-        assert (
-            "Command executed successfully"
-            in (
-                await connection.create_process(TELIOD_QUIT_DAEMON_PARAMS).execute()
-            ).get_stdout()
-        )
-
-        await wait_for_teliod_stop(connection)
+        with pytest.raises(ProcessExecError) as err:
+            await teliod.quit()
```
mathiaspeters, line 53: I wonder if this is necessary since the new teliod test abstraction will kill the teliod instance if there's an error

nat-lab/tests/test_teliod.py
```diff
@@ -173,31 +60,15 @@ async def test_teliod_logs() -> None:
...
-            ["rm", "-f", STDOUT_FILE_PATH, LOG_FILE_PATH]
-        ).execute()
-
-        # Make sure they are indeed deleted
-        for path in [STDOUT_FILE_PATH, LOG_FILE_PATH]:
-            await connection.create_process(["test", "!", "-f", path]).execute()
-
-        await start_teliod(exit_stack, connection, TELIOD_START_PARAMS)
+        teliod = Teliod(connection, exit_stack)
+        await teliod.remove_logs()
```
mathiaspeters: `remove_logs` is called as the first thing in `start` so this line is unnecessary

nat-lab/tests/test_teliod.py
```diff
@@ -208,23 +79,12 @@ async def test_teliod_logs() -> None:
...
-        CONFIG_FILE_PATH_WITH_VPN_MANUAL_SETUP
-        if use_manual_interface_setup
-        else CONFIG_FILE_PATH_WITH_VPN_IPROUTE_SETUP
-    )
-    async with AsyncExitStack() as exit_stack:
-        connection = (
-            await setup_connections(exit_stack, [ConnectionTag.DOCKER_CONE_CLIENT_1])
-        )[0].connection
-
+async def test_teliod_vpn_connection(config_type: ConfigType) -> None:
```
mathiaspeters: I think `IfcConfigType` or something similar might be a clearer name since it's actually about interface configuration. Currently it might get confused with the `Config` class

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,312 @@
...
+from typing import AsyncIterator, List
+from utils.connection import Connection
+from utils.logger import log
+from utils.process import Process, ProcessExecError
+from utils.router import IPStack
+from utils.router.linux_router import LinuxRouter
+
+
+class ConfigType(Enum):
+    DEFAULT = "config.json"
```
tomasz-grz, line 24: Nice!

nat-lab/tests/teliod.py
tomasz-grz: 💯 yes, that's the issue reported in LLT-6461

nat-lab/tests/test_teliod.py
```diff
@@ -16,155 +13,45 @@
...
-            "Command executed successfully"
-            in (
-                await connection.create_process(TELIOD_QUIT_DAEMON_PARAMS).execute()
-            ).get_stdout()
-        )
-
-        await wait_for_teliod_stop(connection)
+        with pytest.raises(ProcessExecError) as err:
+            await teliod.quit()
+        assert err.value.stderr == "Error: DaemonIsNotRunning"
```
tomasz-grz, line 54: I think the original test was first expecting an error, and the second time to succeed.

Try to quit deamon that is not running
Start daemon
Make sure it quits without error

here looks like we expect errors in both tries

nat-lab/tests/teliod.py
sfraczek: The pylint complains `W1203: Use lazy % formatting in logging functions (line 146)`

nat-lab/tests/teliod.py
sfraczek: On https://github.com/NordSecurity/libtelio/pull/1346 it was added handling of that scenario and on the current PR we catch that exception and iterate again the while after sleeping here (https://github.com/NordSecurity/libtelio/pull/1363/files#diff-9110616e8b95ca1b753835304a509716b8c05a567de0fc7b85c6afc99fac2148R179)

nat-lab/tests/test_teliod.py
sfraczek: It is not necessary. It is just testing that part of teliod abstraction where the ending of the process is handled regardless if it's quit or killed. We're just making sure here that there's no traces of the process running. This is done only on this test.

nat-lab/tests/test_teliod.py
sfraczek: here's similar to the original test, the difference is that we add a third (redundant) quit attempt to ensure that there's no traces of the app after quitting to test the teliod abstraction itself:

"Try to quit deamon that is not running
Start daemon
Make sure it quits without error
-> make sure the teliod abstraction quit successfully even if haven't raised any exception"

sfraczek:
I removed #1368

nat-lab/tests/test_teliod.py
tomasz-grz: But we are asserting twice on `assert err.value.stderr == "Error: DaemonIsNotRunning"` and not on `"Command executed successfully"` the second time

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
+import asyncio
```
tomasz-grz, line 1: Since it's more of a helper, should this be in `nat-lab/tests/utils/teliod.py`?

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
...
+        return proc
+
+    @asynccontextmanager
+    async def start(self) -> AsyncIterator["Teliod"]:
+        try:
+            await self.remove_logs()
+
+            async def wait_for_teliod_start():
+                await self.wait_for_teliod_socket()
+                while True:
```
tomasz-grz, line 167: What do you think adding a timeout? like in `wait_for_teliod_socket()`?
```py
while time.monotonic() - start_time < self.START_TIMEOUT_S:
```

or having an attempt `attempt_count`? so that it doesn't hang here for ever

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
...
+                    self.socket_exists(), self.SOCKET_CHECK_INTERVAL_S
+                ):
+                    return
+            except TimeoutError:
+                pass
+            await asyncio.sleep(self.SOCKET_CHECK_INTERVAL_S)
+        raise TimeoutError("teliod did not start within timeout")
+
+    async def wait_for_vpn_connected_state(self):
+        while True:
```
tomasz-grz, line 276: Same regarding timeout or attempt counter

nat-lab/tests/teliod.py
tomasz-grz: 1/3 
I had to re-read it a couple of times to understand it.. which makes me think if it can be written a little bit more clear.

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
...
+                    log.info("Tried to quit but daemon is already not running")
+
+    async def is_alive(self) -> bool:
+        try:
+            stdout, _ = await self.execute_command(Command.is_alive())
+            return "Command executed successfully" in stdout
+        except ProcessExecError as exc:
+            if "Error: DaemonIsNotRunning" not in exc.stderr:
+                raise exc
+            return False
```
tomasz-grz: 2/3
I feel like this bit has some reversed logic, what do you think if we re-write it something like this?

With a custom exception for obtaining identity, to be explicit that we are accepting only that state?
```python
class TeliodObtainingIdentity(Exception):
    pass
```

```python
    try:
        stdout, _ = await self.execute_command(Command.is_alive())
        return "Command executed successfully" in stdout
    except ProcessExecError as exc:
        # explicit exception for when obtaining identity 
        if "Obtaining identity, ignoring" in exc.stdout:
            raise TeliodObtainingIdentity()
        # re-raise any other error
        raise
```

nat-lab/tests/test_teliod.py
sfraczek: This middle block checks if it succeeded asserting that is_alive is true:
https://github.com/NordSecurity/libtelio/pull/1363/files#diff-ad0957b95a003082412bb199243caf3ab119b9af35cd68110a91332239101c73R49-R50
And afterward upon end of the second block the finally teardown is called here:
https://github.com/NordSecurity/libtelio/pull/1363/files#diff-9110616e8b95ca1b753835304a509716b8c05a567de0fc7b85c6afc99fac2148R190-R199
And afterwards in third block we're just making sure here that there's no traces of the process running  - that the error is the same as if it was never started.

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
...
+            await self.remove_logs()
+
+            async def wait_for_teliod_start():
+                await self.wait_for_teliod_socket()
+                while True:
+                    try:
+                        assert (
+                            await self.is_alive()
+                        ), "socket exists but daemon's not running."
+                    except ProcessExecError as exc:
```
tomasz-grz: 3/3
Then here we can explicitly `continue` for `TeliodObtainingIdentity` exception and propagate the rest of the errors:

```python
    try:
        if not await self.is_alive():
            raise RuntimeError("socket exists but daemon's not running.")
        # teliod is alive and ready
        break
    # any other exception should propagate up
    except TeliodObtainingIdentity:
        # still initializing retry
        await asyncio.sleep(self.TELIOD_CMD_CHECK_INTERVAL_S)
        continue
```

nat-lab/tests/teliod.py
sfraczek: The test should timeout itself, it shouldn't hang here forever.

nat-lab/tests/test_teliod.py
tomasz-grz: Ah I see "Command executed successfully" is asserted already inside of `teliod.quit()`.
So teliod calls quit on itself when it goes out of scope?

nat-lab/tests/test_teliod.py
```diff
@@ -30,17 +30,17 @@
...
 )
-async def test_teliod_start(daemonized_mode) -> None:
+async def test_teliod_start(no_detach) -> None:
     async with AsyncExitStack() as exit_stack:
         connection = (
             await setup_connections(exit_stack, [ConnectionTag.DOCKER_CONE_CLIENT_1])
         )[0].connection
 
-        teliod = Teliod(connection, exit_stack, Config(detach=daemonized_mode))
+        teliod = Teliod(connection, exit_stack, Config(detach=not no_detach))
```
tomasz-grz: What do you think about updating `Config()` as well?

```python
Config(no_detach=no_detach))
```

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
...
+
+            if not self.config.no_detach:
+                stdout, stderr = await self.execute_command(Command.start(self.config))
+                await wait_for_teliod_start()
+            else:
+                proc = await self.run_command(Command.start(self.config))
+                await wait_for_teliod_start()
+                stdout, stderr = proc.get_stdout(), proc.get_stderr()
+
+            assert len(stderr) == 0
```
tomasz-grz: Should we print what's in `stderr` and some message, if this assertion fails? Otherwise might be a bit mysterious debugging

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,305 @@
...
+    def __init__(
+        self,
+        config_type: IfcConfigType = IfcConfigType.DEFAULT,
+        detach: bool = True,
+        paths=Paths(),
+    ):
+        self.paths: Paths = paths
+        self.config_type: IfcConfigType = config_type
+        self.no_detach: bool = not detach
+
```
tomasz-grz:
```suggestion
class Config:
    def __init__(
        self,
        config_type: IfcConfigType = IfcConfigType.DEFAULT,
        no_detach: bool = False,
        paths=Paths(),
    ):
        self.paths: Paths = paths
        self.config_type: IfcConfigType = config_type
        self.no_detach: bool = no_detach
```

nat-lab/tests/teliod.py
sfraczek: Do you mean is_alive wouldn't return False, or have you just omitted that?

tomasz-grz: `is_alive()` would return false when "Command executed successfully" is not in stdout.. 
But in practice it means that it would throw an ProcessExecError 🤔 
I guess it could be then something like
```python
        if "Obtaining identity, ignoring" in exc.stdout:
            raise TeliodObtainingIdentity()
        if "Error: DaemonIsNotRunning" in exc.stderr:
            return False
        # re-raise any other error
        raise
```

nat-lab/tests/teliod.py
tomasz-grz: Yes, but the test timeout is 180 seconds, and we do have a timeout in `wait_for_teliod_socket()`, so why not here?

Or we can remove the timeout in `wait_for_teliod_socket()`, and wrap all of `wait_for_teliod_start()` in 
`await asyncio.wait_for(wait_for_teliod_start(), timeout=self.START_TIMEOUT_S)`?

tomasz-grz: But it's a minor suggestion, and it depends if obtaining identity could take more than 3 seconds.. which I feel like it shouldn't.

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,310 @@
...
+                    await self.remove_socket()
+                else:
+                    log.info("Tried to quit but daemon is already not running")
+
+    async def is_alive(self) -> bool:
+        try:
+            stdout, _ = await self.execute_command(Command.is_alive())
+            return "Command executed successfully" in stdout
+        except ProcessExecError as exc:
+            if "Obtaining identity, ignoring" in exc.stdout:
```
tomasz-grz, line 205: This might need to be
```suggestion
            if "Obtaining identity, ignoring" in exc.stderr:
```

tomasz-grz: ~~Or is it `exc.stdout`? Then `if "Error: DaemonIsNotRunning" in exc.stdout:` too~~

tomasz-grz: Never mind..
`Command executed failed: Obtaining identity, ignoring` is in STDOUT
`Error: CommandFailed(IsAlive)` is in STDERR

sfraczek: Thanks, how did you check it?

nat-lab/tests/teliod.py
sfraczek: Not if telio.py is also not in the utils/ subdirectory and it is analogous.

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,310 @@
...
+    @classmethod
+    def start(cls, config: "Config") -> "Command":
+        cmd = [str(Paths.exec_path), "start"]
+        if config.no_detach:
+            cmd.append("--no-detach")
+        cmd.append(str(config.path()))
+        return cls(cmd)
+
+    @classmethod
+    def is_alive(cls) -> "Command":
```
stalowyjez, line 75: Hmm, why these return types are sometimes given as strings and sometimes not?

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,310 @@
...
+
+            log_match = re.search(r"Saving logs to:\s*(.+\.log)", stdout)
+            assert log_match, f"Could not find log path in: {stdout}"
+            log_path = Path(log_match.group(1).strip())
+            assert (
+                log_path == self.paths.lib_log
+            ), f"Log path does not match: '{log_path}' != '{self.paths.lib_log}'"
+
+        except AssertionError as e:
+            raise e
```
stalowyjez: What does this catch do?

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,310 @@
...
+        config: Config = Config(),
+    ) -> None:
+        self._connection: Connection = connection
+        self._exit_stack: AsyncExitStack = exit_stack
+        self.config: Config = config
+
+    async def execute_command(
+        self,
+        cmd: Command,
+    ) -> tuple[str, str]:
```
stalowyjez, line 141: Maybe it could return some named tuple, so one can just use `res.stdout` instead of `res.0`?

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,310 @@
...
+
+    @property
+    def lib_log(self) -> Path:
+        return self.log_dir / "teliod_natlab.log"
+
+    def config_path(self, config_type: IfcConfigType = IfcConfigType.DEFAULT) -> Path:
+        return self.config_dir / config_type.value
+
+
+class Command(list):
```
stalowyjez, line 59: Hmm, do we need this additional layer of abstraction here? To me it is kinda weird that you can just run `connection.run(Command.start())` (or sth like this) skipping `Teliod` instance entirely and `self.execute_command(Command.is_alive())` isn't that much simpler than `self.execute_command([str(Paths.exec_path), "start"])` (especially when I think about a command for Teliod I'd think about "start" itself, wouldn't expect executable path tbh)

nat-lab/tests/teliod.py
```diff
@@ -0,0 +1,310 @@
...
+        self, ip_addresses: List[str], vpn_routes: bool
+    ) -> AsyncIterator:
+        """
+        Setups interface addresses and routes manually.
+
+        This function should only be used when interface config provider
+        is set to 'manual' on the teliod config.
+        This is not checked by this function. (TODO: LLT-6476)
+        """
+        router = LinuxRouter(self._connection, IPStack.IPv4)
```
stalowyjez, line 294: Maybe it could be configurabler with IPv6 as well?

README.md
```diff
@@ -72,7 +72,12 @@ cargo build -p tcli
 
 You will need a lightweight Linux docker image with the some networking utilities,
 which are missing from the basic Ubuntu image, so let's create a new one.
-Make a `docker` directory in `tcli-test` and put there the following simple Dockerfile:
+Make a `docker` directory in `tcli-test`:
```
tomasz-grz: Should this be part of this PR?

tomasz-grz (commented):
+1 🚀 
Great work @lcruz99 and @sfraczek 🙌

Please clean up the commits before merging

nat-lab/tests/teliod.py
sfraczek: String is commonly used for forward declaration - here Command is not yet completely defined and cannot be used without quotes. But I can make Process without quotes returned from run_command

README.md
sfraczek: I intentionally put it here but if you insist I can make a separate one?

nat-lab/tests/teliod.py
lcruz99: As @sfraczek mentioned thiis is how forward declaration (or referencing) is done on Python. Command is not declared and/or defined at this point so we use a type alias for it: https://docs.python.org/3/library/typing.html#typing.TypeAlias

nat-lab/tests/teliod.py
sfraczek: I should have removed it.

nat-lab/tests/teliod.py
lcruz99: Command was not designed to be part of the Teliod object instance because it doesn't have any dependence on it, however it's not entirely separated also, it belongs to the teliod module and you can view it from an alternative POV as Teliod::Command. 

I didn't understand why you mention `self.execute_command([str(Paths.exec_path), "start"])`, we use it as you correctly assumed: [self.execute_command(Command.start(self.config)) ](https://github.com/NordSecurity/libtelio/pull/1363/files#diff-9110616e8b95ca1b753835304a509716b8c05a567de0fc7b85c6afc99fac2148R183)

nat-lab/tests/teliod.py
sfraczek: It is always unpacked as `stdout, stderr = execute_command()` at the moment, or `stdout, _ = ...`.
I think it is ok at this moment because then we could think what else is inside res if we only returned `res = execute_command` it is only used in few places in this file.

nat-lab/tests/teliod.py
sfraczek: I think it makes it easy to navigate code and I know that all the commands are specified in one place. I think that otherwise we would eventually be still thinking about some way to pool the command strings like "start" "is-alive" in one place to avoid typos or something.

nat-lab/tests/teliod.py
lcruz99: There's no usecase for it at this point and it seems to me that it's not yet fully supported either, if that's the case adding IPv6 configuration is pointless and could be misleading.

sfraczek: Thanks. I tried to add ipv6 but the http request to register meshnet node is returning ipv4 address and I don't know how to overcome it.

nat-lab/tests/teliod.py
stalowyjez: @lcruz99 I just meant that the mentioned verion seems clear and simple enough to me, I don't see a lot of value in having this additional class, I just imagine that when one reads sth like `self.execute_command(Command.start(config))` it feels a bit like you are starting something and running execute on the result (but you just return a string). You also don't see the real command immediately, but you need do some additional jumping to get to the actual command which is quick when you are editing the code, but when you see just a snippet from the test failure it slows stuff a bit.

@sfraczek I get the idea, I think that wrapping it in `Teliod::start()` and similar is enough, though.

Anyway, I don't think that's anything important, you can just resolve it

PR #1366, opened 2025-07-22
Prepare release v6.0.1

mathiaspeters:
### Problem
We have an issue with NepTUN on certain devices with `v6.0.0`

### Solution
Release `v6.0.1` with an updated NepTUN version.

Note: Bumping NepTUN from `v2.1.0` to `v2.2.1` might look like a big update but it contains the panic fixes and new config options that are optional, so nothing breaking.

- [x] README.md is updated

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,13 @@
 <!-- Note: this file is auto-generated. See CONTRIBUTING.md for details. -->
 
+### v6.0.1
+### **Pink Soup**
+---
+* LLT-6456: Handle panics in NepTUN to prevent runtime crashes 
+* LLT-6382: Make new internal NepTUN channel sizes configurable
+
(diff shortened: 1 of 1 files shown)


Hasan6979 (commented):
+1

crates/telio-wg/src/adapter/neptun.rs
```diff
@@ -27,6 +27,7 @@ pub struct NepTUN {
 }
 
 impl NepTUN {
+    #[allow(clippy::too_many_arguments)]
```
tomasz-grz, line 30: 😿 Maybe we should pass it the `DeviceConfig` directly?
Not the job for a release PR of course

tomasz-grz (approved):
Looks good +1

PR #1368, opened 2025-07-24
Teliod natlab test waits for is-alive

tomasz-grz:
### Problem

### Solution


gytsto (commented):
+1

nat-lab/tests/test_teliod.py
```diff
@@ -72,7 +72,15 @@ async def wait_for_is_teliod_running(connection):
...
-                return
+                if (
+                    "Command executed successfully"
+                    in (
+                        await connection.create_process(
+                            TELIOD_IS_ALIVE_PARAMS
+                        ).execute()
+                    ).get_stdout()
+                ):
+                    return
```
LukasPukenis, line 83: I have slight concern about the latency of `create_process` since if it's small, it might introduce a nice hot path by constantly spawning process which a simple `sleep(1)` could solve :thinking:

LukasPukenis (approved):
+1 left a small comment

nat-lab/tests/test_teliod.py
tomasz-grz: There is an  `await asyncio.sleep(0.1)` below

LukasPukenis: whoops :D

PR #1369, opened 2025-07-24
allow trailing close in conntracker

Hasan6979:
### Problem
test_firewall_blacklist_tcp nat-lab test fails because TcpState CLOSE comes immediately after TIME_WAIT and we just check for last 3 events upto TIME_WAIT

### Solution
Add optional to check whether events above the last event match the event sequences.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection_tracker.py
+++ b/nat-lab/tests/utils/connection_tracker.py
@@ -198,7 +198,13 @@ class TCPStateSequence(ConnTrackerEventsValidator):
     Note this validator allows for various TCP states on connections, but full sequence of states must *end* in specified sequence.
     """
 
-    def __init__(self, key: str, five_tuple: FiveTuple, sequence: List[TcpState]):
+    def __init__(
+        self,
+        key: str,
+        five_tuple: FiveTuple,
+        sequence: List[TcpState],
+        trailing_state: Optional[TcpState] = None,
+    ):
         if five_tuple.protocol is None or five_tuple.protocol != "tcp":
             raise ValueError(
                 'TcpStateSequence validator is only available for "tcp" protocol five tuples'
@@ -212,6 +218,7 @@ class TCPStateSequence(ConnTrackerEventsValidator):
         self.key = key
         self.five_tuple = five_tuple
         self.sequence = sequence
+        self.trailing_state = trailing_state
 
     def __repr__(self):
         return f"TCPStateSequence(key: {self.key}, five_tuple: {self.five_tuple}, sequence: {self.sequence})"
@@ -254,9 +261,16 @@ class TCPStateSequence(ConnTrackerEventsValidator):
         # Verify whether all conections end up with expected sequence of TCP states
         violations: list[Optional[ConnTrackerViolation]] = []
         for connection in sequences:
-            state_sequence = list(map(lambda c: c.tcp_state, connection))[
-                -len(self.sequence) :
-            ]
+            if self.trailing_state == connection[-1].tcp_state:
+                # Skip the trailing tcp state
+                state_sequence = list(map(lambda c: c.tcp_state, connection))[
+                    -len(self.sequence) - 1 : -1
+                ]
+            else:
+                state_sequence = list(map(lambda c: c.tcp_state, connection))[
+                    -len(self.sequence) :
+                ]
+
             if state_sequence != self.sequence:
                 violations.append(
                     ConnTrackerViolation(
--- a/nat-lab/tests/test_vpn.py
+++ b/nat-lab/tests/test_vpn.py
@@ -471,6 +471,7 @@ async def test_firewall_blacklist_tcp(ipv4: bool) -> None:
                     "telio-kill-blacklisted-connection",
                     FiveTuple(protocol="tcp", dst_ip=serv_ip, dst_port=serv_port),
                     [TcpState.FIN_WAIT, TcpState.LAST_ACK, TcpState.TIME_WAIT],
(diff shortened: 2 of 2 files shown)


nat-lab/tests/utils/connection_tracker.py
```diff
@@ -250,14 +249,20 @@ def find_conntracker_violations(
...
                 -len(self.sequence) :
             ]
+
             if state_sequence != self.sequence:
+                if trailing_close == TcpState.CLOSE:
+                    skip_close_state_sequence = list(
+                        map(lambda c: c.tcp_state, connection)
+                    )[-len(self.sequence) - 1 : -1]
+                    if skip_close_state_sequence == self.sequence:
+                        continue
```
LukasPukenis: I am not sure if I understand this code correctly but at a glance - could we, for readability purposes, just check the last element for being `trailing_close` and removing it if so? This would make it very clear what is happening :)

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -250,14 +249,20 @@ def find_conntracker_violations(
...
-
         # Verify whether all conections end up with expected sequence of TCP states
         violations: list[Optional[ConnTrackerViolation]] = []
         for connection in sequences:
             state_sequence = list(map(lambda c: c.tcp_state, connection))[
                 -len(self.sequence) :
             ]
+
             if state_sequence != self.sequence:
+                if trailing_close == TcpState.CLOSE:
```
stalowyjez: Hmm, couldn't it just be a bool?

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -250,14 +249,20 @@ def find_conntracker_violations(
...
+                    skip_close_state_sequence = list(
+                        map(lambda c: c.tcp_state, connection)
+                    )[-len(self.sequence) - 1 : -1]
```
stalowyjez: Hmm, shouldn't you check here if this last state is really `CLOSE` state?

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -217,7 +217,7 @@ def __repr__(self):
         return f"TCPStateSequence(key: {self.key}, five_tuple: {self.five_tuple}, sequence: {self.sequence})"
 
     def find_conntracker_violations(
-        self, events: List[ConntrackerEvent]
+        self, events: List[ConntrackerEvent], trailing_close: Optional[TcpState]
```
tomaszklak: Would it not be simpler to add that optional parameter to the constructor of `TCPStateSequence`? This way, we could avoid adding unused parameter in other places.

nat-lab/tests/utils/connection_tracker.py
tomaszklak: nit: It took me a moment to understand this slicing. Could you add an explanation what it does in a comment above

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -198,7 +198,13 @@ class TCPStateSequence(ConnTrackerEventsValidator):
     Note this validator allows for various TCP states on connections, but full sequence of states must *end* in specified sequence.
     """
 
-    def __init__(self, key: str, five_tuple: FiveTuple, sequence: List[TcpState]):
+    def __init__(
+        self,
+        key: str,
+        five_tuple: FiveTuple,
+        sequence: List[TcpState],
+        trailing_seq: Optional[TcpState] = None,
```
stalowyjez: Nit: If that's only one state, shouldn't that be `trailing_state` instead of `trailing_sequence`?

nat-lab/tests/utils/connection_tracker.py
```diff
@@ -254,9 +261,16 @@ def find_conntracker_violations(
...
+            if self.trailing_seq == connection[-1].tcp_state:
+                # Skip the trailing tcp state
+                state_sequence = list(map(lambda c: c.tcp_state, connection))[
+                    -len(self.sequence) - 1 : -1
+                ]
+            else:
+                state_sequence = list(map(lambda c: c.tcp_state, connection))[
+                    -len(self.sequence) :
+                ]
+
```
stalowyjez: Nit: Shouldn't we first check against state sequence? I mean, it will probably work as it is, but theoretically, if the sequence would be `[Established, Close]` and expected also `[Established, Close]`, with `Close` in trailing, then it would add some incorrect violation, wouldn't it?

Hasan6979: The expected sequence in this case would not be `[Established, Close]` but rather `[Established, TimeWait]` for eg. The problem is that we normally don't wait for Close to come, and in normal cases it isn't a problem but sometimes we get an "extra" tcp state at the end, which isn't wrong but it came quicker than expected so we want to ignore the last state. This particular case isn't true for all the tests, and has shown up only in 1 test.

nat-lab/tests/utils/connection_tracker.py
Hasan6979: Makes sense. Changed.

nat-lab/tests/utils/connection_tracker.py
stalowyjez: Yeah, I just meant that if we would wait for such state in the one for some test it could be problematic - but I guess in such case one shouldn't add Close as trailing state? IDK, I think it could be a bit cleaner to check if that's not the case, but for testing purposes that should be good enough

stalowyjez (approved):
lgtm, +1

PR #1370, opened 2025-07-25
Extend natlab-mocked core api to return countries and servers

asledzinskiy:
### Problem
Teliod needs 2 api endpoints to select server for connection by teliod.

### Solution
Add one endpoint that returns list of countries and second that returns list of servers based on provided filters

### DIFF
--- a/nat-lab/bin/core-api.py
+++ b/nat-lab/bin/core-api.py
@@ -11,6 +11,8 @@ from http.server import BaseHTTPRequestHandler, HTTPServer
 from itertools import count
 from typing import Dict, Optional
 from uuid import uuid4
+from mocked_core_api_servers_data import get_countries, get_servers
+from urllib.parse import urlparse, parse_qs
 
 DERP_SERVER = {
     "region_code": "nl",
@@ -46,6 +48,7 @@ class CoreApiErrorCode(Enum):
     INVALID_CREDENTIALS = 100104
     AUTHORIZATION_HEADER_NOT_PROVIDED = 100105
     AUTHORIZATION_HEADER_INVALID = 100106
+    RESOURCE_NOT_FOUND = 404
 
 
 @dataclass
@@ -140,6 +143,8 @@ class CoreApiHandler(BaseHTTPRequestHandler):
         self.server: CoreServer
         self.machines_path = "/v1/meshnet/machines"
         self.notifications_path = "/v1/notifications/tokens"
+        self.recommended_servers_path = "/v1/servers/recommendations"
+        self.countries_path = "/v1/countries"
         super().__init__(request, client_address, server)
 
     def _set_headers(
@@ -242,6 +247,12 @@ class CoreApiHandler(BaseHTTPRequestHandler):
         elif self.path.split("/")[-1] == "map":
             machine_id = self.path.split("/")[-2]
             self.handle_get_machine_map(machine_id)
+        elif self.path == self.countries_path:
+            self.handle_get_countries()
+        elif self.path.startswith(self.recommended_servers_path):
+            parsed_url = urlparse(self.path)
+            query_params = parse_qs(parsed_url.query)
+            self.handle_get_servers(query_params)
 
     def do_HEAD(self):
         self._set_headers()
@@ -403,6 +414,29 @@ class CoreApiHandler(BaseHTTPRequestHandler):
             "expires_in": 60,
         }
         self._write_response(response, status_code=HTTPStatus.CREATED)
+
+    @requires_bearer_token
+    def handle_get_countries(self):
+        countries = get_countries()
+        self._write_response(countries)
+
+    @requires_bearer_token
+    def handle_get_servers(self, query_params):
+        print(f"Passed params: {query_params}")
+        filters = {}
+        for key, values in query_params.items():
+            if key.startswith("filters[") and key.endswith("]"):
+                filter_key = key[len("filters["):-1]
+                filters[filter_key] = values[0]
+        servers = get_servers(filters)
+        if not servers:
+            self._send_error_response(
+                CoreApiErrorCode.RESOURCE_NOT_FOUND,
+                "No vpn servers found for provided filters",
+                HTTPStatus.NOT_FOUND,
+            )
+            return
+        self._write_response([asdict(server) for server in servers])
 
 
 def run(mqttc, port=443):
--- a/nat-lab/bin/mocked_core_api_servers_data.py
+++ b/nat-lab/bin/mocked_core_api_servers_data.py
@@ -1,0 +1,423 @@
+from dataclasses import dataclass
+from typing import List, Dict
+
+
+@dataclass
+class City:
+    id: int
+    name: str
+    latitude: float
+    longitude: float
+    dns_name: str
+
+
+@dataclass
+class Country:
+    id: int
+    name: str
+    code: str
+    city: City
+
+
+@dataclass
+class Location:
+    id: int
+    created_at: str
+    updated_at: str
+    longitude: float
+    latitude: float
+    country: Country
+
+
+@dataclass
+class Service:
+    id: int
+    name: str
+    identifier: str
+    created_at: str
+    updated_at: str
+
+
+@dataclass
+class TechnologyPivot:
+    server_id: int
+    technology_id: int
+    status: str
+
+
+@dataclass
+class Metadata:
+    name: str
+    value: str
+
+
+@dataclass
+class Technology:
+    id: int
+    name: str
+    identifier: str
+    created_at: str
+    updated_at: str
+    metadata: List[Metadata]
+    pivot: TechnologyPivot
+
+
+@dataclass
+class GroupType:
+    id: int
+    created_at: str
+    updated_at: str
+    title: str
+    identifier: str
+
+
+@dataclass
+class Group:
+    id: int
+    created_at: str
+    updated_at: str
+    title: str
+    type: GroupType
+
+
+@dataclass
+class SpecificationValue:
+    id: int
+    value: str
+
+
+@dataclass
+class Specification:
+    id: int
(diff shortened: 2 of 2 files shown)


nat-lab/bin/mocked_core_api_servers_data.py
```diff
@@ -0,0 +1,423 @@
...
+        {
+            "id": 35,
+            "name": "Wireguard",
+            "identifier": "wireguard_udp",
+            "created_at": "2019-02-14 14:08:43",
+            "updated_at": "2019-02-14 14:08:43",
+            "metadata": [
+                {
+                    "name": "public_key",
+                    "value": "kjAOzXQRVGpmQdqE2zPsITH8QHmFK83AAPktqWed9wM="
```
asledzinskiy: Public key is dynamically generated by tests so need to figure out how to retrieve it by mocked data

olekoliinyk: What if we add our "internal" POST endpoint and will update it before the test runs? - maybe just random idea

tomasz-grz: I think you can get the key from [config module](https://github.com/NordSecurity/libtelio/blob/87fb7cb57c9c8b39959ef0aae37d01dce610e691/nat-lab/tests/config.py#L145)

Like:
```python
public_key = WG_SERVER["public_key"]
```

olekoliinyk: I guess we can't directly do the import from `tests` to `bin`.
Bin supposed to be fully isolated which runs defined services.

tomasz-grz: Hmm, right, then maybe the POST idea, or reading it from a file?

How does the VPN server get it's private key?

tomasz-grz: It's [here](https://github.com/NordSecurity/libtelio/blob/87fb7cb57c9c8b39959ef0aae37d01dce610e691/nat-lab/tests/mesh_api.py#L377), looks like it's just copying the file to the server container. so we could do similar, in the same step?

mathiaspeters: I don't think we can import the config module and have it working since they will run in different python sessions/interpreters and the key is populated on startup, which should make it interpreter-specific. But the internal POST endpoint would work well here

asledzinskiy: Implemented as internal endpoint

nat-lab/bin/core-api.py
```diff
@@ -404,6 +421,58 @@ def handle_get_notifications_token(self):
...
+            self._send_error_response(
+                CoreApiErrorCode.RESOURCE_NOT_FOUND,
+                "No vpn servers found for provided filters",
+                HTTPStatus.NOT_FOUND,
+            )
+            return
+        self._write_response([asdict(server) for server in servers])
+
+    @requires_bearer_token
+    def handle_public_key(self):
```
tomasz-grz, line 464: This is for preparing the tests with public key?

Can we please add a comment to this, also mentioning that it's not part of the official API, but only mock for testing

nat-lab/bin/core-api.py
```diff
@@ -404,6 +421,58 @@ def handle_get_notifications_token(self):
...
+        content_length = int(self.headers["Content-Length"])
+        post_data = self.rfile.read(content_length).decode("utf-8")
+        json_obj = json.loads(post_data)
+        pub_key = json_obj.get("public_key")
```
tomasz-grz, line 470: Also how do we specify for which server is this `public_key`?
We currently have 2 servers, potentially more could be added.

nat-lab/tests/test_core_api.py
```diff
@@ -374,6 +374,9 @@ async def test_not_able_to_register_same_machine_twice():
         ("/v1/meshnet/machines/uid", "DELETE"),
         ("/v1/meshnet/machines/uid", "PATCH"),
         ("/v1/notifications/tokens", "POST"),
+        ("/v1/countries", "GET"),
+        ("/v1/servers/recommendations", "GET"),
+        ("/v1/public-key", "POST"),
```
tomasz-grz: This is not part of the official API, so should it be `v1`?
What do you think of calling it maybe something like `/test/set-public-key`?

tomasz-grz (commented):
Nice work! 🙏 
Just two points from my side

nat-lab/bin/core-api.py
asledzinskiy: Yes, thanks

nat-lab/tests/test_core_api.py
asledzinskiy: Agree, but maybe /test/public-key ? in REST verb is usually not part of the url, action is defined by HHTP method itself

nat-lab/bin/core-api.py
asledzinskiy: Good point. I can add another parameter - country_id as we will use it anyway in filters.
And save public key under name like - "/tmp/public_key_{country_id}.pub"

nat-lab/bin/mocked_core_api_servers_data.py
```diff
@@ -0,0 +1,377 @@
...
+}
+
+country_to_server_mapping: Dict[int, dict] = {
+    1: server_poland,
+    2: server_germany,
+}
+
+
+def get_countries() -> List[Dict]:
+    countries = [country_poland, country_germany]
```
tomaszklak, line 301: The jira task asks for 3-5 countries, which means we are missing at least one

nat-lab/bin/core-api.py
```diff
@@ -404,6 +421,58 @@ def handle_get_notifications_token(self):
...
+        if not os.path.isfile(SERVER_PUBLIC_KEY_PATH):
+            self._send_error_response(
+                CoreApiErrorCode.RESOURCE_NOT_FOUND,
+                "There is no saved public key for server. "
+                "Use 'public-key' endpoint to save it first",
+                HTTPStatus.NOT_FOUND,
+            )
+            return
+        with open(SERVER_PUBLIC_KEY_PATH, "r", encoding="utf-8") as f:
+            public_key = f.read()
```
tomaszklak: This should be the first thing we do in this function.

nat-lab/bin/core-api.py
```diff
@@ -404,6 +421,58 @@ def handle_get_notifications_token(self):
...
+        self._write_response(countries)
+
+    @requires_bearer_token
+    def handle_get_servers(self, query_params):
+        print(f"Passed params: {query_params}")
+        filters = {}
+        for key, values in query_params.items():
+            if key.startswith("filters[") and key.endswith("]"):
+                filter_key = key[len("filters[") : -1]
+                filters[filter_key] = values[0]
```
tomaszklak, line 439: Should we assert that there was only one value?

nat-lab/bin/mocked_core_api_servers_data.py
```diff
@@ -0,0 +1,377 @@
...
+    return countries
+
+
+def get_servers(filters: Dict[str, str], public_key: str) -> List[Server]:
+    country_id = int(filters.get("country_id", 1))
+    city_id = int(filters.get("country_city_id", 1))
```
tomaszklak, line 307: Why do we extract this value if we don't use it later?

nat-lab/bin/mocked_core_api_servers_data.py
```diff
@@ -0,0 +1,377 @@
...
+    print(f"Passed city and country filters: {city_id}, {country_id} ")
+    vpn_server = country_to_server_mapping.get(country_id, None)
+    if not vpn_server:
+        return []
+    vpn_server["technologies"][1]["metadata"][0]["value"] = public_key
+    return [
+        Server(
```
tomaszklak: Could we maybe avoid all of this typing somehow? Maybe https://github.com/konradhalas/dacite could work for our case?

nat-lab/bin/mocked_core_api_servers_data.py
asledzinskiy: We figured out that we only have 2 vpn servers in nat-lab at the moment and it makes sense at first cover only 2 countries. I'll leave the comment in Jira as @mathiaspeters is on vacation and can't ask him to modify ticket description

nat-lab/bin/core-api.py
asledzinskiy: Probably makes sense as a validation of passed query params

nat-lab/bin/mocked_core_api_servers_data.py
asledzinskiy: city_id was going to be used in next versions of teliod but if it's confusing I can delete it for now

nat-lab/bin/core-api.py
asledzinskiy: I added 'country_id' to public.key file name so can't move it before getting params values

nat-lab/bin/mocked_core_api_servers_data.py
asledzinskiy: Let me explore this options, looks good

nat-lab/bin/mocked_core_api_servers_data.py
tomaszklak: Now that you've wrote your comment, it's clear :)

nat-lab/bin/mocked_core_api_servers_data.py
mathiaspeters: The intention, that I for some reason didn't write in the ticket, was to have at least one country without a VPN server as an edge case. But it's not important

tomasz-grz (approved):
+1

PR #1372, opened 2025-07-28
Enable OpenWRT building

LukasPukenis:
### Problem
We want to be able to produce OpenWRT package for testing and feed validation.

### Solution
Add support to produce the OpenWRT package by using the SDK

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4315,9 +4315,9 @@ dependencies = [
 
 [[package]]
 name = "smallvec"
-version = "1.15.0"
+version = "1.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8917285742e9f3e1683f0a9c4e6b57960b7314d0b08d30d1ecd426713ee2eee9"
+checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 
 [[package]]
 name = "smart-default"
@@ -4695,11 +4695,13 @@ dependencies = [
  "if-addrs",
  "ipnet",
  "mockall",
+ "parking_lot",
  "pnet_packet 0.35.0",
  "proptest",
  "proptest-derive",
  "rand 0.8.5",
  "rustc-hash",
+ "smallvec",
  "sn_fake_clock",
  "telio-crypto",
  "telio-model",
--- a/crates/telio-firewall/Cargo.toml
+++ b/crates/telio-firewall/Cargo.toml
@@ -11,6 +11,7 @@ test_utils = [] # For use in benchmarks to avoid duplication of the 'setup' code
 
 [dependencies]
 enum-map = "2.7.3"
+smallvec = "1.15.1"
 hashlink.workspace = true
 if-addrs.workspace = true
 ipnet.workspace = true
@@ -18,6 +19,7 @@ tracing.workspace = true
 mockall = { workspace = true, optional = true }
 pnet_packet.workspace = true
 rustc-hash.workspace = true
+parking_lot.workspace = true
 
 telio-crypto.workspace = true
 telio-utils.workspace = true
--- a/crates/telio-firewall/benches/firewall_bench.rs
+++ b/crates/telio-firewall/benches/firewall_bench.rs
@@ -156,7 +156,7 @@ pub fn firewall_tcp_inbound_benchmarks(c: &mut Criterion) {
                                     .process_inbound_packet(&peers[which_peer], &param.packet));
                                 which_peer = (which_peer + 1) % peers.len();
                             }
-                            assert_eq!((0, 0), firewall.get_state());
+                            assert_eq!((peers.len(), 0), firewall.get_state());
                         });
                     },
                 );
@@ -239,7 +239,8 @@ pub fn firewall_tcp_inbound_benchmarks(c: &mut Criterion) {
                                     firewall.process_inbound_packet(&public_key.0, &param.packet)
                                 );
                             }
-                            assert_eq!((0, 0), firewall.get_state());
+                            // Any passing SYN will create a conntrack entry
+                            assert_eq!((1, 0), firewall.get_state());
                         });
                     },
                 );
@@ -470,7 +471,7 @@ pub fn firewall_udp_inbound_benchmarks(c: &mut Criterion) {
                                     .process_inbound_packet(&peers[which_peer], &param.packet));
                                 which_peer = (which_peer + 1) % peers.len();
                             }
-                            assert_eq!((0, 0), firewall.get_state());
+                            assert_eq!((0, peers.len()), firewall.get_state());
                         });
                     },
                 );
@@ -553,7 +554,8 @@ pub fn firewall_udp_inbound_benchmarks(c: &mut Criterion) {
                                     firewall.process_inbound_packet(&public_key.0, &param.packet)
                                 );
                             }
-                            assert_eq!((0, 0), firewall.get_state());
+                            // Any UDP packet will create a conntrack entry
+                            assert_eq!((0, 1), firewall.get_state());
                         });
                     },
                 );
--- a/crates/telio-firewall/src/conntrack.rs
+++ b/crates/telio-firewall/src/conntrack.rs
@@ -3,10 +3,10 @@ use std::{
     fmt::{Debug, Formatter},
     io,
     net::IpAddr as StdIpAddr,
-    sync::Mutex,
     time::Duration,
 };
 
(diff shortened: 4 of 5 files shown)


ci/build_libtelio.py
```diff
@@ -571,14 +574,35 @@ def exec_build(args):
...
+    if args.os == "openwrt":
+        # OpenWRT is being built from within the OpenWRT SDK so `cargo build` invocation happens there.
+        # TODO: we probably want to have a nordsecurity repository or even better yet - merge it into the official one and point it here
+        os.system(f"sed -i 's|^src-git packages .*|src-git packages https://github.com/LukasPukenis/openwrt-packages;LLT-6479_openwrt_feed |' /usr/local/openwrt_sdk_{args.arch}/feeds.conf.default")
+
+        os.system(f"/usr/local/openwrt_sdk_{args.arch}/scripts/feeds update -a 2>&1 > /dev/null")
+        os.system(f"/usr/local/openwrt_sdk_{args.arch}/scripts/feeds install -a 2>&1 > /dev/null")
+
+        # Without a proper .config file present, OpenWRT will launch it's TUI which waits for user input
+        os.system(f"cd /usr/local/openwrt_sdk_{args.arch}/ && touch .config && make defconfig 2>&1 > /dev/null")
```
tomasz-grz: What's the default `.config`?

Don't we need to specify target architecture in the `.config` file by adding `CONFIG_TARGET_<board>_<target>=y`?

Also setting these?
```
CONFIG_TARGET_BOARD
CONFIG_TARGET_SUBTARGET
CONFIG_TARGET_ARCH_PACKAGES
```

ci/build_libtelio.py
```diff
@@ -571,14 +574,35 @@ def exec_build(args):
...
-        args.arch,
-        args.debug,
-    )
-    rutils.check_config(config)
-    call_build(config, args)
+    
+    if args.os == "openwrt":
+        # OpenWRT is being built from within the OpenWRT SDK so `cargo build` invocation happens there.
+        # TODO: we probably want to have a nordsecurity repository or even better yet - merge it into the official one and point it here
+        os.system(f"sed -i 's|^src-git packages .*|src-git packages https://github.com/LukasPukenis/openwrt-packages;LLT-6479_openwrt_feed |' /usr/local/openwrt_sdk_{args.arch}/feeds.conf.default")
```
stalowyjez: I guess we should move to that before merging - i.e. I'm not sure if that's a good idea to point to private repositories from libtelio files

stalowyjez: I'd also add some comment to that script itself tbh, or maybe it would be possible to rewrite it a bit with some python code?

ci/build_libtelio.py
```diff
@@ -571,14 +574,35 @@ def exec_build(args):
...
+
+        os.system(f"/usr/local/openwrt_sdk_{args.arch}/scripts/feeds update -a 2>&1 > /dev/null")
+        os.system(f"/usr/local/openwrt_sdk_{args.arch}/scripts/feeds install -a 2>&1 > /dev/null")
+
+        # Without a proper .config file present, OpenWRT will launch it's TUI which waits for user input
+        os.system(f"cd /usr/local/openwrt_sdk_{args.arch}/ && touch .config && make defconfig 2>&1 > /dev/null")
+
+        # Will be produced at /usr/local/openwrt_sdk_{args.arch}/bin/packages/{args.arch}/packages/nordvpn_{GIT_SHA}-r1_{args.arch}.ipk
+        os.system(f"cd /usr/local/openwrt_sdk_{args.arch} && make package/nordvpn/download V=s && make package/nordvpn/check V=s && make package/nordvpn/compile -j$(nproc) V=s 2>&1 > /dev/null")
```
stalowyjez: I guess we usually prefer `subprocess`?

stalowyjez (commented):
These shell commands are a bit hard to read to me - I would split them a bit (maybe extracting them to some shell script and comment there, use some `pushd`/`popd` would also help?)

ci/build_libtelio.py
LukasPukenis: So this code here produced just an empty `.config` file. I was pretty surprised that it worked since initially I added multiple entries into it but later on when refactoring I reduced it down to just removing it at which point it failed because TUI was expecting a terminal. Then I tried to just create it as it is here - and to my surprise it just worked.

I haven't added anything extra since this looks to be enough to produce x86_64 package. Extra flags added might be required for other architectures but I would rather add them when needed and when we can actually exercise them.

I know this is not a satisfactory answer but I didn't dig deeper why it is enough to just have this empty file and then `make defconfig` just prepares it as needed :)

ci/build_libtelio.py
LukasPukenis: I did not consider these commands to be for much inspection since they are pretty standard OpenWRT commands and probably will be left like that for a long time - but if you think it might be more worth to have them more readable by splitting all of them separately I have nothing against it and can surely do so :) 

Regarding the private repo, you're right, @tomasz-grz also asked about this so I will fork it right now :)

ci/build_libtelio.py
LukasPukenis: Hmmm... To be honest didn't think about it - I do not care about the output of these commands, as they are just auxiliary OpenWRT commands to build it's package cache and what not. However one thing I didn't think of - `os.system()` returns an error code and I don't check for it, nor does it throw an exception. With `subprocess` it has potential to be more obvious if it fails due to the exception but I am on the edge if it's worth refactoring to be honest.. What's your take on this one?

stalowyjez: Well, that's not that much of refactoring tbh, I would change that to subprocess ^^'

ci/build_libtelio.py
stalowyjez: I feel like some comments or at least some link to docs which explain why this sed replacement looks like it looks, what all of this make commands do etc. is quite necessary - but tbh I think it could be better to move them to some `*.sh` file, add some explanation how it works there, and just call it from here once would be best option

ci/build_libtelio.py
stalowyjez: I mean this is kinda messy, because this function is practically bash script encoded in strings - so I'd either move it to a proper bash script and just run it here, or write it a bit more in python and use `cwd=..` instead of `cd ..`, `stdout=subprocess.DEVNULL` instead of `> /dev/NULL` etc., so it will be easier to work with it with normal developer tool is the future

ci/build_libtelio.py
LukasPukenis: `ci/build_libtelio.py` is invoking scripts from within the docker container. This menas that adding a shell script like `openwrt.sh` would need to be inside of rust_build_utils and not in this repo unless it's somehow mounted. I don't want to investigate into that as that's not necessary and I don't see an issue by providing the commands in the code as well.

So I will opt out for:
- Add better descriptions to the commands and references why they are needed
- Split the commands and don't chain them together
- `os.system -> subprocess`

@stalowyjez how does that sound?

ci/build_libtelio.py
LukasPukenis: In another comment - https://github.com/NordSecurity/libtelio/pull/1372#discussion_r2245522593 I also mentioned `subprocess`.

Lets leave just that one and resolve this one, would you agree?

ci/build_libtelio.py
LukasPukenis: ACtually found out that the empty `.config` is not even needeed, so it's just `make defconfig`

ci/build_libtelio.py
LukasPukenis: @stalowyjez I've replaced `os.system()` with `subsystem`. Added explanations on each command and some references on top.

ci/build_libtelio.py
```diff
@@ -571,14 +575,94 @@ def exec_build(args):
...
+            try:
+                subprocess.run(cmd, check=True, **kwargs)
+            except subprocess.CalledProcessError as e:
+                raise RuntimeError(f"Command failed [{e}]: {' '.join(cmd)}")
+
+        # Replace the default OpenWRT 'packages' feed with our fork to simulate the package being inlined and paths set up correctly.
+        # The goal for packages are not to live in separate feeds but rather be inlined in the main openwrt packages repo.
+        run([
+            "sed", "-i",
+            "s|^src-git packages .*|src-git packages https://github.com/NordSecurity/openwrt-packages;LLT-6479_openwrt_feed |",
```
LukasPukenis: I will replace the branch name with `main` once the feed will be merged. Now it is like so in order to validate the whole chain of actions

PR #1374, opened 2025-07-30
Add mkdir and cd to readme

sfraczek:
### Problem
As a first time user while following README.md I was looking for a directory `tcli-test` in which I was supposed to create `docker` subfolder inside, but the `tcli-test` directory also has to be created.

### Solution
Make the README.md super clear about making both directories.

- [x] README.md is updated


tomasz-grz (approved):
+2

PR #1388, opened 2025-08-06
Fix log typo

Hasan6979:
### Problem
The log line prints unparked, whereas status above is checking whether thread was parked.

### Solution
Change logging from "unparked" to "parked"

- [ ] README.md is updated
- [ ] Functionality is covered by unit or integration tests

### DIFF
--- a/crates/telio-utils/src/tokio.rs
+++ b/crates/telio-utils/src/tokio.rs
@@ -43,7 +43,7 @@ impl ThreadTracker {
         {
             let delta = now - thread_last_status_change;
             if status == ThreadStatus::Parked && delta > UNPARKED_THRESHOLD {
-                telio_log_debug!("Thread {tid:?} was unparked for too long: {delta:?}");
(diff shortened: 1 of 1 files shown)


crates/telio-utils/src/tokio.rs
```diff
@@ -43,7 +43,7 @@ impl ThreadTracker {
         {
             let delta = now - thread_last_status_change;
             if status == ThreadStatus::Parked && delta > UNPARKED_THRESHOLD {
-                telio_log_debug!("Thread {tid:?} was unparked for too long: {delta:?}");
+                telio_log_debug!("Thread {tid:?} was parked for too long: {delta:?}");
```
tomaszklak, line 46: This is changing state from unparked to parked. So the thread **was** parked from the time it's status was previously changed up to now. So I think the log is correct.

Hasan6979: Yes. It seems so, it was discussed in the natlab meeting and the log looked confusing at first. Closing this.

PR #1389, opened 2025-08-06
Add docs about natlab RPC

mathiaspeters:
### Problem
There has been a knowledge gap about how the communication is done between test host and libtelio instances in containers/VMs in natlab

### Solution
Add docs

- [x] README.md is updated

### DIFF
--- a/nat-lab/README.md
+++ b/nat-lab/README.md
@@ -47,3 +47,34 @@ To update the diagram run
 python3 utils/generate_network_diagram.py docker-compose.yml network.md
 ```
 
+## Architecture
+
+### RPC
+
+When a natlab test is run, the actual test is run on the machine/VM/container that is acting as the test host, but it needs to interact with instances of libtelio on other containers/VMs. Otherwise, we wouldn't be able to enable VPN, check events, etc which are essential to having tests we can trust. In the past, we would start `tcli` on the relevant container/VM and then interact with it over stdout/stdin. Since libtelio v5 we are generating our FFI bindings with UniFFI, and with that we decided that it would be good to use the generated bindings in natlab as well to dogfood not just the libtelio functionality, but also the bindings themselves. Since natlab is written in python the natural choice was to have UniFFI generate python bindings and use those in natlab, but that meant that we need to change how we interact with the remote libtelio instances. We chose Pyro5 for those interactions.
+
+In a nutshell, Pyro5 is a remote object library for python that allows two different python scripts that may or may not be running on different machines to interact as if they were part of the same script. Pyro5 connections have a `proxy` and a `remote`, where the `remote` has some functionality that the `proxy` needs to interact with. `proxy` and `remote` are disctinct roles in a connection and as such, for a single connection only one side can be `proxy` and only one side can be `remote`. That said, it is entirely possible to have multiple connections between two scripts, allowing either side of the connection to act as both `proxy` and `remote` at the same time. In natlab we are not using that kind of bidirectional connection so for us the separation of concern is clear: the test host is the `proxy` and the container/VM in which libtelio is running is the `remote`.
+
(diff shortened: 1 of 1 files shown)


nat-lab/README.md
```diff
@@ -47,3 +47,16 @@ To update the diagram run
 python3 utils/generate_network_diagram.py docker-compose.yml network.md
 ```
 
+## Architecture
+
+### RPC
+
+When a natlab test is run, the actual test is run on the machine/VM/container that is acting as the test host, but it needs to interact with instances of libtelio on other containers/VMs. Otherwise, we wouldn't be able to enabled VPN, check events, etc which are essential to having tests we can trust. In the past, we would start `tcli` on the relevant container/VM and then interact with it over stdout/stdin. Since libtelio v5 we are generating our FFI bindings with UniFFI, and with that we decided that it would be good to use the generated bindings in natlab as well to dogfood not just the libtelio functionality, but also the bindings themselves. Since natlab is written in python the natural choice was to have UniFFI generate python bindings and use those in natlab, but that meant that we need to change how we interact with the remote libtelio instances. We chose Pyro5 for those interactions.
```
LukasPukenis: > we wouldn't be able to enabled VPN, 
typo :)

nat-lab/README.md
```diff
@@ -47,3 +47,16 @@ To update the diagram run
...
+
+On the container/VM where libtelio is to be run, we execute `libtelio_remote.py` which dynamically loads the libtelio binary. It then starts a Pyro5 server, letting Pyro5 select an available port. That port is then written to stdout. The `remote` is started by a natlab test `Client` (from `telio.py`), which can pick up the port number by reading the port number from stdout, and creating a `LibtelioProxy` object (from `libtelio_proxy.py`) and pointing it at the IP of the relevant container/VM and the port it just picked up. This kind of port/ervice discovery is done to avoid port collisions, which was fairly common before when the ports were decided by the test host.
```
LukasPukenis: > which dynamically loads the libtelio binary
maybe clarify that it is supposed to be a library, binary looks very general :)

LukasPukenis (commented):
+1 Nice!

As a fan of diagrams, adding MermaidJS markup to illustrate host-VM interaction would be a great addition that's also nicely rendered in GitHub.

That said, it doesn't lessen the value of the current work :)

nat-lab/README.md
sfraczek: typo `port\ervice`

nat-lab/README.md
```diff
@@ -47,3 +47,16 @@ To update the diagram run
...
+
+Pyro5 can't send arbitrary objects back and forth, only objects explicitly exposed over the Pyro5 layer. To still allow sending strongly typed classes between `proxy` and `remote`, there is a serialization implementation that will transparently convert to and from python dictionaries. In `serialization.py` there is a function `init_serialization` that will configure Pyro5 in the current python interpreter to be able to serialize and deserialize pretty any object. Serializing is one by creating a dict that contains the class name and the data the object holds as a dict, and deserialization is the same but opposite. This way, we can send any objects back and forth without ever having to deal with serialization issues.
```
sfraczek: What do you mean by `pretty` any object? Like `virtually` any object?

nat-lab/README.md
mathiaspeters: Fixed

nat-lab/README.md
mathiaspeters: Fixed and fixed

nat-lab/README.md
mathiaspeters: It was supposed to say `pretty much` but it seems a word got lost somewhere. I added it so now it should be coherent. I think the (de)serialization could handle _any_ object but just to be on the safe side, `pretty much` is used, though I can't think of any object that couldn't be handled

sfraczek (approved):
+1

PR #1390, opened 2025-08-06
[LLT-6406] Teliod is not registering device on core api

lcruz99:
### Problem
While writing tests for the teliod identity file management on https://github.com/NordSecurity/libtelio/pull/1322, there were a few issues found on the identity file logic:
1. Returned status code from API evaluation was checking for specific error instead of HTTP success codes (2xx).
2. According to the documentation, API HTTP errors should be handled with wether they are classified as terminal or not.
3. The identity file state machine from the teliod::core_api module was trivially fetching the machine attributes from the API right after generating new keys and hardware identifier.

Besides the issues mentioned above, tests were also non-existent for this identity file component of the core_api module.

### Solution
This PR adds integration tests for the identity file management functionality of the teliod::core_api module, and tackles the issues mentioned before by:
1. Checking for HTTP success codes
2. Add HTTP status code handling as described in the requirements of the documentation (LLT-5553)*.
3. Adds identity files specific functions to the core_api module that address the state machine and refactor the logic so it follows the documentation (LLT-5553*).

Tests are also added to cover https://github.com/NordSecurity/libtelio/pull/1322 and the identity file management logic that was touched in this PR.

*Besides LLT-5553, LLT-5410 can also be worthwhile taking a look.

- [x] README.md is updated

### DIFF
--- a/clis/teliod/src/config.rs
+++ b/clis/teliod/src/config.rs
@@ -64,7 +64,7 @@ fn reconnect_after_expiry_default() -> Percentage {
     Percentage(90)
 }
 
-#[derive(PartialEq, Eq, Debug, Clone)]
+#[derive(Default, PartialEq, Eq, Debug, Clone)]
 pub struct NordToken(Arc<Hidden<String>>);
 
 impl NordToken {
@@ -81,13 +81,7 @@ impl NordToken {
     }
 
     fn validate(token: &str) -> bool {
-        token.is_empty() || (token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit()))
-    }
-}
-
-impl Default for NordToken {
-    fn default() -> Self {
-        NordToken(Arc::new(Hidden("".to_owned())))
+        token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit())
     }
 }
 
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -106,19 +106,17 @@ impl DeviceIdentity {
         .await
         {
             Ok(id) => id,
-            Err(e) => match e {
-                Error::DeviceNotFound => {
-                    info!("Unable to load identifier due to {e}. Registering ...");
-                    Box::pin(register_machine_with_exp_backoff(
-                        &hw_identifier.to_string(),
-                        private_key.public(),
-                        &config.authentication_token,
-                        &config.http_certificate_file_path,
-                    ))
-                    .await?
-                }
-                _ => return Err(e),
-            },
+            Err(Error::DeviceNotFound) => {
+                info!("Device not found, registering...");
+                Box::pin(register_machine_with_exp_backoff(
+                    &hw_identifier.to_string(),
+                    private_key.public(),
+                    &config.authentication_token,
+                    &config.http_certificate_file_path,
+                ))
+                .await?
+            }
+            Err(e) => return Err(e),
         };
 
         Ok(DeviceIdentity {
@@ -132,24 +130,21 @@ impl DeviceIdentity {
     /// Either the path is provided as parameter or hardcoded path
     /// is used.
     pub fn from_file(path: Option<&Path>) -> Result<DeviceIdentity, Error> {
-        info!("Fetching identity config");
-
         let path = path.unwrap_or(Self::device_identity_path()?);
 
         match std::fs::File::open(path) {
             Ok(file) => {
-                debug!("Found existing identity config {}", path.to_string_lossy());
+                info!("Found existing identity config {}", path.to_string_lossy());
                 serde_json::from_reader(file).map_err(Error::Deserialize)
             }
             Err(e) => {
-                warn!("Reading identity config failed: {e}");
+                info!("Reading identity config failed: {e}");
                 Err(Error::Io(e))
             }
         }
     }
 
-    // Write device identity to the corresponding file.
-    // File path is DEVICE_IDENTITY_FILE_PATH.
+    // Write device identity to file.
     pub fn write(&self) -> Result<(), Error> {
         // User has read/write but others have none
         // Mode for identity file rw-------
@@ -160,6 +155,10 @@ impl DeviceIdentity {
             .write(true)
             .open(Self::device_identity_path()?)?;
         serde_json::to_writer(&mut file, &self)?;
+        info!(
+            "Device identity written to {:?}",
+            Self::device_identity_path()?
+        );
         Ok(())
     }
 
@@ -318,10 +317,36 @@ async fn fetch_identifier_with_exp_backoff(
         match fetch_identifier_from_api(auth_token, cert_path, public_key).await {
             Ok(id) => return Ok(id),
             Err(e) => {
-                warn!(
-                    "Failed to fetch identifier due to {e:?}, will wait for {:?} and retry",
-                    backoff.get_backoff()
-                );
+                warn!("Failed to fetch identifier due to {e:?}.");
+                // More info on error handling: LLT-5553
+                match e {
+                    // Terminal errors
+                    Error::PeerRegistering(_)
+                    | Error::Decode(_)
+                    | Error::InvalidResponse
+                    | Error::UpdateMachine(_)
+                    | Error::ExpBackoffTimeout()
+                    | Error::NoDataLocalDir
+                    | Error::DeviceNotFound
+                    | Error::BackoffBounds(_)
+                    | Error::DeviceIdentityFileAccess => return Err(e),
+
+                    Error::FetchingIdentifier(status_code) => {
+                        if status_code == StatusCode::BAD_REQUEST
+                            || status_code == StatusCode::UNAUTHORIZED
+                            || status_code == StatusCode::FORBIDDEN
+                            || status_code == StatusCode::NOT_FOUND
+                        {
+                            return Err(e);
+                        }
+                        // Server errors (5xx) and other non-terminal errors
+                    }
+
+                    Error::Reqwest(_) | Error::Deserialize(_) | Error::Io(_) => {
+                        // Non-terminal errors
+                    }
+                }
+                warn!("Will wait for {:?} and retry", backoff.get_backoff());
                 wait_with_backoff_delay(&mut backoff, &mut retries).await?;
             }
         }
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -1,7 +1,9 @@
 use base64::{prelude::*, DecodeError};
 use reqwest::{header, Certificate, Client, StatusCode};
-use serde::{Deserialize, Serialize};
+use serde::de::{IgnoredAny, MapAccess, Visitor};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
+use std::fmt;
 use std::fs::{create_dir_all, File, OpenOptions};
 use std::io::Read;
 use std::os::unix::fs::OpenOptionsExt;
@@ -104,6 +106,14 @@ macro_rules! handle_api_error {
     };
 }
 
+/// Result of loading device identity
+enum IdentityLoadResult {
+    /// Identity was loaded from file and synchronized with API
+    Loaded(DeviceIdentity),
+    /// New identity was created and registered
+    Created(DeviceIdentity),
+}
+
 #[derive(Default, Serialize, Deserialize)]
 struct MeshConfig {
     public_key: PublicKey,
@@ -114,7 +124,7 @@ struct MeshConfig {
     traffic_routing_supported: bool,
 }
 
-#[derive(Serialize, Deserialize, Debug, Default)]
+#[derive(Serialize, Debug, Default)]
 pub struct DeviceIdentity {
     pub hw_identifier: uuid::Uuid,
     pub private_key: SecretKey,
@@ -123,35 +133,19 @@ pub struct DeviceIdentity {
 
 impl DeviceIdentity {
     /// Generate a new device identity.
-    /// A new key and hw id are generated, whereas the machine id is got upon
-    /// registering the device from on the API with a [1,180]s exponential backoff.
+    /// A new key and hw id are generated, whereas the machine id is obtained after
+    /// registering the device with the API, which uses an [1,180]s exponential backoff.
     pub async fn new(config: &TeliodDaemonConfig) -> Result<Self, Error> {
         info!("Generating a new device identity..");
         let private_key = SecretKey::gen();
         let hw_identifier = uuid::Uuid::new_v4();
-
-        // LLT-6406: There's likely a bug here since looking for a device with a private key that
-        // we just created above is trivial and will fail. This will always register a new device.
-        let machine_identifier = match fetch_identifier_with_exp_backoff(
+        let machine_identifier = Box::pin(register_machine_with_exp_backoff(
+            &hw_identifier.to_string(),
+            private_key.public(),
             &config.authentication_token,
             &config.http_certificate_file_path,
-            private_key.public(),
-        )
-        .await
-        {
-            Ok(id) => id,
-            Err(Error::DeviceNotFound) => {
-                info!("Device not found, registering...");
-                Box::pin(register_machine_with_exp_backoff(
-                    &hw_identifier.to_string(),
-                    private_key.public(),
-                    &config.authentication_token,
-                    &config.http_certificate_file_path,
-                ))
-                .await?
-            }
-            Err(e) => return Err(e),
-        };
+        ))
+        .await?;
 
         Ok(DeviceIdentity {
             private_key,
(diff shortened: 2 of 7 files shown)


clis/teliod/src/core_api.rs
```diff
@@ -89,36 +133,19 @@ pub struct DeviceIdentity {
 
 impl DeviceIdentity {
     /// Generate a new device identity.
-    /// A new key and hw id are generated, whereas the machine id is fetched from the
-    /// API with a [1,180]s exponential backoff.
+    /// A new key and hw id are generated, whereas the machine id is got upon
+    /// registering the device from on the API with a [1,180]s exponential backoff.
```
jjanowsk: Please clarify the docstring

clis/teliod/src/core_api.rs
```diff
@@ -210,92 +240,187 @@ impl DeviceIdentity {
...
-    };
-    Ok(ExponentialBackoff::new(backoff_bounds)?)
-}
+/// Parses device identity file and handles missing values (LLT-5553).
+/// Returned device identity must be matched against the API using the fetch/update functions (PATCH).
+/// Missing:
+///   - hw_identifier: generates new hw_identifier, API should be updated.
+///   - private_key: generates new key pair, API should be updated.
+///   - machine_identifier: empty string is acceptable as value can be fetched from API.
+impl<'de> Deserialize<'de> for DeviceIdentity {
```
jjanowsk, line 249: Cannot this whole thing be replaced with something like this?
```
#[derive(Serialize, Deserialize, Debug, Default)]
pub struct DeviceIdentity {
    #[serde(default = "default_hw_identifier")]
    pub hw_identifier: uuid::Uuid,
    pub private_key: SecretKey,
    pub machine_identifier: String,
}

fn default_hw_identifier() -> uuid::Uuid {
    warn!("Identity file missing hardware identifier, a new one is generated",);
    uuid::Uuid::new_v4()
}
// ... same for other fields 
```

clis/teliod/src/core_api.rs
```diff
@@ -210,92 +240,187 @@ impl DeviceIdentity {
...
-        )
-        .await?;
-        parsed_device_identity
-    } else {
-        Box::pin(DeviceIdentity::new(config)).await?
+/// Initialize device identity with API synchronization.
+///
+/// This function handles three scenarios:
+/// 1. No existing identity file, creates a new one and registers it [Created]
+/// 2. Existing file but with missing machine_identifier, fetches it from API [Loaded] and registering it if not found [Created]
```
jjanowsk: `and registering` -> `or registers`

clis/teliod/src/core_api.rs
```diff
@@ -210,92 +240,187 @@ impl DeviceIdentity {
...
+/// 3. Complete identity file, validate by updating it on API [Loaded], registering it if not found [Created]
```
jjanowsk: `registering` -> `registers`

clis/teliod/src/core_api.rs
jjanowsk: And in case I have missed something it would be great to have UTs for this deserialization

clis/teliod/src/config.rs
```diff
@@ -68,35 +64,50 @@ fn reconnect_after_expiry_default() -> Percentage {
...
+    }
 }
 
-pub fn generate_hw_identifier() -> Uuid {
-    // Generate hw_identifier
-    debug!("Generating a new hw identifier");
-    Uuid::new_v4()
+impl std::fmt::Display for NordToken {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
Jauler: Does this respect the `Hidden` property? Meaning, that it is printed as plaintext value only for debug builds?

clis/teliod/src/config.rs
```diff
@@ -68,35 +64,50 @@ fn reconnect_after_expiry_default() -> Percentage {
...
+    }
+
+    fn validate(token: &str) -> bool {
+        token.is_empty() || (token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit()))
+    }
+}
+
+impl Default for NordToken {
+    fn default() -> Self {
+        NordToken(Arc::new(Hidden("".to_owned())))
```
Jauler: I wonder, although it is definitely up for discussion. Does it make sense to have `Default` trait for NordToken? As any token constructed this way - will be invalid token.

Jauler: This would also allow us to get rid of `token.is_empty()` case in the `validate()` which kind of is returning `true` for invalid token. :thinking:

nat-lab/data/teliod/config.json
```diff
@@ -2,8 +2,9 @@
     "log_level": "trace",
     "log_file_path": "/var/log/teliod_natlab.log",
     "log_file_count": 0,
-    "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+    "authentication_token": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
```
Jauler, line 5: what token is this :suspect: ?

If the only requirement for it to match one specified in `nat-lab/data/teliod/config_with_vpn_iproute_without_id_file.json`

Maybe we could create something obviously non-real. E.g. something like:
`0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef` and add a comment that this token is for testing purposes only and is not valid generally.

As otherwise we are getting false-positive "leak" reports in H1

nat-lab/tests/test_teliod.py
```diff
@@ -83,87 +65,108 @@ async def test_teliod_logs() -> None:
...
-            authorization_header=CORE_API_BEARER_AUTHORIZATION_HEADER,
-        )
-        device_id = {
-            "hw_identifier": str(hw_identifier),
-            "private_key": list(base64.b64decode(str(private_key))),
-            "machine_identifier": response_data["identifier"],
-        }
+    async with AsyncExitStack() as exit_stack:
+        teliod = await Teliod.new(exit_stack, config_type)
+        node, device_identity = await teliod.register_device_on_core()
```
Jauler, line 70: I wonder, does it make sense, to register `teliod` device to meshnet APIs for VPN-only test case? As `teliod` VPN functionality must be able to function without meshnet.

But I am not sure if this question is in-scope for the current PR.

nat-lab/tests/test_teliod.py
```diff
@@ -83,87 +65,108 @@ async def test_teliod_logs() -> None:
...
-            True,
-            IPStack.IPv4,
-            response_data["ip_addresses"],
-        )
-        api.prepare_all_vpn_servers()
+            assert device_identity == await teliod.read_identity_file()
 
-        return node
 
+async def test_teliod_device_identity_registration() -> None:
```
Jauler, line 90: I am a little bit confused. Is this test verifying anything? 

It seems like we generally check that identity file exists and is a valid JSON. But this looks a bit loose? Should'nt we validate that it matches core_api point of view?

P.S. also why do we need to connect to VPN for the device registration test?

Jauler: Oh I misread a little bit, it configures VPN server to accept client-side public key and then checks that VPN connects.

So we are validating that identity file exists, that its a valid JSON, and at least _some_ valid private key is stored there, although whether it was used during core_api registration - is not verified?

If I am not mistaken - this is still does look like a bit loose validation for device identity.

IMHO better way would be to trigger `teliod` to register itself onto core_api. Then retreive all the information which was used during the registration from core_api and compare it against the information stored in identity file.

nat-lab/tests/test_teliod.py
```diff
@@ -83,87 +65,108 @@ async def test_teliod_logs() -> None:
...
+            status = await teliod.wait_for_meshnet_ip_on_meshmap()
+            device_identity = await teliod.read_identity_file()
+            await teliod.whitelist_device_on_the_vpn_servers(
+                device_identity, [status["meshnet_ip"]]
+            )
+            await teliod.wait_for_vpn_connected_state()
+
 
-        teliod = Teliod(connection, exit_stack, config=Config(config_type))
+async def test_teliod_device_identity_update_when_machine_id_is_missing() -> None:
```
Jauler, line 104: [Some concerns](https://github.com/NordSecurity/libtelio/pull/1390/files#r2269188289) apply to this test as well

nat-lab/tests/test_teliod.py
```diff
@@ -83,87 +65,108 @@ async def test_teliod_logs() -> None:
...
 
-        # we only know the key of the VPN server at runtime and it needs to be in the config before starting teliod
-        await connection.create_process([
-            "sed",
-            "-i",
-            f's#"server_pubkey": .*#"server_pubkey": "{WG_SERVER["public_key"]}"#g',
-            str(teliod.config.path()),
-        ]).execute()
+
+async def test_teliod_device_identity_update_when_keys_are_missing() -> None:
```
Jauler, line 128: [Some concerns](https://github.com/NordSecurity/libtelio/pull/1390/files#r2269188289) apply to this test as well

nat-lab/tests/test_teliod.py
```diff
@@ -83,87 +65,108 @@ async def test_teliod_logs() -> None:
...
             await teliod.wait_for_vpn_connected_state()
 
-            if config_type == IfcConfigType.VPN_MANUAL:
-                await exit_stack.enter_async_context(
-                    teliod.setup_interface(node.ip_addresses, vpn_routes=True)
-                )
 
-            await ping(connection, PHOTO_ALBUM_IP)
-            ip = await stun.get(connection, STUN_SERVER)
+async def test_teliod_device_identity_update_when_hw_id_is_missing() -> None:
```
Jauler, line 155: [Some concerns](https://github.com/NordSecurity/libtelio/pull/1390/files#r2269188289) apply to this test as well

clis/teliod/src/core_api.rs
```diff
@@ -67,159 +124,303 @@ struct MeshConfig {
...
+                    warn!(
+                        "Failed to remove device identity file at {}: {}",
+                        path.display(),
+                        e
+                    );
+                }
+            };
+
+            if let Some(parent) = path.parent() {
+                match std::fs::remove_dir(parent) {
```
Jauler: By any chance, could this be handled by package manager of specific installation?

clis/teliod/src/core_api.rs
```diff
@@ -67,159 +124,303 @@ struct MeshConfig {
...
-            identity_path.push("teliod");
-            if !identity_path.exists() {
-                let _ = create_dir_all(&identity_path);
+/// Parses device identity file and handles missing values (LLT-5553).
+/// Returned device identity must be matched against the API using the fetch/update functions (PATCH).
+/// Missing:
+///   - hw_identifier: generates new hw_identifier, API should be updated.
+///   - private_key: generates new key pair, API should be updated.
+///   - machine_identifier: empty string is acceptable as value can be fetched from API.
+impl<'de> Deserialize<'de> for DeviceIdentity {
```
Jauler, line 249: Could you elaborate a little bit, why we need to expand deserialize/serialize methods for serde? Was there any problems simply defining a `struct` which represents response body model and deserialize/serialize to that?

Or am I completely misreading something here?

clis/teliod/src/daemon.rs
```diff
@@ -84,142 +79,137 @@ fn telio_task(
...
-                                task_set_ip(new_ip_address, tx_channel.clone());
-                            }
-                        }
-                        Err(e) => {
-                            error!("Unable to set meshmap due to {e}");
+    while let Some(cmd) = rx_channel.blocking_recv() {
+        trace!("TelioTask got command {:?}", cmd);
+        match cmd {
+            TelioTaskCmd::UpdateMeshmap(map) => {
+                let some_new_ip_address = map
```
Jauler, line 100: This is probably out of scope for this specific PR. But why only first IP is taken? Should'nt it be all of them?

clis/teliod/src/daemon.rs
```diff
@@ -84,142 +79,137 @@ fn telio_task(
...
-                    // update the interface only if the address is new
-                    if state.interface_ip_address != Some(new_ip_address) {
-                        state.interface_ip_address = Some(new_ip_address);
-                        _ = sys_config
-                            .set_ip(&new_ip_address)
-                            .inspect_err(|e| error!("Failed to set IP with error '{e:?}'"));
-                        task_connect_to_exit_node(config.vpn, tx_channel.clone());
-                    }
+            }
+            TelioTaskCmd::SetIp(new_ip_address) => {
```
Jauler, line 136: Would it make sense to add a validation, whether `new_ip_address` is matching with meshmap if one currently exists? 

As it would produce invalid state, where one IP is set to the interface and another is in meshmap.

But again, may be separate PR or task

clis/teliod/src/daemon.rs
```diff
@@ -84,142 +79,137 @@ fn telio_task(
...
+                    .external_nodes()
+                    .ok()
+                    .and_then(|nodes| {
+                        nodes
+                            .iter()
+                            .find(|n| n.public_key == pubkey)
+                            .map(|n| n.state == NodeState::Connected)
+                    })
+                    .unwrap_or(false);
+                if is_connected {
```
Jauler, line 178: Routes should be configured prior to "Connected" state.

clis/teliod/src/config.rs
lcruz99: Yes, it implements just Display trait for the NordToken, the wrapped type is intact so here just prints the inner type by ref'ing it.

clis/teliod/src/config.rs
lcruz99: IIRC this was discussed during the QNAP UI development, it allows the auth token to be empty on the UI/config file. If not empty then a dummy value should be used, IMO empty token is cleaner and simpler to assert.

nat-lab/data/teliod/config.json
lcruz99: Your suggestion is no different than the previous "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa".. Probably there was some rationale behind this change but I'm not aware of it unfortunately.

@mathiaspeters @olekoliinyk can you help here?

clis/teliod/src/core_api.rs
lcruz99: This deserializer fn parses the contents from the identity file. The caveat was that the device identity file can have any fields missing, not just the hw_identifier, I could follow your suggestion by either modifying DeviceIdentity and wrap the fields on Option's or just create an aux struct for just the parsing purposes which was initial attempt but realised implementing the deserialized was a cleaner path and then found out about the Visitor trait and the result you know already

clis/teliod/src/core_api.rs
lcruz99: changed

clis/teliod/src/core_api.rs
lcruz99: changed

clis/teliod/src/core_api.rs
lcruz99: changed

nat-lab/tests/test_teliod.py
lcruz99: I'd say it's out of scope, the PR focus on improving the already existing logic. Nevertheless, the team is already talking about that topic.

nat-lab/tests/test_teliod.py
lcruz99: The test logic is that teliod is able to start without identity file. This ensures that:
1. teliod is able to generate an identity
2. teliod is able to register that identity into core-api
3. teliod is able to connect to the servers after successful registration

However you pointed right, we can add a match for the identity registered in the core api against the one in the file.

lcruz99: The same match can be added for the followed tests that you also commented the same. wdyt?

clis/teliod/src/core_api.rs
lcruz99: Only the directory removal right? Definitely, for that I can just remove this block or should I modify any already existent platforms?

clis/teliod/src/core_api.rs
lcruz99: https://github.com/NordSecurity/libtelio/pull/1390#discussion_r2278816957

> This deserializer fn parses the contents from the identity file. The caveat was that the device identity file can have any fields missing, not just the hw_identifier, I could follow your suggestion by either modifying DeviceIdentity and wrap the fields on Option's or just create an aux struct for just the parsing purposes which was initial attempt but realised implementing the deserialized was a cleaner path and then found out about the Visitor trait and the result you know already

clis/teliod/src/daemon.rs
lcruz99: It looks like a teliod limitation, probably we should discuss this further.

clis/teliod/src/daemon.rs
lcruz99: I thnk it should be a separate PR or task

clis/teliod/src/daemon.rs
lcruz99: This block had just the indentation changed, not on the scope of the PR. However, do you mean to do the call to ConfigureInterface::set_exit_routes() before checking if we are connected to the VPN?

clis/teliod/src/config.rs
LukasPukenis: I am the one responsible for this technical debt. In this commit: 
https://github.com/NordSecurity/libtelio/commit/4b31b86efe5e58248b924a7443d845eb08314bb0
I've introduced it. It was a band-aid fix for a security issue where the QNAP UI always embedded the token in the HTML form and it was then submitted where backend always expected it. In order to avoid embedding it, an empty value needed to be handled.

Empty value could be handled as "please ignore this value" or "please erase the contents", so obviously the first one was chosen and the empty token became a valid one which is explicitly ignored when handling the REST endpoint.

The correct way would be to have a separate `Token` type something like `UIToken` which would have this edge case and then deserialize properly into `Token` but I do not remember the reasons why I didn't follow through with that.

However, QNAP related code should be removed so the empty token validation should be removed as well. @lcruz99 I suppose the `is_empty()` removal could happen in this PR if it's not too much to ask?

lcruz99: sure, will include it

nat-lab/tests/test_teliod.py
Jauler: I guess lets register a ticket if it is not already done. In the meantime I will resolve this thread.

nat-lab/tests/test_teliod.py
Jauler: > The same match can be added for the followed tests that you also commented the same. wdyt?

yes IMHO the pattern is same, hence the same match fits well.

clis/teliod/src/daemon.rs
Jauler: lets create a ticket. But in the meantime I will resolve this thread.

clis/teliod/src/daemon.rs
Jauler: > This block had just the indentation changed, not on the scope of the PR. 

It may be separate release-blocking ticket. Does not have to be this PR.

> However, do you mean to do the call to ConfigureInterface::set_exit_routes() before checking if we are connected to the VPN?

yes, normally the flow happens like this:
1. contact API, ensure, that subscription is valid for VPN use. Retrieve credentials, or register any keys if needed. Get server list and select VPN server to use
2. Configure OS for VPN usage
3. Attempt to connect to VPN

If we swap 2 and 3 then there is a time gap, where we claim to be connected to VPN, but routing is not yet configured, which enters IP leak territory...

clis/teliod/src/core_api.rs
Jauler: I see @jjanowsk raised the same concern, hence lets continue discussion in his thread. I am resolving this one to avoid dublication.

clis/teliod/src/core_api.rs
Jauler: TBH I am not convinced that it is a cleaner path :thinking: 

The core of it being these two points:
1. If we somehow have invalid config (e.g. existing `hw_id` but not private key or similar) then instead of getting explicit error and handling it - we simply generate partially new identity (but not fully). This is even more so for `machine_identifier` where emtpy string is produced - which is invalid machine identifier.
2. At a higher level, it sounds to me that a decision when to generate new identity is a separate concern, not related to serialization. Hence mangling these two concerns into one, at least to me - makes it harder to understand general flow.

nat-lab/data/teliod/config.json
olekoliinyk: It's a fake token which we use in nat-lab core-api

olekoliinyk: There is a comment in [core-api.py](https://github.com/NordSecurity/libtelio/blob/main/nat-lab/bin/core-api.py#L36) that it is only for testing purposes

PR #1391, opened 2025-08-06
Implement the Login flow for ENS

tomaszklak:
As described in the RFC, this PR adds custom grpc authentication to the ENS endpoint. This authentication schema is used to prove that the client is in possession of it's private key.

- [x] README.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4823,6 +4823,7 @@ dependencies = [
  "telio-utils",
  "thiserror 2.0.12",
  "tracing",
+ "x25519-dalek",
  "zeroize",
 ]
 
@@ -5023,6 +5024,8 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "base64 0.22.1",
+ "blake3",
  "bytes",
  "http",
  "hyper",
@@ -5045,6 +5048,7 @@ dependencies = [
  "tower",
  "tower-service",
  "tracing",
+ "uuid",
 ]
 
 [[package]]
--- a/crates/telio-proto/src/ens.rs
+++ b/crates/telio-proto/src/ens.rs
@@ -1,8 +1,11 @@
-use std::{net::IpAddr, sync::Arc};
+use std::{net::IpAddr, str::FromStr, sync::Arc};
 
+use base64::prelude::{Engine, BASE64_STANDARD};
+use blake3::{derive_key, keyed_hash};
 use grpc::{ConnectionError, Empty};
 use http::Uri;
 use hyper_util::rt::TokioIo;
+use telio_crypto::{SecretKey, SharedSecret};
 use telio_model::PublicKey;
 use telio_sockets::SocketPool;
 use telio_task::io::{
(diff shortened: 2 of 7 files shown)


mathiaspeters:
+1

Cargo.lock
```diff
@@ -5023,6 +5024,8 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "base64 0.22.1",
```
stalowyjez, line 5027: Do we use multiple base64 versions?

crates/telio-proto/src/ens.rs
```diff
@@ -235,35 +225,123 @@ async fn task(
...
+async fn get_login_challenge(
+    external_channel: Channel,
+    vpn_public_key: PublicKey,
+    local_private_key: SecretKey,
+) -> anyhow::Result<AsciiMetadataValue> {
+    let mut login_client = grpc::login_client::LoginClient::new(external_channel);
+
+    let challenge_response = login_client.get_challenge(Empty::default()).await?;
+    let challenge = &challenge_response.get_ref().challenge;
+    let challenge = Uuid::from_str(challenge)?;
```
stalowyjez, line 237: Why UUID here?

nat-lab/bin/ens
```diff
@@ -39,6 +48,46 @@ class EnsServicerStub(ENSServicer):
...
+        if handler_call_details.method == "/ens.ENS/ConnectionErrors":
+            print("metadata: ", handler_call_details.invocation_metadata)
+            metadata = dict(handler_call_details.invocation_metadata)
+            auth_token = metadata.get('authentication')
+            assert auth_token
+
+            decoded_bytes = base64.b64decode(auth_token)
+            assert len(decoded_bytes) == 80
+            
+            client_public_key, challenge_uuid_bytes, received_authentication_tag = decoded_bytes[:32], decoded_bytes[32:48], decoded_bytes[48:]
```
stalowyjez, line 76: Nit: These look like 3 independent definitions, maybe it makes sense to split it into 3 lines

nat-lab/bin/ens
```diff
@@ -39,6 +48,46 @@ class EnsServicerStub(ENSServicer):
...
+            challenge_uuid = uuid.UUID(bytes=challenge_uuid_bytes)
+            # This will rise KeyError when the key is not present:
+            self.grpc_stub.challenges.remove(challenge_uuid)
```
stalowyjez, line 79: So how this will look like from the client's perspective? Will it return some error, or the server will just silently crash?

stalowyjez (approved):
Looks good, +1

Cargo.lock
tomaszklak: telio-proto uses the workspace's version: `base64.workspace = true` as seen in Cargo.toml

crates/telio-proto/src/ens.rs
tomaszklak: I'm not sure I understand the question - where else should the uuid by parsed?

nat-lab/bin/ens
tomaszklak: This shows how the byte array is composed of 3 non-overlapping fields. Also, I wanted to mimic the definition RFC so that it's easy for whoever is reading this code to see that it is matching RFC.

nat-lab/bin/ens
tomaszklak: When the test fails due to this exception, the grpc client will report an warning.

PR #1392, opened 2025-08-07
[LLT-6509] Add self-signed certificate to OpenWrt container

olekoliinyk:
### Problem
To be able to make https requests to nat-lab's core api the self-signed certificate should be available in OpenWrt GW container

### Solution
1. Add core-api's `test.pem` certificate to OpenWrt container

Later on the daemon can be started with config which points to certificate location (tested and it works):
```
{
  "log_level": "trace",
   ....
  "http_certificate_file_path": "/etc/ssl/certs/test.pem",
   ....
}
```

- [x] README.md is updated

### DIFF
--- a/nat-lab/openwrt.Dockerfile
+++ b/nat-lab/openwrt.Dockerfile
@@ -1,5 +1,5 @@
 FROM openwrt/rootfs:x86-64-v24.10.2
 
-RUN mkdir -p /var/lock && opkg update && opkg install kmod-wireguard && opkg install kmod-tun
+RUN mkdir -p /var/lock && opkg update && opkg install kmod-wireguard kmod-tun curl ca-bundle ca-certificates
 
 COPY --chmod=0755 bin/ /opt/bin/
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -507,6 +507,8 @@ services:
     dns:
       - 10.0.80.82
       - 10.0.80.83
+    volumes:
+      - ./data/core_api/test.pem:/etc/ssl/certs/test.pem
 
   ###########################################################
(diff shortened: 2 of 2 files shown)


nat-lab/openwrt.Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM openwrt/rootfs:x86-64-v24.10.2
 
-RUN mkdir -p /var/lock && opkg update && opkg install kmod-wireguard && opkg install kmod-tun
+RUN mkdir -p /var/lock && opkg update && opkg install kmod-wireguard kmod-tun curl ca-bundle ca-certificates
```
LukasPukenis, line 3: It would be nice to remove the installation of:
- kmod-tun
- kmod-wireguard

since `opkg install nordvpn.ipk` should install those.

With the current changes we are not able to validate the nordvpn Makefile. For example if it would miss dependencies it will just work for us but for the end user it will not since the dependencies are missing from the Makefile and not automatically installed :)

PS:
`opkg update` might be needed before `opkg install nordvpn.ipk`

olekoliinyk: But at the point of docker entrypoint we don't have access to public internet so that we wouldn't be able to download them.

LukasPukenis: Relaying the resolution here from slack: `docker-compose.yml` defines `build` step which will build this `Dockerfile` all the time when running NatLab. This means `RUN` is effectively the same as `ENTRYPOINT` due to the usage.

Though for pre-installing packages, please create a separate ticket blocking the release so we would double-check the dependencies :pray:

LukasPukenis (approved):
+1

PR #1397, opened 2025-08-07
[LLT-6515] Install openwrt vpn package during image creation process

olekoliinyk:
### Problem
We must ensure that all dependencies specified in the Makefile are properly installed. Pre-installing them bypasses the standard installation process. 

### Solution
Install vpn package during image creation process so that all dependencies can be correctly downloaded (in docker entrypoint there is no access to public internet)

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/openwrt_docker_entrypoint.sh
+++ b/nat-lab/bin/openwrt_docker_entrypoint.sh
@@ -4,4 +4,6 @@ set -e
 opkg --version
 uname -a
 
+touch /ready
+
 exec sleep infinity
--- a/nat-lab/bin/openwrt_docker_entrypoint.sh
+++ b/nat-lab/bin/openwrt_docker_entrypoint.sh
@@ -4,6 +4,4 @@ set -e
 opkg --version
 uname -a
 
-opkg install /opt/bin/nordvpn_*_x86_64.ipk
-
 exec sleep infinity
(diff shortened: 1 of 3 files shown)


LukasPukenis (commented):
+1

nat-lab/openwrt.Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM openwrt/rootfs:x86-64-v24.10.2
 
-RUN mkdir -p /var/lock && opkg update && opkg install kmod-wireguard kmod-tun curl ca-bundle ca-certificates
+RUN mkdir -p /var/lock && opkg update && opkg install curl ca-bundle ca-certificates
 
 COPY --chmod=0755 bin/ /opt/bin/
+
+RUN opkg install /opt/bin/nordvpn_*_x86_64.ipk
```
gytsto, line 7: We could also try to add safeguard if .ipk failed is not in `bin/` dir due to user error. Try to copy `nordvpn_*.ipk` from it's original path of `libtelio-build` maybe?

gytsto: nvm, tried to do it myself and it's abit messy :smile:

nat-lab/bin/openwrt_docker_entrypoint.sh
```diff
@@ -4,6 +4,4 @@ set -e
 opkg --version
 uname -a
 
-opkg install /opt/bin/nordvpn_*_x86_64.ipk
-
```
gytsto, line 8: missing `touch /ready` here :smile:

olekoliinyk: Hmm, do we need it? if have `ls /ready` in healthcheck?

olekoliinyk: Ah, nevermind :D Got confused

gytsto (approved):
+1

PR #1401, opened 2025-08-13
Align our proto definition with latest one from RFC

tomaszklak:
Small changes have been made to the proto definition. Instead of using one common `Empty` type, distinct types are used to make it easy in the future to change them.

- [x] README.md is updated


gytsto (commented):
+1

PR #1403, opened 2025-08-13
LLT-6517: Prevent deadlocks in firewall

mathiaspeters:
### Problem
We had a deadlock in the firewall code recently. That particular deadlock was already fixed, but we want to make sure there aren't any other deadlocks lurking

### Solution
Go through all places where we use locks (`Mutex` or `RwLock`) in the firewall code and
1. Reduce the scope of the lock where possible
2. Make sure that there are no I/O or async operations happening while a lock is held

Additionally, there is a clippy lint for checking if there are any await while locks are held. That clippy lint is added to all out crates.


crates/telio-dns/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![deny(unsafe_code)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![deny(missing_docs)]
+#![deny(clippy::await_holding_lock)]
```
stalowyjez: A nice idea to add such lint

gytsto (commented):
+1

crates/telio-firewall/src/firewall.rs
```diff
@@ -398,12 +404,15 @@ impl StatefullFirewall {
             _ => (),
         };
 
-        // whitelist read-lock scope
-        let whitelist = self.whitelist.read();
+        let is_in_whitelist = {
```
stalowyjez: Hmm, there a few ways in which packets can be whitelisted, maybe it would be better to say explicitly that this one is allowed incoming connectinos? Idk `is_allowed_incomming_conns`?

crates/telio-firewall/src/firewall.rs
```diff
@@ -431,39 +440,48 @@ impl StatefullFirewall {
...
-        }
 
-        if !ip.check_valid() {
-            telio_log_trace!("Inbound IP packet is not valid, dropping: {:?}", ip);
-            return Ok(false);
-        }
+            if !ip.check_valid() {
+                telio_log_trace!("Inbound IP packet is not valid, dropping: {:?}", ip);
+                return Ok(false);
+            }
```
stalowyjez: Hmm, I think you don't need whitelist for this ip validation, maybe you could reduce the scope even more by separating VPN whitelist usage from the later uses?

crates/telio-firewall/src/firewall.rs
```diff
@@ -357,10 +357,13 @@ impl StatefullFirewall {
 
         match proto {
             IpNextHeaderProtocols::Udp => {
-                let blacklist = self.outgoing_udp_blacklist.read();
                 let link =
                     Conntrack::build_conn_info(&ip, LibfwDirection::LibfwDirectionOutbound)?.0;
-                if blacklist.contains(&SocketAddr::new(link.remote_addr.into(), link.remote_port)) {
+                let is_in_blacklist = {
+                    let blacklist = self.outgoing_udp_blacklist.read();
```
tomaszklak: nit: why not just `let is_in_blacklist = self.outgoing_udp_blacklist.read().blacklist.contains(&SocketAddr::new(link.remote_addr.into(), link.remote_port)); `? If the new artificial scope is not needed, let's not introduce it needlessly - this will be confusing to someone reading it in the future.

This applies to few other places also.

mathiaspeters: Are you sure the lock will only be held for that one line? It seems to me like it should be, but I'm not 100% sure

crates/telio-firewall/src/firewall.rs
mathiaspeters: I was trying to not change the order of statements unless necessary, but I'm happy to change this order too

crates/telio-crypto/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![deny(unsafe_code)]
 #![deny(missing_docs)]
+#![deny(clippy::await_holding_lock)]
```
tomaszklak: nit: Maybe there is a way to avoid copy-pasting it into all sub crates: https://doc.rust-lang.org/clippy/configuration.html ?

crates/telio-firewall/src/firewall.rs
mathiaspeters: I tried it with `std::sync::Mutex` and it worked well so I'll make the change

tomaszklak: I'm sure, the only tricky case is match, which has scope that extends from the matched expression till the end of the matched arm: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=4b0a56d5a3accb0f4b4c9dd090c3535f

tomaszklak: This is also visible if you peek into MIR representation (I've highlighted guard anonymous variable and dropping of it):
Screenshot 2025-08-14 at 11 34 53

mathiaspeters: Fixed wherever possible

crates/telio-firewall/src/firewall.rs
mathiaspeters: Fixed

crates/telio-firewall/src/firewall.rs
mathiaspeters: I renamed some of these

crates/telio-dns/src/lib.rs
mathiaspeters: And it appears it's covered by default for us so I removed it

crates/telio-crypto/src/lib.rs
mathiaspeters: Apparently we don't need to configure it. I wrote a scenario that should be caught by that lint and clippy caught it without explicitly telling it to

PR #1404, opened 2025-08-13
LLT-6538: Ignore 'Connection reset by peer' during is-alive call during teliod test start

mathiaspeters:
### Problem
We've had teliod tests fail because the `is-alive` call failed with error `Connection reset by peer`. We suspect this is a race condition that can be ignored. Ignored here meaning that since we are checking `is-alive` in a loop until we get `True`, the aforementioned error would mean `is-alive: False`

### Solution
Continue looping `is-alive` if getting error `Connection reset by peer`


Hasan6979 (approved):
+2

PR #1408, opened 2025-08-19
Disable package installation in OpenWrt GW container

olekoliinyk:
### Problem
There were start-up failures of openwrt gw container in nat-lab due to mismatched hashes. 

### Solution
Until we moved to qemu based approach we can disable the openwrt-gw containers entirely. 

- [x] README.md is updated

### DIFF
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -234,14 +234,15 @@ services:
     environment:
       CLIENT_GATEWAY_PRIMARY: 192.168.111.254
 
-  openwrt-client-01:
-    hostname: openwrt-client-01
-    <<: *common-client
-    networks:
-      openwrt-net-01:
-        ipv4_address: 192.168.115.100
-    environment:
-      CLIENT_GATEWAY_PRIMARY: 192.168.115.254
+# Temporary disabled until we moved to QEMU based approach
+#  openwrt-client-01:
+#    hostname: openwrt-client-01
+#    <<: *common-client
+#    networks:
+#      openwrt-net-01:
+#        ipv4_address: 192.168.115.100
+#    environment:
+#      CLIENT_GATEWAY_PRIMARY: 192.168.115.254
 
 
   ###########################################################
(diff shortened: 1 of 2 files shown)


nat-lab/docker-compose.yml
```diff
@@ -479,10 +479,8 @@ services:
 
   openwrt-gw-01:
     hostname: openwrt-gw-01
-    build:
-      context: .
-      dockerfile: openwrt.Dockerfile
-    entrypoint: /opt/bin/openwrt_docker_entrypoint.sh
+    image: openwrt/rootfs:x86-64-v24.10.2
+    entrypoint: sleep infinity
```
LukasPukenis: What would happen if these two lines would also be removed?

olekoliinyk: service "openwrt-gw-01" has neither an image nor a build context specified.
I changed the image to `nat-lab:base` as today it fails in nat-lab to pull this image.

nat-lab/docker-compose.yml
```diff
@@ -479,10 +479,8 @@ services:
 
   openwrt-gw-01:
```
mathiaspeters: Couldn't we just comment out the whole thing so it doesn't start at all? If that's an option, let's add a comment saying that it's disabled until QEMU approach is done, as to not confuse someone who might be going through the file in the meantime

mathiaspeters: This is obviously not a blocker, I'm also fine with the current solution

olekoliinyk: I've commented out all openwrt containers and also updated `network.md`

mathiaspeters: Looks great! We just have to make sure to remember to edit `network.md` again when openwrt is added again

olekoliinyk: We have [check-diagram](https://github.com/NordSecurity/libtelio/actions/runs/17091656289/job/48466542276?pr=1408#logs) job for this :)

LukasPukenis:
The PR description and the contents are different - the code changes disable nodes entirely :thinking: ?

PR #1410, opened 2025-08-19
LLT-6211: Connect to VPN by country

tomasz-grz:
### Problem
Initial VPN supuport in Teliod is currently designed to use a direct endpoint (ip, port, public_key) provided from the config file. This approach is not user-friendly, as end users are unlikely to know valid IPs and public keys of NordVPN servers.

To improve usability, the goal is to allow users to specify a country code (e.g., "US" or "AR") in the VPN config. Teliod will then fetch a suitable VPN server automatically using the NordVPN Core API.

####  Updated config file to support connection by country:
`config::VpnConfig` was converted into an enum with `Server` and `Country` variants.
This enables us to maintain the previous functionality.

Config changes from:
```json
  "vpn": {
    "server_endpoint": "10.0.100.1:1023",
    "server_pubkey": "public-key-placeholder"
  }
```
to
```json
  "vpn": {
    "server": {
      "address": "10.0.100.1",
      "public_key": "public-key-placeholder"
    }
  }
```

Connection by country is done with:
```json
  "vpn": {
    "country": "pl"
  }
```

#### Updated config file to support overriding WG endpoint port:
When fetching VPN server configurations from the Core API, no endpoint port is given, and the default WG port `51820` is assumed.
During natlab testing, we have a non standard port of `1023` for WG servers, therefore the default port needs to be overwritten. For this reason the "hidden" `override_default_wg_port` field was added to the config.

#### Added exit node `hostname` to status report:
To help verify the selected VPN server, the `hostname` field was added to status report.

This field is not present from the call `telio.external_nodes()` (in fact it is only present for meshnet peers fetched by the meshmap, but not peers added with `telio.connect_exit_node()` which is the case with VPN servers). 
Therefore a reply from the API is cached as `maybe_connected_vpn_server: Option<VpnServer>` to fill this data.

Example output of `get-status`:
```json
  "exit_node": {
    "identifier": "identifier-placeholder",
    "public_key": "public-key-placeholder",
    "hostname": "pl100.nordvpn.com",
    "endpoint": "10.0.100.1:1023",
    "state": "connected"
  }
```

- [x] README.md is updated

### DIFF
--- a/clis/teliod/README.md
+++ b/clis/teliod/README.md
@@ -71,9 +71,11 @@ needs be absolute, otherwise will be relative to `working-directory` when daemon
     - `ifconfig` - systems using ifconfig command
     - `iproute` - systems using iproute2 command
     - `uci` - OpenWRT systems using uci command
-  - `vpn` - optional vpn config. If omitted, a VPN connection will not be established
-    - `server_endpoint` - The endpoint (ip:port) of the server/endpoint to connect to
-    - `server_pubkey` - The public key of the server/peer to connect to
+- `vpn` - VPN config type. If omitted, a VPN connection will not be established, please select only one. Possible options:
+  - `server` - manually specified endpoint:
+    - `address` - The IP address of the server/endpoint to connect to
+    - `public_key` - The public key of the server/peer to connect to
+  - `country` - Full country name or ISO A2 code, of the desired VPN server location.
 
 And following cli commands:
 
--- a/clis/teliod/src/cgi/api.rs
+++ b/clis/teliod/src/cgi/api.rs
@@ -430,6 +430,7 @@ mod tests {
                 config_provider: InterfaceConfigurationProvider::Manual,
             },
             vpn: None,
+            override_default_wg_port: None,
             authentication_token: NordToken::new(
                 "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
             )
@@ -540,6 +541,7 @@ mod tests {
                 config_provider: InterfaceConfigurationProvider::Manual,
             },
             vpn: None,
+            override_default_wg_port: None,
             authentication_token: NordToken::new(
                 "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
             )
--- a/clis/teliod/src/config.rs
+++ b/clis/teliod/src/config.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::SocketAddr,
+    net::IpAddr,
     num::NonZeroU64,
     path::{Path, PathBuf},
     str::FromStr,
@@ -12,9 +12,8 @@ use std::fs;
 use tracing::{level_filters::LevelFilter, Level};
 use uuid::Uuid;
 
-use telio::{crypto::PublicKey, device::AdapterType, telio_utils::Hidden};
-
 use crate::{configure_interface::InterfaceConfigurationProvider, TeliodError};
+use telio::{crypto::PublicKey, device::AdapterType, telio_utils::Hidden};
 
 #[derive(PartialEq, Eq, Clone, Copy, Debug, SmartDefault)]
 #[repr(transparent)]
@@ -124,6 +123,11 @@ pub struct TeliodDaemonConfig {
     pub adapter_type: AdapterType,
     pub interface: InterfaceConfig,
     pub vpn: Option<VpnConfig>,
+
+    /// Overrides the default WireGuard endpoint port, should be used only for testing.
+    /// The Core API does not include the WireGuard port in its response, and in environments like
+    /// natlab where a non-standard port is used, this allows specifying a custom port manually.
+    pub override_default_wg_port: Option<u16>,
 
     #[serde(
         deserialize_with = "deserialize_nord_token",
@@ -281,6 +285,7 @@ impl Default for TeliodDaemonConfig {
                 config_provider: Default::default(),
             },
             vpn: None,
+            override_default_wg_port: None,
             authentication_token: Default::default(),
             http_certificate_file_path: None,
             device_identity_file_path: None,
@@ -351,10 +356,23 @@ pub struct InterfaceConfig {
     pub config_provider: InterfaceConfigurationProvider,
 }
 
-#[derive(PartialEq, Eq, Deserialize, Serialize, Debug, Copy, Clone)]
-pub struct VpnConfig {
-    pub server_endpoint: SocketAddr,
-    pub server_pubkey: PublicKey,
+/// Direct VPN server endpoint configuration
+#[derive(PartialEq, Eq, Deserialize, Serialize, Clone, Debug)]
+pub struct DirectEndpointConfig {
+    /// IP of VPN server
+    pub address: IpAddr,
+    /// Public key of VPN server
+    pub public_key: PublicKey,
+}
+
+/// Type of VPN server connection, automatic by country, or specified server endpoint
+#[derive(PartialEq, Eq, Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum VpnConfig {
+    /// Direct endpoint and public key of VPN server
+    Server(DirectEndpointConfig),
+    /// Country name or ISO code of VPN server location
+    Country(String),
 }
 
 #[allow(dead_code)]
@@ -453,6 +471,7 @@ mod tests {
                 config_provider: InterfaceConfigurationProvider::Manual,
             },
             vpn: None,
+            override_default_wg_port: None,
             authentication_token: NordToken::new(
                 "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
             )
@@ -497,6 +516,150 @@ mod tests {
 
             assert_eq!(expected, serde_json::from_str(json).unwrap());
         }
+    }
+
+    #[test]
+    fn test_config_example() {
+        let example = TeliodDaemonConfig::from_file("example_teliod_config.json").unwrap();
+
+        assert_eq!(example.adapter_type, AdapterType::NepTUN);
+    }
+
+    #[test]
+    fn test_config_vpn_country() {
+        let expected_config = TeliodDaemonConfig {
+            vpn: Some(VpnConfig::Country("de".to_string())),
+            ..TeliodDaemonConfig::default()
+        };
+
+        let json = r#"{
+            "log_level": "Trace",
+            "log_file_path": "/var/log/teliod_lib.log",
+            "adapter_type": "neptun",
+            "interface": {
+                "name": "nlx",
+                "config_provider": "manual"
+            },
+            "vpn": {
+                "country": "de"
+            },
+            "authentication_token": ""
+            }"#;
+
+        assert_eq!(expected_config, serde_json::from_str(json).unwrap());
+    }
+
+    #[test]
+    fn test_config_vpn_endpoint() {
+        let expected_config = TeliodDaemonConfig {
+            vpn: Some(VpnConfig::Server(DirectEndpointConfig {
+                address: "127.0.0.1".parse().unwrap(),
+                public_key: "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
+                    .parse()
+                    .unwrap(),
+            })),
+            ..TeliodDaemonConfig::default()
+        };
+
+        let json = r#"{
+            "log_level": "Trace",
+            "log_file_path": "/var/log/teliod_lib.log",
+            "adapter_type": "neptun",
+            "interface": {
+                "name": "nlx",
+                "config_provider": "manual"
+            },
+            "vpn": {
+                "server": {
+                    "address": "127.0.0.1",
+                    "public_key": "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
+                }
+            },
+            "authentication_token": ""
+            }"#;
+
+        assert_eq!(expected_config, serde_json::from_str(json).unwrap());
+    }
+
+    #[test]
+    fn test_config_vpn_fails() {
+        let json = r#"{
+            "log_level": "Trace",
+            "log_file_path": "/var/log/teliod_lib.log",
+            "adapter_type": "neptun",
+            "interface": {
+                "name": "nlx",
+                "config_provider": "manual"
+            },
+            "vpn": {
+                "country": "de",
+                "country": "pl",
+            },
+            "authentication_token": ""
+            }"#;
+
+        assert!(serde_json::from_str::<TeliodDaemonConfig>(json).is_err());
+
+        let json = r#"{
+            "log_level": "Trace",
+            "log_file_path": "/var/log/teliod_lib.log",
+            "adapter_type": "neptun",
+            "interface": {
+                "name": "nlx",
+                "config_provider": "manual"
+            },
+            "vpn": {
+                "country": "de",
+            },
+            "vpn": {
+                "country": "pl",
+            },
+            "authentication_token": ""
+            }"#;
+
+        assert!(serde_json::from_str::<TeliodDaemonConfig>(json).is_err());
+
+        let json = r#"{
+            "log_level": "Trace",
+            "log_file_path": "/var/log/teliod_lib.log",
+            "adapter_type": "neptun",
+            "interface": {
+                "name": "nlx",
+                "config_provider": "manual"
+            },
+            "vpn": {
+                "country": "de",
+            },
+            "vpn": {
+                "server": {
+                    "address": "127.0.0.1",
+                    "public_key": "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
+                }
+            },
+            "authentication_token": ""
+            }"#;
+
+        assert!(serde_json::from_str::<TeliodDaemonConfig>(json).is_err());
+
+        let json = r#"{
+            "log_level": "Trace",
+            "log_file_path": "/var/log/teliod_lib.log",
+            "adapter_type": "neptun",
+            "interface": {
+                "name": "nlx",
+                "config_provider": "manual"
+            },
+            "vpn": {
+                "country": "de",
+                "server": {
+                    "address": "127.0.0.1",
+                    "public_key": "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
+                }
+            },
+            "authentication_token": ""
+            }"#;
+
+        assert!(serde_json::from_str::<TeliodDaemonConfig>(json).is_err());
     }
 
     #[test]
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -128,6 +128,7 @@ impl Country {
 #[derive(Debug, PartialEq, Deserialize)]
 pub struct Server {
     station: IpAddr,
+    hostname: Option<String>,
     technologies: Vec<Technology>,
 }
 
@@ -149,6 +150,11 @@ impl Server {
     /// Get IP address of server
     pub fn address(&self) -> IpAddr {
         self.station
+    }
+
+    /// Get hostname of VPN server
+    pub fn hostname(&self) -> Option<String> {
+        self.hostname.to_owned()
     }
 
     /// Get public_key of wireguard service
@@ -878,5 +884,6 @@ mod tests {
                 .parse()
                 .unwrap()
         );
+        assert_eq!(exit_node.hostname, Some("test100.domain.com".to_string()));
     }
 }
--- a/clis/teliod/src/core_api.rs
+++ b/clis/teliod/src/core_api.rs
@@ -2,8 +2,10 @@ use base64::{prelude::*, DecodeError};
 use reqwest::{header, Certificate, Client, StatusCode};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::fmt::Display;
 use std::fs::{create_dir_all, File, OpenOptions};
 use std::io::Read;
+use std::net::IpAddr;
 use std::os::unix::fs::OpenOptionsExt;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, LazyLock};
@@ -25,9 +27,11 @@ const API_BASE: &str = "https://api.nordvpn.com/v1";
 const OS_NAME: &str = "macos";
 #[cfg(target_os = "linux")]
 const OS_NAME: &str = "linux";
+const WIREGUARD_ID: u64 = 35;
 
 type MachineIdentifier = String;
 const MAX_RETRIES: usize = 10;
+const REQUEST_TIMEOUT_SECONDS: u64 = 30;
 
 static DEVICE_IDENTITY_FILE_PATH: LazyLock<Result<PathBuf, Error>> = LazyLock::new(|| {
     dirs::data_local_dir()
@@ -52,7 +56,7 @@ pub enum Error {
     PeerRegistering(StatusCode),
     #[error(transparent)]
     Decode(#[from] DecodeError),
-    #[error("Invalid response recieved from server")]
+    #[error("Invalid response received from server")]
     InvalidResponse,
     #[error("Unable to update machine due to Error: {0}")]
     UpdateMachine(StatusCode),
@@ -68,6 +72,95 @@ pub enum Error {
     BackoffBounds(#[from] BackoffError),
     #[error("Failed to access device identity file")]
     DeviceIdentityFileAccess,
+    #[error("API responded with error: {0}")]
+    ApiErrorResponse(ErrorDetail),
+}
+
+/// Error response returned by the Core API in case of bad requests
+///
+/// For example:
+/// ```json
+/// {
+///  "errors": {
+///    "code": 101101,
+///    "message": "Invalid form data: {...}"
+///  }
+/// }
+/// ```
+#[derive(Debug, Deserialize)]
+struct ErrorResponse {
+    pub errors: ErrorDetail,
+}
+
+/// Error details
+#[derive(Debug, Deserialize)]
+pub struct ErrorDetail {
+    pub code: u32,
+    pub message: String,
+}
+
+impl Display for ErrorDetail {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}: {}", self.code, self.message)
+    }
+}
+
+/// Country description returned by the Core API
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct Country {
+    /// Internal country ID
+    pub id: u64,
+    /// Full country name
+    pub name: String,
+    /// Country ISO A2 code
+    pub code: String,
+}
+
+impl Country {
+    /// Check if the given &str is either full name or ISO code of the country
+    pub fn matches(&self, target: &str) -> bool {
+        let target = target.trim().to_lowercase();
+        self.name.trim().to_lowercase() == target || self.code.trim().to_lowercase() == target
+    }
+}
+
+/// VPN Server description returned by the Core API
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct Server {
+    station: IpAddr,
+    technologies: Vec<Technology>,
+}
+
+/// VPN Server technology stack
+#[derive(Debug, PartialEq, Deserialize)]
+struct Technology {
+    id: u64,
+    metadata: Vec<Metadata>,
+}
+
+/// VPN Server technology stack metadata
+#[derive(Debug, PartialEq, Deserialize)]
+struct Metadata {
+    name: String,
+    value: String,
+}
+
+impl Server {
+    /// Get IP address of server
+    pub fn address(&self) -> IpAddr {
+        self.station
+    }
+
+    /// Get public_key of wireguard service
+    pub fn wg_public_key(&self) -> Option<String> {
+        self.technologies
+            .iter()
+            .filter(|t| t.id == WIREGUARD_ID)
+            .flat_map(|t| &t.metadata)
+            .filter(|m| m.name == "public_key")
+            .map(|m| m.value.to_owned())
+            .next()
+    }
 }
 
 #[derive(Default, Serialize, Deserialize)]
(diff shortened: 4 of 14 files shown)


clis/teliod/src/config.rs
```diff
@@ -351,10 +355,23 @@ pub struct InterfaceConfig {
...
+}
+
+/// Type of VPN server connection, automatic by country, or specified server endpoint
+#[derive(PartialEq, Eq, Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum VpnConfig {
+    /// Direct endpoint and public key of VPN server
+    Server(DirectEndpointConfig),
+    /// Country name or ISO code of VPN server location
+    Country(String),
```
tomasz-grz, line 375: This can be later extended with `City(String)` etc

clis/teliod/src/daemon.rs
```diff
@@ -148,26 +199,46 @@ fn telio_task(
                     // update the interface only if the address is new
                     if state.interface_ip_address != Some(new_ip_address) {
                         state.interface_ip_address = Some(new_ip_address);
-                        _ = sys_config
-                            .set_ip(&new_ip_address)
-                            .inspect_err(|e| error!("Failed to set IP with error '{e:?}'"));
-                        task_connect_to_exit_node(config.vpn, tx_channel.clone());
```
tomasz-grz, line 154: Here we used to still try connecting to exit_node even though `set_ip` failed.

clis/teliod/src/config.rs
```diff
@@ -125,6 +124,10 @@ pub struct TeliodDaemonConfig {
     pub interface: InterfaceConfig,
     pub vpn: Option<VpnConfig>,
 
+    /// Override the hardcoded WG endpoint port,
+    /// used when fetching servers from the Core API
+    pub override_vpn_endpoint_port: Option<u16>,
```
stalowyjez: Maybe that could be a part of `vpn` config field?

stalowyjez: (and maybe it could be just set there, not as an option and just set by default to `DEFAULT_PORT`? idk)

clis/teliod/src/config.rs
```diff
@@ -499,6 +517,67 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_config_example() {
+        let _ = TeliodDaemonConfig::from_file("example_teliod_config.json").unwrap();
```
stalowyjez: Does it check only if it doesn't fail? Maybe some additional asserts would be helpful?

clis/teliod/src/core_api.rs
```diff
@@ -68,6 +72,102 @@ pub enum Error {
...
+                t.metadata.iter().find_map(|m| {
+                    if m.name == "public_key" {
+                        Some(m.value.to_owned())
+                    } else {
+                        None
+                    }
+                })
+            } else {
+                None
+            }
```
stalowyjez: Nit: I think separate filter/map could be actually a bit cleaner, without this `else` tail, but that's just my preference

clis/teliod/src/core_api.rs
```diff
@@ -505,3 +686,98 @@ async fn register_machine(
...
+            [
+                {"id":1,"name":"Afghanistan","code":"AF"},
+                {"id":2,"name":"Albania","code":"AL"},
+                {"id":3,"name":"Algeria","code":"DZ"}
+            ]
+        "#;
+
+        let countries: Vec<Country> = serde_json::from_str(json).unwrap();
+        assert_eq!(countries.len(), 3);
+        assert_eq!(countries[0].code, "AF".to_string());
```
stalowyjez: I guess we can add all of the asserts here? Errors like having 3 Afghanistans in the vector after parsing happens from time to time.

clis/teliod/src/core_api.rs
```diff
@@ -505,3 +686,98 @@ async fn register_machine(
...
+            .first()
+            .and_then(|s| ExitNodeConfig::from_api(s, None).ok())
+            .unwrap();
+        assert_eq!(server.endpoint, "127.0.0.1:51820".parse().unwrap());
+        assert_eq!(
+            server.public_key,
+            "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
+                .parse()
+                .unwrap()
+        );
```
stalowyjez: Same here, I think if we do such test we can assert a bit more

clis/teliod/src/core_api.rs
```diff
@@ -68,6 +72,102 @@ pub enum Error {
     BackoffBounds(#[from] BackoffError),
     #[error("Failed to access device identity file")]
     DeviceIdentityFileAccess,
+    #[error("API responded with error: {0}")]
+    ApiErrorResponse(ErrorDetail),
+}
+
+/// Error responded by the API
+#[derive(Debug, Deserialize)]
+pub struct ErrorResponse {
```
stalowyjez: Does it need to be public?

clis/teliod/src/daemon.rs
```diff
@@ -55,6 +56,45 @@ const EMPTY_TOKEN: &str = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
...
+
+impl ExitNodeConfig {
+    /// Convert Server from API response to ExitNode endpoint configuration
+    /// Optionally override the default WG port
+    pub fn from_api(s: &CoreApiServer, override_port: Option<u16>) -> Result<Self, TeliodError> {
+        Ok(Self {
+            endpoint: s.wg_endpoint(override_port),
+            public_key: s
+                .wg_public_key()
+                .and_then(|s| s.parse().ok())
```
stalowyjez: hmm, I know that's just for a dumb lambda, but why is `s` a good name for a public key?

clis/teliod/src/daemon.rs
```diff
@@ -128,10 +164,25 @@ fn telio_task(
...
+                    // fill it with the information fetched from the API
+                    if let (Some(exit_node), Some(connected_server)) =
+                        (&mut exit_node, &state.maybe_connected_vpn_server)
+                    {
+                        // make sure it's the same node by comparing keys
+                        if exit_node.public_key == connected_server.public_key {
+                            // if the hostname is missing, fill it in
+                            if let (None, Some(hostname)) =
+                                (&exit_node.hostname, &connected_server.hostname)
+                            {
```
stalowyjez: You can just check exit_node.hostname.is_none(), it won't be worse in this case, and maybe `&&` it to the outer `if`?

nat-lab/data/teliod/config_with_vpn_country.json
```diff
@@ -0,0 +1,17 @@
...
+  "authentication_token": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
+  "adapter_type": "neptun",
+  "http_certificate_file_path": "/etc/ssl/server_certificate/test.pem",
+  "device_identity_file_path": "/etc/teliod/data.json",
+  "interface": {
+    "name": "teliod",
+    "config_provider": "iproute"
+  },
+  "vpn": {
+    "country": "pl"
```
stalowyjez: <3

clis/teliod/src/daemon.rs
```diff
@@ -213,6 +284,82 @@ fn telio_task(
...
+                if country_id.is_none() {
+                    warn!("country_id not found for: {target_country}");
+                }
+
+                // Fetch the list of recommended server from the API
+                // if no country_id is found, as a fallback the API will still return a
+                // recommended server based on other metrics
+                match get_recommended_servers(country_id, Some(1), &auth_token, &cert_path).await {
+                    Ok(servers) => {
+                        // TODO: We can store the recommended server list, just in case
```
stalowyjez: Sounds sane, it would be nice to have some associated JIRA item for that imho

nat-lab/tests/test_core_api.py
```diff
@@ -598,17 +600,9 @@ async def test_get_nonexisting_servers():
...
-            f"{CORE_API_URL}/test/public-key",
-            "POST",
-            CORE_API_CA_CERTIFICATE_PATH,
-            data=payload,
-            authorization_header=BEARER_AUTHORIZATION_HEADER,
-            expect_response=False,
-        )
+        # note that there are only 2 country_id servers hardcoded,
+        # even if we register one for any other, this test will fail
+        await register_vpn_server_key(connection, linux_vm_public_key, 5)
```
stalowyjez: maybe let's make some constant for this country id

stalowyjez (commented):
Looks good, only a few minor comments

clis/teliod/src/cgi/api.rs
```diff
@@ -430,6 +430,7 @@ mod tests {
                 config_provider: InterfaceConfigurationProvider::Manual,
             },
             vpn: None,
+            override_vpn_endpoint_port: None,
```
LukasPukenis: Minor: maybe just `port` with a comment about the default value?

clis/teliod/src/config.rs
```diff
@@ -125,6 +124,10 @@ pub struct TeliodDaemonConfig {
     pub interface: InterfaceConfig,
     pub vpn: Option<VpnConfig>,
 
+    /// Override the hardcoded WG endpoint port,
```
LukasPukenis: Is the comment correct? I read it as if the port is used in fetching the servers but it also mentions that it's `WG endpoint port` :thinking:

clis/teliod/src/core_api.rs
```diff
@@ -68,6 +72,102 @@ pub enum Error {
...
+    /// Country ISO A2 code
+    pub code: String,
+}
+
+impl Country {
+    /// Check if country name or code matches the target
+    pub fn matches(&self, target: &str) -> bool {
+        let target = target.trim().to_lowercase();
+        self.name.trim().to_lowercase() == target || self.code.trim().to_lowercase() == target
+    }
```
LukasPukenis, line 124: Very minor, highlighting: Rust way would be to probably implement the `PartialEq` trait for `Country`

clis/teliod/README.md
```diff
@@ -72,8 +72,10 @@ needs be absolute, otherwise will be relative to `working-directory` when daemon
     - `iproute` - systems using iproute2 command
     - `uci` - OpenWRT systems using uci command
   - `vpn` - optional vpn config. If omitted, a VPN connection will not be established
-    - `server_endpoint` - The endpoint (ip:port) of the server/endpoint to connect to
-    - `server_pubkey` - The public key of the server/peer to connect to
+    - `server` - manually specified endpoint:
+      - `endpoint` - The endpoint (ip:port) of the server/endpoint to connect to
+      - `public_key` - The public key of the server/peer to connect to
+    - `country` - Full country name or ISO A2 code, of the desired VPN server location.
```
LukasPukenis: Is it mentioned anywhere what's the priority if both the `country` and `server` are present?

clis/teliod/src/core_api.rs
```diff
@@ -68,6 +72,102 @@ pub enum Error {
...
     #[error("Failed to access device identity file")]
     DeviceIdentityFileAccess,
+    #[error("API responded with error: {0}")]
+    ApiErrorResponse(ErrorDetail),
+}
+
+/// Error responded by the API
+#[derive(Debug, Deserialize)]
+pub struct ErrorResponse {
+    pub errors: ErrorDetail,
```
LukasPukenis, line 92: The name is plural though it stores only a single instance of `ErrorDetail`

clis/teliod/src/core_api.rs
```diff
@@ -466,6 +566,87 @@ pub async fn get_meshmap(
...
+/// Get all known countries list from core API
+/// used to find the internal country ID for filtering VPN servers
+pub async fn get_countries(
+    auth_token: &str,
+    cert_path: &Option<PathBuf>,
+) -> Result<Vec<Country>, Error> {
+    debug!("Getting countries list");
+    let client = http_client(cert_path);
+    let response = client
+        .get(format!("{}/countries", API_BASE))
```
LukasPukenis, line 604: I wonder - is there a default timeout on the client or is this call somehow managed by a timeout at call site(I couldn't find it)? This applies to `/servers` as well. If there's no timeout it might hang forever :)

clis/teliod/src/daemon.rs
```diff
@@ -55,6 +56,45 @@ const EMPTY_TOKEN: &str = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 #[derive(Debug, Default)]
 pub struct TelioTaskStates {
     interface_ip_address: Option<IpAddr>,
+    maybe_connected_vpn_server: Option<ExitNodeConfig>,
```
LukasPukenis, line 64: The `Option` part of the type already is enough, I suggest renaming the member to `connected_vpn_server` or just `vpn_server` :)

clis/teliod/src/daemon.rs
```diff
@@ -213,6 +284,82 @@ fn telio_task(
...
 
+/// Handles establishing a connection to a VPN exit node based on the given configuration.
+///
+/// If a specific server is provided, it uses that directly.
+/// If a country is provided, it queries the API to find a recommended server.
+/// Sends a `ConnectToExitNode` command via the provided channel on success.
+#[allow(mpsc_blocking_send)]
+fn handle_exit_node_connection(
+    vpn_config: VpnConfig,
+    tx_channel: mpsc::Sender<TelioTaskCmd>,
```
LukasPukenis, line 301: Maybe `tx_channel` -> `cmd_chan` ?

clis/teliod/src/config.rs
LukasPukenis, line 375: Since `VpnConfig` is now an enum, how is it deserialized if both the `Server` and `Country` are present? I do not see explicit code to handle that so I suppose it's somehow automatically handled by `serde`? Could you provide some information on this? Thanks :pray:

nat-lab/tests/test_teliod.py
```diff
@@ -164,6 +183,32 @@ async def register_new_device_on_api():
...
+
+        async with teliod.start():
+            log.debug("Teliod started, waiting for connected vpn state...")
+            await teliod.wait_for_vpn_connected_state()
+
+            await ping(connection, PHOTO_ALBUM_IP)
+            ip = await stun.get(connection, STUN_SERVER)
+            assert (
+                ip == WG_SERVER["ipv4"]
+            ), f"wrong public IP when connected to VPN {ip}"
```
LukasPukenis: Nice test however could we also add a second country - "de" as well and check that it really selects `WG_SERVER2`? Since these use mocked APIs it would be nice to have at least two checks so it wouldn't hide an error in the mocked endpoint :)

Additionally a question - is `hostname` visible somewhere and maybe should be checked for as well?

clis/teliod/src/daemon.rs
```diff
@@ -128,10 +164,25 @@ fn telio_task(
...
+                    if let (Some(exit_node), Some(connected_server)) =
+                        (&mut exit_node, &state.maybe_connected_vpn_server)
+                    {
+                        // make sure it's the same node by comparing keys
+                        if exit_node.public_key == connected_server.public_key {
+                            // if the hostname is missing, fill it in
+                            if let (None, Some(hostname)) =
+                                (&exit_node.hostname, &connected_server.hostname)
+                            {
+                                exit_node.hostname = Some(hostname.to_owned());
```
LukasPukenis: To me this code is slightly confusing - basically there's `exit_node` and `connected_server` - do we need both? It looks like just one should suffice and would be a single source of truth :)

clis/teliod/src/cgi/api.rs
Jauler, line 433: IMHO such option should not exist, why not just specify `1.2.3.4` for default port and `1.2.3.4:12345` for overriden port?

clis/teliod/src/config.rs
Jauler: IMHO such option should not exist, why not just specify 1.2.3.4 for default port and 1.2.3.4:12345 for overriden port?

If not, then I would vote for not having default at all :thinking:

.unreleased/teliod_vpn_country
Jauler, line 1: :thinking: It would be nice to prepend jira id to the branch name and `.unreleased` file. e.g. `LLT-6211_teliod_vpn_country`.

clis/teliod/src/core_api.rs
```diff
@@ -68,6 +72,102 @@ pub enum Error {
...
+#[derive(Debug, Deserialize)]
+pub struct Server {
+    station: IpAddr,
+    hostname: Option<String>,
+    technologies: Vec<Tech>,
+}
+
+/// VPN Server technologies stack
+#[derive(Debug, Deserialize)]
+struct Tech {
```
Jauler: Nit: maybe it is worth using full name in the struct name which is part of the model? 

But this is subjective, you may decide to change it or you may decide to leave it and mark this comment as resolved.

clis/teliod/src/core_api.rs
```diff
@@ -466,6 +566,87 @@ pub async fn get_meshmap(
...
 }
 
+/// Get all known countries list from core API
+/// used to find the internal country ID for filtering VPN servers
+pub async fn get_countries(
+    auth_token: &str,
+    cert_path: &Option<PathBuf>,
+) -> Result<Vec<Country>, Error> {
+    debug!("Getting countries list");
+    let client = http_client(cert_path);
```
Jauler, line 602: In the long term - we have to decide whether PDP-based pinning is a must for MVP.

nat-lab/tests/test_core_api.py
```diff
@@ -559,17 +571,7 @@ async def test_get_servers_with_filters():
...
-        await send_https_request(
-            connection,
-            f"{CORE_API_URL}/test/public-key",
-            "POST",
-            CORE_API_CA_CERTIFICATE_PATH,
-            data=payload,
-            authorization_header=BEARER_AUTHORIZATION_HEADER,
-            expect_response=False,
-        )
+        await register_vpn_server_key(connection, linux_vm_public_key, 2)
```
Jauler: what is this magic number "2"?

clis/teliod/src/daemon.rs
```diff
@@ -148,26 +199,46 @@ fn telio_task(
...
                     if state.interface_ip_address != Some(new_ip_address) {
                         state.interface_ip_address = Some(new_ip_address);
-                        _ = sys_config
-                            .set_ip(&new_ip_address)
-                            .inspect_err(|e| error!("Failed to set IP with error '{e:?}'"));
-                        task_connect_to_exit_node(config.vpn, tx_channel.clone());
+                        match sys_config.set_ip(&new_ip_address) {
+                            Ok(_) => {
+                                if let Some(vpn_config) = config.vpn.as_ref() {
+                                    handle_exit_node_connection(
```
Jauler, line 208: I am a bit confused - why do we use `SetIp` command to connect to VPN?

clis/teliod/src/config.rs
tomasz-grz: I thought about it... But then it would be rather ugly from user perspective,

Since it's only used for natlab testing, user should never actually interact with it

That would make VPN config from the simple:
```json
  "vpn": {
    "country": "de"
  }
```

To something like this, to accommodate for the named struct variant :
```json
  "vpn": {
    "country": {
      "name": "de"
    }
  }
```

With the optional port override:
```json
  "vpn": {
    "country": {
      "name": "de",
      "override_vpn_endpoint_port": 1023 
    }
  }
```

If it's implemented as an unnamed tuple, then it's even less intuitive:
```json
  "vpn": {
    "country": ["de", null]
  }
```
or
```json
  "vpn": {
    "country": ["de", 1023]
  }
```

tomasz-grz: I also agree such option should not exist, but unfortunately Core API does not respond with what the WG endpoint port is, and in natlab we use non standard ports.

This is only used for servers fetched from Core API during testing. Manually specified endpoints are in the `1.2.3.4:12345` format

clis/teliod/src/config.rs
tomasz-grz: This indeed overrides the WG endpoint port, but I will try to rephrase it better.

clis/teliod/README.md
tomasz-grz: That would be an invalid schema, but good point I will mention it.

.unreleased/teliod_vpn_country
tomasz-grz: Good point! Updated

clis/teliod/src/config.rs
tomasz-grz: Since that would be an invalid JSON schema, de-serializing with serde would fail.

for example 
```json
            "vpn": {
                "country": "de",
                "server": {
                    "endpoint": "127.0.0.1:51820",
                    "public_key": "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
                }
            },
```

Serde fails with:
```
Error("expected value", line: 10, column: 31)
```

```json
            "vpn": {
                "country": "de"
            },
            "vpn": {
                "server": {
                    "endpoint": "127.0.0.1:51820",
                    "public_key": "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
                }
            },
```
Serde fails with:
```
Error("duplicate field `vpn`", line: 12, column: 17)
```

I can add both cases to tests

clis/teliod/src/config.rs
tomasz-grz: The main motivation for this is to make sure that `example_teliod_config.json` is kept valid and up to date.

I personally don't really believe it will benefit us to check the values it has, since we would have to maintain both the config and the test if we change something, but I added a simple assert for `adapter_type`

clis/teliod/src/core_api.rs
tomasz-grz: It's a good point but that's the JSON schema returned by Core API

```json
{
  "errors": {
    "code": 101101,
    "message": "Invalid form data: {...}"
  }
}
```

I could add a `#[serde(rename = "errors")]`, but felt that it's too much.

clis/teliod/src/core_api.rs
tomasz-grz: Good point

clis/teliod/src/core_api.rs
tomasz-grz: That's a good point, changed!

clis/teliod/src/core_api.rs
tomasz-grz: Hmm it does happen? I am not sure how much it is our responsibility to validate Core API data.

Plus even if we add asserts for duplicates here, doesn't mean what comes from the actual Core API response is valid.. We could add some sort of validation logic but IMHO it would be an overkill, if getting Country ID fails, we simply take a recommended server without the country filter

clis/teliod/src/core_api.rs
tomasz-grz: Added some one more assertion for `hostname`.
We only have 3 fields in `ExitNodeConfig`, `endpoint`, `public_key` and `hostname`.

The main motivation for this test is to verify that our Struct definition is correct, and that the `from_api` method works.

I don't see the value of adding more assertion on the actual values, after deserializing, as that only tests Serde, I think that is the responsibility of Serde developers.

clis/teliod/src/daemon.rs
tomasz-grz: We can have `Some<ExitNodeConfig>` set by the API response, but we are not actually connected until `MaybeSetupVPNRoutes` is successfully completed, which would be misleading.

clis/teliod/src/daemon.rs
tomasz-grz: Good point, thanks for pointing that out

clis/teliod/src/daemon.rs
tomasz-grz: Unfortunately we don't have a single source of truth.

`exit_node` is retrieved from `libtelio` it self, and contains the actual state of connection, but does not contain `hostname` unless it's a meshnet peer...

`maybe_connected_vpn_server` signifies only the intention of the server we want to connect, and the data received from the Core API, but this includes `hostname`.

How would you recommend handling this case?

I can simply remove `hostname` from the status report, then we don't have to keep track of `maybe_connected_vpn_server` at all.
But I thought this was a helpful feature.

nat-lab/tests/test_teliod.py
tomasz-grz: Good point, I'll try to quickly add that.

clis/teliod/src/daemon.rs
tomasz-grz: `tx_channel` is consistent with the rest of the handlers.

Would you say the explicit type definition `mpsc::Sender<TelioTaskCmd>`is not obvious enough?

clis/teliod/src/core_api.rs
tomasz-grz: Nope, good eyes

clis/teliod/src/core_api.rs
tomasz-grz: Good point, I can wrap it into a Exponential Backoff.

nat-lab/tests/test_core_api.py
tomasz-grz: Added consts for magic numbers

clis/teliod/src/daemon.rs
tomasz-grz: To configure the tunnel interface address, and set it up.

clis/teliod/src/core_api.rs
tomasz-grz: I believe this is only used for natlab testing, since we have a mocked Core API

clis/teliod/src/core_api.rs
LukasPukenis: Thanks for explaining. A simple comment should suffice in such case :)

clis/teliod/src/daemon.rs
LukasPukenis: The name is obvious only if I look into the type declaration first :) However if it is consistent with already existing code, it's better to stay consistent then

clis/teliod/src/config.rs
LukasPukenis: Thanks for explaining. Your example includes duplicate key which results in error but I wonder about the case where the key is not duplicated:
```
"vpn": {
                "country": "de",
                "server": {
                    "endpoint": "127.0.0.1:51820",
                    "public_key": "urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6uro="
                }
            },
```

clis/teliod/src/daemon.rs
LukasPukenis: Thanks for taking your time to explain. I can think of an edge case where `connected_server` is pointing to another server already, while the `exit_node` still points to the previous one. 


Since we grab the peers from WireGuard directly, it means it's prone to have a small race condition unless we would lock the mutex on the whole thing (so we wouldn't be able to switch the server while processing `external_nodes()`). Even that might not help since it's not guaranteed(my guess) that WireGuard will process the nodes immediately so for example setting the new `exit_node` might not propagate immediately to WireGuard while the `connected_server` will already be updated.

I must say that having a `hostname` in the status report is a very nice thing to have so I would prefer to leave it. As for the solution, I do not have a good offer except my doubts above :)

clis/teliod/src/config.rs
tomasz-grz: That's the first example, Serde fails with
`Error("expected value", line: 10, column: 31)`

clis/teliod/src/daemon.rs
tomasz-grz: That is exactly it's called `maybe_connected_vpn_server` 😅 and we are checking if the public keys of both are matching.

At the moment there is no possible way to switch `exit_node` during runtime.

tomasz-grz: We can of course also have `verified_connected_vpn_server` in addition to `maybe_connected_vpn_server`, and move `maybe_connected_vpn_server` into `verified_connected_vpn_server` once the connection was verified in `MaybeSetupVPNRoutes()`

but I think it over-complicates it just for displaying a `hostname` to the user..

LukasPukenis: Agreed on the overcomplication, I think it's fine as is. Thanks for elaborating :)

LukasPukenis (commented):
+0.5

clis/teliod/src/daemon.rs
tomasz-grz: Refactored this bit

clis/teliod/src/daemon.rs
tomasz-grz: Added jira number

nat-lab/tests/test_core_api.py
tomasz-grz: Turned into constant

clis/teliod/src/config.rs
tomasz-grz: Added a better comment

clis/teliod/src/cgi/api.rs
tomasz-grz: Renamed to `override_default_wg_port`, since it's a global parameter, it would be very cryptic what kind of `port` it is.

clis/teliod/src/core_api.rs
tomasz-grz: Converted to `PartialEq`

clis/teliod/README.md
tomasz-grz: Updated docs

clis/teliod/src/core_api.rs
tomasz-grz: Added better comment

nat-lab/tests/test_teliod.py
tomasz-grz: Added test for second country and hostnames

clis/teliod/src/core_api.rs
tomasz-grz: Rewrote this

clis/teliod/src/core_api.rs
```diff
@@ -505,3 +741,103 @@ async fn register_machine(
...
+                {"id":3,"name":"Algeria","code":"DZ"}
+            ]
+        "#;
+
+        let countries: Vec<Country> = serde_json::from_str(json).unwrap();
+        assert_eq!(countries.len(), 3);
+        assert_eq!(countries[0].code, "AF".to_string());
+
+        assert!(countries[0].eq("af"));
+        assert!(countries[0].eq("afghanistan"));
```
stalowyjez: Maybe this comparison is a bit strange, i.e. we have kinda "af" == countries[0] == "afghanistan",
Maybe defining it on countries would make more sense, i.e. Country::from_str("af") == Country::from_str("afghanistan") makes more sense to me, what do you think?

clis/teliod/src/daemon.rs
Jauler: yeah, but I would expect it to be vice-versa, `connect-to-vpn` or some sort of similar command would be dependant on SetIp (e.g. calling it at some stage), not vice-versa. 

Sounds a bit unintuitive that one needs to call `SetIp` to connect to VPN.

nat-lab/tests/test_core_api.py
Jauler: :pray:

clis/teliod/src/daemon.rs
stalowyjez: It can be like that

clis/teliod/src/core_api.rs
stalowyjez: I meant to add them only to check the parsing validation - because about the parsing this test is, isn't it?

clis/teliod/src/core_api.rs
stalowyjez: > I don't see the value of adding more assertion on the actual values, after deserializing, as that only tests Serde, I think that is the responsibility of Serde developers.

That's kinda true, but we also affect how it behaves, have some default values etc. so that happens that you have some typo in your serde defs, then you overwrite values with some defaults etc.

clis/teliod/src/core_api.rs
Jauler: I mean... `teliod` does go to API to get server list/countries list/etc. and will do that in prod, therefore pinning is on the table :P

clis/teliod/src/daemon.rs
```diff
@@ -148,26 +202,49 @@ fn telio_task(
...
+                        endpoint: Some(SocketAddr::new(server.address, port)),
                     };
                     match telio.connect_exit_node(&node) {
                         Ok(_) => {
                             // routing for VPN should only be setup once the VPN connection has been established
                             // which could take some amount of time, yet we don't want to block the event loop
                             // this task checks the connection status and if connected sets up routing information
                             // otherwise it sleeps and tries again
-                            task_maybe_setup_vpn_routes(tx_channel.clone(), ip.ip(), pubkey);
+                            task_maybe_setup_vpn_routes(
```
Jauler, line 239: In OpenWRT's case - isn't routing handled by the OS itself? I believe we will probably mess-up a lot of configurations on the GW if we will not adhere to `UCI` :thinking:

clis/teliod/src/config.rs
stalowyjez: I think it's worth it - here you have some state of the VPN connection, but relevant only for country config. Imho you could even have three states here:
 - WaitingForCountryList(Str)
 - WaitingForPort(Country)
 - Ready(Country, Port)
 then you could have a custom parser which returns Country::WaitingForCountryList(str) and you could parse it from `country: "de"` WDYT?

clis/teliod/src/daemon.rs
tomasz-grz: We still need to setup the routing rules, firewall zones, etc. We are adhering to [UCI commands here](https://github.com/NordSecurity/libtelio/blob/59e934855921418525161e923697c705ec50217f/clis/teliod/src/configure_interface.rs#L597)

This PR doesn't touch at all any part of the setup, do you see any potential issues?

clis/teliod/src/config.rs
tomasz-grz: I am not sure what you mean here, could you please elaborate a little bit more?

This is not the VPN connection state, but the configuration file given to teliod.
We are not fetching the port from the API, we only have to manually specify it for natlab tests.

In fact it's used that it's used for both country and direct connection:
- [config_with_vpn_iproute_setup.json](https://github.com/NordSecurity/libtelio/blob/teliod_vpn_country/nat-lab/data/teliod/config_with_vpn_iproute_setup.json)
- [config_with_vpn_country_pl.json](https://github.com/NordSecurity/libtelio/blob/teliod_vpn_country/nat-lab/data/teliod/config_with_vpn_country_pl.json)

clis/teliod/src/core_api.rs
tomasz-grz: Oh right, I see what you mean now 👍 will add the rest of the checks

clis/teliod/src/core_api.rs
tomasz-grz: Sounds to me like a separate task.. and was not part of this ticket description.
Since all of the API call sites will have to be updated not just these.

clis/teliod/src/daemon.rs
tomasz-grz: I didn't want to change the already established flow logic too much, to not make this PR even bigger or break things, but this kind of assumes that VPN connection is optional after configuring meshnet.

The full flow is:
Fetch device identity -> Fetch meshnet map -> Setup tunnel interface with meshnet IP -> _Fetch VPN server_ -> _Setup VPN connection_

Removing meshnet and making VPN not optional is part of a separate task that Luis is working on

clis/teliod/src/core_api.rs
tomasz-grz: We don't have any default values for these structs.
The only data that we de-serialize is `station`, `hostname` and `public_key`, that are already covered by this test.

Which other asserts would you recommend to add?

clis/teliod/src/core_api.rs
tomasz-grz: We are testing here that we can search for a country both by it's ISO code and full name.

How can we construct a `Country::from_str("af")`, without country ID and name, that are not optional?
```rust
pub struct Country {
    pub id: u64,
    pub name: String,
    pub code: String,
}
```

clis/teliod/src/core_api.rs
Jauler: LLT-6592 has been created

Jauler (approved):
I think that in future, we will need to have some sort of more active approach (e.g. for example `desired vs actual state` or similar, in order to handle changes in the OS config or handle lost connection, reconnection, retries etc. in order to keep uptime as high as is reasonable.

But lets go step-by-step therefore +1.0

clis/teliod/src/core_api.rs
tomasz-grz: Also here we are not comparing that `Country == Country`. We want to find the `Country.id` for a user input string.

clis/teliod/src/core_api.rs
tomasz-grz: Added assertion for the whole `vec![Sever]`

clis/teliod/src/core_api.rs
tomasz-grz: Added more asserts, and for whole `vec![Country]`

clis/teliod/src/config.rs
stalowyjez: Let's leave it as it is

PR #1411, opened 2025-08-21
Allow trailing conntracker close event in another test

mathiaspeters:
### Problem
`test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_client_side` failed in a recent nightly because we were expecting conntracker events LAST_ACK, TIME_WAIT but we got TIME_WAIT, CLOSE. We know from before that occasionally, probably due to multithreading in NepTUN, the final close event might arrive early which is fine and we should allow that to happen in our tests

### Solution
Allow trailing `CLOSE` in `test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_client_side`

- [x] README.md is updated


Hasan6979 (approved):
+2

PR #1416, opened 2025-08-26
LLT-6519: Remove meshnet from teliod

lcruz99:
This PR removes all meshnet-related functionality from teliod, refactoring it to focus solely on VPN exit node connections. This change is necessary as meshnet is being sunsetted and the related API endpoints will be unavailable starting December 2025.

#### Core Changes
- Deleted all meshnet-related code from teliod, including the `nc.rs` module
- Replaced device identity with user service credentials for VPN server connections
- Removed meshnet-related configuration options from `config.rs`
- Renamed `configure_interface.rs` to `interface.rs` and refactored to fetch IP addresses directly from the interface using the selected configuration provider
- Removed every code related with meshnet connection handling from `daemon.rs`
- Removed the entire `nc.rs` module which only use was to communicate meshnet map update events
- Refactored the `core_api.rs` to remove all meshnet related code
- Modified CGI API endpoints in `api.rs` and `web.rs` to remove meshnet-related functionality

#### Testing / Nat-lab
- Added service credentials generation and tracking functionality to support the new authentication method
- Removed meshnet settings from teliod configuration files
- Added iptables rule to VPN containers entrypoints to masquerade non-meshnet VPN server addresses
- Removed meshnet-specific tests from `test_teliod.py` and adapted to the VPN-only functionality

#### Breaking Changes
- Teliod no longer supports meshnet connections
- Configuration files must be updated to remove meshnet-related settings
- Device identity authentication is replaced with service credentials

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/teliod.py
+++ b/nat-lab/tests/teliod.py
@@ -26,6 +26,7 @@ from utils.logger import log
 from utils.process import Process, ProcessExecError
 from utils.router import IPStack
 from utils.router.linux_router import LinuxRouter
+from utils.testing import get_current_test_log_path
 
 
 class IgnoreableError(Exception):
@@ -230,6 +231,7 @@ class Teliod:
             await self.config.assert_match_daemon_start(stdout)
             yield self
         finally:
+            log.info("Teliod cleanup: exiting and removing socket (if exists)")
             try:
                 await self.quit()
             except ProcessExecError as exc:
@@ -241,6 +243,9 @@ class Teliod:
                 if await self.socket_exists():
                     log.debug("Dangling socket found, removing it..")
                     await self.remove_socket()
+            finally:
+                log.info("Teliod cleanup: saving logs")
+                await self._save_logs()
 
     async def is_alive(self) -> bool:
         try:
@@ -414,3 +419,43 @@ class Teliod:
             )
             with open(config_path, "w", encoding="UTF-8") as f:
                 f.write(clean_cfg)
+
+    async def _save_logs(self) -> None:
+        if os.environ.get("NATLAB_SAVE_LOGS") is None:
+            return
+
+        log_dir = get_current_test_log_path()
+        os.makedirs(log_dir, exist_ok=True)
+
+        log_files = [
+            (self.config.paths.daemon_log, "teliod.log"),
+            (self.config.paths.lib_log, "teliod_natlab.log"),
+        ]
+
+        for log_path, log_name in log_files:
+            try:
+                process = await self.connection.create_process(
+                    ["cat", str(log_path)], quiet=True
+                ).execute()
+                log_content = process.get_stdout()
+
+                filename = f"{self.connection.tag.name.lower()}_{log_name}"
+
+                if len(filename.encode("utf-8")) > 256:
+                    filename = f"{filename[:251]}.log"
+                    i = 0
+                    while os.path.exists(os.path.join(log_dir, filename)):
+                        filename = f"{filename[:249]}_{i}.log"
+                        i += 1
+
+                with open(
+                    os.path.join(log_dir, filename),
+                    "w",
+                    encoding="utf-8",
+                ) as f:
+                    f.write(log_content)
+
+                log.info("Saved %s to %s", log_name, filename)
+
+            except ProcessExecError as err:
+                log.warning("Failed to save %s: %s", log_name, err)


clis/teliod/src/daemon.rs
```diff
@@ -2,284 +2,156 @@ use futures::pin_mut;
...
-                        _ = sys_config
-                            .set_ip(&new_ip_address)
-                            .inspect_err(|e| error!("Failed to set IP with error '{e:?}'"));
-                        task_connect_to_exit_node(config.vpn, tx_channel.clone());
-                    }
+                let status_report = TelioStatusReport {
+                    telio_is_running: telio.is_running(),
+                    ip_address: interface_configurator.get_ip(),
+                    exit_node,
+                    external_nodes,
```
tomasz-grz: We probably don't need to report `external_nodes` anymore

clis/teliod/src/interface.rs
```diff
@@ -69,16 +71,28 @@ pub enum InterfaceConfigurationProvider {
...
-        info!("Creating interface config provider for {:?}", self);
-        match self {
-            Self::Manual => Box::new(Manual),
-            Self::Ifconfig => Box::new(Ifconfig::new(interface_name)),
-            Self::Iproute => Box::new(Iproute::new(interface_name)),
-            Self::Uci => Box::new(Uci::new(interface_name)),
+#[derive(Default, PartialEq, Eq, Deserialize, Serialize, Debug, Clone)]
+pub struct InterfaceConfig {
+    pub name: String,
+    pub configurator: InterfaceConfigurationProvider,
```
tomaszklak: Why make such a breaking change? And why is it needed as part of meshnet removal?

clis/teliod/src/cgi/web.rs
```diff
@@ -394,16 +395,14 @@ fn meshnet(app: &AppState) -> Markup {
         }
     };
 
-    let meshnet_status_text = if status.meshnet_ip.is_none() {
-        "Connecting..."
+    let meshnet_status_text = if let Some(exit_node) = &status.exit_node {
```
tomaszklak: I think this variable should be renamed and the `meshnet` part should be replaced with maybe `vpn`.

clis/teliod/src/cgi/web.rs
```diff
@@ -394,16 +395,14 @@ fn meshnet(app: &AppState) -> Markup {
...
-            "Meshnet Status ({})",
-            status
-                .meshnet_ip
-                .map(|ip| format!("{ip}"))
-                .unwrap_or_default()
-        )
+        "Disconnected (no exit node)"
     };
 
     html! {
```
tomaszklak: this `html!` contains references to meshnet

clis/teliod/src/main.rs
```diff
@@ -140,8 +137,8 @@ enum TeliodError {
 pub struct TelioStatusReport {
     /// State of telio runner
     pub telio_is_running: bool,
-    /// Assigned mesnet IP address
-    pub meshnet_ip: Option<IpAddr>,
+    /// Assigned IP address
+    pub ip_address: Option<IpAddr>,
     /// Node used in traffic routing, VPN server or meshnet peer
```
tomaszklak: This comment still references meshnet

clis/teliod/src/main.rs
```diff
@@ -140,8 +137,8 @@ enum TeliodError {
...
     pub exit_node: Option<ExitNodeStatus>,
     /// List of meshnet peers
```
tomaszklak: Should the `external_nodes` still be a `Vec`? Maybe it should be plain `Node` or `Option<Node>` ?

nat-lab/tests/conftest.py
```diff
@@ -403,7 +428,7 @@ def pytest_sessionfinish(session, exitstatus):
         collect_dns_server_logs()
         asyncio.run(collect_kernel_logs(session.items, "after_tests"))
         asyncio.run(collect_mac_diagnostic_reports())
-        asyncio.run(save_fakefm_logs())
+        # asyncio.run(save_fakefm_logs())
```
tomaszklak: This shouldn't be commented out

nat-lab/tests/conftest.py
```diff
@@ -472,7 +497,7 @@ async def start_tcpdump_processes():
     connections = [
         await SESSION_SCOPE_EXIT_STACK.enter_async_context(new_connection_raw(gw_tag))
         for gw_tag in ConnectionTag
-        if "_GW" in gw_tag.name
+        if "DOCKER_CONE_GW" in gw_tag.name
```
tomaszklak: Is that correct?

nat-lab/tests/helpers.py
```diff
@@ -471,38 +471,48 @@ async def ping_between_all_nodes(env: Environment) -> None:
...
-    connection,
+    connection: Connection,
     endpoint,
     method,
     ca_cert_path,
     data=None,
     authorization_header=None,
+    extra_headers: Optional[List[str]] = None,
     expect_response=True,
+    basic_auth: Optional[Tuple[str, str]] = None,
```
tomaszklak: nit: Could you add remaining types? :)

tomaszklak:
Should we also change the qpkg_name in `qnap/shared/teliod.cgi`?

clis/teliod/src/daemon.rs
```diff
@@ -2,284 +2,156 @@ use futures::pin_mut;
...
-                    let exit_node = external_nodes
-                        .iter()
-                        .find(|node| node.is_exit)
-                        .map(ExitNodeStatus::from);
+    while let Some(cmd) = rx_channel.blocking_recv() {
+        trace!("TelioTask got command {:?}", cmd);
+        match cmd {
+            TelioTaskCmd::GetStatus(response_tx_channel) => {
+                let external_nodes = telio.external_nodes()?;
+                // Find the identifier of the exit node (VPN server or meshnet peer), if present.
```
tomaszklak: Should this comment still mention meshnet?

clis/teliod/src/cgi/web.rs
```diff
@@ -2,7 +2,8 @@
 
 use std::{collections::HashMap, fs, str::FromStr};
 
-use crate::config::{InterfaceConfig, NordToken, TeliodDaemonConfigPartial};
+use crate::config::{NordToken, TeliodDaemonConfigPartial};
+use crate::interface::InterfaceConfig;
 use crate::TeliodDaemonConfig;
 use lazy_static::lazy_static;
 use maud::{html, Markup, Render};
```
tomaszklak: Line 37-38 configure meshnet icon (can't comment on unchanged lines, so leaving the comment here).

tomaszklak: Staring at line 68, there is a definition of `save_user_intent` which uses a flag to check if there is meshnet - this should be removed (can't comment on unchanged lines, so leaving the comment here).

tomaszklak: Starting at line 101 we define the `/meshnet` endpoint (can't comment on unchanged lines, so leaving the comment here).

tomaszklak: Starting 155 - `title` and link to icon should be updated (can't comment on unchanged lines, so leaving the comment here).

tomaszklak: Line 175 also uses function `meshnet` (can't comment on unchanged lines, so leaving the comment here).

clis/teliod/src/cgi/web.rs
```diff
@@ -210,7 +211,7 @@ fn update_config(app: &mut AppState, request: &CgiRequest) -> Result<TeliodDaemo
             .and_then(|v| LevelFilter::from_str(v).ok()),
         interface: values.get(TUNNEL_NAME).map(|name| InterfaceConfig {
             name: name.to_string(),
-            config_provider: app.config.interface.config_provider.clone(),
+            configurator: app.config.interface.configurator.clone(),
         }),
         ..Default::default()
     };
```
tomaszklak: Lines 276, 278 and 293 contains references to meshnet (can't comment on unchanged lines, so leaving the comment here).

tomaszklak: Starting at line 367 is the `meshnet` function (can't comment on unchanged lines, so leaving the comment here).

tomaszklak:
`clis/teliod/src/cgi/constants.rs` contains references to meshnet.

clis/teliod/src/config.rs
```diff
@@ -81,13 +83,7 @@ impl NordToken {
...
 
     fn validate(token: &str) -> bool {
-        token.is_empty() || (token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit()))
-    }
-}
-
-impl Default for NordToken {
-    fn default() -> Self {
-        NordToken(Arc::new(Hidden("".to_owned())))
+        token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit())
```
tomaszklak, line 88: Removing `token.is_empty()` means that NordToken::default() is not valid, is that expected?

clis/teliod/src/core_api.rs
```diff
@@ -1,212 +1,122 @@
...
+    let status = response.status();
+    let resp = response.text().await?;
+    debug!("Got response: {resp}");
+    if status.is_success() {
+        serde_json::from_str::<Value>(&resp)?
+            .get("nordlynx_private_key")
+            .and_then(|v| v.as_str())
+            .ok_or(Error::InvalidResponse)?
+            .parse::<SecretKey>()
+            .map_err(|_| Error::InvalidResponse)
```
tomaszklak: nit: I think that using a proper struct to do deserialization is more readable:

```rust
#[derive(Deserialize)]
struct NordlynxKeyResponse {
    nordlynx_private_key: SecretKey,
}
// ..
if status.is_success() {
    Ok(serde_json::from_str::<NordlynxKeyResponse>(&resp)?.nordlynx_private_key)
}
```

clis/teliod/src/daemon.rs
```diff
@@ -2,284 +2,156 @@ use futures::pin_mut;
...
-        task_retrieve_meshmap(
-            node_identity,
-            config.http_certificate_file_path.clone(),
-            config.authentication_token.clone(),
-            tx_channel.clone(),
-        );
+    let vpn_config = config.vpn.ok_or(TeliodError::InvalidConfigOption {
+        key: "vpn".to_owned(),
+        msg: "Couldn't get vpn configuration.".to_owned(),
+        value: format!("{:?}", config),
```
tomaszklak: I'm not sure that dumping whole config will add anything useful here, maybe we can replace it with an empty string?

clis/teliod/src/daemon.rs
```diff
@@ -2,284 +2,156 @@ use futures::pin_mut;
...
 fn telio_task(
-    node_identity: Arc<DeviceIdentity>,
     mut rx_channel: mpsc::Receiver<TelioTaskCmd>,
     tx_channel: mpsc::Sender<TelioTaskCmd>,
     config: TeliodDaemonConfig,
+    nordlynx_private_key: SecretKey,
 ) -> Result<(), TeliodError> {
     debug!("Initializing telio device");
 
+    // TODO: Make telio features configurable from teliod config LLT-6587]
```
tomaszklak: nit: typo at the end of comment

clis/teliod/src/daemon.rs
```diff
@@ -355,34 +185,38 @@ pub async fn daemon_event_loop(config: TeliodDaemonConfig) -> Result<(), TeliodE
     let socket = DaemonSocket::new(&DaemonSocket::get_ipc_socket_path()?)?;
     let mut cmd_listener = CommandListener::new(socket, telio_tx.clone());
 
-    let identity_ptr = {
-        let daemon_init_future = daemon_init(config.clone(), telio_tx.clone());
+    let nordlynx_private_key = {
+        let daemon_init_future = request_nordlynx_key(
```
tomaszklak: this variable should also change it's name

clis/teliod/src/interface.rs
```diff
@@ -107,20 +126,14 @@ impl ConfigureInterface for Manual {
 
 /// Implementation using `ifconfig`
 #[derive(Debug)]
-pub struct Ifconfig {
-    interface_name: String,
+pub struct Ifconfig<'a> {
+    interface_name: &'a str,
```
tomaszklak: What is the point of this borrow? This is very much not performance critical - we can afford to copy the interface name (especially since we pass around this type via box dyn trait).

clis/teliod/src/interface.rs
```diff
@@ -183,24 +215,15 @@ impl ConfigureInterface for Ifconfig {
 }
 
 /// Implementation using `iproute2`
-#[derive(Debug)]
-pub struct Iproute {
-    interface_name: String,
+#[derive(Debug, Default)]
+pub struct Iproute<'a> {
+    interface_name: &'a str,
```
tomaszklak: Same, what is the point of this borrow? This is very much not performance critical - we can afford to copy the interface name (especially since we pass around this type via box dyn trait).

clis/teliod/src/interface.rs
```diff
@@ -135,13 +148,13 @@ impl ConfigureInterface for Ifconfig {
 
         info!(
             "Assigning IP address for {} to {}",
-            self.interface_name, cidr_string
+            &self.interface_name, cidr_string
```
tomaszklak: nit: is that `&` really needed?

clis/teliod/src/interface.rs
```diff
@@ -378,8 +428,8 @@ impl ConfigureInterface for Iproute {
 
 /// UCI-based implementation for managing network configuration on OpenWRT devices.
 #[derive(Debug)]
-pub struct Uci {
-    interface_name: String,
+pub struct Uci<'a> {
+    interface_name: &'a str,
```
tomaszklak: Same, what is the point of this borrow? This is very much not performance critical - we can afford to copy the interface name (especially since we pass around this type via box dyn trait).

clis/teliod/src/interface.rs
```diff
@@ -314,10 +337,37 @@ impl ConfigureInterface for Iproute {
...
+            let line = line.trim();
+            if line.contains("inet ") {
+                let parts: Vec<&str> = line.split_whitespace().collect();
+                if let Some(ip_idx) = parts.iter().position(|&s| s == "inet") {
+                    if let Some(ip_cidr) = parts.get(ip_idx + 1) {
+                        let ip_str = ip_cidr.split('/').next()?;
+                        if let Ok(ip) = ip_str.parse::<IpAddr>() {
+                            return Some(ip);
+                        }
+                    }
```
tomaszklak: We already have `ipnet` dependency, could we use https://docs.rs/ipnet/latest/ipnet/enum.IpNet.html for cidr parsing?

nat-lab/bin/core-api.py
```diff
@@ -484,6 +496,40 @@ def handle_public_key(self):
         self.end_headers()
         self.wfile.write(b"Public key saved")
 
+    @requires_basic_authentication
+    def handle_service_credentials(self):
+        current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+        if "credentials" not in _SERVICE_CREDENTIALS_CACHE:
```
tomaszklak: why is `_SERVICE_CREDENTIALS_CACHE` a dictionary if it always contains at most one (`credentials`) key?

nat-lab/tests/conftest.py
```diff
@@ -262,7 +262,32 @@ async def kill_natlab_processes():
     subprocess.run(["sudo", cleanup_script_path]).check_returncode()
 
 
-PRETEST_CLEANUPS = [kill_natlab_processes, clear_ephemeral_setups_set]
+async def reset_service_credentials_cache():
+    try:
+        result = subprocess.run(
+            [
+                "curl",
```
tomaszklak: nit: if this is not run from the docker, but from the host running tests, it would be (imho) slightly nicer to use python libraries to make a http post, instead of calling curl binary.

nat-lab/tests/teliod.py
```diff
@@ -130,17 +140,39 @@ def __init__(
...
+        api: Optional[API] = None,
     ) -> None:
-        self._connection: Connection = connection
+        self._api: API = api if api is not None else API()
         self._exit_stack: AsyncExitStack = exit_stack
+        self.connection: Connection = connection
         self.config: Config = config
 
+    @classmethod
+    async def new(
```
tomaszklak, line 160: Why do we need this separate factory method? Can we merge it into `__init__`?

clis/teliod/src/config.rs
```diff
@@ -278,12 +267,11 @@ impl Default for TeliodDaemonConfig {
             adapter_type: AdapterType::default(),
             interface: InterfaceConfig {
                 name: "nlx".to_string(),
-                config_provider: Default::default(),
+                configurator: Default::default(),
             },
             vpn: None,
```
tomasz-grz: I think `VpnConfig` in `TeliodDaemonConfig` should not be optional anymore.

clis/teliod/src/config.rs
```diff
@@ -425,7 +408,7 @@ mod tests {
                 "adapter_type": "linux-native",
                 "interface": {{
                     "name": "test",
-                    "config_provider": "manual"
+                    "configurator": "manual"
```
tomasz-grz: Always nice to see refactoring, but maybe this could be part of a different PR to make this one a bit less massive? or at least an atomic commit?

clis/teliod/src/cgi/web.rs
```diff
@@ -2,7 +2,8 @@
```
LukasPukenis: These files still contain a string "meshnet" inside of them under `clis/teliod` directory:
```
./teliod/API.md
./teliod/README.md
./teliod/src/cgi/constants.rs
./teliod/src/cgi/web.rs
./teliod/src/daemon.rs
./teliod/src/main.rs
```

I am aware that you're doing QNAP removal as well - https://github.com/NordSecurity/libtelio/pull/1426 , to me it looks like it makes sense to merge QNAP removal first, then rebase this branch since a sanity check of "meshnet" being removed from `teliod` is going to be `grep meshnet` :)

Alternatively, parts except QNAP can be just addressed.

clis/teliod/README.md
```diff
@@ -65,7 +65,7 @@ needs be absolute, otherwise will be relative to `working-directory` when daemon
 - `interface`
   - `name` - Name of tunnel interface to connect to. Note that for macOS
   the name has to be in form `tun#` where `#` can be any integer number
-  - `config_provider` - Provider for configuring the interface address,
+  - `configurator` - Provider for configuring the interface address,
```
LukasPukenis: While we do not yet have teliod released, this would be a breaking change.. I think the previous name was just as good, for example with the manual value it looked more natural.

In all honesty I would refrain from refactoring when removing stuff if it's not necessary - reviewing purely removal PR would be much simpler :)

clis/teliod/src/main.rs
```diff
@@ -25,11 +25,10 @@ mod cgi;
 mod command_listener;
```
LukasPukenis, line 8: Since this PR mentions:
> Replaced device identity with user service credentials for VPN server connections

There's still a mention of `identity` in this file(this diff doesn't touch it so attaching here)
```
teliod/src/command_listener.rs
107:                        CommandResponse::Err("Obtaining identity, ignoring".to_owned()).serialize(),
```

clis/teliod/src/interface.rs
lcruz99: Reading `config_provider` didn't sound that we were going to use it to manipulate the routing table. I think it can be mistaken by something else.

Ughh, you're right, not a necessary change for the context of this PR and had to fix a  few broken tests/config files. I'll revert it back.

nat-lab/tests/helpers.py
```diff
@@ -471,38 +471,48 @@ async def ping_between_all_nodes(env: Environment) -> None:
...
     if authorization_header:
         curl_command.extend(["-H", f"Authorization: {authorization_header}"])
 
-    log.info("Curl command: %s", curl_command)
+    if basic_auth:
+        username, password = basic_auth
+        curl_command.extend(["-u", f"{username}:{password}"])
 
-    process = await connection.create_process(curl_command, quiet=True).execute()
+    process = await connection.create_process(curl_command, quiet=False).execute()
```
LukasPukenis, line 509: nit: maybe we can pass the `quiet` parameter as a `debug` argument, that's defaulted to `False`?

nat-lab/tests/teliod.py
```diff
@@ -130,17 +140,39 @@ def __init__(
...
+        cls,
+        exit_stack: AsyncExitStack,
+        config_type: IfcConfigType,
+        no_detach: bool = False,
+    ) -> "Teliod":
+        connection = (
+            await setup_connections(exit_stack, [ConnectionTag.DOCKER_CONE_CLIENT_1])
```
LukasPukenis: I was a bit puzzled about which container was used for `teliod` tests until I saw it's inlined in this function instead of being a parameter. I think parameter is beneficial since it stands out more. In addition to that I am a bit puzzled about `__init__` vs `new()` as well.. Why not just have one :thinking:

nat-lab/tests/teliod.py
```diff
@@ -130,17 +140,39 @@ def __init__(
...
+        )[0].connection
+        await clean_up_registered_machines_on_api(connection, CORE_API_URL)
```
LukasPukenis: This function seems to be meshnet related, since it's in `teliod.py` is it really needed here?

clis/teliod/src/cgi/web.rs
lcruz99: done

nat-lab/tests/teliod.py
```diff
@@ -306,3 +337,52 @@ async def setup_interface(
...
+            True,
+            IPStack.IPv4,
+            [LIBTELIO_LOCAL_IP],
+        )
+        self._api.prepare_all_vpn_servers()
+
+        return node
+
+    @asynccontextmanager
+    async def setup_vpn_public_key(self, key: str) -> AsyncIterator:
```
LukasPukenis: minor: can we rename `key` to `pubkey` or `server_pubkey`?

clis/teliod/src/cgi/web.rs
lcruz99: done

nat-lab/tests/teliod.py
```diff
@@ -306,3 +337,52 @@ async def setup_interface(
...
+            "teliod",
+            core_response["nordlynx_private_key"],
+            generate_public_key(core_response["nordlynx_private_key"]),
+            True,
+            IPStack.IPv4,
+            [LIBTELIO_LOCAL_IP],
+        )
+        self._api.prepare_all_vpn_servers()
+
+        return node
```
LukasPukenis: :thinking: we have a mocked API server where we send a request for randomly generated credentials. However since `teliod` itself doesn't query this endpoint, is it really necessary to have it? We could probably inline the randomized credentials in this function.

While on this topic - mocking API makes sense only if `teliod` itself uses that API endpoint but I do not see that it does, we don't offer a way for `teliod` to override `nordvpn` API endpoint. That makes me think - why do we need to have a mocked API server if our application doesn't use it. This is more of an open question, not related to this PR directly :thinking:

clis/teliod/src/daemon.rs
```diff
@@ -2,284 +2,156 @@ use futures::pin_mut;
...
+        msg: "Couldn't get vpn configuration.".to_owned(),
+        value: format!("{:?}", config),
+    })?;
+    let exit_node = ExitNode {
+        identifier: uuid::Uuid::new_v4().to_string(),
+        public_key: vpn_config.server_pubkey,
+        allowed_ips: None,
+        endpoint: Some(vpn_config.server_endpoint),
+    };
+    task_connect_to_exit_node(exit_node, tx_channel.clone());
```
LukasPukenis: I think it would be cleaner if `telio_task` would simply handle commands instead of handling commands and also orchestrating VPN connection.

Alternatively - splitting into two functions also works. What do you think about that?

clis/teliod/src/cgi/web.rs
lcruz99: `cgi` module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

clis/teliod/src/cgi/web.rs
lcruz99: `cgi` module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

clis/teliod/src/cgi/web.rs
lcruz99: `cgi` module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

clis/teliod/src/main.rs
lcruz99: changed

clis/teliod/src/main.rs
lcruz99: I agree, it should be plain Node. In fact I think we can drop `external_nodes` and just use exit_node as `Option<Node>`.

nat-lab/tests/conftest.py
lcruz99: Sorry, shouldn't have included this in the commit

nat-lab/tests/conftest.py
lcruz99: Sorry, shouldn't have included this in the commit

nat-lab/tests/helpers.py
lcruz99: oh lazy me, will add :)

clis/teliod/src/daemon.rs
lcruz99: fixed

clis/teliod/src/cgi/web.rs
lcruz99: cgi module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

lcruz99: cgi module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

clis/teliod/src/cgi/web.rs
lcruz99: cgi module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

lcruz99: cgi module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

clis/teliod/src/cgi/web.rs
lcruz99: cgi module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

lcruz99: cgi module is removed on https://github.com/NordSecurity/libtelio/pull/1426, therefore not worthy of reviewing.

clis/teliod/src/config.rs
lcruz99: yes, empty token was being considered valid just for UI functionality reasons, a refactoring was planned but given that QNAP is being dropped so the UI, making the empty token validation not needed anymore. This issue was discussed here: https://github.com/NordSecurity/libtelio/pull/1390#discussion_r2269085986

clis/teliod/src/daemon.rs
lcruz99: agreed, changed

clis/teliod/src/daemon.rs
lcruz99: fixed

clis/teliod/src/daemon.rs
lcruz99: agreed, changed

clis/teliod/src/interface.rs
lcruz99: it's not, it becomes a && which the compiler derefs then. will remove

clis/teliod/src/interface.rs
lcruz99: you're right, unnecessary complexity here. Changed to owned Strings.

clis/teliod/src/interface.rs
lcruz99: you're right, unnecessary complexity here. Changed to owned Strings.

clis/teliod/src/interface.rs
lcruz99: you're right, unnecessary complexity here. Changed to owned Strings.

clis/teliod/src/interface.rs
lcruz99: yess good suggestion, will do

nat-lab/tests/conftest.py
lcruz99: agreed, changed

nat-lab/tests/teliod.py
lcruz99: We cannot write async code on `__init__` due to python limitation. Also both instantiations patterns are required in the tests:
- Direct instantiation when we already have a connection and want manual control
- Factory method when we want a standard test setup with automatic initialization

clis/teliod/src/daemon.rs
lcruz99: changed

clis/teliod/src/config.rs
lcruz99: reverted back

clis/teliod/src/cgi/web.rs
lcruz99: that makes sense, will merge #1426 and only then this PR

clis/teliod/README.md
lcruz99: agreed, reverted back.

clis/teliod/src/main.rs
lcruz99: missed that, changed

nat-lab/tests/helpers.py
lcruz99: it already is defaulted to `False`. When `debug` is used all logs are printed regardless of this value.

lcruz99: https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/utils/connection/docker_connection.py#L128

nat-lab/tests/teliod.py
lcruz99: Agreed, I'll set them as parameters.

As for the `__init__` vs `new()`, I paste the same answer I've given to @tomaszklak above:

We cannot write async code on __init__ due to python limitation. Also both instantiations patterns are required in the tests:

    Direct instantiation when we already have a connection and want manual control
    Factory method when we want a standard test setup with automatic initialization

lcruz99: >  think parameter is beneficial since it stands out more. 

btw, do you mean specifying the container in every call to `new()` or making it providing it as the parameter default value will suffice?

nat-lab/tests/teliod.py
lcruz99: nice catch, this is effectively not needed, will remove.

nat-lab/tests/teliod.py
lcruz99: sure, changed

nat-lab/tests/teliod.py
lcruz99: > ...However since teliod itself doesn't query this endpoint, is it really necessary to have it? ...

It does here: https://github.com/NordSecurity/libtelio/blob/b7ae10c12bcca668aa3428e6b6b476129c5bf279/clis/teliod/src/core_api.rs#L98

clis/teliod/src/daemon.rs
lcruz99: definitely! Will refactor this

clis/teliod/src/core_api.rs
lcruz99: it doesn't directly deserialize to SecretKey, it needs to be parsed first to `str` and only then to `SecretKey`. I'll still add the struct as I agree it improves readability

clis/teliod/src/cgi/web.rs
LukasPukenis: since 1426 was merged, this comment can be resolved :)

nat-lab/tests/helpers.py
LukasPukenis: Oh, I missed that! Sorry.

nat-lab/tests/teliod.py
LukasPukenis: I like that the tests have containers explicitly listed in parameters so it's very obvious which ones are used and also can be parametrized. If `Teliod.new()` would take a parameter for `ConnectionTag.DOCKER_CONE_CLIENT_1`, that should be enough I think :)

nat-lab/tests/teliod.py
LukasPukenis: Oh I missed it entirely :facepalm:  The base is hardcoded in the code:
```
const API_BASE: &str = "https://api.nordvpn.com/v1";
```
but in `docker-compose.yml` I can see an entry on `extra_hosts` that has
```
      - "api.nordvpn.com:10.0.80.86"
```
which "overrides" it. My bad! Thanks for pointing out :)

clis/teliod/src/main.rs
tomasz-grz: We probably don't need this either anymore, since we are not fetching an IP address from the API, but rather fixing it ourselves to a "static" value of `10.5.0.2`

nat-lab/data/teliod/config_with_vpn_country_de.json
```diff
@@ -5,10 +5,9 @@
   "authentication_token": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
   "adapter_type": "neptun",
   "http_certificate_file_path": "/etc/ssl/server_certificate/test.pem",
-  "device_identity_file_path": "/etc/teliod/data.json",
   "interface": {
     "name": "teliod",
-    "config_provider": "iproute"
+    "configurator": "iproute"
```
tomaszklak: I think it should be `config_provider`

nat-lab/data/teliod/config_with_vpn_country_pl.json
```diff
@@ -5,10 +5,9 @@
   "authentication_token": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
   "adapter_type": "neptun",
   "http_certificate_file_path": "/etc/ssl/server_certificate/test.pem",
-  "device_identity_file_path": "/etc/teliod/data.json",
   "interface": {
     "name": "teliod",
-    "config_provider": "iproute"
+    "configurator": "iproute"
```
tomaszklak: I think it should be `config_provider`

clis/teliod/src/cgi/web.rs
tomaszklak: In that case, maybe this PR could be rebased ontop of #1426? This way it will be much more clear.

clis/teliod/src/config.rs
tomaszklak: What I mean is that token created with `NordToken::default()` will not be valid according to it's own rules. Imo, this means that the `Default` should not be implemented or it should be implemented in a way that produces a valid token.

nat-lab/data/teliod/config_with_vpn_country_de.json
lcruz99: changed

nat-lab/data/teliod/config_with_vpn_country_pl.json
lcruz99: changed

nat-lab/tests/teliod.py
lcruz99: done

clis/teliod/src/config.rs
lcruz99: done

nat-lab/bin/core-api.py
lcruz99: was equating a test with multiple credentials which is not applicable in the end. changed

clis/teliod/src/config.rs
lcruz99: agreed, changed so that `NordToken::default()` produces a valid token.

clis/teliod/src/cgi/web.rs
lcruz99: done

clis/teliod/src/config.rs
lcruz99: https://github.com/NordSecurity/libtelio/pull/1416/files#diff-6aef2145f1124adca821f60433aab06532511863dc562ee553547d95c9c1bc80R94

LukasPukenis:
+1 I see that majority of my comments were addressed and the ones unaddressed are non-blocking

clis/teliod/src/config.rs
```diff
@@ -300,31 +318,29 @@ const fn default_log_file_count() -> usize {
...
     /// Direct endpoint and public key of VPN server
-    Server(DirectEndpointConfig),
+    Server(Endpoint),
     /// Country name or ISO code of VPN server location
     Country(String),
 }
 
+impl Default for VpnConfig {
+    fn default() -> Self {
+        VpnConfig::Country("lt".to_string())
```
tomasz-grz, line 340: Why `LT` should be considered the default?
If this is empty, or something like "any" it would pick automatically the recommended VPN server in your area (I believe)..

Maybe we should just have a VpnConfig variant like `VpnConfig::Any` or `VpnConfig::Recommended`? What do you think?

clis/teliod/src/daemon.rs
```diff
@@ -2,475 +2,386 @@ use futures::pin_mut;
...
-    /// But we are not actually connected until MaybeSetupVPNRoutes is successfully completed
-    /// It used mainly to display `hostname` in [TelioStatusReport]
-    maybe_connected_vpn_server: Option<ExitNodeConfig>,
+/// Libtelio and VPN status report
+#[derive(Debug, Serialize, Deserialize, PartialEq, Default, Clone)]
+pub struct TelioStatusReport {
+    /// State of telio runner
+    pub telio_is_running: bool,
+    /// Assigned IP address
+    pub ip_address: Option<IpAddr>,
```
tomasz-grz, line 93: Nit: Do we need this still since we don't assign the IP address dynamically by the API?
But rather always set to`10.5.0.2`?

clis/teliod/src/core_api.rs
```diff
@@ -125,27 +95,43 @@ impl Country {
...
+
+    fn try_from(server: &Server) -> Result<Self, Self::Error> {
+        Ok(Self {
+            identifier: uuid::Uuid::new_v4().to_string(),
+            public_key: server
+                .wg_public_key()
+                .and_then(|key| key.parse().ok())
+                .ok_or(TeliodError::EndpointNoPublicKey)?,
+            allowed_ips: None,
+            endpoint: Some(SocketAddr::new(server.address(), DEFAULT_WIREGUARD_PORT)),
```
tomasz-grz, line 130: Just a reminder, if you are actually using `try_from` or `try_into` in the connection flow, that you need to override `DEFAULT_WIREGUARD_PORT` for natlab tests.

clis/teliod/src/config.rs
```diff
@@ -300,31 +318,29 @@ const fn default_log_file_count() -> usize {
...
-/// Direct VPN server endpoint configuration
-#[derive(PartialEq, Eq, Deserialize, Serialize, Clone, Debug)]
-pub struct DirectEndpointConfig {
-    /// IP of VPN server
+#[derive(PartialEq, Eq, Deserialize, Serialize, Debug, Clone)]
+pub struct Endpoint {
     pub address: IpAddr,
-    /// Public key of VPN server
     pub public_key: PublicKey,
+    pub hostname: Option<String>,
```
tomasz-grz, line 325: nit: `hostname` being part of the user provided config, somehow implies that I can simply connect by providing `"hostname": "pl100.nordvpn.com"`, is that the case?

clis/teliod/src/daemon.rs
lcruz99: > since we don't assign the IP address dynamically by the API?

Exit node IP had always been static, either `100.64.0.1` (meshnet) or `10.5.0.2` (non-meshnet).

If we do not want to include it in the TelioStatusReport shown to the user there would be no usage and I agree with the removal of it. However I say it doesn't hurt in including it in the report.

clis/teliod/src/config.rs
lcruz99: Good question.

Hostname in the config is ignored and and it's used as a custom name for the server the user specifies. Eg: For the following server:
`
{
   "vpn": {
    "server": {
      "address": "10.0.100.1",
      "public_key": "<some_publickey>",
      "hostname": "pl100.nordvpn.com" # is ignored and the user can call it whatever he wants.
    }
  },
}
`

Maybe in the near future this can be used to connect to a server through this given hostname..

clis/teliod/src/core_api.rs
lcruz99: I think the caller should be responsible for overriding it as in https://github.com/NordSecurity/libtelio/blob/c7d7a2813d687b6dce2920bbc8ba9b5c8467ef8b/clis/teliod/src/daemon.rs#L265.

clis/teliod/src/config.rs
tomasz-grz: I guess it's also ok to give it a "nickname" so you remember which server it was 😅

clis/teliod/src/daemon.rs
tomasz-grz: I guess, doesn't hurt to keep it for now

clis/teliod/src/config.rs
tomasz-grz: Created ticket LLT-6641 for this

PR #1417, opened 2025-08-27
LLT-5601: Map more error types to FfiError

mathiaspeters:
### Problem
Most errors that are generated inside libtelio get mapped to `UnknownError` before being passed over the FFI layer. Some of the errors can be mapped to more appropriate error types

### Solution
Map the errors to better types where possible

Note: the list of errors that can be better mapped turned out to be surprisingly small

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -89,14 +89,7 @@ impl From<&serde_json::Error> for TelioError {
 
 impl From<DevError> for TelioError {
     fn from(err: DevError) -> Self {
-        // TODO: Map more error types.
-        match err {
-            DevError::AlreadyStarted => Self::AlreadyStarted,
-            DevError::BadPublicKey => Self::InvalidKey,
-            _ => Self::UnknownError {
-                inner: format!("{err:?}"),
-            },
-        }
+        (&err).into()
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/ffi/types.rs
```diff
@@ -89,14 +89,7 @@ impl From<&serde_json::Error> for TelioError {
...
     fn from(err: DevError) -> Self {
-        // TODO: Map more error types.
-        match err {
-            DevError::AlreadyStarted => Self::AlreadyStarted,
-            DevError::BadPublicKey => Self::InvalidKey,
-            _ => Self::UnknownError {
-                inner: format!("{err:?}"),
-            },
-        }
+        (&err).into()
```
mathiaspeters, line 92: Before I got started I thought that I don't want to have to duplicate everything so I made this simplification. Then it turned out that very few errors could be mapped differently so this cleanup was a bit unnecessary, but still, it reduces duplication

mathiaspeters:
> `BadConfig` is available on the bindings, can it describe a `PollingPeriodZero` error?

I like the idea of this and was about to do it when I realized that doing so would be less informative for the consumer. If the polling period is set to 0, they will currently get `UnknownError { inner: PollingPeriodZero }`, which may not be the best but it's better than just `BadConfig` since that doesn't say anything about what is actually wrong. And thinking about it, using `InvalidKey` both for `BadPublicKey` and `BadPrivateKey` might also be confusing, but I think we're already doing that in places. I think it could be worth it to revisit errors in the FFI, but I think that might be something for after meshnet has been fully sunset

jjanowsk:
> We can, but that would technically be a breaking change and would therefore require a major release

I would argue that this is already a breaking change. Imagine the app handling somehow UnknownError("Not started"). Now they need to change the code to handle NotStarted error.

PR #1419, opened 2025-09-01
Move OpenWrt GW to Qemu based approach

olekoliinyk:
### Problem
With previous container based approach there is an issue that it cannot simulate the `ubus` and it's needed to set-up the network interfaces through `uci`.

### Solution
1. Emulate OpenWrt using QEMU
2. Attach storage to be able to access required files during QEMU run
3. Set-up SSH connection
4. Set-up nat-lab's DNS servers to be able to access nat-lab's core-api

Please note, that currently access to QEMU VM is possible with `127.0.0.1 -p 30022 `.
Proper (with bridging) networking solution will be addressed in a separate PR.

- [x] README.md is updated


nat-lab/bin/openwrt/10-usbmount-initsh.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+cat <<'EOUSB' > /etc/init.d/usbmount
+#!/bin/sh /etc/rc.common
+START=89
```
mathiaspeters, line 4: Is this some magic variable? Just asking because I don't see it being used

nat-lab/bin/openwrt/send-config-to-vm.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+set -ex
+find /var/lib/vmconfig
+rm -rf /tmp/vmconfig
+cp -rv /var/lib/vmconfig /tmp/vmconfig
+mkdir -p /tmp/vmconfig/container.d /tmp/vmconfig/vm.d
+if [ "$(ls -A /usr/local/share/vmconfig/container.d)" ]; then cp /usr/local/share/vmconfig/container.d/* /tmp/vmconfig/container.d; fi
+if [ "$(ls -A /usr/local/share/vmconfig/vm.d)" ]; then cp /usr/local/share/vmconfig/vm.d/* /tmp/vmconfig/vm.d; fi
```
mathiaspeters: Nit: These two if-statements could be formatted differently so they're not just overly long one-liners

mathiaspeters:
+1

nat-lab/bin/openwrt/10-usbmount-initsh.sh
olekoliinyk: It actually tells when this init script will run during system boot process. 
In our case it will be executed when most system services have already started.

nat-lab/bin/openwrt/30-wait-for-network.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -ex
+ubus wait_for network.interface.wan
+sleep 3
```
Jauler: :suspect: do we need to sleep here?

nat-lab/bin/openwrt/40-dns.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -ex
+uci set dhcp.@dnsmasq[0].noresolv="1"
+uci add_list dhcp.@dnsmasq[0].server="10.0.80.82"
+uci add_list dhcp.@dnsmasq[0].server="10.0.80.83"
+uci add_list dhcp.@dnsmasq[0].rebind_domain="api.nordvpn.com"
+uci commit dhcp
+/etc/init.d/dnsmasq restart
+sleep 5
```
Jauler: do we need to sleep here?

nat-lab/openwrt.Dockerfile
```diff
@@ -1,5 +1,46 @@
...
+WORKDIR /ipk-source
+COPY bin/nordvpn_x86_64.ipk /ipk-source/
+COPY data/core_api/test.pem /ipk-source/
+
+RUN mkdir -p /var/lib/qemu-image
+WORKDIR /var/lib/qemu-image
+
+
+RUN mkdir -p /var/lib/qemu && \
+    gunzip -c openwrt-24.10.2-x86-64-generic-ext4-combined.img.gz > /var/lib/qemu/image.raw && \
```
Jauler: I wonder, although I have no strong preference either way, but decided to discuss this.

Does it make sense to convert from raw to qcow? The image itself is not super big, and qemu can run directly out of raw format.

nat-lab/docker-compose.yml
```diff
@@ -478,39 +477,40 @@ services:
...
+      args:
+        LIBTELIO_ENV_SEC_CONTAINER_REGISTRY: $LIBTELIO_ENV_SEC_CONTAINER_REGISTRY
+        LIBTELIO_ENV_NAT_LAB_DEPS_TAG: $LIBTELIO_ENV_NAT_LAB_DEPS_TAG
+    cap_drop:
+      - ALL
+    cap_add:
+      - NET_ADMIN
+      - NET_RAW
+    sysctls:
+      - net.ipv4.ip_forward=1
```
Jauler: This is probably a bit redundant, as it is enabled by default in docker containers

nat-lab/bin/openwrt/30-wait-for-network.sh
olekoliinyk: I think not necessary. Left from debugging 😞

nat-lab/bin/openwrt/40-dns.sh
olekoliinyk: Same as here: 
https://github.com/NordSecurity/libtelio/pull/1419#discussion_r2318384699

nat-lab/openwrt.Dockerfile
olekoliinyk: I was playing with dynamic disk resizing which is only supported in qcow2. 
But agree, in our case the raw format might be better as size of the disk is relatively small.

Jauler (approved):
Network bridging setup is kind needed, but I saw you mentioned this is left for future PRs. +1.0

PR #1420, opened 2025-09-01
Add linters in clippy

Hasan6979:
### Problem
There are lints checked by nagger which are now supported by clippy

### Solution
Use clippy for lint checking for supported lints rather than nagger

- [x] README.md is updated

### DIFF
--- a/clis/derpcli/src/main.rs
+++ b/clis/derpcli/src/main.rs
@@ -61,7 +61,6 @@
 //! ```
 
 #![allow(unknown_lints)]
-#![allow(unwrap_check)]
 
 mod conf;
 mod metrics;
@@ -439,7 +438,7 @@ async fn run_without_clients_config(config: conf::Config) -> Result<()> {
     Ok(())
 }
 
-#[allow(index_access_check)]
+#[allow(clippy::indexing_slicing)]
 fn create_buffer(size: usize, counter: i32) -> String {
     let mut buf = format!("{}B", size);
     let middle_point = buf.len();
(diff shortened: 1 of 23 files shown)


clis/derpcli/src/main.rs
```diff
@@ -61,7 +61,7 @@
 //! ```
 
 #![allow(unknown_lints)]
-#![allow(unwrap_check)]
+#![allow(clippy::unwrap_used)]
```
stalowyjez: `clippy::unwrap_used` was added already but it wasn't allowed here, why it worked before?

Hasan6979: Good point. tbh I don't know why. 😅  as in other places both `unwrap_check` and `clippy::unwrap_used` were added.

tomasz-grz: Possibly because we don't run clippy on `derpcli`.
In CI we used to only run clippy with `cargo clippy --lib`, which doesn't check any binary packages... I added the package flag so that it can also check `teliod`.

If you manually run `just clippy derpcli` there's a ton of errors :D

stalowyjez: Is unwrap really used anywhere in this file? From a quick search it seems it contains only `unwrap_or_default`s and `unwrap_or_else`s?

Hasan6979: We could remove it then, since it isn't really checked in clippy so doesn't make any sense to add allow here.

mathiaspeters (approved):
+1

PR #1422, opened 2025-09-02
Update wg nt with latest upstream changes

Hasan6979:
### Problem
wireguard-nt rust wrappers repository was outdated from the upstream version

### Solution
Update the wg-nt rust wrappers repo and update the version in libtelio


Hasan6979:
Most of the changes comes from cargo lock, because I ended up running `cargo update` otherwise udeps linter was failing

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -421,7 +433,9 @@ impl Adapter for WindowsNativeWg {
             interface_watcher.clear_last_known_configuration();
             interface_watcher.stop();
         }
-        self.adapter.down();
+        if let Err(e) = self.adapter.down() {
+            telio_log_warn!("Failure to turn adapter down {e}");
```
mathiaspeters: Nit: I think it should be either `Failed to turn off adapter` or `Failed to set adapter state to down`

matislovas (dismissed):
+1

mathiaspeters (approved):

PR #1423, opened 2025-09-02
Add control-flow-guard to windows build

matislovas:
Adding `control-flow-guard` to Windows build. Built binaries was tested with [binary-security-check](https://crates.io/crates/binary-security-check) tool.

- [x] README.md is updated

### DIFF
--- a/ci/build_libtelio.py
+++ b/ci/build_libtelio.py
@@ -89,9 +89,9 @@ def post_check_for_windows_static_runtime(config, args):
                     config.rust_target, dll_bin, config.debug
                 )
                 if os.path.isfile(dll_bin_path):
-                    should_link_statically = (
-                        WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC]
-                        in GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"]
+                    should_link_statically = any(
+                        WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC] in s
+                        for s in GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"]
                     )
                     msvc_context = None
                     if not msvc.is_msvc_active():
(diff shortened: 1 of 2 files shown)


ci/build_libtelio.py
```diff
@@ -95,8 +95,15 @@ def post_check_for_windows_static_runtime(config, args):
...
-                        in GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"]
+                        len(
+                            [
+                                s
+                                for s in GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"]
+                                if WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC]
+                                in s
+                            ]
+                        )
+                        != 0
```
LukasPukenis: minor: `any()` could be used so it becomes:
```
any(
    WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC] in s
    for s in GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"]
)
```

ci/build_libtelio.py
```diff
@@ -561,9 +568,11 @@ def exec_build(args):
                 GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"] = ()
 
             GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"] += (
-                WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC],
+                WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC]
+                + f"-C control-flow-guard ",
```
LukasPukenis: just double checking - gluing strings here assumes that already existing one has a space at the end, is that correct?

LukasPukenis: Also additional question - maybe it's worth to put the `CFG` in `rust_build_utils.rust_utils_config`?

matislovas: Yea, the `WINDOWS_RUNTIME_LINKING` already has a space at the end. I mean, this "glueing" is not an elegant solution, but it seems that setting multiple `RUSTFLAGS` can only be done in this way.

matislovas: > Also additional question - maybe it's worth to put the `CFG` in `rust_build_utils.rust_utils_config`?

So that was my dilemma as well. If there is no convenient method to append those `RUSTFLAGS`, maybe we should leave it to add it on per-lib basis.

ci/build_libtelio.py
matislovas: Fixed

ci/build_libtelio.py
LukasPukenis: In rust_build_utils it's defined as
```
WINDOWS_RUNTIME_LINKING = {
    WindowsLinkingMethod.STATIC: " -C target-feature=+crt-static ",
    WindowsLinkingMethod.DYNAMIC: " -C target-feature=-crt-static ",
}
```
it looks like this could easily go there since it's a security flag and we probably always want it?

LukasPukenis (commented):
+1 however would be nice to move the flag setting into `rust_build_utils` if possible to benefit everyone :)

ci/build_libtelio.py
```diff
@@ -557,13 +557,18 @@ def exec_build(args):
...
 
-            if not "RUSTFLAGS" in GLOBAL_CONFIG["windows"]["env"]:
-                GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"] = ()
+            if not GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"]:
+                GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"] = (
+                    WINDOWS_RUNTIME_LINKING[WindowsLinkingMethod.STATIC],
+                    "set",
+                )
+            else:
+                current_flags = GLOBAL_CONFIG["windows"]["env"]["RUSTFLAGS"][0]
```
LukasPukenis, line 522: :thinking: should we hardcode `[0]` instead of taking all the values there are?

LukasPukenis (commented):
+2 

This uses `control_flow_guard` from `rust_build_utils` which is nice

ci/build_libtelio.py
matislovas: The tuple there consists of `("<all _the_flags>", "set")`, so it always will be first member of tuple

PR #1428, opened 2025-09-05
Mark test_find_available_lower_rule_priority and test_find_available_ower_rule_priorities linux only

tomaszklak:
Those tests will not work on macOS.


mathiaspeters (approved):
+2

PR #1432, opened 2025-09-09
LLT-6598: Bump uniffi-generators to v0.28.3-4

mathiaspeters:
### Problem
`uniffi-generators v0.28.3-3` had an issue where in C# the GC could move stuff around in memory, causing callbacks to no longer be where expected, causing segfaults.

### Solution
`uniffi-generators v0.28.3-4` fixes that issue, so we bump it here


matislovas (approved):
+2

matislovas (approved):

matislovas (approved):

PR #1435, opened 2025-09-12
Add networking to OpenWrt qemu VM

asledzinskiy:
- Set up bridging networking in container to access qemu VM outside of the container
- Add necessary VM ip and routing set-up

**Problem:**
With VM’s default -nic user mode, the VM cannot be reached directly from LAN.
It cannot forward traffic between networks so it's not possible to properly test routing or VPN passthrough.

**Solution:**
Use TAP devices and Linux bridges to connect the QEMU VM to the host’s networks:
Bridge connects the TAP device to the host network, allowing the VM to appear as if it’s physically connected to that network.
With at least two interfaces (eth0 on LAN, eth1 on WAN), the QEMU VM can forward packets between networks and act as a router/gateway.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/openwrt/container_net_setup.sh
+++ b/nat-lab/bin/openwrt/container_net_setup.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+set -Eeuox
+
+ETH_COUNT=$(ls /sys/class/net | grep -E '^eth[0-9]+$' | wc -l)
+ADD_ERR="Please add the following setting to your container:"
+
+tuntap="TUN device is missing. $ADD_ERR --device /dev/net/tun"
+WAN_IP="10.0.254.14/16"
+LAN_IP="192.168.115.254/24"
+
+if [ ! -c /dev/net/tun ]; then
+  error "$tuntap" && return 1
+fi
+
+for ((i = 0; i < ETH_COUNT; i++)); do
+  DOCKER_BRIDGE="dockerbridge$i"
+  NET_DEV="eth$i"
+  NET_TAP="qemu$i"
+  {
+    ip link add dev $DOCKER_BRIDGE type bridge
+    rc=$?
+  } || :
+
+  if ((rc != 0)); then
+    error "Failed to create bridge. $ADD_ERR --cap-add NET_ADMIN" && return 1
+  fi
+
+  # We need freshly created bridge to have IP address of the container
+  # For this reason we need to migrate IP from eth0 to dockerbridge.
+  for addr in $(ip --json addr show dev $NET_DEV | jq -c '.[0].addr_info[] | select(.family == "inet")'); do
+    cidr_addr=$(echo $addr | jq -r '[ .local, .prefixlen|tostring] | join("/")')
+    if ! ip addr add dev $DOCKER_BRIDGE $cidr_addr; then
+      error "Failed to add address for $DOCKER_BRIDGE interface"
+      exit 30
+    fi
+  done
+
+  if ! ip addr flush dev $NET_DEV; then
+    error "Failed to clear $NET_DEV interface addresses"
+    exit 30
+  fi
+
+  while ! ip link set $DOCKER_BRIDGE up; do
+    info "Waiting for IP address to become available..."
+    sleep 2
+  done
+
+  # QEMU Works with taps, set tap to the bridge created
+  if ! ip tuntap add dev "$NET_TAP" mode tap; then
+    error "$tuntap" && return 1
+  fi
+
+  while ! ip link set "$NET_TAP" up promisc on; do
+    info "Waiting for TAP to become available..."
+    sleep 2
+  done
+
+  if ! ip link set dev "$NET_TAP" master $DOCKER_BRIDGE; then
+    error "Failed to set IP link!" && return 1
+  fi
+
+  if ! ip link set dev "$NET_DEV" master $DOCKER_BRIDGE; then
+    error "Failed to attach docker interface to bridge"
+  fi
+done
+
+# deleting ips from the bridge to avoid ip conflicts
+# same ips will be assigned to VM interfaces
+ip addr del "$WAN_IP" dev dockerbridge0
+ip addr del "$LAN_IP" dev dockerbridge1
+
+ip link set dockerbridge0 down
+ip link set dockerbridge0 up
+ip link set dockerbridge1 down
+ip link set dockerbridge1 up
--- a/nat-lab/docker-compose.yml
+++ b/nat-lab/docker-compose.yml
@@ -490,8 +490,11 @@ services:
     cap_add:
       - NET_ADMIN
       - NET_RAW
+      - SYS_ADMIN
     security_opt:
       - no-new-privileges:true
+    devices:
+      - "/dev/net/tun"
     networks:
       internet:
         ipv4_address: 10.0.254.14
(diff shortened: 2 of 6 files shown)


nat-lab/openwrt.Dockerfile
```diff
@@ -30,14 +31,13 @@ RUN ln -s /opt/bin/openwrt/10-usbmount-initsh.sh    /usr/local/share/vmconfig/vm
...
     ln -s /opt/bin/openwrt/30-wait-for-network.sh   /usr/local/share/vmconfig/vm.d/30-wait-for-network.sh && \
     ln -s /opt/bin/openwrt/40-dns.sh                /usr/local/share/vmconfig/vm.d/40-dns.sh && \
+    ln -s /opt/bin/openwrt/50-set-ips.sh            /usr/local/share/vmconfig/vm.d/50-set-ips.sh && \
     ln -s /opt/bin/openwrt/99-install-ipks.sh       /usr/local/share/vmconfig/vm.d/99-install-ipks.sh && \
     ln -s /opt/bin/openwrt/serialize-vm-config.sh   /usr/local/bin/serialize-vm-config.sh && \
     ln -s /opt/bin/openwrt/send-config-to-vm.sh     /usr/local/bin/send-config-to-vm.sh && \
     ln -s /opt/bin/openwrt/run-vm.sh                /usr/local/bin/run-vm.sh && \
-    ln -s /opt/bin/openwrt/entrypoint.sh            /usr/local/bin/entrypoint.sh
-
-HEALTHCHECK --interval=30s --timeout=5s --retries=3 CMD \
```
asledzinskiy, line 39: I've deleted that HealthCheck because container IPs are migrated to VM and we can't test ssh from the container.
The only way I see at the moment to check ssh to VM is ready by adding extra service to docker-compose that will be a dependency for openwrt-client:

openwrt-healthcheck:
    image: alpine:latest
    depends_on:
      openwrt-gw-01:
        condition: service_started
    network_mode: host  # so it can reach 192.168.115.254
    command: >
      sh -c "
        apk add --no-cache openssh-client >/dev/null 2>&1 &&
        for i in $(seq 1 10); do
          ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o BatchMode=yes -o ConnectTimeout=5 root@192.168.115.254 'echo ok' && exit 0
          echo 'SSH attempt $i failed, retrying...';
          sleep 20;
        done;
        echo 'SSH failed after 5 attempts' >&2;
        exit 1
      "
    restart: "no"

olekoliinyk:
Can you please fill the pr description with problem and solution sections?

nat-lab/docker-compose.yml
```diff
@@ -490,6 +490,8 @@ services:
     cap_add:
       - NET_ADMIN
       - NET_RAW
+      - SYS_ADMIN
+      - MKNOD
```
olekoliinyk: Why do we need it (MKNOD)?

nat-lab/openwrt.Dockerfile
```diff
@@ -2,6 +2,7 @@ ARG LIBTELIO_ENV_SEC_CONTAINER_REGISTRY
 
 FROM ${LIBTELIO_ENV_SEC_CONTAINER_REGISTRY}/low-level-hacks/vpn/client/libtelio-build/openwrt-24.10.2-x86-64:v0.0.1
 
+RUN apk add --no-cache bash iproute2 jq
```
olekoliinyk: I'm wondering maybe we can move this to base image (openwrt-24.10.2-x86-64:v0.0.1)?

nat-lab/bin/openwrt/container_net_setup.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+set -Eeuox
+
+ETH_COUNT=$(ls /sys/class/net | grep -E '^eth[0-9]+$' | wc -l)
+ADD_ERR="Please add the following setting to your container:"
+
+tuntap="TUN device is missing. $ADD_ERR --device /dev/net/tun"
+
+# Create the necessary file structure for /dev/net/tun
+if [ ! -c /dev/net/tun ]; then
```
olekoliinyk: If we will add to openwrt service in the docker-compose the following
```
    devices:
      - "/dev/net/tun"
```
Would we need this section?

nat-lab/bin/openwrt/container_net_setup.sh
```diff
@@ -0,0 +1,81 @@
...
+  fi
+
+  if ! ip link set dev "$NET_DEV" master $DOCKER_BRIDGE; then
+    error "Failed to attach docker interface to bridge"
+  fi
+done
+
+# deleting ips from the bridge to avoid ip conflicts
+# same ips will be assigned to VM interfaces
+ip addr del 10.0.254.14/16 dev dockerbridge0
```
olekoliinyk: Maybe we can move those ip addresses to variables on top of the script?

nat-lab/bin/openwrt/container_net_setup.sh
asledzinskiy: Let me test that

nat-lab/bin/openwrt/container_net_setup.sh
asledzinskiy: Yes, agree

nat-lab/docker-compose.yml
asledzinskiy: We execute this command in the container 'mknod /dev/net/tun c 10 200' - it fails without MKNOD

nat-lab/openwrt.Dockerfile
asledzinskiy: Yep, agree. I'll re-publish base image with those packages

nat-lab/docker-compose.yml
asledzinskiy: Deleted MKNOD after adding:
    devices:
      - "/dev/net/tun"

nat-lab/openwrt.Dockerfile
olekoliinyk: Thanks

asledzinskiy:
> I don't fully understand the problem and the solution. Do you want OpenWRT VM to be accessible from other containers or what is the goal?

Yes, it needs to be accessible as a gateway for OpenWrt client and also from the host to be able to ssh on VM while executing nat-lab tests.
That approach mostly is a replication of Vagrantless approach to Mac and windows VMs

nat-lab/openwrt.Dockerfile
Jauler: I wonder, once we have removed the healthcheck, is there anything, that waits for OpenWRT to be ready before tests start?

asledzinskiy: No, I was going to add it as part of my next PR with test itself. At the moment I have openwrt-healthcheck that I described above. But also want to investigate adding wait in the test itself

Jauler: In order to avoid this race condition, IMHO we have two options:
- add some other IP address to the container itself (e.g. `192.168.115.253`) which reenables `ssh` again. But we have to be careful to make sure, that no other container ever gets this IP
- we can do something simple, like `nlx-01` does, where it creates a file `/ready` once all the scripts are finished exeucting succesfully

Jauler: @asledzinskiy mentioned, that healthcheck will be added with next PR, as now there are no tests dependant on readyness of the GW yet

Jauler (approved):
+1.0

PR #1438, opened 2025-09-15
Postquantum TLS for ENS

tomaszklak:
This change adds post-quantum TLS for the ENS using `X25519MLKEM768` key exchange algorithm. This is done using `aws-lc-rs` which is a wrapper over C library `aws-lc` and is a reason for some additional work in the build system to get it to (cross) compile:
- aws-lc-rs seem lack support for the less common architectures (eg. mips) so for now ENS is disabled in `teliod` since it has no use for it right now and is the only binary needed for openwrt
- the build system for the `aws-lc-rs` seem to be somewhat fragile which is why exact version of the lib is used. This version compiles and works everywhere except for windows. The windows version is switched at build time in the libtelio-build.
- last minute small change to the naming in the RFC was requested by server side (`Challenge` -> `ChallengeResponse`)


.github/workflows/build.yml
```diff
@@ -85,9 +85,9 @@ jobs:
             target_os: windows
     runs-on: ubuntu-22.04
     container: |-
-      ${{ matrix.target_os == 'linux' && 'ghcr.io/nordsecurity/build-linux-rust1.89.0:v6.3.0' ||
-          matrix.target_os == 'android' && 'ghcr.io/nordsecurity/build-android-rust1.89.0:v6.3.0' ||
-          'ghcr.io/nordsecurity/build-windows-rust1.89.0:v6.3.0' }}
+      ${{ matrix.target_os == 'linux' && 'ghcr.io/nordsecurity/build-linux-rust1.89.0:v7.2.0' ||
+          matrix.target_os == 'android' && 'ghcr.io/nordsecurity/build-android-rust1.89.0:v7.2.0' ||
+          'ghcr.io/nordsecurity/build-windows-rust1.89.0:v7.2.0' }}
```
tomaszklak: TODO: update after https://github.com/NordSecurity/rust_build_utils/pull/64 is merged

.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v3.0.0
+TRIGGERED_REF=LLT-6469-cmake
```
tomaszklak: TODO: Update before merge

ci/env.py
```diff
@@ -4,9 +4,9 @@
 
 LIBTELIO_ENV_MOOSE_RELEASE_TAG = "v17.0.0-libtelioApp"
 LIBTELIO_ENV_NAT_LAB_DEPS_TAG = "v0.0.32"
-LIBTELIO_ENV_ANDROID_BUILDER_TAG = "v6.3.0"
-LIBTELIO_ENV_LINUX_BUILDER_TAG = "v6.3.0"
-LIBTELIO_ENV_WINDOWS_BUILDER_TAG = "v6.3.0"
+LIBTELIO_ENV_ANDROID_BUILDER_TAG = "v7.2.0"
+LIBTELIO_ENV_LINUX_BUILDER_TAG = "v7.2.0"
+LIBTELIO_ENV_WINDOWS_BUILDER_TAG = "v7.2.0"
```
tomaszklak: TODO: update after https://github.com/NordSecurity/rust_build_utils/pull/64 is merged

linkers/d-arm-linux-gnueabi-ar
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
mathiaspeters: Are these linker files really necessary?

crates/telio-proto/src/ens.rs
```diff
@@ -263,42 +282,136 @@ fn authentication_interceptor(
...
+
+    if allow_only_mlkem {
+        provider.kx_groups =
+            vec![tokio_rustls::rustls::crypto::aws_lc_rs::kx_group::X25519MLKEM768];
+    }
+
+    let mut tls_config = ClientConfig::builder_with_provider(Arc::new(provider))
+        .with_safe_default_protocol_versions()
+        .map_err(|e| std::io::Error::other(format!("{e}")))?
+        .dangerous()
```
mathiaspeters, line 408: Is this here so it works for tests? Or is it needed for prod as well?

linkers/d-arm-linux-gnueabi-ar
tomaszklak: It might not be needed if we wouldn't use custom compilers (eg. `d-arm-linux-gnueabi-gcc`). Since we use those, the `aws-lc` build system derives the name of `ar` from it (eg. `d-arm-linux-gnueabi-ar`), which we also need to provide now.

crates/telio-proto/src/ens.rs
tomaszklak: It is needed in prod, since in prod we won't verify the certs.

linkers/d-arm-linux-gnueabi-ar
mathiaspeters: I'm currently removing the previous custom linkers, like `d-arm-linux-gnueabi-gcc`, which is why I'm asking if these are necessary

tomaszklak: It's hard for me to predict to be honest. I would hope that removing wrappers with `d-` prefix will also mean that correct `ar` will be picked up automatically but I'm not completely sure - the `aws-lc-rs` build system turned out very finicky in practice :(

matislovas:
+1

mathiaspeters (approved):
+1

PR #1440, opened 2025-09-15
LLT-6641: Use recommended server as default

lcruz99:
### Solution


clis/teliod/src/core_api.rs
```diff
@@ -442,6 +436,79 @@ async fn get_recommended_servers(
     }
 }
 
+pub async fn get_server_endpoint(config: &TeliodDaemonConfig) -> Result<Endpoint, Error> {
+    let country_id: Option<u64> = match &config.vpn {
+        // Use the endpoint directly if provided in the config
+        VpnConfig::Server(endpoint) => return Ok(endpoint.clone()),
+        // Find VPN exit node based on country
+        VpnConfig::Country(target_country) => {
+            if !(target_country.len() == 2
```
tomasz-grz: Why are we checking for the `length == 2` here? It means that `"country": "poland"` will not work,
But good point checking `is_ascii_alphabetic`

clis/teliod/src/core_api.rs
```diff
@@ -375,15 +375,13 @@ pub async fn get_recommended_servers_with_exp_backoff(
                 wait_with_backoff_delay(&mut backoff, &mut retries).await?;
             }
             Err(e) => {
-                // other error, return immedately
```
tomasz-grz, line 378: Are you waging war on code comments? 😅 🪓

clis/teliod/src/core_api.rs
```diff
@@ -442,6 +436,79 @@ async fn get_recommended_servers(
...
+                        }
+                    }),
+                    Err(e) => {
+                        error!("Getting countries failed due to: {e}");
+                        None
+                    }
+                }
+            }
+        }
+        VpnConfig::Recommended => None,
```
tomasz-grz, line 489: Maybe log here instead?
```rust
trace!("Using a recommended server");
```

clis/teliod/src/core_api.rs
```diff
@@ -442,6 +436,79 @@ async fn get_recommended_servers(
...
+    };
+
+    if country_id.is_none() {
```
tomasz-grz, line 492: Since we could have `VpnConfig::Recommended`, what do you think about removing this check completely, since it's neither an error nor a warning.

And we already log above when failing to find a country code
```rust
```

clis/teliod/src/core_api.rs
```diff
@@ -442,6 +436,79 @@ async fn get_recommended_servers(
     }
 }
 
+pub async fn get_server_endpoint(config: &TeliodDaemonConfig) -> Result<Endpoint, Error> {
```
tomasz-grz: Looking a bit into the future, we would rather want to return the whole list of recommended servers from `core_api.rs` to `daemon.rs` rather than a single one.

Since we want to implement server selection logic (I guess this one we could implement here, but I would still prefer it to be the responsibility of the daemon module) and retry flow next (would be harder to implement here, since we would need to know which servers failed connecting).

clis/teliod/src/core_api.rs
lcruz99: AFAIK `target_country` is following the [ISO 3166](https://www.iso.org/iso-3166-country-codes.html) alpha-2.

lcruz99: Alpha-2 code – a two-letter code that represents a country name, recommended as the general purpose code

clis/teliod/src/core_api.rs
lcruz99: 😈  thought some comments were trivial

clis/teliod/src/core_api.rs
lcruz99: Log has higher relevance when the VpnConfig::Country(_) is configured but it falls back to the recommended server. I've decided to log on both cases

clis/teliod/src/core_api.rs
lcruz99: I disagree, the log you mention warns that the server with the country code was not found and the other informs to that we're not connecting to the server the user requested but using the first recommended server instead.

clis/teliod/src/core_api.rs
lcruz99: fair enough, changed

clis/teliod/src/core_api.rs
lcruz99: I've just seen that full country names are also checked: https://github.com/NordSecurity/libtelio/blob/0cb3bfccc34545c51f3fdeb08e7457050fdc933b/clis/teliod/src/core_api.rs#L93.

However I think we should stick to alpha-2 codes. From ISO 3166:
> Using codes saves time and avoids errors as instead of using a country’s name (which will change depending on the language being used), we can use a combination of letters and/or numbers that are understood all over the world.

tomasz-grz: It's full name **or** ISO alpha-2

https://github.com/NordSecurity/libtelio/blob/5bc97ac6489960317e8cf542c2e22f7efb0cf5c2/clis/teliod/src/core_api.rs#L90-L95

Maybe here a comment would have been more relevant 💭

lcruz99: ok, so you think we should keep as is and support both?

tomasz-grz: Hmm I would say so, since we get the data from the API for both, it is rather trivial to support both.

But maybe I am missing something (like unicode case..), but since we're not supporting localization anywhere either way.. 🤷 Maybe double check with PO or EM?

clis/teliod/src/core_api.rs
tomasz-grz: Maybe then from `warn!("Using a recommended server");` then instead we could make it `debug!("Using a recommended server");`?

Because it will appear even if the user selects VpnConfig::Recommended, as if it's some sort of problem

lcruz99: Not debug as the users running release builds will miss it.

In that case I'll change it to info log level.

clis/teliod/src/core_api.rs
lcruz99: Ok, I'll remove `target_country` length check then.

clis/teliod/src/core_api.rs
```diff
@@ -442,6 +451,64 @@ async fn get_recommended_servers(
...
+pub async fn get_server_endpoints_list(
+    config: &TeliodDaemonConfig,
+) -> Result<Vec<Endpoint>, Error> {
+    let country_id: Option<u64> = match &config.vpn {
+        // Use the endpoint directly if provided in the config
+        VpnConfig::Server(endpoint) => return Ok(vec![endpoint.clone()]),
+        // Find VPN exit node based on country
+        VpnConfig::Country(target_country) => {
+            if !target_country.chars().all(|c| c.is_ascii_alphabetic()) {
+                warn!(
```
tomasz-grz, line 463: I guess this warn is also not relevant anymore

PR #1441, opened 2025-09-15
LLT-5597: Inverval timer for LocalNameServer

tomasz-grz:
### Problem
The [update_timers](https://github.com/NordSecurity/libtelio/blob/v5.1.3/crates/telio-dns/src/nameserver.rs#L97) loop is blocked by the `.recv()` call.
In cases there is no DNS activity for a long time, the Wireguard Session may expire, and will only be re-established on the next DNS request.

### Solution
Add an interval timer `idle_timer`, that periodically (once a `IDLE_TIME_SECONDS` by default 1s) calls `update_timers()` whenever there is no DNS traffic activity.
This should ensure that all the Wireguard keep-alive and handshake messages are transmitted even when there are no DNS requests.

Other refactoring:
- replace `semaphore.try_acquire()` with `semaphore.acquire().await` to fairly queue requests instead of dropping them.
- Add per task scratch arrays, and capture only exact needed bytes from `receiving_buffer` instead of cloning the whole vector in a loop.
```rust
let mut receiving_buffer = vec![0u8; MAX_PACKET];
let mut receiving_buffer = receiving_buffer.clone();
```

```rust
let mut receiving_buffer = vec![0u8; MAX_PACKET];
let in_bytes = Vec::from(receiving_buffer.get(..bytes_read).unwrap_or_default());
```

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/tests/nameserver.rs
+++ b/crates/telio-dns/tests/nameserver.rs
@@ -39,7 +39,6 @@ const MAX_PACKET: usize = 2048;
 
 struct WGClient {
     client_socket: tokio::net::UdpSocket,
-    _client_address: SocketAddr,
     server_address: SocketAddr,
     tunnel: Arc<Mutex<Tunn>>,
 }
@@ -54,15 +53,9 @@ impl WGClient {
             tokio::net::UdpSocket::bind(SocketAddr::from_str("127.0.0.1:0").unwrap())
                 .await
                 .expect("Failed to bind client socket");
-        let client_port = client_socket
-            .local_addr()
-            .expect("Failed to get client local address")
-            .port();
-        let _client_address = ([127, 0, 0, 1], client_port).into();
         let client_private_key = client_secret_key;
         WGClient {
             client_socket,
-            _client_address,
             server_address,
             tunnel: Arc::new(Mutex::new(
                 Tunn::new(client_private_key, server_public_key, None, None, 0, None)
@@ -88,7 +81,7 @@ impl WGClient {
                     .expect("Failed to send handshake message");
             }
             TunnResult::Err(e) => panic!("Encapsulate error: {:?}", e),
-            _ => println!("Unexpected TunnResult during handshake phase"),
+            _ => panic!("Unexpected TunnResult during handshake phase"),
         }
 
         let bytes_read = self
@@ -133,7 +126,7 @@ impl WGClient {
                     .expect("Failed to send the dns request");
             }
             TunnResult::Err(e) => panic!("Encapsulate error: {:?}", e),
-            _ => println!("Unexpected TunnResult while sending the dns request"),
+            _ => panic!("Unexpected TunnResult while sending the dns request"),
         }
 
         let result = timeout(
@@ -381,10 +374,6 @@ impl WGClient {
         buffer.truncate(total_length);
         buffer
     }
-
-    fn _client_address(&self) -> SocketAddr {
-        self._client_address
-    }
 }
 
 #[derive(Copy, Clone, Debug)]
@@ -469,16 +458,16 @@ async fn dns_test_with_server(
     nameserver: Arc<RwLock<LocalNameServer>>,
     ttl_value: TtlValue,
 ) {
-    let client = init_client(local_records, nameserver, ttl_value).await;
+    let (client, _) = init_client_and_server(local_records, nameserver, ttl_value).await;
     client.do_handshake().await;
     client.send_dns_request(query, test_type).await;
 }
 
-async fn init_client(
+async fn init_client_and_server(
     local_records: Option<(String, Records)>,
     nameserver: Arc<RwLock<LocalNameServer>>,
     ttl_value: TtlValue,
-) -> WGClient {
+) -> (WGClient, Arc<Mutex<Tunn>>) {
     if let Some((zone, records)) = local_records {
         nameserver
             .upsert(&zone, &records, ttl_value)
@@ -509,11 +498,9 @@ async fn init_client(
 
     let client = WGClient::new(client_secret_key, server_public_key, server_address).await;
 
-    nameserver
-        .start(server_peer.clone(), server_socket.clone())
-        .await;
+    nameserver.start(server_peer.clone(), server_socket).await;
 
-    client
+    (client, server_peer)
 }
 
 #[tokio::test]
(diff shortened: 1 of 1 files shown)


crates/telio-dns/tests/nameserver.rs
```diff
@@ -39,7 +39,7 @@ const MAX_PACKET: usize = 2048;
 
 struct WGClient {
     client_socket: tokio::net::UdpSocket,
-    client_address: SocketAddr,
+    _client_address: SocketAddr,
```
Jauler: is this rename intentional? :)

crates/telio-dns/tests/nameserver.rs
```diff
@@ -756,3 +747,70 @@ async fn test_tcp_rst() {
...
     assert_eq!(error_kind, ErrorKind::ConnectionRefused);
 }
+
+#[tokio::test]
+async fn test_timers_updated() {
+    let nameserver = LocalNameServer::new(&[IpAddr::V4(Ipv4Addr::new(8, 8, 8, 8))])
+        .await
+        .expect("Failed to create a LocalNameServer");
+
+    let server_socket = tokio::net::UdpSocket::bind(SocketAddr::from_str("127.0.0.1:0").unwrap())
```
Jauler: Would it be possible to reuse `init_client()`? this seems like it is duplicated code

crates/telio-dns/src/nameserver.rs
```diff
@@ -81,129 +108,134 @@ impl LocalNameServer {
...
-                                .decapsulate(None, &[], &mut temporary_buffer)
-                        {
-                            if let Err(e) = socket.send_to(empty, sender_addr).await {
-                                telio_log_warn!("[DNS] Failed to send handshake packet  : {:?}", e);
-                                return;
-                            };
+        loop {
+            tokio::select! {
+                res = socket.recv_from(&mut receiving_buffer) => {
+                    let (bytes_read, sender_addr) = match res {
```
Jauler, line 121: I guess this arm really calls for some refactoring to split it into a few parts/functions.
For example:
1. WG-transport handling (receive)
2. WG decapsulation
3. DNS request itself

But probably it make sense to do that in separate PRs, not mangle into one. As it is unrelated to `.update_timer()` calls or LLT-5597.

tomasz-grz: Good idea, also prefer to keep it separate PR

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Yes to make clippy happy, since `client_address` is not used anywhere, but I didn't want to remove it completely in case it's useful in the future.

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Yes it is.. but `init_client` doesn't expose `server_peer` to enable persistent keepalives after initializing..

I tried to enable keepalives for the server in `init_client()` from the beginning, but then the first `send_dns_request()` assert fails because it is getting a keepalive packet instead of the DNS response.. But maybe for this test we can ignore the DNS response 🤔

tomasz-grz: Let me think if I can massage it in nicely

crates/telio-dns/src/nameserver.rs
```diff
@@ -35,6 +36,7 @@ const MAX_PACKET: usize = 2048;
 const UDP_HEADER: usize = 8;
 const TCP_MIN_HEADER: usize = 20;
 const MAX_CONCURRENT_QUERIES: usize = 256;
+const IDLE_TIME_SECONDS: u64 = 1;
```
jjanowsk: Duration has a const constructor. Use it here to avoid mistakes in the future. I've seen mistakes like this in the past.
```
const IDLE_TIME: Duration = Duration::from_secs(1);
```

crates/telio-dns/tests/nameserver.rs
jjanowsk: Remove it ;)

crates/telio-dns/src/nameserver.rs
```diff
@@ -81,129 +108,134 @@ impl LocalNameServer {
         nameserver: Arc<RwLock<LocalNameServer>>,
         socket: Arc<UdpSocket>,
     ) {
-        let mut receiving_buffer = vec![0u8; MAX_PACKET];
-        let mut sending_buffer = vec![0u8; MAX_PACKET];
+        let mut receiving_buffer = [0u8; MAX_PACKET];
```
jjanowsk: I think it's safer to have it on the heap

crates/telio-dns/src/nameserver.rs
```diff
@@ -81,129 +108,134 @@ impl LocalNameServer {
...
+
+                    let peer = peer.clone();
+                    let socket = socket.clone();
+                    let nameserver = nameserver.clone();
+                    let semaphore = semaphore.clone();
+                    let in_bytes = Vec::from(receiving_buffer.get(..bytes_read).unwrap_or_default());
+
+                    tokio::spawn(async move {
+                        let mut sending_buffer = [0u8; MAX_PACKET];
+                        let mut task_receiving_buffer = [0u8; MAX_PACKET];
```
jjanowsk: Please, place those also on the heap it's another 1kb. They anyway are moved to the heap by the tokio::spawn as a part of the future struct. The upside is that here you do it in one allocation if you move them to the heap explicitly there's gonna be 3 allocations but I think the risk is not worth the reward

jjanowsk (commented):
In general I think it could have also been fixed by setting `persistent_keepalive_interval` in `get_peer` function to e.g. 1 sec? This way the wireguard would drive nameserver peer. But maybe it's safer this way because the intent is clear.

jjanowsk:
> > In general I think it could have also been fixed by setting persistent_keepalive_interval in get_peer function to e.g. 1 sec? This way the wireguard would drive nameserver peer. But maybe it's safer this way because the intent is clear.
> 
> technically yes, then we would receive keepalives often, but IMHO sending network traffic (even if localhost) is not a the best way of triggering some loop in local process code.

I don't disagree, I mentioned in mostly as a funny idea, but I still prefer current solution. My choice of words was not the best ;)

crates/telio-dns/src/nameserver.rs
tomasz-grz: Good point

crates/telio-dns/src/nameserver.rs
tomasz-grz: Could you elaborate on the risks you see?
Expanding the stack size?

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Yeah I wanted to remove it first, but then I thought it might be useful in the future (since somebody put it there in the fist place)

tomasz-grz: Removed

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Refactored this bit a little, I just exposed the server tun from `init_client()`

crates/telio-dns/src/nameserver.rs
tomasz-grz: Replaced

crates/telio-dns/src/nameserver.rs
tomasz-grz: Back on Heap

crates/telio-dns/src/nameserver.rs
tomasz-grz: Also moved to heap allocated

jjanowsk: Yeah, overflowing the stack is the reason. I know that maybe it's becoming less and less relevant on most platforms but when I see allocations in kilobytes I tend to lean towards heap.

PR #1444, opened 2025-09-16
LLT-6524 Rename teliod to NordVPN Lite

jjanowsk:


tomaszklak (commented):
+1, unless teliod had been already released somewhere. In that case since we are changing some of the paths, we should have a migration prepared.

.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v3.1.0
+TRIGGERED_REF=LLT-6524-rename-teliod
```
matislovas: reminder: update ref

matislovas (dismissed):

matislovas (approved):

PR #1448, opened 2025-09-17
Fix teliod tunnel network mask

tomasz-grz:
### Problem
After removal of meshnet, the tunnel interface address changed from the meshnet range `100.64.0.0/10` to a static `10.5.0.2` but the network mask is still set for the meshnet range.
This breaks connectivity in Class A `10.0.0.0` subnets, like our "public internet" in natlab tests.

### Solution
Change the tunnel interface network mask from `255.192.0.0` to `255.255.255.252`

### DIFF
--- a/clis/teliod/src/interface.rs
+++ b/clis/teliod/src/interface.rs
@@ -139,7 +139,7 @@ impl ConfigureInterface for Ifconfig {
 
     fn set_ip(&mut self, ip_address: &IpAddr) -> Result<(), TeliodError> {
         let ip_string = ip_address.to_string();
-        let cidr_suffix = if ip_address.is_ipv4() { "/10" } else { "/64" };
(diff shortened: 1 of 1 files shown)


gytsto (commented):
+1

clis/nordvpnlite/src/interface.rs
lcruz99, line 1: What's the reason for /30 instead of a /32 netmask?

tomasz-grz: If I'm not mistaken, the server peer is at `10.5.0.1`, therefore network `10.5.0.0/30` would give a usable hosts range from `10.5.0.1` to `10.5.0.2`.

If we set the interface to `10.5.0.2/32`, we could loose routing to `10.5.0.1`. But we could also just use `10.5.0.0/16`, as some of the apps do.

lcruz99 (approved):
+1

PR #1449, opened 2025-09-18
Loop log check in teliod test

mathiaspeters:
### Problem
With vagrantless natlab it seems like there's sometimes a small delay between teliod being started and logs being available. It happens in every nightly but only a couple of the pipelines there, so not super common

### Solution
Loop the log check until the expected logs are there, to account for any delays

- [x] README.md is updated


lcruz99 (approved):
+1

PR #1450, opened 2025-09-18
test(nat-lab): Use process.is_done() in run fail

gytsto:
### Problem
Sleep makes test flaky

### Solution
Replace the arbitrary sleep with process.is_done() in the failing-run test to make it deterministic and avoid timing-related flakiness. This ensures ProcessExecError and the return code are captured reliably across supported OS targets.


mathiaspeters:
+1

LukasPukenis (commented):
+1

PR #1452, opened 2025-09-18
Fix local host natlab VM setup

lcruz99:
### Problem
At the moment Windows VM cannot communicate with natlab "internet" network containers due to restrictive nat.
On docker versions >28 the default mode will drop every packet destined to this network that doesn't come from its bridge interface. Docker appends several iptables rules to the `PREROUTING` chain of the `raw` table, one for each host, eg:
```DROP       all  --  !br-123456789abc any anywhere 10.0.10.1
DROP       all  --  !br-123456789abc any anywhere 10.0.10.2
...```

### Solution
Set `com.docker.network.bridge.gateway_mode_ipv4: "nat-unprotected"` for this network.

### CODE REVIEW

File: nat-lab/natlab.py [hunk 1]
```diff
@@ -40,13 +40,13 @@ def start():
...
-    disabled_port_mapping = "ports: []"
     with open("docker-compose.yml", "r", encoding="utf-8") as file:
         filedata = file.read()
-    if original_port_mapping not in filedata:
-        raise RuntimeError("Cannot find expected port mapping compose file")
     if "GITLAB_CI" in os.environ:
+        original_port_mapping = 'ports: ["58001"]'
+        disabled_port_mapping = "ports: []"
+        if original_port_mapping not in filedata:
+            raise RuntimeError("Cannot find expected port mapping compose file")
```
stalowyjez, line 49: Is it somehow connected to this "nat-unprotected" change, or just a random refactor?

stalowyjez (approved):
Looks good, +1

nat-lab/natlab.py
lcruz99: This check was running (and failing) locally when it should only run on CI, not connected to the main issue this PR tackles. I was lazy to not create a separate PR for it, if you desire I still can do it.

stalowyjez (approved):

PR #1453, opened 2025-09-18
Backport uniffi fix

mathiaspeters:
### Problem
A bug was found in version 0.28 of the uniffi C# generator. The solution is to bump the uniffi generator version. This also bumps rust_build_util, which requires cherry picking another commit that adds control flow guard for windows, which they will be happy about.

### Solution
Cherry pick commits

- [x] README.md is updated


.unreleased/LLT-6598
```diff
@@ -0,0 +1 @@
+Bump uniffi-generators to v0.28.3-4 to fix windows issue where GC might relocate callbacks, causing segfaults
```
stalowyjez: Should we have this in both `changelog.md` and `.unreleased`?

mathiaspeters: I didn't think about that at all while cherry-picking, but good point! I removed from `.unreleased`

ci/env.py
```diff
@@ -3,11 +3,11 @@
...
 LIBTELIO_ENV_MOOSE_RELEASE_TAG = "v17.0.0-libtelioApp"
-LIBTELIO_ENV_NAT_LAB_DEPS_TAG = "v0.0.30"
-LIBTELIO_ENV_ANDROID_BUILDER_TAG = "v0.0.1"
-LIBTELIO_ENV_LINUX_BUILDER_TAG = "v0.0.3"
-LIBTELIO_ENV_WINDOWS_BUILDER_TAG = "v0.0.1"
-LIBTELIO_ENV_UNIFFI_GENERATORS_TAG = "v0.28.3-3"
+LIBTELIO_ENV_NAT_LAB_DEPS_TAG = "v0.0.32"
+LIBTELIO_ENV_ANDROID_BUILDER_TAG = "v6.3.0"
+LIBTELIO_ENV_LINUX_BUILDER_TAG = "v6.3.0"
+LIBTELIO_ENV_WINDOWS_BUILDER_TAG = "v6.3.0"
```
Jauler, line 9: is it v6.3.0? I see that control flow guard was added in `v6.4.0` :thinking: 

and submodule points to v7.1.1

mathiaspeters: The PR that's supposed to add CFG to `main` didn't update these, and I just cherry picked the commit :thinking:

mathiaspeters: And looking at the [changes in rust_build_utils](https://github.com/NordSecurity/rust_build_utils/pull/60/files) The docker images don't really need to be updated, but it still happened as part of the cherry-picking

Jauler (approved):
+1.0

PR #1454, opened 2025-09-19
Fix sqlite timeout by splitting it

matislovas:
### Problem
We caught an error, where `sqlite3` binary and `libmoose` accessed db file and caught a deadlock.

### Solution
Offloading some of `busy_timeout` to python-side retries.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/telio.py
+++ b/nat-lab/tests/telio.py
@@ -36,7 +36,7 @@ from utils.connection import Connection, TargetOS
 from utils.connection.docker_connection import DockerConnection, container_id
 from utils.connection_util import get_uniffi_path
 from utils.logger import log
-from utils.moose import MOOSE_LOGS_DIR
+from utils.moose import MOOSE_DB_TIMEOUT_MS, MOOSE_LOGS_DIR
 from utils.output_notifier import OutputNotifier
 from utils.process import Process, ProcessExecError
 from utils.python import get_python_binary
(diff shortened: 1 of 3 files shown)


nat-lab/tests/telio.py
```diff
@@ -1045,19 +1045,27 @@ async def maybe_write_device_fingerprint_to_moose_db(self):
...
+                            f"PRAGMA busy_timeout = {max_timeout};",
+                            (
+                                "INSERT OR REPLACE INTO shared_context (key, val, is_essential) VALUES"
+                                f" ('device.fp._string', '\"{fingerprint}\"', 1)"
+                            ),
+                        ],
+                        quiet=True,
+                    ).execute()
+                    return
+                except Exception:  # pylint: disable=broad-exception-caught
```
gytsto: could we print an exception? and also what will happen if max_retries run out, no exception, no nothing?

nat-lab/tests/test_lana.py
```diff
@@ -171,19 +171,26 @@ async def get_moose_db_file(
...
+                    "--cmd",
+                    f"PRAGMA busy_timeout = {max_timeout};",
+                    f".backup {container_backup_path}",
+                ],
+                quiet=True,
+            ).execute(privileged=True)
+
+            await connection.download(container_backup_path, local_path)
+            return
+        except Exception:  # pylint: disable=broad-exception-caught
```
gytsto: same here

nat-lab/tests/telio.py
matislovas: oops, fixed

nat-lab/tests/test_lana.py
matislovas: fixed

nat-lab/tests/telio.py
```diff
@@ -1045,19 +1045,32 @@ async def maybe_write_device_fingerprint_to_moose_db(self):
...
+                        quiet=True,
+                    ).execute()
+                    return
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    print(
+                        f"maybe_write_device_fingerprint_to_moose_db error: {e}, retrying ..."
+                    )
+                    max_retries -= 1
+            if not max_retries:
+                print(f"Retries exhausted, while trying to write fingerprint to db")
```
gytsto: But shouldn't this raise an exception of some sorts? It fails -> prints something -> does nothing else.
call sites wouldn't know if anything is wrong or not

matislovas: If no file is downloaded, it will fail in `fetch_moose_events`. I mean I can add it here, but didn't seem necessary.

gytsto: yeah, but we can fail before :smile: this would make it more clear when debugging, which place failed

matislovas: I just wanted to leave this as it was, but ok will add

Jauler (dismissed):
Good point from Gytis about exceptions. But it is already raised. Other than that one - +1.0

nat-lab/tests/telio.py
gytsto: Don't wanna be pain in the a$$, but you forgot other function as well xD

gytsto (approved):

PR #1455, opened 2025-09-19
Fix a crash when https is attempted with the default crypto provider

tomaszklak:
When derp connects to http**s** it uses the default crypto provider. If there is only one configured at build time (either ring or aws-lc-rs) this works fine. But if both are present at compile time, rustls will refuse to work without explicitly configuring the one which should be the default.


LukasPukenis (commented):
+1

Looks like a nice fix, though Cargo.lock is a bit unexpected to have this many changes :cold_sweat: 

Small question - is the conditional to use one or another just because of 32bit support(teliod/openwrt)?

tomaszklak:
> is the conditional to use one or another just because of 32bit support(teliod/openwrt)?

Yes, it's due to problem with getting aws-lc-rs to compile on openwrt (at least mips, but also possibly the mediatek architecture).

crates/telio-proto/fuzz/Cargo.toml
```diff
@@ -11,6 +11,7 @@ cargo-fuzz = true
 
 [dependencies]
 libfuzzer-sys = "0.4"
+rustls = { version = "0.23", default-features = false, features = ["ring", "std"] }
```
mathiaspeters, line 14: Can we do `rustls.workspace = true` here? I don't remember if the fuzzers are part of the workspace

mathiaspeters (approved):
+1

crates/telio-proto/fuzz/Cargo.toml
tomaszklak: Tried it, it doesn't work :( probably because of this in fuzzing Cargo.toml

```toml
# Prevent this from interfering with workspaces
[workspace]
members = ["."]
```

PR #1456, opened 2025-09-19
Add nat-lab OpenWrt test on vpn connection

asledzinskiy:
- Cover scenario with connecting OpenWrt device to vpn
- Add healthcheck to wait for ssh to start on VM

- [x] README.md is updated


nat-lab/bin/openwrt/99-install-ipks.sh
```diff
@@ -6,7 +6,7 @@ mkdir -p /mnt/usb
 echo "Mounting vfat USB..."
 mount -t vfat /dev/vdb /mnt/usb && echo "Mount success" || { echo "Mount failed!"; ls /dev; exit 2; }
 ls -l /mnt/usb
-mkdir -p /etc/ssl/certs
-[ -f /mnt/usb/test.pem ] && mv /mnt/usb/test.pem /etc/ssl/certs/ || echo "test.pem not found on USB!"
+mkdir -p /etc/ssl/server_certificate
```
asledzinskiy, line 9: Changing to 'server_certificate' as it is hardcoded in some parts of teliod.py

nat-lab/bin/openwrt/container_net_setup.sh
```diff
@@ -6,12 +6,75 @@ ADD_ERR="Please add the following setting to your container:"
...
 WAN_IP="10.0.254.14/16"
-LAN_IP="192.168.115.254/24"
+LAN_IP="192.168.115.254"
+LAN_IP_MASK="24"
 
 if [ ! -c /dev/net/tun ]; then
   error "$tuntap" && return 1
 fi
 
+find_free_ip() {
```
asledzinskiy: Finding available ip to assign to bridge to be able to check ssh is ready on VM.
Function is re-used from docker_linux repo

nat-lab/tests/teliod.py
```diff
@@ -53,7 +54,7 @@ class Paths:
     run_dir: Path = Path("/run")
 
     def __post_init__(self):
-        if os.environ.get("PYTEST_CURRENT_TEST"):
+        if os.environ.get("PYTEST_CURRENT_TEST") and "teliod" in self.exec_path.parts:
```
asledzinskiy: As we only have OpenWrt package installed in VM and not on the host - modify this statement to be executed for teliod only, not 'nordvpn' (OpenWrt)

nat-lab/tests/test_openwrt.py
```diff
@@ -0,0 +1,72 @@
...
+            gateway_connection,
+            exit_stack,
+            config=Config(IfcConfigType.VPN_OPENWRT, paths=config_path),
+        )
+        await teliod.request_credentials_from_core()
+
+        async with teliod.start():
+            log.debug("Teliod started, waiting for connected vpn state...")
+            try:
+                await teliod.wait_for_vpn_connected_state()
```
asledzinskiy: Sporadically this command fails with ClientTimeoutError. Added 1 retry

nat-lab/tests/utils/connection/ssh_connection.py
```diff
@@ -47,22 +49,29 @@ async def new_connection(
...
-            username="root" if tag is ConnectionTag.VM_MAC else "vagrant",
-            password="vagrant",
+            username=(
+                "root"
+                if tag in [ConnectionTag.VM_MAC, ConnectionTag.VM_OPENWRT_GW_1]
+                else "vagrant"
+            ),
+            password=None if tag in [ConnectionTag.VM_OPENWRT_GW_1] else "vagrant",
             known_hosts=None,
+            connect_timeout=15,
```
asledzinskiy, line 73: Even though we wait for ssh to be ready on VM - there was one time failure with ssh connection from tests so added those timeouts

nat-lab/bin/openwrt/container_net_setup.sh
Jauler: I think this is not the best approach, as it depends on the order of hosts coming up.  E.g. if this script is executed before some host using some IP is up - it may be considered as free, later on resulting in IP address clashes. We had such problem with a similar approach on Vagrantless natlab.

For the reason above, IMHO it is better to just have some deterministic IP address, which we reserve for this purpose intentionally, and thus - avoiding clashes.

nat-lab/data/nordvpnlite/config_openwrt_uci_pl_setup.json
```diff
@@ -0,0 +1,16 @@
+{
+  "log_level": "trace",
+  "log_file_path": "/var/log/teliod_natlab.log",
+  "log_file_count": 0,
+  "authentication_token": "48e9ef50178a68a716e38a9f9cd251e8be35e79a5c5f91464e92920425caa3d9",
```
Jauler, line 5: Can we make this token "obviously-for-testing-purposes-only". Either by value, or by some comment. e.g. something like `aaaaaaaaaaaaaaa....aaaaaaaaaaaa` or similar

As otherwise we likely will get hackerone report about leaked secrets, even though the token is not valid one.

nat-lab/tests/test_openwrt.py
Jauler: IMHO retry is not good idea in this place. As sporadical failures indicate some sort of race condition. 

We should figure out root cause for it, and then fix it properly, as retries will just hide potential problems.

nat-lab/tests/utils/connection/ssh_connection.py
```diff
@@ -119,3 +128,14 @@ async def copy_binaries(self) -> None:
             await utils_win.copy_binaries(self._connection, self)
         elif self.target_os is TargetOS.Mac:
             await utils_mac.copy_binaries(self._connection, self)
+
+    async def upload_file(self, local_file_path: str, remote_file_path: str) -> None:
```
Jauler, line 132: This will only work with VMs or OpenWRT. Could we somehow indicate that to avoid confusion, e.g. for example if `self._connection` is not ssh connection - then throw not supported exception with some explanation, that this method is only supported for VMs

nat-lab/tests/test_openwrt.py
asledzinskiy: Yep, agree, I'll try to organize debug session

nat-lab/bin/openwrt/container_net_setup.sh
asledzinskiy: I'll use 192.168.115.253 then. 192.168.115.0/24 is used for openwrt only so hopefully we won't have a problem with deterministic ip

nat-lab/tests/utils/connection/ssh_connection.py
asledzinskiy: But this is concrete implementation in SshConnection class. There is 'upload_file' method in abstract Connection class.
I've added only implementation for SshConnection but implementation can also be added to DockerConnection if needed

nat-lab/data/nordvpnlite/config_openwrt_uci_pl_setup.json
asledzinskiy: I'd rather add a comment to mark it as a test token. If we want to change the token itself - I'd prefer a separate PR as it requires to change all teliod test configs we use

nat-lab/data/nordvpnlite/config_openwrt_uci_pl_setup.json
```diff
@@ -0,0 +1,16 @@
...
+  "adapter_type": "linux-native",
+  "http_certificate_file_path": "/etc/ssl/server_certificate/test.pem",
+  "interface": {
+    "name": "teliod",
+    "config_provider": "uci"
```
olekoliinyk, line 10: We have as well other config providers. 
Maybe we can parametrize the tests to test each of them?
But if you think that it can be addressed in a separate PR - it's ok then.

asledzinskiy: I think it makes sense to do it in this PR. 
@olekoliinyk And which providers make sense for OpenWrt?
I see currently we have 'iproute' and 'manual'

asledzinskiy: Discussed in private that we will test uci only at the moment but anyway I've added parametrization to the test for future

nat-lab/data/nordvpnlite/config_openwrt_uci_pl_setup.json
asledzinskiy: Json doesn't support comments so if we want to fix it I'd rather have it in a separate PR fixing it everywhere

Jauler: :ok_hand: it can be separate PR.

nat-lab/tests/utils/connection/ssh_connection.py
Jauler: oh sorry I missed the fact that this is `ssh_connection.py`.  But in that case, maybe it would be nice to avoid "silent failures" meaning, that the method gets called, but does nothing (in DockerConnection's case) and then no errors are raised.

IMHO it maybe some exception (e.g. "Not Implemented") raised either in `connection.py` or DockerConnection, which would indicate that upload operation did not actually succeed. At this time, it is less important to add actual implementation of copying files in DockerConnection, but some helpful comment for future people could be nice at the exception raise place.

nat-lab/tests/test_openwrt.py
```diff
@@ -0,0 +1,76 @@
...
+            config=Config(openwrt_config, paths=config_path),
+        )
+        await teliod.request_credentials_from_core()
+
+        async with teliod.start():
+            log.debug("Teliod started, waiting for connected vpn state...")
+            # Add temporary time sleep due to race condition
+            # ConnectToExitNode interrupts getStatus
+            # remove when properly fixed
+            await asyncio.sleep(2)
```
asledzinskiy: @Jauler @olekoliinyk @tomasz-grz I've added temporary time sleep here with comment until there is a stable fix for that. Would you mind such solution ?

nat-lab/data/nordvpnlite/config_openwrt_uci_pl_setup.json
tomasz-grz: Actually, I was thinking before the configs could be dynamically created `tempfile`s for each test scenario.
Because this `nat-lab/data/teliod/` directory could get out of hands if we add more tests.

But that should be part of a different ticket

nat-lab/bin/openwrt/container_net_setup.sh
```diff
@@ -7,6 +7,7 @@ ADD_ERR="Please add the following setting to your container:"
 tuntap="TUN device is missing. $ADD_ERR --device /dev/net/tun"
 WAN_IP="10.0.254.14/16"
 LAN_IP="192.168.115.254/24"
+NEW_BRIDGE_IP="192.168.115.253/24"
```
olekoliinyk: NIT: I think some better name can be defined for this? :)

nat-lab/tests/test_openwrt.py
tomasz-grz: Please add ticket number LLT-6693

tomasz-grz: You could possibly add the sleep inside of
`teliod.wait_for_vpn_connected_state()` so it could benefit other tests too.

I had in mind something like 100ms, but not verified if that is sufficient.

Edit: I added this sleep in my PR to `wait_for_vpn_connected_state()`
https://github.com/NordSecurity/libtelio/pull/1478

nat-lab/tests/nordvpnlite.py
```diff
@@ -84,23 +86,35 @@ def __repr__(self) -> str:
...
+        cmd = [str(config.paths.exec_path), "start"]
         if config.no_detach:
             cmd.append("--no-detach")
         cmd.append(str(config.path()))
         return cls(cmd)
 
     @classmethod
-    def is_alive(cls) -> "Command":
-        return cls([str(Paths.exec_path), "is-alive"])
+    def is_alive(cls, config: Optional["Config"] = None) -> "Command":
```
tomasz-grz: I'm not very familiar with python types system, but what type is `Optional["Config"]` and `"Command"`, how is it different to `Optional[Config]` and `Command`?

nat-lab/tests/teliod.py
```diff
@@ -84,23 +86,35 @@ def __repr__(self) -> str:
...
         return cls(cmd)
 
     @classmethod
-    def is_alive(cls) -> "Command":
-        return cls([str(Paths.exec_path), "is-alive"])
+    def is_alive(cls, config: Optional["Config"] = None) -> "Command":
+        if config:
+            exec_path = str(config.paths.exec_path)
+        else:
+            exec_path = str(Paths.exec_path)
```
tomasz-grz: What do you think exporting this in a helper function like `get_exec_path()` to avoid code duplication?

nat-lab/tests/nordvpnlite.py
```diff
@@ -84,23 +86,35 @@ def __repr__(self) -> str:
...
+        if config:
+            exec_path = str(config.paths.exec_path)
+        else:
+            exec_path = str(Paths.exec_path)
+        return cls([exec_path, "is-alive"])
 
     @classmethod
-    def get_status(cls) -> "Command":
-        return cls([str(Paths.exec_path), "get-status"])
+    def get_status(cls, config: Optional["Config"] = None) -> "Command":
```
tomasz-grz: Can we store `config` in self from `start()`? so that we don't have to pass it for every command?

nat-lab/tests/test_openwrt.py
Jauler: it would be significantly better if we could fix the issue before merging. 

As such sleeps have a tendency to stay for long time. Of course, jira ticket helps a bit.

asledzinskiy: Removing it from this PR as it was addressed in 1478

nat-lab/bin/openwrt/container_net_setup.sh
asledzinskiy: Will rename to DOCKER_BRIDGE_IP

nat-lab/tests/nordvpnlite.py
asledzinskiy: Putting type into quotes allows to reference this Type even if it is not defined at the point the function is declared

nat-lab/tests/teliod.py
asledzinskiy: Yep, agree

nat-lab/tests/nordvpnlite.py
asledzinskiy: This will imply that `start()` should always be executed first otherwise undefined attribute will be referenced.
I understand that `start()` should always be executed first but sounds a bit risky

nat-lab/tests/nordvpnlite.py
jjanowsk: Please move Config above Command, this way there is no reason for using forward reference. And in case of Command returning Command you can import Self from typing module and return it from those methods

nat-lab/tests/nordvpnlite.py
tomasz-grz: Hmm I think we do call `is_alive()` before `start()` sometimes.

It is a little bit unconventional that this is a `class Command` and not a `class NordVpnLite`

Otherwise I would suggest a default constructor with the config set to `None`, and a `set_config()` method..

tomasz-grz: Edit: oh I see there is `class NordVpnLite` with constructor that takes `config: Config = Config()`. 

My only question then is, why is it an `Optional["Config"]` here but for `start(cls, config: "Config")` it's not optional?

nat-lab/tests/nordvpnlite.py
```diff
@@ -284,9 +298,15 @@ async def quit(self) -> None:
 
     async def kill(self) -> None:
         try:
-            await self.connection.create_process(
-                ["killall", "-w", "-s", "SIGTERM", "nordvpnlite"]
-            ).execute()
+            # OpenWrt doesn't support killall -w
+            if "nordvpn" in self.config.paths.exec_path.parts:
+                await self.connection.create_process(
+                    ["killall", "-s", "SIGTERM", "nordvpn"]
```
tomasz-grz, line 275: `nordvpn` is the main Linux app, what it has to do with OpenWRT?

nat-lab/tests/test_openwrt.py
```diff
@@ -0,0 +1,71 @@
...
+    async with AsyncExitStack() as exit_stack:
+        client_connection = (
+            await setup_connections(exit_stack, [ConnectionTag.DOCKER_OPENWRT_CLIENT_1])
+        )[0].connection
+        gateway_connection = (
+            await setup_connections(exit_stack, [ConnectionTag.VM_OPENWRT_GW_1])
+        )[0].connection
+        assert gateway_connection is not None
+        await gateway_connection.create_process(
+            ["mkdir", "-p", "/etc/teliod"]
```
tomasz-grz: `teliod` was renamed to `nordvpnlite` please update all references

asledzinskiy: Thank, done

nat-lab/tests/nordvpnlite.py
asledzinskiy: @LukasPukenis  called the binary 'nordvpn' before renaming of 'teliod' :)
Maybe we can initiate discussion to use nordvpnlite for openwrt as well

nat-lab/tests/nordvpnlite.py
asledzinskiy: I agree that it is inconsistency. For `start` it was config: "Config" initially and I was hesitant to change it as it wasn't related to my test changes.
But I can make it Optional for all for consistency

jjanowsk: TBH this Command abstraction seems out of place here. Please just remove it. There is no reason for its existence. Functions in Nordvpnlite structure should just call execute_command with command string e.g. in start there should be: self.execute_command("start") and execute_command should just take exec path from its config.  Also Nordvpnlite structure should only use paths from its config not from global Paths class since config.paths are anyway initialized to default ones

nat-lab/tests/nordvpnlite.py
jjanowsk: But it is the main app on OpenWRT :) that's why we call it nordvpn on this platform even though it's nordvpnlite under the hood.

nat-lab/tests/test_openwrt.py
```diff
@@ -0,0 +1,71 @@
...
+        5. Ping PHOTO_ALBUM_IP from both openwrt and client node
+    """
+    async with AsyncExitStack() as exit_stack:
+        client_connection = (
+            await setup_connections(exit_stack, [ConnectionTag.DOCKER_OPENWRT_CLIENT_1])
+        )[0].connection
+        gateway_connection = (
+            await setup_connections(exit_stack, [ConnectionTag.VM_OPENWRT_GW_1])
+        )[0].connection
+        assert gateway_connection is not None
```
jjanowsk: Nit: I don't think these types of asserts are useful in Python. setup_connections will most likely throw if it cannot create the connection but even if not you will just get exception from the next line (`gateway_connection.create_process`) with a nice stacktrace so it's just noise IMO

nat-lab/tests/nordvpnlite.py
jjanowsk: Ignore moving part, but you can import Self from typing and fix it in all places that return itself in quotes. Thanks

nat-lab/tests/nordvpnlite.py
LukasPukenis: "nordvpn" is our executable for OpenWRT, it is a rename of the binary that comes in the Makefile at build time (previously "teliod" now "nordvpnlite").

The name is as such because it was agreed earlier with a PO to use "nordvpn" for the package name and as such I renamed the binary as well so it would be uniform.

At this point, where we renamed teliod -> nordvpnlite it might look a bit weird since nordvpnlite is already pretty marketable name(imho) :)

nat-lab/tests/test_openwrt.py
asledzinskiy: I'll double check but I think I've added it because of 'mypy' checker

nat-lab/tests/nordvpnlite.py
asledzinskiy: typing.Self is supported from python 3.11.
I see we have requires-python = ">=3.10" in toml file.
Theoretically on some runners it can be python 3.10

nat-lab/tests/nordvpnlite.py
asledzinskiy: I can't pass only command name like "start" to self.execute_command because in case of `start` i.e. we also need 'config.path' and -no-detach options that we don't need for 'is-alive', 'get-status', 'quit-daemon'.

So if I remove `class Command` completely - I'll still need to pass 'cmd' to self.execute_command, it's just will be part of `class NordVpnLite` methods. 

Also `class Command` has custom __str__ for logging but that can be fixed in log.debug of self.execute_command

nat-lab/tests/nordvpnlite.py
asledzinskiy: `class Command` has been removed, so that section is no longer applicable

nat-lab/tests/nordvpnlite.py
asledzinskiy: Done

nat-lab/tests/nordvpnlite.py
```diff
@@ -266,11 +250,13 @@ async def is_alive(self) -> bool:
...
     async def get_status(self) -> str:
-        status, _ = await self.execute_command(Command.get_status())
+        cmd = [str(self.config.paths.exec_path), "get-status"]
+        status, _ = await self.execute_command(cmd)
         return status
 
     async def quit(self) -> None:
-        stdout, stderr = await self.execute_command(Command.quit_daemon())
+        cmd = [str(self.config.paths.exec_path), "quit-daemon"]
+        stdout, stderr = await self.execute_command(cmd)
```
jjanowsk: Nit: I think it would be nicer to have it like this here and have str(self.config.paths.exec_path) hardcoded in execute_command to not repeat it for every command.
```
async def quit(self) -> None:
    stdout, stderr = await self.execute_command(["quit-daemon"])

...

async def execute_command(
    self,
    args: list,
) -> tuple[str, str]:
        try:
            proc = await self.connection.create_process([str(self.config.paths.exec_path)] + args).execute()
```

jjanowsk (dismissed):
+1

nat-lab/bin/openwrt/container_net_setup.sh
```diff
@@ -7,6 +7,7 @@ ADD_ERR="Please add the following setting to your container:"
 tuntap="TUN device is missing. $ADD_ERR --device /dev/net/tun"
 WAN_IP="10.0.254.14/16"
 LAN_IP="192.168.115.254/24"
+DOCKER_BRIDGE_IP="192.168.115.253/24"
```
tomasz-grz, line 10: Is this the docker management interface? Looks like it's on the same sub-net as `LAN` network, is that ok?

nat-lab/bin/openwrt/container_net_setup.sh
```diff
@@ -7,6 +7,7 @@ ADD_ERR="Please add the following setting to your container:"
 tuntap="TUN device is missing. $ADD_ERR --device /dev/net/tun"
 WAN_IP="10.0.254.14/16"
 LAN_IP="192.168.115.254/24"
```
tomasz-grz, line 9: Looks like we are not following the LAN to WAN addressing convention observed in the rest of the setup. Should the WAN IP be `10.0.254.15` for the LAN network `192.168.115.0/24`?

nat-lab/tests/nordvpnlite.py
asledzinskiy: Done

nat-lab/bin/openwrt/container_net_setup.sh
asledzinskiy: This IP is assigned for health check purposes - to check ssh connection to VM from the container. So we can assign  any non-taken IP from LAN

nat-lab/bin/openwrt/container_net_setup.sh
asledzinskiy: 10.0.254.14 was next available IP assigned to Gateways

Coming after this one:
  cone-gw-05:
    hostname: cone-gw-05
    <<: *common-gw
    networks:
      internet:
        ipv4_address: 10.0.254.13

nat-lab/tests/conftest.py
```diff
@@ -248,6 +250,22 @@ async def perform_setup_checks() -> bool:
     return True
 
 
+async def check_gateway_connectivity() -> bool:
```
asledzinskiy, line 253: @olekoliinyk @jjanowsk @Jauler @tomasz-grz I've added check on gateways connection before starting tcpdump on it

olekoliinyk (approved):

PR #1458, opened 2025-09-19
Improve socat connection to OpenWrt QEMU VM

olekoliinyk:
### Problem

### Solution


matislovas (approved):
+1

PR #1459, opened 2025-09-19
Add backoff to the ENS task

tomaszklak:
Since part of the ENS task logic is to retry connection, we need (exponential) backoff to avoid (d)dosing the VPN server.

- [x] README.md is updated

### DIFF
--- a/crates/telio-proto/src/ens.rs
+++ b/crates/telio-proto/src/ens.rs
@@ -22,7 +22,10 @@ use telio_task::io::{
     chan::{Rx, Tx},
     Chan,
 };
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_info, telio_log_warn};
+use telio_utils::{
+    exponential_backoff::{self, Backoff},
+    telio_log_debug, telio_log_error, telio_log_info, telio_log_warn,
+};
 use tokio::{select, sync::watch, task::JoinHandle};
 use tokio_rustls::TlsConnector;
 use tonic::{
@@ -71,6 +74,9 @@ pub enum Error {
     /// GRPC status error
     #[error("GRPC status error: {0}")]
     StatusError(#[from] tonic::Status),
+    /// Exponential backoff creation error
+    #[error("Exponential backoff creation failed {0}")]
+    ExponentialBackoff(#[from] exponential_backoff::Error),
 }
 
 /// ErrorNotificationService manages tasks started and stopped to consume the ENS grpc error streams
@@ -112,8 +118,9 @@ impl ErrorNotificationService {
         vpn_ip: IpAddr,
         vpn_public_key: PublicKey,
         local_private_key: SecretKey,
+        backoff: impl Backoff,
     ) -> Result<(), Error> {
-        self.start_monitor_on_port(vpn_ip, ENS_PORT, vpn_public_key, local_private_key)
+        self.start_monitor_on_port(vpn_ip, ENS_PORT, vpn_public_key, local_private_key, backoff)
             .await
     }
 
@@ -123,6 +130,7 @@ impl ErrorNotificationService {
         ens_port: u16,
         vpn_public_key: PublicKey,
         local_private_key: SecretKey,
+        backoff: impl Backoff,
     ) -> Result<(), Error> {
         telio_log_info!("Will start ENS monitoring on {vpn_ip}:{ens_port} ({vpn_public_key:?})");
         self.stop().await;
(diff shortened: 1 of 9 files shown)


src/device.rs
```diff
@@ -2084,9 +2086,14 @@ impl Runtime {
...
             if let Some(ens) = self.entities.error_notification_service.as_mut() {
                 telio_log_info!("Starting ENS monitoring");
-                ens.start_monitor(vpn_ip, exit_node.public_key, client_pk)
-                    .await
-                    .map_err(Box::new)?;
+                ens.start_monitor(
+                    vpn_ip,
+                    exit_node.public_key,
+                    client_pk,
+                    ExponentialBackoff::new(Default::default())?,
```
matislovas: Where are those `initial_s` and `maximal_s` values are actually passed to this object?

tomaszklak: No idea... I thought I've added it but clearly I have not. It should be fixed now.

crates/telio-proto/src/ens.rs
```diff
@@ -227,6 +271,7 @@ async fn task(
                     telio_log_warn!("Received error notification for '{vpn_uri}': {error_notification:?}");
                     match error_notification {
                         Ok(Some(error_notification)) => {
+                            backoff.reset();
```
matislovas, line 274: Will this be triggered on first genuine error? Or the server always returns some message when connected. My first thought was moving this `backoff.reset();` right after the `let mut stream = connection.into_inner();`.

matislovas:
+1

crates/telio-proto/src/ens.rs
tomaszklak: > Will this be triggered on first genuine error?

Yes

> My first thought was moving this `backoff.reset();` right after the `let mut stream = connection.into_inner();`.

So that we reset backoff after successfully handshaking? I feel like it's better to reset it here, since we know at that point that everything (ENS related) is working fine. At the `into_inner` part we only learn that it was possible to establish grpc connection.

src/device.rs
```diff
@@ -192,6 +192,8 @@ pub enum Error {
     PollingPeriodZero,
     #[error("Ens failure: {0:?}")]
     EnsFailure(#[from] Box<EnsError>),
+    #[error("Exponential backoff error")]
```
Hasan6979: Printing just "Exponential backoff error" hides the actual error that would have come. Could we also include the error generated in exponential_backoff::Error ?

src/libtelio.udl
```diff
@@ -803,6 +803,23 @@ dictionary FeatureErrorNotificationService {
...
+    /// Initial bound
+    ///
+    /// Used as the first backoff value after ExponentialBackoff creation or reset
+    u32 initial_s;
+
+    /// Maximal bound
+    ///
+    /// A maximal backoff value which might be achieved during exponential backoff
+    /// - if not defined there will be no upper bound for the penalty duration
+    u32? maximal_s;
```
Hasan6979, line 822: Could you also add in the comments the default value of it if not defined.

crates/telio-proto/src/ens.rs
```diff
@@ -148,6 +156,7 @@ impl ErrorNotificationService {
                 tx,
```
Hasan6979, line 156: I think on line 152 we can avoid the vpn_uri.clone() and pass the ownership of vpn_uri and modify the logging afterwards to not include vpn_uri. wdyt ?

tomaszklak: There was no need for that clone. Fixed.

src/libtelio.udl
tomaszklak: Done

src/device.rs
tomaszklak: Done

src/libtelio.udl
jjanowsk: `if not defined there will be no upper bound for the penalty duration [default 120s]` - TBH it's confusing. It seems like a misuse of serde + SmarDefault in that case. When you create this in Rust you get 120s default. Also when the app creates it via get_default_feature_config they'll get 120s. When creating from json it defaults to None. When creating manually via native constructor in the app's language there is no default for optional field unless you specify `= null` in the udl but it can only be null. Also for initial_s I feel like we should have `= 2` specified in the udl (but on the other hand we cannot specify `maximal_s = 120` in the udl which sucks) I think it reqires fixing, but not necessarily in this PR, but I'm mentioning it here since it became apparent now.

jjanowsk (dismissed):
+1

PR #1460, opened 2025-09-19
Randomize server selection in teliod

jjanowsk:
The list of recommended servers that we receive from the core api is sorted by the current load so that's why it makes sense to take the servers from the beginning of that list.  However, the list is not refreshed very often so it can lead to overloading the first server from the list, especially in busy locations. To avoid it, it is recommended to take 20 (default value) server recommendations and pick one at random.

- [x] README.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2852,7 +2852,7 @@ dependencies = [
  "interprocess",
  "ipnet",
  "nix 0.30.1",
- "rand 0.9.2",
+ "rand 0.8.5",
  "regex",
  "reqwest",
  "serde",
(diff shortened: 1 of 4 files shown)


clis/teliod/Cargo.toml
```diff
@@ -34,9 +34,9 @@ smart-default = "0.7.1"
 base64 = "0.22.1"
 daemonize = "0.5.0"
 cfg-if = "1.0.1"
+rand = "0.9.1"
```
mathiaspeters: What do you think about using the workspace version instead? If it means having to fix a bunch of stuff we can keep it like this though

LukasPukenis (commented):
+1

clis/teliod/Cargo.toml
jjanowsk: Done

mathiaspeters (dismissed):

PR #1461, opened 2025-09-22
LLT-6571: Handle legacy boringtun_reset_conns feature config option

mathiaspeters:
### Problem
We have "aliased" names for feature config option to handle legacy boringtun name and newer neptun name. Previously these were not handled properly so that using neptun_reset_conns actually did something and boringtun_reset_conns did nothing

### Solution
Introduce function that makes sure both options are taken into consideration and use that instead of directly referencing `neptun_reset_conns`, where relevant

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1044,7 +1044,7 @@ impl Runtime {
                 fw.process_outbound_packet(peer, packet, sink)
             }
         };
-        let firewall_reset_connections = if features.firewall.neptun_reset_conns {
+        let firewall_reset_connections = if features.firewall.neptun_reset_conns() {
             let fw = firewall.clone();
             let cb = move |exit_pubkey: &PublicKey, sink: &mut dyn io::Write| {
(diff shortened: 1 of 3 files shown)


src/device.rs
```diff
@@ -1044,7 +1044,7 @@ impl Runtime {
                 fw.process_outbound_packet(peer, packet, sink)
             }
         };
-        let firewall_reset_connections = if features.firewall.neptun_reset_conns {
+        let firewall_reset_connections = if features.firewall.neptun_reset_conns() {
```
stalowyjez, line 1047: Nit:  IMO that's a bit strange to have a method and a field with the same name

stalowyjez (approved):
Looks good, +1

src/device.rs
mathiaspeters: I think it's a pretty common rust pattern, but usually for private fields

matislovas:
+1

PR #1462, opened 2025-09-22
LLT-6649: Rework sleep for pq nonet handshake test

mathiaspeters:
### Problem
There shouldn't be a "PQ restarted" log line for 180 seconds after successful handshake. A test checked this by sleeping for 179 seconds, expecting no such log line and then waiting for log line after that. That worked well mostly, but there are steps between connection and that sleep that might take too long and make the time since last handshake actually be more than 180s, causing the test to fail.

### Solution
Get the timestamp of the connection being established and then sleep 179 seconds from that point

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_pq.py
+++ b/nat-lab/tests/test_pq.py
@@ -2,6 +2,7 @@ import asyncio
 import config
 import pytest
 from contextlib import AsyncExitStack
+from datetime import datetime, timedelta
 from helpers import SetupParameters, setup_environment
 from telio import Client
 from utils import stun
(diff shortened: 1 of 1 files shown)


nat-lab/tests/test_pq.py
```diff
@@ -534,15 +540,17 @@ async def test_pq_vpn_handshake_after_nonet(
         ip = await stun.get(client_conn, config.STUN_SERVER)
         assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
 
-        await _connect_vpn_pq(
+        pq_connected = await _connect_vpn_pq(
             client_conn,
             client_alpha,
         )
+        pq_restart = pq_connected + timedelta(seconds=179)
```
stalowyjez: Nit: Isn't it more like `just_before_pq_restart`?

mathiaspeters: I was trying to keep the name short but you are right, it's better to be accurate there. Fixed

PR #1467, opened 2025-09-22
Teliod servers not found log spam

tomasz-grz:
### Problem
We are spamming `Servers not found for '{target_country}' country code.` in the logs for every country we compare, since we are doing it inside the `find_map`.

### Solution
Log only once if no country is found.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/core_api.rs
+++ b/clis/nordvpnlite/src/core_api.rs
@@ -458,9 +458,7 @@ pub async fn get_server_endpoints_list(config: &NordVpnLiteConfig) -> Result<Vec
         // Find VPN exit node based on country
         VpnConfig::Country(target_country) => {
             if !target_country.chars().all(|c| c.is_ascii_alphabetic()) {
-                warn!(
-                    "Invalid ISO country code format: '{target_country}', expected 2-letter code"
-                );
+                warn!("Invalid country format: '{target_country}', non-ascii characters used");
                 None
             } else {
                 match get_countries_with_exp_backoff(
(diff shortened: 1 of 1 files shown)


clis/teliod/src/core_api.rs
```diff
@@ -484,6 +481,10 @@ pub async fn get_server_endpoints_list(
                         None
                     }
                 }
+                .or_else(|| {
+                    warn!("Servers not found for '{target_country}' country code.");
+                    None
+                })
```
lcruz99: Shouldn't this be called against the result of `countries_list.iter().find_map()`?

Otherwise when the `get_countries_with_exp_backoff()` call fails it'll be logged:
```
error!("Getting countries failed due to: {e}");
```

clis/teliod/src/core_api.rs
```diff
@@ -460,9 +460,7 @@ pub async fn get_server_endpoints_list(
         // Find VPN exit node based on country
         VpnConfig::Country(target_country) => {
             if !target_country.chars().all(|c| c.is_ascii_alphabetic()) {
-                warn!(
-                    "Invalid ISO country code format: '{target_country}', expected 2-letter code"
-                );
+                warn!("Invalid country format: '{target_country}'");
```
lcruz99: Add the info that non-ascii characters were used or are not supported to the log message.

clis/teliod/src/core_api.rs
tomasz-grz: Good point

lcruz99 (approved):

PR #1470, opened 2025-09-23
[LLT-6654] Add fetch artifacts implementation to build_libtelio.py

olekoliinyk:
### Problem
Currently there is no option/implementation in `build_libtelio.py` to fetch the files from artifacts stage (relevant case - for OpenWrt).

### Solution
- Extend rust_utils parser
- Add implementation to `build_libtelio.py` which allow to download the artifacts based on job name

- [x] README.md is updated


.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v3.0.0
+TRIGGERED_REF=olekoliinyk/use_ipk_from_build
```
tomasz-grz: needs to be updated?

olekoliinyk: Yeah, will update after merging :)

tomaszklak (approved):

PR #1473, opened 2025-09-24
Log remote server TLS certificate fingerprint for gRPC connections

tomaszklak:
We need to log the fingerprint, so that if it ever happens that if we ever get logs from the possible MitM attack, we will be able to check if the used certificate was our or not.


Jauler (dismissed):
+1.0

nat-lab/bin/ens
```diff
@@ -201,12 +210,18 @@ async def start_grpc_server():
     return server
 
 async def main():
+    cert_pem, key_pem = generate_self_signed_cert(
+        common_name="localhost",
+        dns_names=["localhost"],
+        ip_addrs=["127.0.0.1"],
+    )
+
```
stalowyjez: Hmm, at first sight that was not obvious for me why it was moved - maybe you could just return `grpc_tls_fingerprint` from `generate_self_signed_cert` and then pass it as an argument to `start_http_server`?

stalowyjez (commented):
I think this one place is a bit unclear for the reader, besides that, +1

nat-lab/bin/ens
tomaszklak: Done

PR #1476, opened 2025-09-25
Prepare v6.1.1 release

mathiaspeters:
### Problem
We need to backport some windows fixes to v6.1 line, to make sure it has relevant fixes in case windows decides to update

### Solution
Backport changes and add necessary release stuff, also removed old changelog files from old version in the series

- [x] README.md is updated


.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v2.15.0
+TRIGGERED_REF=v2.15.2
```
mathiaspeters: Removes qnap builds from CI pipeline (already done on main since before) and adds the required dependencies to the `generate-release` job (will be forward ported to main after this release is done)

stalowyjez (commented):
Looks good, +!

.env
Jauler: I think this `ref` does not have correct linux-image->uniffi-generators->uniffi-bindgen-cs version

mathiaspeters: Should be fine now

PR #1477, opened 2025-09-25
LLT-6308: Add chain C-compatible layer to telio firewall

stalowyjez:
### Problem
We want to use libfirewall as a core of telio-firewall ultimately, so we need to convert telio-firewall chains to a form digestable by the libfirewall C API. We need to allocate memory for this chains, while making the chain itself readable, so there are a few things to be done:
 - create proper `LibfwRule` and `LibfwFilter` and `LibfwAssociatedData` pointers containing the needed data
 - keep `recreate_chain` as clean as possible
 - cleanup the allocated data properly

### Solution
This PR adds similar rust structures the the ones which were used in telio-firewall chain, so the chain definition remains almost the same, and the vector of rules created by that is then converted into `FFiChainGuard` which contains a valid `LibfwChain` instance with all pointers pointing to a memory owned by `FfiChainGuard` - in this way we assure that the memory will be free and the same time as ffi chain struct, without any need of unsafe operations.

- [x] README.md is updated

### DIFF
--- a/crates/telio-firewall/src/chain.rs
+++ b/crates/telio-firewall/src/chain.rs
@@ -110,7 +110,7 @@ impl TryFrom<&LibfwNetworkFilter> for NetworkFilter {
                     }
                     unexpected_ip_version => {
                         telio_log_warn!(
-                            "Malformed chain: Unexpeted IP version: {}",
+                            "Malformed chain: Unexpected IP version: {}",
                             unexpected_ip_version
                         );
                         return Err(Error::InvalidChain);
@@ -1282,7 +1282,7 @@ pub mod tests {
                 filter: LibfwFilterData {
                     associated_data_filter: ManuallyDrop::new(LibfwAssociatedData {
                         associated_data_len: assoc_data.len(),
-                        associated_data: assoc_data.leak().as_ptr(),
+                        associated_data: assoc_data.as_ptr(),
                     }),
                 },
             },
@@ -1298,24 +1298,24 @@ pub mod tests {
         let rules = vec![
             LibfwRule {
                 filter_count: filters1.len(),
-                filters: filters1.leak().as_ptr(),
+                filters: filters1.as_ptr(),
                 action: LIBFW_VERDICT_ACCEPT,
             },
             LibfwRule {
                 filter_count: filters2.len(),
-                filters: filters2.leak().as_ptr(),
+                filters: filters2.as_ptr(),
                 action: LIBFW_VERDICT_DROP,
             },
             LibfwRule {
                 filter_count: filters3.len(),
-                filters: filters3.leak().as_ptr(),
+                filters: filters3.as_ptr(),
                 action: LIBFW_VERDICT_REJECT,
             },
         ];
 
         let test_ffi_chain = LibfwChain {
             rule_count: rules.len(),
-            rules: rules.leak().as_ptr(),
+            rules: rules.as_ptr(),
         };
 
         // Compare converted LibfwChain to original Chain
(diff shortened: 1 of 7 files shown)


crates/telio-firewall/src/chain_helpers.rs
```diff
@@ -0,0 +1,648 @@
...
+        }
+    }
+}
+
+type PinnedAssocData = Pin<Box<Vec<u8>>>;
+type PinnedFilters = Pin<Box<Vec<LibfwFilter>>>;
+type PinnedRules = Pin<Box<Vec<LibfwRule>>>;
+
+impl From<&Filter> for (LibfwFilter, Option<PinnedAssocData>) {
+    fn from(value: &Filter) -> (LibfwFilter, Option<PinnedAssocData>) {
```
mathiaspeters: Couldn't we use the signature `fn from(value: &Filter) -> Self` here to avoid duplicating the type?

crates/telio-firewall/src/chain_helpers.rs
```diff
@@ -0,0 +1,648 @@
...
+                            }),
+                        },
+                        LIBFW_FILTER_ASSOCIATED_DATA,
+                        Some(cloned_assoc_data),
+                    )
+                } else {
+                    (
+                        LibfwFilterData {
+                            associated_data_filter: ManuallyDrop::new(LibfwAssociatedData {
+                                associated_data: null_mut(),
```
mathiaspeters: Personal preference: For something like `null()` and `null_mut()` I prefer to fully qualify it as `std::ptr::null()` and `std::ptr::null_mut()` to make it clearer where it's from

mathiaspeters:
+1

crates/telio-firewall/src/chain_helpers.rs
```diff
@@ -0,0 +1,648 @@
...
+    LIBFW_IP_TYPE_V6, LIBFW_NEXT_PROTO_ICMP, LIBFW_NEXT_PROTO_ICMPV6, LIBFW_NEXT_PROTO_TCP,
+    LIBFW_NEXT_PROTO_UDP,
+};
+
+use crate::conntrack::AssociatedData;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct NetworkFilterData {
+    pub(crate) network: IpNet,
+    pub(crate) ports: (u16, u16),
```
LukasPukenis: nit: `ports` -> `port_range`

crates/telio-firewall/src/chain.rs
```diff
@@ -93,17 +84,94 @@ impl NetworkFilterData {
...
+                match value.network_addr.ip_type {
+                    LIBFW_IP_TYPE_V4 => {
+                        IpAddr::V4(unsafe { value.network_addr.ip_data.ipv4_bytes }.into())
+                    }
+                    LIBFW_IP_TYPE_V6 => {
+                        IpAddr::V6(unsafe { value.network_addr.ip_data.ipv6_bytes }.into())
+                    }
+                    unexpected_ip_version => {
+                        telio_log_warn!(
+                            "Malformed chain: Unexpeted IP version: {}",
```
LukasPukenis: typo: `Unexpeted` -> `Unexpected`

LukasPukenis (commented):
+1

crates/telio-firewall/src/chain_helpers.rs
stalowyjez: Changed

crates/telio-firewall/src/chain.rs
stalowyjez: Fixed

crates/telio-firewall/src/chain_helpers.rs
stalowyjez: Changed

crates/telio-firewall/src/chain_helpers.rs
stalowyjez: Qualified

mathiaspeters (approved):

PR #1478, opened 2025-09-25
nordvpnlite increase command timeouts

tomasz-grz:
### Problem
There are two race conditions, occasionally causing a Broken Pipe and Client Timeout errors when requesting `get-status` and `is-alive` commands while `nordvpnlite` daemon is transitioning between states.

Proper solutions should be investigated in LLT-6694 and LLT-6693

### Solution
Increase client command timeouts, and ignore broken pipe errors.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/nordvpnlite.py
+++ b/nat-lab/tests/nordvpnlite.py
@@ -257,7 +257,11 @@ class NordVpnLite:
                 raise IgnoreableError() from exc
             if "Error: DaemonIsNotRunning" in exc.stderr:
                 return False
+            # TODO: remove after LLT-6694
             if "Connection reset by peer" in exc.stderr:
+                raise IgnoreableError() from exc
+            # TODO: remove after LLT-6694
+            if "Broken pipe" in exc.stderr:
                 raise IgnoreableError() from exc
             raise exc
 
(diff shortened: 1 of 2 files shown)


nat-lab/tests/nordvpnlite.py
```diff
@@ -257,8 +257,12 @@ async def is_alive(self) -> bool:
                 raise IgnoreableError() from exc
             if "Error: DaemonIsNotRunning" in exc.stderr:
                 return False
+            # TODO: remove after LLT-6694
             if "Connection reset by peer" in exc.stderr:
                 raise IgnoreableError() from exc
+            # TODO: remove after LLT-6694
+            if "Broken pipe" in exc.stderr:
+                raise IgnoreableError() from exc
```
LukasPukenis, line 265: thoughts:
```
        await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
```
I just wonder if this line isn't enough to make the race condition go away or is it present in other places as well(that should be also patched with sleeps)?

By the way, very nice anchoring to concrete tickets for a more appropriate fix :)

tomasz-grz: It's two different issues (though maybe related), so that would only help with one of them 😞

LukasPukenis: Oh okay, wanted to be sure :)

PR #1482, opened 2025-09-30
Rename for OpenWRT

LukasPukenis:
OpenWRT used nordvpn as a name
but should use NordVPN Lite instead.

### Problem

### Solution


.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v3.2.0
+TRIGGERED_REF=lukasp/rename_openwrt_binary
```
LukasPukenis, line 1: Will fix this in the next PR(since I need to merge libtelio-build with appropriate changes first, then tag it, then point it here).

asledzinskiy:
+1

PR #1483, opened 2025-09-30
LLT-6308: Use libfirewall API functions

stalowyjez:
### Problem
Ultimately, we want to use libfirewall in telio-firewall, so we need to change StatefullFirewall implementation, so it'll start using libfirewall API functions instead of accessing its internals (currently cloned) directly

### Solution
In this PR I clone the rest of the libfirewall API and rewrite StaticFirewall logic, so it calls this API instead of internal libfirewall methods

- [x] README.md is updated

### DIFF
--- a/crates/telio-firewall/src/conntrack.rs
+++ b/crates/telio-firewall/src/conntrack.rs
@@ -1275,7 +1275,9 @@ mod tests {
     };
 
     use crate::{
-        conntrack::{ConnectionState, Conntrack},
+        conntrack::{
+            ConnectionState, Conntrack, LRU_CACHE_ICMP_TTL, LRU_CACHE_TCP_TTL, LRU_CACHE_UDP_TTL,
+        },
         packet::IpPacket,
     };
 
@@ -1481,6 +1483,42 @@ mod tests {
         make_icmp4_with_body(src, dst, icmp_type, &[])
     }
 
+    fn handle_inbound_ipv4_packet(
+        ctk: &Conntrack,
+        packet: &[u8],
+        assoc_data: Option<&[u8]>,
+    ) -> ConnectionState {
+        ctk.track_inbound_ip_packet::<Ipv4Packet>(assoc_data, packet)
+            .expect("Unexpected conntrack error")
+    }
+
+    fn handle_outbound_ipv4_packet(
+        ctk: &Conntrack,
+        packet: &[u8],
+        assoc_data: Option<&[u8]>,
+    ) -> ConnectionState {
+        ctk.track_outbound_ip_packet::<Ipv4Packet>(assoc_data, packet)
+            .expect("Unexpected conntrack error")
+    }
+
+    fn handle_inbound_ipv6_packet(
+        ctk: &Conntrack,
+        packet: &[u8],
+        assoc_data: Option<&[u8]>,
+    ) -> ConnectionState {
+        ctk.track_inbound_ip_packet::<Ipv6Packet>(assoc_data, packet)
+            .expect("Unexpected conntrack error")
+    }
+
+    fn handle_outbound_ipv6_packet(
+        ctk: &Conntrack,
+        packet: &[u8],
+        assoc_data: Option<&[u8]>,
+    ) -> ConnectionState {
+        ctk.track_outbound_ip_packet::<Ipv6Packet>(assoc_data, packet)
+            .expect("Unexpected conntrack error")
+    }
+
     #[test]
     fn conntrack_establishing_outbound_udp_connection() {
         let conntrack = Conntrack::new();
@@ -1496,18 +1534,21 @@ mod tests {
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_udp_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.udp.lock().len(), 1);
 
         // Another outbound packet
         let conn_state = conntrack
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_udp_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.udp.lock().len(), 1);
 
         // Inbound packet should change conn state to established
         let conn_state = conntrack
             .track_inbound_ip_packet::<Ipv4Packet>(None, &inbound_udp_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::Established);
+        assert_eq!(conntrack.udp.lock().len(), 1);
     }
 
     #[test]
@@ -1525,18 +1566,100 @@ mod tests {
             .track_inbound_ip_packet::<Ipv4Packet>(None, &inbound_udp_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.udp.lock().len(), 1);
 
         // Another outbound packet
         let conn_state = conntrack
             .track_inbound_ip_packet::<Ipv4Packet>(None, &inbound_udp_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.udp.lock().len(), 1);
 
         // Inbound packet should change conn state to established
         let conn_state = conntrack
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_udp_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::Established);
+        assert_eq!(conntrack.udp.lock().len(), 1);
+    }
+
+    #[rustfmt::skip]
+    #[test]
+    fn multiple_udp_connections() {
+        struct TestInput {
+            src1: &'static str,
+            src2: &'static str,
+            src3: &'static str,
+            src4: &'static str,
+            src5: &'static str,
+            dst1: &'static str,
+            dst2: &'static str,
+            make_udp: fn(&str, &str) -> Vec<u8>,
+            handle_inbound_packet: fn(&Conntrack, &[u8], Option<&[u8]>) -> ConnectionState,
+            handle_outbound_packet: fn(&Conntrack, &[u8], Option<&[u8]>) -> ConnectionState,
+        }
+
+        let test_inputs = vec![
+            TestInput{
+                src1: "127.0.0.1:1111", src2: "127.0.0.1:2222",
+                src3: "127.0.0.1:3333", src4: "127.0.0.1:4444",
+                src5: "127.0.0.1:5555",
+                dst1: "8.8.8.8:8888", dst2: "8.8.8.8:7777",
+                make_udp: make_udp,
+                handle_inbound_packet: handle_inbound_ipv4_packet,
+                handle_outbound_packet: handle_outbound_ipv4_packet,
+            },
+            TestInput{
+                src1: "[::1]:1111", src2: "[::1]:2222",
+                src3: "[::1]:3333", src4: "[::1]:4444",
+                src5: "[::1]:5555",
+                dst1: "[2001:4860:4860::8888]:8888",
+                dst2: "[2001:4860:4860::8888]:7777",
+                make_udp: make_udp6,
+                handle_inbound_packet: handle_inbound_ipv6_packet,
+                handle_outbound_packet: handle_outbound_ipv6_packet,
+            },
+        ];
+
+        for TestInput { src1, src2, src3, src4, src5, dst1, dst2, make_udp, handle_inbound_packet, handle_outbound_packet } in test_inputs {
+            let ctk = Conntrack::new_with_capacity_and_ttl(3, LRU_CACHE_TCP_TTL, LRU_CACHE_UDP_TTL, LRU_CACHE_ICMP_TTL);
+
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src1), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 1);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src2), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 2);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src3), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src4), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src1, dst1), None), ConnectionState::New);
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src2, dst1), None), ConnectionState::New);
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src1, dst1), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src3, dst1), None), ConnectionState::New);
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src1, dst1), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src4, dst1), None), ConnectionState::New);
+            assert_eq!(handle_outbound_packet(&ctk, &make_udp(src1, dst1), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src3), None), ConnectionState::Established);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src4), None), ConnectionState::Established);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src1), None), ConnectionState::Established);
+            assert_eq!(ctk.udp.lock().len(), 3);
+
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src2), None), ConnectionState::New);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src2), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst1, src5), None), ConnectionState::New);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst2, src1), None), ConnectionState::New);
+            assert_eq!(handle_inbound_packet(&ctk, &make_udp(dst2, src1), None), ConnectionState::New);
+            assert_eq!(ctk.udp.lock().len(), 3);
+        }
     }
 
     #[test]
@@ -1557,42 +1680,49 @@ mod tests {
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_syn_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
 
         // Second oubound SYN packet shouldn't change conn state to established
         let conn_state = conntrack
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_syn_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
 
         // Outbound SYN-ACK packet shouldn't change the state to established
         let conn_state = conntrack
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_syn_ack_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::New);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
 
         // Inbound packet should change conn state to established
         let conn_state = conntrack
             .track_inbound_ip_packet::<Ipv4Packet>(None, &inbound_syn_ack_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::Established);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
 
         // First FIN packet should still get state Established
         let conn_state = conntrack
             .track_inbound_ip_packet::<Ipv4Packet>(None, &inbound_fin_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::Established);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
 
         // Second FIN from the same direction should also get Established
         let conn_state = conntrack
             .track_inbound_ip_packet::<Ipv4Packet>(None, &inbound_fin_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::Established);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
 
         // Inbound FIN should change connection state to closed
         let conn_state = conntrack
             .track_outbound_ip_packet::<Ipv4Packet>(None, &outbound_fin_packet)
             .expect("Unexpected conntrack error");
         assert_eq!(conn_state, ConnectionState::Closed);
+        assert_eq!(conntrack.tcp.lock().len(), 1);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


crates/telio-firewall/src/error.rs
```diff
@@ -0,0 +1,81 @@
...
+    MalformedUdpPacket,
+    MalformedTcpPacket,
+    MalformedIcmpPacket,
+    UnexpectedProtocol,
+    InvalidIcmpErrorPayload,
+    UnexpectedPacketType,
+    NullPointer,
+    NotImplemented,
+    InvalidChain,
+}
```
LukasPukenis: nit: is it worth having such a one to one mapping?

crates/telio-firewall/src/log.rs
```diff
@@ -0,0 +1,109 @@
...
+///
+/// @level - log level as defined in @LIBFW_LOG_LEVEL
+/// @log_line - zero-terminated log line
+///
+pub type LibfwLogCallback = Option<extern "C" fn(level: LibfwLogLevel, log_line: *const c_char)>;
+
+pub(crate) static LOG_CALLBACK: RwLock<LibfwLogCallback> = RwLock::const_new(RawRwLock::INIT, None);
+pub(crate) static MIN_LOG_LEVEL: RwLock<LibfwLogLevel> =
+    RwLock::const_new(RawRwLock::INIT, LibfwLogLevel::LibfwLogLevelInfo);
+
```
LukasPukenis, line 42: I am a bit confused about the logging. Here I can see it clearly that the initial value is, for some reason `info`, then in `StatefulFirewall::new()` it is overwritten with `trace`:
```
    libfw_set_log_callback(LibfwLogLevel::LibfwLogLevelTrace, Some(log_callback));
```

Ideally, I would expect the same log level to be passed as in `set_global_logger()` but that would obviously require more changes since it's a static function. I would like to understand the current implementation better.

crates/telio-firewall/src/log.rs
```diff
@@ -0,0 +1,109 @@
+use parking_lot::{lock_api::RawRwLock, RwLock};
```
LukasPukenis, line 1: Question: I do not understand why do we have `libfw_log_*` macros, are they necessary? They appear to trigger the logger callback which converts Rust string to C string and then lands in firewall.rs::log_callback() where it is converted back to C like string and emitted via callback :thinking:.

If we want additional logs from libfirewall(in addition to already existing logs from within the library) we should use `libtelio_log_*!` macros.

crates/telio-firewall/src/libfirewall_api.rs
```diff
@@ -0,0 +1,478 @@
...
+///
+/// # Safety
+///
+/// This function dereferences pointer to firewall - user must ensure that this is
+/// the pointer returned by `libfw_init`.
+#[no_mangle]
+pub unsafe extern "C" fn libfw_deinit(firewall: *mut LibfwFirewall) {
+    if !firewall.is_null() {
+        drop(Box::from_raw(firewall));
+    }
```
LukasPukenis, line 477: Could we have a log in case it is `null`?(thus unexpected)

crates/telio-firewall/src/conntrack.rs
```diff
@@ -63,6 +47,11 @@ pub(crate) use unwrap_option_or_return;
 
 const KEY_SIZE: usize = 32;
 
+pub(crate) const LRU_CACHE_SIZE: usize = 8192;
+pub(crate) const LRU_CACHE_TCP_TTL: Duration = Duration::from_secs(7440); // 2 hours 4 minutes as per  REQ-5 of RFC5382 section-8
+pub(crate) const LRU_CACHE_UDP_TTL: Duration = Duration::from_secs(120); // 2 minutes as per REQ-5 of RFC4787 section-12
+pub(crate) const LRU_CACHE_ICMP_TTL: Duration = Duration::from_secs(60); // 1 minute as per REQ-2 of RFC5508 section-9
+
```
LukasPukenis, line 53: The references to RFC <3 <3

crates/telio-firewall/src/conntrack.rs
```diff
@@ -244,12 +255,13 @@ impl Conntrack {
                     last_out_pkg_chunk: Some(last_chunk),
                     state: ConnectionState::New,
                 };
-                telio_log_trace!("Inserting new UDP conntrack entry {:?}", e.key());
+                libfw_log_trace!("Inserting new UDP conntrack entry {:?}", e.key());
```
LukasPukenis, line 257: This is one of the logs I asked about previously - it should remain `telio_log_trace!`

crates/telio-firewall/src/log.rs
stalowyjez: This PR aims at using stuff through FFI API and I guess it makes sense to start using loggjng callback as it will be used after moving to libfirewall.so

crates/telio-firewall/src/conntrack.rs
stalowyjez: These is just copied from libfirewall ^^'

crates/telio-firewall/src/conntrack.rs
stalowyjez: As I explained previously, imho it shouldn't ;)

crates/telio-firewall/src/libfirewall_api.rs
stalowyjez: I'm not sure - C `free()` just doesn't take any action when the pointer is NULL and I think it's kinda intuitive that the user will follow such well-known patterns without checking for NULL. I'm not strongly agains that log, though. Or maybe we should set it to NULL after deallocation to msake that more indempotent? IDK

crates/telio-firewall/src/error.rs
stalowyjez: IMHO yes, because it's fairly easy to maintain and having C `LibfwError` easily convertible from and to `Result<Error>` (which can be used with `?` and other rust stuff) sounds like a pretty good thing to have imho

crates/telio-firewall/src/log.rs
stalowyjez: Soo, this part is not  really meant to be a part of `telio-firewall`, its just temporary copy to make it easier to integrate it with libfw interface before starting using `libfirewall.so`

>Ideally, I would expect the same log level to be passed as in set_global_logger()

TBH I don't get why ^^'

Simple explanation is that `Trace` level with the current logging callback just sets it to the same level as telio logging level, which sounds nice. And that's not implementable in `libfirewall` itself, as we don't know what will be in the callback, `Info` level sounds like a nice default to me.

crates/telio-firewall/src/log.rs
LukasPukenis: Ah. Okay if libfirewall will become a separate binary, we will want to use libtelio callback for it and also will relay those logs the application, it makes sense then.

crates/telio-firewall/src/libfirewall_api.rs
LukasPukenis: I have nothing against the guard but was curious :). In general if I am sure that some invariant is invalid I do not like to have code for it.

crates/telio-firewall/src/conntrack.rs
LukasPukenis: Let me ask the other way around - once we move to libfirewall.so are these macros going to go away?

stalowyjez: They will still live inside libfirewall repo (i.e. whole conntrack.rs, chain.rs etc. will be moved there), but this file will be removed from `telio-firewall` after we start using `libfirewall.so`

crates/telio-firewall/benches/firewall_bench.rs
```diff
@@ -114,7 +114,6 @@ pub fn firewall_tcp_inbound_benchmarks(c: &mut Criterion) {
                                     .process_inbound_packet(&peers[which_peer], &param.packet));
                                 which_peer = (which_peer + 1) % peers.len();
                             }
-                            assert_eq!((0, 0), firewall.get_state());
```
Hasan6979, line 117: Can we have some other way to assert this instead of removing it ?

crates/telio-firewall/src/log.rs
```diff
@@ -0,0 +1,109 @@
...
+macro_rules! libfw_log {
+    ($level:expr, $text:literal) => {
+        {
+            use crate::log::{LOG_CALLBACK, MIN_LOG_LEVEL, LibfwLogLevel};
+            if *MIN_LOG_LEVEL.read() <= $level {
+                if let Some(logger) = LOG_CALLBACK.read().as_ref() {
+                    use std::ffi::CString;
+                    let log_message = match CString::new($text) {
+                        Ok(message) => message,
+                        Err(err) => CString::new(format!("INVALID LOG MESSAGE: NULL byte on position: {}", err.nul_position())).expect("There can't be any NULL characters in this string")
```
Hasan6979: Why are we panicking here instead of handling it gracefully ?

crates/telio-firewall/src/log.rs
```diff
@@ -0,0 +1,109 @@
...
+
+macro_rules! libfw_log_warn {
+    ($($arg:expr),+) => {
+        {
+            use crate::log::libfw_log;
+            libfw_log!(LibfwLogLevel::LibfwLogLevelWarn, $($arg),+)
+        }
+    };
+}
+
```
Hasan6979, line 116: nit: merge all the `pub(crate) use XXX ` together and the macros together

crates/telio-firewall/src/firewall.rs
```diff
@@ -112,6 +112,7 @@ pub(crate) trait IpPacket<'a>: Sized + Debug + Packet {
     type Addr: Into<IpAddr> + Into<std::net::IpAddr>;
     type Icmp: Icmp;
 
+    #[allow(dead_code)]
```
Hasan6979: Why are we allowing dead_code here ?

crates/telio-firewall/src/firewall.rs
```diff
@@ -300,29 +299,32 @@ pub struct StatefullFirewall {
...
 }
 
+// Access to internal firewall structs is guarded by locks, so that should be fine
+unsafe impl Sync for StatefullFirewall {}
+unsafe impl Send for StatefullFirewall {}
+
 impl LocalInterfacesObserver for StatefullFirewall {
-    fn notify(&self) {
-        self.recreate_chain();
+    fn notify(&self) {}
```
Hasan6979: Why is this an empty function ?

crates/telio-firewall/src/firewall.rs
```diff
@@ -283,9 +284,7 @@ struct Whitelist {
 /// Statefull packet-filter firewall.
 pub struct StatefullFirewall {
     /// Connection tracking
-    conntrack: Conntrack,
-    /// Firewall rule chain
-    chain: RwLock<InternalChain>,
+    firewall: *mut LibfwFirewall,
```
Hasan6979, line 180: Could you modify the comment of firewall. It explains `Connection tracking`. Could you make it a bit more descriptive ?

crates/telio-firewall/src/firewall.rs
```diff
@@ -431,6 +435,15 @@ impl StatefullFirewall {
     fn recreate_chain(&self) {
         let mut rules = vec![];
 
+        // Drop all IPv6 packets when we don't allow ipv6 traffic
+        const ALL_IPS: IpNet = IpNet::V6(Ipv6Net::new_assert(StdIpv6Addr::UNSPECIFIED, 0));
```
Hasan6979: Since this is just for IPv6, better to specify IPv6

crates/telio-firewall/src/firewall.rs
```diff
@@ -607,129 +620,42 @@ impl StatefullFirewall {
...
+    let packet_data = unsafe { std::slice::from_raw_parts(buffer, buffer_len) };
 
-        Ok(self.chain.read().process_packet(
-            conn_state,
-            &ip,
-            Some(public_key),
-            CtkDirection::Inbound,
-        ) == LibfwVerdict::LibfwVerdictAccept)
-    }
+    let _ = sink.write(packet_data);
```
Hasan6979: Can you handle the error here.

crates/telio-firewall/src/firewall.rs
```diff
@@ -799,85 +725,53 @@ impl Firewall for StatefullFirewall {
...
-                    .track_outbound_ip_packet::<Ipv6Packet>(Some(public_key), buffer)
-                    .unwrap_or_else(|_| {
-                        telio_log_warn!("Conntrack failed for outbound IPv4 packet");
-                        // Some packets are unexpected, but we still let them through, like outbound ICMP reply
-                        CtkConnectionState::New
-                    });
-                self.process_outbound_ip_packet::<Ipv6Packet>(public_key, buffer, sink, conn_state)
-                    .unwrap_or(false)
+        let sink_ptr = &sink as *const &mut dyn io::Write;
+        LibfwVerdict::LibfwVerdictAccept
```
Hasan6979, line 626: Here we are just getting whether it was accepted or not. If it wasn't accepted it could be rejected or dropped, information that would be hidden

crates/telio-firewall/src/firewall.rs
```diff
@@ -799,85 +725,53 @@ impl Firewall for StatefullFirewall {
...
-                let result = self
-                    .process_inbound_ip_packet::<Ipv6Packet>(public_key, buffer, conn_state)
-                    .unwrap_or(false);
-                self.cleanup_conntrack_inbound::<Ipv6Packet>(public_key, buffer, result);
-                result
-            }
-            version => {
-                telio_log_warn!("Unexpected IP version {version} for inbound packet");
-                false
+        LibfwVerdict::LibfwVerdictAccept
```
Hasan6979, line 645: Same for this

crates/telio-firewall/src/firewall.rs
```diff
@@ -799,85 +725,53 @@ impl Firewall for StatefullFirewall {
...
     fn reset_connections(&self, pubkey: &PublicKey, sink: &mut dyn io::Write) -> io::Result<()> {
         telio_log_debug!("Constructing connetion reset packets");
-
-        self.conntrack
-            .reset_tcp_conns(Some(pubkey), |packet: &[u8]| sink.write_all(packet))?;
-        self.conntrack
-            .reset_udp_conns(Some(pubkey), |packet: &[u8]| sink.write_all(packet))?;
+        let sink_ptr = &sink as *const &mut dyn io::Write;
+        unsafe {
+            libfw_trigger_stale_connection_close(
```
Hasan6979, line 663: Handle returned error

crates/telio-firewall/src/firewall.rs
```diff
@@ -799,85 +725,53 @@ impl Firewall for StatefullFirewall {
...
+                self.firewall,
+                pubkey.as_ptr(),
+                pubkey.len(),
+                sink_ptr as *mut c_void,
+                Some(write_to_sink),
+                None,
+            );
+        }
 
         Ok(())
```
Hasan6979: Do we need to return a result ? Either we can return the Result from the above function or we can not return anything in case we are logging the error in this function. It would be better to return the result rather than hardcoding it as Ok(())

crates/telio-firewall/src/firewall.rs
```diff
@@ -1345,34 +1239,28 @@ pub mod tests {
...
         for TestInput { src1, src2, src3, src4, src5, dst1, dst2, make_udp } in test_inputs {
-            let fw = StatefullFirewall::new_custom(3, LRU_TIMEOUT, true, &FeatureFirewall::default());
+            let fw = StatefullFirewall::new(true, &FeatureFirewall::default());
             fw.set_ip_addresses(vec![(StdIpAddr::V4(StdIpv4Addr::new(127, 0, 0, 1))), StdIpAddr::V6(StdIpv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1))]);
             // Should FAIL (no matching outgoing connections yet)
             assert_eq!(fw.process_inbound_packet(&make_peer(), &make_udp(dst1, src1)), false);
             assert_eq!(fw.process_inbound_packet(&make_peer(), &make_udp(dst1, src2)), false);
             assert_eq!(fw.process_inbound_packet(&make_peer(), &make_udp(dst1, src3)), false);
             assert_eq!(fw.process_inbound_packet(&make_peer(), &make_udp(dst1, src4)), false);
-            assert_eq!(fw.conntrack.udp.lock().len(), 0);
```
Hasan6979, line 1355: I don't think removing them without an alternative is a good idea as these tests depend on reading the LRU cache in the firewall.

crates/telio-firewall/src/error.rs
```diff
@@ -0,0 +1,81 @@
+///
+/// Possible FW errors
+///
+#[allow(clippy::enum_variant_names)]
+#[repr(C)]
+#[derive(Copy, Clone, Debug)]
+pub enum LibfwError {
+    LibfwSuccess,
```
mathiaspeters, line 8: This is probably more of a libfirewall comment than a comment for this PR, but it seems to me now like the name of this enum should be `LibfwResult` or something since there is a variant that is very specifically not an error :sweat_smile:

crates/telio-firewall/src/firewall.rs
```diff
@@ -2454,11 +2210,11 @@ pub mod tests {
         for TestInput { us, them, make_icmp, is_v4 } in &test_inputs {
             let mut feature = FeatureFirewall::default();
             // Set number of conntrack entries to 0 to test only the whitelist
-            let fw = StatefullFirewall::new_custom(0, 20, true, &feature);
+            let fw = StatefullFirewall::new(true, &feature);
             fw.set_ip_addresses(vec![(StdIpAddr::V4(StdIpv4Addr::new(127, 0, 0, 1))), StdIpAddr::V6(StdIpv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1))]);
 
             let peer = make_random_peer();
-            assert_eq!(fw.process_outbound_packet(&peer.0, &make_icmp(us, them, IcmpTypes::EchoRequest.into())), true);
+            // assert_eq!(fw.process_outbound_packet(&peer.0, &make_icmp(us, them, IcmpTypes::EchoRequest.into())), true);
```
mathiaspeters: This should probably be either removed or uncommented

mathiaspeters:
Let's maybe wait to merge this until v6.2.0 has been released, since we're planning to release that soon

crates/telio-firewall/benches/firewall_bench.rs
stalowyjez: That would clutter API with some functions for getting libfirewall internal states, so I think that's a bad idea

crates/telio-firewall/src/log.rs
stalowyjez: Because this is already handling gracefully some other error and here we just wrote this string, so we are sure that it won't panic

crates/telio-firewall/src/log.rs
stalowyjez: IDK, would that be clearer? Anyway, this file is just copied from libfirewall and will be removed in the next iteration, so I guess reformatting it here doesn't make a lot of sense.

crates/telio-firewall/src/error.rs
stalowyjez: Yeah, that would be probably better name ^^'
I'll add that to fixes to libfirewall added after this PR

crates/telio-firewall/src/conntrack.rs
LukasPukenis: But here it's libtelio repo so we only need libtelio related macros, especially after moving to `libfirewall.so` unless I miss something.

crates/telio-firewall/src/log.rs
LukasPukenis: My main concern is that I expect 1:1 mapping from libfirewall to libtelio logs, meaning `trace` level log message from libfirewall will be handled as `trace` level log message from within libtelio(such as `telio_log_trace!()`). Same for other levels. If that's achieved then it is ok but I cannot say that I get this impression by looking at the code.

LukasPukenis (commented):
+1

crates/telio-firewall/src/firewall.rs
stalowyjez: It is not used anymore, but this will be removed from `telio-firewall`, as it is used at this stage only in libfirewall internals, so - maybe it would be clearer if I would just remove that now and use `ip_packet.rs`, like in libfirewall?

crates/telio-firewall/src/firewall.rs
stalowyjez: Ooops

crates/telio-firewall/src/firewall.rs
stalowyjez: Hmm, what do you mean?

crates/telio-firewall/src/firewall.rs
stalowyjez: What do you mean by "handling"? Is logging it good enough?

crates/telio-firewall/src/firewall.rs
stalowyjez: Yeah, currently we just don't care here whether it is rejected or dropped here

crates/telio-firewall/src/firewall.rs
stalowyjez: Same

crates/telio-firewall/src/firewall.rs
stalowyjez: But internal LRU behavior should be tested inside libfirewall, not here, at this point telio shouldn't care whether libfirewall even uses LRU cache or not

crates/telio-firewall/src/firewall.rs
stalowyjez: Removed

crates/telio-firewall/src/firewall.rs
stalowyjez: Fixed

crates/telio-firewall/src/firewall.rs
stalowyjez: Hmm, `device.rs` just logs the error, but the same happens inside libfirewall, where we have a nicer error repr, so maybe there is no point in logging that twice? As we made the callback error-less, probably we can make this one error-less too.

crates/telio-firewall/src/firewall.rs
stalowyjez: We probably cannot do better than just logging it probably, but libfirewall part already does that, so what handling would be good here?

crates/telio-firewall/benches/firewall_bench.rs
Hasan6979: Removing it lets go of of the checks this assert makes. Maybe some other way to ensure this ?

crates/telio-firewall/src/log.rs
Hasan6979: I don't think that we should panic here. Better safe :D

crates/telio-firewall/src/firewall.rs
Hasan6979: Rename the const `ALL_IPS` to `ALL_IP_V6` since this is being used to block Ipv6 packets

crates/telio-firewall/src/firewall.rs
Hasan6979: Just for information .. why not ?

crates/telio-firewall/src/firewall.rs
Hasan6979: If libfw already handles the logging, then logging would be redundant.

crates/telio-firewall/src/firewall.rs
Hasan6979: This is not testing the internal LRU functionality but checking whether connections were actually added or not and which ones were added. Removing this removes that check.

crates/telio-firewall/src/firewall.rs
Hasan6979: Yep. Logging works!

crates/telio-firewall/src/firewall.rs
stalowyjez: I mean, we just don't use it in any way, is there any particular usecase for it which you have on your mind?

crates/telio-firewall/benches/firewall_bench.rs
stalowyjez: Hmm, I think that would be rather hard, what this layer should expect from firewall is just result, whether that packet was accepted/rejected/dropped/invalid, but it shouldn't make assumptions about how it works IMHO.
Libfirewall internals should be tested by libfirewall unit tests, to some extent also proptests, but not by a different library ^^'

crates/telio-firewall/src/log.rs
stalowyjez: So maybe just skipping the log in such case (which cannot happen) is better? wdyt?

crates/telio-firewall/src/firewall.rs
stalowyjez: Fixed

crates/telio-firewall/src/firewall.rs
stalowyjez: Removed

crates/telio-firewall/src/firewall.rs
stalowyjez: Added log

crates/telio-firewall/src/firewall.rs
stalowyjez: Hmm, added some checks on this connection adding in conntrack.rs, wdyt?

Hasan6979:
+1

PR #1484, opened 2025-10-01
Prevent OpenWRT DNS leaks

LukasPukenis:
### Problem
**On OpenWRT device:**
Use `nordvpnlite` to connect the OpenWRT device to the VPN(via `nordvpnlite start config.json`).

**On test device:**
1) Set OpenWRT device as a network gateway.
2) Point to OpenWRT as a DNS server or leave DNS set to automatic(OpenWRT will advertise itself via DHCP as a DNS server).
3) Visit https://www.whatismyip.com/ and observe VPN server IP instead of your own.
4) Visit https://dnsleaktest.com/. Observe origin country leaked via DNS.

### Solution
Please see commits

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/config.rs
+++ b/clis/nordvpnlite/src/config.rs
@@ -67,6 +67,7 @@ fn reconnect_after_expiry_default() -> Percentage {
     Percentage(90)
 }
 
+// These IPs are documented but the documentation is not publicly available
 fn dns_default() -> Vec<IpAddr> {
     vec![
         Ipv4Addr::new(103, 86, 96, 100).into(),
--- a/clis/nordvpnlite/src/interface.rs
+++ b/clis/nordvpnlite/src/interface.rs
@@ -607,6 +607,7 @@ impl Uci {
 
         execute(Command::new("uci").args(["commit", "dhcp"]))?;
         execute(Command::new("service").args(["dnsmasq", "restart"]))?;
+        execute(Command::new("/etc/init.d/dnsmasq").args(["reload"]))?;
 
         Ok(())
     }
@@ -640,7 +641,7 @@ impl Uci {
         }
 
         execute(Command::new("uci").args(["commit", "dhcp"]))?;
-        execute(Command::new("service").args(["dnsmasq", "restart"]))?;
+        execute(Command::new("/etc/init.d/dnsmasq").args(["reload"]))?;
 
         Ok(())
     }


clis/nordvpnlite/src/command_listener.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
 };
 
 // TODO: reduce to 1 when investigating LLT-6693
-pub(crate) const TIMEOUT_SEC: u64 = 2;
+pub(crate) const TIMEOUT_SEC: u64 = 15;
```
LukasPukenis: Note to self: remove before merging

clis/nordvpnlite/src/daemon.rs
```diff
@@ -410,6 +410,7 @@ pub async fn daemon_event_loop(config: NordVpnLiteConfig) -> Result<(), NordVpnL
                     Ok(command) => {
                         debug!("Client command {:?} executed successfully", command);
                         if command == ClientCmd::QuitDaemon {
+                            std::thread::sleep(std::time::Duration::from_secs(5));
```
LukasPukenis: Note to self: remove before merging

nat-lab/bin/openwrt/40-dns.sh
```diff
@@ -2,7 +2,6 @@
 set -ex
 uci set dhcp.@dnsmasq[0].noresolv="1"
 uci add_list dhcp.@dnsmasq[0].server="10.0.80.82"
-uci add_list dhcp.@dnsmasq[0].server="10.0.80.83"
```
LukasPukenis, line 5: 10.0.80.82 and 10.0.80.83 are both NatLab DNS servers.

Just one DNS server is enough for testing OpenWRT testcases. I remove the second one here so I could test if setting DNS via configuration file works(I set it to 10.0.80.83 and expect it to be restored afterwards to 10.0.80.82).

mathiaspeters:
I think the config options for DNS should be simplified to 
1. default value if omitted
2. custom value if specified

I don't see a value in using `system` option since that has been shown to be leaky, which means that it should, in my opinion, not even be available as an option when VPN is connected. And I think `recommended` is a good option for the VPN server, but not for the DNS.

Other than that, looks good

LukasPukenis:
@mathiaspeters I had a lot of thought about the points you raised when devleoping and consciously added `system` option to make users be able to use VPN but custom DNS in case they need it, for example if NordVPN DNS would not resolve things or blocks something unecessarily.

I think it's good to have it separated but I can also see your POV.

Both `dns` and `server` default to `recommended` so if both or either is emitted, `recommended` is used accordingly :)

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/config.json
```diff
@@ -8,6 +8,9 @@
     "name": "nordvpnlite",
     "config_provider": "uci"
   },
-  "vpn": "recommended"
```
LukasPukenis, line 11: Note to self: update nordvpnlite/README.md

LukasPukenis:
@mathiaspeters I followed your advice on the configuration. I still like the earlier version of explicitness but realized that having less options limits the paths for bugs to appear which is what I prefer.

clis/nordvpnlite/src/config.rs
```diff
@@ -67,6 +67,13 @@ fn reconnect_after_expiry_default() -> Percentage {
     Percentage(90)
 }
 
+fn dns_default() -> Vec<IpAddr> {
+    vec![
+        Ipv4Addr::new(103, 86, 96, 100).into(),
```
matislovas, line 73: Maybe some additional comment, on where are those server ip's are coming from.

clis/nordvpnlite/src/interface.rs
```diff
@@ -531,14 +553,96 @@ impl Uci {
...
+            .ok();
+
+        for ip in ips {
+            execute(
+                Command::new("uci").args(["add_list", &format!("dhcp.@dnsmasq[0].server={}", ip)]),
+            )?;
+        }
+
+        execute(Command::new("uci").args(["commit", "dhcp"]))?;
+        execute(Command::new("service").args(["dnsmasq", "restart"]))?;
```
matislovas, line 609: From my own xp, the `service` command wasn't always available in OpenWRT (actually idk in which package it is included, maybe in `busybox` ..?), so I leaned more to use `/etc/init.d/<service_name> restart (or reload)`.

matislovas (approved):

clis/nordvpnlite/src/interface.rs
LukasPukenis: Good point. Though I was basing my decision on the following url: https://openwrt.org/docs/guide-user/base-system/dhcp.dnsmasq which uses the `service`. I gave it a quick experiment and it appears as `service` is just a wrapper that also awaits for the command to complete( via init.d we need to await ourselves for the proper status ). I will try to push the service -> init.d conversion and see if the tests still pass. If it's a quick win I will leave it like that instead :)

clis/nordvpnlite/src/config.rs
LukasPukenis: I left a comment that it's internally documented(cannot share the link since it's a public repo) :)

PR #1485, opened 2025-10-02
Verify gRPC connection over TLS against hardcoded root certificate

tomaszklak:
We decided to use one hardcoded root certificate that will be used to issue intermediary certificates that will in turn be used to create leaf certificates. To verify the TLS connection, libtelio needs to know the root certificate which is now hardcoded in the ens.rs.

To make it possible to test this functionality (and have an option to change in production) an option to override the root cert is also provided in the ENS feature. This uncovered an issue with Pyro5 - serialization of byte arrays is not (by default) transparent, one needs to manually extract the slice out of the created dictionary that contains base64 encoded string. To fix this, we use slower method that is using python's native `repr` to serialize.

- [x] README.md is updated


nat-lab/bin/ens
```diff
@@ -96,52 +98,177 @@ class AuthenticationInterceptor(grpc.aio.ServerInterceptor):
...
+    ])
+    
+    root_cert = (
+        x509.CertificateBuilder()
+        .subject_name(root_subject)
+        .issuer_name(root_subject)  # Self-signed
+        .public_key(root_private_key.public_key())
+        .serial_number(x509.random_serial_number())
+        .not_valid_before(datetime.utcnow() - timedelta(days=1))
+        .not_valid_after(datetime.utcnow() + timedelta(days=7))  # 10 years
```
mathiaspeters: This comment looks wrong. 7 days instead of 10 years

tomaszklak: You are right, it's outdated. Fixed.

crates/telio-model/src/features.rs
```diff
@@ -583,6 +584,27 @@ pub struct FeatureErrorNotificationService {
     /// Configuration of the backoff algorithm used by ENS
     #[serde(default)]
     pub backoff: Backoff,
+
+    /// DER encoded root certificate to be used for verification of all TLS connections
+    /// to gRPC ENS endpoint in place of the hardcoded one
+    pub root_certificate_override: Option<Vec<u8>>,
```
LukasPukenis, line 590: question: could you elaborate a bit on the "in place of the hardcoded one" - does it refer to OS cert list?

nat-lab/bin/ens
```diff
@@ -232,3 +419,5 @@ async def main():
 
 if __name__ == "__main__":
     asyncio.run(main())
+
+
```
LukasPukenis: nit: 2 newlines

nat-lab/tests/conftest.py
```diff
@@ -453,6 +454,7 @@ def pytest_runtest_setup():
 
 # pylint: disable=unused-argument
 def pytest_sessionstart(session):
+    Pyro5.config.SERPENT_BYTES_REPR = True
```
LukasPukenis, line 457: question: curious, how to reproduce issue where this is the fix?

nat-lab/tests/test_vpn.py
```diff
@@ -914,6 +919,138 @@ async def test_ens(
         await client_alpha.wait_for_log(fingerprint)
 
 
+@pytest.mark.parametrize(
+    "alpha_setup_params, public_ip",
+    [
+        pytest.param(
+            SetupParameters(
+                connection_tag=ConnectionTag.DOCKER_CONE_CLIENT_1,
+                adapter_type_override=TelioAdapterType.NEP_TUN,
```
LukasPukenis: question: as per my understanding ENS should be indifferent to adapter types, is it worth to test with multiple adapters?

nat-lab/tests/test_vpn.py
```diff
@@ -914,6 +919,138 @@ async def test_ens(
...
+                stun_limits=(1, 1),
+                vpn_1_limits=(
+                    (1, 1)
+                    if vpn_conf.conn_tag == ConnectionTag.DOCKER_VPN_1
+                    else (0, 0)
+                ),
+            )
+        )
+        assert alpha_setup_params.features.error_notification_service
+        alpha_setup_params.features.error_notification_service.allow_only_pq = False
```
LukasPukenis: PQ safe connection was mentioned as a requirement in the RFC IIRC, why is it configurable?

nat-lab/tests/test_vpn.py
```diff
@@ -914,6 +919,138 @@ async def test_ens(
...
+async def test_ens_will_not_emit_errors_from_incorrect_tls_session(
+    alpha_setup_params: SetupParameters,
+    public_ip: str,
+) -> None:
+    vpn_conf = VpnConfig(config.WG_SERVER, ConnectionTag.DOCKER_VPN_1, True)
+    fingerprint = await get_grpc_tls_fingerprint(vpn_conf.server_conf["ipv4"])
+    root_certificate = await get_grpc_tls_root_certificate(vpn_conf.server_conf["ipv4"], incorrect=True)
+    root_certificate = base64.b64decode(root_certificate)
+    error_code_id = 0
+    error_code = VpnConnectionError.UNKNOWN
```
LukasPukenis: isn't `error_code_id == error_code` since `VpnConnectionError.UNKNOWN = 0`? I wonder if these two refer to the same thing.

nat-lab/tests/test_vpn.py
```diff
@@ -914,6 +919,138 @@ async def test_ens(
...
+        with pytest.raises(asyncio.TimeoutError):
+            await client_alpha.wait_for_state_peer(
+                vpn_conf.server_conf["public_key"],
+                [NodeState.CONNECTED],
+                [PathType.DIRECT],
+                True,
+                True,
+                timeout=15,
+                vpn_connection_error=error_code,
+            )
```
LukasPukenis: question: I am unfamiliar with ENS(should read RFC) but I wonder how are the errors mapped to the session, for example if two errors are pushed from the server?

crates/telio-proto/src/ens.rs
```diff
@@ -85,6 +86,8 @@ pub struct ErrorNotificationService {
     tx: Tx<(ConnectionError, PublicKey)>,
     socket_pool: Arc<SocketPool>,
     allow_only_mlkem: bool,
+    // DER encoded root certificate to be use for verification of TLS
+    root_certificate: Vec<u8>,
```
LukasPukenis, line 90: curious: where did you take the root certificate from? :)

nat-lab/tests/test_ens.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if alpha_setup_params.connection_tag == ConnectionTag.VM_MAC:
+            await ensure_interface_router_property_expectations(client_conn)
+
+        ip = await stun.get(client_conn, config.STUN_SERVER)
+        assert ip == public_ip, f"wrong public IP before connecting to VPN {ip}"
+
+        vpn_connection, *_ = await setup_connections(exit_stack, [vpn_conf.conn_tag])
+        await exit_stack.enter_async_context(
+            client_alpha.get_router().block_tcp_port(993)
```
LukasPukenis: nit: 993 could be extracted as `ENS_PORT`

LukasPukenis (commented):
+1

Left few questions that made me curious but the changes themselves look good to me

nat-lab/tests/conftest.py
tomaszklak: I think it's enough to send `bytes` python type across Pyro connection. In this particular case, it was triggered by the new feature config flag that allow for setting root certificate override. On the rust side it's a `Vec<u8>` which on the python side is reprsented by `bytes`. The problem is happening when a whole `Features` is sent to the `libtelio_remote.py` when the `libtelio` is to be started.

nat-lab/tests/test_vpn.py
tomaszklak: Currently, there is no way to create python grpc server that will use (any) post-quantum key exchange algorithm.

nat-lab/tests/test_vpn.py
tomaszklak: The app is not required to react to any of the errors returned by ENS. Which means that it's possible (although unlikely in practice) that the first error is ignored and a subsequent error is received for the same session. And contrived example of that is the server maintenance error which could be ignored by the app, and the session can be kept open - after all that error indicates upcoming (hours) maintenance, not an immediate one.

LukasPukenis: Here I wondered(but improperly formulated my question) about the case where I connect to server A, then to server B and would receive messages dedicated to the session with server A. That was my misunderstanding - I read the RFC and it states that the ENS service is running on each VPN server so this cannot happen :) For some reason I imagined ENS as some separate entity.

tomaszklak: Yes, each VPN server will now run both the NordLynx and ENS services at the same time.

nat-lab/bin/ens
tomaszklak: Done

nat-lab/tests/test_vpn.py
tomaszklak: Good point, removed duplicate Linux versions.

nat-lab/tests/test_vpn.py
tomaszklak: You are right, I forgot that I can access this value in Python usinge the `value` field.

nat-lab/tests/test_ens.py
tomaszklak: Done

nat-lab/tests/test_vpn.py
LukasPukenis: Oh okay, it's fine then :)

nat-lab/tests/conftest.py
LukasPukenis: Thank you for the explanation <3

mathiaspeters (approved):
+1

PR #1487, opened 2025-10-02
LLT-6694: select loop race condition

tomasz-grz:
### Problem
When client connections are accepted inside of `try_recv_quit()`, there is a possibility the connection might break, before the reply is sent to the client and the connection is properly closed. 
This can happen if the `try_recv_quit` future is canceled halfway in, after `request_nordlynx_key()` is done:

```rust
select! {
  nordlynx_private_key = &mut api_request_future => break nordlynx_private_key?,
  res = cmd_listener.try_recv_quit() => {...},
  _ = signals.next() => {...}
};
```

This is possibly causing an occasional `Error: Io(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })` error ( and suspected `Error: Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })` error)

There is also a bit of code duplication between regular event loop and handling early events while fetching the keys.

### Solution
- `IsAlive` command now returns true all stages, indicating that the daemon is already alive.
- New command response added, `DaemonInitializing`, indicating that the daemon is not ready to process events (is fetching credentials or TelioTask has not started yet).
- Separate `accept_client_connection()` to match quickly on a `tokio::select!` branch. The reply is sent inside the select branch, where there is no danger of it being cancelled.
- Refactored `handle_client_connection()` to handle both cases, early quit and normal processing.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -206,7 +206,7 @@ impl CommandListener {
                         CommandResponse::Err(format!("Invalid command: {command_str}")).serialize(),
                     )
                     .await?;
-                Err(NordVpnLiteError::InvalidCommand(command_str))
+                Err(err.into())
             }
         }
     }
@@ -350,7 +350,7 @@ mod tests {
         let response = client_send_command(&path, command).await;
         let cmd = daemon.await.unwrap();
 
-        assert_matches!(cmd, Err(NordVpnLiteError::InvalidCommand(_)));
+        assert_matches!(cmd, Err(NordVpnLiteError::ParsingError(_)));
         assert_matches!(response, Ok(CommandResponse::Err(_)));
     }
 
--- a/nat-lab/tests/nordvpnlite.py
+++ b/nat-lab/tests/nordvpnlite.py
@@ -238,21 +238,21 @@ class NordVpnLite:
             stdout, _ = await self.execute_command(["is-alive"])
             return "Command executed successfully" in stdout
         except ProcessExecError as exc:
-            if "Obtaining nordlynx key, ignoring" in exc.stdout:
-                raise IgnoreableError() from exc
             if "Error: DaemonIsNotRunning" in exc.stderr:
                 return False
-            # TODO: remove after LLT-6694
-            if "Connection reset by peer" in exc.stderr:
-                raise IgnoreableError() from exc
-            # TODO: remove after LLT-6694
-            if "Broken pipe" in exc.stderr:
-                raise IgnoreableError() from exc
             raise exc
 
     async def get_status(self) -> str:
-        status, _ = await self.execute_command(["get-status"])
-        return status
+        try:
+            status, _ = await self.execute_command(["get-status"])
+            return status
+        except ProcessExecError as exc:
+            if "Daemon is not ready, ignoring" in exc.stdout:
+                raise IgnoreableError() from exc
+            # TODO: remove after LLT-6693
+            if "ClientTimeoutError" in exc.stdout:
+                raise IgnoreableError() from exc
+            raise exc
 
     async def quit(self) -> None:
         stdout, stderr = await self.execute_command(["quit-daemon"])
@@ -326,11 +326,28 @@ class NordVpnLite:
         # TODO: remove after LLT-6693
         await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
         while True:
-            status = json.loads(await self.get_status())
-            if status["exit_node"]:
-                if status["exit_node"]["state"] == "connected":
+            try:
+                status = json.loads(await self.get_status())
+                if status["exit_node"]:
+                    if status["exit_node"]["state"] == "connected":
+                        return
+                await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
+            except IgnoreableError:
+                await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
+                continue
+
+    async def wait_for_telio_running_status(self):
+        # TODO: remove after LLT-6693
+        await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
+        while True:
+            try:
+                status = json.loads(await self.get_status())
+                if status["telio_is_running"]:
                     return
-            await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
+                await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
+            except IgnoreableError:
+                await asyncio.sleep(self.NORDVPNLITE_CMD_CHECK_INTERVAL_S)
+                continue
 
     @asynccontextmanager
     async def setup_interface(self, vpn_routes: bool) -> AsyncIterator:
--- a/nat-lab/tests/test_nordvpnlite.py
+++ b/nat-lab/tests/test_nordvpnlite.py
@@ -40,7 +40,7 @@ async def test_nordvpnlite_logs() -> None:
 
         nordvpnlite = await NordVpnLite.new(exit_stack, connection=connection)
         async with nordvpnlite.start():
-            pass
+            await nordvpnlite.wait_for_telio_running_status()
 
         expected_log_contents = {
             str(nordvpnlite.config.paths.daemon_log): "task started",


clis/nordvpnlite/src/command_listener.rs
```diff
@@ -159,52 +170,39 @@ impl CommandListener {
         }
     }
 
-    pub async fn try_recv_quit(&mut self) -> Result<(), NordVpnLiteError> {
-        let mut connection = self.socket.accept().await?;
+    pub async fn accept_client_connection(&mut self) -> Result<DaemonConnection, NordVpnLiteError> {
+        Ok(self.socket.accept().await?)
```
tomaszklak, line 171: Is the `accept` called here, cancel safe?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -159,52 +170,39 @@ impl CommandListener {
...
-            Err(e) => {
+            Err(err) => {
+                error!("Failed parsing command {command_str}: {err:?}");
                 connection
                     .respond(
                         CommandResponse::Err(format!("Invalid command: {command_str}")).serialize(),
                     )
                     .await?;
-                Err(e.into())
+                Err(NordVpnLiteError::InvalidCommand(command_str))
```
tomaszklak: It would be good idea to capture the `err` and include it in the returned error value.

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -257,13 +255,43 @@ mod tests {
...
     // Broken client, closes connection without waiting for response
-    async fn broken_client_send_command(path: &str, cmd: &str) -> std::io::Result<()> {
+    async fn broken_client_send_command(path: &str, cmd: &str) -> std::io::Result<CommandResponse> {
         let mut client_stream = UnixStream::connect(&Path::new(path)).await?;
         client_stream
             .write_all(format!("{}\n", cmd).as_bytes())
             .await?;
 
-        Ok(())
+        Ok(CommandResponse::Ok)
```
tomaszklak: nit: I'm not sure why did we change the return type here, can you explain?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -121,6 +122,16 @@ impl CommandListener {
         }
     }
 
+    // Quick command handling before telio task is initialized
+    fn early_process_command(&mut self, command: &ClientCmd) -> CommandResponse {
+        match command {
+            ClientCmd::QuitDaemon => CommandResponse::Ok,
+            ClientCmd::IsAlive => CommandResponse::Ok,
```
tomaszklak: Why are we returning `Ok` in the initializing phase for those 2 commands? I could maybe see how `IsAlive` returns `Ok` (but it's surprising) but `QuitDaemon`? Please explain the reasoning - am I missing something important about nordvpnlite here?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -298,25 +339,77 @@ mod tests {
...
+            let connection = listener.accept_client_connection().await.unwrap();
+            listener
+                .handle_client_command(DaemonState::Ready, connection)
+                .await
+        });
+        let response = client_send_command(&path, command).await;
+        let cmd = daemon.await.unwrap();
 
-        assert!(matches!(cmd, Err(NordVpnLiteError::InvalidCommand(_))));
+        assert!(
```
tomaszklak: nit: you could use https://docs.rs/assert_matches/latest/assert_matches/macro.assert_matches.html - it produces nice error messages already with both the expected pattern and the actual value.

nat-lab/tests/nordvpnlite.py
```diff
@@ -238,21 +238,18 @@ async def is_alive(self) -> bool:
...
 
     async def get_status(self) -> str:
-        status, _ = await self.execute_command(["get-status"])
-        return status
+        try:
+            status, _ = await self.execute_command(["get-status"])
+            return status
+        except ProcessExecError as exc:
+            if "Daemon is not ready, ignoring" in exc.stdout:
+                raise IgnoreableError() from exc
```
tomaszklak, line 251: TIL, thanks! :)

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -159,52 +170,39 @@ impl CommandListener {
...
-                    )
-                    .await?;
-                Err(NordVpnLiteError::InvalidCommand(command_str))
+            Ok(command) => {
+                // daemon is still initializing, only QuitDaemon and IsAlive allowed
+                let response = if state < DaemonState::Ready {
+                    self.early_process_command(&command)
+                } else {
+                    self.process_command(&command).await?
+                };
```
LukasPukenis: thoughts: inlining the `early_process_command` function here might make it more obvious since it isn't used anywhere else, also it would eliminate the comment :)

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -257,13 +255,43 @@ mod tests {
     }
 
     // Broken client, closes connection without waiting for response
-    async fn broken_client_send_command(path: &str, cmd: &str) -> std::io::Result<()> {
+    async fn broken_client_send_command(path: &str, cmd: &str) -> std::io::Result<CommandResponse> {
```
LukasPukenis: It looks like it's a normal function that connects and writes to the socket, why the word _broken_?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -257,13 +255,43 @@ mod tests {
...
         let mut client_stream = UnixStream::connect(&Path::new(path)).await?;
         client_stream
             .write_all(format!("{}\n", cmd).as_bytes())
             .await?;
 
-        Ok(())
+        Ok(CommandResponse::Ok)
+    }
+
+    async fn test_command_helper(
```
LukasPukenis, line 275: question: could we eliminate `test_command_helper` and `broken_client_send_command()` and use the normal flow instead? I think they are not really necessary and we could use dependency injection if needed to simulate, for example, a broken client. That way we would use and validate the same code which runs in production.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -120,6 +120,15 @@ impl ExitNodeStatus {
     }
 }
 
+/// Current stage of the daemon
+/// It is ininitializing state while fetching credentials and before starting the telio task
```
LukasPukenis:
```suggestion
/// It is in initializing state while fetching credentials and before starting the telio task
```

clis/nordvpnlite/src/daemon.rs
```diff
@@ -351,15 +360,25 @@ pub async fn daemon_event_loop(config: NordVpnLiteConfig) -> Result<(), NordVpnL
...
-                        Ok(_) => {
-                            info!("Received quit command, exiting");
-                            return Ok(())
-                        },
-                        Err(e) => {
-                            debug!("Received command {:?} while obtaining service credentials, ignoring", e);
+                connection_result = cmd_listener.accept_client_connection() => {
+                    match connection_result {
+                        Ok(connection) => {
+                            match cmd_listener.handle_client_command(DaemonState::Initializing, connection).await {
```
LukasPukenis: Why isn't the daemon state bind to it and instead explicitly passed? I think this is also used just for testing purposes :thinking:

clis/nordvpnlite/src/daemon.rs
```diff
@@ -120,6 +120,15 @@ impl ExitNodeStatus {
...
 
+/// Current stage of the daemon
+/// It is ininitializing state while fetching credentials and before starting the telio task
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Default)]
+pub enum DaemonState {
+    #[default]
+    Initializing,
+    Ready,
+}
+
```
LukasPukenis: Thoughts: IMO we could eliminate the daemon state and some of the complexity if we would introduce a separate command for vpn like `vpn <on|off>`, that way the daemon would either be running and ready to accept commands(just open socket and listen) or not running at all. I think it would result in a much simpler design and less edge cases.

tomasz-grz: Hmm I'm afraid that introducing a separate `vpn on|off` command might introduce even more complexity regarding the states..

I do like the idea of having a separate command though, but that sounds like a completely different task\requirement from this.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: It is not absolutely clear to me what you mean here 😅 could you please rephrase?

Do you mean why do we pass `DaemonState::Initializing` as a literal value, instead of a variable? 🤔

This is not used for testing in this instance, but for the "early" quit if we want to abort fetching the credentials.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: Good catch 👀

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: As mentioned in the code comment above, we close the client socket before reading a response, which means that the daemon `handle_client_command()` will fail at `connection.respond()`.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: We do validate some code path for `CommandListener` that the daemon runs in production here. Just without the Daemon TelioTask part as there is no way to "retrieve" the events it is receiving there. They are not really necessary but I used them mostly for TDD without having to run the daemon/client all the time.

I think testing the full path with the proper Client and Daemon with the TelioTask running would be an integration test and not a unit test anymore, and should be covered by natlab. Could you give maybe an example how we could use dependency injection here?

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Good idea!

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Yes, these are the IPC messages that the daemon is sending back to the client after receiving a command from the client.

While the daemon is `Initializing`, it is fetching credentials from the CoreAPI. A process that can take a while or even get stuck (usually if you have misconfigured NordLayer).
In case it is stuck in this phase, we would like it to still be able react nicely to`nordvpnlite quit-daemon` and clean up after itself.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Because `broken_client_send_command()` and `client_send_command()` had mismatching return types.

```rust
        let daemon_response = if broken_client {
            broken_client_send_command(&path, &command).await
        } else {
            client_send_command(&path, &command).await
        };
```

Though in this case... logically `CommandResponse::Ok` is not correct, since the `broken_client_send_command` does never actually read a `CommandResponse` from the daemon.. 🤔

tomasz-grz: Maybe I'll do something like this:

```rust
        let daemon_response = if broken_client {
            broken_client_send_command(&path, &command).await.unwrap();
            None
        }
```

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Hmm yeah fair point

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: If this is wrapping `UnixListener::accept()` it should be indeed [cancel safe](https://docs.rs/tokio/latest/tokio/net/struct.UnixListener.html#cancel-safety).

Otherwise it still reduces the surface area of dropping the connection prematurely.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: But I can remove these states again by maybe exposing `early_process_command` and `process_command`, I mostly wanted to reduce code duplication from `handle_client_command`..

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Hmm actually good idea!

clis/nordvpnlite/src/daemon.rs
tomasz-grz: Or can simply be a boolean? Though I personally prefer an enum to make it more explicit

```rust
    pub async fn handle_client_command(
        &mut self,
        ready: bool,
        mut connection: DaemonConnection,
    ) -> Result<ClientCmd, NordVpnLiteError>
```

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: But in that case it would be a serder error like `Err(ParsingError(Error("expected value", line: 1, column: 1)))` instead of `Err(NordVpnLiteError::InvalidCommand("sudo"))`. Is that more descriptive?

We do log the serde error above, so it's not lost.
```rust
    error!("Failed parsing command {command_str}: {err:?}");
```

clis/nordvpnlite/src/command_listener.rs
LukasPukenis: It's in `mod tests` which explains it, I was rather looking into it as a normal function :D However still, the function doesn't break anything on it's own..?

clis/nordvpnlite/src/daemon.rs
LukasPukenis: Sure thing :) so to me it sort of looks weird that we pass `DaemonState::Initializing` - why do we even need to do this? After all, the daemon state should be an internal thing Daemon is aware of itself so I questioned it's existence.

clis/nordvpnlite/src/daemon.rs
LukasPukenis: I understand that it looks out of scope of this PR but this PR increases the complexity of introducing more state management IMO and thinking about `vpn <on,off>` makes me think that we should eliminate it altogether - daemon starts, tries to bind to socket and read config, fails, that's it - no more state tracking that it started but is fetching the credentials and other stuff. I am not saying that the current PR doesn't fix the problem though, what I am saying is that maybe we can eliminate the problem and complexity in one go :)

clis/nordvpnlite/src/command_listener.rs
LukasPukenis: To me it just sort of red flagged to see unit tests that:
- establish socket
- custom command listener
- pass daemon state explicitly

This looks like it bypasses the `nordvpnlite` itself but obviously, it is expected to some degree since these are unit tests. I just wonder if this is the right approach. Feels more suited to NatLab.

I will try to think about it for more time but in the meanwhile this comment is a no blocker(marked as a question) :)

clis/nordvpnlite/src/command_listener.rs
tomaszklak: In that case I think it would be good idea to return both the command and the error. But it's a minor thing, feel free to ignore it and treat this as a nitpick :)

clis/nordvpnlite/src/command_listener.rs
tomaszklak: But does it mean that sending quit during initialization will be ignored?

clis/nordvpnlite/src/command_listener.rs
tomaszklak: Since this PR is solving an issue caused by cancellation, I think it would be good to add a comment here explaining why we think that this is cancel safe (or at least why it is better this way).

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: `CommandResponse::Ok` will be sent to the client side. And the daemon event loop will get `Ok(ClientCmd::QuitDaemon)` returned from [handle_client_command()](https://github.com/NordSecurity/libtelio/blob/0217d0603f198313fea5831b6a56d2138449bffc/clis/nordvpnlite/src/command_listener.rs#L188), breaking the [event loop](https://github.com/NordSecurity/libtelio/blob/0217d0603f198313fea5831b6a56d2138449bffc/clis/nordvpnlite/src/daemon.rs#L369) and cancelling the daemon startup.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: After some refactoring the "self awareness" of the daemon was moved into `TelioContext`, which is only part of TelioTask...
So in-fact, Dameon is not aware at all of itself, only the TelioTask is, which we have no access to here as it's spawned as a separate task when the Daemon is in `Ready` state..

Therefore declaring a variable for `DaemonState` would make little sense as it would be never read by anything, that's why I preferred using a Literal Value.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: Alright, I removed `DaemonState` in favor of bool

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Updated

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Updated

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: It breaks what would be the daemon's socket connection, by closing it, so there is no-where to write the `CommandResponse`.

This is meant to test that the daemon's `CommandListener` can handle the client unexpectedly closing.

Under normal circumstances the client should wait and read the `CommandResponse` from the Daemon before closing the socket connection.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: This is **the** `CommandListener` used by the daemon, there is nothing custom about it. The rest is just the scaffolding around it since we don't have a full Daemon or Client process here.

And precisely it bypasses `nordvpnlite`, since it is only meant to test IPC communication part and "command & response" flow of the `CommandListener` module. As we can easily access both sides of the IPC here in one test, not something we can do in natlab. We could only parse the daemon logs there to cover these scenarios.

I thought having more code coverage is better, but I can certainly remove all the tests added here.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: Removed `DaemonState`

clis/nordvpnlite/src/daemon.rs
tomasz-grz: Updated

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: I agree it would be better for error handling in general, but in this case we simply ignore all errors from the daemon side.

So there would be no difference in bubbling up the exact Serde error.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Updated

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Added comment

clis/nordvpnlite/src/command_listener.rs
LukasPukenis: Thanks for taking your time to explain, it's clearer to me now. No need to remove the tests, they are useful.

clis/nordvpnlite/src/command_listener.rs
tomaszklak: This reasoning means that returning anything more than an indication of an error is pointless, so we should also drop the `command_str` from the enum variant. I think we can find soon enough that this code will be also run in a different context (maybe there is some unreleased device that might use it ;>). But like I said, this is minor - feel free to keep it as it is if you feel changing this is not worth the trouble.

clis/nordvpnlite/src/daemon.rs
LukasPukenis: I reviewed the code again and noticed few things I didn't initially catch on. This PR seems to be an improvement over existing code so I am considering this a good change.

Just to connect the dots - in one of the threads I suggested an idea of `vpn <on,off>` with hopes that the daemon launch would simply establish command socket and daemonize the process, thus resulting in a single phase command handling. That was my idea with the suggestion :)

clis/nordvpnlite/src/command_listener.rs
LukasPukenis: Okay.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Fair point, updated!

LukasPukenis (approved):

PR #1488, opened 2025-10-02
Add OpenWrt tests on reconnect to vpn

asledzinskiy:
- Add tests on disconnect and re-connect to vpn with the same config and config with the different country

### Problem
*--Check that user can disconnect from vpn and then connect again--*

### DIFF
--- a/nat-lab/tests/test_openwrt.py
+++ b/nat-lab/tests/test_openwrt.py
@@ -1,13 +1,59 @@
 import pytest
-from config import WG_SERVER, PHOTO_ALBUM_IP, STUN_SERVER
+from config import WG_SERVER, WG_SERVER_2, PHOTO_ALBUM_IP, STUN_SERVER
 from contextlib import AsyncExitStack
 from helpers import setup_connections
 from nordvpnlite import NordVpnLite, Config, IfcConfigType, Paths
 from pathlib import Path
 from utils import stun
-from utils.connection import ConnectionTag
+from utils.connection import Connection, ConnectionTag
 from utils.logger import log
 from utils.ping import ping
+
+
+async def check_gateway_and_client_ip(
+    gateway_connection: Connection,
+    client_connection: Connection,
+    expected_ip: str | int,
+) -> None:
+    """
+    Verify that both the OpenWRT gateway and the client device obtain the expected
+    IP address when connected to or disconnected from the VPN.
+
+    This function performs two checks:
+      1. Ensures that the gateway can reach PHOTO_ALBUM_IP and that
+         its public IP, as reported by STUN, matches the expected IP.
+      2. Ensures that the client device can also reach the same host and that
+         its public IP matches the expected IP.
+
+    Args:
+        gateway_connection (Connection):
+            An active SSH or Docker connection to the OpenWRT gateway.
+        client_connection (Connection):
+            An active SSH or Docker connection to the client device behind the gateway.
+        expected_ip (str):
+            The public IP address that both the gateway and client are expected to have
+            when connected to or disconnected from the VPN.
+
+    Raises:
+        AssertionError:
+            If either the gateway or the client reports a public IP different from
+            `expected_ip`.
+    Returns:
+        None
+    """
+    await ping(gateway_connection, PHOTO_ALBUM_IP)
+    gw_ip = await stun.get(gateway_connection, STUN_SERVER)
+    assert gw_ip == expected_ip, (
+        f"OpenWRT gateway has wrong public IP when connected to VPN: {gw_ip}. "
+        f"Expected value: {expected_ip}"
+    )
+
+    await ping(client_connection, PHOTO_ALBUM_IP)
+    client_ip = await stun.get(client_connection, STUN_SERVER)
+    assert client_ip == expected_ip, (
+        f"Client device has wrong public IP when connected to VPN: {client_ip}. "
+        f"Expected value: {expected_ip}"
+    )
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
             )
         await wait_for_log_line(logread_proc)
         log.info("Network has been reloaded")
+
+
```
asledzinskiy, line 324: I thought to parametrize and make it as one test but because if slight difference between the tests I don't like the readability and that in case of reconnect to the same country we will have unnecessary second instance of Nordvpnlite

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+    Steps:
+        1. Prepare vpn servers
+        2. Start NordVPN Lite in OpenWRT container
+        3. Check ip address of both openwrt container and client node is equal to vpn ip address
+        4. Ping PHOTO_ALBUM_IP from both openwrt and client node
+        5. Disconnect from vpn
+        6. Check PHOTO_ALBUM_IP is reachable and ip addresses are equal to public ips
+        7. Connect to vpn again with the same config (start NordVPN Lite)
+        8. Check ip address of both openwrt container and client node is equal to vpn ip address
+        9. Ping PHOTO_ALBUM_IP from both openwrt and client node
```
LukasPukenis, line 340: I must say that I really enjoy these descriptions. Even though they have a tendency to become stale in regular code, the testcases themselves are not usually modified often so this is very appreciated <3

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+            f"data/nordvpnlite/{IfcConfigType.VPN_OPENWRT_UCI_PL.value}",
+            f"/etc/nordvpnlite/{IfcConfigType.VPN_OPENWRT_UCI_PL.value}",
+        )
+
+        config_path = Paths(exec_path=Path("nordvpnlite"))
+        nordvpnlite = NordVpnLite(
+            gateway_connection,
+            exit_stack,
+            config=Config(IfcConfigType.VPN_OPENWRT_UCI_PL, paths=config_path),
+        )
```
LukasPukenis: nit: these could go into a separate preparation function since it's a repeating code and it's irrelevant to the testcase, we only care about the config file used - `IfcConfigType.VPN_OPENWRT_UCI_PL` here :)

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+
+        async with nordvpnlite.start():
+            log.debug("NordVPN Lite started, waiting for connected vpn state...")
+            await nordvpnlite.wait_for_vpn_connected_state()
+            await check_gateway_and_client_ip(
+                gateway_connection, client_connection, WG_SERVER["ipv4"]
+            )
+            logread_proc = await start_logread_process(
+                gateway_connection, exit_stack, NETWORK_RESTART_LOG_LINE
+            )
```
LukasPukenis, line 359: Maybe it's worth to inline this into the `nordvpnlite.start()`? The dalaying will be present _probably_ in a majority of the testcases anyway.

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+        await wait_for_log_line(logread_proc)
+
+        log.debug("Check connection after disconnect from vpn")
+        await check_gateway_and_client_ip(
+            gateway_connection, client_connection, "10.0.0.0"
```
LukasPukenis: Is it possible to use constants and not direct IP here?

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+        5. Disconnect from vpn
+        6. Check PHOTO_ALBUM_IP is reachable and ip addresses are equal to public ips
+        7. Connect to vpn again with the same config (start NordVPN Lite)
+        8. Check ip address of both openwrt container and client node is equal to vpn ip address
+        9. Ping PHOTO_ALBUM_IP from both openwrt and client node
+
+    """
+    async with AsyncExitStack() as exit_stack:
+        client_connection = (
+            await setup_connections(exit_stack, [ConnectionTag.DOCKER_OPENWRT_CLIENT_1])
```
gytsto: nit: `new_connection_by_tag()` could be used here, returns `Connection` straight away

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+        )[0].connection
+        gateway_connection = (
+            await setup_connections(exit_stack, [ConnectionTag.VM_OPENWRT_GW_1])
```
gytsto: nit: `new_connection_by_tag()` could be used here, returns `Connection` straight away

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+        async with nordvpnlite.start():
+            log.debug("NordVPN Lite started, waiting for connected vpn state...")
+            await nordvpnlite.wait_for_vpn_connected_state()
+            await check_gateway_and_client_ip(
+                gateway_connection, client_connection, WG_SERVER["ipv4"]
+            )
+            logread_proc = await start_logread_process(
+                gateway_connection, exit_stack, NETWORK_RESTART_LOG_LINE
+            )
+        await wait_for_log_line(logread_proc)
```
gytsto, line 360: Shouldn't we wait for log while nordvpnlite is still going? otherwise we have nothing to wait for, because logs won't be appended. Or just change method to `check_log`

Anyway, looks like `OutputNotifier` could be used also

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+        async with nordvpnlite.start():
+            log.debug("Reconnect to VPN...")
+            await nordvpnlite.wait_for_vpn_connected_state()
+            await check_gateway_and_client_ip(
+                gateway_connection, client_connection, WG_SERVER["ipv4"]
+            )
+            logread_proc = await start_logread_process(
+                gateway_connection, exit_stack, NETWORK_RESTART_LOG_LINE
+            )
+        await wait_for_log_line(logread_proc)
```
gytsto, line 377: Same as before

nat-lab/tests/test_openwrt.py
```diff
@@ -328,3 +328,150 @@ async def test_openwrt_simulate_network_down() -> None:
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.openwrt
+async def test_openwrt_vpn_reconnect_different_country() -> None:
```
gytsto, line 383: all comments applies from previous test

nat-lab/tests/test_openwrt.py
asledzinskiy: Sounds reasonable, will move it to separate function

nat-lab/tests/test_openwrt.py
asledzinskiy: We had this discussion [here](https://github.com/NordSecurity/libtelio/pull/1518#discussion_r2436330163) and decided that network reload after 'quit-daemon' is specific to openwrt and therefore it doesn't make sense to make it a part of Nordvpnlite class

nat-lab/tests/test_openwrt.py
asledzinskiy: Yes, let me move it to constants

nat-lab/tests/test_openwrt.py
asledzinskiy: Thanks, will check

nat-lab/tests/test_openwrt.py
asledzinskiy: I described it in this [PR](https://github.com/NordSecurity/libtelio/pull/1518) 
In OpenWrt when 'quit-daemon' is executed there is an asynchronous network reload that nordvpnlite client itself doesn't wait for.
That's why we start polling logs right before clean up -> then `quit-daemon` happens as part of context manager clean up -> then we wait for network reload log to appear in dmesg

nat-lab/tests/test_openwrt.py
asledzinskiy: Done

nat-lab/tests/test_openwrt.py
asledzinskiy: Done

nat-lab/tests/test_openwrt.py
asledzinskiy: Done

nat-lab/tests/test_openwrt.py
asledzinskiy: Done

gytsto (commented):
+1

PR #1491, opened 2025-10-06
Ignore device stopping errors when collecting logs

LukasPukenis:
### Problem
In NatLab when stopping the device fails, logs are not collected.

### Solution
Ignore failure to stop the device and still proceed with log collection.


nat-lab/tests/telio.py
```diff
@@ -579,14 +579,36 @@ async def on_stderr(stderr: str) -> None:
...
+                                "[%s] Exception while stopping device: %s. Ignoring.",
+                                self._node.name,
+                                e,
+                            )
                     else:
                         log.info(
                             "[%s] Test cleanup: We don't have LibtelioProxy instance, Stop() not called.",
                             self._node.name,
                         )
-                    self._quit = True
```
stalowyjez, line 589: Shouldn't `_quit` be set to `True` anyway?

nat-lab/tests/telio.py
```diff
@@ -579,14 +579,36 @@ async def on_stderr(stderr: str) -> None:
...
+                                "[%s] Couldn't stop the device. Ignoring.",
+                                self._node.name,
+                            )
+                        # pylint: disable=broad-except
+                        except Exception as e:
+                            log.exception(
+                                "[%s] Exception while stopping device: %s. Ignoring.",
+                                self._node.name,
+                                e,
+                            )
```
stalowyjez: Hmm, does it make sense to create two cases when we do the same in both of them?

stalowyjez (commented):
Looks good, just two questions, besides them, +1

nat-lab/tests/telio.py
LukasPukenis: As I understand it should be true on success always so here I try to adhere to the same convention and only set it if there wasn't an exception :)

nat-lab/tests/telio.py
LukasPukenis: You're right, an umbrella exception would also handle just fine but I wanted to be explicit that it is wrapped in `asyncio.timeout` :)

nat-lab/tests/telio.py
```diff
@@ -579,14 +579,36 @@ async def on_stderr(stderr: str) -> None:
...
+                        self._node.name,
+                    )
                     if self._libtelio_proxy:
-                        await self.stop_device()
+                        log.info(
+                            "[%s] Test cleanup: will stop the device",
+                            self._node.name,
+                        )
+                        try:
+                            async with asyncio.timeout(20):
```
lcruz99: 20 is a magic number, please create a const for it.

nat-lab/tests/telio.py
lcruz99: I wonder if the next test in the pytest queue will be able to launch the proxy process in this case that the stop() fails and we swallow the exception. In that case, the next fails and the user will not know - at least initially - that the previous test was actually the culprit.

Wdyt if we catch the exception but store it in order to run the rest of the cleanup and then at the end still raise it?

lcruz99: For example locally I use the pytest flag `-x` to stop execution on any test failure, with this change it will make me debug the wrong test.

nat-lab/tests/telio.py
LukasPukenis: :heavy_check_mark:

nat-lab/tests/telio.py
LukasPukenis: That is a good point, I save the exception till the end and raise it if it happened before.

tomasz-grz (approved):
+1

PR #1492, opened 2025-10-06
LLT-5706: Fix link detection enhanced detection ping

lcruz99:
### Problem
The ip stack is never being set on the wireguard implementation because the [future](https://github.com/NordSecurity/libtelio/pull/1492/files#diff-36327182a60f061fcbee1b4da122b99e82b6d4d213e4b3e94bf297354680a720L1838) is never waited. Because of this, link detection doesn't send any ICMP requests (when the enhanced detection is enabled).

### Solution
Wait for the future. This PR also embeds log improvements on the enhanced detection module.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/link_detection/enhanced_detection.rs
+++ b/crates/telio-wg/src/link_detection/enhanced_detection.rs
@@ -63,23 +63,23 @@ impl Runtime for State {
 
     async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
         if let Some(targets) = self.ping_channel.recv().await {
-            telio_log_debug!("ping channel received trigger");
+            telio_log_debug!("{} received a trigger", Self::NAME);
             Self::guard(async move {
                 let curr_ip_stack = targets.1;
-
-                // We do not want to ping anything, unless we have an IP address assigned
-                if let Some(ips) = curr_ip_stack.as_ref() {
+                // We do not want to ping anything, unless we have an IP address/stack assigned
+                if let Some(ipstack) = curr_ip_stack.as_ref() {
                     for target in targets.0 {
                         let t = match target {
                             IpAddr::V4(v4) => (Some(v4), None),
                             IpAddr::V6(v6) => (None, Some(v6)),
                         };
                         // Skip unsupported combinations
-                        match (ips, target) {
+                        match (ipstack, target) {
                             (IpStack::IPv4, IpAddr::V6(_)) | (IpStack::IPv6, IpAddr::V4(_)) => {
                                 continue
                             }
-                            _ => telio_log_debug!("Unsupported target combination"),
+                            // Supported combo, nothing to do here
+                            _ => (),
                         }
                         if let Ok(t) = DualTarget::new(t) {
                             if let Err(e) = self.pinger.send_ping(&t).await {
@@ -87,9 +87,9 @@ impl Runtime for State {
                             }
                         }
                     }
-                } else {
-                    telio_log_debug!("No targets to ping");
                 }
+                telio_log_debug!("No targets left to ping");
+
                 Ok(())
             })
         } else {


crates/telio-wg/src/link_detection/enhanced_detection.rs
```diff
@@ -63,33 +63,33 @@ impl Runtime for State {
...
                             if let Err(e) = self.pinger.send_ping(&t).await {
                                 telio_log_warn!("Failed to ping peer: {:?}", e);
                             }
                         }
                     }
-                } else {
-                    telio_log_debug!("No targets to ping");
                 }
+                telio_log_debug!("No targets left to ping");
+
```
stalowyjez, line 92: Hmm, why is this log helpful?

crates/telio-wg/src/link_detection/enhanced_detection.rs
```diff
@@ -63,33 +63,33 @@ impl Runtime for State {
 
     async fn wait(&mut self) -> WaitResponse<'_, Self::Err> {
         if let Some(targets) = self.ping_channel.recv().await {
```
stalowyjez, line 65: It seems that we have all the info needed stored in the channel message - maybe we could verify it before sending and just don't send them to pinger if they don't match with the IP stack?

crates/telio-wg/src/link_detection/enhanced_detection.rs
lcruz99: The previous log was not accurate because it was in reality just checking if the peer had any ipstack set. With this change you are acknowledged when the ipstack is not set, the pinging failed or succeded.

crates/telio-wg/src/link_detection/enhanced_detection.rs
lcruz99: I don't think that checking if the node address is compatible with the defined ipstack should be responsability of the LinkDetection module.

stalowyjez:
@lcruz99 I mean, from what I understand, previously the IpStack was set by the lambda in the map when it was not `None`, but when it is `None`, the other function doesn't send any pings anyway, so it seems to me, that some pings, which are unnecessary could be sent previously, but I don't see any additional pings which could be sent, but the PR description implies that they should. It seems to me, that I'm missing sth obvious, but I'd like to understand that before accepting ^^'

src/device.rs
```diff
@@ -1833,14 +1833,11 @@ impl Runtime {
...
         // Update for proxy and derp config
         if let Some(config) = config {
-            let _ = get_ip_stack(config.this.ip_addresses.as_ref().map_or(&[], |v| v))
-                .ok()
-                .map(|ip_stack| async {
-                    self.entities
-                        .wireguard_interface
-                        .set_ip_stack(Some(ip_stack))
-                        .await
-                });
```
lcruz99, line 1843: > @lcruz99 I mean, from what I understand, previously the IpStack was set by the lambda in the map when it was not None, but when it is None, the other function doesn't send any pings anyway, so it seems to me, that some pings, which are unnecessary could be sent previously, but I don't see any additional pings which could be sent, but the PR description implies that they should. It seems to me, that I'm missing sth obvious, but I'd like to understand that before accepting ^^'

@stalowyjez I didn't fully understand your comment so I'll split it by parts.

> previously the IpStack was set by the lambda in the map

It wasn't because the future was never awaited, that was the bug.

> but when it is None, the other function doesn't send any pings anyway, so it seems to me, that some pings, which are unnecessary could be sent previously, but I don't see any additional pings which could be sent

is this a contradiction? 🤔 which unnecessary pings could be sent previously when the ipstack is None? [If the ipstack is None, no pings are sent.](https://github.com/NordSecurity/libtelio/pull/1492/files#diff-c4fe43361a140a64c5b28402d77510922fe05275627627b6efd21fd58cdf8c3dL71)

lcruz99: > but I don't see any additional pings which could be sent, but the PR description implies that they should.

The pings the PR description refer to are triggered from the link_detection mod.

lcruz99:
@jjanowsk 

> but shouldn't this PR remove xfail marks from some tests?

The only xfail marks that exist were referring to other bug that was tackled on https://github.com/NordSecurity/libtelio/pull/1498.

> I understand from jira tickets that this bug caused some instability in them? 

Not really as apps were implementing the ICMP pinging on their side, irrelevating the enhanced detection.

> Also is it possible to prepare a test that fails 100% of times without the fix to show that it fixes it?

On the works..

PR #1493, opened 2025-10-06
Add OpenWrt ip leaks test

asledzinskiy:
- Add test checking that OpenWrt gateway and client always connect through vpn

### Problem
While being connected to the vpn server, OpenWrt gateway and client should always connect to other resources with vpn ip address, not their public ip address.

### Solution

- [x] README.md is updated


olekoliinyk:
+1.0

PR #1498, opened 2025-10-06
Llt 5073 link detection fix wireguardnt

lcruz99:
### Problem

### Solution


crates/telio-network-monitors/src/monitor.rs
```diff
@@ -115,11 +118,23 @@ impl NetworkMonitor {
                                 }
                                 is_paused
                             };
+                            if is_paused {
+                                telio_log_debug!("Received OS interface change notification but the module is paused.");
+                            } else {
+                                telio_log_debug!("Received OS interface change notification");
+                            }
```
tomasz-grz, line 125: Looks like this if statement is used for debugging, and not following any functional code paths?

Maybe instead we could add a log line in `else` of the next `if !is_paused` statement?

Or at least turn this into `telio_log_debug!("Received OS interface change notification. Module is paused: {is_paused}");` what do you think?

crates/telio-relay/src/derp.rs
```diff
@@ -714,9 +714,20 @@ impl Runtime for State {
                         telio_log_info!("Disconnecting from DERP server, due to transmission tasks error");
 
                         self.last_disconnection_reason = match err {
-                            Ok(Ok(())) => RelayConnectionChangeReason::ConfigurationChange,
-                            Ok(Err(err)) => err.into(),
-                            _ => RelayConnectionChangeReason::ClientError,
+                            Ok(Ok(())) => {
+                                let reason = RelayConnectionChangeReason::ConfigurationChange;
+                                telio_log_error!("DERP transmission task Ok: {:?} -> {reason:?}", err);
```
tomasz-grz: Why is this considered an error? Maybe `telio_log_debug`?

crates/telio-wg/src/link_detection.rs
```diff
@@ -12,33 +14,70 @@ use telio_model::{
...
+/// - Network change awareness for improved reliability (windows-only, LLT-5073)
+/// - Configurable RTT thresholds and downgrade behavior
 pub struct LinkDetection {
     cfg_max_allowed_rtt: Duration,
     enhanced_detection: Option<EnhancedDetection>,
     ping_channel: chan::Tx<(Vec<IpAddr>, Option<IpStack>)>,
     peers: HashMap<PublicKey, State>,
     use_for_downgrade: bool,
+    nw_change_observer_rx: Option<chan::Rx<()>>,
+    _observer_handle: Option<Arc<LinkDetectionObserver>>,
```
tomasz-grz: I would personally like to have a comment regarding `_observer_handle`.
Why do we need it if it's not accessed?
Is it to preserve a reference and prevent it from being dropped elsewhere?

crates/telio-wg/src/link_detection.rs
```diff
@@ -12,33 +14,70 @@ use telio_model::{
...
+    pub tx: chan::Tx<()>,
+}
+
+impl LocalInterfacesObserver for LinkDetectionObserver {
+    fn notify(&self) {
+        telio_log_debug!("Notified about local interfaces change");
+        // don't block or panic if the receiver is gone
+        let tx = self.tx.clone();
+        tokio::spawn(async move {
+            let _ = tx.send(()).await;
```
tomasz-grz: In which cases is the receiver gone?

Maybe still worth to log an error if send fails?

crates/telio-network-monitors/src/monitor.rs
lcruz99: IMHO I think it's trivial.. But if you ask me, I think it sums up as having clearer logs because the user/dev doesn't have to read `Module is paused: {is_paused}`, in the other hand using `telio_log_debug!("Received OS interface change notification. Module is paused: {is_paused}");` saves 3 lines of code.

crates/telio-relay/src/derp.rs
lcruz99: For the time being I was considering the previous log: "Disconnecting from DERP server, due to transmission tasks error".

But good question, taking another look also at the other logs within this mod, I would instead change all these logs to debug level. wdyt?

crates/telio-wg/src/link_detection.rs
lcruz99: > Is it to preserve a reference and prevent it from being dropped elsewhere?

exactly, I imagined LinkDetectionObserver docs would suffice. Have you checked them? Nevertheless I can still add it here if you insist

crates/telio-wg/src/link_detection.rs
lcruz99: The receiver is an attribute of LinkDetection just like the observer handle, when one's dropped so the other which should happen only on the LinkDetection destructor/drop. Thus I didn't think the send() check was needed also considering that this is not a critical path.

crates/telio-wg/src/link_detection.rs
```diff
@@ -65,22 +104,66 @@ impl LinkDetection {
...
+                // was indeed disabled.
+                #[cfg(target_os = "windows")]
+                {
+                    if self.enhanced_detection.is_none() {
+                        telio_log_warn!("Link state not reliable for the next few minutes, the interface might have been disabled.");
+                        telio_log_warn!("If a packet is not received in the next 3s the link will be considered down, consider enabling enhanced detection for increased accuracy.");
+                    }
+                    for (public_key, state) in &self.peers {
+                        if let Ok(mut stats) = state.stats.lock() {
+                            stats.first_tx_after_rx = Some(Instant::now());
```
Jauler: This will produce false-negative link state event in case there is no traffic for `10s + max allowed rtt` after network change. 

I think better way would be to take a look at `.is_up()` here: https://github.com/NordSecurity/wireguard-nt-rust-wrapper/blob/libtelio-main/src/adapter.rs#L642

or some other way of forcing traffic to occur (e.g. ICMP ping after network change notificaiton)

Unfortunately having false negatives like this is not really acceptable...

lcruz99: > This will produce false-negative link state event in case there is no traffic for 10s + max allowed rtt after network change.

I've tried to explain that in the comment above 😅 .. btw, it's not 3s but `10s + max allowed rtt + delay (3s)` isn't it?

Shouldn't we expect a packet during that period considering the KEEPALIVE_TIMEOUT and REKEY_TIMEOUT constants?

> I think better way would be to take a look at .is_up() here: https://github.com/NordSecurity/wireguard-nt-rust-wrapper/blob/libtelio-main/src/adapter.rs#L642

The issue is not when the WG adapter is disabled but the interfaces from the following hop (eg: eth/wifi), [we are not tracking them..](https://github.com/NordSecurity/libtelio/pull/1498/files/511e026fa87d878cb5273255c975986d03c26791#diff-cb7633f0bae97d2fd16b373f202ec8f3090ac21c844ad4ad22a6d79f7663e134R134)

> or some other way of forcing traffic to occur (e.g. ICMP ping after network change notificaiton)

That's what I've tried to achieve using the link detection normal functioning by triggering the `first_tx_after_rx` timer. 

Nevertheless, another alternative I've considered was to send ICMP ping by directly triggering its pinger, however we are not allowed to send ICMP pings when enhanced detection is disabled.. 🤔

lcruz99: > Unfortunately having false negatives like this is not really acceptable...

It would be less accurate if we tipped the connection UP when it was in fact down 😅 (false positive)

Jauler: > The issue is not when the WG adapter is disabled but the interfaces from the following hop (eg: eth/wifi), [we are not tracking them..](https://github.com/NordSecurity/libtelio/pull/1498/files/511e026fa87d878cb5273255c975986d03c26791#diff-cb7633f0bae97d2fd16b373f202ec8f3090ac21c844ad4ad22a6d79f7663e134R134)

Oh yeah, you are right, then in this situation `is_up()` is not suitable.

> Shouldn't we expect a packet during that period considering the KEEPALIVE_TIMEOUT and REKEY_TIMEOUT constants?

It is not guaranteed to happen, as KEEPALIVE_TIMEOUT + REKEY_TIMEOUT is 15s (here keepalive timeout refers to passive keepalive, not active), for this reason it is not guaranteed that traffic will flow in this timeframe. But we can solve this by generating our own packet (for example ping) under these circumstances.

Techincally it would also be possible to check whether, according to local OS config (e.g. routing tables) the WG endpoint is reachable. But this is probably overkill option to implement for such a problem.

lcruz99: > But we can solve this by generating our own packet (for example ping) under these circumstances.

Even when systems have the enhanced detection feature disabled?

crates/telio-relay/src/derp.rs
tomasz-grz: What type is `err` here? Shouldn't it be something like `Result<Result<JoinHandle>>`? I would also maybe change the name depending on that.

But I expect an `Ok(Ok())` result to be a successful case and not an error.

In this case we would have the log:
```sh
ERROR: DERP transmission task Ok: Ok(Ok()) -> ConfigurationChange
```

crates/telio-wg/src/link_detection.rs
tomasz-grz: I would add a tiny note so it's easier to understand without jumping around.

crates/telio-wg/src/link_detection.rs
```diff
@@ -65,15 +110,64 @@ impl LinkDetection {
...
+            #[cfg(target_os = "windows")]
+            _observer_handle,
         }
     }
 
+    /// Inserts a new peer.
+    ///
+    /// # Parameters
+    /// - `public_key`: The public key identifying the peer
+    /// - `stats`: Shared statistics for tracking the peer's traffic
```
tomasz-grz: As a non technical user, I would be terrified to see this sentence somewhere in the source code of my VPN app 😅 What do you think of rephrasing to make it sound a little bit less malignant?

lcruz99: fair enough, changed

crates/telio-wg/src/link_detection.rs
lcruz99: added

crates/telio-relay/src/derp.rs
lcruz99: `err` is `Result<Result<JoinError>>`. I'll change the log level to debug, and amend the log message

nat-lab/tests/test_network_monitor.py
```diff
@@ -38,5 +39,8 @@ async def test_network_monitor(
 
         await asyncio.sleep(DEFAULT_WAITING_TIME)
         await client_alpha.restart_interface()
-
-        await client_alpha.wait_for_log("Updating local addr cache")
+        await client_alpha.wait_for_log(
+            "Detected network interface modification, notifying..",
+            count=4,
```
Jauler: Could you clarifiy, why it is <=4 , where does this 4 come from?

crates/telio-wg/src/wg.rs
```diff
@@ -707,6 +696,12 @@ impl State {
             }
         }
 
+        // This is a workaround for WireguardNT (windows): #LLT-5073
```
Jauler: aren't this called on every `update()` cycle, which is happening every second or less?

Jauler: oh sorry, I thought that this will send pings uncoditionally, but it just checks whether there are some network change notifications and only then does necessary state changes.

crates/telio-wg/src/wg.rs
```diff
@@ -707,6 +696,12 @@ impl State {
...
+        #[cfg(target_os = "windows")]
+        if let Some(link_detection) = self.link_detection.as_mut() {
+            link_detection.handle_network_changes().await;
```
Jauler: This does not seem to be related to notifications in any way.  Therefore is'nt it a bit confusing to have it in  `update_send_notifications_events()`? E.g. maybe having it in wg.rs `update()` makes more sense? 

Or even in DynamicWG runtime `wait()`

crates/telio-wg/src/link_detection.rs
```diff
@@ -181,8 +283,12 @@ pub struct LinkDetectionUpdateResult {
 #[derive(Debug)]
 enum StateVariant {
     Down,
-    PossibleDown { deadline: Instant },
+    PossibleDown {
+        deadline: Instant,
+    },
     Up,
+    #[cfg(target_os = "windows")]
+    Unreliable,
```
Jauler: TBH `Unreliable` seems to be very non-descriptive state. E.g. without explicit context on `LLT-5073` there is not clear at all what is the difference between `PossibleDown` and `Unreliable`. 

Maybe calling it `NetworkChanged` would make things more explicit? WDYT?

crates/telio-wg/src/link_detection.rs
```diff
@@ -296,20 +402,31 @@ impl State {
...
-                    Self::build_result(StateDecision::Ping, LinkState::Up)
+                    if ping_enabled {
+                        return Self::build_result(StateDecision::Ping, LinkState::Up);
+                    }
                 } else {
                     // Current link_state is Up
                     telio_log_debug!("definitely up");
-                    Self::build_result(StateDecision::NoAction, LinkState::Up)
                 }
+                Self::build_result(StateDecision::NoAction, LinkState::Up)
```
Jauler, line 345: correct me if I am wrong, but by moving this `Self::build_result(StateDecision::NoAction, LinkState::Up)` out of `else` condition - arent we breaking link detection completely when `ping_enabled` is false?

crates/telio-wg/src/link_detection.rs
```diff
@@ -12,36 +14,80 @@ use telio_model::{
...
+/// accessing the LocalInterfacesObserver::notify() implementation.
+#[cfg(target_os = "windows")]
+#[derive(Debug)]
+pub struct LinkDetectionObserver {
+    /// Channel transmitter for sending network change notifications
+    pub tx: chan::Tx<()>,
+}
+
+#[cfg(target_os = "windows")]
+impl LocalInterfacesObserver for LinkDetectionObserver {
```
Jauler: Do we need separate struct for `LinkDetectionObserver`? 

Maybe we can implement `LocalInterfaceObserver` trait for `LinkDetection` struct itself, to remove the need for additional tokio-tasks being spawned, channels being used etc.?

crates/telio-wg/src/link_detection.rs
Jauler: yes

crates/telio-wg/src/link_detection.rs
Jauler, line 208: Could you update this state diagram? 🙏

Jauler (commented):
I know this goes into the territory of "scope creep" for the LLT-5073, but I am looking at the link-detection code and can't help myself, but see opportunity to simplify it significantly by removing "enhanced detection". 

Now that we learned (or realized that previously we missed) that pretty much all WG implementations we are using will issue handshake after missed passive keepalive, "enhanced detection" is somewhat unnecessary. And it already complicates code for this PR.

What  do you think about removing it in this PR?

nat-lab/tests/test_network_monitor.py
lcruz99: Good catch, it's 1 for startup + 1 for first configuration + 1 for [deconfigure_interface()](https://github.com/NordSecurity/libtelio/blob/c2694f5c8add9f0bc952f75bc131b88d69f22bf0/nat-lab/tests/utils/router/linux_router.py#L68) + 1 for [setup_interface()](https://github.com/NordSecurity/libtelio/blob/c2694f5c8add9f0bc952f75bc131b88d69f22bf0/nat-lab/tests/utils/router/linux_router.py#L48).

I'll document and use a constant instead.

crates/telio-wg/src/link_detection.rs
lcruz99: I completely agree, will change

crates/telio-wg/src/link_detection.rs
lcruz99: If `ping_enabled` is false the `self.variant` is set to `PossibleDown` and the `Self::build_result(StateDecision::NoAction, LinkState::Up)` is returned. LinkDetection then proceeds as usual the next iteration to the `PossibleDown` match branch.

crates/telio-wg/src/link_detection.rs
lcruz99: I've preferred to decouple this to LinkDetectionObserver over having to make LinkDetection Send + Sync:

My first idea was what you've suggested, but that meant sharing the LinkDetection object with both the [NetworkMonitor](https://github.com/NordSecurity/libtelio/blob/LLT-5073_link_detection_fix_wireguardnt/src/device.rs#L1088) and the [WG implementation](https://github.com/NordSecurity/libtelio/blob/LLT-5073_link_detection_fix_wireguardnt/src/device.rs#L1161). WG requires interior mutability on the LinkDetection, so lock() will be required for every LinkDetection method called that modifies it.

lcruz99: If we opt for that road, we still have to make the LinkDetection access on LocalInterfacesObserver::notify() Sync  or else block the monitor task until we have the lock guard for LinkDetection.

crates/telio-wg/src/wg.rs
lcruz99: agreed, changed

crates/telio-wg/src/link_detection.rs
lcruz99: Updated diagram description

crates/telio-wg/src/link_detection.rs
lcruz99: See https://github.com/NordSecurity/libtelio/pull/1498/commits/43f908451d21b76e72663f57c75558e325ac92e5

lcruz99: Ignore ^^, moved Observer to windows specific mod: https://github.com/NordSecurity/libtelio/pull/1498/files#diff-52f258cd141889ef5c7680372166358202141f58bc81878713548e8205899437R25

crates/telio-wg/src/link_detection.rs
```diff
@@ -94,18 +143,14 @@ impl LinkDetection {
...
-            telio_log_debug!(
-                "ping_enabled: {}, result.should_ping: {}. result.link_detection_update_result: {:?}",
-                ping_enabled,
-                result.should_ping,
-                result.link_detection_update_result
-            );
-            if ping_enabled && result.should_ping {
+            let result = state.update(self.cfg_max_allowed_rtt);
+
+            telio_log_debug!("{:?}", result.link_detection_update_result);
```
tomasz-grz: Leftover from development?
Could you remove it or add some context to the debug log if you would like to keep it?

crates/telio-wg/src/link_detection.rs
```diff
@@ -94,18 +143,14 @@ impl LinkDetection {
...
+            #[cfg(target_os = "windows")]
```
tomasz-grz, line 152: Not exactly fond of all the ` #[cfg(target_os = "windows")]` that this PR is introducing.

Is there a better way to handle this? Maybe a single module with no no-op implementation for all the other platforms? What do you think?

A bit how the `telio-sockets/src/protector.rs` module is implemented.

crates/telio-wg/src/link_detection.rs
```diff
@@ -174,19 +230,27 @@ pub struct LinkDetectionUpdateResult {
...
 #[derive(Debug)]
 enum StateVariant {
     Down,
-    PossibleDown { deadline: Instant },
+    PossibleDown {
+        deadline: Instant,
+    },
     Up,
+    #[cfg(target_os = "windows")]
+    NetworkChanged,
```
tomasz-grz, line 235: Why is this state variant only for Windows?

tomasz-grz: Also please add the state to the diagram/comments

crates/telio-wg/src/windows/enhanced_link_detection.rs
```diff
@@ -1,36 +1,63 @@
+#![cfg(windows)]
```
tomasz-grz: So if I understand correctly Enhanced Link Detection becomes a Windows only feature? Why, could you please update the PR description?

.unreleased/LLT-5073
Jauler, line 1: please 🙏  add some changelog entry. This functionality is definitely relevant for the integrators.

crates/telio-wg/src/link_detection.rs
Jauler: 🤔 but state diagram as well as description is still missing "NetworkChanged" state you added 🤔.

crates/telio-wg/src/link_detection.rs
Jauler: oh, ok.

.unreleased/LLT-5073
lcruz99: done

crates/telio-wg/src/windows/enhanced_link_detection.rs
lcruz99: added

crates/telio-wg/src/link_detection.rs
lcruz99: added

crates/telio-wg/src/link_detection.rs
lcruz99: IIUC you mean to abstract and then provide concrete LinkDetection mods for the different targets?

I agree it's a lot of macros, not pretty, but I can't think of a better alternative. Abstracting LinkDetection IMHO is not worth the effort and will create a lot of repeated code.

crates/telio-wg/src/link_detection.rs
lcruz99: It is intentional, the context is given by the previous logs (just like the one in push_update()). I'll add the public key but it can be too verbose.

crates/telio-wg/src/link_detection.rs
```diff
@@ -1,79 +1,128 @@
...
     uapi::UpdateReason,
     wg::{BytesAndTimestamps, WG_KEEPALIVE},
 };
 
-mod enhanced_detection;
-
+/// Component that manages the link state of each peer.
+///
+/// `LinkDetection` monitors peer connectivity by analyzing traffic patterns and when enhanced
+/// detection is enabled performing active probing through ICMP ping to determine if a peer link is up or down.
```
Jauler:
```suggestion
/// `LinkDetection` monitors peer connectivity by analyzing traffic patterns
```

crates/telio-wg/src/link_detection.rs
Jauler: IMHO it is not a big deal with all those macros - they are compile-time attributes, hence little to no chance of various prod issues.

But while this kind of link detection upgrade is requested by windows - IMHO it is useful generally. Hence we may just remove all the windows conditional compilation and make this "pinging" upon network change a general concept for all platforms.

Just that IMHO it is probably better to do that not immediately, but after some proof that it works as intended, e.g. by spinning some time in prod on one platform.

nat-lab/tests/test_events_link_state.py
```diff
@@ -546,250 +472,10 @@ async def test_event_link_state_peer_doesnt_respond(
...
-                        ConnectionTag.DOCKER_CONE_CLIENT_2,
-                        TelioAdapterType.NEP_TUN,
-                    ),
-                ],
-                enhanced_detection=False,
-            ),
-        ),
-    ],
-)
-async def test_event_link_state_without_enhanced_detection(
```
Jauler, line 655: Is this test removal intentional? It seems to be testing standard link detection.  Not the one with the pings.

Jauler (commented):
nice work! This definately looks now much nicer :).

But I have a quick question - do we have any kind of automated test covering the scenario? I kind of did not find one, but maybe I missed something?

crates/telio-wg/src/link_detection.rs
lcruz99: done

nat-lab/tests/test_events_link_state.py
lcruz99: Yes, this test by itself is avoiding all expected outgoing traffic from beta to alpha so that beta's link state on alpha node goes down. I think it doesn't test anything more than other tests such:
```
test_event_link_detection_after_disabling_ethernet_adapter() (variants)
test_event_link_state_delayed_packet()
```

lcruz99:
> But I have a quick question - do we have any kind of automated test covering the scenario? I kind of did not find one, but maybe I missed something?

@Jauler The test_event_link_detection_after_disabling_ethernet_adapter(), the LLT-5073 bug was reproducible in its variants:
```
test_event_link_detection_after_disabling_ethernet_adapter_direct_path()
test_event_link_detection_after_disabling_ethernet_adapter_with_vpn()
```

Added here: https://github.com/NordSecurity/libtelio/pull/1240

nat-lab/tests/test_events_link_state.py
Jauler: oh yeah, it does seem to overlap with other test. Note really the `*_disabling_ethernet_*` variants, as those are different scenarios, but delayed packets or offline peers is something overlapping.

Jauler (dismissed):
+1.0

crates/telio-wg/src/link_detection.rs
tomasz-grz: I see, thanks

tomasz-grz: NIT: If it's too verbose, should we make some of the additional logs as `trace`?

crates/telio-wg/src/link_detection.rs
tomasz-grz: Fair point

crates/telio-wg/src/windows/enhanced_link_detection.rs
tomasz-grz: Thank you

tomasz-grz (approved):
+1

PR #1499, opened 2025-10-07
Use external source for the `ens.proto`

tomaszklak:
### Problem
The `ens.proto` file definitions are use both on the client and server side. Currently both sides have a copy of the file, which leads to duplication and can lead to (it actually already happened, but was caught by manual testing on staging) discrepancies.

### Solution
Move the `ens.proto` file to an external repo that can be referenced by both client and server sides to get the official version. To avoid having to deal with yet another git submodule, the proto is distributed as a rust crate. Since we need the definition both for building rust library and testing with python, the python side now picks the file from the `dist` directory where other build artifacts are already stored.


crates/telio-proto/Cargo.toml
```diff
@@ -12,6 +12,7 @@ enable_ens = ["rustls/aws_lc_rs", "dep:aws-lc-rs", "dep:rustls-webpki"]
 [dependencies]
 http = "1.3.1"
 hyper-util = "0.1.15"
+llt-proto = { git = "https://github.com/NordSecurity/llt-proto.git", rev = "de5fdab93d1f64fbea3c5e59aebf236714e98a5d" }
```
tomaszklak: 🚨 TODO: update with a released tag 🚨

src/device.rs
```diff
@@ -2721,6 +2721,27 @@ fn get_config_peer<'a>(config: Option<&'a Config>, public_key: &'a PublicKey) ->
         .find(|p| p.base.public_key == *public_key)
 }
 
+fn convert_connection_error(
```
stalowyjez, line 2724: Wouldn't be it a bit more idiomatic to add that `From<ConnectionError>` impl to `telio_model::mesh::VpnConnectionError`?

nat-lab/natlab.py
```diff
@@ -163,6 +164,12 @@ def get_docker_version():
 def generate_grpc(path):
     os.makedirs("bin/grpc_protobuf", exist_ok=True)
     include = os.path.dirname(path)
+    if not os.path.exists(path):
+        print(
+            f"You need to have {path} generated. You can do it by building libtelio, for example with:"
```
stalowyjez, line 199: Hmm, couldn't we just generate one in such case?

src/device.rs
tomaszklak: You are correct, it would be more idiomatic and I did consider it. The problem is that for it to be possible the telio-model would have to also depend on the llt-proto which I think is worse than having this conversion function.

nat-lab/natlab.py
tomaszklak: I see two problems with that:
- `natlab.py` starting compilation of libtelio would be very surprising
- we don't know how to build it (debug or release)

Which is why I think that this way is better, since most of the time this issue will not come up

src/device.rs
stalowyjez: I guess if VpnConnectionError is kinda linked to that error, maybe such dependency makes sense? IDK

PR #1502, opened 2025-10-08
Fix double-free tun fd

LukasPukenis:
Tunnel FD is stored as OwnedFd which has its own
drop() in which it does explicit fd close. An explicit close is thus unecessary, we suspect it might be causing rare crashes on Android.


crates/telio-wg/src/wg.rs
```diff
@@ -1030,10 +1030,6 @@ impl Runtime for State {
         if let Some(link_detection) = self.link_detection {
             link_detection.stop().await;
         }
-        #[cfg(unix)]
-        use std::os::fd::AsRawFd;
-        #[cfg(unix)]
-        self.cfg.tun.map(|tun| nix::unistd::close(tun.as_raw_fd()));
```
tomasz-grz, line 1036: When would the `fd` be closed though then? When `WireGuard` adapter is dropped or next `set_tun()` is called?

Or do we need to set here?
```rust
self.cfg.tun = None;
```

LukasPukenis: Very good question, I was thinking about that myself but came to realize this: here is device.rs::stop() implementation:

```
pub fn stop(&mut self) {
        if let Some(rt) = self.rt.take() {
            if let Some(art) = &self.async_runtime {
                let _ = art.block_on(rt.stop());
                self.flush_events();
            }
        }
    }
```

which eventually triggers the code in this PR code via `rt.stop()`. At the end of this function, all the task destructors are executed, meaning it includes the `cfg` and in turn `cfg.tun` itself.

Making the `cfg.tun = None` explicitly wouldn't hurt really but also isn't particularly necessary IMHO.

tomasz-grz: Ahh I see, thanks for clarifying!

tomasz-grz (commented):
+1 from my side

crates/telio-wg/src/wg.rs
tomaszklak, line 1036: Do you think it would be possible to create a test that catches this problem? I suspect that introducing 'fd sanitizer' could be costly but maybe we could (only in tests) use custom wrapper around file descriptor / `OwnedFd` which would panic on double close?

LukasPukenis: I realized that in order to properly sanitize in tests I would need to intercept `libc::close` and I had no idea how to do that easily. For example we might pass the file descriptor to some crate(maybe we even do) so I gave up on that idea. For `drop()` being correct I would trust Rust and `OwnedFd` implementation. That was my initial take :)

However, now looking again into it I have one idea - use `LD_PRELOAD` and check if `close` tries to close the same FD but that would also mean we would need to book-keep the file descriptors opened by our process(since we might open(3), close(3), open(3), close(3)). I do not see that being done easily or quickly :).

Maybe you have some specific implementation in your mind that's easy to do in this case?

tomaszklak: Yeah, a wrapper type will be not enough and the LD_PRELOAD idea is a good one. Although I feel it's too big and complex for this PR. If you also think it would be useful, please create a separate ticket for such test(s).

LukasPukenis: The fix looks too big for this PR as you say so I created LLT-6726.

PR #1568, opened 2025-11-07
[LLT-6804] - Replace token placeholder in default nordvpnlite config

olekoliinyk:
### Problem
There is inconsistency in the token placeholder between the example config and nordvpnlite service config. 

### Solution
Use 'aaa' format of placeholder in nordvpnlite service config

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/config.json
+++ b/clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/config.json
@@ -1,5 +1,5 @@
 {
-  "authentication_token": "<REPLACE_WITH_YOUR_TOKEN>",
+  "authentication_token": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
   "vpn": "recommended",
   "log_level": "error",
   "log_file_path": "/var/log/nordvpnlite.log",


lcruz99 (approved):
+2

Jauler (approved):
+2.0

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "authentication_token": "<REPLACE_WITH_YOUR_TOKEN>",
```
tomasz-grz, line 2: The token in the default confing should be explicitly something like`"PASTE_YOUR_AUTHENTICATION_TOKEN_HERE"`.
This will fail to deserialize on start and an [error message will appear](https://github.com/NordSecurity/libtelio/blob/da7e673298fe01efbc4c529d0c2692daa7ce1397/clis/nordvpnlite/src/config.rs#L88), to help prompt the user to provide an authentication token.

If you replace it with a deserializable token like `"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"`, the only way to tell that the user has not added a valid token is after making the API calls, which can only be viewed in the logs, and the user will have no idea that something is wrong.

tomasz-grz (changes requested):

olekoliinyk:
Closing as @tomasz-grz raised a valid point.

PR #1676, opened 2026-02-06
Wireguard-NT orphaned devices cleanup

matislovas:
### Problem
Wireguard-NT `create()` and `open()` methods are launching the orphaned devices cleaning [routine](https://github.com/WireGuard/wireguard-nt/blob/50086f0f88a2afdc5d9862064f599938b99dc712/api/adapter.c#L534), but this is launched in separate thread from the one, that is creating device. So there will be a race condition on adapter creation and cleanup.

### Solution
We're moving the cleanup code to `libtelio` side and initiating cleanup before creation.

P.S. the tests will be added in the other PR.

- [x] README.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -169,7 +169,7 @@ uniffi = "0.28.3"
 url = "2.5"
 uuid = { version = "1.19", features = ["v4"] }
 wireguard-uapi = { git = "https://github.com/NordSecurity/wireguard-uapi-rs.git", rev = "5f5d9952ced4faf10f56d98153e4a8e560102e5c", features = ["xplatform", "zeroize"], default-features = false  }
-winapi = { version = "0.3", features = ["netioapi", "ws2def"] }
+winapi = { version = "0.3", features = ["netioapi", "ws2def", "cfgmgr32"] }
 windows = { version = "0.61", features = [
   "Win32_Networking_WinSock",
   "Win32_NetworkManagement_IpHelper",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -169,10 +169,11 @@ uniffi = "0.28.3"
 url = "2.5"
 uuid = { version = "1.19", features = ["v4"] }
 wireguard-uapi = { git = "https://github.com/NordSecurity/wireguard-uapi-rs.git", rev = "5f5d9952ced4faf10f56d98153e4a8e560102e5c", features = ["xplatform", "zeroize"], default-features = false  }
-winapi = { version = "0.3", features = ["netioapi", "ws2def", "cfgmgr32"] }
+winapi = { version = "0.3", features = ["netioapi", "ws2def", "cfgmgr32", "setupapi", "winreg", "iphlpapi", "errhandlingapi", "ifdef"] }
 windows = { version = "0.61", features = [
   "Win32_Networking_WinSock",
   "Win32_NetworkManagement_IpHelper",
+  "Win32_NetworkManagement_Ndis",
 ] }
 winreg = "0.50.0"
 
--- a/crates/telio-wg/src/adapter/windows_native_wg.rs
+++ b/crates/telio-wg/src/adapter/windows_native_wg.rs
@@ -1,51 +1,46 @@
 use super::{Adapter, Error as AdapterError, Tun as NativeTun};
-use crate::uapi::{Cmd, Cmd::Get, Cmd::Set, Interface, Peer, Response};
-#[cfg(windows)]
-use crate::windows::service;
-#[cfg(windows)]
-use crate::windows::tunnel::interfacewatcher::InterfaceWatcher;
+use crate::{
+    uapi::{Cmd, Cmd::Get, Cmd::Set, Interface, Peer, Response},
+    windows::{service, tunnel::interfacewatcher::InterfaceWatcher},
+};
 use async_trait::async_trait;
-#[cfg(windows)]
 use sha2::{Digest, Sha256};
-use std::io::Error as IOError;
-use std::slice::Windows;
-use std::sync::{Arc, Mutex, MutexGuard};
-use std::time::Duration;
-use std::{mem, option, result};
+use std::{
+    collections::HashMap,
+    io::Error as IOError,
+    mem, option,
+    ptr::{self, null},
+    result,
+    slice::Windows,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
+
 use telio_utils::{
     telio_log_debug, telio_log_error, telio_log_info, telio_log_trace, telio_log_warn,
 };
 use tokio::time::sleep;
-#[cfg(windows)]
+use utf16_lit::utf16_null;
 use uuid::Uuid;
-#[cfg(windows)]
-use winapi::um::cfgmgr32::{CM_Get_DevNode_Status, CR_SUCCESS, DN_HAS_PROBLEM};
-#[cfg(windows)]
-use windows::{
-    core::GUID,
-    Win32::{
-        Devices::{
-            DeviceAndDriverInstallation::{
-                SetupDiCallClassInstaller, SetupDiDestroyDeviceInfoList, SetupDiEnumDeviceInfo,
-                SetupDiGetClassDevsExW, SetupDiGetDevicePropertyW, SetupDiSetClassInstallParamsW,
-                GUID_DEVCLASS_NET, HDEVINFO, SP_CLASSINSTALL_HEADER, SP_DEVINFO_DATA,
-                SP_REMOVEDEVICE_PARAMS,
-            },
-            Properties::{DEVPROPID_FIRST_USABLE, DEVPROPKEY, DEVPROP_TYPE_STRING},
-        },
-        Foundation::{ERROR_NO_MORE_ITEMS, HANDLE},
-    },
+use windows::core::GUID;
+use windows::core::PCWSTR;
+use windows::Win32::Devices::DeviceAndDriverInstallation::GUID_DEVCLASS_NET;
+use windows::Win32::Devices::DeviceAndDriverInstallation::{
+    CM_Get_DevNode_Status, SetupDiCallClassInstaller, SetupDiEnumDeviceInfo,
+    SetupDiGetDeviceRegistryPropertyW, SetupDiSetClassInstallParamsW, CM_DEVNODE_STATUS_FLAGS,
+    CM_PROB, CR_SUCCESS, DIF_REMOVE, DI_REMOVEDEVICE_GLOBAL, DN_HAS_PROBLEM, SPDRP_FRIENDLYNAME,
+    SP_CLASSINSTALL_HEADER, SP_REMOVEDEVICE_PARAMS,
 };
-#[cfg(windows)]
+use windows::Win32::NetworkManagement::Ndis::GUID_DEVINTERFACE_NET;
 use winreg::{enums::*, RegKey, HKEY};
-#[cfg(windows)]
 use wireguard_nt::{
     self, set_logger, Error as WireGuardNTError, SetInterface, SetPeer, WIREGUARD_STATE_UP,
 };
 use wireguard_uapi::xplatform;
 
+const REMOVAL_SLEEP_SECS: u64 = 2;
+
 /// Telio wrapper around wireguard-nt
-#[cfg(windows)]
 pub struct WindowsNativeWg {
     // Object holding the adapter handle and associated wireguard functionality
     adapter: Arc<wireguard_nt::Adapter>,
@@ -63,7 +58,6 @@ pub struct WindowsNativeWg {
 }
 
 /// Error type implementation for wg-nt
-/// #[cfg(any(windows, doc))]
 #[cfg_attr(docsrs, doc(cfg(windows)))]
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -78,7 +72,6 @@ enum AdapterState {
 }
 
 // Windows native associated functions
-#[cfg(windows)]
 impl WindowsNativeWg {
     pub fn new(
         adapter: &Arc<wireguard_nt::Adapter>,
@@ -118,7 +111,7 @@ impl WindowsNativeWg {
         name: &str,
         path: &str,
         enable_dynamic_wg_nt_control: bool,
-    ) -> Result<Self, AdapterError> {
+    ) -> std::result::Result<Self, AdapterError> {
         // try to load dll
         match unsafe { wireguard_nt::load_from_path(path) } {
             Ok(wg_dll) => {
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -207,7 +208,7 @@ telio-pinger = { version = "0.1.0", path = "./crates/telio-pinger" }
 
 [profile.release]
 opt-level = "s"
-lto = true
+# lto = true
```
lcruz99: leftover?

matislovas: oh ... aha :smile: fixed!

gytsto (dismissed):
works good +1

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -166,6 +182,144 @@ impl WindowsNativeWg {
...
+                index += 1;
+            }
+
+            let _ = SetupDiDestroyDeviceInfoList(dev_info);
+
+            if removed_any {
+                telio_log_info!(
+                    "Orphaned WireGuard-NT adapters were removed; sleeping 2 seconds for system to settle"
+                );
+                sleep(Duration::from_secs(2)).await;
```
stalowyjez: Do we really need this timeout? Maybe there is some better way to ensure that system is in the desired state?

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+/// * `device_instance_id` - Device instance ID (e.g. `SWD\\WireGuard\\...`)
+/// * `verify_in_ip_stack` - If true, also require the interface to appear in
+///   GetIfEntry2 (IP stack). Slightly stricter/later.
+///
+/// # Returns
+/// * `Ok(())` if the adapter interface is created and registered.
+/// * `Err(win32_error)` on failure (e.g. ERROR_DEVICE_NOT_READY).
+fn is_network_adapter_interface_ready(
+    adapter_luid: u64,
+    verify_in_ip_stack: bool,
```
stalowyjez: Do we need this flag when we always verify it anyway?

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+            dev_info,
+            device_instance_id.as_ptr(),
+            ptr::null_mut(),
+            0,
+            &mut dev_info_data,
+        )
+    };
+    if open_ok == 0 {
+        let err = unsafe { GetLastError() };
+        unsafe { SetupDiDestroyDeviceInfoList(dev_info) };
```
stalowyjez: Hmm, could we have some drop guard (i.e. raii wrapper) around that?

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+        return Err(AdapterInterfaceReadyError::Other(err));
+    }
+
+    // 1. Network driver must be initialized.
+    is_network_driver_initialized()?;
```
stalowyjez, line 458: I guess here you lost `SetupDiDestroyDeviceInfoList`? (this is why I think some guard around that would be better)

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+    if unsafe {
+        RegQueryValueExW(
+            drv_key,
+            if_type_name.as_ptr(),
+            ptr::null_mut(),
+            ptr::null_mut(),
+            &mut if_type as *mut u32 as *mut u8,
+            &mut val_size,
+        )
+    } != 0
```
stalowyjez: I think it would be a bit cleaner to extract these single checks to a separate functions

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+    {
+        unsafe {
+            RegCloseKey(drv_key);
+            SetupDiDestroyDeviceInfoList(dev_info);
+        }
+        return Err(AdapterInterfaceReadyError::IfTypeNotSet);
+    }
+    telio_log_debug!("IfType set");
+
+    unsafe { RegCloseKey(drv_key) };
```
stalowyjez: This could be some guard as well. Maybe this whole block could be also a separate function? IDK

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+    }
+    Err(AdapterInterfaceReadyError::DeviceInstanceIdNotFound)
+}
+
+fn is_network_driver_initialized() -> Result<(), AdapterInterfaceReadyError> {
+    unsafe {
+        let enumerator_utf16: Vec<u16> = WIREGUARD_ENUMERATOR
+            .encode_utf16()
+            .chain(std::iter::once(0))
+            .collect();
```
stalowyjez: I think I saw it more than once and it looks like sth which could be const - also there is some crate with ready to use
```
const EXAMPLE_NULL: &[u16] = &utf16_null!("example");
```

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+            PCWSTR(enumerator_utf16.as_ptr()),
+            None,
+            DIGCF_PRESENT,
+            Some(HDEVINFO::default()),
+            PCWSTR(ptr::null()),
+            None,
+        ) {
+            Ok(info) => info,
+            Err(_) => return Err(AdapterInterfaceReadyError::DriverNotInitialized),
+        };
```
stalowyjez: `map_err`?

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+            {
+                (status & (DN_DRIVER_LOADED.0)) != 0 && (status & (DN_STARTED.0)) != 0
+            } else {
+                return Err(AdapterInterfaceReadyError::DriverNotInitialized);
+            }
+        } else {
+            return Err(AdapterInterfaceReadyError::DriverNotInitialized);
+        };
+
+        let _ = winrs_SetupDiDestroyDeviceInfoList(dev_info);
```
stalowyjez: Hmm, maybe some comment why you can just skip this result would be helpful

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: I am sure there is, but honestly I haven't found :smile:

crates/telio-wg/src/windows/service.rs
matislovas: Actually, it returns bool, if the call succeeded or not (as in all other cases in this file). So I'll omit `let _ = ` part only.

crates/telio-wg/src/windows/service.rs
matislovas: It is inside `is_network_driver_initialized` function or am I missing sth there?

crates/telio-wg/src/windows/service.rs
matislovas: I am not very sure about this check, so maybe we should keep it, if we will get failure in checking the `IPstack`?

crates/telio-wg/src/windows/service.rs
stalowyjez: @matislovas I think with this `?` you miss the destruction which you did e.g. on line 395

crates/telio-wg/src/windows/service.rs
matislovas: Addressed

crates/telio-wg/src/windows/service.rs
matislovas: Oh yeah, addressed

crates/telio-wg/src/windows/service.rs
matislovas: Done

crates/telio-wg/src/windows/service.rs
matislovas: Done

crates/telio-wg/src/windows/service.rs
matislovas: Moved to separate function

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -11,14 +11,30 @@ use std::io::Error as IOError;
...
+};
 use telio_utils::{
     telio_log_debug, telio_log_error, telio_log_info, telio_log_trace, telio_log_warn,
 };
 use tokio::time::sleep;
 #[cfg(windows)]
 use uuid::Uuid;
 #[cfg(windows)]
+use winapi::um::cfgmgr32::{CM_Get_DevNode_Status, CR_SUCCESS};
+#[cfg(windows)]
```
stalowyjez: I wonder why we need this windows configs here - shouldn't be the whole module removed when the build does not target windows?

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +163,386 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+        SetupDiOpenDeviceInfoW(
+            dev_info,
+            device_instance_id.as_ptr(),
+            ptr::null_mut(),
+            0,
+            &mut dev_info_data,
+        )
+    };
+    if open_ok == 0 {
+        unsafe { SetupDiDestroyDeviceInfoList(dev_info) };
```
stalowyjez: Hmm, shouldn't we use RAII also here?

crates/telio-wg/src/windows/service.rs
stalowyjez: We can just remove it in such case

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -166,6 +182,144 @@ impl WindowsNativeWg {
...
+    /// WireGuard-NT's `api/adapter.c`. It enumerates all `GUID_DEVCLASS_NET`
+    /// devices created by the WireGuard software enumerator and removes any
+    /// whose devnode has `DN_HAS_PROBLEM` set.
+    #[cfg(windows)]
+    async fn cleanup_orphaned_devices() {
+        unsafe {
+            let mut removed_any = false;
+
+            // Get list of all WireGuard network adapters.
+            let enumerator_utf16: Vec<u16> = service::WIREGUARD_ENUMERATOR
```
stalowyjez: It seems it should be doable in compile time.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -166,6 +182,144 @@ impl WindowsNativeWg {
...
+                .encode_utf16()
+                .chain(std::iter::once(0))
+                .collect();
+
+            let dev_info: HDEVINFO = match SetupDiGetClassDevsExW(
```
stalowyjez: Soo, you have this `ScopedDeviceInfoSet` already - why won't use it here as well?

crates/telio-wg/src/windows/service.rs
```diff
@@ -1,18 +1,193 @@
...
+        if (dev_info as isize) == INVALID_HANDLE_VALUE {
+            return Err(AdapterInterfaceReadyError::DriverNotInitialized);
+        }
+
+        // Enumerate the first device and cache its SP_DEVINFO_DATA so callers
+        // can use the DevInst without having to enumerate again.
+        let mut dev_info_data: SP_DEVINFO_DATA = unsafe { std::mem::zeroed() };
+        dev_info_data.cbSize = std::mem::size_of::<SP_DEVINFO_DATA>() as DWORD;
+
+        let ok = unsafe { SetupDiEnumDeviceInfo(dev_info, 0, &mut dev_info_data) };
```
stalowyjez: That could be created with `ScopedDeviceInfoSet` (maybe renamed) instance method taking an index, which would make it simpler to use in the loop in windows_native_wg.rs, probably

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -166,6 +182,144 @@ impl WindowsNativeWg {
...
+                    == CR_SUCCESS
+                    && (status & (DN_HAS_PROBLEM.0)) == 0
+                {
+                    index += 1;
+                    continue;
+                }
+
+                // Best-effort: fetch device friendly name for logging.
+                const FRIENDLY_NAME_BUF_WCHARS: usize = 260;
+                let mut name_buf = [0u8; FRIENDLY_NAME_BUF_WCHARS * 2];
```
stalowyjez: Nit: This buffer could be defined before the loop and then reused, probably

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -166,6 +182,144 @@ impl WindowsNativeWg {
...
+                            .map(|s| String::from_utf16_lossy(s).trim().to_string())
+                            .unwrap_or_default();
+                        if s.is_empty() {
+                            format!("devinst {}", dev_info_data.DevInst)
+                        } else {
+                            s
+                        }
+                    }
+                    Err(_) => format!("devinst {}", dev_info_data.DevInst),
+                };
```
stalowyjez, line 269: Nit: Maybe we can also have some `get_device_frtiendly_name` function, or sth

crates/telio-wg/src/adapter/windows_native_wg.rs
stalowyjez: Maybe you could have an outer loop which runs while there are none orphaned nodes? I guess a run in which all devices are in the correct state would be quite convincing.

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +297,290 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+    }
+}
+
+pub fn str_to_utf16(s: &str) -> Vec<u16> {
+    s.encode_utf16().chain(std::iter::once(0)).collect()
+}
+
+fn get_device_instance_id_for_adapter_luid(
+    adapter_luid: u64,
+) -> Result<Vec<u16>, AdapterInterfaceReadyError> {
```
stalowyjez, line 369: Maybe it could return the scoped info when it matches the luid?

crates/telio-wg/src/windows/service.rs
```diff
@@ -88,3 +297,290 @@ fn is_service_running(service_name: &str) -> WindowsResult<WinServiceStatus> {
...
+    };
+    if cr != CR_SUCCESS {
+        return Err(AdapterInterfaceReadyError::Other(cm_map_cr_to_win32_err(
+            cr,
+            ERROR_GEN_FAILURE,
+        )));
+    }
+    if interface_list.first().copied() == Some(0) {
+        return Err(AdapterInterfaceReadyError::NDISDeviceInterfaceNotExists);
+    }
```
stalowyjez, line 547: Hmm, so `0` means that this is the list end? But then what is the meaning of the size? Maybe it actually matters and we could just check `list_len > 0`, wdyt?

stalowyjez (commented):
Looks much better, I think, just a few minor comments

crates/telio-wg/src/windows/service.rs
matislovas: Y? `device_instance_id` is used further down in `CM_Get_Device_Interface_ListW` function calls.

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: But this is used only in one place, so why bother? Comment above should be enough

crates/telio-wg/src/windows/service.rs
matislovas: But `list_len` is only the input of the size, it does not output, how the buffer was filled https://learn.microsoft.com/en-us/windows/win32/api/cfgmgr32/nf-cfgmgr32-cm_get_device_interface_listw

matislovas: And we're actually checking it with `CM_Get_Device_Interface_List_SizeW` before, and if it returns `list_len < 1`, it return `Err`.

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: done

crates/telio-wg/src/windows/service.rs
matislovas: done

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: done

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: done

crates/telio-wg/src/windows/service.rs
matislovas: done

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: done

crates/telio-wg/src/adapter/windows_native_wg.rs
stalowyjez: Hmm, my intuition is that when you have such a self-contained part of a long function it just makes the code easier to read if it is extracted and have some nice name, i.e. you don't need to search where it ends when you are not interested in it, but just see that's a function, in this case, getting friendly name, and easily skip it. But that's a nit anyway, co you can just resolve it

stalowyjez (approved):

PR #1709, opened 2026-03-12
LLT-7053: Raw DNS forwarder

tomasz-grz:
### Problem

Non `.nord` DNS queries are forwarded through hickory-server's ForwardAuthority zone. We want to remove  `hickory` dependencies as it adds unnecessary overhead, requires additional maintenance and was a source of bugs.

### Solution

Add a `RawForwarder`, that sends DNS queries directly to upstream resolvers as UDP packets.
The forwarder will be integrated in a subsequent PR

- DNS message use internal ID rewriting to support multiple concurrent queries
- each upstream is tried in order, with per query timeouts
- single socket bound to the tunnel interface is reused
- raw bytes are forwarded and returned unchanged

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/src/forwarder.rs
+++ b/crates/telio-dns/src/forwarder.rs
@@ -1,0 +1,920 @@
+//! DNS packet forwarder for remote resolvers
+//!
+//! This module provides a layer for forwarding DNS queries to upstream resolvers.
+//! Multiple queries can be in flight concurrently.
+//!
+//! Upstream resolvers are tried in order they are set.
+//! If no response is received after the set timeout, the next
+//! upstream resolver is tried.
+
+use crate::{bind_tun::bind_to_tun, packet_encoder::DNS_HEADER_OFFSET};
+use rand::RngExt;
+use std::{collections::HashMap, io, net::SocketAddr, sync::Arc, time::Duration};
+use telio_utils::{sleep_until, telio_log_debug, telio_log_warn, Instant};
+use thiserror::Error;
+use tokio::{
+    net::UdpSocket,
+    sync::{mpsc, oneshot, Mutex},
+};
+
+/// Maximum buffer size for response packet
+const FORWARDER_BUFFER_SIZE: usize = 4096;
+/// Channel size for forward messages
+const CHANNEL_SIZE: usize = 256;
+/// Default timeout for upstream DNS queries
+const DEFAULT_QUERY_TIMEOUT: Duration = Duration::from_secs(2);
+/// Maximum number of DNS IDs
+const DNS_ID_SPACE: u32 = 65536;
+
+/// Send a result through the response channel, logging a warning if the receiver was dropped
+macro_rules! send_channel_response {
+    ($tx:expr, $result:expr) => {
+        if $tx.send($result).is_err() {
+            telio_log_warn!("Response channel dropped");
+        }
+    };
+}
+
+/// Errors returned when forwarding a DNS query
+#[derive(Error, Debug)]
+pub enum ForwardError {
+    /// Failed upstream socket bind operation
+    #[error("Failed socket bind operation: {0}")]
+    SocketBindFailed(#[from] io::Error),
+    /// Failed to send a DNS query to the upstream resolver
+    #[error("Failed to send DNS query: {0}")]
+    SendFailed(io::Error),
+    /// No upstreams configured
+    #[error("No upstream resolvers configured")]
+    NoUpstreams,
+    /// The upstream resolvers did not respond within the configured timeout
+    #[error("DNS query timed out")]
+    Timeout,
+    /// The forwarder channel was closed
+    #[error("Forwarder channel closed")]
+    ChannelClosed,
+    /// Too many concurrent requests
+    #[error("Too many concurrent requests in flight")]
+    TooManyRequests,
+    /// The DNS packet is too short
+    #[error("DNS packet too short")]
+    PacketTooShort,
+}
+
+/// DNS query forwarder bound to the tunnel interface
+///
+/// Forwards raw DNS queries to upstream resolvers.
+/// Next upstream resolver is tried if no response is received after `timeout`.
+#[derive(Clone, Debug)]
+pub(crate) struct RawForwarder {
+    tx: mpsc::Sender<ForwardQuery>,
+    upstreams: Arc<Mutex<UpstreamState>>,
+    timeout: Arc<Mutex<Duration>>,
+}
+
+/// Upstream resolver list with a generation counter to detect changes
+#[derive(Clone, Debug, Default)]
+struct UpstreamState {
+    addrs: Vec<SocketAddr>,
+    generation: u64,
+}
+
+/// Internal message for forwarding a DNS query
+struct ForwardQuery {
+    /// Raw DNS packet bytes to forward
+    query_bytes: Vec<u8>,
+    /// Channel to send the response back to the caller
+    respond_to: oneshot::Sender<Result<Vec<u8>, ForwardError>>,
+}
+
+/// In progress DNS query awaiting an upstream response
+struct PendingQuery {
+    /// Original request ID
+    original_id: u16,
+    /// Raw DNS packet bytes
+    query_bytes: Vec<u8>,
+    /// Current upstream index for request
+    upstream_index: usize,
+    /// Generation of the upstream list when this query was last dispatched
+    upstream_generation: u64,
+    /// Channel to send the response back to the caller
+    respond_to: oneshot::Sender<Result<Vec<u8>, ForwardError>>,
+    /// Instant when request times out
+    deadline: Instant,
+}
+
+/// Extract the 16-bit transaction ID of a DNS packet
+fn get_dns_id(packet: &[u8]) -> Result<u16, ForwardError> {
+    if packet.len() < DNS_HEADER_OFFSET {
+        return Err(ForwardError::PacketTooShort);
+    }
+
+    // This is ok because the size is checked above
+    #[allow(clippy::indexing_slicing)]
+    Ok(u16::from_be_bytes([packet[0], packet[1]]))
+}
+
+/// Overwrite the 16-bit transaction ID of a DNS packet
+fn set_dns_id(packet: &mut [u8], id: u16) -> Result<(), ForwardError> {
+    if packet.len() < DNS_HEADER_OFFSET {
+        return Err(ForwardError::PacketTooShort);
+    }
+    let bytes = id.to_be_bytes();
+
+    // This is ok because the size is checked above
+    #[allow(clippy::indexing_slicing)]
+    {
+        packet[0] = bytes[0];
+        packet[1] = bytes[1];
+    }
+    Ok(())
+}
+
+/// Allocate an unused internal DNS transaction ID, returns `None` if all IDs are in use
+fn allocate_id(pending: &HashMap<u16, PendingQuery>, next_id: &mut u16) -> Option<u16> {
+    for _ in 0..DNS_ID_SPACE {
+        let candidate = *next_id;
+        *next_id = next_id.wrapping_add(1);
+        if !pending.contains_key(&candidate) {
+            return Some(candidate);
+        }
+    }
+    None
+}
+
+impl RawForwarder {
+    /// Create a new DNS forwarder
+    pub(crate) async fn new() -> Result<Self, ForwardError> {
+        let socket = Arc::new(UdpSocket::bind("0.0.0.0:0").await?);
+
+        bind_to_tun(&socket)?;
+
+        let (tx, rx) = mpsc::channel(CHANNEL_SIZE);
+        let upstreams = Arc::new(Mutex::new(UpstreamState::default()));
+        let timeout = Arc::new(Mutex::new(DEFAULT_QUERY_TIMEOUT));
+
+        tokio::spawn(Self::run(socket, rx, upstreams.clone(), timeout.clone()));
+
+        Ok(RawForwarder {
+            tx,
+            upstreams,
+            timeout,
+        })
+    }
+
+    /// Forward a DNS query to the configured upstream resolver
+    pub(crate) async fn query(&self, request: &[u8]) -> Result<Vec<u8>, ForwardError> {
+        let (respond_to, recv) = oneshot::channel();
+        let msg = ForwardQuery {
+            query_bytes: request.to_vec(),
+            respond_to,
+        };
+
+        self.tx
+            .send(msg)
+            .await
+            .map_err(|_| ForwardError::ChannelClosed)?;
+
+        recv.await.map_err(|_| ForwardError::ChannelClosed)?
+    }
+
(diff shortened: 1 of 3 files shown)


crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+    /// The forwarder channel was closed
+    #[error("Forwarder channel closed")]
+    ChannelClosed,
+    /// Too many concurrent requests
+    #[error("Too many concurrent requests in flight")]
+    TooManyRequests,
+    /// The DNS packet is too short
+    #[error("DNS packet too short")]
+    PacketTooShort,
+}
```
sfraczek, line 62: nit: SocketBind and Send read like operation names, while the rest of the variants (NoUpstreams, Timeout, ChannelClosed, TooManyRequests, PacketTooShort) describe an error state. Consider renaming them for consistency.

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+            locked.first().cloned()
+        };
+
+        let upstream_addr = match first_upstream {
+            Some(addr) => addr,
+            None => {
+                send_channel_response!(msg.respond_to, Err(ForwardError::NoUpstreams));
+                return;
+            }
+        };
```
sfraczek: nit: Consider this:
```suggestion
let Some(upstream_addr) = upstreams.lock().await.first().cloned() else {
    send_channel_response!(msg.respond_to, Err(ForwardError::NoUpstreams));
    return;
};
```

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+        telio_log_debug!("Forwarder exiting");
+    }
+
+    /// Return the earliest deadline
+    fn next_deadline(pending: &HashMap<u16, PendingQuery>) -> Option<Instant> {
+        pending.values().map(|e| e.deadline).min()
+    }
+
+    /// Handle new query
+    async fn handle_new_query(
```
sfraczek, line 288: nit: Consider extracting the body into a `Result`-returning helper (e.g. `prepare_query`) and dispatching the result once at the end. The current shape repeats the `match { Ok(v) => v, Err(e) => { send_channel_response!(...); return; } }` pattern four times — collapsing them with `?` would significantly shorten the function. But it is readable as it is so just a nitpick. Not sure if it would be readable after doing this.

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+        upstreams: &Arc<Mutex<Vec<SocketAddr>>>,
+        timeout: &Arc<Mutex<Duration>>,
+        pending: &mut HashMap<u16, PendingQuery>,
+    ) {
+        let now = Instant::now();
+        let expired_ids: Vec<(u16, bool)> = pending
+            .iter()
+            .filter(|(_, entry)| entry.deadline <= now || entry.respond_to.is_closed())
+            .map(|(&id, entry)| (id, entry.respond_to.is_closed()))
+            .collect();
```
sfraczek, line 395: nit: `is_closed()` is called twice per entry — once in `filter`, once in `map`. A single-pass `filter_map` is possible:

```rust
let expired_ids: Vec<(u16, bool)> = pending
    .iter()
    .filter_map(|(&id, entry)| {
        let closed = entry.respond_to.is_closed();
        (entry.deadline <= now || closed).then_some((id, closed))
    })
    .collect();
```

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+                        Ok((n, src)) => {
+                            // validate the src IP
+                            let is_known_upstream = {
+                                let locked = upstreams.lock().await;
+                                locked.iter().any(|u| u.ip() == src.ip())
+                            };
+                            if !is_known_upstream {
+                                telio_log_warn!("Received DNS response from unknown source: {src}, ignoring");
+                                continue;
+                            }
```
sfraczek, line 242: nit: this branch has no test

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+fn allocate_id(pending: &HashMap<u16, PendingQuery>, next_id: &mut u16) -> Option<u16> {
+    for _ in 0..DNS_ID_SPACE {
+        let candidate = *next_id;
+        *next_id = next_id.wrapping_add(1);
+        if !pending.contains_key(&candidate) {
+            return Some(candidate);
+        }
+    }
+    None
+}
```
sfraczek, line 143: `allocate_id` walks `pending` linearly and bails out after a full sweep — correctness depends on IDs being released back into the pool once a query is delivered or times out. That contract is not directly tested anywhere; a regression here (e.g. an entry forgotten in `pending` on some error path) would show up as a slow leak, not a unit test failure.

Easy smoke test: drive the forwarder through more queries than `DNS_ID_SPACE` and assert no `TooManyRequests`. The looping echo stub used by `spawn_multi_stub` is a good base — `spawn_stub` exits after one packet so it cannot serve enough queries here.

```rust
#[tokio::test]
async fn ids_are_released_after_response() {
    let (addr, _h) = spawn_multi_stub(DNS_ID_SPACE as usize + 1000).await;
    let forwarder = RawForwarder::new().await.unwrap();
    forwarder.set_upstreams(vec![addr]).await;

    for i in 0..(DNS_ID_SPACE as u32 + 1000) {
        let req = make_dns_packet(i as u16, b"x");
        forwarder.query(&req).await.expect("id pool exhausted — leak in `pending`");
    }
}
```

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+
+                    pending.insert(internal_id, entry);
+                }
+                None => {
+                    telio_log_warn!("All upstreams exhausted for request: {internal_id}");
+                    send_channel_response!(entry.respond_to, Err(ForwardError::Timeout));
+                }
+            }
+        }
+    }
```
sfraczek, line 448: Bug: `upstream_index` is captured at submit time but resolved here against `current_upstreams`, which may have been replaced via `set_upstreams` while this query was pending. After a swap, `entry.upstream_index + 1` can point at an unrelated resolver — possibly the same one we just timed out on.

Suggestion: store a snapshot of the upstream list (or the next `SocketAddr` directly) inside `PendingQuery` so retry behavior is independent of mutations to the shared list.

Regression test that fails on the current implementation with `count_a == 2` (the retry hits the same blackhole that just timed out):

```rust
// Demonstrates that `PendingQuery::upstream_index` is unstable across
// `set_upstreams`: the index is captured at submit time but resolved later
// against the *current* upstream list, so a reorder/replace mid-flight makes
// the retry land on the wrong resolver — possibly the same one we just
// timed out on.
//
// Sequence:
//   1. upstreams = [addr_a]; submit one query → goes to addr_a (index 0).
//   2. mid-flight, swap upstreams to [echo_addr, addr_a].
//   3. timeout fires; retry uses next_index = upstream_index + 1 = 1.
//   4. current_upstreams[1] is addr_a → blackhole gets a SECOND packet.
//
// The unambiguous bug signal is `count_a == 2`. Whether the query should
// ultimately succeed via echo or fail with Timeout depends on the eventual
// fix design (snapshot at submit time vs. "use latest list, skip already-tried").
#[tokio::test]
async fn retry_targets_wrong_upstream_after_list_reorder() {
    use std::sync::atomic::{AtomicUsize, Ordering};

    // Counting blackhole: records every datagram but never sends a reply.
    let socket_a = UdpSocket::bind(\"127.0.0.1:0\").await.unwrap();
    let addr_a = socket_a.local_addr().unwrap();
    let count_a = Arc::new(AtomicUsize::new(0));
    let count_a_task = count_a.clone();
    let _stub_a = tokio::spawn(async move {
        let mut buf = vec![0u8; 4096];
        loop {
            if socket_a.recv_from(&mut buf).await.is_ok() {
                count_a_task.fetch_add(1, Ordering::SeqCst);
            }
        }
    });

    let (echo_addr, _echo) = spawn_stub(StubBehavior::Echo).await;

    let forwarder = RawForwarder::new().await.unwrap();
    forwarder.set_upstreams(vec![addr_a]).await;
    forwarder.set_timeout(Duration::from_millis(100)).await;

    let f = forwarder.clone();
    let query_handle = tokio::spawn(async move {
        let request = make_dns_packet(TEST_PACKET_ID, TEST_DNS_PAYLOAD);
        f.query(&request).await
    });

    tokio::time::sleep(Duration::from_millis(20)).await;
    assert_eq!(count_a.load(Ordering::SeqCst), 1, "initial query did not reach addr_a yet");

    // Reorder upstreams BEFORE timeout fires:
    //   index 0 -> echo_addr
    //   index 1 -> addr_a   <-- this is what handle_timeouts will pick
    forwarder.set_upstreams(vec![echo_addr, addr_a]).await;

    let _result = query_handle.await.unwrap();

    assert_eq!(
        count_a.load(Ordering::SeqCst),
        1,
        "addr_a was retried after upstream list reorder — got {} packets, expected 1",
        count_a.load(Ordering::SeqCst)
    );
}
```

crates/telio-dns/src/forwarder.rs
```diff
@@ -0,0 +1,803 @@
...
+    let bytes = id.to_be_bytes();
+
+    // This is ok because the size is checked above
+    #[allow(clippy::indexing_slicing)]
+    {
+        packet[0] = bytes[0];
+        packet[1] = bytes[1];
+    }
+    Ok(())
+}
```
sfraczek, line 131: Larger refactor option (only worth it if more header inspection is on the horizon — `get_flags`, `is_response`, `rcode`, `qdcount`, etc.): split validation from access. A `validate_dns_header` gate validates once and returns a typed reference; getters/setters take the already-validated `&[u8; 12]`, do not check length, do not return `Result`, do not need `#[allow]`.

```rust
fn validate_dns_header(packet: &[u8]) -> Result<&[u8; DNS_HEADER_OFFSET], ForwardError> {
    packet.first_chunk().ok_or(ForwardError::PacketTooShort)
}

fn validate_dns_header_mut(
    packet: &mut [u8],
) -> Result<&mut [u8; DNS_HEADER_OFFSET], ForwardError> {
    packet.first_chunk_mut().ok_or(ForwardError::PacketTooShort)
}

fn get_dns_id(header: &[u8; DNS_HEADER_OFFSET]) -> u16 {
    u16::from_be_bytes([header[0], header[1]])
}

fn set_dns_id(header: &mut [u8; DNS_HEADER_OFFSET], id: u16) {
    let bytes = id.to_be_bytes();
    header[0] = bytes[0];
    header[1] = bytes[1];
}
```

Indexing `[0]`/`[1]` on `&[u8; 12]` is compile-time safe (the type guarantees the length), so clippy stays quiet without `#[allow]`. Call sites validate once on entry to `handle_new_query` / `handle_response` and then pass the typed header around.

Trade-off: two layers instead of one for the current ID-only use case. If ID is all you will ever read from the header, this is over-engineered. If more header fields show up later, this scales without duplicating length checks across every getter.

crates/telio-dns/src/forwarder.rs
tomasz-grz: Good idea

crates/telio-dns/src/forwarder.rs
tomasz-grz: Would this still hold the lock on `upstreams` while doing `send_channel_response`?

crates/telio-dns/src/forwarder.rs
tomasz-grz: It's not so straight forward, since it only checks for the IP address, and in unit tests all the injected responses would be from localhost 🤔 It's probably better tested in `nat-lab`

Unless we change the check to include source port as well, it would be more strict. But I don't know how often could servers use ephemeral ports for their responses (for example for load balancing).. probably need to check this more

crates/telio-dns/src/forwarder.rs
tomasz-grz: Not in scope to have other header inspection at the moment.

crates/telio-dns/src/forwarder.rs
tomasz-grz: Hmmm, It's a good point, but I'm afraid storing a snapshot would would be too many allocations per query , it also would mean that all pending requests would have "stale" list after an update.. But I don't expect `set_upstreams` to be called so often during runtime. 

Another option would be to reset the `upstream_index` to `0` for all PendingQueries whenever `set_upstreams` is called 🤔

crates/telio-dns/src/forwarder.rs
tomasz-grz: ids are removed from `pending` in `handle_timeouts` and `handle_response`.
But I'll add the test.

crates/telio-dns/src/forwarder.rs
tomasz-grz: Updated

crates/telio-dns/src/forwarder.rs
tomasz-grz: Changed

crates/telio-dns/src/forwarder.rs
tomasz-grz: Updated

crates/telio-dns/src/forwarder.rs
tomasz-grz: I think this is better tested in nat-lab

sfraczek (commented):
Checking all threads...

crates/telio-dns/src/forwarder.rs
sfraczek: I haven't thought about that. It's safer to have it separate.

crates/telio-dns/src/forwarder.rs
tomasz-grz: To be honest that's why I added the `send_channel_response` macro..

Something like this? That result triple is cursed though..
```rust
    async fn prepare_new_query(
        bytes: &[u8],
        pending: &mut HashMap<u16, PendingQuery>,
        next_id: &mut u16,
    ) -> Result<(u16, u16, Vec<u8>), ForwardError> {
        let original_id = get_dns_id(bytes)?;
        let internal_id = allocate_id(pending, next_id).ok_or(ForwardError::TooManyRequests)?;

        let mut rewritten = bytes.to_owned();
        set_dns_id(&mut rewritten, internal_id)?;

        Ok((original_id, internal_id, rewritten))
    }
```

tomasz-grz: Updated

~Another option would be to have `handle_new_query` return a result and have the caller `send_channel_response!`~ 🤔  Never mind, that makes `forward_msg` ownership quite awkward

djkarwowski:
+1

sfraczek (dismissed):
+1

sfraczek (approved):

PR #1714, opened 2026-03-18
LLT-7043: pytest fixtures for helpers

gytsto:
### Problem

Nat-lab test files use a factory-pattern for environment setup where each test manually calls `setup_mesh_nodes()` or `setup_environment()` at runtime, managing `AsyncExitStack` lifecycle inline. This leads to:

- Environment setup and teardown time is included in the test's reported runtime, inflating test durations and making it difficult to distinguish actual test logic performance from infrastructure overhead
- Duplicated setup/teardown boilerplate across every test function
- Inconsistent error handling during teardown (e.g. iptables cleanup failures crashing teardown)
- Proliferation of specialized fixture variants (`env_vpn_conntracked`, `env_pq_nlx_vpn`, `env_pq_nlx_vpn_rekey`, etc.) that differ only in minor parameter mutations
- Tight coupling between test logic and infrastructure orchestration, making tests harder to read and maintain

### Solution

Introduce a `helpers_fixtures.py` module with two generic setup-phase fixtures — `env_mesh` and `env` — that handle all environment lifecycle in pytest's fixture phase. By moving setup into fixtures, environment provisioning and teardown time is no longer counted as part of the test's execution duration, giving more accurate test timing metrics. Key design decisions:

- **Dynamic node resolution**: `_resolve_setup_params()` uses `request.getfixturevalue()` to auto-detect which `alpha_setup_params`, `beta_setup_params`, `gamma_setup_params` are parametrized on the test, eliminating the need for separate 1-node/2-node/3-node fixture variants
- **Compositional `vpn_tags` fixture**: A default fixture returning `[]` that tests override at class/module level to inject VPN server connections, replacing dedicated VPN-specific fixtures
- **Autouse mutation fixtures**: Test classes define small autouse fixtures (e.g. `_mutate_pq_version`, `_mutate_conntracker`) to apply parameter mutations, replacing 5 specialized environment fixtures with the generic `env`/`env` + composition
- **Robust teardown**: `linux_router.py` wraps all iptables teardown commands in `try/except ProcessExecError` to prevent cascading failures

This reduces the fixture count from 22 specialized fixtures to 2 generic ones (plus 1 kept specialized: `env_mesh_3node_ring_fw` which requires pre-setup API manipulation). Each of the 19 converted test files gets its own commit for reviewability.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/helpers_fixtures.py
+++ b/nat-lab/tests/helpers_fixtures.py
@@ -1,0 +1,237 @@
+"""Pytest fixtures wrapping helpers.setup_environment / setup_connections."""
+
+from __future__ import annotations
+
+import asyncio
+import pytest
+import pytest_asyncio
+from contextlib import AsyncExitStack
+from tests.helpers import (
+    SetupParameters,
+    setup_connections as _setup_connections,
+    setup_environment as _setup_environment,
+    setup_mesh_nodes as _setup_mesh_nodes,
+    Environment,
+)
+from tests.helpers_vpn import VpnConfig
+from tests.mesh_api import API
+from tests.utils.connection import ConnectionTag
+from tests.utils.connection_util import ConnectionManager
+from typing import AsyncGenerator, Callable, Awaitable, List, Optional, Union, Tuple
+
+
+# ---------------------------------------------------------------------------
+# 1) Shared AsyncExitStack
+# ---------------------------------------------------------------------------
+@pytest_asyncio.fixture(name="exit_stack")
+async def _exit_stack() -> AsyncGenerator[AsyncExitStack, None]:
+    """Provide an AsyncExitStack whose lifetime spans the entire test."""
+    stack = AsyncExitStack()
+    yield stack
+    try:
+        await asyncio.wait_for(stack.aclose(), timeout=60)
+    except asyncio.TimeoutError:
+        raise RuntimeError(
+            "exit_stack fixture teardown timed out after 60s — "
+            "likely a context manager is stuck during cleanup"
+        ) from None
+
+
+# ---------------------------------------------------------------------------
+# 2) Factory fixture for setup_connections
+# ---------------------------------------------------------------------------
+@pytest.fixture(name="setup_connections_factory")
+def _setup_connections_factory(
+    exit_stack: AsyncExitStack,
+) -> Callable[..., Awaitable[List[ConnectionManager]]]:
+    """Return a callable that forwards to helpers.setup_connections
+    using the shared *exit_stack*.
+
+    Usage inside a test or another fixture::
+
+        managers = await setup_connections_factory([ConnectionTag.DOCKER_VPN_1])
+    """
+
+    async def _factory(
+        connection_parameters: List[
+            Union[
+                ConnectionTag,
+                Tuple[ConnectionTag, Optional[list]],
+            ]
+        ],
+    ) -> List[ConnectionManager]:
+        return await _setup_connections(exit_stack, connection_parameters)
+
+    return _factory
+
+
+# ---------------------------------------------------------------------------
+# 2b) Factory fixture for setup_mesh_nodes
+# ---------------------------------------------------------------------------
+@pytest.fixture(name="setup_mesh_nodes_factory")
+def _setup_mesh_nodes_factory(
+    exit_stack: AsyncExitStack,
+) -> Callable[..., Awaitable[Environment]]:
+    """Return a callable that forwards to helpers.setup_mesh_nodes
+    using the shared *exit_stack*."""
+
+    async def _factory(
+        instances: List[SetupParameters],
+        is_timeout_expected: bool = False,
+        provided_api=None,
+        vpn=None,
+    ) -> Environment:
+        return await _setup_mesh_nodes(
+            exit_stack,
+            instances,
+            is_timeout_expected=is_timeout_expected,
+            provided_api=provided_api,
+            vpn=vpn,
+        )
+
+    return _factory
+
+
+# ===========================================================================
+# VPN TAGS — composable VPN configuration
+# ===========================================================================
+
+
+@pytest.fixture(name="vpn_tags")
+def _vpn_tags() -> List[ConnectionTag]:
+    """Default: no VPN servers. Override in test module or conftest to add VPN."""
+    return []
+
+
+# ===========================================================================
+# ENVIRONMENT FIXTURES
+# ===========================================================================
+
+
+# --- Optional VPN-side connection (for ping-back) ---
+
+
+@pytest_asyncio.fixture(name="vpn_server_connection")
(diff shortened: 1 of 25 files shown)


nat-lab/tests/test_pq.py
```diff
@@ -246,17 +248,14 @@ async def test_pq_vpn_rekey(
...
-        await client_alpha.wait_for_log("Successful PQ REKEY")
+        async with new_connection_by_tag(ConnectionTag.VM_LINUX_NLX_1) as nlx_conn:
+            preshared_before = inspect_preshared_key(nlx_conn)
+            await client_alpha.wait_for_log("Successful PQ REKEY")
 
-        preshared_after = inspect_preshared_key(nlx_conn)
-        assert (
-            preshared_after != preshared_before
-        ), "Preshared key not changed on the nlx server"
+            preshared_after = inspect_preshared_key(nlx_conn)
```
matislovas: This should be `await`'ed ..? Or no?

gytsto: looks like it should, but even before my changes it was not awaited, so idk how it worked :smile: will fix it

gytsto: this is fixed now

nat-lab/tests/utils/router/linux_router.py
```diff
@@ -384,36 +384,48 @@ async def disable_path(self, address: str) -> AsyncIterator:
...
+                        "--source",
+                        address,
+                        "--jump",
+                        "DROP",
+                    ],
+                    quiet=True,
+                ).execute()
+            except ProcessExecError:
+                log.warning(
+                    "disable_path cleanup: INPUT rule for %s already removed",
```
tomasz-grz: Should we check for the ProcessExecError error message like we do in DockerProcess, in case it failed with for some other reason? Otherwise log or re-raise the Exception

nat-lab/tests/utils/process/docker_process.py
```diff
@@ -137,30 +155,39 @@ async def mark_as_done():
...
-                            text=True,
-                        )
-                        if proc.returncode != 0:
-                            log.warning(
-                                "[%s] Cleanup failed: %s",
+                        while inspect["Pid"] == 0 and inspect["ExitCode"] is None:
+                            inspect = await self._execute.inspect()
+                            await asyncio.sleep(0.01)
+                        if inspect["ExitCode"] is None:
+                            proc = subprocess.run(
```
tomasz-grz: Is `subrocess.run()` a blocking call?

nat-lab/tests/utils/process/docker_process.py
```diff
@@ -137,30 +155,39 @@ async def mark_as_done():
...
-                            stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE,
-                            text=True,
-                        )
-                        if proc.returncode != 0:
-                            log.warning(
-                                "[%s] Cleanup failed: %s",
+                        while inspect["Pid"] == 0 and inspect["ExitCode"] is None:
+                            inspect = await self._execute.inspect()
+                            await asyncio.sleep(0.01)
```
tomasz-grz: I wonder if we can use `await asyncio.wait_for()` somehow to avoid the spin lock with the `sleep()`

nat-lab/tests/helpers_fixtures.py
```diff
@@ -0,0 +1,237 @@
...
+# Auto-detection of *_setup_params fixtures
+# ---------------------------------------------------------------------------
+
+_PARAM_NAMES = ["alpha_setup_params", "beta_setup_params", "gamma_setup_params"]
+
+
+def _resolve_setup_params(request: pytest.FixtureRequest) -> List[SetupParameters]:
+    """Dynamically resolve available *_setup_params fixtures.
+
+    Tries alpha, beta, gamma in order. Stops at the first missing fixture.
```
tomasz-grz: Is there a case where we can expect only "alpha" and "gamma" to be defined?

nat-lab/tests/utils/router/linux_router.py
gytsto: fixed

nat-lab/tests/utils/process/docker_process.py
gytsto: it was, fixed

nat-lab/tests/utils/process/docker_process.py
gytsto: added

nat-lab/tests/helpers_fixtures.py
gytsto: there usually is not, but you pin pointed an issue, fixed

matislovas:
+1

nat-lab/tests/utils/process/docker_process.py
```diff
@@ -84,31 +85,25 @@ async def execute(
...
+                return self
+            raise
         exit_code = inspect["ExitCode"]
 
         # 0 success
-        # suppress 137 linux sigkill, since we kill those processes
-        if exit_code and exit_code not in [0, 137]:
+        # 137 = SIGKILL (128+9), 143 = SIGTERM (128+15) — both are expected
+        # when the process is intentionally killed during cleanup
+        if exit_code and exit_code not in [0, 137, 143]:
```
tomasz-grz: What do you think about defining these as "const"? like `SIGKILL = 137`?

tomasz-grz: What if a process is unexpectedly terminated with SIGTERM in "normal" operation, would it still be ignored and hide any potential issue?

If so, we could ignore SIGTERM maybe only after a`kill_sent` flag is set by `kill_exec_if_running`?

nat-lab/tests/helpers_fixtures.py
tomasz-grz: Should we add a log for visibility at least in case of `pytest.FixtureLookupError`?
To indicate that "beta" was ignored, maybe that could save some debugging headaches later on.

tomasz-grz (commented):
Some small nits.
I wonder if such a big refactor deserves additional pair of eyes 👀

nat-lab/tests/helpers_fixtures.py
gytsto: now that I tought better, there is no thing, where beta is being skipped in general, so I reverted a change :smile:

nat-lab/tests/utils/process/docker_process.py
gytsto: addressed

tomasz-grz:
+1, could you just maybe squish some of the feedback commits?

gytsto:
will hold this open for a bit without merging, don't wanna merge it right now, since nat-lab is in very good shape right now

matislovas (approved):

PR #1724, opened 2026-03-25
Llt 7085 - Refactor telio-dns: Replace string-based error handling with strongly-typed enums across library and tests

Damos-dev:
### Problem

The `telio-dns` crate was using `String` as the error type across both public API and internal logic. This made error handling non-idiomatic and limited:

- no ability to match on specific error types
- error handling required string parsing at call sites
- inconsistent error propagation across modules
- reduced clarity of failure modes in the public API

Additionally, there was no clear separation between errors that are part of the public API contract and those used internally for DNS packet processing.

---

### Solution

Refactored error handling in `telio-dns` to use typed enums instead of `String`,
with a clear separation between public and internal error domains.

Key changes:

- Introduced a public `Error` enum (in `error.rs`) and `Result<T>` alias
- Introduced DnsIoError enum to wrap io::Error with context
- Replaced all `Result<_, String>` in the public API with typed errors
- Updated public traits (`DnsResolver`, `NameServer`) and constructors to return `Error`
- Defined a separate internal `PacketError` enum (in `nameserver.rs`) for DNS packet parsing and runtime errors
- Ensured internal errors are not exposed through the public API
- Replaced manual `format!`-based error construction with structured error variants
- Updated code outside of the `telio-dns` ( `device.rs`) to use typed errors and `From` for propagation

Additionally:
- Replaced `DnsResolverError(String)` with `DnsResolverError(#[from] telio_dns::error::Error)`
  to enable automatic error propagation via `?`

This change introduces a clean boundary between:
- **public API errors** (stable, typed, exposed via `Error`)
- **internal runtime errors** (private, implementation-specific via `PacketError`)

As a result, the API will be more robust, type-safe, and easier to extend without breaking external contracts.

### To do in the future (not subject of this PR):
- Refactor the codebase to enable dependency injection for key components, especially in the DNS server logic. This would allow unit tests to simulate errors from external modules (such as Neptune) or the OS, making it possible to test error propagation and handling in the public API.
- Currently, it is not feasible to fully test error scenarios in the DNS server using unit tests, because the code structure does not allow for mocking or simulating failures in dependencies like Neptune or system calls. As a result, some error-handling paths in the public API cannot be reliably covered by unit tests.

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/src/dns.rs
+++ b/crates/telio-dns/src/dns.rs
@@ -1,4 +1,8 @@
-use crate::{bind_tun, LocalNameServer, NameServer, Records};
+use crate::{
+    bind_tun,
+    error::{DnsIoError, Error, Result},
+    LocalNameServer, NameServer, Records,
+};
 use async_trait::async_trait;
 use ipnet::IpNet;
 use neptun::noise::Tunn;
@@ -25,14 +29,9 @@ pub trait DnsResolver {
     /// Stop the server.
     async fn stop(&self);
     /// Insert or update zone records used by the server.
-    async fn upsert(
-        &self,
-        zone: &str,
-        records: &Records,
-        ttl_value: TtlValue,
-    ) -> Result<(), String>;
+    async fn upsert(&self, zone: &str, records: &Records, ttl_value: TtlValue) -> Result<()>;
     /// Configure list of forward DNS servers for zone '.'.
-    async fn forward(&self, to: &[IpAddr]) -> Result<(), String>;
+    async fn forward(&self, to: &[IpAddr]) -> Result<()>;
     /// Get public key of this DNS server.
     fn public_key(&self) -> PublicKey;
     /// Get Peer of this DNS server with selected allowed IPs.
@@ -66,26 +65,17 @@ impl LocalDnsResolver {
         forward_ips: &[IpAddr],
         tun: Option<&Tun>,
         dns_features: &FeatureDns,
-    ) -> Result<Self, String> {
+    ) -> Result<Self> {
         let socket = UdpSocket::bind(SocketAddr::from(([127, 0, 0, 1], 0)))
             .await
-            .map_err(|e| {
-                telio_log_error!("Failed to bind dns socket: {:?}", e);
-                format!("Failed to bind dns socket: {e:?}")
-            })?;
+            .map_err(DnsIoError::BindDnsSocket)?;
 
         let socket_port = socket
             .local_addr()
-            .map_err(|e| {
-                telio_log_error!("Failed to get socket port: {:?}", e);
-                format!("Failed to get socket port: {e:?}",)
-            })?
+            .map_err(DnsIoError::GetSocketAddr)?
             .port();
 
-        bind_tun::set_tun(tun).map_err(|e| {
-            telio_log_debug!("Failing setting tun: {:?}", e);
-            format!("{e}")
-        })?;
+        bind_tun::set_tun(tun).map_err(DnsIoError::ConfigureTunnel)?;
 
         // DNS secret key
         let dns_secret_key = SecretKey::gen();
@@ -105,14 +95,17 @@ impl LocalDnsResolver {
         Ok(LocalDnsResolver {
             socket: Arc::new(socket),
             secret_key: dns_secret_key.clone(),
-            peer: Arc::new(Mutex::new(Tunn::new(
-                StaticSecret::from(dns_secret_key.into_bytes()),
-                telio_public_key,
-                None,
-                None,
-                0,
-                None,
-            )?)),
+            peer: Arc::new(Mutex::new(
+                Tunn::new(
+                    StaticSecret::from(dns_secret_key.into_bytes()),
+                    telio_public_key,
+                    None,
+                    None,
+                    0,
+                    None,
+                )
+                .map_err(|e| Error::CreateTunnel(e.to_string()))?,
+            )),
             socket_port,
             nameserver,
             auto_switch_ips,
@@ -133,19 +126,14 @@ impl DnsResolver for LocalDnsResolver {
         self.nameserver.stop().await;
     }
 
-    async fn upsert(
-        &self,
-        zone: &str,
-        records: &Records,
-        ttl_value: TtlValue,
-    ) -> Result<(), String> {
+    async fn upsert(&self, zone: &str, records: &Records, ttl_value: TtlValue) -> Result<()> {
         telio_log_debug!("Dns - upsert {:?} {:?}", zone, records);
-        Ok(self.nameserver.upsert(zone, records, ttl_value).await?)
+        self.nameserver.upsert(zone, records, ttl_value).await
     }
 
-    async fn forward(&self, to: &[IpAddr]) -> Result<(), String> {
+    async fn forward(&self, to: &[IpAddr]) -> Result<()> {
         telio_log_debug!("Dns - forward {:?}", to);
-        Ok(self.nameserver.forward(to).await?)
+        self.nameserver.forward(to).await
     }
 
     fn public_key(&self) -> PublicKey {
--- a/crates/telio-dns/src/error.rs
+++ b/crates/telio-dns/src/error.rs
@@ -1,0 +1,61 @@
+use crate::forwarder::ForwardError;
+use crate::zone::NordZoneError;
+use std::{io, net::AddrParseError};
+use thiserror::Error;
+
+/// Result type used by the telio-dns crate
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An IO error during DNS operations
+#[derive(Debug, Error)]
+pub enum DnsIoError {
+    /// Binding the DNS UDP socket
+    #[error("Failed to bind DNS socket: {0}")]
+    BindDnsSocket(io::Error),
+    /// Getting the local address of the DNS socket
+    #[error("Failed to get DNS socket local address: {0}")]
+    GetSocketAddr(io::Error),
+    /// Configuring tunnel binding for DNS
+    #[error("Failed to configure tunnel binding: {0}")]
+    ConfigureTunnel(io::Error),
+}
+
+/// Error type for the telio-dns crate
+///
+/// This enum represents all errors in the public interfaces
+/// that can occur when configuring or running the DNS resolver
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum Error {
+    /// An IO error during DNS socket or tunnel operations
+    #[error(transparent)]
+    Io(#[from] DnsIoError),
+
+    /// Failed to create WireGuard tunnel
+    #[error("Failed to create WireGuard tunnel: {0}")]
+    CreateTunnel(String),
+
+    /// Invalid DNS name
+    #[error("Invalid DNS name: {0}")]
+    InvalidDnsName(#[from] hickory_server::proto::error::ProtoError),
+
+    /// Record does not belong to the specified DNS zone
+    #[error("Record '{0}' does not belong to zone '{1}'")]
+    RecordOutsideZone(String, String),
+
+    /// Invalid IP address
+    #[error("Invalid IP address: {0}")]
+    AddressParse(#[from] AddrParseError),
+
+    /// Failed to create the raw DNS forwarder
+    #[error("Failed to create DNS forwarder: {0}")]
+    CreateForwarder(#[from] ForwardError),
+
+    /// Failed to upsert a record into a zone
+    #[error("Upsert failed: {0}")]
+    UpsertFailed(#[from] NordZoneError),
+
+    /// Nameserver is stopped
+    #[error("Nameserver stopped")]
+    Stopped,
+}
--- a/crates/telio-dns/src/error.rs
+++ b/crates/telio-dns/src/error.rs
@@ -59,3 +59,33 @@ pub enum Error {
     #[error("Nameserver stopped")]
     Stopped,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Test helper: creates io::Error with given message.
+    fn fake_io_error(msg: &str) -> io::Error {
+        io::Error::other(msg)
+    }
+
+    // Tests DnsIoError display format: "{context}: {source}" from thiserror::Error implementation
+    #[test]
+    fn dns_io_error_display() {
+        let err = DnsIoError::BindDnsSocket(fake_io_error("port busy"));
+        assert_eq!(err.to_string(), "Failed to bind DNS socket: port busy");
+    }
+
+    // Tests Error::Io — conversion from DnsIoError via #[from] from thiserror::Error implementation
+    #[test]
+    fn error_io_from_dns_io_error() {
+        let io_err = DnsIoError::GetSocketAddr(fake_io_error("no address"));
+        let err = Error::from(io_err);
+        assert!(matches!(err, Error::Io(_)));
+        // transparent display should show inner DnsIoError message
+        assert_eq!(
+            err.to_string(),
+            "Failed to get DNS socket local address: no address"
+        );
+    }
+}
--- a/crates/telio-dns/src/forward.rs
+++ b/crates/telio-dns/src/forward.rs
@@ -31,7 +31,7 @@ use hickory_server::{
 use telio_utils::{telio_log_debug, telio_log_info, telio_log_trace, telio_log_warn};
 use tokio::net::UdpSocket;
 
-use crate::bind_tun;
+use crate::{bind_tun, error::Error};
 
 #[derive(Default, Clone)]
 pub struct TelioRuntimeProvider(TokioRuntimeProvider);
@@ -141,7 +141,7 @@ impl ForwardAuthority {
         origin: Name,
         _zone_type: ZoneType,
         config: ForwardConfig,
-    ) -> Result<Self, String> {
+    ) -> Result<Self, Error> {
         telio_log_info!("loading forwarder config: {}", origin);
 
         let name_servers = config.name_servers;
--- a/crates/telio-dns/src/nameserver.rs
+++ b/crates/telio-dns/src/nameserver.rs
@@ -1,3 +1,4 @@
+use crate::error::Result as DnsResult;
 use crate::{
     forwarder::RawForwarder,
     packet_decoder::{find_nord_query, normalize_qname, parse_dns_query_packet, DnsParseError},
@@ -116,16 +117,11 @@ pub trait NameServer {
     /// Stop the server.
     async fn stop(&self);
     /// Configure list of forward DNS servers for zone '.'.
-    async fn forward(&self, to: &[IpAddr]) -> Result<(), String>;
+    async fn forward(&self, to: &[IpAddr]) -> DnsResult<()>;
     /// Configure list of forward DNS servers with explicit socket addresses.
-    async fn forward_to_addrs(&self, to: &[SocketAddr]) -> Result<(), String>;
+    async fn forward_to_addrs(&self, to: &[SocketAddr]) -> DnsResult<()>;
     /// Insert or update zone records used by the server.
-    async fn upsert(
-        &self,
-        zone: &str,
-        records: &Records,
-        ttl_value: TtlValue,
-    ) -> Result<(), String>;
+    async fn upsert(&self, zone: &str, records: &Records, ttl_value: TtlValue) -> DnsResult<()>;
 }
 
 /// Helper to update wg timers
@@ -169,13 +165,9 @@ impl LocalNameServer {
     pub async fn new(
         forward_ips: &[IpAddr],
         use_raw_forwarder: bool,
-    ) -> Result<Arc<RwLock<Self>>, String> {
+    ) -> DnsResult<Arc<RwLock<Self>>> {
         let raw_forwarder: Option<RawForwarder> = if use_raw_forwarder {
-            Some(
-                RawForwarder::new()
-                    .await
-                    .map_err(|e| format!("Failed to create forwarder: {e:?}"))?,
-            )
+            Some(RawForwarder::new().await?)
         } else {
             None
         };
@@ -403,7 +395,7 @@ impl LocalNameServer {
                     let response = forwarder
                         .query(&raw_query)
                         .await
-                        .map_err(|e| format!("Forward failed: {e:?}"))?;
+                        .map_err(|e| PacketError::LookupFailed(Box::new(e)))?;
 
                     telio_log_debug!("Forwarder responded");
                     if let Some(dns_packet) = DnsPacket::new(&response) {
@@ -416,9 +408,8 @@ impl LocalNameServer {
                     let resolver = Resolver::new();
                     let zones = nameserver.zones().await;
 
-                    let forward = MessageRequest::from_bytes(&raw_query).map_err(|_| {
-                        String::from("Failed to build MessageRequest from request packet")
-                    })?;
+                    let forward = MessageRequest::from_bytes(&raw_query)
+                        .map_err(|_| PacketError::DecodeMessageRequestFailed)?;
                     let dns_request =
                         Request::new(forward, request_info.dns_source(), Protocol::Udp);
                     telio_log_debug!("DNS request: {:?}", &dns_request);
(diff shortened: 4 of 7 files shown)


nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/libtelio?pullRequest=1724) <br/>All committers have signed the CLA.

crates/telio-dns/src/dns.rs
```diff
@@ -30,9 +30,9 @@ pub trait DnsResolver {
         zone: &str,
         records: &Records,
         ttl_value: TtlValue,
-    ) -> Result<(), String>;
+    ) -> crate::error::Result<()>;
     /// Configure list of forward DNS servers for zone '.'.
-    async fn forward(&self, to: &[IpAddr]) -> Result<(), String>;
+    async fn forward(&self, to: &[IpAddr]) -> crate::error::Result<()>;
```
tomasz-grz: I would personally prefer to see just `Result<()>` here, or at least keep the `Result<(), Error>`

you can import it with `use` like `use crate::error::Result;`

Or in this case 
```rust
use crate::{bind_tun, LocalNameServer, NameServer, Records, error::Error, error::Result};
```

crates/telio-dns/src/dns.rs
```diff
@@ -66,25 +66,25 @@ impl LocalDnsResolver {
         forward_ips: &[IpAddr],
         tun: Option<&Tun>,
         exit_dns: Option<FeatureExitDns>,
-    ) -> Result<Self, String> {
+    ) -> crate::error::Result<Self> {
         let socket = UdpSocket::bind(SocketAddr::from(([127, 0, 0, 1], 0)))
             .await
             .map_err(|e| {
                 telio_log_error!("Failed to bind dns socket: {:?}", e);
```
tomasz-grz: Are these logs redundant? Will the returned Error be logged anyway?

crates/telio-dns/src/error.rs
```diff
@@ -0,0 +1,92 @@
+use std::{io, net::AddrParseError};
+
+// use hickory_server::proto::rr::Name;
+// use hickory_server::proto::rr::LowerName;
+// use neptun::noise::errors::WireGuardError;
```
tomasz-grz: Please clean up

crates/telio-dns/src/error.rs
```diff
@@ -0,0 +1,92 @@
...
+/// Error type for the telio-dns crate
+///
+/// This enum represents all errors in the public interfaces 
+/// that can occur when configuring or running the DNS resolver
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum Error {
+    /// Failed to bind DNS socket
+    #[error("Failed to bind DNS socket: {0}")]
+    BindDnsSocket(#[source] io::Error),
```
tomasz-grz: Using `#[from]` attribute is generally preferable, because it gives you the `From`, `Into` and source implementations for "free".

crates/telio-dns/src/error.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Nameserver is stopped
+    #[error("Nameserver stopped")]
+    Stopped,
+}
+
+impl Error {
+    /// Create `BindDnsSocket` error
+    pub fn bind_dns_socket(err: io::Error) -> Self {
+        Self::BindDnsSocket(err)
+    }
```
tomasz-grz: What are these intended for?

If you use the [From](https://doc.rust-lang.org/core/convert/trait.From.html) trait by the `#[from]`, I believe you don't need to explicitly do any error type conversions.

crates/telio-dns/src/nameserver.rs
```diff
@@ -38,6 +40,61 @@ const TCP_MIN_HEADER: usize = 20;
 const MAX_CONCURRENT_QUERIES: usize = 256;
 const IDLE_TIME: Duration = Duration::from_secs(1);
 
+#[derive(Debug, Error)]
+#[non_exhaustive]
```
tomasz-grz: It's possibly fine to omit `#[non_exhaustive]` for internal errors.

crates/telio-dns/src/zone.rs
```diff
@@ -145,23 +146,23 @@ impl Authority for AuthoritativeZone {
...
+    ) -> std::result::Result<Self::Lookup, LookupError> {
         self.zone.search(request_info, lookup_options).await
     }
 
     async fn get_nsec_records(
         &self,
         name: &LowerName,
         lookup_options: LookupOptions,
-    ) -> Result<Self::Lookup, LookupError> {
+    ) -> std::result::Result<Self::Lookup, LookupError> {
```
tomasz-grz: The`Result<T, E>` is already included as part of the "Standard library", you don't need to explicitly specify `std::result::`

crates/telio-dns/src/error.rs
```diff
@@ -0,0 +1,92 @@
...
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum Error {
+    /// Failed to bind DNS socket
+    #[error("Failed to bind DNS socket: {0}")]
+    BindDnsSocket(#[source] io::Error),
+
+    /// Failed to get DNS socket local address
+    #[error("Failed to get DNS socket local address: {0}")]
+    GetDnsSocketLocalAddr(#[source] io::Error),
```
tomasz-grz: Since these variants are already in the DNS crate, what do you think dropping it from the error type name?
Something like `GetLocalAddr` or simply `LocalAddr` would be less verbose in my opinion.

crates/telio-dns/src/dns.rs
```diff
@@ -108,7 +108,8 @@ impl LocalDnsResolver {
                 None,
                 0,
                 None,
-            )?)),
+            ).map_err(|e| Error::CreateTunnel(e.to_string()))?,
```
tomasz-grz: Instead of manual conversion, could we use the `#[from]` attribute here to have the types converted just by `?`

crates/telio-dns/src/error.rs
Damos-dev: Fixed.

crates/telio-dns/src/dns.rs
tomasz-grz: Or actually 
```rust
use crate::{bind_tun, LocalNameServer, NameServer, Records, error::{Error, Result}};
```

crates/telio-dns/src/zone.rs
```diff
@@ -174,7 +175,7 @@ pub(crate) struct ForwardZone {
 }
 
 impl ForwardZone {
-    pub(crate) async fn new(name: &str, ips: &[IpAddr]) -> Result<Self, String> {
+    pub(crate) async fn new(name: &str, ips: &[IpAddr]) -> Result<Self> {
```
tomasz-grz: I see, you are replacing the `std::result::Result` type here with the `crate::error::Result` type def.

I would recommend simply to do `Result<Self, Error>` here.
Since you already import it `use crate::error::Error`

FIY you can also rename imports, if you have clashes or want to make more explicit.

```rust
use crate::error::Error as DnsError;

/// then this would become
Result<Self, DnsError>
```

In theory it should be possible rename the imported Result (though I wouldn't recommend it).
```rust
use crate::error::Result as DnsResult;

/// then write this as 
DnsResult<Self>
```
This way it will also not clash with std Result

crates/telio-dns/src/dns.rs
Damos-dev: Done as `use crate::{bind_tun, error::{Error,Result}, LocalNameServer, NameServer, Records};` to keep proper order.

crates/telio-dns/src/dns.rs
Damos-dev: You are right!
With the recent change from string-based errors to typed enums, we no longer need to log errors at the point where they are created. Instead, errors can be propagated and handled (and logged) at the call site.

**At the moment, I don’t see any other logging** "at the boundary" where this code is called, which is why I left the existing logs in place. It’s possible they are still relied upon elsewhere (e.g. in tests or other integration points).

If there are no such dependencies, I’d suggest moving towards a proper “log at the boundary” approach: remove the logs from here and handle logging in device.rs, where the error is actually observed and acted upon. 

I think this change should be handled as a separate subtask and commit. Grouping all related logging changes into a focused commit with a clear  message - would make the intent and idea much easier to understand, rather than mixing it into change of the error handling approach. I have made LLT-7174 for that.

crates/telio-dns/src/zone.rs
Damos-dev: Good point - I like this approach and it does make the code a bit cleaner and more explicit.

One thing to keep in mind, though, is that aliasing like this can make it harder to find all usages later, using a simple text search. It will require a two-step process: first locating the alias definitions, and then searching for their usages.

So while it does help avoid clashes with `std::result::Result` and improves readability, it also introduces a bit of indirection in terms of future code analysis.

Overall, I agree it’s a reasonable trade-off, especially for clarity, but it’s worth being aware of this downside in the future.

crates/telio-dns/src/error.rs
Damos-dev: The first one can benefit from the "#[from]" but the rest will conflict as "thiserror" will try to generate impl for them as well, and it will cause a conflict. So we have now changed  ` BindDnsSocket(#[from] io::Error)` to use `#from` but two other errors from `io::error` will have to use "source" and later map_err.

crates/telio-dns/src/zone.rs
Damos-dev: Solved by aliasing crate::error::Result as TelioResult.

crates/telio-dns/src/zone.rs
Damos-dev: Anyway - solved by aliasing new Result with a different name.

crates/telio-dns/src/error.rs
Damos-dev: Error type will give us more detailed info and I would like to keep it. Yet.

crates/telio-dns/src/error.rs
Damos-dev: It's been changed to `?` but two other errors will have to use `map_err`.

crates/telio-dns/src/nameserver.rs
Damos-dev: At the moment we don't have any code what will benefit from that - so removed.

crates/telio-dns/src/dns.rs
Damos-dev: Not in this case, when in our `enum Error` in `error.rs` we have 3 different enum cases:
`#[error("Failed to bind DNS socket: {0}")]
    BindDnsSocket(#[from] io::Error),

    /// Failed to get DNS socket local address
    #[error("Failed to get DNS socket local address: {0}")]
    GetDnsSocketLocalAddr(#[source] io::Error),

    /// Failed to configure tunnel binding
    #[error("Failed to configure tunnel binding: {0}")]
    ConfigureTunnel(#[source] io::Error),` 
and all of them are based on the same io::Error. In this case `#[from]` will be not able to decide - which one to choose. Here `map_err` looks like a proper approach.

.unreleased/LLT-7085
```diff
@@ -0,0 +1 @@
+Refactor error handling in telio-dns: remove simple String and replace it with enums for both: public (public API) and private (internal) errors
```
tomasz-grz: We add a message here only if there are breaking changes to the App facing public API,
since the `telio-dns` module is not facing directly the final app, it should be ok to skip the message here.

crates/telio-dns/src/dns.rs
```diff
@@ -129,19 +120,14 @@ impl DnsResolver for LocalDnsResolver {
...
         telio_log_debug!("Dns - upsert {:?} {:?}", zone, records);
-        Ok(self.nameserver.upsert(zone, records, ttl_value).await?)
+        self.nameserver.upsert(zone, records, ttl_value).await
     }
 
-    async fn forward(&self, to: &[IpAddr]) -> Result<(), String> {
+    async fn forward(&self, to: &[IpAddr]) -> Result<()> {
         telio_log_debug!("Dns - forward {:?}", to);
-        Ok(self.nameserver.forward(to).await?)
+        self.nameserver.forward(to).await
```
tomasz-grz, line 136: Nice 👍

crates/telio-dns/src/error.rs
tomasz-grz: Then please make the other errors follow the same convention,

maybe something like `ConfigureDnsTunnel`, `CreateDnsTunnel` and their error descriptions, to better distinguish them from the actual VPN tunnel errors

crates/telio-dns/src/error.rs
tomasz-grz: Exactly, you can use `map_err` to convert the error types. I don't see these functions used anywhere.

crates/telio-dns/src/forward.rs
```diff
@@ -141,7 +141,7 @@ impl ForwardAuthority {
         origin: Name,
         _zone_type: ZoneType,
         config: ForwardConfig,
-    ) -> Result<Self, String> {
+    ) -> TelioResult<Self> {
```
tomasz-grz:
```suggestion
    ) -> Result<Self, Error> {
```

crates/telio-dns/src/forward.rs
```diff
@@ -31,7 +31,7 @@ use hickory_server::{
 use telio_utils::{telio_log_debug, telio_log_info, telio_log_trace, telio_log_warn};
 use tokio::net::UdpSocket;
 
-use crate::bind_tun;
+use crate::{bind_tun, error::Result as TelioResult};
```
tomasz-grz:
```suggestion
use crate::{bind_tun, error::Error};
```

crates/telio-dns/src/nameserver.rs
```diff
@@ -49,14 +107,9 @@ pub trait NameServer {
     /// Stop the server.
     async fn stop(&self);
     /// Configure list of forward DNS servers for zone '.'.
-    async fn forward(&self, to: &[IpAddr]) -> Result<(), String>;
+    async fn forward(&self, to: &[IpAddr]) -> TelioResult<()>;
```
tomasz-grz:
```suggestion
    async fn forward(&self, to: &[IpAddr]) -> Result<(), Error>;
```

crates/telio-dns/src/nameserver.rs
tomasz-grz: `#[non_exhaustive]` is still there 👀

crates/telio-dns/src/error.rs
tomasz-grz: I feel like the error variants of `nameserver::PacketError` are much more "human readable" than these, and personally prefer that style.

crates/telio-dns/src/nameserver.rs
tomasz-grz: If you really want to use the type alias here, maybe `DnsResult` would be a more suitable name?

tomasz-grz (commented):
Nice, this is taking shape.

Please update the PR name

crates/telio-dns/src/error.rs
tomasz-grz: If I'm not mistaken you don't need to derive `#[source]` in order to use `map_err`

crates/telio-dns/src/zone.rs
tomasz-grz: I personally don't see the benefit of using a type alias when there are multiple error types in the module.

tomasz-grz: If you really need to use an alias, since it's not a global libtelio type, I would be more in favor of something like `DnsResult`.

.unreleased/LLT-7085
mathiaspeters: I was also going to add that I think this file should be left empty. The purpose of the changelog is to communicate to apps what has changed for them, and this change shouldn't affect them

crates/telio-dns/src/forward.rs
Damos-dev: It's been changed to `DnsResult` as shadowing was required due to two Result definitions available.

crates/telio-dns/src/forward.rs
Damos-dev: It's been changed to DnsResult as shadowing was required due to two Result definitions available.

crates/telio-dns/src/nameserver.rs
Damos-dev: I have used DnsResult as suggested.

crates/telio-dns/src/error.rs
Damos-dev: Way of handling io errors has been redesigned.

crates/telio-dns/src/error.rs
Damos-dev: Helpers removed due to a different approach.

crates/telio-dns/src/nameserver.rs
Damos-dev: Done.

crates/telio-dns/src/zone.rs
Damos-dev: Used DnsResult. Aliasing was required.

crates/telio-dns/src/error.rs
Damos-dev: Due to redesigned way of handling io errors - such complex names are no longer exists.

.unreleased/LLT-7085
Damos-dev: done

crates/telio-dns/src/forward.rs
```diff
@@ -31,7 +31,7 @@ use hickory_server::{
 use telio_utils::{telio_log_debug, telio_log_info, telio_log_trace, telio_log_warn};
 use tokio::net::UdpSocket;
 
-use crate::bind_tun;
+use crate::{bind_tun, error::Result as DnsResult};
```
mathiaspeters, line 34: Is there a reason this is aliased to `DnsResult`?

crates/telio-dns/src/zone.rs
```diff
@@ -12,19 +9,17 @@ use hickory_server::{
...
     str::FromStr,
 };
 use telio_model::features::TtlValue;
-use telio_utils::{telio_log_debug, telio_log_warn};
-use thiserror::Error;
+use telio_utils::telio_log_warn;
 
-pub(crate) const NORD_ZONE: &str = "nord.";
-pub(crate) const NORD_ZONE_SUFFIX: &str = ".nord.";
+use crate::error::{Error, Result as TelioResult};
```
mathiaspeters: Is there a reason this is aliased to `TelioResult`? And is there a reason it's aliased differently than in `forward.rs` where it was aliased to `DnsResult`?

crates/telio-dns/src/nameserver.rs
```diff
@@ -1,8 +1,7 @@
+use crate::error::Result as TelioResult;
```
mathiaspeters: Same question about aliasing

mathiaspeters (commented):
Looks good in general, just one question asked at three different places

Damos-dev:
> Nice, this is taking shape.
> 
> Please update the PR name

Done

crates/telio-dns/src/zone.rs
Damos-dev: - [x] change to DnsResult

crates/telio-dns/src/nameserver.rs
Damos-dev: - [x] change to DnsResult

crates/telio-dns/src/forward.rs
Damos-dev: yes - in the same scope we have two different definitions of Result:
`std::io::Result` and `crate::error::Result` both called Result, but represent different types (with different error types inside).
Because of that compiler don't know which Result should be used and we have a name conflict.

In theory we can use in all places `error::Result` but aliasing looks just better, cleaner and more OK with "Rust spirit".

mathiaspeters: I looked through the imports and couldn't see any other `Result`, but maybe I missed it. Either way it's fine, I was mostly curious

mathiaspeters:
+1

crates/telio-dns/src/forward.rs
```diff
@@ -141,7 +141,7 @@ impl ForwardAuthority {
         origin: Name,
         _zone_type: ZoneType,
         config: ForwardConfig,
-    ) -> Result<Self, String> {
+    ) -> DnsResult<Self> {
```
tomasz-grz: I don't think there is a need for alias here.
An alias would be useful only if you have only one error type in the whole module and don't want to repeatedly type `Result<T, E>`, so you replace it with `Result<T>`. 

```suggestion
    ) -> Result<Self, Error> {
```

In fact, looks like `try_from_config` is not returning any errors at all..

crates/telio-dns/src/error.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl fmt::Display for DnsIoContext {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::BindDnsSocket => write!(f, "Failed to bind DNS socket"),
+            Self::GetSocketAddr => write!(f, "Failed to get DNS socket local address"),
+            Self::ConfigureTunnel => write!(f, "Failed to configure tunnel binding"),
+        }
+    }
+}
```
tomasz-grz: Why don't we just use `#[derive(Error)]`?

```rust
pub enum DnsIoContext {
    /// Binding the DNS UDP socket
    #[error("Failed to bind DNS socket")]
    BindDnsSocket,
    etc...
}
```

Like we do else where?

crates/telio-dns/src/error.rs
```diff
@@ -0,0 +1,148 @@
...
+    source: io::Error,
+}
+
+// traits to create error from io::Error with context
+impl DnsIoError {
+    /// Create a new `DnsIoError` for the given context
+    pub fn new(context: DnsIoContext, source: io::Error) -> Self {
+        Self { context, source }
+    }
+
```
tomasz-grz: I still find it somehow strange to have "constructors" for error enum types in rust..

Maybe we could get rid of one layer and have something like:
```rust
#[derive(Debug, Error)]
pub enum DnsIoError {
    /// Binding the DNS UDP socket
    #[error("Failed to bind DNS socket: {0}")]
    BindDnsSocket(io::Error),
    /// Getting the local address of the DNS socket
    #[error("Failed to get DNS socket local address: {0}")]
    GetSocketAddr(io::Error),
    /// Configuring tunnel binding for DNS
    #[error("Failed to configure tunnel binding: {0}")]
    ConfigureTunnel(io::Error),
}
```

Then you can create the error types like this, without the need for any constructors or helpers:
```rust
        let socket = UdpSocket::bind(SocketAddr::from(([127, 0, 0, 1], 0)))
            .await
            .map_err(DnsIoError::BindDnsSocket)?;

        let socket_port = socket
            .local_addr()
            .map_err(DnsIoError::GetSocketAddr)?
            .port();

        bind_tun::set_tun(tun).map_err(DnsIoError::ConfigureTunnel)?;
```

`.map_err(DnsIoError::BindDnsSocket)` is short hand for `.map_err(|e| {DnsIoError::BindDnsSocket(e)})`

crates/telio-dns/src/forward.rs
Damos-dev: Changes applied.

crates/telio-dns/src/zone.rs
Damos-dev: and no `std::result::Result used anymore`

crates/telio-dns/src/zone.rs
```diff
@@ -192,15 +193,11 @@ pub(crate) struct AuthoritativeZone {
...
-        name: &str,
-        records: &Records,
-        ttl_value: TtlValue,
-    ) -> Result<Self, String> {
+    pub(crate) async fn new(name: &str, records: &Records, ttl_value: TtlValue) -> DnsResult<Self> {
         // TODO: rewrite code so that this assert is not needed.
         for domain in records.keys() {
             if !domain.contains(name) {
-                return Err(format!("{domain} does not end with {name}"));
+                return Err(Error::record_outside_zone(domain, name));
```
tomasz-grz: Why not just `Error::RecordOutsideZone(domain, name)`?

tomasz-grz (commented):
Small nit, otherwise looking good 👍

crates/telio-dns/src/zone.rs
Damos-dev: Good point! 
Suggested approach is already used in the test. Then we can remove whole helper 'pub fn record_outside_zone(domain: impl Into<String>, zone: impl Into<String>) -> Self {'.

- [x] remove helper and use struct instead

tomasz-grz (approved):
+1

Damos-dev:
I had to manually rebase on top of `main` to go over cargo deny error related to `scc` and `RUSTSEC-2026-0205`. And then trigger merge again.

PR #1738, opened 2026-04-09
LLT-7084: Refactor conftest.py into focused helper modules

gytsto:
### Problem
The conftest.py file grew to over 1200 lines, mixing unrelated concerns such as setup checks, pre-test routines, log collection, and Windows monitoring. This made it difficult to navigate and maintain.

### Solution
Extract four dedicated helper modules from conftest.py:
- helpers_setup_checks.py: environment validation fixtures
- helpers_pretest.py: pre-test preparation logic
- helpers_log_collection.py: test log gathering utilities
- helpers_windows_monitoring.py: Windows VM monitoring helpers

conftest.py now delegates to these modules, reducing it to ~254 lines and giving each concern a clear, single-responsibility home.

- [x] README.md is updated


mathiaspeters:
I think the idea is great and I'll have a look later today, but a first thought: what do you think about organizing these files under something like `nat-lab/tests/conftest` or something similar instead of letting them live directly in `nat-lab/tests`. Might make it easier to know what is part of conftest and what isn't, and it keeps the `tests` dir from becoming even more unmanageable

nat-lab/tests/conftest.py
```diff
@@ -155,683 +116,38 @@ def pytest_make_parametrize_id(config, val):
...
-    not_running = [
-        tag.name for tag, status in SESSION_IS_CONTAINER_RUNNING.items() if not status
-    ]
-    setup_log.debug(
-        "Not running (%s): %s", len(not_running), ", ".join(not_running).lower()
-    )
-
-
 def pytest_runtestloop(session):
+    global SESSION_VM_MARKS, SESSION_IS_CONTAINER_RUNNING
```
mathiaspeters: Is it necessary to mark these as globals? `SESSION_IS_CONTAINER_RUNNING` is only referenced in this function, and both of them are declared at the top of the file

nat-lab/tests/conftest_helpers/windows_monitoring.py
```diff
@@ -0,0 +1,194 @@
...
+from datetime import datetime
+from tests.utils.connection import ConnectionTag
+from tests.utils.connection.docker_connection import container_id
+from tests.utils.connection_util import new_connection_raw, is_running
+from tests.utils.logger import setup_log
+from tests.utils.tcpdump import make_local_tcpdump, make_tcpdump
+from typing import List
+
+
+async def start_tcpdump_processes(
```
mathiaspeters: I'm not sure where this should go, maybe in `log_collection.py` or `pretest.py`, but I don't think it should live in this file since it's not windows-specific

mathiaspeters (commented):
A couple of small comments, but looks good in general

nat-lab/tests/conftest.py
gytsto: fixed

nat-lab/tests/conftest_helpers/windows_monitoring.py
gytsto: moved to pretest

mathiaspeters (dismissed):
+2

mathiaspeters (approved):

PR #1740, opened 2026-04-09
Backport ENS TLS certificate signature verification to v6.2

tomasz-grz:
### Problem
ENS TLS verified certificates but not the signatures. The fix was merged to main, but we want to release a hotfix.

### Solution
Back port https://github.com/NordSecurity/libtelio/pull/1725 builtin signature verifier for both certificate and signature verification. This backport does not include setting the public key for the NLX server in natlab simply because it's not necessary.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-7169
+++ b/.unreleased/LLT-7169
@@ -1,0 +1,1 @@
+Fix: Incorrect certificate validation in ENS service
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5023,6 +5023,7 @@ name = "telio-proto"
 version = "0.1.0"
 dependencies = [
  "anyhow",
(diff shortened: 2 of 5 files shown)


tomaszklak (commented):
This PR is missing public key setting for NLX server in tests: https://github.com/NordSecurity/libtelio/pull/1725/changes#diff-a91e5f9ea38f2b6c0e059f4f4ad014acb21345cc7a344656df63eb5379d51afaR295

If this is by design, I think it would be good idea do mention this in the PR description and clarify the by-design differences from the original (and reasons for them).

Other than that, seems like a verbatim copy, so +1 :)

Jauler (approved):
+1.0

PR #1742, opened 2026-04-14
LLT-6727: natlab command execution times

gytsto:
### Problem
Investigating test failures due to timeouts is difficult since we do not know how long the commands were executed for. In some cases we can interpolate between nearby calls.

### Solution
Add "Finished ... in Xms" log lines after every command execution in NatLab, so that test failure investigation can show how long each command took. Duration is logged at INFO level using `time.monotonic()` for accuracy.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/connection/docker_connection.py
+++ b/nat-lab/tests/utils/connection/docker_connection.py
@@ -181,7 +181,10 @@ class DockerConnection(Connection):
         self, command: List[str], kill_id=None, term_type=None, quiet=False
     ) -> "Process":
         process = DockerProcess(
-            self._container, container_id(self.tag), command, kill_id
+            self._container,
+            container_id(self.tag),
+            command,
+            kill_id,
         )
 
         if not quiet:
--- a/nat-lab/tests/utils/connection/ssh_connection.py
+++ b/nat-lab/tests/utils/connection/ssh_connection.py
@@ -158,7 +158,11 @@ class SshConnection(Connection):
             assert False, f"not supported target_os '{self.target_os}'"
 
         return SshProcess(
-            self._connection, self.tag.name, command, escape_argument, term_type
+            self._connection,
+            self.tag.name,
+            command,
+            escape_argument,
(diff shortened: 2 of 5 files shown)


nat-lab/tests/utils/connection/docker_connection.py
```diff
@@ -179,7 +179,11 @@ def create_process(
         self, command: List[str], kill_id=None, term_type=None, quiet=False
     ) -> "Process":
         process = DockerProcess(
-            self._container, container_id(self.tag), command, kill_id
+            self._container,
+            container_id(self.tag),
+            command,
+            kill_id,
+            tag_name=self.tag.name,
```
LukasPukenis: nit: can we pass `self.tag` instead of `container_id(self.tag)` and `tag_name=self.tag.name`?

nat-lab/tests/utils/connection/ssh_connection.py
```diff
@@ -157,7 +157,12 @@ def create_process(
...
 
         return SshProcess(
-            self._connection, self.tag.name, command, escape_argument, term_type
+            self._connection,
+            self.tag.name,
+            command,
+            escape_argument,
+            term_type,
+            tag_name=self.tag.name,
         )
```
LukasPukenis, line 166: Do we need to pass `self.tag.name` twice?

LukasPukenis (commented):
+1

Just tiny nits regarding `tag_name`.

EDIT: +2

nat-lab/tests/utils/connection/docker_connection.py
gytsto: removed

nat-lab/tests/utils/connection/ssh_connection.py
gytsto: removed

LukasPukenis (approved):

jjanowsk (approved):
+1

PR #1744, opened 2026-04-14
[LLT-7222] Fix LuCI TypeError in openwrt nordvpnlite

olekoliinyk:
### Problem
When clicking Save in the NordVPN Lite settings page, a `TypeError: Section ID` required error was thrown from LuCI's `form.js`.
This happened because `isValid()` was called on a form option without passing the required `section_id` argument, which the LuCI framework requires to locate the corresponding UI element in the DOM.

### Solution
Pass the section name 'config' to `isValid()`, consistent with how formvalue() is already called in the same handler.

- [x] README.md is updated


gytsto (approved):
+2

djkarwowski (approved):
+1

PR #1748, opened 2026-04-17
LLT-6224: test: replace wg adapter cleanup test with looped clean/dirty shutdown variants

gytsto:
### Problem
On Windows, we’re sometimes getting errors during interface setup, but currently do not have good reproduction of this situation. It would be beneficial to create a test, that creates and destroys interface multiple times and in case of failure, catches various OS logs.

### Solution
Replace single-iteration test_wg_adapter_cleanup with two looped tests that run 5 iterations each to detect stability issues:

- test_wg_adapter_cleanup_loop_clean_shutdown: verifies repeated clean start/stop cycles properly remove WireGuard adapters
- test_wg_adapter_cleanup_loop_dirty_shutdown: verifies repeated dirty shutdowns (taskkill) followed by restarts properly detect and clean up orphaned adapters

Extract QUERY_CMD as module-level constant to reduce duplication. Add pytest.mark.timeout(240) to both tests.

- [x] README.md is updated


nat-lab/tests/test_wg_adapter.py
```diff
@@ -2,26 +2,30 @@
...
 from tests.utils.bindings import TelioAdapterType
 from tests.utils.connection import ConnectionTag
 from tests.utils.connection_util import new_connection_by_tag
 from tests.utils.logger import log
 from tests.utils.process import ProcessExecError
 
+QUERY_CMD = [
+    "reg",
+    "query",
+    r"HKLM\SYSTEM\CurrentControlSet\Control\Class\{4d36e972-e325-11ce-bfc1-08002be10318}",
```
tomaszklak: I think there are 3 places where this path/uuid is copied in the nat-lab tests. Could you make a one common constant somewhere and use it in place of the copies. Also, pleas add a comment to that constant with link to the documentation (msdn?) that explains what this path/uuid is.

tomaszklak (commented):
one small comment but apart from that +1 - I really like the dirty test! ❤️

nat-lab/tests/test_wg_adapter.py
```diff
@@ -2,26 +2,30 @@
...
-        "query",
-        r"HKLM\SYSTEM\CurrentControlSet\Control\Class\{4d36e972-e325-11ce-bfc1-08002be10318}",
-        "/s",
-        "/f",
-        "DeviceInstanceID",
-    ]
-    # Run libtelio and kill it dirty so it would leave hanging wg-nt adapter
-    try:
+async def test_wg_adapter_cleanup_loop_clean_shutdown(conn_tag: ConnectionTag):
+    for i in range(5):
```
tomasz-grz: Do we need 5 times, or is 2 consecutive runs is enough?

Can this be a "const"? or even better, some configurable test flag?

nat-lab/tests/test_wg_adapter.py
gytsto: addressed

nat-lab/tests/test_wg_adapter.py
gytsto: this was intentionally made in loop of 5, to "stress test" abit. Moved interation count to pytest parameters tho

tomaszklak (commented):
+1

tomasz-grz (approved):
+1

PR #1751, opened 2026-04-20
Bump rust nightly version in fuzzing jobs

mathiaspeters:
### Problem
One of our fuzzing jobs fails because a dependency is incompatible with rust 1.91.0 which is what the nightly version we use in the fuzzing jobs corresponds to

### Solution
Bump rust nightly to a version that corresponds to rust 1.95.0, which is compatible with that dependency

- [x] README.md is updated


sfraczek (approved):
+1

olekoliinyk:
+1.0

olekoliinyk (approved):

PR #1755, opened 2026-04-21
Fix CI for release/v6.2

mathiaspeters:
### Problem
"Vanilla v6.2.1" was no longer building due to a couple of issues

### Solution
- Bump nightly rust version in fuzzing jobs
- Ignore pylint warning
- Bump aws-lc-rs and for that to build on all platforms, bump libtelio-build

- [x] README.md is updated


crates/telio-proto/Cargo.toml
```diff
@@ -17,7 +17,7 @@ prost = "0.14.1"
 protobuf = "3.7.2"
 rustls = { version = "0.23", default-features = false, features = ["std"] }
 rustls-webpki = { version = "0.103.4", optional = true }
-aws-lc-rs = { version = "=1.13.3", features = ["bindgen"], optional = true }
+aws-lc-rs = { version = "=1.15.0", features = ["bindgen"], optional = true }
```
tomaszklak: This is not the latest release in 1.15.x family, but also there are already 1.16.x releases. Can you bump (without too much work) to the latest 1.15 or better yet 1.16 ?

tomaszklak:
I think it would be nice to mention in the commit message, to which released version nightly-2025-05-09 maps to (1.87.0?).

Also, I think it would be good idea to explain clearly (in aws-lc-rs update commit and fuzzing fixing commit) what problems are those commits fixing.

crates/telio-proto/Cargo.toml
mathiaspeters: I was aiming to keep the update as minimal as possible, which is why I only bumped it to `1.15.0`, it is the first version that builds successfully. But `1.16.3` also works

mathiaspeters: Updated

mathiaspeters:
@tomaszklak I added some explanations to those commits

tomaszklak (approved):
+2

PR #1763, opened 2026-04-30
test: reduce WG adapter cleanup iterations and increase timeout

gytsto:
### Problem
The test assumes that every dirty shutdown will leave an orphaned WireGuard adapter, but in practice the adapter is cleaned up by the WireGuard-NT driver itself or by the Windows PnP system before the Phase 2 check occurs. The orphaned adapter only persists after multiple rapid create-kill cycles accumulate stale device state.

Combined with the slow execution speed of Windows VM commands (especially in CI), the 240-second timeout is insufficient for 5 iterations where only the last one triggers the actual test logic.

### Solution
Lower iteration count from 5 to 3 in both clean and dirty shutdown loop tests to reduce flakiness, and raise TEST_WG_ADAPTER_CLEANUP_TIMEOUT from 240s to 300s to give the reduced iterations enough headroom.

- [x] README.md is updated


asledzinskiy (approved):
+2

PR #1765, opened 2026-05-05
LLT-7053: Integrate DNS forwarder

tomasz-grz:
### Problem

Non `.nord` DNS queries are forwarded through hickory-server's ForwardAuthority zone. We want to remove  `hickory` dependencies as it adds unnecessary overhead, requires additional maintenance and was a source of bugs.

### Solution

Integrate `RawForwarder` into `LocalNameServer` behind a new `use_raw_forwarder` feature flag.

- Add `use_raw_forwarder: Option<bool>` to `FeatureDns` (default: None, uses hickory-server)
- When `use_raw_forwarder` is true, `LocalNameServer` routes non `.nord` queries through `RawForwarder` instead of hickory `Resolver` and `ClonableZones::lookup()`.
- The hickory-server path remains the default
- Parametrize nat-lab DNS tests to run with both hickory and raw forwarder

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/src/nameserver.rs
+++ b/crates/telio-dns/src/nameserver.rs
@@ -1,5 +1,6 @@
 use crate::{
-    packet_decoder::{find_nord_query, normalize_qname, parse_dns_query_packet, DnsParseError},
+    forwarder::RawForwarder,
+    packet_decoder::{find_nord_query, normalize_qname, parse_dns_query_packet},
     packet_encoder::DnsResponseBuilder,
     resolver::Resolver,
     zone::{AuthoritativeZone, ClonableZones, ForwardZone, NordZone, Records, NORD_ZONE},
@@ -12,7 +13,7 @@ use hickory_server::{
 };
 use neptun::noise::{Tunn, TunnResult};
 use pnet_packet::{
-    dns::DnsQuery,
+    dns::{DnsPacket, DnsQuery},
     ip::{IpNextHeaderProtocol, IpNextHeaderProtocols},
     ipv4::{checksum, Ipv4Packet, MutableIpv4Packet},
     ipv6::{Ipv6Packet, MutableIpv6Packet},
@@ -40,6 +41,7 @@ const UDP_HEADER: usize = 8;
 const TCP_MIN_HEADER: usize = 20;
 const MAX_CONCURRENT_QUERIES: usize = 256;
 const IDLE_TIME: Duration = Duration::from_secs(1);
+const DNS_PORT: u16 = 53;
 
 /// NameServer is a server that stores the DNS records.
 #[async_trait]
@@ -53,6 +55,8 @@ pub trait NameServer {
     async fn stop(&self);
     /// Configure list of forward DNS servers for zone '.'.
     async fn forward(&self, to: &[IpAddr]) -> Result<(), String>;
+    /// Configure list of forward DNS servers with explicit socket addresses.
+    async fn forward_to_addrs(&self, to: &[SocketAddr]) -> Result<(), String>;
     /// Insert or update zone records used by the server.
     async fn upsert(
         &self,
@@ -94,16 +98,30 @@ pub struct LocalNameServer {
     nord_zone: NordZone,
     zones: Arc<ClonableZones>,
     task_handle: Option<JoinHandle<()>>,
+    forwarder: Option<RawForwarder>,
 }
 
 impl LocalNameServer {
     /// Create a new `LocalNameServer` with forwarding dns servers from `forward_ips`
     /// configured for zone `.`.
-    pub async fn new(forward_ips: &[IpAddr]) -> Result<Arc<RwLock<Self>>, String> {
+    pub async fn new(
+        forward_ips: &[IpAddr],
+        use_raw_forwarder: bool,
+    ) -> Result<Arc<RwLock<Self>>, String> {
+        let raw_forwarder: Option<RawForwarder> = if use_raw_forwarder {
+            Some(
+                RawForwarder::new()
+                    .await
+                    .map_err(|e| format!("Failed to create forwarder: {e:?}"))?,
+            )
+        } else {
+            None
+        };
         let ns = Arc::new(RwLock::new(LocalNameServer {
             nord_zone: NordZone::new(),
             zones: Arc::new(ClonableZones::new()),
             task_handle: None,
+            forwarder: raw_forwarder,
         }));
         ns.forward(forward_ips).await?;
         Ok(ns)
@@ -291,23 +309,69 @@ impl LocalNameServer {
                     .build()
                     .map_err(|e| format!("Failed building DNS response packet: {e:?}"))?
             }
-            PayloadDestination::Forward(question) => {
-                telio_log_debug!("Forwarding request dns");
-                let resolver = Resolver::new();
-                telio_log_debug!("Getting DNS zones");
-                let zones = nameserver.zones().await;
+            PayloadDestination::Forward(raw_query) => {
+                let raw_forwarder = {
+                    let ns = nameserver.read().await;
+                    ns.forwarder.clone()
+                };
+                if let Some(forwarder) = raw_forwarder {
+                    telio_log_debug!(
+                        "Forwarding raw DNS request from port {:?}",
+                        request_info.dns_source_port(),
+                    );
+                    // TODO: LLT-7054: Remove once migration is fully validated
+                    if let Some(dns_packet) = DnsPacket::new(&raw_query) {
+                        telio_log_debug!(
+                            "Request: id: {} is_response: {} opcode: {:?} recursion_desired: {} query_count: {} rcode: {:?}",
+                            dns_packet.get_id(),
+                            dns_packet.get_is_response(),
+                            dns_packet.get_opcode(),
+                            dns_packet.get_is_recursion_desirable(),
+                            dns_packet.get_query_count(),
+                            dns_packet.get_rcode(),
+                        );
+                        for query in dns_packet.get_queries() {
+                            telio_log_debug!(
+                                "  qtype: {} qclass: {} qname: {}",
+                                query.qtype,
+                                query.qclass,
+                                query.get_qname_parsed()
+                            );
+                        }
+                    }
 
-                let dns_request = Request::new(question, request_info.dns_source(), Protocol::Udp);
-                telio_log_debug!("DNS request: {:?}", &dns_request);
+                    let response = forwarder
+                        .query(&raw_query)
+                        .await
+                        .map_err(|e| format!("Forward failed: {e:?}"))?;
 
-                zones
-                    .lookup(&dns_request, resolver.clone())
-                    .await
-                    .map_err(|e| format!("Lookup failed {e}"))?;
+                    telio_log_debug!("Forwarder responded");
+                    if let Some(dns_packet) = DnsPacket::new(&response) {
+                        telio_log_debug!("Response: {:#?}", dns_packet);
+                    }
 
-                let dns_response = resolver.0.lock().await;
-                telio_log_debug!("Nameserver response: {:?}", &dns_response);
-                dns_response.to_vec()
+                    response
+                } else {
+                    telio_log_debug!("Forwarding DNS request (hickory-server)");
+                    let resolver = Resolver::new();
+                    let zones = nameserver.zones().await;
+
+                    let forward = MessageRequest::from_bytes(&raw_query).map_err(|_| {
+                        String::from("Failed to build MessageRequest from request packet")
+                    })?;
+                    let dns_request =
+                        Request::new(forward, request_info.dns_source(), Protocol::Udp);
+                    telio_log_debug!("DNS request: {:?}", &dns_request);
+
+                    zones
+                        .lookup(&dns_request, resolver.clone())
+                        .await
+                        .map_err(|e| format!("Lookup failed {e}"))?;
+
+                    let dns_response = resolver.0.lock().await;
+                    telio_log_debug!("Nameserver response: {:?}", &dns_response);
+                    dns_response.to_vec()
+                }
             }
         };
 
@@ -382,7 +446,7 @@ impl LocalNameServer {
         }
 
         // Validate DNS port
-        if udp_request.get_destination() != 53 {
+        if udp_request.get_destination() != DNS_PORT {
             return Err(String::from("Invalid DNS port"));
         }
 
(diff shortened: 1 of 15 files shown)


gytsto (commented):
lgtm +1

nat-lab/tests/test_dns.py
```diff
@@ -28,6 +28,11 @@
 from tests.utils.router import IPStack
 from typing import List, Optional
 
+DNS_FORWARDER_PARAMS = [
```
LukasPukenis, line 31: This PR might introduce additional wor in NatLab due to parametrization. I ran:
```
TELIO_BIN_PROFILE=debug uv run python3 run_local.py -k test_dns -v --notypecheck --nobuild
```
and the results for this branch are:
```
collected 625 items / 587 deselected / 38 selected
```
versus `main` branch:
```
collected 654 items / 584 deselected / 70 selected
```

I wonder what is the additional time needed to run all of these tests?

crates/telio-dns/tests/nameserver.rs
```diff
@@ -1231,3 +1287,178 @@ async fn test_timers_updated() {
...
+
+    let nameserver = LocalNameServer::new(&[], USE_RAW_FORWARDER)
+        .await
+        .expect("Failed to create a LocalNameServer");
+    nameserver
+        .forward_to_addrs(&[stub_addr])
+        .await
+        .expect("Failed to set stub upstream");
+
+    let response = timeout(
```
LukasPukenis, line 1370: nit: can we lower timeout values in all of the tests from 60 down to say 5?

crates/telio-dns/src/nameserver.rs
```diff
@@ -53,6 +55,8 @@ pub trait NameServer {
     async fn stop(&self);
     /// Configure list of forward DNS servers for zone '.'.
     async fn forward(&self, to: &[IpAddr]) -> Result<(), String>;
+    /// Configure list of forward DNS servers with explicit socket addresses.
+    async fn forward_to_addrs(&self, to: &[SocketAddr]) -> Result<(), String>;
```
LukasPukenis, line 59: nit: i found it a bit difficult to understand the relationship between `forward_to_addrs` and `forward`. I suggest to rename `forward_to_addrs` to `set_upstream`.

Additionally, I do not see the reason for `forward_to_addrs` to be part of the trait, I would make it a private function instead unless I am missing something.

crates/telio-dns/src/nameserver.rs
```diff
@@ -94,16 +98,24 @@ pub struct LocalNameServer {
     nord_zone: NordZone,
     zones: Arc<ClonableZones>,
     task_handle: Option<JoinHandle<()>>,
+    /// Use the raw DNS forwarder instead of hickory-server
+    use_raw_forwarder: bool,
+    forwarder: Option<RawForwarder>,
```
LukasPukenis: It looks like `use_raw_forwarder` can be eliminated since `forwarder` is an `Option` already.

In `new()` based on the `use_raw_forwarder` value the `forwarder` can be set to Some/None and would convey they same information.

crates/telio-dns/tests/nameserver.rs
```diff
@@ -599,6 +602,56 @@ async fn init_client_and_server(
...
+    buf[6..8].copy_from_slice(&1_u16.to_be_bytes());
+
+    buf.extend_from_slice(&[0xC0, 0x0C]);
+    buf.extend_from_slice(&1_u16.to_be_bytes());
+    buf.extend_from_slice(&1_u16.to_be_bytes());
+    buf.extend_from_slice(&300_u32.to_be_bytes());
+    buf.extend_from_slice(&4_u16.to_be_bytes());
+    buf.extend_from_slice(&answer_ip.octets());
+
+    buf
```
LukasPukenis, line 645: nit: maybe we can use some create which is available to us to craft the packet?

crates/telio-dns/tests/nameserver.rs
```diff
@@ -1231,3 +1287,178 @@ async fn test_timers_updated() {
...
+    .await
+    .expect("Test timeout")
+    .expect("Expected some DNS response");
+
+    assert!(response.no_error());
+    assert_eq!(response.a_addrs(), vec![expected_ip]);
+}
+
+#[tokio::test]
+async fn dns_request_forward_timeout_returns_no_response() {
```
LukasPukenis: I was looking where the timeout is enforced and it seems that it's deeper in `send_dns_request()`. I suggest to remove the timeout from `send_dns_request` and instead add timeout in this test in which case it's very explicit.

crates/telio-dns/tests/nameserver.rs
```diff
@@ -36,6 +37,7 @@ const UDP_HEADER: usize = 8;
 const TCP_MIN_HEADER: usize = 20;
 const ICMP_HEADER: usize = 8;
 const MAX_PACKET: usize = 2048;
+const USE_RAW_FORWARDER: bool = true;
```
LukasPukenis, line 41: Can we instead of having `USE_RAW_FORWARDER` (defined in two places) parametrize the parameter for `use_raw_forwarder` with `rstest`?

crates/telio-dns/src/nameserver.rs
tomasz-grz: `forward()` takes an  IP address only and is used by the old hickory server, where as `forward_to_addrs()` takes a Socket Address (IP + port number) used by the new raw forwarder.
`forward()` hard codes the port number to 53 for production and calls `forward_to_addrs()` internally.

It's mainly to be able to mock multiple forwarders for intermediate integration tests locally (otherwise they would clash on port 53).

Also if it was a private function we would not be able to use in `tests/nameserver.rs` tests.

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Some of the tests would only work with the RawForwarder, but good idea to parameterize them

crates/telio-dns/src/nameserver.rs
tomasz-grz: Good point, I added `use_raw_forwarder` only after everything else was already in place.

tomasz-grz: Updated

nat-lab/tests/test_dns.py
tomasz-grz: Yes we possibly don't need to test the forwarder for meshnet cases

crates/telio-dns/src/nameserver.rs
LukasPukenis: Sounds good :)

nat-lab/tests/test_dns.py
LukasPukenis: If it for some reason doesn't add a lot of additional time, it's good, but I would expect at least at minimum like 30seconds per each testcase so minimizing would be nice :)

crates/telio-dns/tests/nameserver.rs
LukasPukenis: Raw forwarder specific tests could use an explicit `true` passed to them if there are few of them. Existing tests tested hickory forwarder, so always passing `true` means that hickory is not tested anymore but we still support it so tests should still ensure it works.

I haven't dig too deep though so consider this with a pinch of salt :)

LukasPukenis:
+1

nat-lab/tests/test_dns.py
tomasz-grz: They would eventually go away once we remove hickory, but good point, I'll try to compare how this impacts test times

crates/telio-dns/tests/nameserver.rs
tomasz-grz: There is a forwarder timeout, that is handled internally by `RawForwarder`.
`UpstreamStubBehavior::BlackHole` (aka don't reply) specifies the behavior of the stub upstream server. So the `RawForwarder` will timeout and try the next upstream, until it runs out of upstreams.

There is an additional timeout for receiving a response in the test (the one in `send_dns_request`), which one do you mean?
The expected test result is specified by `DnsTestType::NonRespondingForwardServer`, ensuring that we don't receive any `Ok()` result when a test is not expecting one (`DnsTestType::is_correct()`).

Removing the `DnsTestType` approach would probably mean refactoring the whole test framework. How would you recommend doing it instead?

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Reduced it to 5.

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Although `dns_parser` it claims it supports building answers, there are absolutely no methods for answers in it's Builder..

`pnet_packet` also works with raw byte slices and requires similar level of manual construction. Bringing in another dependency is probably not worth it for a quick test helper.

I documented the crafting a little bit, is it better now?

crates/telio-dns/tests/nameserver.rs
tomasz-grz: Added `rstest` parametrization to the tests that use the forwarding paths.

The Raw forwarder specific and rest of the local tests still use `USE_RAW_FORWARDER = true` since hickory was already tested and will go away.

nat-lab/tests/test_dns.py
tomasz-grz: Reduced the number of parametrized tests in `test_dns.py`.

But added some to `test_dns_through_exit.py`  🙈

crates/telio-dns/tests/nameserver.rs
LukasPukenis: I assumed the "timeout" part in the function name is referring to timeout which is enforced in `send_dns_request()`, but it looks like a wrong assumption!

nat-lab/tests/test_dns.py
```diff
@@ -602,6 +629,7 @@ async def test_set_meshnet_config_dns_update(
                             ConnectionTag.DOCKER_CONE_CLIENT_1
                         ),
                         derp_servers=[],
+                        features=_dns_features(False),
```
mathiaspeters: In the places where `_dns_features` is called with literals, how about we use name arguments? `_dns_features(use_raw_forwarder=False)` is a lot clearer in my opinion

nat-lab/tests/test_dns_through_exit.py
```diff
@@ -10,9 +10,21 @@
...
+    pytest.param(False, id="hickory"),
+    pytest.param(True, id="raw"),
+]
+
+
+def _dns_features(use_raw_forwarder: bool, **kwargs):
+    features = default_features(**kwargs)
+    features.dns.use_raw_forwarder = use_raw_forwarder
+    return features
+
```
mathiaspeters: Is it possible to import these from `test_dns.py` instead of redefining?

crates/telio-dns/src/nameserver.rs
```diff
@@ -291,23 +309,68 @@ impl LocalNameServer {
...
+                let raw_forwarder = {
+                    let ns = nameserver.read().await;
+                    ns.forwarder.clone()
+                };
+                if let Some(forwarder) = raw_forwarder {
+                    telio_log_debug!(
+                        "Forwarding raw DNS request from port {:?}",
+                        request_info.dns_source_port(),
+                    );
+                    if let Some(dns_packet) = DnsPacket::new(&raw_query) {
```
mathiaspeters, line 323: This block feels like something that should maybe not be here. I'd be okay with it if it was done like with the response, to log the whole packet at once, but unpacking the whole thing like this feels a bit noisy

tomasz-grz: Yeah I was thinking about this again, it's very verbose (and sensitive) logging..
 
I think there might be a bug with `pnet_packet` and DnsPacket (or DnsQuery) panics whenever I try to log the whole thing with just `{:#?}`. Also it doesn't resolve the `qname`s.

I could remove it completely now, or add a `TODO` note to remove after validation (together with hickory cleanup), I think it's mostly useful for debugging. What do you think?

nat-lab/tests/test_dns_through_exit.py
tomasz-grz: Good point, I'll look into it

nat-lab/tests/test_dns.py
tomasz-grz: Good idea!

crates/telio-dns/src/nameserver.rs
tomasz-grz: Yes that panicked again, on debug print of `DnsPacket`..

It's only doing it on natlab test runs or locally when using `dig`, but not in integration tests.
```
telio::ffi:236 panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pnet_packet-0.35.0/src/dns.rs:325:54:
range start index 3136 out of range for slice of length 75
```

Packets otherwise seem to be passing all the tests, and appear correct when "unpacking" like before.. I wonder if this requires more investigation 🤔 I checked it briefly before and it seemed like buggy debug print in `pnet_packet`.

nat-lab/tests/test_dns.py
tomasz-grz: Updated

nat-lab/tests/test_dns_through_exit.py
tomasz-grz: Updated

nat-lab/tests/test_dns_through_exit.py
```diff
@@ -16,12 +17,6 @@
 ]
```
mathiaspeters: Can we import this too from `test_dns` ? 😄

crates/telio-dns/src/nameserver.rs
```diff
@@ -291,23 +309,68 @@ impl LocalNameServer {
...
+                        .await
+                        .map_err(|e| format!("Forward failed: {e:?}"))?;
 
-                zones
-                    .lookup(&dns_request, resolver.clone())
-                    .await
-                    .map_err(|e| format!("Lookup failed {e}"))?;
+                    telio_log_debug!("Forwarder responded");
+                    if let Some(dns_packet) = DnsPacket::new(&response) {
+                        telio_log_debug!("Response: {:#?}", dns_packet);
```
mathiaspeters, line 350: Is this log not also crashing then? 🤔

crates/telio-dns/src/nameserver.rs
mathiaspeters: I would vote for removing these logs completely, but I can accept them being here if we make sure to remove them in LLT-7054

crates/telio-dns/src/nameserver.rs
tomasz-grz: Nope 🤷

mathiaspeters (approved):
+1

PR #1770, opened 2026-05-15
Bump version in Cargo and add changelog

mathiaspeters:
### Problem
The changelog was manually updated on the release page and the version in `Cargo.toml` was never updated

### Solution
Update the changelog to add section for `v7.0.0` and set the version in `Cargo.toml` to the "next" version, in this case `v7.1.0`

- [x] README.md is updated


changelog.md
```diff
@@ -1,5 +1,46 @@
 <!-- Note: this file is auto-generated. See CONTRIBUTING.md for details. -->
 
+### v7.0.0
+### ****
```
lcruz99, line 4: codename missing?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "telio"
-version = "6.1.0"
+version = "7.1.0"
```
Jauler, line 3: it probably should be `v7.0.1` and not `v7.1.0`

mathiaspeters: I disagree. Patch releases usually don't come out of `main` because there are too many commits, instead we usually create a release branch and cherry pick. So updating the patch version here would mean that we would have to update this version before we make the next release from `main`, which in my opinion negates the purpose of bumping this after a release in the first place

changelog.md
mathiaspeters: If you have a suggestion please give it. As far as I understand code names are not necessary and are usually based on what the releaser ate for lunch that day and I haven't had lunch yet :smile:

Damos-dev: Then add codename: 
`** No Lunch ** `
😄

lcruz99:
+1

Cargo.toml
Jauler: Ah, ok, I was thinking for some reason, that this is preparation for v7.0.1 release

Jauler:
+1.0

Jauler (approved):

PR #1771, opened 2026-05-15
Do not run setup nlx server if no nlx container is running

lcruz99:
### Problem
`nlx-01` container is part of the speciality containers and should not be required to be running by default.

At the moment natlab runs will fail setup check if this container is not running which is often the case when run locally (most of the tests don't require it).

### Solution
Add check to `setup_nlx_vpn_server()` to skip if the container is not running.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest_helpers/setup_checks.py
+++ b/nat-lab/tests/conftest_helpers/setup_checks.py
@@ -306,8 +306,14 @@ async def setup_check_arp_cache(session_vm_marks: set[str]):
         raise Exception("ARP cache not ready for VMs: " + ", ".join(failures))
 
 
-async def setup_nlx_vpn_server() -> None:
+async def setup_nlx_vpn_server(
+    session_is_container_running: dict[ConnectionTag, bool],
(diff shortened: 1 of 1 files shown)


nat-lab/tests/conftest_helpers/setup_checks.py
```diff
@@ -382,6 +388,10 @@ async def perform_setup_checks(
                     await asyncio.wait_for(
                         asyncio.shield(target(session_vm_marks)), timeout
                     )
+                elif target is setup_nlx_vpn_server:
+                    await asyncio.wait_for(
+                        asyncio.shield(target(session_is_container_running)), timeout
+                    )
```
mathiaspeters: What do you think about adding this as an additional condition to line 383 instead? Since the body of that if branch and this elif branch do the same thing

lcruz99: It also works, will change

mathiaspeters (approved):
+2

PR #1772, opened 2026-05-15
LLT-7073: Bump NepTUN version

djkarwowski:
### Problem
Current NepTUN version shows degraded performance on Apple platforms.

### Solution
Bump NepTUN to version v2.3.1, which provides the fix to speed degradation issue on Apple platforms. 

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-7073
+++ b/.unreleased/LLT-7073
@@ -1,0 +1,1 @@
+Bump NepTUN to v2.3.1 to fix speed degradation on Apple platforms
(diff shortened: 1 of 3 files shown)


mathiaspeters (commented):
Looking at the commit history, I think it would have been better to tag the NepTUN commit v2.4.0 since there's quite a few changes in there that at least I would consider bigger than a patch release. But more of a "for next time" than a blocker. 

+1

lcruz99 (dismissed):
+1

olekoliinyk (dismissed):

olekoliinyk (approved):

PR #1773, opened 2026-05-18
LLT-7243: collect Windows crash dumps in nat-lab

gytsto:
## Problem

Windows nat-lab tests don't capture crash dumps when a process inside the Windows VM (e.g. tcli, daemon) crashes. The result is a test failure with no post-mortem artifact, forcing manual re-runs with added instrumentation.

## Solution

Collect WER local dumps off the Windows VM on test teardown.

## Changes

- **`nat-lab/tests/log_collector.py`** — pulls `C:\CrashDumps\*.dmp` from the Windows VM via qemu-guest-agent alongside the existing log capture.
- **`nat-lab/tests/telio.py`** — minor wiring so the collector runs for Windows clients.
- **`ci/env.py`** — pins `LIBTELIO_ENV_NAT_LAB_WINDOWS_VM_TAG` to a dockur_windows image digest that has WER crash dump collection enabled.

- [x] README.md is updated


.unreleased/LLT-7243_collect_windows_crash_dumps_in_nat_lab
```diff
@@ -0,0 +1 @@
+Collect Windows Error Reporting crash dumps from the Windows nat-lab VM on test teardown
```
mathiaspeters: Nit: we can leave the changelog empty here since it doesn't concern the apps teams

mathiaspeters (commented):
+1

I'm wondering if we (as a separate ticket) should create a `dump_collector.py` file where we put the dump-related code, since the log collector is becoming a bit bigger with mixed log/dump code

.unreleased/LLT-7243_collect_windows_crash_dumps_in_nat_lab
gytsto: addressed

Jauler (dismissed):
I guess it would be nice to intentionally crash something (e.g. by causing segfault) and checking if dump appears unless its already has been done

But in general LGTM, +1.0

Jauler (approved):

PR #1775, opened 2026-05-19
[LLT-7132] Add Mac VM to performance tests

olekoliinyk:
### Problem
Performance test metrics are missing the nep_tun data which runs on macos vm.

### Solution
1. Parametrize `test_vpn_connection_performance` to run it on linux and macos vms
2. Copy the required binaries
3. Add additional data to metrics: `platform` and `adapter_type`

- [x] README.md is updated


.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v5.15.5
+TRIGGERED_REF=LLT-7132_add_neptun_to_performance_tests
```
olekoliinyk: To be updated

LukasPukenis:
+1

nat-lab/performance_tests/test_vpn_connection_performance.py
```diff
@@ -176,7 +177,29 @@ async def collect_download_metrics(
...
+        ),
+        pytest.param(
+            SetupParameters(
+                connection_tag=ConnectionTag.VM_MAC,
+                adapter_type_override=TelioAdapterType.NEP_TUN,
+                is_meshnet=False,
+                run_tcpdump=False,
+            ),
+            marks=pytest.mark.mac,
+        ),
```
tomasz-grz, line 201: In general I'm a proponent of annotating the test parameters with an [id](https://docs.pytest.org/en/stable/example/parametrize.html#set-marks-or-test-id-for-individual-parametrized-test) to avoid long congested test names.


```suggestion
         pytest.param(
            SetupParameters(
                connection_tag=ConnectionTag.VM_MAC,
                adapter_type_override=TelioAdapterType.NEP_TUN,
                is_meshnet=False,
                run_tcpdump=False,
            ),
            marks=pytest.mark.mac,
            id="macos",
        ),
```

tomasz-grz (commented):
I would say +1 but looks like natlab is failing.
Also I couldn't find in which job this test was run.. is it excluded from regular runs?

nat-lab/performance_tests/conftest.py
```diff
@@ -154,6 +156,8 @@ def pytest_runtestloop(session):
     if not session.config.option.collectonly:
         if os.environ.get("NATLAB_SAVE_LOGS") is not None:
             asyncio.run(collect_kernel_logs(session.items, "before_tests"))
+        session_vm_marks = get_session_vm_marks(session.items)
+        asyncio.run(copy_vm_binaries_if_needed(session_vm_marks))
```
asledzinskiy, line 160: may be you need to rebase as I see this code in `main` branch

asledzinskiy:
+1

nat-lab/performance_tests/conftest.py
olekoliinyk: Hmm, can you point me where do you see it?

nat-lab/performance_tests/test_vpn_connection_performance.py
olekoliinyk: Done

olekoliinyk:
> I would say +1 but looks like natlab is failing.
> Also I couldn't find in which job this test was run.. is it excluded from regular runs?

It's run only for schedules. But here is the pipeline id: 4226915

The natlab was failing on `test_openwrt_vpn_connection[ConfigPresetName.VPN_OPENWRT_UCI_PL-openwrt-24.10-armv8]` which is unrelated to the changes.

tomasz-grz (dismissed):
+1

asledzinskiy (approved):

PR #1782, opened 2026-05-22
nordvpnlite fix wait for start

tomasz-grz:
### Problem

There is a race condition between `TelioContext::new()` and `handle_exit_node_connection()`:
1. `handle_exit_node_connection()` starts making API calls.
2. `TelioContext::new()` calls `set_ip()` -> `/etc/init.d/network reload` and disrupts the network stack.
3. Current in flight API calls can fail.

### Solution

1. Move `handle_exit_node_connection()` to run after `TelioContext::new()` completes.
2. Broaden retry logic (exponential backoff for all API calls).
3. Add additional DNS logs to restart test.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/nordvpnlite.py
+++ b/nat-lab/tests/nordvpnlite.py
@@ -290,31 +290,18 @@ class NordVpnLite:
             await self.remove_logs()
             await self.save_config()
 
-            async def wait_for_nordvpnlite_start():
-                await self.wait_for_nordvpnlite_socket()
-                while True:
-                    try:
-                        if not await self.is_alive():
-                            raise RuntimeError(
-                                "socket exists but daemon's not running."
-                            )
-                        break
-                    except IgnoreableError:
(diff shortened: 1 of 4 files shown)


nat-lab/tests/nordvpnlite.py
djkarwowski: With the introduction of the `wait_for_nordvpnlite_start(self)` instance method in line 442, do we still need this local function (which is pretty much the same)?

djkarwowski (commented):
I've left just one comment, but overall looks good to me

+1

nat-lab/tests/nordvpnlite.py
tomasz-grz: Good catch, fixed

mathiaspeters (approved):
+1

lcruz99:
this looks a nice improvement, however would be nice to add some logs or ticket ID to the PR, besides the usual commit history cleaning before merge :)

PR #1783, opened 2026-05-25
LLT-7321: Reject outgoing packets with stale addresses

lcruz99:
### Problem
After connecting to exit-node, past stale connections might still go through and pollute NAT with outdated addresses.

### Solution
Reject outbound packets (incoming on tunnel's POV) whose source IP are non-tunnel addresses.

Unit and integration tests added.

- [x] README.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -159,6 +159,8 @@ pub struct FirewallState {
     pub whitelist: Whitelist,
     /// Local node ip addresses
     pub ip_addresses: Vec<StdIpAddr>,
+    /// Whether an exit node is currently active
+    pub exit_node_present: bool,
 }
 
 /// Configuration for firewall initialization.
@@ -181,7 +183,7 @@ pub struct StatefulFirewall {
     config: FirewallConfig,
     /// Current firewall state
     state: RwLock<FirewallState>,
-    /// Last known host interface IPs reported by network monitor.
+    /// Last known host interfaces reported by network monitor.
     interface_ips: RwLock<Vec<StdIpAddr>>,
     /// Chain configuration function
     configure_chain_fn: ConfigureChainFn,
@@ -196,7 +198,6 @@ unsafe impl Send for StatefulFirewall {}
 impl LocalInterfacesObserver for StatefulFirewall {
     fn notify(&self, addrs: &[StdIpAddr]) {
         telio_log_debug!("Firewall notified about network changes: {:?}", addrs);
-
         *self.interface_ips.write() = addrs.to_vec();
         self.refresh_chain();
     }
@@ -376,9 +377,19 @@ impl StatefulFirewall {
     }
 }
 
-fn dst_net_all_ports_filter(net: IpNet, inverted: bool) -> Filter {
+fn filter_dst_ip_all_ports(net: IpNet, inverted: bool) -> Filter {
     Filter {
         filter_data: FilterData::DstNetwork(NetworkFilterData {
+            network: net,
+            port_range: (0, 65535),
+        }),
+        inverted,
+    }
+}
+
+fn filter_src_ip_all_ports(net: IpNet, inverted: bool) -> Filter {
+    Filter {
+        filter_data: FilterData::SrcNetwork(NetworkFilterData {
             network: net,
             port_range: (0, 65535),
         }),
@@ -399,18 +410,18 @@ fn get_local_area_networks_filters(
 
     // Include packets which are going to local IPv4 networks
     let mut ipv4_local_area_network_filters = ipv4_local_area_networks
-        .map(|network| vec![dst_net_all_ports_filter(IpNet::V4(network), false)]);
+        .map(|network| vec![filter_dst_ip_all_ports(IpNet::V4(network), false)]);
 
     // Exclude packets from the exclude range
     if let Some(exclude_range) = exclude_ip_range {
         for local_net_filters in ipv4_local_area_network_filters.iter_mut() {
-            local_net_filters.push(dst_net_all_ports_filter(IpNet::V4(exclude_range), true));
+            local_net_filters.push(filter_dst_ip_all_ports(IpNet::V4(exclude_range), true));
         }
     }
 
     let mut ipv6_local_area_network_filters = vec![
         // Include packets which are going to local network
-        dst_net_all_ports_filter(
+        filter_dst_ip_all_ports(
             IpNet::V6(Ipv6Net::new_assert(
                 // TODO: Actually it Unique Local Address range should be (from what I found) fc00::/7,
                 // but it seems that we assume (and our tests do) that it is fc00::/6
@@ -420,7 +431,7 @@ fn get_local_area_networks_filters(
             false,
         ),
         // Exclude packets which are going to local interfaces
-        dst_net_all_ports_filter(
+        filter_dst_ip_all_ports(
             IpNet::V6(Ipv6Net::new_assert(
                 StdIpv6Addr::new(0xfd74, 0x656c, 0x696f, 0, 0, 0, 0, 0),
                 64,
@@ -432,10 +443,10 @@ fn get_local_area_networks_filters(
     for ip in local_ifs_addrs.iter() {
         if ip.is_ipv4() {
             for local_network_filters in ipv4_local_area_network_filters.iter_mut() {
-                local_network_filters.push(dst_net_all_ports_filter(IpNet::from(*ip), true));
+                local_network_filters.push(filter_dst_ip_all_ports(IpNet::from(*ip), true));
             }
         } else {
-            ipv6_local_area_network_filters.push(dst_net_all_ports_filter(IpNet::from(*ip), true));
+            ipv6_local_area_network_filters.push(filter_dst_ip_all_ports(IpNet::from(*ip), true));
         }
     }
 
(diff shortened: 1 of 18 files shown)


crates/telio-network-monitors/src/monitor.rs
```diff
@@ -1,13 +1,12 @@
...
 //! Get notified when network paths change
 //! and update local IP address cache
-use crate::{local_interfaces::gather_local_interfaces, local_interfaces::GetIfAddrs};
+use crate::local_interfaces::{gather_local_interfaces, GetIfAddrs};
 use if_addrs::Interface;
 use once_cell::sync::Lazy;
 use parking_lot::Mutex;
 use std::{
     io,
-    net::IpAddr,
```
mathiaspeters, line 10: Nit: Is there a reason why this import was removed?

nat-lab/tests/test_mesh_firewall.py
```diff
@@ -698,3 +713,171 @@ async def test_mesh_firewall_tcp_stuck_in_last_ack_state_conn_kill_from_client_s
...
                 await conntrack.wait_for_no_violations()
+
+
+# LLT-7321: a TCP connection established BEFORE being connected to VPN
+# carries the client's physical interface IP as its source.
+#
+# After connecting to VPN, the kernel reroutes it through the tunnel
+# but still with the wrong source IP. Exit node's receive this packet,
+# poisoning their NAT table which causes return traffic to be misrouted.
+@pytest.mark.xfail(reason="LLT-7321")
```
mathiaspeters: If this PR fixes the issue, do we need this `xfail` anymore?

djkarwowski: Is the xfail mark still needed after applying the actual fix?

src/device/wg_controller.rs
```diff
@@ -645,6 +647,15 @@ async fn consolidate_firewall<F: Firewall>(
             .ok_or(Error::IpNotSet)?;
     }
 
+    if is_vpn_exit_node {
+        state
+            .ip_addresses
+            .push(IpAddr::V4(Ipv4Addr::new(10, 5, 0, 2)));
```
tomasz-grz: Nit: Should we add a `const` for this in `telio-model/src/constants.rs` and use it, like with `VPN_EXTERNAL_IPV4`? 

Actually what was the `VPN_EXTERNAL_IPV4` (`10.5.0.1`), the VPN server node? And `10.5.0.2` is the local tunnel interface address, or the other way around?

tomasz-grz (commented):
Nice work +1, just small nit

crates/telio-firewall/src/firewall.rs
```diff
@@ -376,7 +377,7 @@ impl StatefulFirewall {
     }
 }
 
-fn dst_net_all_ports_filter(net: IpNet, inverted: bool) -> Filter {
+fn get_filter_dst(net: IpNet, inverted: bool) -> Filter {
```
mathiaspeters: Nit: I think the previous name was better, though it wasn't great either. The fact that it mentioned `all_ports` I think gives more context what the filter is actually doing. I think `filter_dst_ip_all_ports` and `filter_src_ip_all_ports` would be better names

mathiaspeters (commented):
For the libfirewall pytest mark, all the tests in `test_tp_lite_stats.py` should also get the mark

What do you think of doing an in-branch test of this before we merge to make sure the fix is working and isn't breaking anything? It should be tested by the android team anyways to make sure it works, so I'm thinking let's do it early

+0.75

djkarwowski (commented):
LGTM, +0.5

nat-lab/tests/test_mesh_firewall.py
lcruz99: it should have been removed, thanks

crates/telio-firewall/src/firewall.rs
Damos-dev: Maybe `get_filter_dst_all_ports`?

src/device/wg_controller.rs
Damos-dev: Yes, ie: in other places (natlab tests) we using `LIBTELIO_LOCAL_IP` for `10.5.0.2`

src/device/wg_controller.rs
```diff
@@ -1564,8 +1583,57 @@ mod tests {
...
+        let mut requested_state = create_requested_state(vec![]);
+        requested_state.exit_node = Some(ExitNode {
+            public_key: pub_key_vpn,
+            ..Default::default()
+        });
+
+        let mut expected_ips = vec![
+            IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+            IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1)),
+            IpAddr::V4(Ipv4Addr::new(10, 5, 0, 2)),
```
Damos-dev: Same comment as @tomasz-grz  : maybe use const for 10.5.0.2?

src/device/wg_controller.rs
```diff
@@ -1564,8 +1583,57 @@ mod tests {
...
+        let pub_key_vpn = SecretKey::gen().public();
+
+        let mut requested_state = create_requested_state(vec![]);
+        requested_state.exit_node = Some(ExitNode {
+            public_key: pub_key_vpn,
+            ..Default::default()
+        });
+
+        let mut expected_ips = vec![
+            IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
```
Damos-dev: Maybe `Ipv4Addr::LOCALHOST` ?

src/device/wg_controller.rs
```diff
@@ -1564,8 +1583,57 @@ mod tests {
...
+            IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1)),
```
Damos-dev: Maybe `Ipv6Addr::LOCALHOST`?

src/device/wg_controller.rs
```diff
@@ -1551,6 +1562,14 @@ mod tests {
             ..Default::default()
         });
 
+        let mut expected_ips = vec![
+            IpAddr::V4(Ipv4Addr::LOCALHOST),
+            IpAddr::V6(Ipv6Addr::LOCALHOST),
+            IpAddr::V4(Ipv4Addr::new(10, 5, 0, 2)),
```
Damos-dev: same for const for `10.5.0.2`

crates/telio-firewall/src/firewall.rs
```diff
@@ -477,12 +489,28 @@ pub(crate) fn configure_chain(
...
-                dst_net_all_ports_filter(IpNet::from(blacklist_entry.ip), false),
+                get_filter_dst(IpNet::from(blacklist_entry.ip), false),
             ],
             action: LibfwVerdict::LibfwVerdictReject,
         });
     }
 
+    // LLT-7321: reject outbound packets whose src is not a known tunnel IP.
+    // Filters within a rule are ANDed (see LibfwRule docs).
+    if state.exit_node_present && !state.ip_addresses.is_empty() {
```
Damos-dev, line 500: It will be good to have a `natlab` tests with flag `allow_peer_local_network_access` set and tested an attempt to access host in a local network via `meshnet` in two scenarios:

1. with exit node present
2. without exit node present

Damos-dev (commented):
+0.75

crates/telio-firewall/src/firewall.rs
```diff
@@ -477,12 +489,28 @@ pub(crate) fn configure_chain(
...
+        let mut filters = vec![Filter {
+            filter_data: FilterData::Direction(Direction::Outbound),
+            inverted: false,
+        }];
+        for ip in &state.ip_addresses {
+            filters.push(get_filter_src(IpNet::from(*ip), true));
```
Jauler: Due to a fact that filters are AND'ed - would'nt this just simply never match if there are more than one IP address? 

Note: multiple IPs which does seem to happen in case of mesh+vpn

Jauler: Ah, wait, forgot that the rule is inverted. Which makes it correct.

Ignore my comment.

Jauler (dismissed):
+1.0

nat-lab/tests/test_mesh_firewall.py
lcruz99: it should have been removed, thanks

src/device/wg_controller.rs
lcruz99: yes, fixed

src/device/wg_controller.rs
lcruz99: yes, fixed

src/device/wg_controller.rs
lcruz99: > Actually what was the VPN_EXTERNAL_IPV4 (10.5.0.1), the VPN server node? And 10.5.0.2 is the local tunnel interface address, or the other way around?

10.5.0.1 for remote, 10.5.0.2 for local

lcruz99: fixed

src/device/wg_controller.rs
lcruz99: fixed

src/device/wg_controller.rs
lcruz99: fixed

crates/telio-network-monitors/src/monitor.rs
lcruz99: fixed

crates/telio-firewall/src/firewall.rs
lcruz99: you're right, changed

crates/telio-firewall/src/firewall.rs
lcruz99: At the moment, without routing through a node you won't be able to reach its LAN.

lcruz99: anw, I think this is out of scope for this PR

tomasz-grz (approved):

PR #1784, opened 2026-05-26
Add natlab test for TP-Lite with reconnects

mathiaspeters:
### Problem
Android has an issue with a different VPN protocol where TP-Lite stats callback is not called after multiple reconnects, we should add a regression test here to make sure it works and stays working

### Solution
Add natlab test

- [x] README.md is updated


mathiaspeters:
The bug this was supposed to be a regression test for had a completely different root cause that this test doesn't cover, and there already is a test checking that the callback works across reconnects

PR #1787, opened 2026-06-01
LLT-7361: Add option to force plaintext DNS when TP-Lite stats collected is enabled

mathiaspeters:
### Problem
Stats collection can only work on plaintext DNS packets, but some platforms, like android, eagerly use DoT and DoH. That means that the stats collection becomes mostly useless.

### Solution
Add a config option that allows apps to force plaintext DNS by adding rules to libfw to block traffic to all non-53 ports on TP-Lite DNS servers

- [x] README.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -161,6 +161,8 @@ pub struct FirewallState {
     pub ip_addresses: Vec<StdIpAddr>,
     /// Whether an exit node is currently active
     pub exit_node_present: bool,
+    /// List of DNS server IPs for which only plaintext DNS should be allowed
+    pub force_plaintext_dns_for_servers: Option<Vec<StdIpAddr>>,
 }
 
 /// Configuration for firewall initialization.
@@ -335,6 +337,11 @@ impl StatefulFirewall {
         config: TpLiteStatsOptions,
         collect_stats_cb: Box<dyn TpLiteStatsCallback>,
     ) -> Result<(), Error> {
+        let force_plaintext_dns_for_servers = config
+            .force_plaintext_dns
+            .unwrap_or(false)
+            .then(|| config.dns_server_ips.clone());
+
         let config = serde_json::to_string(&config).map_err(|err| {
             telio_log_warn!("Failed to convert TP-Lite config to json. {err:?}");
             Error::InvalidTpLiteConfig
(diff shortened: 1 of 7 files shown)


gytsto (commented):
+1

crates/telio-firewall/src/firewall.rs
```diff
@@ -159,6 +159,8 @@ pub struct FirewallState {
     pub whitelist: Whitelist,
     /// Local node ip addresses
     pub ip_addresses: Vec<StdIpAddr>,
+    /// List of DNS server IPs for which only plaintext DNS should be allowed
+    pub force_plaintext_dns_for_servers: Option<Vec<StdIpAddr>>,
```
djkarwowski, line 165: Nit: I feel like `force_plaintext_dns_for_servers` reads more like a boolean. Maybe something like `plaintext_only_dns_servers` would better reflect the actual semantics behind this value.

crates/telio-firewall/src/firewall.rs
```diff
@@ -334,6 +336,11 @@ impl StatefulFirewall {
         config: TpLiteStatsOptions,
         collect_stats_cb: Box<dyn TpLiteStatsCallback>,
     ) -> Result<(), Error> {
+        let force_plaintext_dns_for_servers = config
```
djkarwowski, line 340: Nit: should be aligned with [this name](https://github.com/NordSecurity/libtelio/pull/1787/changes#r3332690608) if you choose to update it.

crates/telio-firewall/src/firewall.rs
```diff
@@ -350,6 +357,11 @@ impl StatefulFirewall {
             let mut cb = self.tp_lite_stats_cb.callback.write();
             std::mem::swap(&mut old_cb, &mut cb);
         }
+        if let Some(thing) = force_plaintext_dns_for_servers {
```
djkarwowski: Nit: While `thing` is just a short-lived binding, it still feels more like a placeholder. Maybe a bit more descriptive name could still enhance readability a little.

djkarwowski (commented):
+1

Looks good, just added few nits

crates/telio-firewall/src/firewall.rs
mathiaspeters: I get what you mean, but the servers themselves accept non-plaintext DNS. So the name shouldn't communicate "these servers only allow plaintext DNS", but rather "make sure that only plaintext DNS is sent to these servers", if that makes sense as a distinction

crates/telio-firewall/src/firewall.rs
mathiaspeters: This was 100% a placeholder that I forgot to change, thanks!

Fixed

crates/telio-firewall/src/firewall.rs
djkarwowski: Thank you for the additional explanation. Let's stick with the original idea.

djkarwowski (dismissed):

djkarwowski (approved):

PR #1788, opened 2026-06-01
LLT-6821: Zeroize authentication token after its last use

tomaszpatejko:
### Problem
NordVPN authentication token was zeroized only when NordVPNLite configuration struct was dropped. The token could still be read after its last use and before it was dropped which is incorrect from security standpoint.

### Solution
Explicitly zeroize NordVPN authentication token after its last use, which is after nordlynx private key is obtained.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-6821
+++ b/.unreleased/LLT-6821
@@ -1,0 +1,1 @@
+nordvpnlite: Zeroize authentication token after its last use
--- a/clis/nordvpnlite/src/config.rs
+++ b/clis/nordvpnlite/src/config.rs
@@ -55,6 +55,10 @@ impl NordToken {
             "<PASTE_YOUR_AUTHENTICATION_TOKEN_HERE>".to_owned().into(),
         ))
     }
+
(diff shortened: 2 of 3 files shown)


nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/libtelio?pullRequest=1788) <br/>All committers have signed the CLA.

gytsto (commented):
+1

.unreleased/LTT-6821
```diff
@@ -0,0 +1 @@
+Zeroize authentication token after its last use
```
mathiaspeters: Nit: I think this line should include that it's only affecting nordvpnlite

mathiaspeters (dismissed):
+1

.unreleased/LTT-6821
tomaszpatejko: Done.

mathiaspeters (approved):

PR #1792, opened 2026-06-02
Fix performance metrics

olekoliinyk:
### Problem
- `iperf3` server catches `SIGTERM` internally and exits with code `1` instead of `143`, causing `DockerProcess` to raise `ProcessExecError` after a tightening change in exit code handling
- `CPU_CORES: "1"` starved `NepTUN` by sharing the single `vCPU` with `QEMU` and `macOS` daemons

### Solution
- Catch and suppress `ProcessExecError` with `returncode == 1` in `IperfServer.run()`
- `CPU_CORES: "4"` gives `NepTUN` a dedicated core, matching real hardware behavior

Tested on: `31422842`

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/iperf3.py
+++ b/nat-lab/tests/utils/iperf3.py
@@ -9,6 +9,7 @@ from tests.utils.connection import Connection, TargetOS
 from tests.utils.logger import log
 from tests.utils.output_notifier import OutputNotifier
 from tests.utils.process import Process
+from tests.utils.process.process import ProcessExecError
 from typing import AsyncIterator, Any, Dict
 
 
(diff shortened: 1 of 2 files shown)


asledzinskiy:
+1

nat-lab/tests/utils/iperf3.py
```diff
@@ -125,8 +126,12 @@ async def execute(self) -> None:
 
     @asynccontextmanager
     async def run(self) -> AsyncIterator["IperfServer"]:
-        async with self._process.run(stdout_callback=self.on_stdout):
-            yield self
+        try:
+            async with self._process.run(stdout_callback=self.on_stdout):
+                yield self
+        except ProcessExecError as e:
+            if e.returncode != 1:
```
gytsto, line 136: can you add a comment why this thing is suppressed and maybe debug log, that it was killed but all good, we got it handled (something like that) 😄

gytsto (commented):
+1

asledzinskiy (approved):

PR #1794, opened 2026-06-02
Warning on stdout/stderr about using a recommended server

tomasz-kumor:
## Problem

NordVPNLite users are not informed via the stdout/stderr about the recommended server being used when a non-existent country/country code is requested, or when there is a problem obtaining the countries list.

## Solution

Add a message to notify users that the recommended server is being used. Warn about the invalid country/country code, and explain how to list valid options.

## Changes
* `logging.rs` defined `nordvpnlite_logstd_<warn|error>` macros
* `core_api.rs ` used  `nordvpnlite_logstd_<warn|error>` and improved the logging for a scenario with a fallback to a recommended server

## Verification

Create a `config.json` file (with valid authentication_token) i.e.: 
```
{
  "authentication_token": "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
  "vpn": {
    "country": "UNSUPPORTED"
  },
  "log_level": "info",
  "log_file_path": "/var/log/nordvpnlite_log_file.log",
  "log_file_count": 0,
  "adapter_type": "linux-native",
  "interface": {
    "name": "utun10",
    "config_provider": "manual"
  }
}
```
Launch nordvpnlite in `no-detach` mode.
```
 ./nordvpnlite start --config-file ~/config.json --no-detach
```
Check that the following logs are visible on the terminal
```
WARNING: Servers not found for 'UNSUPPORTED' country code. To list supported countries/country codes, run: nordvpnlite countries
WARNING: Falling back to a recommended server

```

Change in `config.json` the `country` to unsupported country code with non--ASCII character i.e.: PL_LS and check that the a message in visible on the terminal.

In `detach` mode the `stdout`/`stderr` outputs are redirected by default to `/var/log/nordvpnlite.log` and the logs mentioned above will be written there.

 

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/core_api.rs
+++ b/clis/nordvpnlite/src/core_api.rs
@@ -19,6 +19,7 @@ use telio::telio_utils::exponential_backoff::{
 
 use crate::config::{Endpoint, NordToken, NordVpnLiteConfig, NordlynxKeyResponse, VpnConfig};
 use crate::NordVpnLiteError;
+use crate::{nordvpnlite_logstd_error, nordvpnlite_logstd_warn};
 
 const API_BASE: &str = "https://api.nordvpn.com/v1";
 const WIREGUARD_ID: u64 = 35;
@@ -432,44 +433,62 @@ async fn get_recommended_servers(
 }
 
 pub async fn get_server_endpoints_list(config: &NordVpnLiteConfig) -> Result<Vec<Endpoint>, Error> {
-    let country_id: Option<u64> = match &config.vpn {
+    enum CountryResolution {
+        Recommended,
+        Found(u64),
+        FallbackToRecommended,
+    }
+
+    let resolution = match &config.vpn {
         // Use the endpoint directly if provided in the config
         VpnConfig::Server(endpoint) => return Ok(vec![endpoint.clone()]),
         // Find VPN exit node based on country
         VpnConfig::Country(target_country) => {
             if !target_country.chars().all(|c| c.is_ascii_alphabetic()) {
-                warn!("Invalid country format: '{target_country}', non-ascii characters used");
-                None
+                nordvpnlite_logstd_warn!(
+                    "Invalid country format: '{target_country}', non-ascii characters used. To list supported countries/country codes, run: nordvpnlite countries"
+                );
+                CountryResolution::FallbackToRecommended
             } else {
                 match get_countries_with_exp_backoff(config.http_certificate_file_path.as_deref())
                     .await
                 {
-                    Ok(countries_list) => countries_list
-                        .iter()
-                        .find_map(|country| {
-                            // search for country full name or iso code
-                            if country.matches(target_country) {
-                                Some(country.id)
-                            } else {
-                                None
-                            }
-                        })
-                        .or_else(|| {
-                            warn!("Servers not found for '{target_country}' country code.");
+                    Ok(countries_list) => match countries_list.iter().find_map(|country| {
+                        // search for country full name or iso code
+                        if country.matches(target_country) {
+                            Some(country.id)
+                        } else {
                             None
-                        }),
+                        }
+                    }) {
+                        Some(id) => CountryResolution::Found(id),
+                        None => {
+                            nordvpnlite_logstd_warn!(
+                                "Servers not found for '{target_country}' country code. To list supported countries/country codes, run: nordvpnlite countries"
+                            );
+                            CountryResolution::FallbackToRecommended
+                        }
+                    },
                     Err(e) => {
-                        error!("Getting countries failed due to: {e}");
-                        None
+                        nordvpnlite_logstd_error!("Getting countries failed due to: {e}");
+                        CountryResolution::FallbackToRecommended
                     }
                 }
             }
         }
-        VpnConfig::Recommended => None,
+        VpnConfig::Recommended => CountryResolution::Recommended,
     };
 
-    if country_id.is_none() {
-        info!("Using a recommended server");
+    let country_id = match resolution {
+        CountryResolution::Found(id) => Some(id),
+        CountryResolution::Recommended => {
+            info!("Using a recommended server");
+            None
+        }
+        CountryResolution::FallbackToRecommended => {
+            nordvpnlite_logstd_warn!("Falling back to a recommended server");
+            None
+        }
     };
 
     // Fetch the list of recommended server from the API
--- a/clis/nordvpnlite/src/logging.rs
+++ b/clis/nordvpnlite/src/logging.rs
@@ -50,3 +50,19 @@ pub async fn setup_logging<P: AsRef<Path>>(
 
     Ok(tracing_worker_guard)
 }
+
+#[macro_export]
+macro_rules! nordvpnlite_logstd_warn {
+    ($($arg:tt)*) => {
+        ::tracing::warn!($($arg)*);
+        eprintln!("WARNING: {}", format_args!($($arg)*));
+    };
+}
+
+#[macro_export]
+macro_rules! nordvpnlite_logstd_error {
+    ($($arg:tt)*) => {
+        ::tracing::error!($($arg)*);
+        eprintln!("ERROR: {}", format_args!($($arg)*));
+    };
+}


nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/libtelio?pullRequest=1794) <br/>All committers have signed the CLA.

clis/nordvpnlite/src/core_api.rs
```diff
@@ -456,10 +460,15 @@ pub async fn get_server_endpoints_list(config: &NordVpnLiteConfig) -> Result<Vec
...
                             warn!("Servers not found for '{target_country}' country code.");
+                            println!(
+                                "WARNING: Servers not found for country/country code: '{target_country}'.\n\
+                                To list supported countries/country codes, run: nordvpnlite countries"
+                            );
                             None
                         }),
                     Err(e) => {
                         error!("Getting countries failed due to: {e}");
+                        println!("WARNING: Getting countries failed due to: {e}");
```
tomasz-grz: FIY: You can use `eprintln!` to write errors to `stderr`

tomasz-grz (commented):
+1

Just a note that in the typical use cases (OpenWRT) nordvpnlite is started by `procd` and is not attached to the terminal, therefore this message will not be directly visible to the user.

Maybe worth to add a field to `TelioStatusReport`?

clis/nordvpnlite/src/core_api.rs
tomasz-kumor: Yes, I know but I wanted to log it as a warning so I used println

tomasz-kumor:
> +1
> 
> Just a note that in the typical use cases (OpenWRT) nordvpnlite is started by `procd` and is not attached to the terminal, therefore this message will not be directly visible to the user.
> 
> Maybe worth to add a field to `TelioStatusReport`?

In such case the stdout/stderr could be redirected to a file.

clis/nordvpnlite/src/core_api.rs
tomaszpatejko: I'm probably missing something but if I read the code correctly, we are in erroneous situation: `get_countries_with_exp_backoff` returned `Err` and the error message is printed out with `error!` in line 470, so using `println!` to print out the same  message as a warning could be a bit confusing and redundant.

tomasz-kumor: You are right, reporting the same message as error and warning is confusing. I will print it as `eprintln!`. Regarding redundancy, `error!`, `info!` and `warn!`are redirected to the file, they are not visible on terminal when nordvpnlite is launched in `no-detach` mode.

mathiaspeters:
It's probably possible to configure the logger to both print to file and to terminal, and do that for `no-detach`. It might get spammy, but maybe it's possible to have different log level to the terminal and cap it at `info` or `warn`

clis/nordvpnlite/src/core_api.rs
tomasz-grz: `debug!`, `error!`, `info!` and `warn!` are the logging macros from [tracing](https://github.com/tokio-rs/tracing).
In our case they are always directed to a log file (managed by [tracing_subscriber](https://github.com/NordSecurity/libtelio/blob/main/clis/nordvpnlite/src/logging.rs)) (different from the stdout & stderr redirect).

In real case, these two log lines will appear in two different log files. (or `stdout` & `stderr` if launched with `no-detach`)

tomasz-grz:
Good idea from @mathiaspeters . You possibly just need to add another `tracing_subscriber` in `logging.rs`

tomasz-kumor:
@mathiaspeters @tomasz-grz @tomaszpatejko 

The idea with adding 2nd `layer` and `tracing_subscriber` is good if we want to log on `stdout/stderr` all the information that are also traced to a file. But in my opinion it is to much for a user. I was thinking about adding some parameter that could control which logs should be logged to stdout/stderr. It is possible but it will require quite complex logic to filter out that parameter from what is logged. Without that filtering this new parameter will be logged as part of a log message to a file and stdtout/stderr. The simplest solution that I came up with is to define custom `info!`, `warn!`, `error!` macros that optionally as the first parameter take an argument that controls the behavior. 

Let me know what you think.

tomaszpatejko:
@tomasz-kumor I think the idea is fine. I'm just wondering whether there is any policy regarding macros in the project. Should they be avoided because of cost of maintaining them? Or are they fine to use?

I'm new to the project that's why I'm asking. @mathiaspeters @tomasz-grz

tomaszpatejko:
+1

tomasz-kumor:
Rebased on top of latest main

mathiaspeters:
Adding a new macro for specifically the nordvpnlite crate would be a good option, as long as the extra parameter is optional. And let's follow the `telio_log_<level>!` naming, but specific to nordvpnlite

mathiaspeters:
> I'm just wondering whether there is any policy regarding macros in the project. Should they be avoided because of cost of maintaining them? Or are they fine to use?

We don't have a policy regarding macros. My personal opinion is to avoid them because they can be harder to work with, but sometimes they are simply the best option

mathiaspeters:
I'm not a big fan of the name `TraceOnStdout` but I can't think of a better name so +1 from me

tomasz-grz:
Thanks for working on this, a few thoughts from me:

I am personally not fully convinced regarding custom macros here.
Could you help me understand what they archive beyond than the already existing logging and tracing functionality can provide?

For example you can define a second tracing subscriber layer that prints to stderr only the WARN and ERROR logs only from a specific module, eg `core_api`, no custom macros are necessary for this:

```rust
    // existing configurable log layer
    let file_layer = fmt::layer()
        .with_writer(non_blocking_writer)
        .with_ansi(false)
        .with_line_number(true)
        .with_level(true)
        .with_filter(log_level);

    //  stderr logs only for warn/error from the core_api module
    let stderr_layer = fmt::layer()
        .with_writer(std::io::stderr)
        .with_ansi(true)
        .with_line_number(true)
        .with_level(true)
        .with_filter(EnvFilter::new("nordvpnlite::core_api=warn"));

    Registry::default()
        .with(file_layer)
        .with(stderr_layer)
        .init();
```

I like the more explicit message regarding the fallback server selection, it's nice improvement on it's own.

But I think we should clarify the problem we're actually solving. Who are the "users" in this case? If it's end users, they won't see stdout/stderr in most cases since the daemon runs detached in 95% of the time, and this would just create duplicate logs in two files.
If it's us, this is developer experience improvement, which is good, but should be framed that way, and I would suggest we keep it simple and not over engineer the logging side of things.

tomasz-kumor:
> Thanks for working on this, a few thoughts from me:
> 
> I am personally not fully convinced regarding custom macros here. Could you help me understand what they archive beyond than the already existing logging and tracing functionality can provide?
> 
> For example you can define a second tracing subscriber layer that prints to stderr only the WARN and ERROR logs only from a specific module, eg `core_api`, no custom macros are necessary for this:
> 
> ```rust
>     // existing configurable log layer
>     let file_layer = fmt::layer()
>         .with_writer(non_blocking_writer)
>         .with_ansi(false)
>         .with_line_number(true)
>         .with_level(true)
>         .with_filter(log_level);
> 
>     //  stderr logs only for warn/error from the core_api module
>     let stderr_layer = fmt::layer()
>         .with_writer(std::io::stderr)
>         .with_ansi(true)
>         .with_line_number(true)
>         .with_level(true)
>         .with_filter(EnvFilter::new("nordvpnlite::core_api=warn"));
> 
>     Registry::default()
>         .with(file_layer)
>         .with(stderr_layer)
>         .init();
> ```
> 
> I like the more explicit message regarding the fallback server selection, it's nice improvement on it's own.
> 
> But I think we should clarify the problem we're actually solving. Who are the "users" in this case? If it's end users, they won't see stdout/stderr in most cases since the daemon runs detached in 95% of the time, and this would just create duplicate logs in two files. If it's us, this is developer experience improvement, which is good, but should be framed that way, and I would suggest we keep it simple and not over engineer the logging side of things.

I don't know the story behind the https://nordsec.atlassian.net/browse/LLT-6853, but I think that by "user" they mean the user who uses the app in `no-detach` mode. Wouldn't adding a new layer be too much spam? I don't know if all error/warning should be presented to the end user. @matislovas what is your opinion?

matislovas:
> Thanks for working on this, a few thoughts from me:
> 
> I am personally not fully convinced regarding custom macros here. Could you help me understand what they archive beyond than the already existing logging and tracing functionality can provide?
> 
> For example you can define a second tracing subscriber layer that prints to stderr only the WARN and ERROR logs only from a specific module, eg `core_api`, no custom macros are necessary for this:
> 
> ```rust
>     // existing configurable log layer
>     let file_layer = fmt::layer()
>         .with_writer(non_blocking_writer)
>         .with_ansi(false)
>         .with_line_number(true)
>         .with_level(true)
>         .with_filter(log_level);
> 
>     //  stderr logs only for warn/error from the core_api module
>     let stderr_layer = fmt::layer()
>         .with_writer(std::io::stderr)
>         .with_ansi(true)
>         .with_line_number(true)
>         .with_level(true)
>         .with_filter(EnvFilter::new("nordvpnlite::core_api=warn"));
> 
>     Registry::default()
>         .with(file_layer)
>         .with(stderr_layer)
>         .init();
> ```
> 
> I like the more explicit message regarding the fallback server selection, it's nice improvement on it's own.
> 
> But I think we should clarify the problem we're actually solving. Who are the "users" in this case? If it's end users, they won't see stdout/stderr in most cases since the daemon runs detached in 95% of the time, and this would just create duplicate logs in two files. If it's us, this is developer experience improvement, which is good, but should be framed that way, and I would suggest we keep it simple and not over engineer the logging side of things.

@tomasz-grz  the original intention was, that when user runs the following command in the terminal:
```
nordvpnlite connect fr
```
he/she should see the error message in stdout/stderr, that we haven't connected to a server in France (if for some reason it is not available at the moment), and print the actual connected country.

If I get this right, to see that error message currently, user has issue the command, and than go check the logs for error. We do want to print this particular error into stdout/stderr instead of logging it into file only.

tomasz-grz:
@tomasz-kumor What do you mean regarding too much spam? Like printing too many logs?

The user is not intended to run the daemon usually with `--no-detach` mode, that's mostly intended for debugging and the system service manager. 

@matislovas AFAIK there is no `nordvpnlite connect fr` command at the moment. The connection is defined through the config file.

The daemon is usually launched by the system service manager like `procd`, and the user is expected to interact with the daemon through an IPC "client" command (or a GUI that implements the IPC).

When the daemon prints something to stdout, the user will not usually see it directly, as it's caught by `procd` and redirected to system service logs (separate from the log file managed by tracing_subscriber).

So if you want to add a client command, like `nordvpnlite connect fr` you would probably want to send the status back trough IPC and then have the [client process](https://github.com/NordSecurity/libtelio/blob/49a405a2233ea58e47099d783fca2b433c82afee/clis/nordvpnlite/src/main.rs#L104) print it to stdout, not the daemon process.

Currently the client command `nordvpnlite status` prints the [actual connected VPN server](https://github.com/NordSecurity/libtelio/blob/49a405a2233ea58e47099d783fca2b433c82afee/clis/nordvpnlite/src/daemon.rs#L106), that's why I suggested adding a field to `TelioStatusReport` if we wanted to display something to the end user.

I think adding a line like this for easier debugging, is absolutely fine (and doesn't need custom macros), but it's not going to be obvious to most end users.
```rust
println!("Servers not found for '{target_country}' country code.");
// or
println!("WARNING: Getting countries failed due to: {e}");
```

tomasz-kumor:
@tomasz-grz I received some information about the task background. The main use case is not the `no-detach` mode with a user launching the app in a Linux terminal, as I previously thought, but a scenario where `nordvpnlite` is running on an embedded device such as an OpenWrt router. The goal is to inform the user that — for example, due to a misconfiguration — the requested country or country code could not be selected, and that the recommended server is being used instead. In such a scenario, we want to capture this information in the device log. Because we are dealing with embedded devices, we want to limit the amount of logged information to the most important entries only. That is why I think that repeating on `stdout/stderr` all warning and error logs that are currently traced is not a good idea. It is also about whether the user can do anything about the error or warning. In the case of a country or country code misconfiguration, they can.

clis/nordvpnlite/src/main.rs
```diff
@@ -12,6 +12,7 @@ mod core_api;
 mod daemon;
 mod interface;
 mod logging;
+mod logging_macros;
```
tomasz-grz: We already have a `logging.rs` module, what do you think about combining them?

clis/nordvpnlite/src/logging_macros.rs
```diff
@@ -0,0 +1,60 @@
...
+    }};
+}
+
+/// wrapping tracing::trace
+#[macro_export]
+macro_rules! nordvpnlite_log_trace {
+    ($($arg:tt)*) => {{
+        ::tracing::trace!($($arg)*);
+    }};
+}
```
tomasz-grz: These don't seem to be doing anything additional, why do we want to replace tracing macros?

clis/nordvpnlite/src/logging_macros.rs
```diff
@@ -0,0 +1,60 @@
...
+#[macro_export]
+macro_rules! nordvpnlite_log_info {
+    (TraceOnStdout, $($arg:tt)*) => {{
+        ::tracing::info!($($arg)*);
+        ::std::println!("{}", ::std::format!($($arg)*));
+    }};
+    ($($arg:tt)*) => {{
+        ::tracing::info!($($arg)*);
+    }};
+}
```
tomasz-grz: Just a note, that since this is not gated by `log_level` (by default in production for OpenWRT we have `error` level as minimum), these messages will be directed to syslogs but not the regular log file.

clis/nordvpnlite/src/logging_macros.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// `nordvpnlite_log_error!` macro with optional `TraceOnStdout`.
+///
+/// - `nordvpnlite_log_error!(TraceOnStdout, ...)` → tracing::error! + eprintln! to stderr
+/// - `nordvpnlite_log_error!(...)` → tracing::error! only
+#[macro_export]
+macro_rules! nordvpnlite_log_error {
+    (TraceOnStdout, $($arg:tt)*) => {{
```
tomasz-grz: What do you think about removing the optional`TraceOnStdout`, and only calling it from the lines we want to appear both in `syslog` and `logfile`? (maybe renamed to `logstd_error!`?)

This way you don't need to replace all tracing `error!` and `warn!` calls in the codebase, and it's more explicit which message we want to appear both in `syslog` and `logfile`.

clis/nordvpnlite/src/core_api.rs
```diff
@@ -469,7 +492,7 @@ pub async fn get_server_endpoints_list(config: &NordVpnLiteConfig) -> Result<Vec
     };
 
     if country_id.is_none() {
-        info!("Using a recommended server");
+        nordvpnlite_log_info!(TraceOnStdout, "Using a recommended server");
```
tomasz-grz: This message will appear also if the user themselves set `vpn: recommended`.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -327,32 +344,32 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
             // TODO: LLT-6460 - We can store the recommended server list, just in case
             // one server fails to connect, try the next one.
             if let Some(endpoint) = endpoints.first() {
-                debug!("Selected exit node: {:#?}", endpoint);
+                nordvpnlite_log_debug!("Selected exit node: {:#?}", endpoint);
```
tomasz-grz: This is where we log to which server we are going to connect.

At this point it might be possible compare if it's the same as user defined in `NordVpnLiteConfig`

tomasz-grz (commented):
Thanks for clarifying the use case.

On OpenWRT the messages will definitely not reach the user through stdout or stderr.
These errors are already present in the log file, we will be repeating them in the syslogs.

The user will either have to `cat /var/log/nordvpnlite.log` or `logread | grep -i nordvpnlite` in either case. In some old GUI prototype we did have a tab that displayed the log file, so it could be a bit more "user" friendly.

What do you think about adding a field like `used_recommended: bool` or `used_fallback: bool` to [ExitNodeStatus](https://github.com/NordSecurity/libtelio/blob/49a405a2233ea58e47099d783fca2b433c82afee/clis/nordvpnlite/src/daemon.rs#L106)? This way it could be easier to visualize in GUI or with the `status` command.

clis/nordvpnlite/src/core_api.rs
tomasz-kumor: fixed

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: I have added the fallback logic to core_api, where based on country resolution state I log different messages.

clis/nordvpnlite/src/logging_macros.rs
tomasz-kumor: Fixed, but keep the naming convention closed to the @mathiaspeters suggestion.

clis/nordvpnlite/src/logging_macros.rs
tomasz-kumor: That was added as I did not want to mix tracing::macros with the new ones in code and for consistency I have switched everywhere to the new ones.  But since I have changed the approach this was removed.

clis/nordvpnlite/src/main.rs
tomasz-kumor: Good

tomasz-kumor:
> What do you think about adding a field like used_recommended: bool or used_fallback: bool to [ExitNodeStatus](https://github.com/NordSecurity/libtelio/blob/49a405a2233ea58e47099d783fca2b433c82afee/clis/nordvpnlite/src/daemon.rs#L106)? This way it could be easier to visualize in GUI or with the status command.

It could be useful as it would be faster than checking logs, especially when nordvpnlite launched as a daemon. But I would suggest to not add this as part of this PR.

tomasz-kumor:
Rebased

tomasz-kumor:
Added missing `.unreleased/LLT-6853_connection_to_non_existing_country_should_create_a_log_entry`

tomasz-grz (approved):
+1 Nice, I like it! `CountryResolution::FallbackToRecommended` could be used later to signal TelioStatusReport / and the GUI

tomaszpatejko (approved):

PR #1795, opened 2026-06-02
LLT-7321: backport block packets with wrong src ip from entering tunnel

mathiaspeters:
### Problem
Fix was merged to main, but we need to backport it to v7.0 releases

### Solution
Do backport, but skip some unimportant commits to keep hotfix small. All skipped commits relate to natlab tests and are non-critical. Additionally, had to slightly adjust natlab tests to use standard `setup_mesh_nodes` and `setup_connections` instead of using the newly introduced factories, as well as creating the exit stack in the test instead of providing it as a fixture.

Skipped commits: 
- 7625b7a: Hardcode tunnel local ipv4 - 10.5.0.2
- 370bef7: Add *.so files to .gitignore
- 2e0bf2c: Add libfirewall pytest mark
- 59139df: Fix tcpdump syntax error from unjoined filter predicates
- e0064db: Fail fast in netcat helpers when process exits early

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-7321
+++ b/.unreleased/LLT-7321
@@ -1,0 +1,1 @@
+Block outgoing packets with wrong source IP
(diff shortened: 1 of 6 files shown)


tomaszpatejko:
+1

PR #1798, opened 2026-06-03
Do not start ENS when routing through meshnet peer

matislovas:
### Problem
ENS starts its monitoring, when app start routing either through VPN server or meshnet's peer. We are not interested in enabling ENS, when routing through the later.

### Solution
`ens.start_monitoring()` is called only when routing through VPN server.

- [x] README.md is updated


gytsto (commented):
+1

Paintree (approved):
+1

PR #1800, opened 2026-06-03
Fix tp-lite dns whitelisting API

jjanowsk:
The list of whitelisted domains must be configurable at runtime via API call, not via feature config. This way users can udpate the list without the need to restart the tunnel.

- [x] README.md is updated


src/ffi.rs
```diff
@@ -938,6 +938,25 @@ impl Telio {
...
+    pub fn set_tp_lite_whitelisted_domains(&self, domains: Vec<String>) -> FfiResult<()> {
+        telio_log_info!(
+            "Telio::set_tp_lite_whitelisted_domains entry with instance id: {}. domains: {:?}",
+            self.id,
+            domains
+        );
+        catch_ffi_panic(|| {
+            self.device_op(true, |dev| {
+                dev.set_tp_lite_whitelisted_domains(domains.clone())
+                    .map_err(TelioError::from)
```
Jauler: could we add `.log_result(..)` In a similar fashion like other methods, for example at line 936

src/device.rs
```diff
@@ -2132,6 +2146,27 @@ impl Runtime {
...
+        #[cfg(feature = "enable_firewall")]
+        match &self.entities.firewall {
+            Some(fw) => {
+                fw.set_tp_lite_whitelisted_domains(domains);
+                Ok(())
+            }
+            None => Err(Error::FirewallDisabled),
+        }
+        #[cfg(not(feature = "enable_firewall"))]
+        Ok(())
```
Jauler: Lets fail here. 

Its a bit confusing to have `Ok(())` without actually doing what was requested.

The best would be to have `Error::FirewallUnsupported` in case it is disabled via compile time cargo feature. and like we have above `Error:FirewallDisabled` if it is runtime disabled.

matislovas:
+0.5

Jauler (approved):
+1.0

PR #1801, opened 2026-06-03
luci-app-nordvpnlite: update to meet all feature.

pesa1234:
Dear NordVPN team, as suggest I reopen it and I hope now it will be reviewed by the team.

Extend the NordVPN Lite LuCI backend ACLs and RPC handlers so the settings page can query service state, runtime status, country lists and per-server WireGuard metadata.

Rework the settings view to support recommended, country and specific server selection, fetch country and server data on demand, expose service start/restart/stop/enable/disable controls, and restart the service after Save & Apply while polling for the updated runtime state.

Example:
immagine
immagine
immagine

Changes:

- Port the expanded LuCI settings view from the standalone luci-app-nordvpnlite app.

- Add service enabled flag handling through UCI.

- Add service status and control actions for start, restart, stop, enable and disable.

- Add runtime status lookup and display for Telio state, tunnel IP and exit node details.

- Add VPN selection modes for recommended server, country code and specific server.

- Add country list lookup from the nordvpnlite command.

- Add NordVPN server lookup using the public API, hostname normalization and automatic IP/public key fill.

- Expand the rpcd ucode backend with the new config, service, lookup and runtime status methods.

- Expand the LuCI ACL grants and update the menu ACL dependency for the app.

- Add runtime dependencies required by the new LuCI backend: curl, jq and ca-bundle.

### Problem
Not so useful Luci app

### Solution
After merged is more complete

Run tested on openwrt snapshot

### DIFF
--- a/clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/ucode/nordvpnlite.uc
+++ b/clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/ucode/nordvpnlite.uc
@@ -1,20 +1,270 @@
 'use strict';
 
 const CONFIG_FILE = '/etc/nordvpnlite/config.json';
+const INIT_SCRIPT = '/etc/init.d/nordvpnlite';
+const SERVICE_NAME = 'nordvpnlite';
+const UCI_CONFIG = 'nordvpnlite';
+const UCI_SECTION = 'settings';
+const SERVERS_API_URL = 'https://api.nordvpn.com/v1/servers?limit=20000';
+const VALID_ACTIONS = ['start', 'stop', 'restart', 'reload', 'enable', 'disable'];
+const SERVER_LOOKUP_TIMEOUT = 45;
 let fs = require('fs');
 let log = require('log');
+
+function has_service() {
+	let st = fs.stat(INIT_SCRIPT);
+	return st && st.type == 'file' && st.user_exec != false;
+}
+
+function service_action(action) {
+	return system(sprintf('env -i %s %s >/dev/null 2>&1', INIT_SCRIPT, action));
+}
+
+function shell_quote(value) {
+	if (value == null || value == '')
+		return "''";
+
+	return "'" + replace(value, "'", "'\\''") + "'";
+}
+
+function has_command(command) {
+	return system(sprintf('command -v %s >/dev/null 2>&1', command)) == 0;
+}
+
+function read_command_output(command) {
+	let pp = fs.popen(command, 'r');
+	if (!pp)
+		return null;
+
+	let output = pp.read('all');
+	let exit_code = pp.close();
+
+	if (exit_code != 0 || output == null)
+		return null;
+
+	return trim(output);
+}
+
+function config_enabled() {
+	let value = read_command_output(sprintf('uci -q get %s.%s.enabled 2>/dev/null', UCI_CONFIG, UCI_SECTION));
+
+	if (value == null || value == '')
+		return true;
+
+	switch (lc(trim(value))) {
+		case '0':
+		case 'off':
+		case 'false':
+		case 'no':
+		case 'disabled':
+			return false;
+		default:
+			return true;
+	}
+}
+
+function write_config_enabled(enabled) {
+	let value = enabled ? '1' : '0';
+
+	return system(sprintf(
+		"uci -q set %s.%s=settings >/dev/null 2>&1 && " +
+		"uci -q set %s.%s.enabled='%s' >/dev/null 2>&1 && " +
+		"uci -q commit %s >/dev/null 2>&1",
+		UCI_CONFIG, UCI_SECTION,
+		UCI_CONFIG, UCI_SECTION, value,
+		UCI_CONFIG
+	)) == 0;
+}
+
+function fetch_server_data_with_jq(hostname) {
+	let url = shell_quote(SERVERS_API_URL);
+	let jq_filter = shell_quote(
+		'first(.[] | select(.hostname == $h) | {' +
+			'hostname,' +
+			'address: .station,' +
+			'supports_wireguard_udp: ([.technologies[] | select(.identifier == "wireguard_udp")] | length > 0),' +
+			'public_key: ([.technologies[] | select(.identifier == "wireguard_udp") | .metadata[] | select(.name == "public_key") | .value][0])' +
+		'})'
+	);
+	let quoted_hostname = shell_quote(hostname);
+	let commands = [];
+
+	if (!has_command('jq'))
+		return {
+			server: null,
+			error: 'jq is required for NordVPN server lookup.'
+		};
+
+	if (has_command('curl')) {
+		push(commands, sprintf(
+			'curl --connect-timeout 10 --max-time %d -sS %s 2>/dev/null | jq -c -r --arg h %s %s 2>/dev/null',
+			SERVER_LOOKUP_TIMEOUT, url, quoted_hostname, jq_filter
+		));
+	}
+
+	if (has_command('uclient-fetch')) {
+		push(commands, sprintf(
+			'uclient-fetch -qO- %s 2>/dev/null | jq -c -r --arg h %s %s 2>/dev/null',
+			url, quoted_hostname, jq_filter
+		));
+	}
+
+	if (has_command('wget')) {
+		push(commands, sprintf(
+			'wget -T %d -qO- %s 2>/dev/null | jq -c -r --arg h %s %s 2>/dev/null',
+			SERVER_LOOKUP_TIMEOUT, url, quoted_hostname, jq_filter
+		));
+	}
+
+	if (!length(commands)) {
+		return {
+			server: null,
+			error: 'No supported HTTP client found (curl, uclient-fetch or wget).'
+		};
+	}
+
+	for (let command in commands) {
+		let output = read_command_output(command);
+
+		if (!output)
+			continue;
+
+		try {
+			return {
+				server: json(output),
+				error: null
+			};
+		} catch (e) {
+			log.ERR('Failed to parse NordVPN server lookup response: %J', e);
+		}
+	}
+
+	return {
+		server: null,
+		error: sprintf('Lookup timed out or failed for %s.', hostname)
+	};
+}
+
+function fetch_countries() {
+	if (!has_command(SERVICE_NAME)) {
+		return {
+			countries: null,
+			error: sprintf('%s command not found.', SERVICE_NAME)
+		};
+	}
+
+	let output = read_command_output(sprintf('%s countries 2>/dev/null', SERVICE_NAME));
+	let countries = [];
+
+	if (!output) {
+		return {
+			countries: null,
+			error: sprintf('Unable to read available countries from %s.', SERVICE_NAME)
+		};
+	}
+
+	for (let line in split(output, /\r?\n/)) {
+		let entry = trim(line);
+		let parts = null;
+
+		if (entry == '')
+			continue;
+
+		parts = match(entry, /^(.*):\s*([A-Z]{2})$/);
+		if (!parts)
+			continue;
+
+		push(countries, {
+			name: trim(parts[1]),
+			code: parts[2]
+		});
+	}
+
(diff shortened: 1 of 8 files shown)


jjanowsk:
Hi! Just to let you know, I've just started looking at it. Most likely I'll send you my feedback tomorrow. I've also send the ipk to our QA engineer so he can give his opinion on the UI/UX. Thanks again for the contribution and for your patience

jjanowsk (commented):
Hi! I'll leave my comments free standing because they are mostly UI/UX related. Once we polish that I'll review the code itself.

Major:
1. I'm having trouble understanding what is the point of "Enabled" button. Can you please explain that?
2. It's a little counter intuitive when you install the app, paste the "Authentication token" and try to start the service for the first time. It will fail, because the token is not yet saved in the config file. It's obvious to us what's happening, but not very clear to the user. It might be hard to figure out that he needs to first click the "Save & Apply" button, and only then he should click the "Start" button. Moreover the app just hangs in that situation.
3. Similar situation happens when the user changes configuration e.g. from recommended server to the server from the country list. Changing those settings and clicking restart does not apply the new configuration. That's again counter-intuitive. 
4. Touching on the server configuration. I'm not sure I'm sold on the idea of selecting "Specific server" via hostname. Right now the nordvpnlite config allows only configuration via pub key and IP address and we could leave it this way in the OpenWRT UI. The problem I see with the current solution is that you need to query ~30MB server list and filter it in memory on the device. It can be slow. Also, by dropping this feature we can get rid of extra dependencies for our app. 
5. Why is clicking "Disable autostart", blocking "Start", "Restart", "Stop" buttons? It seems to be a bug.

Minor:
1. Status texts should not end with ".". E.g. "Stopped" not "Stopped."
2. Field names and field values are unaligned, at least in chrome:
image

pesa1234:
Hi @jjanowsk,
thank you for the detailed feedback. I’ve addressed all the points.

The service switch is a runtime master switch stored in /etc/config/nordvpnlite. Unlike autostart, it determines whether the init script is allowed to start the daemon.
I renamed it to “Enable service”, clarified its description and aligned the embedded init script with the OpenWrt package implementation.
Start and Restart now validate and save the current form configuration before controlling the service. A startup failure also produces a notification instead of leaving the UI apparently hanging.

Consequently, changing the selected country or other settings and clicking Restart now applies the new configuration.

I removed hostname-based server lookup. Specific server configuration now accepts the IP address and WireGuard public key directly, matching the native nordvpnlite configuration. The full server-list download and the curl, jq and ca-bundle dependencies were removed.

Autostart is now independent from runtime service controls. Disabling autostart no longer blocks Start, Restart or Stop.

I also removed trailing periods from status texts and adjusted the status field alignment.

Thanks again for the review!

jjanowsk:
Hi @pesa1234! Thanks for the updates. I was on vacation, but I'm back now and I'll review it this week

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/view/nordvpnlite/settings.js
```diff
@@ -29,70 +61,894 @@ var defaultConfig = {
...
         o.validate = function (_, value) {
-            return (value.length === 0 || value.length === 2) ? true : 'Please provide a two-letter country code.';
-        }
+            if (view.vpn_mode_option.formvalue('config') !== 'country')
+                return true;
 
-        return m.render();
-    },
+            value = String(value || '').trim().toUpperCase();
+            return /^[A-Z]{2}$/.test(value) ? true : _('Please provide a two-letter country code.');
```
jjanowsk, line 629: This throws `_ is not a function` TypeError because `_` is shadowed by the first parameter name.

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/view/nordvpnlite/settings.js
```diff
@@ -29,70 +61,894 @@ var defaultConfig = {
...
+
+        o = this.server_public_key_option = s.option(form.Value, 'server_public_key', _('Server Public Key'));
+        o.depends('vpn_mode', 'server');
+        o.placeholder = '';
+        o.description = _('Enter the WireGuard public key of the NordVPN server.');
+        o.validate = function (_, value) {
+            if (view.vpn_mode_option.formvalue('config') !== 'server')
+                return true;
+
+            return String(value || '').trim() !== '' ? true : _('Please provide the server public key.');
```
jjanowsk, line 677: Similar problem with `_` shadowing here.

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/ucode/nordvpnlite.uc
```diff
@@ -1,23 +1,195 @@
...
+const SERVICE_NAME = 'nordvpnlite';
+const UCI_CONFIG = 'nordvpnlite';
+const UCI_SECTION = 'settings';
+const VALID_ACTIONS = ['start', 'stop', 'restart', 'reload', 'enable', 'disable'];
 let fs = require('fs');
 let log = require('log');
 
+function has_service() {
+	let st = fs.stat(INIT_SCRIPT);
+	return st && st.type == 'file' && st.user_exec != false;
```
jjanowsk: `st.perm.user_exec`

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/ucode/nordvpnlite.uc
```diff
@@ -1,23 +1,195 @@
...
+	let output = pp.read('all');
+	let exit_code = pp.close();
+
+	if (exit_code != 0 || output == null)
+		return null;
+
+	return trim(output);
+}
+
+function config_enabled() {
```
jjanowsk: I still don't know what we are gaining by adding this enabled flag to be honest. I would gladly get rid of it unless you have some valid reason for keeping it. It would clean up the logic in this PR by not having to deal with weird states like: `Running (service disabled)`.

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/view/nordvpnlite/settings.js
```diff
@@ -29,70 +61,894 @@ var defaultConfig = {
...
         } catch (err) {
             ui.addNotification(_('Save failed'), E('p', err ? String(err) : _('Unknown error')));
+            return false;
         }
     },
 
-    handleSaveApply: null,
+    handleSave: null,
+
+    handleSaveApply: function () {
```
jjanowsk: Save and apply previously was only saving the configuration and I think it was a correct decision. I've asked you to also save the configuration when you click start, so the configuration visible in the UI is applied to the start. But now you've also added restarting to the save button, even when the service is not running. This I think is incorrect. (Maybe you have not added it now, maybe I've missed that during the first review.)

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/view/nordvpnlite/settings.js
```diff
@@ -29,70 +61,894 @@ var defaultConfig = {
...
+        if (!widget)
+            return;
+
+        if (typeof widget.clearChoices === 'function')
+            widget.clearChoices(true);
+
+        if (typeof widget.addChoices === 'function')
+            widget.addChoices(keys, choices);
+
+        widget.setValue(code !== '' ? code : '');
```
jjanowsk: it's just equivalent to: `widget.setValue(code);`

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/luci/ucode/nordvpnlite.uc
```diff
@@ -1,23 +1,195 @@
...
+
+function fetch_countries() {
+	if (!has_command(SERVICE_NAME)) {
+		return {
+			countries: null,
+			error: sprintf('%s command not found.', SERVICE_NAME)
+		};
+	}
+
+	let output = read_command_output(sprintf('%s countries 2>/dev/null', SERVICE_NAME));
```
jjanowsk: In case of lack of internet this can hang for a long time. I think we should add some kind of timeout mechanism here.

jjanowsk (changes requested):
Hi! Found a few technical issues. The main issue for me though is this enable button. I still don't understand why we would need that tbh. You can try to convince me, because maybe I'm missing someting, but in my opinion it is just adding unnecessary  complications without any benefit.

asledzinskiy:
One minor scenario that is not working for me:
- Pre-condition - `nordvpnlite` is not running
- Go to UI and click `Get status`  button

Actual result - banner with message 'Failed to get status'
Expected - `Service Status` is displaying `Not Running`

asledzinskiy:
And also another failed test case:
1. Open Services → NordVPN Lite.
2. Set VPN Selection = Specific server.
3. Leave Server Public Key empty 
4. Click Save & Apply.
5. Expected: validation error. Actual: red `TypeError: _ is not a function` banner

Was able to fix it with:
```
sed -i 's/function (_, value)/function (section_id, value)/g' \
  /www/luci-static/resources/view/nordvpnlite/settings.js
```

pesa1234:
> Hi! Found a few technical issues. The main issue for me though is this enable button. I still don't understand why we would need that tbh. You can try to convince me, because maybe I'm missing someting, but in my opinion it is just adding unnecessary complications without any benefit.

All the request should be addressed, thanks for point it... please check

jjanowsk (commented):
Hi @pesa1234! Thanks again for your contribution. I've checked the code and it's +1 from me. There are now few minor things left, and I can handle them for you if you agree. Of course your commits will be left marked as authored by you, but I can just move them to my branch and do those administrative things. If you prefer to do it then I need you to:
1. Rebase to fix `deny` and `black` lints
2. Create `.unreleased/luci_improvements` file,  and add short description there e.g.: `nordvpnlite: OpenWRT UI improvements: Added service control`
3. Take content of dod.yml file from the root of libtelio's repo and put it in the top of your PR description. Mark all three points as done. (This part TBH I'm not sure will work for external contribution, so that's why I would prefer to just do it for you to avoid any hustle. We don't have many external contributors;))

Few small issues left which I can also fix:
1. You've changed the behaviour of Save&Apply as I've asked you, but I think now the name does not match. We should use Save handler for that.
2. timeout is not default on all openwrt images. Please add +coreutils-timeout to the dependency list.
3. logout ucode method is unreachable. Let's get rid of it.

Let me know if you want me to handle that. If so I'll prepare new PR and link this one there.

pesa1234:
Hi @jjanowsk, thanks a lot for the review and for the +1!

I can take care of the remaining technical fixes on my side if you want — just let me know:

* use the correct Save handler;
* add `coreutils-timeout`;
* remove the `logout` call.

If you’re happy with that, I’ll update this PR accordingly.

For the repository-specific/admin parts, especially the DoD workflow and anything that is easier to handle internally, I’d really appreciate your help.

I can also test the updated package on a real OpenWrt device afterwards, including the LuCI UI, service controls, Save/Start/Restart behaviour, and runtime status.

Thanks again for helping me get this into shape for merging!

One more thing: as long as there is no free plan, even with limited bandwidth or a restricted server selection, I won’t be able to prepare a PR for the official OpenWrt packages repository.

It might be worth considering a small free tier, as that could make it easier for more OpenWrt users and contributors to test the package and potentially increase adoption.


Andrea

jjanowsk:
Ok, so please do those 3 + rebase, and I'll take if from there. Thanks again! Regarding your suggestion, it's outside of my  scope, but I'll pass the information to the relevant people.

Just one note about the rebase: Please do it first, and force push here, and only then do your fixes. This way I'll have nice "compare" button here on github to just see your latest changes. Thanks!

pesa1234:
@jjanowsk can you please check now, all the things should be done

jjanowsk:
Merged. Thanks again @pesa1234 for your contribution!

PR #1809, opened 2026-06-09
refactor(nat-lab): split the Client god-class into a libtelio_client package (LLT-5791)

gytsto:
### Problem

`nat-lab/tests/telio.py`'s `Client` had grown into a ~1160-line god-class mixing logging, tcpdump/coredumps, event waiting, VPN, analytics and tp-lite stats on top of actual device control. (It's the class `too-many-public-methods` is disabled for in `pyproject.toml`.)

LLT-5791

### Solution

Behavior-preserving refactor — split `Client` by composition into facades and move the client code into a `tests/libtelio_client/` package:

- `client.log` — log / process output
- `client.events` — event/state waiters
- `client.vpn` — VPN / exit-node control
- `client.analytics` — analytics triggers
- `client.tp_lite` — tp-lite stats
- `Runtime`/`Events` split out of the client module
- tcpdump / coredump / system-log clearing moved to connection-level setup (no longer client-owned)

Import is now `from tests.libtelio_client import Client`; calls use `client.<facet>.*`. nat-lab suite green.

- [x] README.md is updated


nat-lab/tests/libtelio_client/client.py
```diff
@@ -0,0 +1,508 @@
...
+                                " INSERT OR REPLACE INTO opt_in (id, value) VALUES (1, 1);"
+                            ),
+                        ],
+                        quiet=True,
+                    ).execute()
+                    return
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    print(
+                        f"maybe_write_device_fingerprint_to_moose_db error: {e}, retrying ..."
+                    )
```
Paintree: Maybe this could be converted to log?

Paintree (dismissed):
+1

nat-lab/tests/libtelio_client/client.py
gytsto: Sure thing, will do

gytsto: Done in dd330640 — converted to `log.warning`. Also applied the same change to the identical retry-error `print` in `test_lana.py`'s `get_moose_db_file` helper.

nat-lab/tests/test_pq.py
```diff
@@ -30,7 +30,7 @@ async def _connect_vpn_pq(
 ) -> datetime:
     wg_server = config.NLX_SERVER
 
-    await client.connect_to_vpn(
+    await client.vpn.connect_to_vpn(
```
tomasz-grz: In general I quite like this "categorization"

On the other hand `client.vpn.connect_to_vpn` is quite a mouthful and conveys redundant information, maybe something like `client.vpn.connect_to` would be cleaner, what do you think? 

`log.wait_for_log` -> `log.wait_for`
`log.get_log` -> `log.get` etc.

Hope it's trivial to do with a search and replace, so only NIT

tomasz-grz (commented):
+1, just a small NIT

Paintree (dismissed):

nat-lab/tests/test_pq.py
gytsto: quite good suggestion, will do

gytsto: updated

tomasz-grz (approved):

PR #1814, opened 2026-06-10
nordvpnlite streamline cleanup

tomasz-grz:
### Problem
The UCI provider for OpenWRT calls `reload_network` (uci commit network + /etc/init.d/network reload) twice during cleanup. Contributing to timeout on slow VM simulators during `nordvpnlite stop` command.

### Solution
- Increase IPC TIMEOUT_SEC from 20 to 30 to accommodate slower cleanup on emulated devices.
- Replace cleanup_exit_routes and cleanup_interface trait methods with a single cleanup method on the ConfigureInterface trait.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -12,7 +12,7 @@ use crate::{
     daemon::{NordVpnLiteError, TelioStatusReport},
 };
 
-pub(crate) const TIMEOUT_SEC: u64 = 20;
+pub(crate) const TIMEOUT_SEC: u64 = 30;
 
 #[derive(Parser, Debug, PartialEq)]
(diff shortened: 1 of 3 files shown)


asledzinskiy:
+1

matislovas:
+0.5

Jauler (approved):
While I am 90% sure that the `stop` command is taking 20+ seconds not because of slow emulated VMs, I mean, if network reload would take 20s due to slow VM - then like 100x heavier operation, like booting the OS - should take 2000s? But it does'nt. does it? My guess would be, that some sort of timeout is being waited for during this operation, causing the delays. It would be cool to figure out exactly what is happening, but whatever the initial issue root cause is - the cleanup still makes sense. +1.0.

PR #1815, opened 2026-06-10
test(nat-lab): no burst after waking from sleep (LLT-4961)

gytsto:
### Problem

[LLT-4961] There was no test guarding against libtelio bursting events/logs after a device wakes from sleep — the regression behind [LLT-4947] (excessive logs on wake), fixed in [LLT-4948] by switching the tokio interval missed-tick behaviour from `Burst` to `Delay`. Adding the test was deferred, and the ticket then sat blocked on "how to simulate a multi-day sleep instantly".

### Solution

Add `nat-lab/tests/test_no_burst_after_sleep.py` (parametrized for Linux, Windows and macOS) plus reusable `paused_container()` / `backing_container_id()` helpers in `docker_connection.py`.

Key insight that unblocked it: a multi-day sleep isn't needed. The `Burst`-vs-`Delay` difference is observable after missing just a handful of ticks of the 5s WireGuard-consolidation interval, and freezing a container advances `CLOCK_MONOTONIC` (the clock tokio's `Interval` uses) instantly while the process does no work.

Two complementary tests, both asserting the per-tick `WG consolidation triggered by tick event` log does **not** burst on wake:

- **`test_no_burst_after_monotonic_jump`** (Linux / Windows / macOS) — cgroup-freezes the client (`docker pause`), so `CLOCK_MONOTONIC` jumps forward and tokio sees the ticks missed while "asleep". Models process-freeze style sleep (mobile background / Android Doze, VM pause / live-migration, CPU starvation) — the real trigger of LLT-4947. Self-validating: reads the guest's `time.monotonic()` and asserts it actually jumped, so it cannot pass vacuously.
- **`test_no_burst_after_system_suspend`** (Windows / macOS VMs) — genuinely suspends the VM via the QEMU monitor (`stop`/`cont`, i.e. `virsh suspend`) and asserts the mesh reconnects with no burst. The guest clock freezes on Windows (true suspend-to-RAM, no missed ticks) and jumps on macOS (missed ticks); both must be burst-free. ACPI S3 via the guest agent is unsupported in the dockur images, so the hypervisor-level suspend is used.

No user-facing change → empty `.unreleased` entry.

Verified locally and in the nat-lab CI pipeline (green) on all three platforms.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_no_burst_after_sleep.py
+++ b/nat-lab/tests/test_no_burst_after_sleep.py
@@ -1,0 +1,287 @@
+"""Regression tests for LLT-4961 / LLT-4948 - libtelio must not replay the tokio
+interval ticks it missed while a device was asleep in a single burst.
+
+Two suspend mechanisms are exercised. tokio `Interval`/`Instant` run on
+CLOCK_MONOTONIC (QPC on Windows, mach_absolute_time on macOS), which - unlike
+CLOCK_BOOTTIME - does NOT count time the OS spent suspended (see telio-utils
+`instant.rs`, which keeps a separate boot_time clock exactly for that reason).
+Either the clock jumps forward on resume (tokio sees overdue ticks, which the
+Delay fix must not replay in a burst) or it freezes (no ticks missed, nothing to
+burst); both cases must be burst-free:
+
+* `test_no_burst_after_monotonic_jump` - the container/VM is frozen with the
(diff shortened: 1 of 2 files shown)


djkarwowski (commented):
+1

stalowyjez:
Lgtm, +1

stalowyjez (approved):

PR #1817, opened 2026-06-11
Use general test timeout in openwrt restart test

asledzinskiy:
### Problem
OpenWrt armv8 VM is slower due to full QEMU emulation so reboot takes more than for other VMs. Periodically it doesn't feet 45 seconds timeout set in the test 

### Solution
Instead of fixed 45 seconds timeout switch to general test timeout and increase armv8 test timeout to 240 seconds

- [x] README.md is updated


nat-lab/tests/test_openwrt.py
```diff
@@ -651,8 +652,7 @@ async def test_openwrt_router_restart(
             except Exception as e:  # pylint: disable=broad-exception-caught
                 log.debug("OpenWrt router is still rebooting")
                 log.debug(e)
-                if attempt < 2:
-                    await asyncio.sleep(15)
+                await asyncio.sleep(15)
```
gytsto, line 656: Maybe we can shorten this sleep to 1 sec? Or will it be too excessive?

nat-lab/tests/test_openwrt.py
```diff
@@ -52,6 +52,7 @@
         ConnectionTag.DOCKER_OPENWRT_CLIENT_2,
         ConnectionTag.VM_OPENWRT_GW_2,
         id="openwrt-24.10-armv8",
+        marks=pytest.mark.timeout(240),
```
gytsto: We have timeouts.py for pytest timeout consts, please add it there

nat-lab/tests/test_openwrt.py
asledzinskiy: Reboot takes around 30-40 seconds usually so 1 second is to often

nat-lab/tests/test_openwrt.py
asledzinskiy: Done

gytsto (approved):
+1, just one thing, PR description doesn't meet our template of problem/solution

tomasz-grz (approved):
+1

PR #1818, opened 2026-06-11
feat(nat-lab): replace WinDump with pktmon for Windows packet capture

gytsto:
### Problem
Packet capture on Windows VMs has been disabled since LLT-5942 — WinDump (WinPcap driver) was suspected of destabilizing the VMs and only ever captured one hardcoded adapter.

### Solution
Drop WinDump, capture with the in-box `pktmon` instead (`PktmonCapture`):

- all adapters present at start, full payloads, no third-party driver
- .etl converted to pcapng on the VM; download/cleanup unchanged
- SSH can't be excluded at capture time, so it's stripped post-download with local tcpdump
- 16 GB circular buffer; capture failures are non-fatal to the test run
- `TcpDump`/`build_tcpdump_command` are now Linux/Mac-only

**Known limitation:** pktmon enumerates adapters at start, so libtelio's wireguard adapter (created later) isn't captured — decrypted tunnel traffic shows up in the counterpart node's pcap instead. `--comp all` and `netsh trace` share this; fixing needs a session roll, left out of scope.

Verified live and via this PR's CI artifacts.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/tcpdump.py
+++ b/nat-lab/tests/utils/tcpdump.py
@@ -2,15 +2,14 @@ import asyncio
 import os
 import secrets
 import subprocess
-from asyncio import Event, wait_for, sleep
+from asyncio import Event, wait_for
 from contextlib import asynccontextmanager, AsyncExitStack
 from datetime import datetime
-from tests.config import WINDUMP_BINARY_WINDOWS
 from tests.utils.connection import TargetOS, Connection
 from tests.utils.connection_util import ConnectionTag
 from tests.utils.logger import log
 from tests.utils.output_notifier import OutputNotifier
-from tests.utils.process import Process
+from tests.utils.process import Process, ProcessExecError
 from tests.utils.testing import get_current_test_log_path
 from typing import AsyncIterator, Optional
 
@@ -19,10 +18,16 @@ PCAP_FILE_PATH = {
     TargetOS.Mac: "/var/root/dump.pcap",
     TargetOS.Windows: "C:\\workspace\\dump.pcap",
 }
+PKTMON_LOG_FILE_WINDOWS = "C:\\workspace\\pktmon.etl"
+# Circular-buffer cap, raised from the 512 MB default so long captures don't
+# rotate away their start (VM has ~56 GB free).
+PKTMON_MAX_LOG_SIZE_MB = 16 * 1024
 TCPDUMP_START_EVENT_TIMEOUT_S = 10
 
 
 class TcpDump:
+    """tcpdump-based packet capture for Linux and Mac; use PktmonCapture on Windows."""
+
     interfaces: Optional[list[str]]
     connection: Connection
     process: Process
@@ -43,6 +48,9 @@ class TcpDump:
         count: Optional[int] = None,
         session: bool = False,
     ) -> None:
+        assert (
+            connection.target_os != TargetOS.Windows
+        ), "tcpdump is not available on Windows, use PktmonCapture instead"
         self.connection = connection
         self.interfaces = interfaces
         self.output_file = output_file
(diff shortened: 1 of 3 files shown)


nat-lab/tests/utils/tcpdump.py
```diff
@@ -19,10 +18,19 @@
     TargetOS.Mac: "/var/root/dump.pcap",
     TargetOS.Windows: "C:\\workspace\\dump.pcap",
 }
+PKTMON_LOG_FILE_WINDOWS = "C:\\workspace\\pktmon.etl"
+# pktmon logs to a circular buffer: once the file is full, new packets
+# overwrite the oldest ones. The default cap (512 MB) could rotate away the
```
stalowyjez: Idk, I feel like the explanation what is circular buffer is not needed here

nat-lab/tests/utils/tcpdump.py
```diff
@@ -120,9 +131,115 @@ async def run(self) -> AsyncIterator["TcpDump"]:
...
+    - captures full packet payloads (pktmon truncates to 128 bytes by default)
+    - no third-party capture driver needed on the VM
+    - `pktmon stop` flushes synchronously
+
+    Limitations compared to tcpdump:
+    - pktmon enumerates adapters once, at session start, so adapters created
+      later are NOT captured. In particular libtelio's wintun/wireguard
+      adapter is created after the capture starts, which means decrypted
+      tunnel-inner traffic is missing from Windows pcaps (unlike Linux/Mac,
+      where `tcpdump -i any` picks up new interfaces). The encrypted tunnel
```
stalowyjez: Isn't that a bit severe drawback? Maybe we could reload it after that or sth?

nat-lab/tests/utils/tcpdump.py
```diff
@@ -245,6 +384,36 @@ async def make_local_tcpdump():
...
+                    # a broken capture tool must not fail the test itself
+                    # (capture on Windows already had to be disabled once for
+                    # destabilizing the VMs, see LLT-5942).
+                    log.error("Continuing without packet capture on %s", conn.tag)
+                else:
+                    raise e
+
+
 @asynccontextmanager
 async def make_tcpdump(
```
stalowyjez, line 386: Maybe this name could be changed if it might be not-tcpdump on windows

stalowyjez (commented):
Overall +1

It doesn't look as good as tcpdump, but for sure that's an improvement if we don't have a packet capture now. The comments seem to me to be a bit too verbose.

Maybe some refactor allowing running it after tunnel is on, so pktmon can see also decrypted in-tunnel traffic would make it a bit nicer?

Paintree (dismissed):
+1

nat-lab/tests/utils/tcpdump.py
gytsto: didn't want to expand scope of this, will add this in different PR if you don't mind

nat-lab/tests/utils/tcpdump.py
gytsto: It was never tcpdump anyway, so I just left it as is. Also tcpdump is so widely known and everyone knows whats the purpose of this, to basically get `pcap` :smile:

nat-lab/tests/utils/tcpdump.py
gytsto: Also with windump it only managed 1 default interface, so there was no tunnel data either. So for now I am just re-enabling original feature, with, hopefully no crashes as it used to crash with windump

nat-lab/tests/utils/tcpdump.py
gytsto: I don't mind someone reminding me general just for me

gytsto:
> Overall +1
> 
> It doesn't look as good as tcpdump, but for sure that's an improvement if we don't have a packet capture now. The comments seem to me to be a bit too verbose.
> 
> Maybe some refactor allowing running it after tunnel is on, so pktmon can see also decrypted in-tunnel traffic would make it a bit nicer?

So for tunnel data, I will do something in separate PR, since this PR is just to restore original feature with no crashes. And for the comments, they don't effect any functionality NOR mislead you also. So I'll keep verbose comments, there might someone who will be thankful for this general explanation :smile:

nat-lab/tests/utils/tcpdump.py
gytsto: after refinement, I reconsider my position and reduced amount of comments :smile:

nat-lab/tests/utils/tcpdump.py
stalowyjez: K, makes sense

stalowyjez: >didn't want to expand scope of this, will add this in different PR if you don't mind

Nit: Hmm, maybe it would be a bit cleaner in such case to have a ticket + TODOs for that? wdyt?

stalowyjez (approved):

nat-lab/tests/utils/tcpdump.py
gytsto: follow up ticket -> LLT-7429

PR #1820, opened 2026-06-11
LLT-7389: Replace "established" with "original" filter

stalowyjez:
### Problem
Incoming connection from a peer from which connections were allowed remains established after disallowing, so the packets are still accepted while they are expected to be blocked.

### Solution
Replace ESTABLISHED filter with ORIG_IP one, which matches packets for connections which were started by the owner of the IP. Add some tests to ensure that correct packets are accepted, including relevant ICMP errors.

- [x] README.md is updated

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-TRIGGERED_REF=test-orig-ip
+TRIGGERED_REF=v5.17.0
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-TRIGGERED_REF=v5.16.0
+TRIGGERED_REF=test-orig-ip
--- a/.unreleased/LLT-7389_replace_established_with_origip_filter
+++ b/.unreleased/LLT-7389_replace_established_with_origip_filter
@@ -1,0 +1,1 @@
+Close established incomming connections after meshnet remote access permission is stripped
--- a/crates/telio-firewall/src/chain_helpers.rs
+++ b/crates/telio-firewall/src/chain_helpers.rs
@@ -15,8 +15,8 @@ use crate::libfirewall::{
     LIBFW_CONTRACK_STATE_RELATED, LIBFW_DIRECTION_INBOUND, LIBFW_DIRECTION_OUTBOUND,
     LIBFW_FILTER_ASSOCIATED_DATA, LIBFW_FILTER_CONNTRACK_STATE, LIBFW_FILTER_DIRECTION,
     LIBFW_FILTER_DNS_QUERY_DOMAIN, LIBFW_FILTER_DST_NETWORK, LIBFW_FILTER_ICMP_TYPE,
-    LIBFW_FILTER_NEXT_LVL_PROTO, LIBFW_FILTER_SRC_NETWORK, LIBFW_FILTER_TCP_FLAGS,
-    LIBFW_ICMP_TYPE_DESTINATION_UNREACHABLE, LIBFW_ICMP_TYPE_ECHO_REPLY,
+    LIBFW_FILTER_NEXT_LVL_PROTO, LIBFW_FILTER_ORIG_SRC_IP, LIBFW_FILTER_SRC_NETWORK,
+    LIBFW_FILTER_TCP_FLAGS, LIBFW_ICMP_TYPE_DESTINATION_UNREACHABLE, LIBFW_ICMP_TYPE_ECHO_REPLY,
     LIBFW_ICMP_TYPE_ECHO_REQUEST, LIBFW_ICMP_TYPE_PARAMETER_PROBLEM,
     LIBFW_ICMP_TYPE_REDIRECT_MESSAGE, LIBFW_ICMP_TYPE_ROUTER_ADVERTISEMENT,
     LIBFW_ICMP_TYPE_ROUTER_SOLICITATION, LIBFW_ICMP_TYPE_TIMESTAMP,
@@ -93,6 +93,7 @@ pub(crate) enum FilterData {
     TcpFlags(u8),
     IcmpType(IcmpType),
     DnsQueryDomain(Vec<String>),
+    OrigSrcIp(IpAddr),
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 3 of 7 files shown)


.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v5.15.11
+TRIGGERED_REF=test-orig-ip
```
stalowyjez: Needs update

.unreleased/LLT-7389_replace_established_with_origip_filter
```diff
@@ -0,0 +1 @@
+Established filters replaced with Original IP one
```
jjanowsk: It would be better to have something more meaningful to the app teams in the changelog. E.g. "Close established incomming connections after meshnet remote access permission is stripped"

crates/telio-firewall/src/chain_helpers.rs
```diff
@@ -120,24 +121,28 @@ impl From<&[u8]> for LibfwAssociatedData {
     }
 }
 
+fn libfw_ip_addr(ip: IpAddr) -> LibfwIpAddr {
```
jjanowsk: Why not keep the `impl From<LibfwIpAddr> for LibfwIpAddr` convention?

crates/telio-firewall/src/firewall.rs
```diff
@@ -638,27 +638,12 @@ pub(crate) fn configure_chain(
             });
         }
 
-        // Accept packets for locally initiated connections
+        // Accept incoming packets only for connections initiated from this IP
```
jjanowsk: The comment was confusing to me at first. I guess "Accept incoming packets for connections initiated from this IP" would be clearer. With the current wording I imagine this rule in my head:
```
filter_data: FilterData::OrigSrcIp(*ip),
inverted: true
action: RuleAction::Drop,
```

sfraczek (dismissed):
+1

crates/telio-firewall/tests/firewall_integration_tests.rs
```diff
@@ -533,11 +652,11 @@ fn ipv6_blocked() {
...
-        // Should PASS (matching outgoing connections exist in LRUCache)
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src3)), is_ipv4);
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src4)), is_ipv4);
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src1)), is_ipv4);
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src2)), is_ipv4);
+        // Dropped: peer-initiated flows (peer's packets seen first), not host-initiated
+        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src3)), false);
+        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src4)), false);
+        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src1)), false);
+        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src2)), false);
```
jjanowsk, line 665: You could change the lines above a little so some packets are are actually host-initiated and are allowed to pass in this part

crates/telio-firewall/tests/firewall_integration_tests.rs
```diff
@@ -533,11 +652,11 @@ fn ipv6_blocked() {
         assert_eq!(fw.process_outbound_packet_sink(&make_peer(), &mut make_udp(src4, dst1)), is_ipv4);
         assert_eq!(fw.process_outbound_packet_sink(&make_peer(), &mut make_udp(src1, dst1)), is_ipv4);
 
-        // Should PASS (matching outgoing connections exist in LRUCache)
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src3)), is_ipv4);
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src4)), is_ipv4);
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src1)), is_ipv4);
-        assert_eq!(fw.process_inbound_packet(&make_peer(), &mut make_udp(dst1, src2)), is_ipv4);
+        // Dropped: peer-initiated flows (peer's packets seen first), not host-initiated
```
jjanowsk, line 661: Is it actually expected that the conntrack entries for dropped packets are kept alive? It seems strange tbh, but maybe I'm missing something

jjanowsk: In theory it leads to the situation where the remote side can block some rules from working by crafting malicious packets. Maybe it's a stretch but worth thinking about

jjanowsk (commented):
Looks good in general, few small comments, however I'm curious about the conntrack behavior in general (kind of not strictly related to this PR)

crates/telio-firewall/tests/firewall_integration_tests.rs
stalowyjez: That should be improved at some point, basically it's second conntrack hook implementation in libfirewall, link on slack

crates/telio-firewall/tests/firewall_integration_tests.rs
stalowyjez: Changed

crates/telio-firewall/src/firewall.rs
stalowyjez: Changed

.unreleased/LLT-7389_replace_established_with_origip_filter
stalowyjez: Changed

crates/telio-firewall/src/chain_helpers.rs
stalowyjez: Changed

jjanowsk (dismissed):
+1

tomaszklak (commented):
+0.5

jjanowsk (approved):

mathiaspeters (approved):
+0.5

PR #1822, opened 2026-06-16
Add TP-Lite tests for VPN enable/disable interactions

asledzinskiy:
### Problem
We need to check TP-Lite statistics behave correctly across VPN state changes to confirm stats collection is independent of the VPN session and survives connect/disconnect without dropping data.

### Solution
- Add test on repeatedly toggling stats on/off while being under load focusing on race conditions and missing-reference bugs
- Add 3 tests verifying TP-Lite statistics behave correctly across VPN state changes

- [x] README.md is updated


mathiaspeters (commented):
+1

nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -292,6 +303,225 @@ async def test_tp_lite_stats_requires_firewall_feature(
...
+
+        This mimics a user rapidly toggling the Threat Protection setting (and
+        the app re-applying it on top of itself) while protection is busy. The
+        goal is to catch crashes or corrupted data caused by switching the
+        feature on and off while it is under load.
+
+        Steps:
+            1. Connect to VPN
+            2. Start background DNS lookups
+            3. Repeat 20 times:
```
djkarwowski: A small thing: it looks like the docstring went out of sync with the code. `NUM_TOGGLE_ITERATIONS` is set to 10, but here it says 20.

djkarwowski (commented):
One small comment, but overall looks good.

+1

nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -72,6 +78,35 @@ async def _trigger_stats_collection(
...
+    while True:
+        try:
+            await query_dns(
+                connection,
+                domains[i % len(domains)],
+                dns_server=TP_LITE_DNS_IP,
+                options=["-type=a", "-timeout=1"],
+                quiet=True,
+            )
+        except ProcessExecError:  # blocked lookups fail; ignore the failure here
```
lcruz99: add string matching to the exception, because `ProcessExecError` can carry a error other than the expected.

nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -292,6 +303,225 @@ async def test_tp_lite_stats_requires_firewall_feature(
...
+            3. Repeat 20 times:
+                a. Turn stats on
+                b. Read the stats while reporting is in progress.
+                c. Turn stats on again without turning it off first
+                d. Turn stats off
+                e. Read the stats after turning off (must still return).
+            4. Stop the background traffic and wait for it to finish.
+            5. Confirm the feature still works: turn it on, look up a blocked
+               domain, and verify it is reported exactly once, then turn it off.
+        """
```
lcruz99, line 338: great docs!

nat-lab/tests/test_tp_lite_stats.py
asledzinskiy: done

nat-lab/tests/test_tp_lite_stats.py
asledzinskiy: done

lcruz99:
I understand these added tests target the lib stability, should they also validate the stats eg: during the iterations?

asledzinskiy:
> I understand these added tests target the lib stability, should they also validate the stats eg: during the iterations?

In case of `test_tp_lite_stats_repeated_enable_disable_under_traffic` - dns traffic is constantly generated and it is not deterministic so it is only possible to check that collected statistic is > 0 on every iteration. 

3 vpn tests are checking statistics during VPN state change but in limited scope as it is validated by other tests.

lcruz99 (approved):

PR #1823, opened 2026-06-16
LLT-7406: Link detection RTT tests

tomasz-grz:
### Problem
After the removal of Enhanced Link Detection ([PR #1498](https://github.com/NordSecurity/libtelio/pull/1498)), the public "contract" of RTT value changed, effectively halving the hold down time from `2 * (WG_KEEPALIVE + rtt)` to `(WG_KEEPALIVE + rtt) + POSSIBLE_DOWN_DELAY`.

This was not obvious and not communicated clearly to the app teams, and they reported a large jump in link down events reported from production when using the same config values.

### Solution

- Add deterministic unit tests that characterize and pin down the expected `rtt` parameter behavior. In case it silently changes in the future.
- Add direct connection scenario to link detection nat lab tests.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/link_detection.rs
+++ b/crates/telio-wg/src/link_detection.rs
@@ -527,4 +527,170 @@ mod tests {
         assert_eq!(possible_down.current_link_state(), LinkState::Up);
         assert_eq!(up.current_link_state(), LinkState::Up);
     }
+
+    /// Characterize the link state detection window for a given `rtt`,
+    /// and returns the measured hold-down after which `LinkState::Down` is reported.
+    ///
+    /// Polls once per second, like the real WireGuard UAPI cadence, and asserts:
+    /// * **within the window**: the link is reported only `Up`
+    /// * **at the deadline**: a single `Down` notify
+    /// * **after the deadline**: `Down` holds and is not re-notified.
+    async fn measure_hold_down_secs(rtt_secs: u64) -> u64 {
+        let rtt = Duration::from_secs(rtt_secs);
+        let start = Instant::now();
+        let rx = 1;
+        let tx = 1;
+
+        let stats = Arc::new(Mutex::new(BytesAndTimestamps::new(
+            Some(rx),
+            Some(tx),
+            start,
+        )));
+        time::advance(ONE_SECOND).await;
+        // A tx delta of 1 (!= KEEPALIVE_PACKET_SIZE) counts as real data and arms first_tx_after_rx.
+        stats.lock().unwrap().update(rx, tx + 1, Instant::now());
+        let first_tx = Instant::now();
+
+        let mut state = State {
+            stats: stats.clone(),
+            variant: StateVariant::Up,
+        };
+
+        // Track number of down notifications and the time of the first down state
+        let mut first_down_at: Option<u64> = None;
+        let mut down_notifications_count = 0u32;
+
+        // 5 seconds to check Down state holds
+        let down_window = 5;
+        // Down event deadline: `13 + rtt`
+        let test_duration = (WG_KEEPALIVE.as_secs() + rtt_secs)
+            + State::POSSIBLE_DOWN_DELAY.as_secs()
+            + down_window;
+
+        for _ in 0..test_duration {
+            time::advance(ONE_SECOND).await;
+            let now = Instant::now();
+            let elapsed = now.duration_since(first_tx).as_secs();
+            let result = state.update(rtt);
+            let update = result.link_detection_update_result;
+            let is_down_notify =
+                update.should_notify && matches!(update.link_state, Some(LinkState::Down));
+
+            if is_down_notify {
+                down_notifications_count += 1;
+                first_down_at.get_or_insert(elapsed);
+            }
+
+            if first_down_at.is_none() {
+                // within the rtt window only Up may be reported
+                assert!(
+                    !update.should_notify,
+                    "rtt={}: unexpected notify at {}s, before the deadline",
+                    rtt_secs, elapsed
+                );
+                assert!(
+                    matches!(update.link_state, Some(LinkState::Up)),
+                    "rtt={}: reported link_state != Up at {}s (within window)",
+                    rtt_secs,
+                    elapsed
+                );
+                assert!(
+                    matches!(state.current_link_state(), LinkState::Up),
+                    "rtt={}: current_link_state != Up at {}s (within window)",
+                    rtt_secs,
+                    elapsed
+                );
+            } else {
+                // after the deadline Down must hold
+                assert!(
+                    matches!(state.current_link_state(), LinkState::Down),
+                    "rtt={}: link not Down without inbound at {}s",
+                    rtt_secs,
+                    elapsed
+                );
+            }
+        }
+
+        assert_eq!(
+            down_notifications_count, 1,
+            "rtt={rtt_secs}: expected exactly one Down notify, got {down_notifications_count}"
+        );
+        first_down_at.unwrap_or_else(|| {
+            panic!(
+                "link never reported Down within the measurement window for rtt={}",
+                rtt_secs
+            )
+        })
+    }
+
+    #[test_log::test(tokio::test(start_paused = true))]
+    async fn test_hold_down_time() {
+        for rtt in [0u64, 5, 15, 20, 25, 30] {
+            let expected = (WG_KEEPALIVE.as_secs() + rtt) + State::POSSIBLE_DOWN_DELAY.as_secs();
+            let measured = measure_hold_down_secs(rtt).await;
+            assert_eq!(
+                measured, expected,
+                "rtt={rtt}: expected {expected}s hold-down, measured {measured}s"
+            );
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


crates/telio-wg/src/link_detection.rs
```diff
@@ -527,4 +527,155 @@ mod tests {
...
+    /// and returns the measured hold-down after which `LinkState::Down` is reported.
+    ///
+    /// Polls once per second, like the real WireGuard UAPI cadence, and asserts:
+    /// * **within the window**: the link is reported only `Up`
+    /// * **at the deadline**: a single `Down` notify
+    /// * **after the deadline**: `Down` holds and is not re-notified.
+    async fn measure_hold_down_secs(rtt_secs: u64) -> u64 {
+        let rtt = Duration::from_secs(rtt_secs);
+        let start = Instant::now();
+
```
lcruz99, line 543: create a var for each counter [rx, tx] for modification and tracking

crates/telio-wg/src/link_detection.rs
```diff
@@ -527,4 +527,155 @@ mod tests {
...
+        stats.lock().unwrap().update(1, 2, Instant::now());
+        let first_tx = Instant::now();
+
+        let mut state = State {
+            stats: stats.clone(),
+            variant: StateVariant::Up,
+        };
+
+        let mut down_at: Option<u64> = None;
+        let mut down_notifications = 0u32;
```
lcruz99: comment or make a pair/struct out of these two vars.

crates/telio-wg/src/link_detection.rs
```diff
@@ -527,4 +527,155 @@ mod tests {
...
+
+        // deadline: `13 + rtt`
```
lcruz99: replace 13 magic number

crates/telio-wg/src/link_detection.rs
```diff
@@ -527,4 +527,155 @@ mod tests {
...
+        // + 17 seconds for Down state hold
```
lcruz99: +17? I suppose you're assuming rtt = 20s? It might be what Android currently uses but that can change anytime.

crates/telio-wg/src/link_detection.rs
```diff
@@ -527,4 +527,155 @@ mod tests {
...
+
+    /// Characterize the link state detection window for a given `rtt`,
+    /// and returns the measured hold-down after which `LinkState::Down` is reported.
+    ///
+    /// Polls once per second, like the real WireGuard UAPI cadence, and asserts:
+    /// * **within the window**: the link is reported only `Up`
+    /// * **at the deadline**: a single `Down` notify
+    /// * **after the deadline**: `Down` holds and is not re-notified.
+    async fn measure_hold_down_secs(rtt_secs: u64) -> u64 {
+        let rtt = Duration::from_secs(rtt_secs);
```
lcruz99, line 539: unused `rtt`?

crates/telio-wg/src/link_detection.rs
```diff
@@ -527,4 +527,155 @@ mod tests {
...
+        let start = Instant::now();
```
lcruz99, line 540: instant also unused?

crates/telio-wg/src/link_detection.rs
tomasz-grz: used by `let result = state.update(rtt);` on line 562

crates/telio-wg/src/link_detection.rs
tomasz-grz: Used on the next line `Arc::new(Mutex::new(BytesAndTimestamps::new(Some(1), Some(1), start)));`, but I guess it can be in-lined, though I think it's a bit more readable like this.

crates/telio-wg/src/link_detection.rs
tomasz-grz: This test is tracking the time it takes for Down event, so we're not modifying `rx` and `tx` values, that's done in the next test below (with defined `rx` and `tx` variables)

crates/telio-wg/src/link_detection.rs
tomasz-grz: Added comment

crates/telio-wg/src/link_detection.rs
tomasz-grz: `rtt` is variable in this case and tested in the range `[0, 5, 15, 20]`

crates/telio-wg/src/link_detection.rs
tomasz-grz: Replaced

djkarwowski (commented):
+1

crates/telio-wg/src/link_detection.rs
lcruz99: oh missed it..

crates/telio-wg/src/link_detection.rs
lcruz99: > ... we're not modifying rx and tx values ...

`BytesAndTimestamps::new(Some(1), Some(1), start)))`
and
`stats.lock().unwrap().update(1, 2, Instant::now());`

lcruz99: replace those magic numbers

crates/telio-wg/src/link_detection.rs
lcruz99: 1. why 17s? 🤔 
2. Android is using rtt=20s, should we extend the range or is [0..20]s enough? AFAIK is not guaranteed to have rtt=20s as upper limit.

nat-lab/tests/test_events_link_state.py
```diff
@@ -131,6 +131,18 @@ def _generate_setup_parameter_pair(
...
+        _generate_setup_parameter_pair(
+            [
+                (ConnectionTag.DOCKER_CONE_CLIENT_1, TelioAdapterType.NEP_TUN),
+                (ConnectionTag.DOCKER_CONE_CLIENT_2, TelioAdapterType.NEP_TUN),
+            ],
+            direct=True,
+        ),
+    )
+]
+
```
lcruz99: This constant is only one `pytest.param()` object it doesn't cover every possible combinations, unlike `FEATURE_ENABLED_PARAMS_RELAY`. So I think it can be misleading.

Just use this pytest.param() directly on the test.

crates/telio-wg/src/link_detection.rs
tomasz-grz: 1. Just arbitrary number of "ticks" after the deadline, my intention was to check that the event stays Down, and does not get reported, but since we are manually controlling the tx and rx counts here, it will never happen, possibly `+1` is just enough 🤔
2. I can add 25 and 30. or do you recommend any other values?

nat-lab/tests/test_events_link_state.py
tomasz-grz: Updated `FEATURE_ENABLED_PARAMS`

crates/telio-wg/src/link_detection.rs
tomasz-grz: Updated

crates/telio-wg/src/link_detection.rs
tomasz-grz: 1. set to 5.
2. Added 25 and 30

crates/telio-wg/src/link_detection.rs
```diff
@@ -538,11 +538,20 @@ mod tests {
...
+            Some(initial_tx),
+            start,
+        )));
         time::advance(ONE_SECOND).await;
         // A tx delta of 1 (!= KEEPALIVE_PACKET_SIZE) counts as real data and arms first_tx_after_rx.
-        stats.lock().unwrap().update(1, 2, Instant::now());
+        stats
+            .lock()
+            .unwrap()
+            .update(initial_rx, initial_tx + 1, Instant::now());
```
lcruz99: I mean actual rx/tx counters that you increase/decrease.

```
let rx = tx = 0;

let stats = Arc::new(Mutex::new(BytesAndTimestamps::new(
    Some(rx),
    Some(tx),
    start,
)));
time::advance(ONE_SECOND).await;
// A tx delta of 1 (!= KEEPALIVE_PACKET_SIZE) counts as real data and arms first_tx_after_rx.
tx += 1;
stats
    .lock()
    .unwrap()
    .update(rx, tx, Instant::now());
```

tomasz-grz:
```rust
stats
       .lock()
       .unwrap()
       .update(initial_rx, initial_tx + 1, Instant::now());
```

tomasz-grz: This test does not accumulate `tx` so there's no point of making it `mut`

tomasz-grz:
```rust
let rx = tx = 0;
```
This is not c++ 👀

tomasz-grz: renamed to just `tx` and `rx`

lcruz99 (approved):
+1

PR #1824, opened 2026-06-17
feat(nat-lab): capture Windows tunnel adapter traffic in pcaps (LLT-7429)

gytsto:
## Problem

pktmon enumerates adapters only at session start, so libtelio's wireguard/wintun tunnel adapter (created later) was missing from Windows pcaps — decrypted tunnel traffic only showed up in the peer's pcap. Follow-up to #1818.

## Solution

- Capture with `--comp all`; poll for new adapters and roll the session (single `cmd /c "pktmon stop && pktmon start"`) when the tunnel adapter appears.
- Merge the resulting segments into one continuous `<guest>.pcap` per node, normalising in the same pass: wrap the tunnel adapter's bare-IP packets in Ethernet framing, drop the SSH control channel. Replaces the old tcpdump-based strip.

Known limitations: best-effort (depends on the tunnel adapter living long enough to roll); the brief stop→start gap shows as a small discontinuity; raw segments are kept if the merge fails.

Closes LLT-7429.

- [x] README.md is updated


tomasz-grz (commented):
+1

nat-lab/tests/utils/tcpdump.py
```diff
@@ -164,50 +202,142 @@ async def _exec_quiet(self, command: list[str], ignore_failure: bool) -> None:
...
                 str(PKTMON_MAX_LOG_SIZE_MB),
                 "--file-name",
-                PKTMON_LOG_FILE_WINDOWS,
+                etl_path,
             ],
             ignore_failure=False,
         )
+        self.log_files.append(etl_path)
+
+    async def _roll(self, current: set[str]) -> None:
```
stalowyjez: The logic of this method looks good, I'd change the name to something like `reload_pktmon_with_new_interfaces` or sth similar, so it's obvious what it does

nat-lab/tests/utils/tcpdump.py
```diff
@@ -164,50 +202,142 @@ async def _exec_quiet(self, command: list[str], ignore_failure: bool) -> None:
...
+                "[%s] failed to roll pktmon capture: %s",
+                self.connection.tag,
+                e.stderr or e.stdout,
+            )
+            return
+        self._baseline |= current
+        log.info(
+            "[%s] rolled pktmon capture to segment %d after new adapter(s)",
+            self.connection.tag,
+            self._segment,
```
stalowyjez: Maybe this new adapter name could be included? Maybe it could be also present in a segment name?

nat-lab/tests/utils/tcpdump.py
gytsto: renamed

nat-lab/tests/utils/tcpdump.py
gytsto: good suggestion about including new names into print, but in the segment I didn't. It might not be safe due special characters or spaces in adapter names (for example: Ethernet 2 or something)

stalowyjez (dismissed):
Lgtm, +1

asledzinskiy (approved):

PR #1826, opened 2026-06-19
Upgrade iperf on Windows

asledzinskiy:
### Problem
*--The iperf3 client can hang (mostly on Windows) running until the global timeout--*

### Solution
*--Upgrade iperf on Windows to the latest version that fixes hang issues--*

- [x] README.md is updated


mathiaspeters:
+1

gytsto (commented):
Code itself is OK, but why do we need those retries. We are basically covering some sort of bug or failure in the framework. I'd rather fail on 1st try and then debug the issue, than have it covered by restarts.

asledzinskiy:
> Code itself is OK, but why do we need those retries. We are basically covering some sort of bug or failure in the framework. I'd rather fail on 1st try and then debug the issue, than have it covered by restarts.

That's not our framework bug, it is Windows iperf periodically hanging.
Let me upgrade the version of iperf in dockur_windows and see if issue reproduces

gytsto (approved):
+1

PR #1830, opened 2026-06-23
refactor(nordvpnlite): remove unnecessary async from setup_logging

tomasz-kumor:
## Problem

`setup_logging` is defined as async function but in in fact it is synchronous function as it does not do any async operations.

## Solution
* Removed async keyword from `setup_logging`
* Moved `setup_logging` call before runtime `block_on()`
* Removed `Box::pin` wrapper from `daemon_event_loop` as the future returned from `daemon_event_loop` does not contain recursion, self-referential structure and pinning is not required by the API. There is no clear reason why this future should be pinned.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/main.rs
+++ b/clis/nordvpnlite/src/main.rs
@@ -82,19 +82,15 @@ fn main() -> Result<(), NordVpnLiteError> {
             }
         }
 
+        let _tracing_worker_guard = logging::setup_logging(
+            &config.log_file_path,
+            config.log_level,
+            config.log_file_count,
+        )?;
+
         // Run the daemon event loop
         let rt = tokio::runtime::Runtime::new()?;
-        rt.block_on(async {
-            // Setup daemon logging
-            let _tracing_worker_guard = logging::setup_logging(
-                &config.log_file_path,
-                config.log_level,
-                config.log_file_count,
-            )
-            .await?;
-
-            Box::pin(daemon::daemon_event_loop(config)).await
-        })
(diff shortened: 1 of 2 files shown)


clis/nordvpnlite/src/main.rs
```diff
@@ -82,19 +82,15 @@ fn main() -> Result<(), NordVpnLiteError> {
...
-        rt.block_on(async {
-            // Setup daemon logging
-            let _tracing_worker_guard = logging::setup_logging(
-                &config.log_file_path,
-                config.log_level,
-                config.log_file_count,
-            )
-            .await?;
-
-            Box::pin(daemon::daemon_event_loop(config)).await
```
tomasz-grz, line 96: Is the future small enough for the stack and there are concerns of overflows if it grows?

tomasz-grz (commented):
+1

Would be nice to have the motivation documented in the PR description of why we drop `daemon_event_loop` box.

lcruz99 (approved):
+1

clis/nordvpnlite/src/main.rs
tomasz-kumor: There is no recursion in that feature so I don't think so it is a problem. I couldn't find in the history the reason why the `Box::pin()` was added. Did we have this kind of problem before?

lcruz99: clippy emits a warning and ask for the box pinning I think when the future passes a certain threshold

djkarwowski (approved):
+1

PR #1831, opened 2026-06-23
Fail fast in netcat helpers when process exits early

lcruz99:
### Problem

The netcat helpers block indefinitely when the underlying process exits before reaching an expected state, until natlab times out.

### Solution
Wait on the event and raise NetCatError as soon as the process is gone.

- [x] README.md is updated


mathiaspeters:
+1

Paintree (approved):
+1

PR #1832, opened 2026-06-23
Ignore pre-existing ip rule when adding fwmark rule

lcruz99:
### Problem
Adding the fwmark routing rule fails when the rule already exists.

### Solution
Catch ProcessExecError and ignore the "File exists" case (logging a warning) for both the IPv4 and IPv6 rules, matching how existing routes are handled.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/utils/router/linux_router.py
+++ b/nat-lab/tests/utils/router/linux_router.py
@@ -145,20 +145,25 @@ class LinuxRouter(Router):
                         raise exception
                     log.warning(exception.stderr)
 
-            await self._connection.create_process([
-                "ip",
-                "rule",
-                "add",
-                "priority",
-                ROUTING_PRIORITY,
-                "not",
-                "from",
-                "all",
-                "fwmark",
-                FWMARK_VALUE,
-                "lookup",
-                ROUTING_TABLE_ID,
-            ]).execute()
+            try:
+                await self._connection.create_process([
+                    "ip",
+                    "rule",
+                    "add",
+                    "priority",
+                    ROUTING_PRIORITY,
+                    "not",
+                    "from",
+                    "all",
+                    "fwmark",
+                    FWMARK_VALUE,
+                    "lookup",
+                    ROUTING_TABLE_ID,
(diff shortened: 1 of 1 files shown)


jjanowsk (commented):
Why is it necessary? If the rule already exists does it mean that the cleanup of the previous test did not work as expected, or is there a legit scenario where this can happen?

lcruz99:
> Why is it necessary? If the rule already exists does it mean that the cleanup of the previous test did not work as expected, or is there a legit scenario where this can happen?

Probably on CI it isn't, but locally this is handy, preventing the restart of the whole setup.

Anyway, this exception shouldn't be the reason for a test failure.

Jauler (commented):
I kind of have the same concern like @jjanowsk.

Even if it is targeted only at local runs - could you elaborate, what leads to a case where duplicate routes are attempted to be inserted?

lcruz99:
@Jauler @jjanowsk This happens for instance when a test fails/crashes/exits earlier and doesn't properly cleanup. This change just follows the same workaround like other commands in the same function.

It makes the natlab suite more resilient, instead a restart of the whole natlab environment is required (or bad containers restart)

jjanowsk:
I'm not hard against adding it, especially that we already do have this logic for other commands in here, but in my opinion the proper answer to:  "This happens for instance when a test fails/crashes/exits earlier and doesn't properly cleanup" would be to actually make sure the test always properly cleanup. This way we just add a workaround for this one command but there may be other garbage left causing other problems elsewhere. Fixing the root cause would be preferable IMO.

lcruz99:
@jjanowsk I agree, however the "test fails/crashes/exits earlier" scenarios is mostly when something unrelated with the test happens, for instance when the pytest receives repeated SIGINTs.

jjanowsk:
+1, I don't want to block you if this really annoys you in local testing. However, handling SIGINT correctly so we can interrupt a test and get a proper cleanup would be a sweet improvement in the future ;)

Jauler (approved):
+1.0

PR #1833, opened 2026-06-23
Add libfirewall pytest marker

lcruz99:
### Problem
Libtelio doesn't panic if the libfirewall binary is not found, and when that happens consequent calls are no-op, so nat-lab tests can either fail or pass depending on their design.

### Solution
This commit registers a "libfirewall" pytest marker that skips any test when the binary is missing

- [x] README.md is updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ tcli.log
 *.nix
 *.prepared*
 *.dylib
+*.so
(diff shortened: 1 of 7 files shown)


tomasz-grz (commented):
+1 🚀

mathiaspeters (approved):
+1

PR #1834, opened 2026-06-23
Hardcode tunnel local IPv4 as a shared constant

lcruz99:
### Problem
The tunnel local address 10.5.0.2 is duplicated as a private LOCAL_IP constant in both nordvpnlite and telio-pq.

### Solution
Define it once as LOCAL_TUNNEL_IPV4 and removes the duplicated definitions.

- [x] README.md is updated

### DIFF
--- a/crates/telio-pq/src/proto.rs
+++ b/crates/telio-pq/src/proto.rs
@@ -20,12 +20,13 @@ use pnet_packet::{
 use pqcrypto_kyber::{ffi::PQCLEAN_KYBER768_CLEAN_CRYPTO_SECRETKEYBYTES, kyber768};
 use pqcrypto_traits::kem::{Ciphertext, PublicKey, SecretKey, SharedSecret};
 use rand::{rand_core::UnwrapErr, rngs::SysRng};
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn, Hidden};
 use tokio::net::{ToSocketAddrs, UdpSocket};
+
+use telio_model::constants::LOCAL_TUNNEL_IPV4;
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn, Hidden};
 
 const SERVICE_PORT: u16 = 6480;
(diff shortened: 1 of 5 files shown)


crates/telio-pq/src/proto.rs
```diff
@@ -20,12 +20,13 @@ use pnet_packet::{
...
-use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn, Hidden};
 use tokio::net::{ToSocketAddrs, UdpSocket};
 
+use telio_model::constants::LOCAL_TUNNEL_IPV4;
+use telio_utils::{telio_log_debug, telio_log_error, telio_log_warn, Hidden};
+
 const SERVICE_PORT: u16 = 6480;
 const LOCAL_PORT_RANGE: RangeInclusive<u16> = 49152..=u16::MAX; // dynamic port range
-const LOCAL_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 2);
 const REMOTE_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 1);
```
djkarwowski, line 30: `telio_model::constants` also defines:
```
/// VPN IPv4 Non-Meshnet Address
pub const VPN_EXTERNAL_IPV4: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 1);
```

Would it make sense to use that value instead of locally defined `REMOTE_IP`?

djkarwowski (commented):
+1

crates/telio-pq/src/proto.rs
lcruz99: maybe, there's also 100.64.0.1 and other constants 🤔

djkarwowski: Anyway, it's just a (somehow related) observation, so I'm resolving this thread :)

tomaszpatejko:
+1

tomaszpatejko (approved):

PR #1836, opened 2026-06-23
Allow disabling forcing plaintext DNS for TP-Lite

mathiaspeters:
### Problem
When enabling force plaintext DNS for TP-Lite stats, the firewall rule would only be set if there was a set of servers provided to force plaintext for. What that means is that if you first enable stats collection with force plaintext enabled and then enable collection again with force plaintext disabled, without disabling stats collection in between, plaintext will still be forced for the initial set of servers.

### Solution
Update firewall if either there is a new set of servers provided, or if there previously a set of servers

- [x] README.md is updated

### DIFF
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -366,11 +366,11 @@ impl StatefulFirewall {
             let mut cb = self.tp_lite_stats_cb.callback.write();
             std::mem::swap(&mut old_cb, &mut cb);
         }
-        if let Some(server_ips) = force_plaintext_dns_for_servers {
-            let mut state = self.get_state();
-            state.force_plaintext_dns_for_servers = Some(server_ips);
-            self.apply_state(state);
-        }
+
+        let mut state = self.get_state();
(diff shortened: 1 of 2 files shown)


crates/telio-firewall/src/firewall.rs
```diff
@@ -366,9 +366,11 @@ impl StatefulFirewall {
...
         }
-        if let Some(server_ips) = force_plaintext_dns_for_servers {
-            let mut state = self.get_state();
-            state.force_plaintext_dns_for_servers = Some(server_ips);
+        let mut state = self.get_state();
+        if force_plaintext_dns_for_servers.is_some()
+            || state.force_plaintext_dns_for_servers.is_some()
+        {
+            state.force_plaintext_dns_for_servers = force_plaintext_dns_for_servers;
             self.apply_state(state);
```
tomaszklak: Can we remove the `if` and always execute those 2 lines?

mathiaspeters: Sure, fixed both in this PR and the backport PR

gytsto (approved):
+1

tomaszklak (approved):
+1

PR #1839, opened 2026-06-23
LLT-6852: Advertise VPN MTU (1420) to LAN via DHCP option 26 on OpenWRT

sfraczek:
### Problem
On OpenWRT with the VPN connected, LAN clients keep MTU 1500. Towards the outside the MTU has to account for the WireGuard overhead (~80B), so it should be 1420. The router never tells clients to lower it.

### Solution
- nordvpnlite advertises DHCP option 26 (interface-mtu)=1420 on the LAN while connected, restores `dhcp.lan.dhcp_option` on disconnect.
- only Linux clients honour option 26; Windows/macOS ignore it.

- [x] README.md is updated


clis/nordvpnlite/src/interface.rs
```diff
@@ -12,6 +12,10 @@ use telio::telio_utils::LIBTELIO_FWMARK;
 // Copied from NordVPN Linux app
 const DEFAULT_ROUTING_TABLE_ID: u32 = 205;
 
+const TUNNEL_MTU: u32 = 1420;
```
matislovas: Do we always assume that underlying interface has MTU of 1500?

clis/nordvpnlite/src/interface.rs
```diff
@@ -530,6 +551,39 @@ impl Uci {
             )?;
         }
 
+        let dhcp_mtu_option_prefix = format!("{DHCP_OPTION_INTERFACE_MTU},");
+        let advertised_mtu = self
```
matislovas: There's another format supported by `dnsmasq`:
```
option:mtu,1420
```

clis/nordvpnlite/src/interface.rs
sfraczek: That is how I understood the requirements. Previously it was already assuming MTU of 1500. I can try to implement it as derived from the WAN MTU.

matislovas: I think we had this discussion previously. Maybe @Jauler  or @tomasz-grz will remember ..?

clis/nordvpnlite/src/interface.rs
```diff
@@ -530,6 +551,39 @@ impl Uci {
...
+            .next_back()
+            .map_or(TUNNEL_MTU, |existing| existing.min(TUNNEL_MTU));
+        debug!(
+            "Advertising MTU {advertised_mtu} to LAN via DHCP option {DHCP_OPTION_INTERFACE_MTU}"
+        );
+        execute(Command::new("uci").args(["delete", "dhcp.lan.dhcp_option"]))
+            .map_err(|e| {
+                debug!("uci couldn't delete dhcp.lan.dhcp_option: {}", e);
+            })
+            .ok();
```
Paintree: This is multiplied in 2 places in the same form, or in 4 places in similar form. Maybe it would be worth cleaning up a bit?

Paintree (dismissed):
+1

clis/nordvpnlite/src/interface.rs
sfraczek: implemented as derived

clis/nordvpnlite/src/interface.rs
sfraczek: thanks. done.

clis/nordvpnlite/src/interface.rs
sfraczek: done

matislovas:
+1

matislovas (dismissed):

Paintree (dismissed):

matislovas (approved):

PR #1842, opened 2026-06-24
Add tp-lite test on force_plaintext_dns flag

asledzinskiy:
### Problem
*--Tp-lite force_plaintext_dns  flag needs to be covered with nat-lab test--*

### Solution
*--Add test checking that when force_plaintext_dns is enabled traffic is blocked to non plain text dns--*

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_tp_lite_stats.py
+++ b/nat-lab/tests/test_tp_lite_stats.py
@@ -12,7 +12,7 @@ from tests.utils.bindings import (
     DnsRedirect,
 )
 from tests.utils.connection import ConnectionTag
-from tests.utils.dns import query_dns
+from tests.utils.dns import query_dns, query_dns_port
 from tests.utils.logger import log
 from tests.utils.process import ProcessExecError
 from typing import Optional
@@ -25,6 +25,10 @@ BLOCKED_NXDOMAIN = "blocked-malware.com"
 BLOCKED_NULL_IP = "blocked-ads.com"
 # Resolves normally on the TP-Lite server
 ALLOWED_DOMAIN = "google.com"
+# Address the TP-Lite server returns for ALLOWED_DOMAIN
+ALLOWED_DOMAIN_IP = "142.250.179.206"
+
+NON_PLAINTEXT_DNS_PORTS = ["443", "853"]
 
 # Standard (non-blocking) DNS server. Unlike the TP-Lite server it resolves
 # blocked-malware.com instead of returning NXDOMAIN.
@@ -49,7 +53,10 @@ def _features_with_firewall():
     return features
 
 
-def _tp_lite_config(dns_server_ips: Optional[list[str]] = None):
+def _tp_lite_config(
+    dns_server_ips: Optional[list[str]] = None,
+    force_plaintext_dns: Optional[bool] = None,
+):
     return TpLiteStatsOptions(
         dns_server_ips=(
             dns_server_ips if dns_server_ips is not None else [TP_LITE_DNS_IP]
@@ -58,7 +65,7 @@ def _tp_lite_config(dns_server_ips: Optional[list[str]] = None):
         blocked_domains_buffer_size=None,
         cache_size=None,
         max_open_requests=None,
-        force_plaintext_dns=None,
+        force_plaintext_dns=force_plaintext_dns,
     )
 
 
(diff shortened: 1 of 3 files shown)


nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -210,6 +215,64 @@ async def test_tp_lite_stats_basic(
...
+
+        Steps:
+            1. Connect to VPN.
+            2. Enable stats collection with force_plaintext_dns=False, query the DNS server
+               on port 853 and verify it is allowed through.
+            3. Enable stats collection with force_plaintext_dns=True.
+            4. Query the DNS server on port 853 and verify it is now blocked.
+            5. Query a blocked domain over plaintext (port 53), trigger a stats
+               flush, and verify it is reported exactly once.
+            6. Disable stats collection.
```
mathiaspeters: It would be nice to add steps here to again enable stats collection with force_plaintext_dns=False and try to query port 853 to verify that it works again. There was a bug that's being fixed in https://github.com/NordSecurity/libtelio/pull/1836 where DoT/DoH would still be blocked. And as a separate but related point, it would be nice to also query port 853 on the other DNS server to verify that it's only the specified server that has plaintext forced

asledzinskiy: Thanks, will add that

asledzinskiy: Done. Test will be failing until #1836  is merged

nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -210,6 +215,80 @@ async def test_tp_lite_stats_basic(
...
+        [pytest.param(_alpha_setup_params_with_firewall())],
+    )
+    @pytest.mark.asyncio
+    async def test_tp_lite_force_plaintext_dns(
+        self,
+        alpha_setup_params: SetupParameters,  # pylint: disable=unused-argument
+        env: Environment,
+    ) -> None:
+        """With force_plaintext_dns enabled, the firewall blocks traffic to the
+        configured DNS server on any port other than 53 (e.g. DoT/DoH), so only
```
matislovas, line 233: We're testing plaintext and `DoT` here, but are not touching the `DoH` case (port 443). Is this by design?

nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -25,6 +25,8 @@
 BLOCKED_NULL_IP = "blocked-ads.com"
 # Resolves normally on the TP-Lite server
 ALLOWED_DOMAIN = "google.com"
+# Address the TP-Lite server returns for ALLOWED_DOMAIN
+ALLOWED_DOMAIN_IP = "142.250.179.206"
```
matislovas, line 29: Can we somehow share this address and `google.com` domain between test file and both dns-server binaries?

nat-lab/tests/test_tp_lite_stats.py
asledzinskiy: force_plaintext_dns checks port only and everything that is not `port 53` is blocked so it seems to me that adding a check for 443 port doesn't make sense as functionally it is the same code logic as for 853 port.

matislovas: I mean, this 443 is a standard for `https` (and not standard for `DNS`), so I guess it would be a good idea to check also, if it is not a big hustle. If it is - than we can assume that `853` assumes `all other ports, except 53`.

asledzinskiy: I'll check it. Shouldn't be a big hustle, should be able to start tp-lite dns server on 443 as well

nat-lab/tests/test_tp_lite_stats.py
asledzinskiy: I can import that IP from `/bin/dns-server` - but we don't do such imports anywhere in tests as `/bin` folder is separate from tests and used for infrastructure only

nat-lab/tests/test_tp_lite_stats.py
asledzinskiy: Done

matislovas:
+1

mathiaspeters (dismissed):

mathiaspeters (approved):

PR #1844, opened 2026-06-25
LLT-6736: Add command 'reload' and support SIGHUP

tomasz-kumor:
### Problem
Currently to change the nordvpnlite configuration while the deamon is running we need to stop it, change the config file and restart the daemon.

```
./nordvpnlite stop
// change config.json file
./nordvpnlite start -c ~/config.json
```

The packets may leak during this operation.

### Solution
This PR doesn't address the packets leak problem yet but prepares a base for future changes.
It introduces:
* CLI `reload` command support
* SIGHUP signal support

Both triggers causes the the daemon to restart with a new configuration.

### Verification

Happy path verification

### 

* CLI reload
```
nordvpnlite start -c ~/config.json
nordvpnlite status
// change anything in the config.json (i.e. country code/log level)
nordvpnlite reload
nordvpnliste status

```

* SIGHUP signal
```
nordvpnlite start -c ~/config.json
nordvpnlite status
// change anything in the config.json
kill -HUP $(pidof nordvpnlite)
nordvpnliste status
```

Daemon does not restart when config.json not changed
 ```
nordvpnlite start -c ~/config.json
kill -HUP $(pidof nordvpnlite) // nordvpnlite reload
// check /var/log/nordvpnlite.log if telio daemon did not restart
```

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/nordvpnlite.init
+++ b/clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/nordvpnlite.init
@@ -22,7 +22,6 @@ stop_service() {
 }
 
 reload_service() {
-  stop
-  sleep 1
-  start
+  echo "Reloading nordvpnlite configuration"
+  /usr/sbin/nordvpnlite reload
 }
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -4,11 +4,11 @@ use clap::Parser;
 use serde::{Deserialize, Serialize};
 use telio_core::telio_task::io::chan;
 use tokio::sync::oneshot;
-use tracing::{error, trace};
+use tracing::{error, info, trace};
 
 use crate::{
     comms::{DaemonConnection, DaemonSocket},
-    config::Endpoint,
+    config::{Endpoint, RunningConfig},
     daemon::{NordVpnLiteError, TelioStatusReport},
 };
 
@@ -24,6 +24,8 @@ pub enum ClientCmd {
     IsAlive,
     #[clap(name = "stop", about = "Stop daemon execution")]
     QuitDaemon,
+    #[clap(name = "reload", about = "Reload config file and restart the daemon")]
+    Reload,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -126,14 +128,27 @@ pub struct CommandListener {
     socket: DaemonSocket,
     /// Channel to send commands to telio task
     telio_task_tx: chan::Tx<TelioTaskCmd>,
+    config: RunningConfig,
+    /// Holds the new running config from a client triggered reload
+    pending_config: Option<RunningConfig>,
 }
 
 impl CommandListener {
-    pub fn new(socket: DaemonSocket, telio_task_tx: chan::Tx<TelioTaskCmd>) -> CommandListener {
+    pub fn new(
+        socket: DaemonSocket,
+        telio_task_tx: chan::Tx<TelioTaskCmd>,
+        config: RunningConfig,
+    ) -> CommandListener {
         CommandListener {
             socket,
             telio_task_tx,
+            config,
+            pending_config: None,
         }
+    }
+
+    pub fn take_pending_config(&mut self) -> Option<RunningConfig> {
+        self.pending_config.take()
     }
 
     // Main command handling communicating with telio task
@@ -175,6 +190,33 @@ impl CommandListener {
                 // cleanup
                 handle_response(response_rx, |_| Ok(CommandResponse::Ok)).await
             }
+            ClientCmd::Reload => {
+                match RunningConfig::from_file(&self.config.path) {
+                    Err(e) => {
+                        error!("Config file changed but failed to parse: {e}");
+                        Ok(CommandResponse::Err(e.to_string()))
+                    }
+                    Ok(new_config) if new_config.hash == self.config.hash => {
+                        info!("Config file unchanged, ignoring reload request");
+                        Ok(CommandResponse::Ok)
+                    }
+                    Ok(new_config) => {
+                        trace!("Config file changed, proceeding with reload");
+                        self.pending_config = Some(new_config);
+                        let (response_tx, response_rx) = oneshot::channel();
+                        #[allow(mpsc_blocking_send)]
+                        self.telio_task_tx
+                            .send(TelioTaskCmd::Quit(response_tx))
+                            .await
+                            .map_err(|e| {
+                                error!("Error sending command: {}", e);
+                                NordVpnLiteError::CommandFailed(ClientCmd::Reload)
+                            })?;
+                        // Wait for teardown to complete before responding to the client
+                        handle_response(response_rx, |_| Ok(CommandResponse::Ok)).await
+                    }
+                }
+            }
             ClientCmd::IsAlive => Ok(CommandResponse::Ok),
         }
     }
@@ -211,7 +253,9 @@ impl CommandListener {
                     match &command {
                         ClientCmd::QuitDaemon => CommandResponse::Ok,
                         ClientCmd::IsAlive => CommandResponse::Ok,
-                        ClientCmd::GetStatus => CommandResponse::DaemonInitializing,
+                        ClientCmd::GetStatus | ClientCmd::Reload => {
+                            CommandResponse::DaemonInitializing
+                        }
                     }
                 };
                 connection.respond(response.serialize()).await?;
@@ -236,6 +280,7 @@ mod tests {
     use crate::{CommandResponse, DaemonSocket};
     use assert_matches::assert_matches;
     use std::path::Path;
+    use temp_file::TempFile;
     use tokio::{
         io::{AsyncReadExt, AsyncWriteExt},
         net::UnixStream,
@@ -245,13 +290,34 @@ mod tests {
 
     const TEST_SOCKET_PATH: &str = "test_socket";
 
+    // A minimal valid NordVpnLiteConfig JSON used across reload tests.
+    const VALID_CONFIG_JSON: &str = r#"{
+        "log_level": "Info",
+        "log_file_path": "test.log",
+        "adapter_type": "linux-native",
+        "interface": { "name": "utun10", "config_provider": "manual" },
+        "authentication_token": "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+    }"#;
+
     // Create a random socket path for the test, since tests run in parallel they can deadlock
     fn make_socket_path() -> String {
         format!("{}_{}", TEST_SOCKET_PATH, rand::random::<u16>())
     }
 
-    // Helper to create a fake command listener
-    fn make_command_listener(path: &str) -> CommandListener {
+    // Write VALID_CONFIG_JSON to a temp file and load it as a RunningConfig.
+    // Returns both so the TempFile stays alive (and thus the file exists) for the duration of the test.
+    fn make_running_config() -> (RunningConfig, TempFile) {
+        let file = TempFile::new()
+            .unwrap()
+            .with_contents(VALID_CONFIG_JSON.as_bytes())
+            .unwrap();
+        let config = RunningConfig::from_file(file.path()).unwrap();
+        (config, file)
+    }
+
+    // Helper to create a fake command listener.
+    // Returns the listener and the TempFile backing its RunningConfig so the file stays alive.
+    fn make_command_listener(path: &str) -> (CommandListener, TempFile) {
         let (tx, mut task_rx) = mpsc::channel(1);
 
         // spawn a fake telio task
@@ -270,7 +336,34 @@ mod tests {
 
         let socket = DaemonSocket::new(Path::new(path)).unwrap();
 
-        CommandListener::new(socket, tx)
+        let (config, config_file) = make_running_config();
+        (CommandListener::new(socket, tx, config), config_file)
+    }
+
+    fn make_reload_listener(
+        path: &str,
+    ) -> (CommandListener, mpsc::Receiver<TelioTaskCmd>, TempFile) {
+        let (tx, task_rx) = mpsc::channel(1);
+        let socket = DaemonSocket::new(Path::new(path)).unwrap();
+        let (config, config_file) = make_running_config();
+        (
+            CommandListener::new(socket, tx, config),
+            task_rx,
+            config_file,
+        )
+    }
+
+    fn spawn_quit_responder(
+        mut task_rx: mpsc::Receiver<TelioTaskCmd>,
+    ) -> tokio::task::JoinHandle<bool> {
+        tokio::spawn(async move {
+            if let Some(TelioTaskCmd::Quit(tx)) = task_rx.recv().await {
+                tx.send(()).unwrap();
+                true
+            } else {
+                false
+            }
+        })
     }
 
     // Simulate client sending command and waiting for response
@@ -305,7 +398,7 @@ mod tests {
         Result<ClientCmd, NordVpnLiteError>,
     ) {
         let path = make_socket_path();
-        let mut listener = make_command_listener(&path);
+        let (mut listener, _config_file) = make_command_listener(&path);
 
         let command = serde_json::to_string(&command).unwrap();
         let daemon = tokio::spawn(async move {
@@ -364,7 +457,7 @@ mod tests {
     #[tokio::test]
     async fn test_command_invalid() {
         let path = make_socket_path();
-        let mut listener = make_command_listener(&path);
+        let (mut listener, _config_file) = make_command_listener(&path);
 
         let command = "garbage";
         let daemon = tokio::spawn(async move {
@@ -381,7 +474,7 @@ mod tests {
     #[tokio::test]
     async fn test_command_invalid_broken() {
         let path = make_socket_path();
-        let mut listener = make_command_listener(&path);
+        let (mut listener, _config_file) = make_command_listener(&path);
 
         let command = "garbage";
         let daemon = tokio::spawn(async move {
@@ -423,5 +516,95 @@ mod tests {
 
         assert_eq!(cmd.unwrap(), ClientCmd::GetStatus);
         assert_eq!(response.unwrap(), CommandResponse::DaemonInitializing);
+    }
+
+    #[tokio::test]
+    async fn test_command_reload() {
+        let (response, cmd) = test_command_helper(ClientCmd::Reload, true, false).await;
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Reload);
+    }
+
+    #[tokio::test]
+    async fn test_command_early_reload() {
+        let (response, cmd) = test_command_helper(ClientCmd::Reload, false, false).await;
+
+        assert_eq!(cmd.unwrap(), ClientCmd::Reload);
+        assert_eq!(response.unwrap(), CommandResponse::DaemonInitializing);
+    }
+
+    #[tokio::test]
+    async fn test_command_reload_unchanged() {
+        use tokio::time::{timeout, Duration};
+
+        let path = make_socket_path();
+        let (mut listener, task_rx, _config_file) = make_reload_listener(&path);
+
+        let command = serde_json::to_string(&ClientCmd::Reload).unwrap();
+        let daemon = tokio::spawn(async move {
+            let connection = listener.accept_client_connection().await.unwrap();
+            listener.handle_client_command(true, connection).await
+        });
+
+        let responder = spawn_quit_responder(task_rx);
+
+        let response = timeout(Duration::from_secs(3), client_send_command(&path, &command))
+            .await
+            .expect("test timed out waiting for client response");
+        let cmd = timeout(Duration::from_secs(3), daemon)
+            .await
+            .expect("test timed out — daemon task hung")
+            .unwrap();
+        let quit_was_incorrectly_sent = responder.await.unwrap();
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Reload);
+        assert!(
+            !quit_was_incorrectly_sent,
+            "TelioTaskCmd::Quit must not be sent when config is unchanged"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_command_reload_changed() {
+        use tokio::time::{timeout, Duration};
+
+        let socket_path = make_socket_path();
+        let (mut listener, task_rx, config_file) = make_reload_listener(&socket_path);
+
+        let alt_config = VALID_CONFIG_JSON.replace("utun10", "utun11");
+        std::fs::write(config_file.path(), alt_config.as_bytes()).unwrap();
+
+        let command = serde_json::to_string(&ClientCmd::Reload).unwrap();
+        let daemon = tokio::spawn(async move {
+            let connection = listener.accept_client_connection().await.unwrap();
+            let result = listener.handle_client_command(true, connection).await;
+            let has_pending = listener.take_pending_config().is_some();
+            (result, has_pending)
+        });
+
+        let responder = spawn_quit_responder(task_rx);
+
+        let response = client_send_command(&socket_path, &command).await;
+        let (cmd, had_pending) = timeout(Duration::from_secs(3), daemon)
+            .await
+            .expect("test timed out — daemon task hung")
+            .unwrap();
+        let quit_was_sent = timeout(Duration::from_secs(3), responder)
+            .await
+            .expect("test timed out — responder task hung")
+            .unwrap();
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Reload);
+        assert!(
+            quit_was_sent,
+            "TelioTaskCmd::Quit must be sent when config has changed"
+        );
+        assert!(
+            had_pending,
+            "pending_config must be set after a successful reload"
+        );
     }
 }
--- a/clis/nordvpnlite/src/config.rs
+++ b/clis/nordvpnlite/src/config.rs
@@ -1,6 +1,6 @@
 use std::{
     fs,
-    io::Write,
+    io::{BufReader, Write},
     net::IpAddr,
     net::Ipv4Addr,
     os::unix::fs::PermissionsExt,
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: clis/nordvpnlite/src/command_listener.rs [hunk 1]
```diff
@@ -125,16 +127,31 @@ pub struct CommandListener {
...
+    pub fn new(
+        socket: DaemonSocket,
+        telio_task_tx: chan::Tx<TelioTaskCmd>,
+        config_file: ConfigFile,
+    ) -> CommandListener {
         CommandListener {
             socket,
             telio_task_tx,
+            config_file,
+            pending_config: None,
```
tomaszpatejko, line 146: What is the difference between contents of `config_file` and `pending_config`? I can see they both store hash value.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -341,15 +344,61 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
...
+    /// Carries the already parsed new config and the new config file hash
+    Reload(Box<NordVpnLiteConfig>, u64),
+}
+
+pub async fn daemon_event_loop(
+    mut config: NordVpnLiteConfig,
+    config_path: String,
+    mut logging_handle: logging::LoggingHandle,
+) -> Result<(), NordVpnLiteError> {
+    let mut config_file = ConfigFile::new(config_path);
```
tomaszpatejko: Does it make sense to calculate hash lazily, on demand when it's known that reload actually happened? If I'm not mistaken hashes and `config_file` variable are only needed when configuration file needs to be reloaded.

clis/nordvpnlite/src/logging.rs
```diff
@@ -2,18 +2,26 @@ use std::path::Path;
...
+use tracing_subscriber::{reload, Layer, Registry};
 
 use crate::NordVpnLiteError;
 
-pub fn setup_logging<P: AsRef<Path>>(
+type DynLayer = Box<dyn Layer<Registry> + Send + Sync>;
+
+pub struct LoggingHandle {
+    _worker_guard: WorkerGuard,
+    layer_reload_handle: reload::Handle<DynLayer, Registry>,
```
stalowyjez: Maybe `fmt_layer_reload_handle`? Basically both of them are for reloading some layer, so maybe their names should be more consistent?

clis/nordvpnlite/src/logging.rs
```diff
@@ -37,18 +45,68 @@ pub fn setup_logging<P: AsRef<Path>>(
...
+        .filter_reload_handle
+        .modify(|level| *level = log_level)
+        .map_err(|e| NordVpnLiteError::TracingError(e.to_string()))?;
+
+    handle
+        .layer_reload_handle
+        .modify(|layer| *layer = new_layer)
+        .map_err(|e| NordVpnLiteError::TracingError(e.to_string()))?;
+
+    drop(std::mem::replace(&mut handle._worker_guard, new_guard));
```
stalowyjez: Do you need explicit drop here? I mean, if it is replaced, shouldn't the old value be dropped by the end of the function anyway?

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: In this place I am calculating a hash for an initial config file. When reload is requested the hash is calculated one more time and compared with that initial value to check if config file has changed. If not, reload it not continued.

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: config_file hold a path and a `hash` of an already applied config file (prio to reload or from a previous reload). 
When reload is requested, the `pending_config` holds a new configuration `NordVpnLiteConfig` and a `hash` of a config file from which that configuration was obtained. 
During a reload the config file is read to a memory once. Then the hash is calculated and compared with the previous value. If hash is changed the configuration is parsed and stored in `pending_config` and reload continues.

tomaszpatejko: Understood.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -448,10 +502,30 @@ pub async fn daemon_event_loop(mut config: NordVpnLiteConfig) -> Result<(), Nord
...
             },
-            // Handle interrupt signals for clean shutdown
+            // Handle interrupt signals for clean shutdown or reload
             signal = signals.next() => {
                 match signal {
-                    Some(s @ SIGHUP | s @ SIGTERM | s @ SIGINT | s @ SIGQUIT) => {
+                    Some(SIGHUP) => {
+                        match config_file.read_if_changed() {
+                            None => {
+                                trace!("Config file unchanged, ignoring reload request");
```
stalowyjez: Should that be only trace? I think it might be valuable to a user who sent this SIGHUP, maybe info would be better?

tomaszpatejko:
+1

tomaszpatejko (approved):

tomaszpatejko (dismissed):

clis/nordvpnlite/src/config.rs
```diff
@@ -19,6 +19,53 @@ use telio::crypto::PublicKey;
 
 pub(crate) const LOCAL_IP: Ipv4Addr = Ipv4Addr::new(10, 5, 0, 2);
 
+#[derive(Clone)]
+pub struct ConfigFile {
+    pub path: String,
+    hash: u64,
+}
```
stalowyjez: I'm not sure if that's really needed, maybe making Config hashable and comparing them would be simpler?

stalowyjez (commented):
Lgtm, just a few minor comments/questions

clis/nordvpnlite/src/logging.rs
tomasz-kumor: Ok I will change it. Previously I had a simple `handle._worker_guard = new_guard` but after AI review I changed it. Unfortunately I don't remember and can't find what was the motivation for that change.
```

File: clis/nordvpnlite/src/config.rs (hunk 6 above)
Comment (tomasz-kumor) [reply clis/nordvpnlite/src/config.rs]: Could you explain it. Do you mean to make a `NordVpnLiteConfig` structure hashable? In the task description I have a point "The reload command should check the config file’s hash - if it has changed, than we should execute the actual reload."  The benefit of this approach it, that we are not parsing a config file if it's content did not change.

Comment (stalowyjez) [reply clis/nordvpnlite/src/config.rs]: >Do you mean to make a NordVpnLiteConfig structure hashable?

Yes, that's what I meant. But now I think just implementing `Eq` would be easier when considering this path

>The benefit of this approach it, that we are not parsing a config file if it's content did not change.

That's right. On the other hand with "just compare configs" you don't need to hash the file and if there are some differences in files which lead to the same configs we won't reload it which also sounds beneficial, I think that keeping stuff simple is more important argument here. I.e. it's not that I'm completely against hashing approach, but I'm just not sure if this added complexity is really justified

Comment (stalowyjez) [reply clis/nordvpnlite/src/config.rs]: I.e. comparing to I/O operations of reading the file and reloading nordvpnlite with a new config this parsing sound like a quite cheap thing to me

Comment (tomasz-kumor) [reply clis/nordvpnlite/src/config.rs]: It is not possible to implement `Eq` for `NordVpnLiteConfig` becasue of the `AdapterType::Custom(Arc<dyn Adapter>)` that uses `dyn trait` that doesn't implement `Eq`.  `Eq` was removed from `NordVpnLiteConfig` by

```
commit 6b73171e96b4a40361f2eb226e4d1476557d64b9
Author: Tomasz Kłak <tomasz.klak@nordsec.com>
Date:   Fri Oct 17 14:01:32 2025 +0200

    Add a custom WG adapter type
    
    This allows for any application that uses libtelio to provide an WG
    configuration mechanism that is different from one of the few built in
    ones.
```

Comment (stalowyjez) [reply clis/nordvpnlite/src/config.rs]: Ok, so it's probably hard to make it hashable too

Comment (tomasz-kumor) [reply clis/nordvpnlite/src/config.rs]: yes the same problem

File: clis/nordvpnlite/src/daemon.rs [hunk 7]
```diff
@@ -348,15 +351,61 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
...
+                // loop continues and run_daemon called again with new config
+            }
+        }
+    }
+    Ok(())
+}
+
+async fn run_daemon(
+    mut config: NordVpnLiteConfig,
+    config_file: &ConfigFile,
```
tomasz-grz: Could these two be unified? It's not very obvious what are the differences between the two parameters

clis/nordvpnlite/src/daemon.rs
```diff
@@ -348,15 +351,61 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
...
+/// whether to exit or reload the config and restart.
+enum LoopOutcome {
+    Exit,
+    /// Carries the already parsed new config and the new config file hash
+    Reload(Box<NordVpnLiteConfig>, u64),
+}
+
+pub async fn daemon_event_loop(
+    mut config: NordVpnLiteConfig,
+    config_path: String,
```
tomasz-grz: `P: AsRef<Path>` so the function could take any path parameter, or `PathBuf` if we explicitly want it owned?

clis/nordvpnlite/src/config.rs
```diff
@@ -20,6 +20,53 @@ use telio::{
 
 use crate::{interface::InterfaceConfig, NordVpnLiteError};
 
+#[derive(Clone)]
+pub struct ConfigFile {
+    pub path: String,
```
tomasz-grz: What do you think about [PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) for a more ergonomic way of working with owned paths?

tomasz-kumor:
@stalowyjez do you have any other comments?

clis/nordvpnlite/src/main.rs
```diff
@@ -82,15 +82,19 @@ fn main() -> Result<(), NordVpnLiteError> {
...
             config.log_level,
             config.log_file_count,
         )?;
 
         // Run the daemon event loop
         let rt = tokio::runtime::Runtime::new()?;
-        rt.block_on(daemon::daemon_event_loop(config))
+        rt.block_on(daemon::daemon_event_loop(
+            config,
+            opts.config_path.clone(),
```
tomasz-grz: How does this work with relative config and log paths when we detach? (since the daemon changes working directory, either to `/` or `opts.working_directory`)

Do we need to convert to canonical paths before detaching?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -174,6 +191,33 @@ impl CommandListener {
...
             }
+            ClientCmd::Reload => {
+                match self.config_file.read_if_changed() {
+                    None => {
+                        trace!("Config file unchanged, ignoring reload request");
+                        Ok(CommandResponse::Ok)
+                    }
+                    Some(Err(e)) => {
+                        error!("Config file changed but failed to parse: {e}");
+                        Err(e)
```
tomasz-grz: Would `Ok(CommandResponse::Err(e))` here return the error to the client caller?

clis/nordvpnlite/src/config.rs
```diff
@@ -20,6 +20,53 @@ use telio::{
...
+        self.hash = hash;
+    }
+
+    fn hash_bytes(bytes: &[u8]) -> u64 {
+        use std::collections::hash_map::DefaultHasher;
+        use std::hash::{Hash, Hasher};
+        let mut hasher = DefaultHasher::new();
+        bytes.hash(&mut hasher);
+        hasher.finish()
+    }
```
tomasz-grz: I wonder how would this approach handle whitespace or formatting changes in the config file?
Would it trigger reload causing a reconnect? 

Maybe we could just use `PartialEq` to compare the pending config against the current config?

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/nordvpnlite.init
```diff
@@ -22,7 +22,6 @@ stop_service() {
 }
 
 reload_service() {
-  stop
-  sleep 1
-  start
+  echo "Reloading nordvpnlite configuration"
+  /usr/sbin/nordvpnlite reload
```
tomasz-grz, line 26: This would no longer restart a stopped daemon, is that an acceptable change?

Should we do `/usr/sbin/nordvpnlite reload || start` as a fallback?

clis/nordvpnlite/src/daemon.rs
```diff
@@ -348,15 +351,61 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
...
+) -> Result<(), NordVpnLiteError> {
+    let mut config_file = ConfigFile::new(config_path);
+
+    loop {
+        match run_daemon(config.clone(), &config_file).await? {
+            LoopOutcome::Exit => break,
+            LoopOutcome::Reload(new_config, hash) => {
+                info!("Reloading config from {}", config_file.path);
+                config = *new_config;
+                config.resolve_env_token();
```
tomasz-grz: Just a thought

If we reload a running daemon, with a valid config but invalid token, it will take down the active connection and kill the daemon.

Is this acceptable risk? 
We could validate the new token before reloading for example, though I would probably add this as a separate task.

tomasz-grz (commented):
Thanks for tackling this, nice work!
Though I'll add some more critical thoughts from my side here..

Also would be nice to see some `nat-lab` test cases for the reload scenarios (at least the happy path) 🙏

clis/nordvpnlite/src/config.rs
tomasz-kumor: With `PartialEq` the problem is that changes to AdapterType::Custom will not be catch. https://github.com/NordSecurity/libtelio/pull/1844#discussion_r3512110993

clis/nordvpnlite/openwrt/feed/net/nordvpnlite/files/nordvpnlite.init
tomasz-kumor: I am not sure if reload should work this way. If daemon is stopped then there is no need to call `reload` after changing the configuration

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: I don't think that we should add such a check. If a user change a configuration and provide invalid access token it it his/her fault. I don't think that it is a risk. Reloading nordvpnlite with a wrong configuration will cause a failure.

clis/nordvpnlite/src/main.rs
tomasz-kumor: For `log_file_path` there is already a statement in readme:
```
log_file_path - a path to store the daemon's logs, needs be absolute, otherwise will be relative to working-directory when daemonized
```

I assumed similar rule for `config-file`. 

IMO it make sense that relative path is determined towards the working directory.

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: I was thinking about that but wasn't sure if I should. What is in the `ConfigFile` describes the config file, path to it and it's hash.  `NordVpnLiteConfig` contains parsed content of a config file.  That's why I wasn't sure if I should mix them.

clis/nordvpnlite/src/config.rs
tomasz-kumor: Adding a whitespace to a file should trigger a reload. At least that was my team intention during this task refinement.

clis/nordvpnlite/src/main.rs
tomasz-grz: yes but the working directory of the daemon changes after it's "detached".
So if you launch `nordvpnlite start -c config.json` from `/home/user/`, the daemon later tries to read `/config.json` on reload, instead of `/home/user/config.json`.
We probably need to call [canonicalize()](https://doc.rust-lang.org/beta/std/path/struct.PathBuf.html#method.canonicalize) before passing the `config_path` into the `daemon_event_loop`

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: here is what we get in log file:
```
2026-07-03T10:59:21.854830Z ERROR nordvpnlite::command_listener: 201: Config file changed but failed to parse: invalid type: unit variant, expected newtype variant at line 12 column 13
```

clis/nordvpnlite/src/main.rs
tomasz-kumor: Yes, you are correct. But is such case, when config file is relative, shouldn't the user launch the app with `--working-directory /home/user/`? I assumed that we should follow similar logic as for `log_file_path`

clis/nordvpnlite/src/config.rs
tomasz-grz: Yes, but that means it will trigger a reconnect -> packet leak, even though nothing changed in the config functionally.
If you are aware of that and the risk is accepted, then it's fine by me.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Yes but what do you get in the cli calling `nordvpnlite reload`? And does the daemon keep running the old config?

clis/nordvpnlite/src/main.rs
tomasz-grz: Actually [log path is canonicalized](https://github.com/NordSecurity/libtelio/blob/f5b419e90d02fba7711192b8debc5b52e95f8bdf/clis/nordvpnlite/src/config.rs#L158) when reading the config, but canonicalizing again after detach would result a different path.

tomasz-grz: Before the changes, the config and log path were resolved only once before "detaching" the daemon.

So if you `nordvpnlite start -c config.json` from `/home/user/`, the config is read and parsed from `/home/user/config.json`.

After `nordvpnlite reload`
The daemon will try to read the config from `/config.json`, but since it doesn't exist, it will:
1. Create a [new default config](https://github.com/NordSecurity/libtelio/blob/f5b419e90d02fba7711192b8debc5b52e95f8bdf/clis/nordvpnlite/src/config.rs#L162)
2. Since the default config does not contain a valid API token, the daemon will exit with error breaking the VPN connection.

As you mentioned `nordvpnlite start -c config.json --working-directory /home/user/` would solve this, provided the user gave the correct `--working-directory`, IMHO this is not a very smooth and polished user experience.

I'm just pointing this out, If you are already aware and this was a conscious design decision, please document the expected behavior in the read me.

clis/nordvpnlite/src/daemon.rs
tomasz-grz: I would personally prefer something like `RunningConfig { parsed, path, hash }`, but I'm not sure if it would make it more complicated with the borrow checker..

clis/nordvpnlite/src/main.rs
```diff
@@ -82,15 +82,19 @@ fn main() -> Result<(), NordVpnLiteError> {
...
+            logging_handle,
```
tomasz-grz: `logging_handle` will be dropped inside `daemon_event_loop` since it's moved here. Could this lead to unflushed logs during teardown?

clis/nordvpnlite/src/daemon.rs
```diff
@@ -348,15 +351,61 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
...
+    /// Carries the already parsed new config and the new config file hash
+    Reload(Box<NordVpnLiteConfig>, u64),
+}
+
+pub async fn daemon_event_loop(
+    mut config: NordVpnLiteConfig,
+    config_path: String,
+    mut logging_handle: logging::LoggingHandle,
+) -> Result<(), NordVpnLiteError> {
+    let mut config_file = ConfigFile::new(config_path);
```
tomasz-grz: Looking at this again, this is called after possible detaching of the daemon.

In case of a relative `config_path` like we already discussed, I suspect the ` std::fs::read(&path)` could already fail when calculating the hash, even before reload is called.

Do you think it would be worth to at least add a log to? 
```rust
        let hash = std::fs::read(&path)
            .map(|b| Self::hash_bytes(&b))
            .unwrap_or(0);
```

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: Will change it to PathBuf

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: On CLI I am getting
```
Error: InvalidResponse("EOF while parsing a value at line 1 column 0")
```

daemon is not restarted and continues with the old configuration

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: Added `RunningConfig`

clis/nordvpnlite/src/config.rs
tomasz-kumor: Changed

clis/nordvpnlite/src/main.rs
tomasz-kumor: I have canonicalized the config path

clis/nordvpnlite/src/main.rs
tomasz-kumor: Changed to keep the ownership of `logging_handle` in the main. Just in case.

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: This comment is no longer relevant

tomasz-kumor:
Added two new improvements:
- config file is read in fixed size chunks during hash calculation. No risk of loading a huge file into the memory.
- tracing subscriber is reload only if any of its parameters changed.

clis/nordvpnlite/src/config.rs
tomasz-kumor: We will address the potential packet leak during reload in the future. We are aware of this risk and this PR is just first step to address this issue.

stalowyjez (dismissed):
@tomasz-kumor lgtm, +0,5 so @tomasz-grz also can give you some points ;)

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: This is possibly because we are dropping the IPC connection before sending a `CommandResponse` back

clis/nordvpnlite/src/daemon.rs
```diff
@@ -396,15 +445,15 @@ pub async fn daemon_event_loop(mut config: NordVpnLiteConfig) -> Result<(), Nord
                 },
                 _ = signals.next() => {
                     warn!("Interrupted while obtaining service credentials - stopping");
-                    return Ok(());
+                    return Ok(LoopOutcome::Exit);
```
tomasz-grz, line 449: If I'm not mistaken this might quit the daemon if it receives the `SIGHUP` while it's fetching the credentials from API (with exp back-off on slow connection this could take up to 180s).

Should we maybe ignore `SIGHUP` signals here, like we would do if you send `nordvpnlite reload` during API fetching?

clis/nordvpnlite/src/config.rs
```diff
@@ -20,6 +20,47 @@ use telio::{
...
+    pub path: PathBuf,
+    pub hash: u64,
+}
+
+impl RunningConfig {
+    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, NordVpnLiteError> {
+        let path = path.as_ref().to_path_buf();
+        let parsed = NordVpnLiteConfig::from_file(&path)?;
+        let path = path.canonicalize()?;
+        let hash = Self::hash_file(&path).unwrap_or(0);
```
tomasz-grz, line 35: We are opening and reading the file twice in `NordVpnLiteConfig::from_file` and `RunningConfig::hash_file`, maybe `NordVpnLiteConfig::from_file` could return the hash?

clis/nordvpnlite/src/config.rs
```diff
@@ -20,6 +20,47 @@ use telio::{
...
+#[derive(Clone, Debug)]
+pub struct RunningConfig {
+    pub parsed: NordVpnLiteConfig,
+    pub path: PathBuf,
+    pub hash: u64,
+}
+
+impl RunningConfig {
+    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, NordVpnLiteError> {
+        let path = path.as_ref().to_path_buf();
```
tomasz-grz, line 32: Also should we `canonicalize` before reading the config?


```suggestion
        let path = path.as_ref().canonicalize()?;
```

clis/nordvpnlite/src/main.rs
```diff
@@ -133,6 +133,10 @@ async fn client_main(cmd: Cmd) -> Result<(), NordVpnLiteError> {
                         println!("Daemon is already stopped");
                         Ok(())
                     }
+                    ClientCmd::Reload => {
+                        println!("Daemon is not running, cannot reload");
+                        Err(NordVpnLiteError::DaemonIsNotRunning)
```
tomasz-grz: Do we need an explicit case here? Should be already caught by 
```rust
_ => Err(NordVpnLiteError::DaemonIsNotRunning),
```

tomasz-grz (commented):
Nice work, +1 besides some NITs

clis/nordvpnlite/src/config.rs
tomasz-kumor: Not necessarily but I will change it.  The path that we are storing for reload need to be canonicalized.

clis/nordvpnlite/src/config.rs
tomasz-kumor: I was thinking about it and my conclusion was that it is better to read the config twice than loading the whole content of a config file into a memory. We can't be sure how big will be the config file. What if the user try to load a huge file. This approach seams to be safer.

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: But what if the user gets bored while waiting for credentials and decides to change something in the configuration and reload it. I think we should let them to do so.

clis/nordvpnlite/src/main.rs
tomasz-kumor: Ok, I will change it as the default log "Error: DaemonIsNotRunning" is enough descriptive.

clis/nordvpnlite/src/config.rs
tomasz-kumor: Ok, I kept it as it is because of the failing `running_config_file_not_found_creates_default_and_returns_token_error test`. The `canonicalize()` method requires the file to exist to resolve the the path. Yours suggestion would require to change the behavior where the `NordVpnLiteConfig::from_file` creates the config file if not exists.

tomasz-grz (dismissed):

tomasz-kumor:
Rebased

tomasz-grz (approved):

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -126,16 +128,29 @@ pub struct CommandListener {
...
             socket,
             telio_task_tx,
+            config,
+            pending_config: None,
         }
     }
 
+    pub fn take_pending_config(&mut self) -> Option<RunningConfig> {
+        self.pending_config.take()
+    }
```
stalowyjez, line 152: Nit: what is the advantage of `self.take_pending_config()` over `self.pending_config.take()` ?

stalowyjez (approved):

PR #1846, opened 2026-06-26
LLT-7444: boot x86_64 OpenWRT nat-lab VM with QEMU -M microvm

gytsto:
## Problem
[LLT-7444](https://nordsec.atlassian.net/browse/LLT-7444): try QEMU's `microvm` machine type in nat-lab — boot OpenWRT with `-M microvm` for a leaner/faster VM, and answer whether it's possible for ARM.

## Solution
Convert the x86_64 OpenWRT gateway (`openwrt-gw-01`) to `-M microvm,pcie=on,rtc=on` with direct kernel boot. `microvm` has no firmware/BIOS, so the kernel is extracted from the combined image at build time and booted via `-kernel`; virtio-pci net/blk and the existing combined image (`root=/dev/vda2`) are kept, so OpenWRT's stock drivers work unchanged.

- ~2× faster boot locally (~5s vs ~10s to serial console), same `-m 256M`
- Config-serialization and healthcheck paths are untouched (boot still reaches the ttyS0 console banner that `send-config-to-vm.sh` waits for)
- Kernel extraction (`extract-kernel.sh`) is gated by the `OPENWRT_DIRECT_KERNEL` build arg, set only on the x86_64 GW

## Notes
- `microvm` is **x86_64-only** — there is no aarch64 `microvm`; ARM already uses the minimal `-M virt`. The aarch64/mipsel gateways are unchanged.
- Validated locally (boot + ext4 rootfs mount + console banner). nat-lab CI is the authoritative check.

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/openwrt/extract-kernel.sh
+++ b/nat-lab/bin/openwrt/extract-kernel.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+# Pull OpenWrt's bzImage out of the x86 ext4-combined image so QEMU -M microvm
+# can direct-boot it (microvm has no firmware/BIOS to run the image's GRUB).
+# Partition 1 is a small ext2 "kernel" partition holding /boot/vmlinuz; carve it
+# out with dd and read the kernel with debugfs (no privileged loop mount, so it
+# works inside an unprivileged docker build). LLT-7444.
+set -eu
+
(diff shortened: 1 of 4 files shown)


gytsto:
closing this due to limited usage of microvm and low performance boost

PR #1847, opened 2026-06-26
Fix primary interface selection on apple platforms

mathiaspeters:
### Problem
There are times when the primary interface list provided by the OS is empty, in such times we should treat every interface as potentially relevant.

### Solution
Treat all interfaces as primary if the OS doesn't provide any primary interfaces

- [x] README.md is updated


tomaszklak (commented):
+1

gytsto (approved):
+1

PR #1855, opened 2026-06-30
nat-lab: minor build, startup, VM and CLI improvements

gytsto:
### Problem

A handful of nat-lab rough edges slowed local dev and CI and risked incorrect orchestration:

- The base image was rebuilt with `--no-cache` on every `./natlab.py start`.
- Lightweight services had bare healthchecks, so `docker compose up --wait` sat on Docker's 30s default interval for services that are ready in well under a second.
- The x86 OpenWrt VM ran under TCG software emulation even though `/dev/kvm` was already mapped into the container.
- Health/restart logic matched containers by substring, so `cone-gw-01` also matched `fullcone-gw-01` (and `symmetric`/`internal-symmetric`): health could read the wrong container and restarts could kill the wrong sibling.
- Per-test tcpdump capture + downloads and redundant Docker-API calls ran even locally, where the captures are never collected.
- `natlab.py --help` (and no-args, invalid command, bad flag) required a running docker daemon and dumped the entire ~58-service list inline.

### Solution

Minor, self-contained improvements — one commit per concern:

- **Build / orchestration** — cache the base image build for fast local iteration (`--rebuild` / `NATLAB_BUILD_NO_CACHE` to force a clean rebuild). **CI keeps `--no-cache`**, so CI images stay fully reproducible and never reuse a persisted layer — BuildKit's cache is content-addressed anyway, so a changed `Dockerfile`/`bin/` script can't serve a stale image. Resolve services to exact container IDs via `docker compose ps -q` (fixes `cone-gw-01`⊂`fullcone-gw-01` substring collisions); `--no-deps` on the recovery recreate; remove dead version-compat code.
- **Startup** — fast healthcheck anchor (2s interval/start, 15 retries) for lightweight services (~30s off the parallel startup floor); drop an unused anchor. VM/openwrt healthchecks keep their own timings.
- **OpenWrt VM** — enable KVM for x86 (probe + `-cpu host`, TCG fallback for runners without nested virt); reorder `openwrt.Dockerfile` so editing a script no longer re-triggers the image decompression.
- **Test setup/teardown** — gate per-test tcpdump capture/download on `NATLAB_SAVE_LOGS`; snapshot running containers once instead of inspecting every container per tag; reset miniupnpd only before tests that use a UPNP gateway (with a conservative fallback); dedupe the per-node log fetch; tighten two timeouts.
- **CLI** — `--help`/errors no longer require docker and read cleanly; helpful `--services-to-start` validation; fixed a stale help reference and a `-o` typo.

- [x] README.md is updated

### DIFF
--- a/nat-lab/natlab.py
+++ b/nat-lab/natlab.py
@@ -310,9 +310,7 @@ def _report_container_failures(
     raise RuntimeError(f"Containers failed to start: {failed}; see docker logs above")
 
 
-def check_containers(
-    services_to_start, check_only=False
-) -> Tuple[List[str], List[str]]:
+def find_failing_containers(services_to_start) -> Tuple[List[str], List[str]]:
     missing_services: List[str] = []
     unhealthy_services: List[str] = []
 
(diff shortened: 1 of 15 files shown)


nat-lab/tests/utils/connection_util.py
```diff
@@ -451,28 +489,6 @@ async def is_running(tag: ConnectionTag) -> bool:
...
-                    stderr=asyncio.subprocess.DEVNULL,
-                ),
-                timeout=IS_VM_RUNNING_PING_TIMEOUT,
-            )
-            returncode = await asyncio.wait_for(proc.wait(), timeout=2.0)
-            if returncode == 0:
-                return True
-            else:
-                log.debug("%s haven't replied to ICMP request at %s", tag, primary_ip)
+        log.debug("%s haven't replied to ICMP request at %s", tag, primary_ip)
```
olekoliinyk: Shouldn't it be secondary_ip here?

nat-lab/natlab.py
```diff
@@ -281,60 +313,21 @@ def _report_container_failures(
...
-                    if not container_state_raw or container_state_raw.lower() == "null"
-                    else json.loads(container_state_raw)
-                )
-            except json.JSONDecodeError:
-                container_state_json = None
-
-            if isinstance(container_state_json, dict):
-                status = (container_state_json.get("Status") or "").lower()
-                if status == "unhealthy":
-                    logs = container_state_json.get("Log") or []
```
olekoliinyk, line 331: We wan't print the log anymore in case container is unhealthy?

nat-lab/tests/utils/tcpdump.py
```diff
@@ -394,7 +394,7 @@ async def make_tcpdump(
                 await _start_capture(exit_stack, conn, session)
             yield
     finally:
-        if download:
+        if download and os.environ.get("NATLAB_SAVE_LOGS") is not None:
```
olekoliinyk: Maybe just: 
`if download and os.environ.get("NATLAB_SAVE_LOGS"):`

nat-lab/tests/utils/diagnostics.py
```diff
@@ -13,7 +14,10 @@ async def connection_diagnostics(
 ) -> AsyncIterator[None]:
     await clear_core_dumps(connection)
     await clear_system_log(connection)
-    if run_tcpdump:
+    # Per-test packet captures are only useful when logs are collected (CI sets
+    # NATLAB_SAVE_LOGS). Skipping the per-connection tcpdump spawn locally removes
+    # a process + pcap download from every test's connection setup.
+    if run_tcpdump and os.environ.get("NATLAB_SAVE_LOGS") is not None:
```
olekoliinyk: Maybe just: 
`if run_tcpdump and os.environ.get("NATLAB_SAVE_LOGS"):`

asledzinskiy:
+1

nat-lab/tests/utils/connection_util.py
gytsto: Good catch — fixed to log `secondary_ip` in the secondary-ping branch

nat-lab/tests/utils/tcpdump.py
gytsto: updated

nat-lab/tests/utils/diagnostics.py
gytsto: updated

nat-lab/natlab.py
gytsto: Somehow missed that check flag had to be set for logs to be printed. Now logs will be printed at the end of start if any container is missing or unhealthy

olekoliinyk:
+1.0

olekoliinyk (dismissed):

olekoliinyk (approved):

PR #1858, opened 2026-06-30
LLT-7321: Make firewall reject packets with non-tunnel src ips

lcruz99:
### Problem
Connections opened before the VPN tunnel comes up (analytics, OS DNS, browser prefetch) are bound to the physical NIC's address. When those sockets later send through the now-active tunnel, the inner packet still carries that stale source IP. The exit node interprets this as a client-IP change and starts redirecting all subsequent return traffic to the wrong destination, tearing down active connections. This breaks NordWhisper and intermittently NordLynx shortly after startup.

  ### Solution
Once a VPN exit node is connected and the integrating app has supplied the tunnel interface's source IPs, the firewall rejects any outbound packet whose source IP is not one of those addresses. The check is opt-in and inert until the app provides IPs (passing an empty list disables it).

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_vpn.py
+++ b/nat-lab/tests/test_vpn.py
@@ -24,6 +24,7 @@ from tests.utils.netcat import NetCatClient
 from tests.utils.ping import ping
 from tests.utils.process import ProcessExecError
 from tests.utils.router import IPProto, IPStack
+from tests.utils.tcpdump import make_tcpdump
 
 
 class TestVpnConnection:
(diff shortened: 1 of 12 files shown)


nat-lab/tests/test_vpn.py
```diff
@@ -768,3 +769,95 @@ async def test_vpn_connection_private_key_change(
...
+                        ConnectionTag.DOCKER_CONE_CLIENT_1,
+                        vpn_1_limits=(1, 1),
+                    ),
+                    is_meshnet=False,
+                ),
+                id="neptun",
+            ),
+        ],
+    )
+    async def test_vpn_reject_wrong_tunnel_src_ip(
```
stalowyjez, line 794: Hmm, I would consider also an integration test for this one, as they are easier to run and debug

stalowyjez (commented):
Lgtm, +1

src/ffi.rs
```diff
@@ -680,6 +680,23 @@ impl Telio {
         })
     }
 
+    /// Set the source IP(s) currently assigned to the tunnel interface so the
+    /// firewall can reject outbound packets with a wrong source IP. Empty to
+    /// disable.
```
mathiaspeters: The doc comment in the udl file is better in my opinion, can we copy-paste that here?

lcruz99: sure, will do

gytsto (approved):
+1

mathiaspeters (approved):
+1

nat-lab/tests/test_vpn.py
lcruz99: added

PR #1863, opened 2026-07-02
LLT-6918: Remove async-std dependency

stalowyjez:
### Problem
It seems that we have async-std crate, which is unmaintanied, added as our telio-rely dependency and we don't use it at all

### Solution
Remove async-std dependency

- [x] README.md is updated


tomaszklak (approved):
+2

PR #1865, opened 2026-07-03
Reset handshake timers on network change

mathiaspeters:
### Problem
Android reported some nonet scenarios and from reading the logs two things could be found: wireguard session were expired for longer than they should after waking up from doze mode and after a network change

### Solution
Make the timers for handshakes suspend-aware and reset all handshake timers on `notify_network_change`

- [x] README.md is updated

### DIFF
--- a/crates/telio-pq/src/conn.rs
+++ b/crates/telio-pq/src/conn.rs
@@ -4,7 +4,7 @@ use tokio::task::JoinHandle;
 
 use telio_model::features::FeaturePostQuantumVPN;
 use telio_task::io::chan;
-use telio_utils::{reset_after, telio_log_debug, telio_log_warn};
+use telio_utils::{telio_log_debug, telio_log_warn, SuspendAwareTicker};
 
 use crate::proto;
 
@@ -23,7 +23,11 @@ impl ConnKeyRotation {
     ) -> Self {
         telio_log_debug!("Starting PQ task");
 
-        let rekey_interval = Duration::from_secs(features.rekey_interval_s as _);
+        let mut rekey_interval = features.rekey_interval_s;
+        if rekey_interval == 0 {
+            rekey_interval = 1;
+        }
+        let rekey_interval = Duration::from_secs(rekey_interval as _);
         let request_retry = Duration::from_secs(features.handshake_retry_interval_s as _);
         let pq_version = features.version;
 
@@ -68,11 +72,10 @@ impl ConnKeyRotation {
                 .await;
 
             telio_log_debug!("Rekey interval: {}s", rekey_interval.as_secs());
-            let mut interval = telio_utils::interval(rekey_interval);
+            let mut ticker = SuspendAwareTicker::new_after(rekey_interval, rekey_interval);
 
-            interval.tick().await; // This call returns immedietly
             loop {
-                interval.tick().await;
+                ticker.tick().await;
 
                 // Dylint is unhappy about the `rekey` future size
                 // and asks for using `Box::pin` to move it on the heap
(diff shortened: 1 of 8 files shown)


sfraczek (commented):
+1

crates/telio-utils/src/repeated_actions.rs
```diff
@@ -84,12 +85,23 @@ where
...
             },
         )
     }
 
+    /// Reset all action timers so they fire immediately on the next poll.
+    ///
+    /// This is useful after a network change: rather than waiting up to one full
+    /// period for the next scheduled keepalive, all peers receive a ping as soon
+    /// as `select_action` is polled again
+    pub fn reset_all_actions(&mut self) {
```
tomasz-grz, line 99: Running `cargo test --release -p telio-utils` locally on macOS produces the following failing tests for me,
but without `--release` flag they pass 🤔 

```sh
failures:

---- repeated_actions::tests::execute_actions stdout ----

thread 'repeated_actions::tests::execute_actions' panicked at crates/telio-utils/src/repeated_actions.rs:263:13:
actual = 1.666µs, expected = 100ms

---- repeated_actions::tests::update_action_interval stdout ----

thread 'repeated_actions::tests::update_action_interval' panicked at crates/telio-utils/src/repeated_actions.rs:348:21:
actual = 200ms, expected = 100ms

thread 'repeated_actions::tests::update_action_interval' panicked at crates/telio-utils/src/repeated_actions.rs:375:10:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(1), "actual = 200ms, expected = 100ms", ...)

---- repeated_actions::tests::execute_action_after_a_long_sleep_shouldnt_call_all_the_missed_actions stdout ----
hello at 8µs
hello at 16.083µs
hello at 20.791µs

thread 'repeated_actions::tests::execute_action_after_a_long_sleep_shouldnt_call_all_the_missed_actions' panicked at crates/telio-utils/src/repeated_actions.rs:232:9:
assertion failed: start.elapsed() >= Duration::from_secs(5)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    repeated_actions::tests::execute_action_after_a_long_sleep_shouldnt_call_all_the_missed_actions
    repeated_actions::tests::execute_actions
    repeated_actions::tests::update_action_interval
```

crates/telio-pq/src/conn.rs
```diff
@@ -68,11 +68,10 @@ impl ConnKeyRotation {
                 .await;
 
             telio_log_debug!("Rekey interval: {}s", rekey_interval.as_secs());
-            let mut interval = telio_utils::interval(rekey_interval);
+            let mut ticker = SuspendAwareTicker::new_after(rekey_interval, rekey_interval);
```
tomasz-grz, line 75: What happens if `rekey_interval` is 0? Is there any possibility that it could constantly create re-key attempts?

crates/telio-utils/src/repeated_actions.rs
mathiaspeters: That's sort of by design. `crate::Instant` which is used behind the scenes there uses `boot_time` in release mode and `tokio::time` in debug mode, and only `tokio::time` works well with the tests

crates/telio-pq/src/conn.rs
mathiaspeters: The old implementation would panic if it was set to 0. I changed it now to force it to be at least 1, but that also means we're not respecting the config

tomasz-grz (approved):
+1

PR #1866, opened 2026-07-03
LLT-6920: Remove explicit defaults from FeaturesDefaultsBuilder constructor

oleksii-khomiakovskyi:
### Problem
Default `Features` config explicitly redefined inside `FeaturesDefaultsBuilder` constructor.

### Solution
`FeaturesDefaultsBuilder` constructor reuses `Features::default()` as a base line, and redefines only needed fields.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -199,7 +199,7 @@ pub enum RttType {
 pub struct FeatureLana {
     /// Path of the file where events will be stored. If such file does not exist, it will be created, otherwise reused
     pub event_path: String,
-    /// Whether the events should be sent to produciton or not
+    /// Whether the events should be sent to production or not
     pub prod: bool,
 }
 
@@ -254,7 +254,7 @@ pub enum PathType {
     Direct,
 }
 
-/// Enable meshent direct connection
+/// Enable meshnet direct connection
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, SmartDefault)]
 #[serde(default)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
@@ -453,7 +453,7 @@ pub struct FeatureFirewall {
     /// Customizable private IP range to treat certain private IP ranges
     /// as public IPs for testing purposes.
     pub exclude_private_ip_range: Option<Ipv4Net>,
-    /// Blackist for outgoing connections
+    /// Blacklist for outgoing connections
     #[serde(default)]
     pub outgoing_blacklist: Vec<FirewallBlacklistTuple>,
 }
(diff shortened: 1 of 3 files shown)


src/ffi/defaults_builder.rs
```diff
@@ -172,3 +151,37 @@ impl Default for FeaturesDefaultsBuilder {
...
+        assert_eq!(features.flush_events_on_stop_timeout_seconds, None);
+        assert_eq!(features.link_detection, None);
+        assert_eq!(features.error_notification_service, None);
+
+        assert!(features.hide_user_data);
+        assert!(features.hide_thread_id);
+        assert!(!features.ipv6);
+        assert!(!features.nicknames);
+        assert!(!features.multicast);
+    }
```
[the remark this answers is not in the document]
oleksii-khomiakovskyi: Test makes sure there is no changes in value produced by the builder constructor using all the fields.

crates/telio-model/src/features.rs
```diff
@@ -515,7 +515,7 @@ pub struct FeatureDns {
     pub use_raw_forwarder: Option<bool>,
 }
 
-/// Newtype for TTL value to ensure that the default function returns the actual default value and not 0.
+/// New type for TTL value to ensure that the default function returns the actual default value and not 0.
```
mathiaspeters: I think this one should stay as `Newtype` as that's what the pattern is referred to in Rust

src/ffi/defaults_builder.rs
```diff
@@ -172,3 +151,37 @@ impl Default for FeaturesDefaultsBuilder {
 fn default<T: Default>() -> T {
     T::default()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new_builder_default_features() {
```
mathiaspeters, line 160: What do you think about instead of doing the asserts against some literal (`default()`, etc), you define a separate variable like `let expected_features = Features::default()` and compare against that? I think that would be a better test that `FeaturesDefaultBuilder` is matching `Features::default()`

crates/telio-model/src/features.rs
oleksii-khomiakovskyi: Oh, got it. Reverted.

src/ffi/defaults_builder.rs
oleksii-khomiakovskyi: Make sense, updated.

crates/telio-model/src/features.rs
```diff
@@ -254,7 +254,7 @@ pub enum PathType {
     Direct,
 }
 
-/// Enable meshent direct connection
+/// Enable meshnet direct connection
```
tomasz-grz, line 257: Reminds me of the time we were wondering why `mehsnet` tickets were not showing up on the Jira board 🤣

tomasz-grz (commented):
+1

mathiaspeters (approved):
+1

PR #1869, opened 2026-07-07
Add support for post-quantum in nordvpnlite

tomaszklak:
### Problem
Nordvpnlite can only create non post-quantum safe vpn tunnels.

### Solution
Add new config option (disabled by default) that will control if the vpn tunnel is post-quantum secure.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/daemon.rs
+++ b/clis/nordvpnlite/src/daemon.rs
@@ -275,10 +275,15 @@ impl TelioTaskCmd {
                             .unwrap_or(DEFAULT_WIREGUARD_PORT),
                     )),
                 };
-                match ctx.telio.connect_exit_node(&node) {
+                let (connect, kind): (fn(_, _) -> _, _) = if exit_node.post_quantum {
+                    (Device::connect_vpn_post_quantum, "post quantum ")
+                } else {
+                    (Device::connect_exit_node, "")
+                };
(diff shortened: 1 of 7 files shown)


clis/nordvpnlite/src/daemon.rs
```diff
@@ -275,10 +275,20 @@ impl TelioTaskCmd {
                             .unwrap_or(DEFAULT_WIREGUARD_PORT),
                     )),
                 };
-                match ctx.telio.connect_exit_node(&node) {
+                let connect = if exit_node.post_quantum {
```
sfraczek: Seems like we could avoid checking twice for post quantum:
```rust
let (connect, kind) = if exit_node.post_quantum {
    (Device::connect_vpn_post_quantum, "post quantum ")
} else {
    (Device::connect_exit_node, "")
};
```

sfraczek (commented):
Consider adding pq to readme.md. While at it, you could add enable_firewall as well.

sfraczek (dismissed):
+1

mathiaspeters (dismissed):
+1

clis/nordvpnlite/src/daemon.rs
tomaszklak: Done

sfraczek (approved):

PR #1874, opened 2026-07-08
Reduce the size of libtelio/nordvpnlite in release mode

tomaszklak:
### Problem
Size of the release artifacts matter in at least two cases. First, mobile app stores use the application size as one of the signals to determine the likelihood of suggesting any given app to the user. Second openwrt devices have limited storage.

### Solution
Three different optimizations are part of this PR:

- switch from opt-level "s" to "z"
- switch from full regex crate to a smaller regex-lite
- change the type of telio crate to only cdylib. When it was set to both cdylib and lib, rust compiler considered any pub symbol to be reachable and this reduced the set of symbols that compiler and linker considered dead code

build type | in MiB | in MiB | in MiB |  
-- | -- | -- | -- | --
  | main | after | delta | delta%
android aarch64 | 13.3 | 8.8 | 4.5 | 33.83%
android armv7 | 8.1 | 5.7 | 2.4 | 29.63%
android i686 | 14 | 9 | 5 | 35.71%
android x86_64 | 15.3 | 10.8 | 4.5 | 29.41%
ios | 12.2 | 7.4 | 4.8 | 39.34%
mac aarch64 | 12.3 | 7.5 | 4.8 | 39.02%
tvos | 12.3 | 7.3 | 5 | 40.65%
linux aarch64 | 13.8 | 9.2 | 4.6 | 33.33%
linux armv7 | 11.2 | 7.8 | 3.4 | 30.36%
linux armv5 | 11.5 | 8 | 3.5 | 30.43%
linux i686 | 14.7 | 9.5 | 5.2 | 35.37%
linux x86_64 | 15.8 | 11.3 | 4.5 | 28.48%
windows x86_64 | 14 | 10.7 | 3.3 | 23.57%
windows aarch64 | 13.3 | 10 | 3.3 | 24.81%
openwrt aarch64 | 8.3 | 6.7 | 1.6 | 19.28%
openwrt armv7 | 7.3 | 5.7 | 1.6 | 21.92%
openwrt mips | 9.8 | 8.2 | 1.6 | 16.33%
openwrt mipsel | 9.8 | 8.2 | 1.6 | 16.33%
openwrt x86_64 | 10.6 | 8.9 | 1.7 | 16.04%

- [x] README.md is updated


clis/tcli/Cargo.toml
```diff
@@ -27,15 +27,15 @@ tracing.workspace = true
...
+regex-lite.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
 thiserror.workspace = true
 time.workspace = true
 tokio = { workspace = true, features = ["full"] }
 
-telio = { path = "../.." }
+telio-core.workspace = true
```
gytsto: `ens` and `default_features` are now `false` on `tcli`, not sure if it was used on the tool itself

gytsto (commented):
other than one comment, looks good +1

clis/tcli/Cargo.toml
tomaszklak: Fixed

Cargo.toml
```diff
@@ -211,7 +211,7 @@ telio-pinger = { version = "0.1.0", path = "./crates/telio-pinger" }
 telio-core = { version = "0.1.0", path = "./crates/telio-core", default-features = false }
 
 [profile.release]
-opt-level = "s"
+opt-level = "z"
```
matislovas, line 214: I wonder what the impact on performance will be with `z` option 🤔

matislovas:
+1

matislovas (approved):

Cargo.toml
tomaszklak: I plan to take a look, after couple of days, at the speeds charts that we collect from nightly and see if there is any impact.

PR #1875, opened 2026-07-08
Fix firewall consolidation

jjanowsk:
Make consolidate_firewall preserve firewall state set by enable/set_tp_lite_* calls.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_tp_lite_stats.py
+++ b/nat-lab/tests/test_tp_lite_stats.py
@@ -271,6 +271,11 @@ class TestTpLiteStats:
             _tp_lite_config(force_plaintext_dns=True)
         )
 
+        # LLT-7452: This sleep is added as a regression test in order to validate that
+        # the configuration set by enable_stats_collection is not cleared by the
+        # firewall consilidation. Consolidation runs every 5 seconds.
+        await asyncio.sleep(10)
+
         for port in NON_PLAINTEXT_DNS_PORTS:
             with pytest.raises(ProcessExecError):
                 await query_dns_port(connection, port, ALLOWED_DOMAIN, TP_LITE_DNS_IP)
@@ -667,6 +672,11 @@ class TestDnsWhitelisting:
                 )
             ],
         )
+
+        # LLT-7452: This sleep is added as a regression test in order to validate that
+        # the configuration set by set_domain_whitelist is not cleared by the
+        # firewall consilidation. Consolidation runs every 5 seconds.
+        await asyncio.sleep(10)
 
         # blocked-malware.com is NXDOMAIN at the TP-Lite (blocking) server, but it
         # is now whitelisted, so the firewall DNATs the query to the standard DNS
--- a/src/device/wg_controller.rs
+++ b/src/device/wg_controller.rs
@@ -14,7 +14,7 @@ use tokio::sync::Mutex;
 use telio_crypto::PublicKey;
 use telio_dns::DnsResolver;
 #[cfg(feature = "enable_firewall")]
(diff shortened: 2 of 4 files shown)


src/device/wg_controller.rs
```diff
@@ -610,25 +610,26 @@ async fn consolidate_firewall<F: Firewall>(
...
 ) -> Result {
     let mut state = FirewallState::default();
-    let peers: Vec<_> = iter_peers(requested_state).collect();
+    let old_state = firewall.get_state();
 
-    state.whitelist.port_whitelist = peers
-        .iter()
+    state.force_plaintext_dns_for_servers = old_state.force_plaintext_dns_for_servers;
+    state.tp_lite_whitelisted_domains = old_state.tp_lite_whitelisted_domains;
+    state.tp_lite_dns_redirects = old_state.tp_lite_dns_redirects;
```
mathiaspeters: What do you think about doing `firewall.get_state().clone()` here instead so that we don't run into the same issue if we add more fields to the state but forget to update here? I think we can skip it for the backport, but for this one targeting `main` it could be good

jjanowsk: I was thinking about it but then it requires extra logic in the existing parts for cleaning fields that we no longer need. Right now we just don't set them in consolidate_firewall function, because it is always starting from empty default. So in the end I've decided to leave it like this.

The other option I also considered and spent most of the time contemplating it, was to simply add those values to the RequestedState and just copy it to the firewall in the conolidation method. Skip having set_tp_lite_* methods in firewall entirely. That would be more consistent with the rest but there were also problems with this solution, so in the end I ended up with this one, which to be fair, I'm still not a fan of, but I think is the cleanest of them all.

mathiaspeters: We could create something like a `clone_for_consolidate` function for the state, or anything else that lives next to the definition of the struct field, or just a comment on the state struct reminding you to update here if necessary, but I will leave it up to you

mathiaspeters:
+1

nat-lab/tests/test_tp_lite_stats.py
```diff
@@ -679,6 +689,7 @@ async def test_dns_whitelisting_redirects_blocked_domain(
             expected_output=[WHITELIST_RESOLVED_IP],
             options=["-type=a"],
         )
+        await asyncio.sleep(1)
```
stalowyjez: And why this sleep is added?

src/device/wg_controller.rs
```diff
@@ -610,25 +610,26 @@ async fn consolidate_firewall<F: Firewall>(
     dns_pubkey: Option<PublicKey>,
 ) -> Result {
     let mut state = FirewallState::default();
-    let peers: Vec<_> = iter_peers(requested_state).collect();
+    let old_state = firewall.get_state();
```
stalowyjez: Well, I probably should write that on the previous PRs, but firewall state is recreated from scratch here by design, so making a part of that persistent doesn't seem to be a very elegant solution to me - I'd consider either moving lists to requested state and converting them here or just creating another field in StatefulFirewall, as we did with interface IPs

stalowyjez: I.e. I see no reason for them to be a part of state which is by-design set only by this consolidation function, while they are by-design not touched by consolidation function, but set in completely different place

stalowyjez (dismissed):
The new version looks good to me, +1

mathiaspeters (dismissed):
Looks good 👍 
+1

stalowyjez (approved):

PR #1876, opened 2026-07-08
LLT-7470: Increase nordvpnlite client command timeout to 60 seconds

sfraczek:
## What

Bumps the nordvpnlite CLI client IPC response timeout `TIMEOUT_SEC` from 30s to 60s.

## Why

Natlab tests intermittently fail because the CLI client trips its 30s IPC response timeout while the daemon is slow to reply during startup / heavy load. This timeout has already been extended several times (1 → 2 → 10 → 20 → 30). Doubling the client-side safety net to 60s reduces flakiness while keeping a guard against a genuinely stuck daemon (removing it entirely would let the CLI hang forever on a deadlocked daemon).

Jira: LLT-7470

## Notes

A separate, self-contained follow-up hardens the IPC protocol itself (newline-framed responses so the client no longer depends on the socket closing) under LLT-7476 / branch `LLT-7476-ipc-response-framing`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-7470
+++ b/.unreleased/LLT-7470
@@ -1,0 +1,1 @@
+Increase nordvpnlite client command timeout to 60 seconds
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -12,7 +12,7 @@ use crate::{
     daemon::{NordVpnLiteError, TelioStatusReport},
 };
 
-pub(crate) const TIMEOUT_SEC: u64 = 30;
+pub(crate) const TIMEOUT_SEC: u64 = 60;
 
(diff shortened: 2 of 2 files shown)


gytsto (approved):
+2

tomasz-grz:
I would prefer if we investigate what is making the VM so slow, we can't bump the timeout indefinitely..
Otherwise +1 😅

gytsto:
> I would prefer if we investigate what is making the VM so slow, we can't bump the timeout indefinitely.. Otherwise +1 😅

It is just x86-64 -> ARM emulation. If we start failing at 60 seconds timeout, I'd suggest removing openwrt-ARM from testsuite until we have ARM ready natlab runners and environment in general

PR #1877, opened 2026-07-08
LLT-7476: Newline-terminate nordvpnlite IPC responses

sfraczek:
### What

Hardens the nordvpnlite IPC protocol so the CLI client no longer depends on the daemon closing the socket to detect the end of a reply.

- Daemon responses are newline-terminated via `writeln!` (symmetric with requests), and the client reads them with `read_line` instead of `read_to_string`.
- `read_command` guards the trailing-newline strip with `ends_with('\n')` instead of an unconditional `pop()`, matching the client side (a peer that omits the terminator no longer loses its last character).
- Both read paths treat a 0-byte read (peer closed without sending) as an explicit `UnexpectedEof` error instead of an empty string that fails to deserialize with an opaque message.
- `respond()` `debug_assert`s the payload contains no newline; a unit test proves `CommandResponse::serialize` never emits a raw newline (compact JSON escapes them), enforcing the single-line framing invariant.
- Tests exercise the real client framing path (`send_command`) and cover the daemon-closed-without-response case.

### Why

The old `read_to_string` only returns once the daemon closes the connection, coupling end-of-message detection to connection teardown - a documented deadlock risk if the daemon ever held the socket open. Framing the response makes the protocol robust regardless of connection lifecycle.

Jira: LLT-7476 (follow-up to LLT-7470).

- [x] README.md is updated


clis/nordvpnlite/src/comms.rs
```diff
@@ -164,8 +188,23 @@ impl DaemonConnection {
...
+        // `CommandResponse::serialize` uses compact JSON which escapes embedded newlines, so
+        // this only trips on a programming error (e.g. switching to pretty-printed JSON or
+        // sending a raw multi-line string). See `test_response_serialize_has_no_raw_newline`.
+        debug_assert!(
+            !response.contains('\n'),
+            "IPC response must not contain a newline; it would corrupt message framing"
+        );
+        let mut buffer = Vec::<u8>::new();
+        writeln!(buffer, "{response}")?;
+        self.stream.write_all(&buffer).await?;
```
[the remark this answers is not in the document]
sfraczek: Done - the response String is owned, so the newline is now appended in place (single allocation-free push + one write) instead of the intermediate buffer.

clis/nordvpnlite/src/comms.rs
```diff
@@ -152,19 +162,33 @@ impl DaemonConnection {
...
         let mut command_buffer = String::new();
         let mut reader = BufReader::new(&mut self.stream);
-        reader.read_line(&mut command_buffer).await?;
-        // Removing the trailing newline used as a message ending according to platform.
-        command_buffer.pop();
+        if reader.read_line(&mut command_buffer).await? == 0 {
+            return Err(Error::new(
+                ErrorKind::UnexpectedEof,
+                "Client closed the connection without sending a command",
+            ));
```
djkarwowski, line 169: Do we have a test covering the scenario when the client closes without sending a command?

djkarwowski (commented):
+1

clis/nordvpnlite/src/comms.rs
sfraczek: Done, added a test.

djkarwowski (dismissed):

clis/nordvpnlite/src/comms.rs
```diff
@@ -152,19 +162,33 @@ impl DaemonConnection {
...
     /// # Arguments
     ///
     /// * `response` - The response produced by handling a command.
-    pub async fn respond(&mut self, response: String) -> Result<()> {
+    pub async fn respond(&mut self, mut response: String) -> Result<()> {
+        debug_assert!(
+            !response.contains('\n'),
+            "IPC response must not contain a newline; it would corrupt message framing"
+        );
+        response.push('\n');
```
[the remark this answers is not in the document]
sfraczek: All `respond()` callers pass `CommandResponse::serialize()`, which is `serde_json::to_string` (compact JSON) - raw newlines are escaped to `\n`, so a newline can't reach this path. The `debug_assert` documents that invariant and catches it in tests if a future non-JSON caller is added; enforcing it in release would only trade a caught-in-tests bug for a runtime error on something that can't currently occur. Leaving as-is.

.unreleased/LLT-7476
```diff
@@ -0,0 +1 @@
+Newline-terminate nordvpnlite IPC responses so the client detects the end of a reply via read_line instead of relying on the daemon closing the connection
```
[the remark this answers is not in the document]
sfraczek: N/A

djkarwowski (dismissed):

.unreleased/LLT-7476
jjanowsk: Not sure that we want to have it in a changelog.

jjanowsk (dismissed):
+1

.unreleased/LLT-7476
sfraczek: removed

djkarwowski (approved):

PR #1885, opened 2026-07-09
LLT-6287: Reset WG-NT listen port on dynamic down

sfraczek:
### Problem

With dynamic WG-NT control enabled, bringing the adapter down after the last peer is removed leaves the previously bound listen port cached inside the wireguard-nt driver. On the next up transition the driver re-binds that exact port; if another process took it in the meantime, the adapter never comes up - the driver logs `Could not bind socket ... (0xc0000043)` and libtelio retries `Failed to set adapter state to Up (os error 32)` until it gives up, so reconnecting to VPN fails.

### Solution

After the dynamic down transition send a `listen_port=0` UAPI set to the driver, so the next up binds a fresh ephemeral port instead of the stale one.

Covered by a new nat-lab test which occupies the previous listen port on the Windows VM and asserts that a subsequent connect still brings the adapter up. Verified red/green on a local nat-lab: the test fails without the fix (`0xc0000043` in the client log) and passes with it. The branch also gates NLX dmesg collection on the `nlx` session mark, since local runs with `--skip-nlx` and `NATLAB_SAVE_LOGS` crashed the whole pytest session before any test ran.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_adapter.py
+++ b/nat-lab/tests/test_adapter.py
@@ -1,5 +1,6 @@
 import asyncio
 import pytest
+import re
 from enum import Enum
 from tests import config
 from tests.helpers import SetupParameters, Environment
@@ -17,6 +18,7 @@ from tests.utils.connection_util import (
     ConnectionManager,
 )
 from tests.utils.process import ProcessExecError
+from tests.utils.python import get_python_binary
 from typing import Awaitable, Callable, List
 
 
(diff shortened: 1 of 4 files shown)


nat-lab/tests/test_adapter.py
```diff
@@ -252,6 +254,82 @@ async def test_adapter_state_for_vpn_and_dns(
...
+        client_alpha, *_ = env.clients
+
+        server_ip = config.WG_SERVER["ipv4"]
+        server_port = config.WG_SERVER["port"]
+        server_public_key = config.WG_SERVER["public_key"]
+        assert (
+            isinstance(server_ip, str)
+            and isinstance(server_port, int)
+            and isinstance(server_public_key, str)
+        )
```
mathiaspeters: I think we can skip these assertions

mathiaspeters (commented):
+1

nat-lab/tests/test_adapter.py
sfraczek: Done

Paintree (approved):
+1

mathiaspeters (approved):

PR #1886, opened 2026-07-10
LLT-6756: Store authentication token separately from config

oleksii-khomiakovskyi:
### Problem
The authentication token is stored as part of the `nordvpnlite` configuration.
The user must know where this file is located and edit it before launching the daemon.
Daemon can't start with the default config until it will be manually updated.

### Solution
Authentication token removed from the general config and stored in a separate file.
Added `login` and `logout` commands to store credentials similar to NordVPN Linux application. 
Keep support for passing the token via environment variables.
Produced default config file require no manual updates.
Previous config supported and will be automatically migrated during at first daemon `start`.

This solution provides the groundwork for the following improvements to store secrets in a secure location - SR-WR2-104.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/auth.rs
+++ b/clis/nordvpnlite/src/auth.rs
@@ -1,0 +1,347 @@
+use std::{
+    fs::{self, OpenOptions},
+    io::{BufReader, Write},
+    os::unix::fs::{OpenOptionsExt, PermissionsExt},
+    path::Path,
+};
+
+use serde::{Deserialize, Serialize};
+
+use telio_core::{crypto::SecretKey, telio_utils::Hidden};
+
+use crate::{NordVpnLiteError, DEFAULT_FILE_PERMISSIONS};
+
+#[derive(PartialEq, Eq, Debug, Clone, Serialize, Deserialize)]
+#[serde(try_from = "Hidden<String>")]
+pub struct NordToken(Hidden<String>);
+
+impl NordToken {
+    pub fn new(token: &str) -> Result<Self, NordVpnLiteError> {
+        if Self::validate(token) {
+            Ok(NordToken(token.to_owned().into()))
+        } else {
+            Err(NordVpnLiteError::InvalidConfigToken {
+                msg: "Invalid authentication token format".to_owned(),
+            })
+        }
+    }
+
+    fn validate(token: &str) -> bool {
+        token.len() == 64 && token.chars().all(|c| c.is_ascii_hexdigit())
+    }
+}
+
+impl TryFrom<Hidden<String>> for NordToken {
+    type Error = NordVpnLiteError;
+
+    fn try_from(token: Hidden<String>) -> Result<Self, Self::Error> {
+        NordToken::new(&token.0)
+    }
+}
+
+impl std::fmt::Display for NordToken {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl AsRef<str> for NordToken {
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl std::ops::Deref for NordToken {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[derive(Deserialize)]
+pub struct NordlynxKeyResponse {
+    pub nordlynx_private_key: String,
+}
+
+impl NordlynxKeyResponse {
+    pub fn into_secret_key(self) -> Result<SecretKey, NordVpnLiteError> {
+        self.nordlynx_private_key.parse::<SecretKey>().map_err(|_| {
+            NordVpnLiteError::InvalidResponse("Failed to parse nordlynx private key".to_owned())
+        })
+    }
+}
+
+#[derive(PartialEq, Serialize, Deserialize, Debug, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct NordVpnLiteAuth {
+    authentication_token: NordToken,
+}
+
+impl NordVpnLiteAuth {
+    /// Create a new NordVpnLiteAuth with the given token
+    pub fn new(token: NordToken) -> Self {
+        NordVpnLiteAuth {
+            authentication_token: token,
+        }
+    }
+
+    /// Get the authentication token
+    pub fn get_token(&self) -> NordToken {
+        self.authentication_token.clone()
+    }
+
+    /// Deserialize the NordVpnLiteAuth from an existing file at the given path
+    pub fn from_existing_file<P: AsRef<Path>>(path: P) -> Result<Self, NordVpnLiteError> {
+        let path = path.as_ref();
+        println!("Reading auth from: {}", path.display());
+
+        match fs::File::open(path) {
+            Ok(file) => {
+                let auth: NordVpnLiteAuth = serde_json::from_reader(BufReader::new(file))?;
+                Ok(auth)
+            }
+            Err(err) => Err(err.into()),
+        }
+    }
+
+    /// Serialize the NordVpnLiteAuth and write it to a file at the given path
+    pub fn to_file<P: AsRef<Path>>(&self, path: P) -> Result<(), NordVpnLiteError> {
+        let path = path.as_ref();
+
+        // Create the file with restrictive permissions set at creation time.
+        match OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .mode(DEFAULT_FILE_PERMISSIONS)
+            .open(path)
+        {
+            Ok(mut file) => {
+                // Ensure permissions are correct even if the file already existed
+                // (in which case `.mode()` on create is ignored by the OS).
+                let mut permissions = file.metadata()?.permissions();
+                permissions.set_mode(DEFAULT_FILE_PERMISSIONS);
+                fs::set_permissions(path, permissions)?;
+
+                let auth_json: Hidden<String> = serde_json::to_string_pretty(&self)?.into();
+                file.write_all(auth_json.as_bytes())?;
+                Ok(())
+            }
+            Err(err) => Err(err.into()),
+        }
+    }
+
+    /// Checks if NORD_TOKEN env var is set and contains a valid token
+    /// If so, returns a NordVpnLiteAuth, otherwise None
+    pub fn resolve_env_token() -> Option<Self> {
+        if let Ok(token) = std::env::var("NORD_TOKEN") {
+            println!("Overriding token from env");
+            match NordToken::new(&token) {
+                Ok(nord_token) => {
+                    return Some(NordVpnLiteAuth::new(nord_token));
+                }
+                Err(e) => {
+                    eprintln!("Token from env not valid: {}", e);
+                }
+            };
+        }
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{EnvVarHelper, NORD_TOKEN_ENV, VALID_TOKEN};
+    use serial_test::serial;
+    use temp_file::TempFile;
+
+    #[test]
+    fn test_new_rejects_invalid_token() {
+        let too_short = "abcd1234";
+        assert_matches::assert_matches!(
+            NordToken::new(too_short),
+            Err(NordVpnLiteError::InvalidConfigToken { .. })
+        );
+
+        let non_hex = "g".repeat(64);
+        assert_matches::assert_matches!(
+            NordToken::new(&non_hex),
+            Err(NordVpnLiteError::InvalidConfigToken { .. })
+        );
+
+        let one_bad = "a".repeat(63) + "w";
+        assert_matches::assert_matches!(
+            NordToken::new(&one_bad),
+            Err(NordVpnLiteError::InvalidConfigToken { .. })
+        );
+
+        let too_short_63 = "a".repeat(63);
+        assert_matches::assert_matches!(
+            NordToken::new(&too_short_63),
+            Err(NordVpnLiteError::InvalidConfigToken { .. })
+        );
+
+        let too_long_65 = "a".repeat(65);
+        assert_matches::assert_matches!(
+            NordToken::new(&too_long_65),
+            Err(NordVpnLiteError::InvalidConfigToken { .. })
+        );
+    }
+
+    #[test]
+    fn test_new_accepts_valid_token() {
+        assert!(NordToken::new(VALID_TOKEN).is_ok());
+    }
+
+    #[test]
+    fn test_nord_token_display() {
+        let token = NordToken::new(VALID_TOKEN).unwrap();
+        assert_eq!(format!("{}", token), VALID_TOKEN);
+    }
+
+    #[test]
+    fn test_nord_token_deref() {
+        let token = NordToken::new(VALID_TOKEN).unwrap();
+        let as_str: &str = &token;
+        assert_eq!(as_str, VALID_TOKEN);
+    }
+
+    #[test]
+    fn test_nordlynx_into_secret_key_valid() {
+        let hex_key = VALID_TOKEN;
+        let expected: SecretKey = hex_key.parse().expect("hex key should parse");
+        let response = NordlynxKeyResponse {
+            nordlynx_private_key: hex_key.to_owned(),
+        };
+        let parsed = response
+            .into_secret_key()
+            .expect("valid nordlynx key should parse");
+        assert_eq!(parsed, expected);
+    }
+
+    #[test]
+    fn test_nordlynx_into_secret_key_invalid() {
+        let response = NordlynxKeyResponse {
+            nordlynx_private_key: "not-a-valid-key".to_owned(),
+        };
+        assert_matches::assert_matches!(
+            response.into_secret_key(),
+            Err(NordVpnLiteError::InvalidResponse(_))
+        );
+    }
+
+    #[test]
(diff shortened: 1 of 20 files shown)


clis/nordvpnlite/src/auth.rs
```diff
@@ -0,0 +1,358 @@
...
+
+impl<'de> Deserialize<'de> for NordToken {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let raw_string: Hidden<String> = de::Deserialize::deserialize(deserializer)?;
+        NordToken::new(&raw_string).map_err(de::Error::custom)
+    }
+}
```
tomaszklak: Can't we use `#derive` to implement both traits? Maybe with `#[serde(transparent)]`.

clis/nordvpnlite/src/config.rs
```diff
@@ -153,53 +71,82 @@ impl NordVpnLiteConfig {
...
-                        .to_owned(),
-                })
+    /// Construct a NordVpnLiteConfig by deserializing an existing file at the given path
+    pub fn from_existing_file<P: AsRef<Path>>(path: P) -> Result<Self, NordVpnLiteError> {
+        let path = path.as_ref();
+        nordvpnlite_logstd_info!("Reading config from: {}", path.display());
+
+        match fs::File::open(path) {
+            Ok(file) => {
+                let mut config: NordVpnLiteConfig = serde_json::from_reader(file)?;
```
tomaszklak: This will read the `file` using one `read` syscall per byte. Please wrap it in `BufReader`.

clis/nordvpnlite/src/main.rs
```diff
@@ -144,8 +150,386 @@ async fn client_main(cmd: Cmd) -> Result<(), NordVpnLiteError> {
...
         _ => Err(NordVpnLiteError::InvalidCommand(format!("{cmd:?}"))),
     }
 }
+
+/// Store NordVPN authentication credentials in the auth file.
+///
+/// The token is validated before the config is read, then written to the
+/// auth file, overwriting any previously stored credentials.
+fn handle_login(opts: LoginOpts) -> Result<(), NordVpnLiteError> {
+    println!("Storing NordVPN authentication credentials");
```
tomaszklak, line 174: Why not use one of the nordvpnlite_logstd here and in couple of other places below that call `println` directly?

clis/nordvpnlite/src/config.rs
```diff
@@ -28,91 +26,11 @@ fn dns_default() -> Vec<IpAddr> {
...
 #[derive(PartialEq, Serialize, Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct NordVpnLiteConfig {
-    #[serde(
-        deserialize_with = "deserialize_nord_token",
-        serialize_with = "serialize_nord_token"
-    )]
-    pub authentication_token: NordToken,
+    /// Authentication credentials file path
+    pub auth_file_path: String,
```
tomaszklak, line 89: This is not backwards compatible change, right? If so, I think we should have a migrate step that will not force our users to re-authenticate.

clis/nordvpnlite/src/logging.rs
```diff
@@ -51,18 +51,26 @@ pub fn setup_logging<P: AsRef<Path>>(
     Ok(tracing_worker_guard)
 }
 
+#[macro_export]
+macro_rules! nordvpnlite_logstd_info {
+    ($($arg:tt)*) => {{
+        ::tracing::info!($($arg)*);
+        ::std::println!("{}", ::std::format_args!($($arg)*));
+    }};
+}
```
tomasz-grz, line 110: What is the motivation for implementing this macro?

`stdout` is already redirected to a log file so we will end up with duplicate messages. We need to be mindful with the amount of logs we write, since some devices have only a couple hundred MB flash.

Moreover and this macro seems to be only used before `tracing` is initialized, `::tracing::info!` will not be caught by the `tracing_subscriber` in those cases.
We should probably move `logging::setup_logging` before `NordVpnLiteConfig::from_file`, if we want to keep this.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -402,8 +406,6 @@ pub async fn daemon_event_loop(mut config: NordVpnLiteConfig) -> Result<(), Nord
         }
     };
 
-    config.authentication_token.zeroize();
```
tomasz-grz: Why was this removed?
Is the new `NordToken` implementation properly zeroized from memory on drop?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -70,6 +70,32 @@ pub(crate) struct DaemonOpts {
     pub stdout_path: String,
 }
 
+#[derive(Parser, Debug)]
+pub(crate) struct LoginOpts {
+    /// Configuration file to read authentication credentials path
+    #[clap(
+        long = "config-file",
+        short = 'c',
+        default_value = "/etc/nordvpnlite/config.json"
```
tomasz-grz: I would personally prefer if the operations are performed only on the currently loaded config by the daemon.
Otherwise this creates some ambiguity...

If daemon is started with `nordvpnlite start --config-file a.json` and the user does `nordvpnlite login --config-file b.json`, the running daemon itself is still not authenticated and would need to be restarted with a different config path before VPN connection can be established.

Or is this solving some specific requirement?

clis/nordvpnlite/src/config.rs
```diff
@@ -153,53 +71,82 @@ impl NordVpnLiteConfig {
...
             Ok(file) => {
                 let mut config: NordVpnLiteConfig = serde_json::from_reader(file)?;
                 config.log_file_path = Self::resolve_log_path(&config.log_file_path)?;
+                config.auth_file_path = Self::resolve_auth_path(&config.auth_file_path)?;
                 Ok(config)
             }
             // Create a default config if it does not exist
             Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
-                eprintln!("File not found, creating default config");
+                nordvpnlite_logstd_info!(
```
tomasz-grz: I would personally prefer to have errors printed to `stderr` instead of `stdout`.

Also same note regarding not initialized `tracing` at this stage.

clis/nordvpnlite/src/config.rs
```diff
@@ -153,53 +71,82 @@ impl NordVpnLiteConfig {
...
-                    ..Default::default()
-                };
+                let default_config = NordVpnLiteConfig::default();
+
+                // Create the file with restrictive permissions set at creation time.
+                let mut file = fs::OpenOptions::new()
+                    .write(true)
+                    .create(true)
+                    .truncate(true)
+                    .mode(0o640)
```
tomasz-grz: Is this redundant with the `set_mode()` below?

clis/nordvpnlite/src/config.rs
tomasz-grz: We could provide `auth_file_path` as an optional command line argument to the daemon instead of adding it to the config, and keeping `authentication_token: Optional<NordToken>`, would that would be a more backwards compatible change?

clis/nordvpnlite/src/auth.rs
oleksii-khomiakovskyi: yes, it is possible. Just `#[serde(transparent)]` not executing validation from the `NordToken::new`. But using `#[serde(try_from = "...")]` seems to be generating similar code.

clis/nordvpnlite/src/config.rs
oleksii-khomiakovskyi: Good point. Updated.

clis/nordvpnlite/src/main.rs
tomasz-grz: The daemon is running mostly in detached mode, and all `stdout` and `stderr` is redirected to a log file, in addition to the `tracing` logs.

There's little benefit to duplicate the logs unless it's an absolutely critical error message we want to make sure the user does not miss.

And we don't want to DOS the device with too many logs since most of them have limited flash memory.

oleksii-khomiakovskyi: Sure, updated with `nordvpnlite_logstd_info` macro.

clis/nordvpnlite/src/config.rs
tomasz-grz: Also should we create a `const` if the file permission is used more than once?

clis/nordvpnlite/src/command_listener.rs
tomasz-grz, line 30: Shouldn't we also add `Login` and `Logout` here?

clis/nordvpnlite/src/daemon.rs
oleksii-khomiakovskyi: Yes. I removed `Arc` around `Hidden<String>` because we don't need to share it. And `auth_token` was moved into `SecretKey` receiving scope. Right after it will be zeroized by Drop trait implemented in `Hidden`.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Ah I understand now, `nordvpnlite login` and `nordvpnlite logout` are actually daemon commands, and not client commands.

How does it work with an actively running daemon, and the `nordvpnlite reload` command?
Since it is launched automatically by initd as a service on systems like OpenWRT for example.

What is the exact user flow from installing nordvpnlite to login?

clis/nordvpnlite/src/config.rs
oleksii-khomiakovskyi: If I understand it correctly `.mode()` will be applied in case of new file creation. But in case file already exists this mode will not be used. That is why `set_mode` also important to make sure permissions will be updated for existing file also. Previously it was `create` + `set_mode`. But in this case there is a small window between creation file with default permission and setting required permissions.

clis/nordvpnlite/src/config.rs
oleksii-khomiakovskyi: yes, you are right. Was thing about this, but was not sure if we need something for this. I'll try implement old config migration functionality.

clis/nordvpnlite/src/command_listener.rs
oleksii-khomiakovskyi: I was trying to keep it simple like it was before. So, the prerequisite for successful execution `start` command is creating a valid auth file or env variable. Otherwise user will receive a hint like
```
Error: InvalidConfigToken { msg: "Please retrieve your authentication token from https://my.nordaccount.com and use the following command to set it: `nordvpnlite login --token <your_token>`." }
```

I was thinking about use cases like this: user adds his token first and then runs daemon.
For default paths:
```
sudo ./nordvpnlite login --token=<TOKEN>
sudo ./nordvpnlite start
```
If user needs a custom config file he is also able to do that just adding same `--config-file file`
```
sudo ./nordvpnlite login --token=<TOKEN> --config-file config.json
sudo ./nordvpnlite start --config-file config.json
```

Passing confg file A to `login` and config B to `start` command is possible, but most likely will be similar to the case without credential, so start will fail.

clis/nordvpnlite/src/config.rs
oleksii-khomiakovskyi: Initially this branch was leading to creating incorrect config file that was missing auth information. Now it produces fully valid default config.
@tomasz-grz do you think this still should be an error ?

clis/nordvpnlite/src/command_listener.rs
oleksii-khomiakovskyi: `Login` / `Logout` commands could be executed any time independent from the running daemon.
Those commands become an alternative to manual config update.
Expected that `login` will be used before the daemon start.
During daemon work you are able execute `login` / `logout` to modify auth credential any time, but to apply changes demon needs to be restarted.

1) installing nordvpnlite
2) `nordvpnlite login --token=<TOKEN>` or setting `export NORD_TOKEN=<TOKEN>` - creating auth credential 
3) `nordvpnlite start` - starting daemon
4) `nordvpnlite reload` - expect this is the same as (stop + start), should work if auth file present or ENV still configured  
6) `nordvpnlite logout` - cleanup credentials
7) `nordvpnlite stop` - stop daemon

clis/nordvpnlite/src/config.rs
oleksii-khomiakovskyi: I added config migration as part of the `start` command. If user has an existing config with configured valid `authentication_token` field just `nordvpnlite start` should be enough to start daemon without any other steps.

`Scenario 1` -  old valid config, fresh binary. 
```
nordvpnlite start
```
Token moved to a dedicated file, that will be created by default path `/etc/nordvpnlite/auth.json`.
Auth info removed from the config.

`Scenario 2` -  no config, fresh binary
```
nordvpnlite login --token=<TOKEN>
nordvpnlite start
```
Login store auth token into `/etc/nordvpnlite/auth.json`.
Default config does not contain auth info. 

`Scenario 3` -  old valid config TOKEN-1, fresh binary, login TOKEN-2 used
```
nordvpnlite login --token=<TOKEN-2>
nordvpnlite start
```
Login store TOKEN-2 into auth file.
Auth TOKEN-1 info removed from the config, but not overwrites existing TOKEN-2.
Daemon start will use credential stored by the `login` command.

clis/nordvpnlite/src/command_listener.rs
oleksii-khomiakovskyi: I did PR rebase, so `reload` command already available. 
Since auth info was moved to a dedicated file `reload` doesn't account any auth changes any more. User could update the auth credentials any time, but to apply it `stop`/`start` sequence needed. 
I think this behavior more consistent across 2 ways of passing credentials. Currently auth token change in config file is accounted by the `reload`, but env changes does not.

clis/nordvpnlite/src/config.rs
oleksii-khomiakovskyi: Added `DEFAULT_FILE_PERMISSIONS` constant.

clis/nordvpnlite/src/logging.rs
oleksii-khomiakovskyi: Yep, it was causing some logs duplication. Removed that extra macro, and updated code to show message once into `stdout` / `stderr`.

tomaszklak (commented):
+1

djkarwowski (commented):
+1

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -72,6 +72,32 @@ pub(crate) struct DaemonOpts {
...
+    pub token: String,
+}
+
+#[derive(Parser, Debug)]
+pub(crate) struct LogoutOpts {
+    /// Configuration file to read authentication credentials path
+    #[clap(
+        long = "config-file",
+        short = 'c',
+        default_value = "/etc/nordvpnlite/config.json"
```
tomasz-grz: Would be nice to have default values for `/etc/nordvpnlite/config.json` as a `const` since it's used in 3 places at least.

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Do we need the `--token` flag? Is it possible to pass it directly as an argument `nordvpnlite login <TOKEN>`?

clis/nordvpnlite/src/command_listener.rs
tomasz-grz: Since the daemon is launched automatically by service manager, wouldn't it be something more like:
1. install nordvpnlite
2. `nordvpnlite stop`
3. `nordvpnlite login --token=<TOKEN>`
4. `nordvpnlite start`

Which in my opinion not super smooth user flow.

But yeah, with the new `reload` command this is improving significantly:
1. install nordvpnlite
2. `nordvpnlite login --token=<TOKEN>`
3. `nordvpnlite reload`

I would personally try to strive for as less steps as possible:
1. install nordvpnlite
2. `nordvpnlite login --token=<TOKEN>`

But this would require quite a big refactor of the daemon event loop, and introducing a proper state machine (though this refactor can be worth if we want to have `connect` and `disconnect` commands later as well).
Since this is already a big improvement over what we currently have, consider it as a NIT 😅

tomasz-grz: Now that I think of it, probably the `nordvpnlite stop` isn't necessary since the daemon would fail to launch initially right after the install (because of the missing login file / token in config) 🤔 

In that case `nordvpnlite reload` wouldn't work either, would it?

clis/nordvpnlite/src/command_listener.rs
oleksii-khomiakovskyi: Not really, but I'd like to keep syntax with `--token` to follow syntax like in the `nordvpn` linux app. 
Made both variants possible.

clis/nordvpnlite/src/command_listener.rs
oleksii-khomiakovskyi: updated

clis/nordvpnlite/src/command_listener.rs
oleksii-khomiakovskyi: There is no big difference to previous version related to the daemon start.
Daemon will not be able to start if auth file or env is not configured beforehand.
So, likely sequence will be:

1. nordvpnlite daemon failed to start
2. `nordvpnlite login <TOKEN>` or setting `export NORD_TOKEN=<TOKEN>`
3. `nordvpnlite start`

The `reload` seems to be working only if demon already running.

Got your idea about having `connect`/`disconnect` and updating daemon event loop.
I believe there is a separate task LLT-6855 to add those commands, so we are moving in this direction.

tomasz-grz (approved):
+1

PR #1888, opened 2026-07-10
Start provider paused when battery opt is on

stalowyjez:
### Problem
`test_upnp_port_lease_duration` is flaky because UPnP provider is sometimes run in the beginning when there are no peers and should be paused

### Solution
Start UPnP endpoint provider paused, so it won't produce endpoints when there are no peers and will be unlocked by telio when peers are added in wg consolidate

- [x] README.md is updated


Paintree (approved):
+1

sfraczek (approved):
+1

PR #1891, opened 2026-07-13
LLT-6304 Check meshnet status, before doing dynamic wg nt control

matislovas:
### Problem
When `enable_dynamic_wg_nt_control=true` then `WireguardNT` adapter is brought up or down by `libtelio` depending on the peer count. As a result services like `NordDrop` may fail to bind to meshnet's IP address if meshnet is enabled, but there are no peers since the adapter would not be up.

### Solution
Provide a callback interface to adapter, to check and take into account meshnet status, while deciding whether to enable or disable the adapter. Meshnet's status is considered `enabled`, when mesh config is set some value (even without any peers in it).

```
|Mesh status | Peer count > 0    | Adapter status
|    1       |       1           |      Up
|    0       |       1           |      Up
|    1       |       0           |      Up
|    0       |       0           |      Down
```

- [x] README.md is updated

### DIFF
--- a/crates/telio-core/src/device.rs
+++ b/crates/telio-core/src/device.rs
@@ -340,6 +340,9 @@ pub struct Entities {
     // Wireguard interface
     wireguard_interface: Arc<DynamicWg>,
 
+    // Shared meshnet status for WireGuard callbacks
+    meshnet_on: Arc<Mutex<bool>>,
+
     // Internal DNS server
     dns: Arc<Mutex<DNS>>,
 
@@ -1229,7 +1232,14 @@ impl Runtime {
         #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
         set_tunnel_interface(&socket_pool, config.tun.as_ref(), config.name.as_ref());
 
+        let meshnet_on = Arc::new(Mutex::new(false));
+
         let requested_device_config = RequestedDeviceConfig::from(&config);
+
+        let enable_dynamic_wg_nt_control = features
+            .wireguard
+            .enable_dynamic_wg_nt_control
+            .then(|| meshnet_on_checker(meshnet_on.clone()));
 
         // tests runtime use wg::MockedAdapter
         cfg_if! {
@@ -1253,7 +1263,7 @@ impl Runtime {
                         firewall_process_inbound_callback,
                         firewall_process_outbound_callback,
                         firewall_reset_connections,
-                        enable_dynamic_wg_nt_control: features.wireguard.enable_dynamic_wg_nt_control,
+                        enable_dynamic_wg_nt_control,
                         skt_buffer_size : Runtime::sanitize_neptun_config(features.wireguard.skt_buffer_size, config.adapter.clone()),
                         inter_thread_channel_size : Runtime::sanitize_neptun_config(features.wireguard.inter_thread_channel_size, config.adapter.clone()),
                         max_inter_thread_batched_pkts : Runtime::sanitize_neptun_config(features.wireguard.max_inter_thread_batched_pkts, config.adapter.clone()),
@@ -1278,7 +1288,7 @@ impl Runtime {
                             firewall_process_inbound_callback,
                             firewall_process_outbound_callback,
                             firewall_reset_connections,
-                            enable_dynamic_wg_nt_control: features.wireguard.enable_dynamic_wg_nt_control,
+                            enable_dynamic_wg_nt_control,
                             skt_buffer_size: features.wireguard.skt_buffer_size,
                             inter_thread_channel_size: features.wireguard.inter_thread_channel_size,
                             max_inter_thread_batched_pkts: features.wireguard.max_inter_thread_batched_pkts,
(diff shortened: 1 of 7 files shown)


nat-lab/tests/test_adapter.py
```diff
@@ -359,7 +359,7 @@ async def test_adapter_state_for_meshnet(
     alpha_setup_params: SetupParameters,
     env: Environment,
 ) -> None:
-    # Creates and enables meshnet without any nodes
+    # 1. Creates and enables meshnet without any nodes
```
djkarwowski: Tiny nit: It looks like there was a plan to add numbers to the comments describing the steps, but only the first one was actually numbered.

crates/telio-wg/src/adapter.rs
```diff
@@ -93,6 +96,15 @@ pub trait Adapter: Send + Sync {
     ///
     /// Only the custom adapters can be cloned this way.
     fn clone_box(&self) -> Option<Box<dyn Adapter>>;
+
+    /// Ensure that adapter is UP or DOWN
+    async fn check_adapter_state(
```
djkarwowski: From the `check_` prefix in the method name I'd expect it to return the current adapter state without actually influencing it, which is rather contradictory to its real purpose (which both the return type and the comment reveals accurately).

If I understand it correctly, it is because `ensure_adapter_state` would cause a name conflict on the implementation side. Maybe something like `ensure_expected_adapter_state` would better reflect the methods purpose and its expected idempotent behavior (following the convention of `ensure_adapter_state` that the actual implementation wraps)? Or to keep public API surface more terse, maybe using `ensure_adapter_state` for the trait method and renaming the private one to a more verbose variant would be an option?

crates/telio-core/src/device.rs
```diff
@@ -2140,6 +2157,26 @@ impl Runtime {
...
+            .wireguard_interface
+            .check_adapter_state(
+                peers_cnt,
+                self.features
+                    .wireguard
+                    .enable_dynamic_wg_nt_control
+                    .then(|| {
+                        let meshnet_on = self.entities.meshnet_on.clone();
+                        Arc::new(move || meshnet_on.try_lock().map(|value| *value).unwrap_or(false))
+                            as Arc<dyn Fn() -> bool + Send + Sync>
```
djkarwowski: This closure seems to be mostly repeating what's on lines 1241-1243. It's a small one, but still maybe it would be worth to extract it to a little function. What do you think?

crates/telio-core/src/device.rs
```diff
@@ -2140,6 +2157,26 @@ impl Runtime {
...
+                    }),
+            )
+            .await
+        {
+            telio_log_debug!("Failed to check adapter state: {}", err);
```
djkarwowski: I'd also consider adjusting this log message, as it corresponds with the name that does not fully reflect the action taken underneath.

crates/telio-wg/src/wg.rs
```diff
@@ -77,6 +77,14 @@ pub trait WireGuard: Send + Sync + 'static {
     async fn reset_existing_connections(&self, exit_pubkey: PublicKey) -> Result<(), Error>;
     /// Set the ip stack for the adapter
     async fn set_ip_stack(&self, ip_stack: Option<IpStack>) -> Result<(), Error>;
+    /// Ensure that adapter is UP or DOWN
+    async fn check_adapter_state(
```
djkarwowski: The same naming concern as in the other comment.

crates/telio-core/src/device.rs
matislovas: Fixed

crates/telio-core/src/device.rs
matislovas: Fixed

crates/telio-wg/src/adapter.rs
matislovas: Yeah, naming clash was the primary concern. Fixed!

nat-lab/tests/test_adapter.py
matislovas: Removed

crates/telio-wg/src/wg.rs
matislovas: Fixed

crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
```diff
@@ -320,8 +320,16 @@ impl InterfaceWatcher {
...
-                        telio_log_info!("Skipping adatper.up() due to empty peer list");
+                    let meshnet_enabled = self
+                        .enable_dynamic_wg_nt_control
+                        .as_ref()
+                        .is_some_and(|is_meshnet_on_cb| is_meshnet_on_cb());
+                    if self.enable_dynamic_wg_nt_control.is_some()
+                        && (!has_peers && !meshnet_enabled)
+                    {
+                        telio_log_info!(
+                            "Skipping adatper.up() due to empty peer list and meshnet disabled"
```
gytsto: typo
```suggestion
                            "Skipping adapter.up() due to empty peer list and meshnet disabled"
```

gytsto (dismissed):
lgtm +1

djkarwowski (commented):
Looks good!

+1

gytsto (approved):

PR #1892, opened 2026-07-14
Align LogCensor with other copies

tomaszklak:
LogCensor ended up in couple of our libraries copied with small changes. This is a commit that updates
the original in libtelio with features from the copies.

Once this is delivered, other libraries will be able to use telio-utils crate (via git dependency) to use common LogCensor.

- [x] README.md is updated

### DIFF
--- a/crates/telio-utils/src/log_censor.rs
+++ b/crates/telio-utils/src/log_censor.rs
@@ -1,4 +1,3 @@
-use rand::RngExt;
 use regex_lite::{Captures, Match, Regex, RegexBuilder};
 use std::{
     net::{Ipv4Addr, Ipv6Addr},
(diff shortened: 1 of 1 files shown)


mathiaspeters:
+1

sfraczek (dismissed):
+1 it's good. 
One question though. Import of log censor drags entire telio-utils dependencies. 
(backtrace, tokio, boot-time, parking_lot, futures, hashlink, …), whereas LogCensor itself only needs regex-lite + blake3 + rand.
Do you think it's worth making it a separate crate?

tomaszklak:
> Do you think it's worth making it a separate crate?

No, the organizational cost of moving one small `.rs` into separate crate is not worth the trouble in my opinion.

sfraczek (approved):

PR #1893, opened 2026-07-15
nat-lab: fix SSH-read hang and teardown races behind the 2h CI wall-kill

gytsto:
## Problem

Natlab shards could hang to GitLab's 2h wall-kill, and one failing test took down the rest of the shard. Root causes:

- **SSH read hang (the 2h killer):** `asyncssh.connect()` set no keepalive/timeouts. When a test left a VM half-alive (TCP up, guest frozen, no RST), any SSH read blocked forever — e.g. `pytest_sessionfinish`'s `collect_logs` SSHing a degraded Mac VM hung 86 min. `wait_for` cancellation alone can't break this (channel close does I/O on the dead transport).
- **Whole-shard cascade:** when a test left a VM unreachable (e.g. VM_MAC after a failed suspend, `Errno 113`), `pytest_runtest_teardown`'s log collection raised on the dead connection, so `LOG_COLLECTORS.clear()` never ran. The dead collector persisted and every following test's teardown re-tried it and errored (`previous item was not torn down properly`) — one failing test became `1 failed -> 27 errors`.
- **Freeze/suspend reconnect race:** a just-unfrozen/resumed guest can accept a socket before it can run anything, so a single `reconnect()` raced recovery and failed with `Errno 113` (e.g. `test_no_burst_after_monotonic_jump[VM_WINDOWS_1]`).
- **Teardown crash every run:** `make_local_tcpdump` called `kill()` on an already-exited local tcpdump → `ProcessLookupError` aborted the whole session `exit_stack` teardown.
- A failed suspend test could also leave the VM paused, cascading setup ERRORs.

## Solution

- **Root fix (hang):** asyncssh `keepalive_interval=15` / `keepalive_count_max=4` (+ connect/login timeouts) so a frozen peer is dropped and reads raise `ConnectionLost` in ~75s instead of hanging.
- **Cascade fix:** isolate each collector in `pytest_runtest_teardown` with try/except and move `LOG_COLLECTORS.clear()` + log-file teardown into a `finally`, so one dead VM can't fail every following test's teardown.
- **Freeze/suspend readiness gate (in-test):** after unfreeze/resume, poll until a real command (`echo`) runs over the session, bounded to 90s, raising a clear `ConnectionError` if the guest never recovers — confirms the guest is actually up, and fails loud on genuine non-recovery. `SshConnection` is unchanged.
- Guard `process.kill()` in `make_local_tcpdump` against `ProcessLookupError`; demote docker kill-by-natlab-id exit 3 (already gone) from warning to debug.
- Send QEMU `cont` from a `finally` so an interrupted suspend always resumes.
- Backstops for non-SSH stalls: bound `pytest_sessionfinish` cleanup with a 600s `asyncio.wait_for`, and add `timeout=` to the sync `docker logs`/`docker cp`/`dmesg` calls in log collection.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -242,31 +242,28 @@ def pytest_runtest_teardown(item, nextitem):  # pylint: disable=unused-argument
     async def collect_all_logs():
         async with AsyncExitStack() as stack:
             for log_collector in LOG_COLLECTORS:
-                log.info(
-                    "[%s] Will run post-test log collection for %s",
-                    log_collector.node_name,
-                    log_collector.tag,
-                )
-                connection = await stack.enter_async_context(
-                    new_connection_raw(log_collector.tag)
-                )
-                await log_collector.cleanup(connection)
-                log.info(
-                    "[%s] Done running post-test log collection for %s",
-                    log_collector.node_name,
-                    log_collector.tag,
-                )
+                try:
+                    connection = await stack.enter_async_context(
+                        new_connection_raw(log_collector.tag)
+                    )
+                    await log_collector.cleanup(connection)
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    log.warning(
+                        "[%s] post-test log collection failed for %s: %s",
+                        log_collector.node_name,
+                        log_collector.tag,
+                        e,
+                    )
 
-    _SESSION.runner.run(collect_all_logs())
-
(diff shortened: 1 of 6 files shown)


jjanowsk (commented):
+1

tomaszpatejko:
+1

tomaszpatejko (dismissed):

jjanowsk (approved):

PR #1894, opened 2026-07-16
Automate Libtelio release process

asledzinskiy:
### Problem
*--Currently in order to release new Libtelio version we need to prepare additional PR with updated changelog.md, deleted .unreleased files and bumped cargo version. It prevents us from releasing latest commit passed nightly tests--*

### Solution
*--Introduce tag-driven release process. After tagging latest commit passed nightly tests - CI is triggered and `generate-release` job automatically updates changelog, prepare a clean up PR with deleted .unreleased files and bumped Cargo version to next expected released version. With the new process we always keep future released version in Cargo.toml--*

Done work:
- Update ci/release.py to be used as orchestrator that builds the changelog block from .unreleased/, inserts it into changelog.md on gh-pages and opens a post-release cleanup PR

- [x] README.md is updated


asledzinskiy:
Testing results:
auto-created PR - https://github.com/NordSecurity/libtelio/pull/1896
updated changelog - https://nordsecurity.github.io/libtelio/changelog/

ci/generate_changelog.py
```diff
@@ -45,6 +47,113 @@ def _match_ticket(input_str: str) -> Optional[str]:
     return None
 
 
+def version_key(version: str):
```
tomaszklak: can we use pre existing python semver library instead of implementing it on our own?

ci/generate_changelog.py
```diff
@@ -1,22 +1,22 @@
 """OS module for file operations,
 argparse module for command line arguments parsing,
 sys module for graceful exiting,
-regex for searching for old series name. re is does not support unicode characters,
+re for regex matching,
 Optional for optional type."""
```
tomaszklak: nit: imo, that kind of docstring is just noise

ci/github_helpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def github_api_request(method, repo, path, token, payload=None):
+    """Call the GitHub REST API and return (status_code, json). Returns (0, {...}) on failure."""
+    url = f"https://api.github.com/repos/{repo}/{path}"
+    data = json.dumps(payload).encode() if payload is not None else None
+    req = urllib.request.Request(url, data=data, method=method)
+    req.add_header("Authorization", f"Bearer {token}")
+    req.add_header("Accept", "application/vnd.github+json")
+    req.add_header("X-GitHub-Api-Version", "2022-11-28")
```
tomaszklak: I think we should use the latest version (`2026-03-10`).

ci/release.py
```diff
@@ -1,210 +1,288 @@
...
-            capture_output=True,
-            text=True,
-            shell=True,
-        )
-        .stdout.strip()
-        .split("\n")
+    return branch if refs.strip() else "main"
+
+
+def next_cargo_version(tag: str, source_branch: str, override: str) -> str:
```
tomaszklak: I don't see anything cargo related in the behaviour of this function. Is the name correct?

tomaszklak (commented):
I'm non sure that using markdown as the source of truth is a good idea, it's a format meant to be easy to work with for humans, but we want to treat it like a miniature **database** file.

If we really want to use md like that, I think that using regexes to interact with it is also a bad idea - I think we should use some pre-existing parser that will produce AST on which we will do our modifications/queries.

ci/release.py
asledzinskiy: Agree, will rename to `next_release_version`

ci/generate_changelog.py
asledzinskiy: Agree, I can use `packaging.version.Version` for that as it seems more suitable for our version naming as more strict `python-semver` library

asledzinskiy:
> I'm non sure that using markdown as the source of truth is a good idea, it's a format meant to be easy to work with for humans, but we want to treat it like a miniature **database** file.
> 
> If we really want to use md like that, I think that using regexes to interact with it is also a bad idea - I think we should use some pre-existing parser that will produce AST on which we will do our modifications/queries.

I would go with AST modification approach. 
If we want to change the way we store changelog - we will need some design session as I can't make such decision on my own.

asledzinskiy:
> > I'm non sure that using markdown as the source of truth is a good idea, it's a format meant to be easy to work with for humans, but we want to treat it like a miniature **database** file.
> > If we really want to use md like that, I think that using regexes to interact with it is also a bad idea - I think we should use some pre-existing parser that will produce AST on which we will do our modifications/queries.
> 
> I would go with AST modification approach. If we want to change the way we store changelog - we will need some design session as I can't make such decision on my own.

I used the parser only to locate version headings to insert new block. Full tree manipulation would re-serialize the whole file, reformatting every existing entry and original styling will be lost

tomaszklak:
> reformatting every existing entry and original styling will be lost

I don't think that would be a problem. But if I'm the only one who would prefer to have AST based changelog mutation, than feel free to ignore this, it's not critical after all.

Also, please squash commit 45a9f49fd41cfe8cc872684a3ecf8acb6a75eaeb.

ci/generate_changelog.py
```diff
@@ -45,6 +42,119 @@ def _match_ticket(input_str: str) -> Optional[str]:
...
+    headings = []
+    for i, token in enumerate(tokens):
+        if token.type == "heading_open" and token.tag == "h3" and token.map:
+            heading_text = tokens[i + 1].content.strip()
+            if VERSION_HEADING_REGEX.match(heading_text):
+                headings.append((heading_text, offsets[token.map[0]]))
+    return headings
+
+
+def insert_block_sorted(body: str, block: str, out_version: str) -> str:
```
mathiaspeters: I don't have any strong feelings about this so feel free to ignore but I feel like the changelog sections should be sorted by the time they were released, not by the semver version of the releases

mathiaspeters (commented):
+1

To add my two cents to the AST discussion: having to reformat the whole doc should not be a problem, and if whatever lib we're using to update it is good then it would apply a good and consistent formatting of the document and we don't have to care about that

.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v5.17.5
+TRIGGERED_REF=LLT-6792_add_auto_generate_release
```
Jauler: adding a reminder to change this to normal version once everything is merged up

ci/release.py
```diff
@@ -1,210 +1,289 @@
...
-        subprocess.run(
-            "git symbolic-ref refs/remotes/origin/HEAD",
-            capture_output=True,
-            text=True,
-            shell=True,
-        )
-        .stdout.strip()
-        .replace("refs/remotes/origin/", "")
+def resolve_source_branch_for_tag(github_repo: str, tag: str) -> str:
+    """Return the branch the tag was cut from: release/vX.Y if it exists, else main."""
```
Jauler, line 39: Would it make sense to do something on the lines of:
```bash
git branch -r --contains <tag>
```

Jauler (commented):
I have a question. What happens with the gh-pages published changelog if we first release v7.0.0 and later do some bugfix backport or whatever and release v6.2.5 for example? 

I wonder if it will retain "newer" (in terms of version number, in the example it would be v7.0.0) entries, e.g. ones from `main` instead of `release/v6.2` after the release process.

Jauler:
And one more question, if we are already publishing github pages, does it really make sense to also have changelog committed in the repository?

Maybe we can simply publish it as gh-pages from artifact of github actions job and this way simplify the flow a bit?

ci/release.py
asledzinskiy: This command may return multiple branches containing this commit and we still will need to filter and pick the branch

asledzinskiy:
> I have a question. What happens with the gh-pages published changelog if we first release v7.0.0 and later do some bugfix backport or whatever and release v6.2.5 for example?
> 
> I wonder if it will retain "newer" (in terms of version number, in the example it would be v7.0.0) entries, e.g. ones from `main` instead of `release/v6.2` after the release process.

 gh-pages changelog.md is a single shared file and `release.py` doesn't rebuild it fully, it inserts new entry (for v6.2.5) based on semver order. So v7.0.0 will be kept and new  v6.2.5 release entry will be inserted into correct semver position.
 
 
 Mathias raised a suggestion to append new entries always to the top instead of semver insert. So that we have release changelog based on time when it was released (not version)

asledzinskiy:
> And one more question, if we are already publishing github pages, does it really make sense to also have changelog committed in the repository?
> 
> Maybe we can simply publish it as gh-pages from artifact of github actions job and this way simplify the flow a bit?

The problem with GitHub Pages is that you can't read it through API to get source block in order to modify it and POST it back. So we need some durable source of truth to append to.

ci/release.py
```diff
@@ -1,210 +1,289 @@
...
-    with open("Cargo.toml") as f:
-        content = f.read()
 
-    if not re.search(r"^\[package\]", content, re.MULTILINE) or not re.search(
-        r'name\s*=\s*"telio"', content
-    ):
-        print('This does not appear to be a "libtelio" repo.')
-        sys.exit(1)
+def is_final_tag(tag: str) -> bool:
+    """True for a final release (v8.1.0), False for a pre-release (v8.0.0-rc2)."""
```
jjanowsk: We call them "stable" tags everywhere. Let's stick to that

ci/release.py
```diff
@@ -1,210 +1,289 @@
...
-        .split("\n")
+    return branch if refs.strip() else "main"
+
+
+def next_release_version(tag: str, source_branch: str, override: str) -> str:
+    if override:
+        return override.lstrip("v")
+    major, minor, patch = (int(x) for x in tag.lstrip("v").split("-")[0].split("."))
+    if source_branch == "main":
+        return f"{major}.{minor + 1}.0"  # next minor
```
jjanowsk, line 55: I guess that will need some manual adjustments from time to time, but it should not be a problem. I guess minor bump as a default for main is a good default

jjanowsk (commented):
My 2 cents regarding sorting: I would prefer to not sort the changelog by the version, but by the release date. Same way we've been doing so far. So that would also simplify markdown manipulation and kind of got rid of Tomasz's objections.

Other then that +1

asledzinskiy:
> My 2 cents regarding sorting: I would prefer to not sort the changelog by the version, but by the release date. Same way we've been doing so far. So that would also simplify markdown manipulation and kind of got rid of Tomasz's objections.
> 
> Other then that +1

Thanks, I'm changing to sorting by release date then.

matislovas:
+1. Let's see how that Github page will look like. I have some ideas on how to represent the release branches and backports, but those can be left to future updates :smile:

jjanowsk (approved):

PR #1895, opened 2026-07-17
LLT-4837: Fix data race in wg-nt MTU monitor callbacks

sfraczek:
### Problem
The route change and interface change callbacks are registered on separate
handles, so they can run concurrently on different threads. Both mutated the
monitor's cached MTU state without any synchronisation between them.

This finishes the root-cause work deferred by the temporary deadlock fix
in #307.

### Solution
Guard that state with a single mutex taken only by the callbacks, so
CancelMibChangeNotify2 in stop() can never wait on a callback that waits on
us. Also drop the per-handle mutexes in MtuMonitor and InterfaceWatcher,
since the callbacks no longer lock them.

The first commit is preparatory cleanup (removes write-only state) so the
fix commit stays minimal.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
+++ b/crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
@@ -27,12 +27,18 @@ use winapi::shared::{
 };
 
 pub struct InterfaceWatcher {
-    // Quickhack: Adapter is Sync+Send, so it cannot hold any substructures with raw ptr / HANDLE
-    iface_cb_handle: Arc<Mutex<usize>>, // iface_cb_handle: HANDLE,
+    // OS change-notification handle; see the `unsafe impl Send` below.
+    iface_cb_handle: HANDLE,
 
     watched_adapter: Arc<Mutex<AdapterConfiguration>>,
     enable_dynamic_wg_nt_control: IsMeshnetEnabledCb,
 }
+
+// SAFETY: `iface_cb_handle` is an opaque OS change-notification handle. It is
+// never dereferenced, only handed back to CancelMibChangeNotify2, which may be
+// called from any thread. The watcher is therefore safe to move between
+// threads.
+unsafe impl Send for InterfaceWatcher {}
 
 struct AdapterConfiguration {
     luid: u64,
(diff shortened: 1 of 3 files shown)


crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
```diff
@@ -28,8 +27,9 @@ use winapi::shared::{
 };
 
 pub struct InterfaceWatcher {
-    // Quickhack: Adapter is Sync+Send, so it cannot hold any substructures with raw ptr / HANDLE
-    iface_cb_handle: Arc<Mutex<usize>>, // iface_cb_handle: HANDLE,
+    // Stored as usize, not HANDLE: a raw-pointer field would make the whole
+    // struct !Send, and it is shared across threads as Arc<Mutex<InterfaceWatcher>>.
+    iface_cb_handle: usize,
```
tomaszklak: The 'trick' of using `usize` just hides an important property of the code (that it's not `Send` by it self and making it `Send` requires explicit `impl` with appropriate safety comment.

Please add a 3rd commit (preferably in between the already existing ones) that changes the types of handles (this one, but also `route_cb_handle` and `iface_cb_handle`) to be `HANDLE`.

crates/telio-wg/src/windows/tunnel/mtumonitor.rs
```diff
@@ -45,81 +46,62 @@ impl MtuMonitor {
...
-        if let Ok(mut route_cb_handle) = self.route_cb_handle.clone().lock() {
-            unsafe {
-                if 0 != *route_cb_handle {
-                    CancelMibChangeNotify2(*route_cb_handle as HANDLE);
-                    *route_cb_handle = 0;
-                }
+        unsafe {
+            if 0 != self.route_cb_handle {
+                CancelMibChangeNotify2(self.route_cb_handle as HANDLE);
+                self.route_cb_handle = 0;
```
tomasz-grz: Now that `route_cb_handle` lock is removed, is there any possibility that `stop()` or `Drop` might be called while `route_change_callback` is executing?

crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
sfraczek: done

tomaszklak (commented):
+1

crates/telio-wg/src/windows/tunnel/mtumonitor.rs
sfraczek: The reason `stop()`/`Drop` racing a running callback is safe is that `CancelMibChangeNotify2` blocks until the in-flight callback returns, and neither side touches a field the other uses (`stop()` only writes the handle fields, which the callbacks never read).

tomasz-grz (approved):
+1

PR #1899, opened 2026-07-21
nat-lab: bound fixture setup/teardown with a phase timeout

gytsto:
## Problem
nat-lab runs pytest with `--timeout=...` and `timeout_func_only=true`, so the per-test timeout guards **only the call phase**. A wedged SSH/RPC in fixture **setup** or **teardown** is unbounded — it hangs the shard until the 2h GitLab job ceiling.

Observed: `test_no_burst_after_monotonic_jump[VM_MAC-NEPTUN-CONE_CLIENT_2-]` printed its nodeid then produced no further output for ~1h33m until the job was killed (`execution took longer than 2h0m0s`). No `PASSED` line + func-only timeout not firing ⇒ hang was in the `env_mesh` **setup** phase (unbounded `ClientLog.wait_for` → `flush_logs` RPC to a wedged VM_MAC), which nothing bounds.

## Solution
Register a plugin that arms a `SIGALRM` phase timeout (300s) around **setup** and **teardown**. A stuck fixture now fails fast and names the phase+test (`setup exceeded 300s for <nodeid>`) instead of burning the whole 2h job. `SIGALRM` interrupts the blocked syscall via EINTR. Handler is saved/restored, so it coexists with pytest-timeout's call-phase alarm (verified: call-phase hang still reported by pytest-timeout, not this).

Existing `pytest_runtest_setup`/`teardown` hooks are untouched.

## Notes / limitations
- Bounds the symptom (no more 2h burn; culprit phase+test named). *Why* VM_MAC wedges after the cgroup-freeze is a separate investigation.
- 300s chosen from CI durations: slowest whole test observed ~212s (setup+call+teardown summed); a single phase is a fraction of that.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -280,7 +280,7 @@ class _PhaseTimeout:
             raise TimeoutError(f"{phase} exceeded {seconds}s for {nodeid}")
 
         old = signal.signal(signal.SIGALRM, _fire)
-        signal.setitimer(signal.ITIMER_REAL, seconds)
+        signal.setitimer(signal.ITIMER_REAL, seconds, seconds)
         return old
 
     @staticmethod
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -5,6 +5,7 @@ import logging
 import os
 import Pyro5  # type: ignore
 import pytest
+import signal
 import threading
 from contextlib import AsyncExitStack
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


djkarwowski (commented):
LGTM

+1

nat-lab/tests/conftest.py
```diff
@@ -268,6 +269,48 @@ async def collect_all_logs():
...
+
+
+class _PhaseTimeout:
+    @staticmethod
+    def _arm(seconds, phase, nodeid):
+        def _fire(signum, frame):  # pylint: disable=unused-argument
+            raise TimeoutError(f"{phase} exceeded {seconds}s for {nodeid}")
+
+        old = signal.signal(signal.SIGALRM, _fire)
+        signal.setitimer(signal.ITIMER_REAL, seconds)
```
sfraczek: The one-shot timer bounds teardown only until its first firing. The teardown phase runs all the test's fixture finalizers in a loop (`SetupState.teardown_exact` in `_pytest/runner.py`), catching each finalizer's exception and carrying on - so once the alarm has fired, the rest of the phase runs with no alarm armed. If a second finalizer also blocks on the same wedged VM, the shard hangs unbounded again.

Minimal repro on pytest 8.3.3 - this plugin verbatim, timeout set to 1s, one test:

```python
@pytest.fixture
def fixture_b():
    yield
    time.sleep(4)  # second wedged finalizer

@pytest.fixture
def fixture_a(fixture_b):
    yield
    time.sleep(4)  # first wedged finalizer

def test_two_hanging_finalizers(fixture_a):
    pass
```

One-shot: the run takes 5.03s against the 1s phase bound - finalizer A is cut at 1s, finalizer B then sleeps its full 4s unbounded. With the optional interval argument the alarm re-arms while the phase is still running, so every subsequent hang is cut too - the same repro finishes in 2.01s with both timeouts reported as an ExceptionGroup:

```suggestion
        signal.setitimer(signal.ITIMER_REAL, seconds, seconds)
```

Normal path unchanged - `_disarm` cancels a repeating timer just the same. Setup does not need this: a fixture exception aborts the setup phase outright.

gytsto: Fixed

sfraczek (approved):
+1

PR #1902, opened 2026-07-21
LLT-7319: Fix nat-lab for docker >=28.2

stalowyjez:
### Problem
In Docker 28.2 there are some firewall rules added which drops packets to nat-lab-internet from other networks, which breaks most of the nat-lab tests.

### Solution
Docker 28.0 also introduces `nat-unprotected` option for docker networks, which skips these rules for some particular network. In this PR we add a patch in nat-lab.py when such docker version is discovered.

- [x] README.md is updated


djkarwowski (commented):
+1

mathiaspeters (dismissed):
+1

djkarwowski (approved):

PR #1904, opened 2026-07-23
Bump neptun version

stalowyjez:
### Problem
NepTUN has a few minor fixes, which would be nice to have in telio too and uses dispatch2 instead of dispatch, so we can get rid of the latter one

### Solution
Bump NepTUN dependency version

- [x] README.md is updated


tomasz-grz (approved):
+2

PR #1905, opened 2026-07-23
nat-lab: keep macOS libtelio remote alive across VM suspend (LLT-4961)

gytsto:
### Problem
`test_no_burst_after_system_suspend[VM_MAC]` fails: after the VM resumes, SSH recovers but the libtelio RPC daemon is dead (Pyro port refused), so the post-resume ping hangs into the test timeout. macOS sshd reaps the exec'd `libtelio_remote` child when the SSH connection drops across the suspend; Windows OpenSSH keeps the orphan, which is why only mac fails.

### Solution
- `libtelio_remote.py`: `os.setsid()` on darwin so the daemon detaches into its own session and survives the SSH teardown/reconnect.
- `ssh_process.py`: tolerate transport loss from the background launcher task at teardown (session is rebuilt via `reconnect()`), so a runner stall past the ~60s asyncssh keepalive window doesn't turn into a teardown error. Body exceptions and `ProcessExecError` still propagate.

Validated on the real Mac VM: `system_suspend` + `monotonic_jump` pass; daemon PID confirmed identical across the suspend even when the old socket is genuinely dead.

- [x] README.md is updated

### DIFF
--- a/.unreleased/LLT-4961_mac_daemon_survive_suspend
+++ b/.unreleased/LLT-4961_mac_daemon_survive_suspend
@@ -1,0 +1,1 @@
+Fix nat-lab macOS VM suspend test: the libtelio remote process is now detached (setsid) so it survives the SSH session being torn down across a VM suspend, and SSH transport loss on a reconnected long-lived process no longer fails teardown.
--- a/nat-lab/tests/uniffi/libtelio_remote.py
+++ b/nat-lab/tests/uniffi/libtelio_remote.py
@@ -283,6 +283,15 @@ def main():
     container_ip = sys.argv[2]
     port = int(sys.argv[3])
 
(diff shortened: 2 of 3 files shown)


gytsto:
Closing: root cause is a **macOS guest kernel panic** on resume from the QEMU stop/cont suspend — `NMIPI spinlock timeout` in the AppleVirtIO-net RX path (confirmed by the guest `Kernel-*.panic` crash report in the failing job's artifacts; `coredumps/` empty = no userspace crash). It's a rare infra flake (~1/24 nightly pipelines ≈ 4%), not a libtelio or nat-lab-harness bug. The changes here (setsid is a proven no-op; the ssh_process teardown catch is unrelated) do not address it. Follow-up: retry-mark the mac variant + dockur ticket for the virtio-net resume panic.

PR #1906, opened 2026-07-23
Add cert printing on failed derp handshake

stalowyjez:
### Problem
We had a problem with debugging problems with DERP connections

### Solution
Let's add printing certs on failing Derp connections, so we have some more info what happens inside

- [x] README.md is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -188,6 +188,7 @@ x25519-dalek = { version = "2.0.1", features = [
 ] }
 zeroize = { version = "1.9.0", features = ["derive"] }
 hyper-util = "0.1.20"
+x509-parser = { version = "0.18.1", default-features = false }
 
 telio-crypto = { version = "0.1.0", path = "./crates/telio-crypto" }
 telio-dns = { version = "0.1.0", path = "./crates/telio-dns" }
--- a/crates/telio-relay/src/derp/http.rs
+++ b/crates/telio-relay/src/derp/http.rs
@@ -20,8 +20,9 @@ use webpki_roots::TLS_SERVER_ROOTS;
 
 use crate::{Config, DerpKeepaliveConfig};
 
-use rustls_platform_verifier::ConfigVerifierExt;
+use rustls_platform_verifier::Verifier as PlatformVerifier;
 use telio_crypto::{PublicKey, SecretKey};
+use telio_proto::CertLoggingVerifier;
 use tokio::time::Interval;
 use tokio::{
     io::{split, AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
@@ -29,7 +30,11 @@ use tokio::{
     time::timeout,
 };
 use tokio_rustls::{
-    rustls::{pki_types::ServerName, ClientConfig, RootCertStore},
+    rustls::{
+        client::{danger::ServerCertVerifier, WebPkiServerVerifier},
+        pki_types::ServerName,
+        ClientConfig, RootCertStore,
+    },
     TlsConnector,
 };
 use url::{Host, Url};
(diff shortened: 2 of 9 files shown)


crates/telio-relay/src/derp/http.rs
```diff
@@ -468,4 +604,123 @@ mod tests {
...
+
+        let fingerprint = hex::encode(Sha256::digest(self_signed.der().as_ref()));
+        let expected = format!(
+            "DERP TLS cert verification failed for DnsName(\"derp.example.com\"): \
+             invalid peer certificate: UnknownIssuer. \
+             Presented cert: issuer=[CN=Evil Proxy CA], subject=[CN=Evil Proxy CA], \
+             validity=[Jan  1 00:00:00 2024 +00:00 .. Jan  1 00:00:00 2034 +00:00], \
+             sha256={fingerprint}, intermediates=0"
+        );
+        assert_eq!(*logs.lock().unwrap(), vec![expected]);
```
gytsto: nit: This might break with any dependency bump without actual behavior change. Consider using `contains` instead of 1:1 string comparison

Cargo.toml
```diff
@@ -187,7 +187,8 @@ x25519-dalek = { version = "2.0.1", features = [
   "static_secrets",
 ] }
 zeroize = { version = "1.9.0", features = ["derive"] }
-hyper-util = "0.1.20"
+hyper-util = { version = "0.1.20", features = ["tokio"] }
```
gytsto: nit: consider moving `tokio` to workspace where it's actually needed instead of global usage

.unreleased/LLT-5911
```diff
@@ -0,0 +1 @@
+Log DERP server address on connection failure and presented certificate details (issuer, subject, validity, fingerprint) on TLS verification failure
```
gytsto: nit: I didn't see any DERP server address being printed anywhere 😄

gytsto (dismissed):
few nits (can be ignored), but looks good +1

.unreleased/LLT-5911
stalowyjez: oops, fixed ^^'

Cargo.toml
stalowyjez: moved

crates/telio-relay/src/derp/http.rs
stalowyjez: changed to `contains`

gytsto: perfect, that's like how I've imagined :smile:

.unreleased/LLT-5911
```diff
@@ -0,0 +1 @@
+Log presented certificate details (issuer, subject, validity, fingerprint) when DERP TLS certificate verification fails
```
jjanowsk: I think it's not important to the app teams. Now once we have automated changelog generation it's even more important to decide it at the moment of PR creation. Previously the author of the release could clean it up. So I would leave it empty.

jjanowsk (commented):
+1

.unreleased/LLT-5911
stalowyjez: Removed

crates/telio-relay/src/derp/http.rs
```diff
@@ -174,6 +191,125 @@ async fn try_connect(
...
+        Ok((_, cert)) => format!(
+            "DERP TLS cert verification failed for {:?}: {}. Presented cert: issuer=[{}], subject=[{}], validity=[{} .. {}], sha256={}, intermediates={}",
+            server_name,
+            err,
+            cert.issuer(),
+            cert.subject(),
+            cert.validity().not_before,
+            cert.validity().not_after,
+            fingerprint,
+            intermediates.len()
```
lcruz99: Could we log something extra about the intermediates or is it trivial?

crates/telio-relay/src/derp/http.rs
```diff
@@ -174,6 +191,125 @@ async fn try_connect(
     }
 }
 
+/// Decorator around the actual certificate verifier which logs details of the
```
lcruz99: telio-proto implements related functions, it would be good to DRY this and even create a new mod for certificate mgmt, if possible.. (on a separate PR/ticket probably)

crates/telio-relay/src/derp/http.rs
stalowyjez: We could, but is there anything particular you would like to log about them?

lcruz99: No, I'm just asking if it would be useful to show details from intermediate certs other than `intermediates.len()`

crates/telio-relay/src/derp/http.rs
stalowyjez: Added some refactor, is it what you meant?

lcruz99: yess! thanks

crates/telio-relay/src/derp/http.rs
stalowyjez: I think the number is just enough for now, so I'd leave it as it is - we can easily add more info in the feature if it will be needed

jjanowsk (dismissed):

jjanowsk (approved):

PR #1908, opened 2026-07-24
Make neptun an optional dependency of telio-sockets

tomaszklak:
When telio-sockets is used as a dependency of some other than libtelio library, there is no need to force neptun as a dependency.

- [x] README.md is updated


stalowyjez (approved):
Lgtm, +1

jjanowsk (approved):
+1

PR #1909, opened 2026-07-24
LLT-6855: Separation of daemon lifecycle from VPN connection state management

tomasz-kumor:
### Problem

Currently the VPN connection state is tightly coupled with the daemon lifecycle and there is not way to connect/disconnect VPN exit node without changing the NordVPN Lite daemon state.

### Solution

* Separate nordvpnlite daemon lifecycle management from VPN connection state management.
  *  Introduced `nordvpnlite start --do-not-connect` flag to allow separate launching the daemon from establishing VPN connection.
  * Introduced dedicated CLI `connect`/`disconnect` commands that allows to control the VPN connection state

### Testing

1. Run daemon without establishing VPN connection
```
# ./nordvpnlite start --do-not-connect -c config.json
```
Check that daemon is running but VPN connection is not established
```
# ./nordvpnlite status
{
  "telio_is_running": true,
  "ip_address": null,
  "exit_node": null
}
```

Connect to the VPN exit node & check that VPN connection is established
```
# ./nordvpnlite connect
Command executed successfully
# /nordvpnlite status
{
  "telio_is_running": true,
  "ip_address": null,
  "exit_node": {
    "identifier": "2432424-234536f34f3-4235f34-323",
    "public_key": "cnwurvevke4747tb43y3v74659g45gh45vd5e",
    "hostname": "lt.nordvpn.com",
    "endpoint": "181.88.88.88:51320",
    "state": "connected"
  }
}
```
Disconnect from the VPN exit node
```
# ./nordvpnlite disconnect
Command executed successfully
# ./nordvpnlite status
{
  "telio_is_running": true,
  "ip_address": null,
  "exit_node": null
}
```
2. Check legacy behavior

```
# ./nordvpnlite stop
# ./nordvpnlite start -c config.json
...
# /nordvpnlite status
{
  "telio_is_running": true,
  "ip_address": null,
  "exit_node": {
    "identifier": "2432424-234536f34f3-4235f34-323",
    "public_key": "cnwurvevke4747tb43y3v74659g45gh45vd5e",
    "hostname": "lt.nordvpn.com",
    "endpoint": "181.88.88.88:51320",
    "state": "connected"
  }
}
```

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/README.md
+++ b/clis/nordvpnlite/README.md
@@ -37,6 +37,7 @@ There is a command for running the daemon:
   * `--config-file <path_to_config_file>` - specify alternative configuration file,
   * `--no-detach` - run the nordvpnlite in the foreground as a regular process,
   without detaching from the terminal.
+  * `--do-not-connect` - starts daemon without establishing VPN connection.
   * `--stdout-path` - Redirect daemon standard output to the specified file,
   Defaults to `/var/log/nordvpnlite.log`, ignored when used with `--no-detach`
   Some early logs may still be printed to stdout before redirection.
--- a/clis/nordvpnlite/README.md
+++ b/clis/nordvpnlite/README.md
@@ -94,6 +94,8 @@ And following cli commands:
 * `nordvpnlite is-alive` - query if the daemon is running
 * `nordvpnlite stop` - stop daemon execution
 * `nordvpnlite reload` - reload the configuration file and restart the daemon
+* `nordvpnlite connect` - connect to the VPN exit node (non-blocking). Use `nordvpnlite status` to check the connection status
+* `nordvpnlite disconnect` - disconnect from the current VPN exit node
 * `nordvpnlite countries` - list countries with available VPN servers
 * `nordvpnlite login <token>` (or `nordvpnlite login --token <token>`) - store the authentication token obtained
   from [my.nordaccount.com](https://my.nordaccount.com) into the auth file
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -26,8 +26,6 @@ pub enum ClientCmd {
     GetStatus,
     #[clap(hide = true)]
     IsAlive,
-    #[clap(name = "stop", about = "Stop daemon execution")]
-    QuitDaemon,
     #[clap(name = "reload", about = "Reload config file and restart the daemon")]
     Reload,
 }
@@ -267,22 +265,6 @@ impl CommandListener {
                 })
                 .await
             }
-            ClientCmd::QuitDaemon => {
-                trace!("Quitting telio task");
-                let (response_tx, response_rx) = oneshot::channel();
-                #[allow(mpsc_blocking_send)]
-                self.telio_task_tx
-                    .send(TelioTaskCmd::Quit(response_tx))
-                    .await
-                    .map_err(|e| {
-                        error!("Error sending command: {}", e);
-                        NordVpnLiteError::CommandFailed(ClientCmd::QuitDaemon)
-                    })?;
-                // Wait for a response from TelioTask
-                // this essentially blocks the client quit command until the daemon initiated
-                // cleanup
-                handle_response(response_rx, |_| Ok(CommandResponse::Ok)).await
-            }
             ClientCmd::Reload => {
                 match RunningConfig::from_file(&self.config.path) {
                     Err(e) => {
@@ -344,7 +326,6 @@ impl CommandListener {
                 } else {
                     // Early command handling before TelioTask is initialized
                     match &command {
-                        ClientCmd::QuitDaemon => CommandResponse::Ok,
                         ClientCmd::IsAlive => CommandResponse::Ok,
                         ClientCmd::GetStatus
                         | ClientCmd::Reload
@@ -574,14 +555,6 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_command_quit() {
-        let (response, cmd) = test_command_helper(ClientCmd::QuitDaemon, true, false).await;
-
-        assert_eq!(response.unwrap(), CommandResponse::Ok);
-        assert_eq!(cmd.unwrap(), ClientCmd::QuitDaemon);
-    }
-
-    #[tokio::test]
     async fn test_command_is_alive() {
         let (response, cmd) = test_command_helper(ClientCmd::IsAlive, true, false).await;
 
@@ -638,14 +611,6 @@ mod tests {
         let (_, cmd) = test_command_helper(ClientCmd::GetStatus, true, true).await;
 
         assert_matches!(cmd, Err(NordVpnLiteError::Io(_)));
-    }
-
-    #[tokio::test]
-    async fn test_command_early_quit() {
-        let (response, cmd) = test_command_helper(ClientCmd::QuitDaemon, false, false).await;
-
-        assert_eq!(cmd.unwrap(), ClientCmd::QuitDaemon);
-        assert_eq!(response.unwrap(), CommandResponse::Ok);
     }
 
     #[tokio::test]
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -71,6 +71,10 @@ pub(crate) struct DaemonOpts {
     /// Ignored with no-detach flag.
     #[clap(long = "stdout-path", default_value = "/var/log/nordvpnlite.log")]
     pub stdout_path: String,
+
+    /// Do not connect to the exit node
+    #[clap(long = "do-not-connect")]
+    pub do_not_connect: bool,
 }
 
 #[derive(Parser, Debug)]
@@ -128,8 +132,8 @@ pub(crate) struct LogoutOpts {
 #[command(version = concat!(env!("CARGO_PKG_VERSION"), " (", env!("LIBTELIO_COMMIT_SHA"), ") ", env!("BUILD_PROFILE")))]
 pub enum Cmd {
     #[clap(about = "Runs the nordvpnlite event loop")]
-    Start(DaemonOpts),
-    #[clap(flatten)]
+    Daemon(DaemonOpts),
+    #[command(flatten)]
     Client(ClientCmd),
     #[clap(about = "Show countries with available VPN servers")]
     Countries,
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -71,6 +71,10 @@ pub(crate) struct DaemonOpts {
     /// Ignored with no-detach flag.
     #[clap(long = "stdout-path", default_value = "/var/log/nordvpnlite.log")]
     pub stdout_path: String,
+
+    /// Do not connect to the exit node
+    #[clap(long = "do-not-connect")]
+    pub do_not_connect: bool,
 }
 
 #[derive(Parser, Debug)]
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -1,16 +1,15 @@
 use std::net::IpAddr;
 
-use clap::Parser;
-use serde::{Deserialize, Serialize};
-use telio_core::telio_task::io::chan;
-use tokio::sync::oneshot;
-use tracing::{error, info, trace};
-
 use crate::{
     comms::{DaemonConnection, DaemonSocket},
     config::{Endpoint, RunningConfig},
     daemon::{NordVpnLiteError, TelioStatusReport},
 };
+use clap::Parser;
+use serde::{Deserialize, Serialize};
+use telio_core::telio_task::io::chan;
+use tokio::sync::oneshot;
+use tracing::{error, info, trace};
 
 pub(crate) const TIMEOUT_SEC: u64 = 60;
 const DEFAULT_CONFIG_PATH: &str = "/etc/nordvpnlite/config.json";
@@ -19,6 +18,10 @@ const DEFAULT_CONFIG_PATH: &str = "/etc/nordvpnlite/config.json";
 #[clap()]
 #[derive(Serialize, Deserialize)]
 pub enum ClientCmd {
+    #[clap(about = "Connect to the exit node")]
+    Connect,
+    #[clap(about = "Disconnect from the exit node")]
+    Disconnect,
     #[clap(name = "status", about = "Retrieve the status report")]
     GetStatus,
     #[clap(hide = true)]
@@ -40,8 +43,8 @@ pub struct ExitNodeConfig {
 pub enum TelioTaskCmd {
     // Get telio status
     GetStatus(oneshot::Sender<TelioStatusReport>),
-    // Connect to exit node with endpoint and optional hostname
-    ConnectToExitNode(ExitNodeConfig),
+    Connect(oneshot::Sender<Result<(), String>>),
+    Disconnect(oneshot::Sender<Result<(), String>>),
     // Break the receive loop to quit the daemon and exit gracefully
     Quit(oneshot::Sender<()>),
 }
@@ -230,6 +233,40 @@ impl CommandListener {
                 })
                 .await
             }
+            ClientCmd::Connect => {
+                trace!("Connect");
+                let (response_tx, response_rx) = oneshot::channel();
+                #[allow(mpsc_blocking_send)]
+                self.telio_task_tx
+                    .send(TelioTaskCmd::Connect(response_tx))
+                    .await
+                    .map_err(|_| {
+                        error!("Failed to send Connect command to telio task");
+                        NordVpnLiteError::CommandFailed(ClientCmd::Connect)
+                    })?;
+                handle_response(response_rx, |result| match result {
+                    Ok(()) => Ok(CommandResponse::Ok),
+                    Err(e) => Ok(CommandResponse::Err(e)),
+                })
+                .await
+            }
+            ClientCmd::Disconnect => {
+                trace!("Disconnect");
+                let (response_tx, response_rx) = oneshot::channel();
+                #[allow(mpsc_blocking_send)]
+                self.telio_task_tx
+                    .send(TelioTaskCmd::Disconnect(response_tx))
+                    .await
+                    .map_err(|_| {
+                        error!("Failed to send Disconnect command to telio task");
+                        NordVpnLiteError::CommandFailed(ClientCmd::Disconnect)
+                    })?;
+                handle_response(response_rx, |result| match result {
+                    Ok(()) => Ok(CommandResponse::Ok),
+                    Err(e) => Ok(CommandResponse::Err(e)),
+                })
+                .await
+            }
             ClientCmd::QuitDaemon => {
                 trace!("Quitting telio task");
                 let (response_tx, response_rx) = oneshot::channel();
@@ -309,9 +346,10 @@ impl CommandListener {
                     match &command {
                         ClientCmd::QuitDaemon => CommandResponse::Ok,
                         ClientCmd::IsAlive => CommandResponse::Ok,
-                        ClientCmd::GetStatus | ClientCmd::Reload => {
-                            CommandResponse::DaemonInitializing
-                        }
+                        ClientCmd::GetStatus
+                        | ClientCmd::Reload
+                        | ClientCmd::Connect
+                        | ClientCmd::Disconnect => CommandResponse::DaemonInitializing,
                     }
                 };
                 connection.respond(response.serialize()).await?;
@@ -378,10 +416,15 @@ mod tests {
                     let status_report = TelioStatusReport::default();
                     response_tx_channel.send(status_report).unwrap();
                 }
+                TelioTaskCmd::Connect(response_tx_channel) => {
+                    response_tx_channel.send(Ok(())).unwrap();
+                }
+                TelioTaskCmd::Disconnect(response_tx_channel) => {
+                    response_tx_channel.send(Ok(())).unwrap();
+                }
                 TelioTaskCmd::Quit(response_tx_channel) => {
                     response_tx_channel.send(()).unwrap();
                 }
-                _ => {}
             }
         });
 
@@ -469,6 +512,65 @@ mod tests {
 
         let error_response = CommandResponse::Err("Test error".to_string());
         assert_eq!(error_response.serialize(), "{\"Err\":\"Test error\"}");
+    }
+
+    #[tokio::test]
+    async fn test_command_connect() {
+        let (response, cmd) = test_command_helper(ClientCmd::Connect, true, false).await;
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Connect);
+    }
+
+    async fn test_command_error_helper(command: ClientCmd, error_msg: &str) {
+        let path = make_socket_path();
+        let (tx, mut task_rx) = mpsc::channel::<TelioTaskCmd>(1);
+
+        let err = error_msg.to_string();
+        task::spawn(async move {
+            match task_rx.recv().await.unwrap() {
+                TelioTaskCmd::Connect(response_tx) => {
+                    response_tx.send(Err(err)).unwrap();
+                }
+                TelioTaskCmd::Disconnect(response_tx) => {
+                    response_tx.send(Err(err)).unwrap();
+                }
+                _ => {}
+            }
+        });
+
+        let socket = DaemonSocket::new(Path::new(&path)).unwrap();
+        let (config, _config_file) = make_running_config();
+        let mut listener = CommandListener::new(socket, tx, config);
+
+        let command_str = serde_json::to_string(&command).unwrap();
+        let daemon = tokio::spawn(async move {
+            let connection = listener.accept_client_connection().await.unwrap();
+            listener.handle_client_command(true, connection).await
+        });
+
+        let response = client_send_command(&path, &command_str).await;
+        let cmd = daemon.await.unwrap();
+
+        assert_eq!(
+            response.unwrap(),
+            CommandResponse::Err(error_msg.to_string())
+        );
+        assert_eq!(cmd.unwrap(), command);
+    }
+
+    #[tokio::test]
+    async fn test_command_connect_disconnect_error() {
+        test_command_error_helper(ClientCmd::Connect, "simulated connect failure").await;
+        test_command_error_helper(ClientCmd::Disconnect, "simulated disconnect failure").await;
+    }
+
+    #[tokio::test]
+    async fn test_command_disconnect() {
+        let (response, cmd) = test_command_helper(ClientCmd::Disconnect, true, false).await;
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Disconnect);
     }
 
     #[tokio::test]
--- a/clis/nordvpnlite/src/command_listener.rs
+++ b/clis/nordvpnlite/src/command_listener.rs
@@ -1,16 +1,15 @@
 use std::net::IpAddr;
 
-use clap::Parser;
-use serde::{Deserialize, Serialize};
-use telio_core::telio_task::io::chan;
-use tokio::sync::oneshot;
-use tracing::{error, info, trace};
-
 use crate::{
     comms::{DaemonConnection, DaemonSocket},
     config::{Endpoint, RunningConfig},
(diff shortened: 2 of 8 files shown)


clis/nordvpnlite/src/daemon.rs
```diff
@@ -132,6 +136,13 @@ impl ExitNodeStatus {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum VpnConnectionState {
+    LoggedOut,
```
tomaszpatejko: What does `VpnConnectionState::LoggedOut` encode? How is it used?

tomasz-kumor: `LoggedOut` is an initial state. It encodes situation where the user did not provide NordVPN authentication credentials (`nordvpnlite login`).

Without valid credentials the `nordvpnlite start` fails and the daemon is not launched.

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -129,7 +133,7 @@ pub(crate) struct LogoutOpts {
 pub enum Cmd {
     #[clap(about = "Runs the nordvpnlite event loop")]
     Start(DaemonOpts),
-    #[clap(flatten)]
+    #[command(flatten)]
```
lcruz99: what's the reason for this change?

clis/nordvpnlite/src/daemon.rs
tomaszpatejko: Understood. I'm wondering whether this initial state is actually needed. I'm just wondering whether using `VpnConnectionState` as 
```
enum VpnConnectionState {
    Connect,
    Disconnect,
}

connection_state: Option<VpnConnectionState>;
```
with `connection_state = None` set initially for the situation you mentioned, would be sufficient. This would mean that the user has not attempted establishing any connection yet.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -466,15 +470,16 @@ async fn run_daemon(config: RunningConfig) -> Result<LoopOutcome, NordVpnLiteErr
...
-    if init_done_rx.await.is_ok() {
-        // TODO: This can be triggered through nordvpnlite command to allow the user to stop/restart.
-        let config_clone = config.parsed.clone();
-        let tx_clone = telio_tx.clone();
-        tokio::spawn(async move {
-            handle_exit_node_connection(&config_clone, tx_clone).await;
-            debug!("Exit node connection task completed");
-        });
+    if !do_not_connect {
+        // Wait for interface setup to complete before initiating connection.
```
lcruz99: trivial comment, next line is self-explanatory. I know you're not the author but since you're touching it I couldn't avoid

clis/nordvpnlite/src/daemon.rs
```diff
@@ -466,15 +470,16 @@ async fn run_daemon(config: RunningConfig) -> Result<LoopOutcome, NordVpnLiteErr
...
+        if init_done_rx.await.is_ok() {
+            let config_clone = config.parsed.clone();
+            let tx_clone = telio_tx.clone();
+            tokio::spawn(async move {
```
lcruz99: IMO is this one that deserves a comment instead

clis/nordvpnlite/src/main.rs
```diff
@@ -102,7 +102,11 @@ fn main() -> Result<(), NordVpnLiteError> {
 
         // Run the daemon event loop.
         let rt = tokio::runtime::Runtime::new()?;
-        rt.block_on(daemon::daemon_event_loop(config, &mut logging_handle))
+        rt.block_on(daemon::daemon_event_loop(
+            config,
+            &mut logging_handle,
+            opts.do_not_connect,
```
lcruz99: perhaps we can rename it to manual mode? Passing it to the runtime it seems like is not going to connect, ever.

clis/nordvpnlite/README.md
```diff
@@ -37,6 +37,7 @@ There is a command for running the daemon:
   * `--config-file <path_to_config_file>` - specify alternative configuration file,
   * `--no-detach` - run the nordvpnlite in the foreground as a regular process,
   without detaching from the terminal.
+  * `--do-not-connect` - starts daemon without establishing VPN connection.
```
lcruz99, line 40: https://github.com/NordSecurity/libtelio/pull/1909/changes/dac60d2c749612267433945e5b0f22175f2fe29b#r3784619117

tomasz-kumor: The name of this flag was suggested in https://nordsec.atlassian.net/browse/LLT-6855

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: Reverted

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: Still not my change but added a comment

clis/nordvpnlite/src/main.rs
tomasz-kumor: Changed to `connect_on_startup`

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: IMO changing it to optional only complicates code. `None` is a state similar to `LoggedOut` so we still have three states. In the context of VPN state keeping it as it is now I think make sense.

tomasz-kumor: I have slightly modified the logic and followed your idea. Internally in the Telio context we have just two states `connected`/`disconnected`. The decision whether on startup/reload connect to the exit node or not is moved higher to `daemon_event_loop` where I used optional VPN state to differentiate between normal daemon launch and reload.

tomaszpatejko: Looks fine.

tomaszpatejko (dismissed):

Cargo.lock
lcruz99, line 1: what's the reason for this downgrade?

tomasz-kumor: I just called
```
cargo update -p h2
```
which was suggested as a fix by `deny` CI job

tomasz-kumor: Ok same fix was already merged on main (see 30c250e6e57a15bf16fd2478e878e005f50e2d8e)

clis/nordvpnlite/src/daemon.rs
```diff
@@ -361,24 +477,41 @@ async fn handle_exit_node_connection(config: &NordVpnLiteConfig, tx: mpsc::Sende
...
+    Reload {
+        config: Box<RunningConfig>,
+        connection_state: VpnConnectionState,
+    },
 }
 
 pub async fn daemon_event_loop(
     mut config: RunningConfig,
     logging_handle: &mut logging::LoggingHandle,
+    auto_connect: bool,
```
lcruz99, line 501: Perhaps this param belongs to the config? Not saying to ditch the CLI flag, both can coexist

clis/nordvpnlite/src/interface.rs
```diff
@@ -392,6 +392,18 @@ impl ConfigureInterface for Iproute {
...
+            "-6",
+            "route",
+            "del",
+            "default",
+            "dev",
+            &self.interface_name,
+        ])) {
+            error!("Failed to remove IPv6 default route: {e}");
+        }
+
```
lcruz99, line 406: Seeing this makes me wonder: is there anything else to be cleaned for IPv6? (eg: tables, routes, fwmarks)

clis/nordvpnlite/src/daemon.rs
```diff
@@ -132,6 +136,12 @@ impl ExitNodeStatus {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum VpnConnectionState {
+    Disconnected,
+    Connected,
+}
+
```
lcruz99: This enum is needless, the exit node status should be checked instead. See examples in the code

lcruz99: isn't what TelioTaskCmd::GetStatus is doing? IIUC, you have created `enum VpnConnectionState` which is intrinsically Daemon's object, however this is an exclusive property of telio, the protocol lib.

nat-lab/tests/test_nordvpnlite.py
```diff
@@ -311,3 +312,239 @@ async def test_nordvpnlite_pq_vpn_connection(config_provider: str) -> None:
...
+        nordvpnlite = await NordVpnLite.new(
+            exit_stack,
+            config_data=NordVpnLiteConfig(vpn=VPNConfig(country="pl")),
+            do_not_connect=True,
+        )
+        await nordvpnlite.request_credentials_from_core()
+
+        async with nordvpnlite.start():
+            await nordvpnlite.wait_for_telio_running_status()
+
```
lcruz99, line 475: You should assert here that the exit node is disconnected

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -226,6 +234,40 @@ impl CommandListener {
                 })
                 .await
             }
+            ClientCmd::Connect => {
+                trace!("Connect");
```
lcruz99, line 239: Maybe we can wrap self.telio_task_tx.send() and trace!() on every call. Wdyt?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +510,65 @@ mod tests {
         (daemon_response, client_command)
     }
```
lcruz99, line 528: You're not the author but if you don't mind could you document test_command_helper()?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +510,65 @@ mod tests {
...
+        assert_eq!(
+            response.unwrap(),
+            CommandResponse::Err(error_msg.to_string())
+        );
+        assert_eq!(cmd.unwrap(), command);
+    }
+
+    #[tokio::test]
+    async fn test_command_connect_disconnect_error() {
+        test_command_error_helper(ClientCmd::Connect, "simulated connect failure").await;
```
lcruz99: '"simulated connect failure"' message doesn't add any info, pls add more details like what was the test intention and what failed

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +510,65 @@ mod tests {
...
+        test_command_error_helper(ClientCmd::Disconnect, "simulated disconnect failure").await;
```
lcruz99: same as above

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +510,65 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_command_connect() {
+        let (response, cmd) = test_command_helper(ClientCmd::Connect, true, false).await;
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Connect);
+    }
+
+    async fn test_command_error_helper(command: ClientCmd, error_msg: &str) {
```
lcruz99, line 531: pls move helper functions to the top of the mod, in such a way that the tests are an uninterrupted sequence of code blocks.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -262,55 +328,105 @@ impl TelioTaskCmd {
...
+                        .is_err()
+                    {
+                        error!("Connect: receiver dropped before result could be sent");
+                    }
+                    return Ok(TelioTaskOutcome::Continue);
+                }
+
+                let endpoint_result = ctx
+                    .tokio_handle
+                    .block_on(get_server_endpoints_list(&ctx.config))
```
lcruz99: This API call can block the whole daemon for long time, and I think would be best to avoid moving the runtime around.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -145,12 +155,17 @@ struct TelioContext {
     interface_config_provider: Box<dyn ConfigureInterface>,
     config: NordVpnLiteConfig,
     exit_node: Option<ExitNodeStatus>,
+    connection_state: VpnConnectionState,
+    /// Handle to the tokio runtime. Used to run async tasks (e.g. HTTP calls) from
```
lcruz99, line 154: > Used to run async tasks

 You're using it to run sync tasks.

tomasz-kumor: I am calling the get_server_endpoints_list() which is async.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -244,6 +261,55 @@ impl TelioContext {
         }
         Ok(())
     }
+    fn connect_to_exit_node(
+        &mut self,
+        exit_node_config: ExitNodeConfig,
+    ) -> Result<(), NordVpnLiteError> {
+        self.interface_config_provider
+            .set_exit_routes(&exit_node_config.endpoint.address, &exit_node_config.dns)
```
lcruz99, line 285: set_exit_routes() can be called twice if handle_exit_node_connection() is still fetching the endpoints, this should be fixed by making the endpoints fetch async though.

nat-lab/tests/test_nordvpnlite.py
```diff
@@ -311,3 +312,239 @@ async def test_nordvpnlite_pq_vpn_connection(config_provider: str) -> None:
...
+            assert status.get("exit_node") is not None, (
+                "Expected VPN connection to remain active, "
+                f"but exit_node={status.get('exit_node')}"
+            )
+            assert status["exit_node"]["state"] == "connected", (
+                f"Expected exit_node state 'connected', "
+                f"got {status['exit_node']['state']!r}"
+            )
+
+            log.debug("Confirmed: VPN connection intact after duplicate connect")
```
lcruz99, line 372: You're asserting that nordvpnlite reports connected, you should (also) assert that the connection is working independently of what the app says.

nat-lab/tests/test_nordvpnlite.py
```diff
@@ -311,3 +312,239 @@ async def test_nordvpnlite_pq_vpn_connection(config_provider: str) -> None:
...
+            await nordvpnlite.disconnect()
+            await nordvpnlite.wait_for_vpn_disconnected_state()
+
+            status = json.loads(await nordvpnlite.get_status())
+            assert status.get("exit_node") is None, (
+                f"Expected no VPN connection after disconnect, "
+                f"but exit_node={status.get('exit_node')}"
+            )
+
+            log.debug("Confirmed: VPN disconnected")
```
lcruz99, line 548: I think these tests can be already split into their own categories/classes/mods, e.g.: TestLog, TestConnection.. just like libtelio.

tomasz-kumor:
Rebased on top of main

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: Not really. It would require a bigger refactoring. Right now the `RunningConfig` contains the parsed data from the configuration file and some information about the configuration file itself.  Command line parameter does not fit to it now.

clis/nordvpnlite/src/interface.rs
tomasz-kumor: This was added to make the cleanup implementation symetric to `set_exit_routes()`. Without that there were sometimes problems with reconnecting after the VPN connection was disconnected.

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: That's a good point. I have changed the implementation that if exit node was in connected or connecting state after reload, the app will try to re-establish the VPN connection.

nat-lab/tests/test_nordvpnlite.py
tomasz-kumor: fixed

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: This change won't remove too much duplications. I would prefer not try to refactor this code in scope of this PR.

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: Added

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: Slightly improved it by adding comments and documenting the helper function.

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: same as above

clis/nordvpnlite/src/command_listener.rs
tomasz-kumor: Fixed

nat-lab/tests/test_nordvpnlite.py
tomasz-kumor: How should I do that?

nat-lab/tests/test_nordvpnlite.py
tomasz-kumor: They could be but there are not that many of them yet. 3 tests for connect/disconnect command and 2 for reload.

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: Fixed, see the top commit

clis/nordvpnlite/src/daemon.rs
tomasz-kumor: Fixed, see the top commit

clis/nordvpnlite/src/daemon.rs
```diff
@@ -244,6 +277,66 @@ impl TelioContext {
...
+fn reject_connect(
+    tx: oneshot::Sender<Result<(), String>>,
+    err: NordVpnLiteError,
+) -> Result<TelioTaskOutcome, NordVpnLiteError> {
+    warn!("{err}");
+    if tx.send(Err(err.to_string())).is_err() {
+        error!("Connect: receiver dropped before result could be sent");
+    }
+    Ok(TelioTaskOutcome::Continue)
 }
```
lcruz99, line 340: How does this look like? can you share a snippet of the logs?

clis/nordvpnlite/src/daemon.rs
```diff
@@ -145,12 +151,21 @@ struct TelioContext {
     interface_config_provider: Box<dyn ConfigureInterface>,
     config: NordVpnLiteConfig,
     exit_node: Option<ExitNodeStatus>,
+    /// Handle to the tokio runtime. Used to run async tasks (e.g. HTTP calls) from
+    /// within the spawn_blocking context of start_listening_commands
+    tokio_handle: tokio::runtime::Handle,
+    /// Sender to send commands back to the telio task loop
+    telio_tx: mpsc::Sender<TelioTaskCmd>,
+    /// Flag to track if a connect is pending to avoid double-connect race
+    pending_connect: bool,
```
lcruz99, line 160: perhaps this should be a state machine, updated periodically or/and whenever getstatus is called

clis/nordvpnlite/src/interface.rs
lcruz99: so let's change the fn name to unset_exit_routes()

clis/nordvpnlite/src/interface.rs
lcruz99, line 384: Not affected by this PR, but I think we should log something here. eg: "Unsupported".

clis/nordvpnlite/src/daemon.rs
```diff
@@ -92,6 +92,12 @@ pub enum NordVpnLiteError {
     IpRoute,
     #[error("Endpoint has no PublicKey")]
     EndpointNoPublicKey,
+    #[error("Already connected to exit node")]
+    AlreadyConnected,
+    #[error("Connection to exit node already in progress")]
+    ConnectionPending,
+    #[error("Not connected to exit node")]
+    NotConnected,
```
lcruz99, line 100: Abstract or reuse errors, these errors all have the same base, e.g.: CommandFailed.

I mean, these are error details not errors themselves.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -178,21 +193,39 @@ impl TelioContext {
...
+            pending_connect: false,
         })
     }
 
     fn start_listening_commands(
         &mut self,
         mut rx_channel: mpsc::Receiver<TelioTaskCmd>,
-    ) -> Result<(), NordVpnLiteError> {
+    ) -> Result<bool, NordVpnLiteError> {
+        let mut was_connected = false;
```
lcruz99, line 206: Why is `was_connected` added? Also find it a bit confusing/non-intuitive that `fn start_listening_commands()` returns it.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -262,55 +355,127 @@ impl TelioTaskCmd {
...
-                            exit_node.endpoint.address,
-                            exit_node.endpoint.public_key,
-                            exit_node.endpoint.hostname.as_deref().unwrap_or_default()
-                        );
-                        let external_nodes = ctx.telio.external_nodes()?;
-                        let exit_node =
-                            external_nodes.iter().find(|node| node.is_exit).map(|node| {
-                                ExitNodeStatus::from_node(node, exit_node.endpoint.hostname)
+            TelioTaskCmd::Disconnect(response_tx) => {
+                let result = (|| {
```
lcruz99, line 359: why the closure?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -226,6 +235,51 @@ impl CommandListener {
...
+                    .await
+                    .map_err(|_| {
+                        error!("Failed to send Disconnect command to telio task");
+                        NordVpnLiteError::CommandFailed(ClientCmd::Disconnect)
+                    })?;
+                handle_response(response_rx, |result| match result {
+                    Ok(()) => Ok(CommandResponse::Ok),
+                    Err(e) => Ok(CommandResponse::Err(e)),
+                })
+                .await
```
lcruz99, line 281: any reason for blocking here and not on the upper ClientCmd::Connect arm?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -305,9 +359,10 @@ impl CommandListener {
                     match &command {
                         ClientCmd::QuitDaemon => CommandResponse::Ok,
                         ClientCmd::IsAlive => CommandResponse::Ok,
-                        ClientCmd::GetStatus | ClientCmd::Reload => {
-                            CommandResponse::DaemonInitializing
-                        }
+                        ClientCmd::GetStatus
+                        | ClientCmd::Reload
+                        | ClientCmd::Connect
+                        | ClientCmd::Disconnect => CommandResponse::DaemonInitializing,
```
lcruz99, line 365: at this point we should consider replacing these commands with `_`

lcruz99:
OpenWrt tests at https://github.com/NordSecurity/libtelio/blob/kumor/LLT-6855_only_one_way_of_stopping_nordvpnlite/nat-lab/tests/test_openwrt.py should be adapted, they are still using start/stopping the daemon on every reconnect. It's critical that these new changes are tested on that platform.

To avoid missing them, maybe is a good idea to put them in the same dir or couple them smw
```
nat-lab
├──tests
         ├──nordvpnlite
                 ├── test_nordvpnlite.py
                 └── test_openwrt.py
```

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +526,76 @@ mod tests {
         (daemon_response, client_command)
     }
 
+    /// Sends `command` to a fake daemon that replies with `error_msg`
+    /// and asserts that the error is propagated to the client.
+    async fn test_command_error_helper(command: ClientCmd, error_msg: &str) {
+        let path = make_socket_path();
```
lcruz99, line 532: this is not a test but a helper, remove the `test_` prefix.

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +526,76 @@ mod tests {
...
+        });
+
+        let response = client_send_command(&path, &command_str).await;
+        let cmd = daemon.await.unwrap();
+
+        assert_eq!(
+            response.unwrap(),
+            CommandResponse::Err(error_msg.to_string())
+        );
+        assert_eq!(cmd.unwrap(), command);
```
lcruz99, line 565: assertions should be made on tests, not on helpers.

clis/nordvpnlite/src/daemon.rs
```diff
@@ -178,21 +193,39 @@ impl TelioContext {
...
         &mut self,
         mut rx_channel: mpsc::Receiver<TelioTaskCmd>,
-    ) -> Result<(), NordVpnLiteError> {
+    ) -> Result<bool, NordVpnLiteError> {
+        let mut was_connected = false;
         while let Some(cmd) = rx_channel.blocking_recv() {
             trace!("TelioTask got command {:?}", cmd);
+
+            if matches!(cmd, TelioTaskCmd::Quit(_)) {
+                was_connected = self.is_connected()?;
```
lcruz99, line 211: why return if not connected?

clis/nordvpnlite/src/interface.rs
lcruz99: also, add also error handling for the previous commands. The cleanup should handle command failures.

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +526,76 @@ mod tests {
         (daemon_response, client_command)
     }
 
+    /// Sends `command` to a fake daemon that replies with `error_msg`
+    /// and asserts that the error is propagated to the client.
+    async fn test_command_error_helper(command: ClientCmd, error_msg: &str) {
```
lcruz99, line 531: no point of using a string literal for error_msg, use a owned type instead ie: `String`

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +526,76 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_command_connect() {
+        let (response, cmd) = test_command_helper(ClientCmd::Connect, true, false).await;
+
+        assert_eq!(response.unwrap(), CommandResponse::Ok);
+        assert_eq!(cmd.unwrap(), ClientCmd::Connect);
+    }
+
+    #[tokio::test]
+    async fn test_command_disconnect_error() {
```
lcruz99, line 577: where's the assertions?

clis/nordvpnlite/src/command_listener.rs
```diff
@@ -458,6 +526,76 @@ mod tests {
...
+        // Check that the telio error raised during disconnecting is propagated to the client
+        test_command_error_helper(ClientCmd::Disconnect, "disconnect failure").await;
+    }
+
+    #[tokio::test]
+    async fn test_command_connect_returns_ok_immediately() {
```
lcruz99, line 583: what is this test doing? is literally the same as `test_command_connect()`

clis/nordvpnlite/src/daemon.rs
lcruz99: or if `external_nodes()` fail at https://github.com/NordSecurity/libtelio/blob/c336ab9fdb9ce4f3944139758c13e897ffc35943/clis/nordvpnlite/src/daemon.rs#L225

nat-lab/tests/test_nordvpnlite.py
lcruz99: https://github.com/NordSecurity/libtelio/blob/c336ab9fdb9ce4f3944139758c13e897ffc35943/nat-lab/tests/test_nordvpnlite.py#L102-L106

lcruz99: add those asserts wherever adequate

clis/nordvpnlite/src/daemon.rs
```diff
@@ -262,55 +355,127 @@ impl TelioTaskCmd {
...
+                Ok(TelioTaskOutcome::Continue)
+            }
             TelioTaskCmd::Quit(response_tx_channel) => {
-                ctx.telio.stop();
                 _ = ctx.interface_config_provider.cleanup().inspect_err(|e| {
                     error!("Failed to cleanup interface and routes with error '{e:?}'")
                 });
 
+                // Has to be after cleanup to avoid issues with missing device
+                ctx.telio.stop();
```
lcruz99, line 477: this is a workaround for a faulty cleanup. If the device is missing it shouldn't fail cleanup but instead handle that failure. It should be resilient.

PR #1910, opened 2026-07-27
nat-lab: bump macOS VM image to v0.0.11 (spinlock watchdog fix)

gytsto:
### Problem
The macOS guest kernel-panics (`NMIPI for spinlock acquisition timeout`) on resume from the QEMU `stop`/`cont` suspend in `test_no_burst_after_system_suspend[VM_MAC]` when the CI host is loaded (~4% of nightly runs; guest `Kernel-*.panic` in job artifacts). XNU's spinlock watchdog counts TSC ticks and the guest TSC jumps by the full pause, so a lock contended at the stop instant reads as a pause-long spin at resume and trips the sub-second default threshold. The 60s cgroup freeze in `test_no_burst_after_monotonic_jump[VM_MAC]` has the same exposure.

### Solution
Bump the nat-lab macOS VM image to dockur_macos v0.0.11, which sets boot-arg `slto_us=180000000` — the watchdog threshold moves above any pause used by the sleep tests (30s stop/cont, 60s freeze). Validated against a forced-load reproducer (4 vCPUs, 2× host CPU oversubscription): 4/20 panics at baseline → 0/25 with the boot-arg.

- [x] README.md is updated


Paintree (approved):
+2

PR #1912, opened 2026-07-27
Make log download parallel and add common timeout

stalowyjez:
### Problem
Recently we got a test timeout and no artifacts left after nat-lab job

### Solution
Let's make log download parallel and add them a common 4 minute timeout, so we can handle log download failures gracefully

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/conftest.py
+++ b/nat-lab/tests/conftest.py
@@ -62,6 +62,9 @@ END_TASKS: threading.Event = threading.Event()
 CURRENT_TEST_LOG_FILE = None
 _LIBFIREWALL_SO = os.path.join(os.path.dirname(__file__), "uniffi", "libfirewall.so")
 
+TEARDOWN_PHASE_TIMEOUT_S = 300
+COLLECT_ALL_LOGS_TIMEOUT_S = TEARDOWN_PHASE_TIMEOUT_S - 60
+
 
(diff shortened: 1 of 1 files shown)


nat-lab/tests/conftest.py
```diff
@@ -56,6 +56,9 @@
 SETUP_CHECK_DUPLICATE_IP_RETRIES = 1
 
 SESSIONFINISH_CLEANUP_TIMEOUT_S = 600
+# Kept below TEARDOWN_PHASE_TIMEOUT_S so the graceful asyncio timeout
+# (warn and keep already-collected logs) wins over the SIGALRM phase abort.
```
sfraczek: What do you think about something like this?
```python
COLLECT_ALL_LOGS_TIMEOUT_S = TEARDOWN_PHASE_TIMEOUT_S - 60.
```
I think we can skip the comment entirely if we make the value derived from teardown_phase_timeout in code. The graceful timeout + warning is easy to follow in code.

sfraczek (dismissed):
+1

nat-lab/tests/conftest.py
stalowyjez: fixd

sfraczek (approved):
+2

PR #1913, opened 2026-07-28
LLT-5433: make the per-test error allowlist actually apply

sfraczek:
### Problem
The per-test error allowlist never applied. `Client._allowed_errors` started as `None` and `allow_errors()` replaced the list instead of extending it, so `LogCollector` - which captures the list when the client starts, before any test body runs - kept the initial `None`.

### Solution
Initialise `_allowed_errors` to an empty list and extend it in place, so the reference `LogCollector` holds stays valid.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/libtelio_client/client.py
+++ b/nat-lab/tests/libtelio_client/client.py
@@ -76,7 +76,7 @@ class Client:
         self._libtelio_proxy: Optional[LibtelioProxy] = None
         self._proxy_port = ""
         self._fingerprint: Optional[tuple[str, str]] = None
-        self._allowed_errors: Optional[List[re.Pattern]] = None
+        self._allowed_errors: List[re.Pattern] = []
         self._log = ClientLog(self)
         self._events_facade = ClientEvents(self)
         self._vpn = ClientVpn(self)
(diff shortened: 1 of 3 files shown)


tomaszpatejko:
+1

tomaszpatejko (approved):

tomasz-grz (approved):
+1

olekoliinyk:
Hey @sfraczek 
Can you please elaborate more why we hide such error? `neptun::device.*Fatal read error`
Shouldn't it be investigated in the first place why it occurs?

sfraczek:
> Hey @sfraczek Can you please elaborate more why we hide such error? `neptun::device.*Fatal read error` Shouldn't it be investigated in the first place why it occurs?

I see this error truly whitelisted only in `test_adapter_gone_event`. The test deletes the interface, then if it succeeds, it adds this error to whitelist of this test because this is expected error to be found in the logs then. This code shouldn't hide it anywhere else. Because the checking happens after the test has executed, if the executions stops sooner than the allowlisting happens, the whitelist should stay empty and the error in logs should be surfaced.

PR #1915, opened 2026-07-29
WG dynamic control bugfix

matislovas:
### Problem
In recent PR https://github.com/NordSecurity/libtelio/pull/1891, there was a bug left out. When disabling meshnet, with VPN still connected, we tear down the adapter. Root cause - we were only countings meshnet peers in `set_config()`, but excluding the possible VPN peer.

### Solution
Added additional check (counting exit node as peer), when turning meshnet off. Added explicit test to address the case.

- [x] README.md is updated

### DIFF
--- a/crates/telio-core/src/device.rs
+++ b/crates/telio-core/src/device.rs
@@ -1989,7 +1989,6 @@ impl Runtime {
             .iter()
             .map(|p| p.base.public_key)
             .collect();
-        let peers_cnt = peers.len();
 
         // Update for proxy and derp config
         if let Some(config) = config {
@@ -2150,6 +2149,22 @@ impl Runtime {
             }
         }
 
+        // Note: If the exit_node is a meshnet peer it will be counted twice
+        // (and it does not take into account magic_dns virtual peer),
+        // but since we only need this count for
+        // `ensure_expected_adapter_state()` to be `wg_peers_cnt > 0`,
+        // to NOT disable the wg-nt adapter, it does not really matter, how will
+        // exit_node will be counted, as long as it is taken into account.
+        let wg_peers_cnt = {
+            (self
+                .requested_state
+                .meshnet_config
+                .as_ref()
+                .and_then(|config| config.peers.as_ref())
+                .map_or(0, |peers| peers.len()))
+                + (self.requested_state.exit_node.as_ref().map_or(0, |_| 1))
+        };
+
         wg_controller::consolidate_wg_state(&self.requested_state, &self.entities, &self.features)
             .boxed()
             .await?;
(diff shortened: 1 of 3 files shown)


gytsto (commented):
+1

crates/telio-core/src/device.rs
```diff
@@ -2149,6 +2148,15 @@ impl Runtime {
                 }
             }
         }
+        let wg_peers_cnt = {
+            (self
+                .requested_state
+                .meshnet_config
+                .as_ref()
+                .and_then(|config| config.peers.as_ref())
```
tomasz-grz, line 2163: Does this cover also magic dns virtual peer?

crates/telio-core/src/device.rs
```diff
@@ -2149,6 +2148,15 @@ impl Runtime {
...
+                .map_or(0, |peers| peers.len()))
+                + (self.requested_state.exit_node.as_ref().map_or(0, |_| 1))
```
tomasz-grz, line 2165: If Exit node is Meshnet peer, would it be counted twice?

tomasz-grz (commented):
Looks mostly good, just some small details

crates/telio-core/src/device.rs
matislovas: Yeah, this is a valid point. It will be counted twice, but it won't effect the result, because in this case, it only needs to know that `peer_count > 0`. Maybe I can leave a comment about it ..?

crates/telio-core/src/device.rs
matislovas: It does only count the peers which are in mesh config. Can the magic dns be enabled without meshnet?

crates/telio-core/src/device.rs
tomasz-grz: Yes please

crates/telio-core/src/device.rs
tomasz-grz: I don't think its part of the mesh config, since it's a WG virtual peer.
🤔 Yes Magic DNS can be enabled without Meshnet, for VPN only.

But I wonder if then the `exit_node` case might "save" it?

I don't think we would have a case with with no VPN and no Meshnet... What would happen if you have meshnet enabled with 0 peers, and no VPN?

matislovas: The adapter would still be on, that was the idea of original PRs https://github.com/NordSecurity/libtelio/pull/1891 .

matislovas: This was the reason, why libdrop was not working sometimes for windows :)

crates/telio-core/src/device.rs
matislovas: Added comment

tomasz-grz (approved):
+1

PR #1918, opened 2026-07-30
Detect sleep burst by tick spacing instead of count

gytsto:
## Problem
`test_no_burst_after_system_suspend[VM_WINDOWS_1]` flaked due to inproper logic of check. Infra lag might result the output and it gets asserted on hardcoded amount of logs.

## Solution
Instead of checking how much log prints there were, check for the timestamp spacing between them. If its less than 1 second, we assume that the burst has happened.

- [x] README.md is updated


Paintree (approved):
+1

mathiaspeters (approved):
+1

PR #1919, opened 2026-07-30
Add extra debug messages in case of BadZipfile in fetch artifacts

olekoliinyk:
### Problem
In case we've got `BadZipfile` when fetching the artifacts - it's not clear why it's happening.

### Solution
Add extra debug messaged to `_get_artifacts` method

- [x] README.md is updated


djkarwowski (approved):
+2

PR #1924, opened 2026-08-03
LLT-7539: Censor logged domains

tomasz-grz:
### Problem
Domains without trailing dot (like `example.com`) were left uncensored in the logs.

### Solution
- Match domains with or without the trailing dot (`example.com` and `example.com.`), in URLs (`http:://example.com`), and in wildcard notation (`*.example.com`).
- The protocol, wildcard, and paths stay readable (`http://DOMAIN(H1)/search?q=x`).
- Hosts are canonicalized before hashing (lowercased, `www.` stripped), to produce the same DOMAIN(H1).
- False positives are filtered out: matches preceded with path segments (`telio-utils/src/lib.rs`), file names with extensions used in the codebase (`main.rs`, `config.json`, etc), and enum forms (`TelioLogLevel.DEBUG`).
- Accepted trade offs: unknown extensions are over censored (`update.zip`), and mixed-case domains (`Example.COM`) are leaked.
- Unit tests with deterministic helper functions
- Fixed multi byte char offset

- [x] README.md is updated

### DIFF
--- a/crates/telio-utils/src/log_censor.rs
+++ b/crates/telio-utils/src/log_censor.rs
@@ -27,6 +27,9 @@ pub enum LogCensorMode {
     EmptyString,
 }
 
+const FILE_EXTENSION_ALLOWLIST: &[&str] =
+    &["rs", "json", "toml", "log", "so", "dylib", "dll", "exe"];
+
 impl Default for LogCensor {
     fn default() -> Self {
         #[allow(clippy::expect_used)]
@@ -68,8 +71,11 @@ impl Default for LogCensor {
                 )
                 |
                 (?<DOMAIN>
+                    (?<SCHEME>[a-zA-Z][a-zA-Z0-9+.\-]*://)?
+                    (\*?\.)?
                     (([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)+
-                    ([A-Za-z]|[A-Za-z][A-Za-z]*[A-Za-z])\.
+                    [a-zA-Z]{2,}
+                    \.?
                 )
                 ",
         )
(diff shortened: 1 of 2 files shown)


crates/telio-utils/src/log_censor.rs
```diff
@@ -102,12 +111,83 @@ impl LogCensor {
...
+                .iter()
+                .any(|deny| ext.eq_ignore_ascii_case(deny))
+        })
+    }
+
+    /// A match is treated as a domain only when its casing is
+    /// unambiguous.
+    ///
+    /// Mixed forms like `TelioLogLevel.DEBUG`
+    /// are are most likely enums that appear in logs.
```
stalowyjez: "are are" -> "are"?

crates/telio-utils/src/log_censor.rs
```diff
@@ -102,12 +111,83 @@ impl LogCensor {
...
+            .rsplit('.')
+            .next()
+            .is_some_and(|tld| tld.chars().all(|c| c.is_ascii_lowercase()));
+        if tld_is_lowercase {
+            return false;
+        }
+        !candidate
+            .chars()
+            .filter(|c| c.is_ascii_alphabetic())
+            .all(|c| c.is_ascii_uppercase())
```
stalowyjez: Is there any reason for them to be different? Also, maybe we could do that in a single iteration?

crates/telio-utils/src/log_censor.rs
```diff
@@ -27,6 +27,12 @@ pub enum LogCensorMode {
     EmptyString,
 }
 
+/// File extensions that could appear in logs as `name.ext` and
+/// would otherwise be indistinguishable from a domain (`.rs` is a
+/// real TLD).
```
stalowyjez: I guess the name here is quite self-explanatory, maybe this comment is not really needed

Paintree (dismissed):
+1

stalowyjez (commented):
looks rather good, just a few nits, +1

crates/telio-utils/src/log_censor.rs
tomasz-grz: It's checking for two things, if the TLD is all lowercase or the string is all upper case.

So:
`EXAMPLE.COM` -> censored
`example.com` -> censored
`Example.com` -> censored
`Example.COM` -> pass
`Example.Com` -> pass

Rewrote it as single iteration.

crates/telio-utils/src/log_censor.rs
tomasz-grz: Removed

crates/telio-utils/src/log_censor.rs
tomasz-grz: Removed

Paintree (approved):

PR #1931, opened 2026-08-05
LLT-7558: Domain accessible after being removed from whitelist

tomaszpatejko:
### Problem

Domain is still accessible after it has been removed from whitelist.
The following use case trigger the bug:
1. Enable libtelio with libfirewall;
2. Try to access a domain. Whitelist is empty initially so the DNS query is sent to the blocking TP-Lite DNS server. The query result is NXDOMAIN that indicates that the domain is blocked. 
3. Whitelist the domain. Nonblocking DNS server responds with an IP address that is cached in DNS resolver cache.
4. Reblock the domain by clearing the whitelist with an empty domain list.
5. Try to access the domain again. Bug: the IP address is served from resolver cache that was written is step 3. Expected behaviour should be that blocking TP-Lite DNS server is queried for an IP address and response should be NXDOMAIN.

### Solution

Flush DNS resolver's cache when whitelist is set. This was no stale cache entries will be present and correct TP-Lite DNS server with be consulted for blocked domain's IP address.

AI disclosure: Claude was used for the following:
- log analysis and rootcausing,
- natlab test generation,
- fix suggestion and generation.

The work was also human-reviewed.

- [x] README.md is updated

### DIFF
--- a/crates/telio-dns/src/forward.rs
+++ b/crates/telio-dns/src/forward.rs
@@ -178,6 +178,17 @@ impl ForwardAuthority {
             resolver,
         })
     }
+
+    /// Drop every entry from the resolver's cache.
+    ///
+    /// Removes both positive and negative (NXDOMAIN) entries. Needed when
+    /// something outside DNS changes what an answer ought to be - see LLT-7558,
+    /// where a TP-Lite whitelist change alters whether a query gets redirected,
+    /// but a cached answer would be returned without the query ever reaching
+    /// the firewall.
+    pub fn flush_cache(&self) {
+        self.resolver.clear_cache();
+    }
 }
 
 #[async_trait::async_trait]
--- a/crates/telio-firewall/src/firewall.rs
+++ b/crates/telio-firewall/src/firewall.rs
@@ -429,10 +429,27 @@ impl StatefulFirewall {
     /// Set the TP-Lite DNS whitelisting configuration: the whitelisted domains
     /// and the (blocking, standard) DNS server redirect pairs. Both are applied
     /// together, reconfiguring the firewall to DNAT-redirect matching queries.
-    pub fn set_tp_lite_domain_whitelist(&self, domains: Vec<String>, redirects: Vec<DnsRedirect>) {
-        self.configured_state.write().tp_lite_whitelisted_domains = domains;
-        self.configured_state.write().tp_lite_dns_redirects = redirects;
+    ///
+    /// Returns `true` when the set of whitelisted domains differs from the one
+    /// previously configured. Callers use this to decide whether caches keyed on
+    /// the whitelist need invalidating - see LLT-7558. Changing only the
+    /// redirects does not affect which domains bypass blocking, so it does not
+    /// set the flag.
+    pub fn set_tp_lite_domain_whitelist(
+        &self,
+        domains: Vec<String>,
+        redirects: Vec<DnsRedirect>,
+    ) -> bool {
+        let domains_changed = {
+            let mut configured_state = self.configured_state.write();
+            let changed = configured_state.tp_lite_whitelisted_domains != domains;
+            configured_state.tp_lite_whitelisted_domains = domains;
+            configured_state.tp_lite_dns_redirects = redirects;
+            changed
+        };
+
         self.refresh_chain();
+        domains_changed
     }
 }
 
--- a/.unreleased/LLT-7558_domain_available_after_removing_from_whitelist
+++ b/.unreleased/LLT-7558_domain_available_after_removing_from_whitelist
@@ -1,0 +1,1 @@
+Flush DNS resolver cache when whitelist is modified.
--- a/crates/telio-core/src/device.rs
+++ b/crates/telio-core/src/device.rs
@@ -796,7 +796,7 @@ impl Device {
     ) -> Result {
         self.async_runtime()?.block_on(async {
             task_exec!(self.rt()?, async move |rt| {
-                Ok(rt.set_tp_lite_domain_whitelist(domains, redirects))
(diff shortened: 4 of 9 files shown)


matislovas:
You can squash `c8ac042` commit with the `3ec2acf`.

crates/telio-dns/src/forward.rs
```diff
@@ -178,6 +178,17 @@ impl ForwardAuthority {
...
+
+    /// Drop every entry from the resolver's cache.
+    ///
+    /// Removes both positive and negative (NXDOMAIN) entries. Needed when
+    /// something outside DNS changes what an answer ought to be - see LLT-7558,
+    /// where a TP-Lite whitelist change alters whether a query gets redirected,
+    /// but a cached answer would be returned without the query ever reaching
+    /// the firewall.
+    pub fn flush_cache(&self) {
+        self.resolver.clear_cache();
```
matislovas, line 190: Does this blocks until all requests in-flight are finished, and only than clears the cache?

crates/telio-firewall/src/firewall.rs
```diff
@@ -429,10 +429,25 @@ impl StatefulFirewall {
...
+    /// previously configured. Callers use this to decide whether caches keyed on
+    /// the whitelist need invalidating - see LLT-7558.
+    pub fn set_tp_lite_domain_whitelist(
+        &self,
+        domains: Vec<String>,
+        redirects: Vec<DnsRedirect>,
+    ) -> bool {
+        let domains_changed = {
+            let mut configured_state = self.configured_state.write();
+            let changed = configured_state.tp_lite_whitelisted_domains != domains;
```
matislovas, line 443: Shouldn't we compare the `redirects` as well?

tomaszpatejko: I don't think `redirects` should be compared in this situation.

If I understand the comment for `set_tp_lite_domain_whitelist`, `redirects` contains pairs of DNS servers used for blocking or passing through a domain. So the change in this list does not drive the decision whether resolver cache should be flushed.

crates/telio-dns/src/forward.rs
tomaszpatejko: If I understand the code correctly, TrustDNS' resolver uses `sync::Cache` from `moka-rs` crate for its cache implementation, and the entire cache invalidation is delegated to that crate.

Moka-rs' `sync::Cache::invalidate_all` method does not remove any entries and just records current time as invalidation time for all the entries in the cache. Check the comment here:  https://github.com/moka-rs/moka/blob/a616ec19e8d4ed938caf8b2c88090331d778d5da/src/sync/cache.rs#L1658

Actual removal of invalidated entries is delegated to `Housekeeper` that is triggered by `get`/`insert` methods. `Housekeeper` does not block reads but it blocks writes.

matislovas: But as I see it, it still can cause problems with requests-in-flight. Before flushing the cache, we should check, if there are some and either:
1. Wait for them to finish and put to cache.
2. Cancel them, for them not to end in cache.

crates/telio-firewall/src/firewall.rs
matislovas: My idea was to check both - the domains and redirects:

```
let changed_domains = configured_state.tp_lite_whitelisted_domains != domains;
let changed_redirects= configured_state.tp_lite_dns_redirects != redirects;
...
(changed_domains || changed_redirects)
```

`redirects` can influence DNAT usage, so my idea was to compare them as well.

crates/telio-dns/src/forward.rs
matislovas: The only situation, where we can avoid this problem with request-in-flight is that cache time entry (used to invalidate entries) is actually `request_started` and NOT `request_put_to_cache`

crates/telio-firewall/src/firewall.rs
jjanowsk: I think we can get rid of this discussion by flashing on every call to set_tp_lite_domain_whitelist. Realistically it's only called when: a) user added/or removed something from whitelist, b) user enabled or disabled tp. In both cases we want to flush the cache. So I don't think it's worth optimizing for some hard to imagine scenario.

crates/telio-dns/src/forward.rs
tomaszpatejko: I think I don't understand the issue you are trying to point out here. Could you elaborate a bit more?

jjanowsk: Imagine the DNS request was sent for a domain. Let's say it was sent to regular DNS server, because the domain was whitelisted. Now the user removes this domain from the whitelist. App will reconfigure our firewall rules and we will flush the cache. At this time the DNS response arrives and it is added to the cache. That's the issue. The problem is I'm not sure it is easily fixable with hickory resolver but maybe there is some way to do that. Maybe even by recreating it? I don't know you would need to research that part.

matislovas: Yeah, and invalidating cache entries by time does not remove the problem.

crates/telio-firewall/src/firewall.rs
tomaszpatejko: Agreed. The cache is flushed whenever `set_tp_lite_domain_whitelist` is called. Could you have a look again?

crates/telio-dns/src/forward.rs
jjanowsk: We've decided to release with this as a know issue and it's tracked under: LLT-7677

matislovas (approved):
+1

jjanowsk (approved):
+1

PR #1933, opened 2026-08-05
Increase openwrt-gw-03 healthcheck timeout

asledzinskiy:
### Problem
*--Current openwrt-gw-03 timeout is 195 seconds that is way less than 600 seconds we allocate on VM boot defined by QEMU_CONFIG_TIMEOUT. Docker kills container after 3 minutes and try to re-create it multiple times that causes nat-lab instability --*

### Solution
*--Align health check  timeout with 600 seconds timeout we set for VM--*

- [x] README.md is updated


olekoliinyk:
+2.0

olekoliinyk (approved):

PR #1934, opened 2026-08-06
Enable client side keepalives for the ENS connection

tomaszklak:
### Problem
If after the ENS connection is established, the connection dies in a silent manner (socket is open but no traffic is received) we will not detect this and the ENS connection will stay useless.

### Solution
Enable client side keepalives on the grpc/http2 connection used by ENS. With the default configuration we should detect the problem and try to reconnect after (120+20) seconds.

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -566,6 +566,21 @@ pub struct FeatureErrorNotificationService {
     /// DER encoded root certificate to be used for verification of all TLS connections
     /// to gRPC ENS endpoint in place of the hardcoded one
     pub root_certificate_override: Option<Vec<u8>>,
+
+    /// Interval between the keep alive messages sent over the ENS connection, in seconds.
+    /// When the underlying tcp connection stops working, the new connection will
+    /// be re-created after at most keepalive_interval_s + keepalive_timeout_s seconds.
+    /// When set to null, the keep alives are disabled.
+    /// Setting it to 0 is an error.
+    #[default(Some(120))]
+    pub keepalive_interval_s: Option<u32>,
+
+    /// How long to wait for a response to a keep alive message before considering the ENS
+    /// connection dead, in seconds. Only used when keepalive_interval_s is set.
+    /// When set to null, the underlying http client default is used
+    /// Setting it to 0 is an error.
+    #[default(Some(20))]
+    pub keepalive_timeout_s: Option<u32>,
 }
 
 impl std::fmt::Debug for FeatureErrorNotificationService {
@@ -574,6 +589,8 @@ impl std::fmt::Debug for FeatureErrorNotificationService {
             .field("buffer_size", &self.buffer_size)
             .field("allow_only_pq", &self.allow_only_pq)
             .field("backoff", &self.backoff)
+            .field("keepalive_interval_s", &self.keepalive_interval_s)
+            .field("keepalive_timeout_s", &self.keepalive_timeout_s)
             .field(
                 "root_certificate_override",
                 &self
@@ -867,7 +884,9 @@ mod tests {
                         buffer_size: 42,
                         allow_only_pq: true,
                         backoff: Default::default(),
-                        root_certificate_override: None
+                        root_certificate_override: None,
+                        keepalive_interval_s: Some(120),
+                        keepalive_timeout_s: Some(20),
                     })
                 }
             );
@@ -1061,6 +1080,8 @@ mod tests {
                         maximal_s: Some(120),
                     },
                     root_certificate_override: None,
+                    keepalive_interval_s: Some(120),
+                    keepalive_timeout_s: Some(20),
                 }),
                 error_notification_service
             );
@@ -1076,6 +1097,8 @@ mod tests {
                     allow_only_pq: true,
                     backoff: Default::default(),
                     root_certificate_override: None,
+                    keepalive_interval_s: Some(120),
+                    keepalive_timeout_s: Some(20),
                 }),
                 error_notification_service
             );
@@ -1093,6 +1116,8 @@ mod tests {
                         maximal_s: Some(120),
                     },
                     root_certificate_override: None,
+                    keepalive_interval_s: Some(120),
+                    keepalive_timeout_s: Some(20),
                 }),
                 error_notification_service
             );
@@ -1111,6 +1136,8 @@ mod tests {
                         maximal_s: None
                     },
                     root_certificate_override: None,
+                    keepalive_interval_s: Some(120),
+                    keepalive_timeout_s: Some(20),
                 }),
                 error_notification_service
             );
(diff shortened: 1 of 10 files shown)


tomaszklak:
The exact default values are still discussed.

crates/telio-proto/src/ens/ens_impl.rs
```diff
@@ -96,6 +103,15 @@ impl ErrorNotificationService {
             );
         }
 
+        if keepalive.interval == Some(Duration::ZERO) {
```
matislovas, line 106: If we pass `None`, than the defaults will still be used. Maybe we should reserve `Some(0)` case, for disabling keepalives? Unlikely to be used, but just maybe ...? 😃

matislovas: And it is a bit ambiguous against the comment in `features.rs` - `/// Setting it to 0 is an error.`.

tomaszklak: > If we pass None, than the defaults will still be used.

Hm, I don't think I understand your point here.

> Maybe we should reserve Some(0) case, for disabling keepalives?

What's wrong with disabling them with None?

> And it is a bit ambiguous against the comment in features.rs - /// Setting it to 0 is an error..

Why? Because we don't fail when such user error happens? Not every error is a fatal error :) Do you think it should be a fatal error and telio should refuse to work when the value is zero?

src/libtelio.udl
```diff
@@ -909,6 +909,21 @@ dictionary FeatureErrorNotificationService {
     /// DER encoded root certificate to be used for verification of all TLS connections
     /// to gRPC ENS endpoint in place of the hardcoded one
     bytes? root_certificate_override;
+
+    /// Interval between the keep alive messages sent over the ENS connection, in seconds.
+    /// When the underlying tcp connection stops working, the new connection will
+    /// be re-created after at most keepalive_interval_s + keepalive_timeout_s seconds.
+    /// When set to null, the keep alives are disabled.
+    /// Setting it to 0 is an error.
+    /// [default 120s]
```
sfraczek, line 919: PR description says 45

crates/telio-proto/src/ens/ens_impl.rs
```diff
@@ -96,6 +103,15 @@ impl ErrorNotificationService {
...
+            telio_log_warn!("Keepalive interval set to 0, reseting to default");
+            keepalive.interval = Some(Duration::from_secs(120));
+        }
+        if keepalive.timeout == Some(Duration::ZERO) {
+            telio_log_warn!("Keepalive timeout set to 0, reseting to default");
```
sfraczek:
```suggestion
            telio_log_warn!("Keepalive interval set to 0, resetting to default");
            keepalive.interval = Some(Duration::from_secs(120));
        }
        if keepalive.timeout == Some(Duration::ZERO) {
            telio_log_warn!("Keepalive timeout set to 0, resetting to default");
```

crates/telio-model/src/features.rs
```diff
@@ -566,6 +566,21 @@ pub struct FeatureErrorNotificationService {
     /// DER encoded root certificate to be used for verification of all TLS connections
     /// to gRPC ENS endpoint in place of the hardcoded one
     pub root_certificate_override: Option<Vec<u8>>,
+
+    /// Interval between the keep alive messages sent over the ENS connection, in seconds.
+    /// When the underlying tcp connection stops working, the new connection will
+    /// be re-created after at most keepalive_interval_s + keepalive_timeout_s seconds.
```
sfraczek: The `interval + timeout` bound covers detecting the dead connection, but the new connection is only created after the ENS restart backoff elapses on top of that (2 s initial with the default bounds)

```suggestion
    /// When the underlying tcp connection stops working, the dead connection will be
    /// detected after at most keepalive_interval_s + keepalive_timeout_s seconds and
    /// a new connection will be created after the ENS backoff elapses.
```

src/libtelio.udl
```diff
@@ -909,6 +909,21 @@ dictionary FeatureErrorNotificationService {
     /// DER encoded root certificate to be used for verification of all TLS connections
     /// to gRPC ENS endpoint in place of the hardcoded one
     bytes? root_certificate_override;
+
+    /// Interval between the keep alive messages sent over the ENS connection, in seconds.
+    /// When the underlying tcp connection stops working, the new connection will
+    /// be re-created after at most keepalive_interval_s + keepalive_timeout_s seconds.
```
sfraczek: Same wording as in `features.rs`:

```suggestion
    /// When the underlying tcp connection stops working, the dead connection will be
    /// detected after at most keepalive_interval_s + keepalive_timeout_s seconds and
    /// a new connection will be created after the ENS backoff elapses.
```

crates/telio-proto/src/ens/ens_impl.rs
```diff
@@ -382,7 +405,29 @@ async fn create_external_channel(
...
+    let endpoint = if let Some(timeout) = keepalive.timeout {
+        endpoint.keep_alive_timeout(timeout)
+    } else {
+        endpoint
+    };
+
+    // Strictly this is not needed in our case since we have a long lived connection
+    // that we want to keep alive. This setting helps in the case where there is
+    // **no** active rpc connection and we want to make a new rpc call after a while.
+    let endpoint = if keepalive.interval.is_some() || keepalive.timeout.is_some() {
```
sfraczek: Nit: hyper ignores both `keep_alive_timeout` and `keep_alive_while_idle` when the interval is not set ([docs](https://docs.rs/hyper/1.9.0/hyper/client/conn/http2/struct.Builder.html#method.keep_alive_timeout): "Does nothing if `keep_alive_interval` is disabled"), so the `|| keepalive.timeout.is_some()` half of this condition has no effect - with only a timeout configured, no pings are sent regardless.

```suggestion
    let endpoint = if keepalive.interval.is_some() {
```

sfraczek (dismissed):
+1

crates/telio-proto/src/ens/ens_impl.rs
matislovas: What I've understood from comments in code is that:
`None` - defaults are used
`Some(0)` -  is error, but still defaults will be used.

So my proposal was to leave `None` for defaults and use `Some(0)` to disable keepalives (for testing or whatever purposes).

tomaszklak: If you omit `keepalive_interval_s`/`keepalive_timeout_s` you will get default values (`120`/`20`). If you set them explicitly to `None`/`Null` they will be `None`. When you set them to `0` it will emit a warning and default value will be used.

That's how I **think** it will behave.

matislovas:
+1

matislovas (approved):

PR #1937, opened 2026-08-07
Extend nat-lab.sh for natlab-subset sharding

stalowyjez:
### Problem
Every nat-lab shard is quite heavy running the whole nat-lab

### Solution
Let's add additional options for running nat-lab, so we can divide tests into shards which run only particular subset of all nat-lab containers, thus reducing the overall load

- [x] README.md is updated

### DIFF
--- a/nat-lab/run_local.py
+++ b/nat-lab/run_local.py
@@ -218,22 +218,23 @@ def _run_tests(args) -> None:
     original_durations_data = {}
     input_path = None
 
+    durations_path: Optional[Path] = None
+
+    if args.input_durations:
+        input_path = Path(args.input_durations)
+        original_durations_data = load_json(input_path)
+        durations_path = input_path
+    elif args.output_durations:
+        durations_path = Path(args.output_durations)
+
+    if durations_path is not None:
+        pytest_cmd.extend([
+            "--store-durations",
+            f"--durations-path={durations_path.absolute()}",
+        ])
+
     if "splits" in pytest_opts:
-        if args.input_durations:
-            input_path = Path(args.input_durations)
-            original_durations_data = load_json(input_path)
-            pytest_cmd.extend([
-                "--store-durations",
-                f"--durations-path={input_path.absolute()}",
-                "--splitting-algorithm=least_duration",
-            ])
-        elif args.output_durations:
-            output_path = Path(args.output_durations)
-            pytest_cmd.extend([
-                "--store-durations",
-                f"--durations-path={output_path.absolute()}",
-                "--splitting-algorithm=least_duration",
-            ])
+        pytest_cmd.append("--splitting-algorithm=least_duration")
 
     pytest_cmd += get_pytest_arguments(args)
 
@@ -242,7 +243,7 @@ def _run_tests(args) -> None:
 
     pytest_result = run_command(pytest_cmd, allow_failure=True)
 
-    if "splits" in pytest_opts and args.output_durations and args.input_durations:
+    if args.output_durations and args.input_durations:
         output_path = Path(args.output_durations)
         if input_path:
             merged_data = load_json(input_path)
--- a/nat-lab/run_local.py
+++ b/nat-lab/run_local.py
@@ -218,22 +218,27 @@ def _run_tests(args) -> None:
     original_durations_data = {}
     input_path = None
 
+    if args.input_durations:
+        input_path = Path(args.input_durations)
+        original_durations_data = load_json(input_path)
+        durations_path: Optional[Path] = input_path
+    elif args.output_durations:
+        durations_path = Path(args.output_durations)
+    else:
+        durations_path = None
+
+    # Record timings whenever a durations path was given, even if this run is not
+    # split. Buckets small enough to fit a single CI shard have no `parallel:` and
+    # so never see --splits; gating storage on splitting would leave them absent
+    # from compiled_test_durations.json and impossible to rebalance later.
+    if durations_path is not None:
+        pytest_cmd.extend([
+            "--store-durations",
+            f"--durations-path={durations_path.absolute()}",
+        ])
+
     if "splits" in pytest_opts:
-        if args.input_durations:
-            input_path = Path(args.input_durations)
-            original_durations_data = load_json(input_path)
-            pytest_cmd.extend([
-                "--store-durations",
-                f"--durations-path={input_path.absolute()}",
-                "--splitting-algorithm=least_duration",
-            ])
-        elif args.output_durations:
-            output_path = Path(args.output_durations)
-            pytest_cmd.extend([
-                "--store-durations",
-                f"--durations-path={output_path.absolute()}",
-                "--splitting-algorithm=least_duration",
-            ])
+        pytest_cmd.append("--splitting-algorithm=least_duration")
 
     pytest_cmd += get_pytest_arguments(args)
 
@@ -242,7 +247,7 @@ def _run_tests(args) -> None:
 
     pytest_result = run_command(pytest_cmd, allow_failure=True)
 
-    if "splits" in pytest_opts and args.output_durations and args.input_durations:
+    if args.output_durations and args.input_durations:
         output_path = Path(args.output_durations)
         if input_path:
             merged_data = load_json(input_path)
--- a/nat-lab/tests/test_ens.py
+++ b/nat-lab/tests/test_ens.py
@@ -803,7 +803,7 @@ async def test_ens_connection_error_unknown(
             setup_environment(
                 exit_stack,
                 [alpha_setup_params],
-                vpn=[ConnectionTag.VM_LINUX_NLX_1, ConnectionTag.DOCKER_VPN_1],
+                vpn=[ConnectionTag.DOCKER_VPN_1],
             )
         )
 
--- a/nat-lab/README.md
+++ b/nat-lab/README.md
@@ -113,7 +113,8 @@ uv run python3 natlab.py check-containers
 
 ### Start modifiers (skip heavy services)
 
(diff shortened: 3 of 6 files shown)


gytsto (commented):
+1

nat-lab/tests/test_nordvpnlite.py
```diff
@@ -22,6 +22,10 @@
 from tests.utils.ping import ping
 from tests.utils.process.process import ProcessExecError
 
+# CI load balancing only - see the shard_* markers in pyproject.toml. These tests need
+# nothing beyond the plain docker services.
+pytestmark = pytest.mark.shard_windows
```
tomasz-grz: Does this mean that the tests will run only on Windows container? 😕 This doesn't seem right since nordvpnlite was never intended to run on Windows

nat-lab/tests/test_upnp_connection.py
```diff
@@ -18,6 +18,10 @@
 from tests.utils.process import ProcessExecError
 from tests.utils.router import new_router, IPStack
 
+# CI load balancing only - see the shard_* markers in pyproject.toml. The windows/mac
+# params below keep their own marks, so they still run on their own shards.
+pytestmark = pytest.mark.shard_openwrt
```
tomasz-grz: I'm confused, does this includes or excludes the tests from OpenWRT?

If the former, why would we want to test UPNP only on OpenWRT?

nat-lab/tests/test_ens.py
```diff
@@ -803,7 +803,7 @@ async def test_ens_connection_error_unknown(
             setup_environment(
                 exit_stack,
                 [alpha_setup_params],
-                vpn=[ConnectionTag.VM_LINUX_NLX_1, ConnectionTag.DOCKER_VPN_1],
```
tomasz-grz, line 806: Was `VM_LINUX_NLX_1` redundant here? Or why are we reducing the test scope?

nat-lab/tests/test_dns.py
```diff
@@ -26,6 +26,10 @@
 from tests.utils.router import IPStack
 from typing import List, Optional
 
+# CI load balancing only - see the shard_* markers in pyproject.toml. These tests need
+# nothing beyond the plain docker services.
+pytestmark = pytest.mark.shard_mac
```
tomasz-grz: > These tests need nothing beyond the plain docker services.

Is this true for the macOS VM though?

Also, how does this interact with the mac shard, none of the tests below specify `ConnectionTag.VM_MAC`?

stalowyjez: The tests don't need anything more than lite-mode, but because we have a bit too much litemode tests, it makes sense to move them to other shards to balance it a bit

nat-lab/tests/test_ens.py
stalowyjez: It connects only to the one in `vpn_conf`, so `DOCKER_VPN_1`, thus NLX one seems to be redundant

nat-lab/tests/test_nordvpnlite.py
stalowyjez: Nah, it only will run on Windows shards, it will use the pure docker containers only (i.e. even if this test file contains some tests using additional containers, it is always used with `DOCKER_ONLY`, so only these using the basic containers will run)

nat-lab/tests/test_upnp_connection.py
stalowyjez: windows and mac tests are excluded from this load balancing, because we always use it with `DOCKER_ONLY` filter in CI config, so only these using basic containers are under this mark and these will be moved to openwrt shard to better balance the tests between them

nat-lab/tests/test_nordvpnlite.py
tomasz-grz: Could we document somehow more explicitly what `Windows shards` entail and how Windows shards is different from regular shard?

Reading it for the first time gives the impression that only Windows containers / VMs are started, and/or that only Windows tests are run in this shard. Does the Windows shard start all the regular natlab containers + Windows VMs?

I wonder if there is a more "elegant way" to assign the shards, personally find it puzzling at first glance to see `mark windows` inside of `nordvpnlite.py`.

nat-lab/tests/test_dns.py
tomasz-grz: I see so this just manually "assigns" the test to the shard that runs the macOS tests?

What happens if I accidentally mark `test_openwrt.py` with `shard_mac`?

nat-lab/tests/test_openwrt.py
```diff
@@ -652,21 +652,29 @@ async def test_openwrt_router_restart(
...
-                await nordvpnlite.wait_for_vpn_connected_state()
-                await check_gateway_and_client_ip(
-                    gateway_connection, client_connection, WG_SERVER["ipv4"], gw_tag
-                )
-                await exit_stack.enter_async_context(
-                    gateway_connection.create_process(["reboot"]).run()
-                )
-                await wait_until_unreachable_after_reboot(gateway_connection)
+            # TODO(LLT-7676): This restart is a workaround, shouldn't be needed when LLT-7676 is done
+            await nordvpnlite.remove_logs()
```
gytsto, line 660: why these changes in this PR?

gytsto (commented):
besides nordvpnlite changes, everything looks good

nat-lab/tests/test_openwrt.py
stalowyjez: So this is a fix for the issue which @olekoliinyk mentioned - when this test starts first, nordvpnlight is running and starting it again fails

nat-lab/tests/test_dns.py
stalowyjez: removed all of the marks

nat-lab/tests/test_upnp_connection.py
stalowyjez: removed all of the marks

nat-lab/tests/test_nordvpnlite.py
stalowyjez: removed all of the marks

nat-lab/tests/test_openwrt.py
gytsto: but is it in this PR? why not just merge it separately into main and then rebase this branch

olekoliinyk: The issue was found particularly in this branch - when `test_openwrt_router_restart` being executed in the first order in pytest sessions - so it was added here to unblock the PR and verify that changes are working.
We could have aded it to main - but it will show us nothing as this test is executed in the last order.

nat-lab/run_local.py
```diff
@@ -218,22 +218,23 @@ def _run_tests(args) -> None:
...
+    elif args.output_durations:
+        durations_path = Path(args.output_durations)
+
+    if durations_path is not None:
+        pytest_cmd.extend([
+            "--store-durations",
+            f"--durations-path={durations_path.absolute()}",
+        ])
+
     if "splits" in pytest_opts:
```
tomasz-grz, line 236: `splits=${CI_NODE_TOTAL}` was removed from `gitlab.ci`, or is there still something setting it?

nat-lab/tests/test_openwrt.py
tomasz-grz: Also not super excited about a workaround, but I like that it's marked clearly with a ticket. 👍 
Was this less of an issue with the previous implementation?

tomasz-grz (dismissed):
Nice! Looks pretty good now +1 👍 🙏

nat-lab/tests/test_openwrt.py
olekoliinyk: We haven't seen this issue before because `test_openwrt_router_restart` was executed when procd already died (crash loop is already over).

What happened here:
1. Installs nordvpnlite and starts the procd service
2. Daemon has no credentials, exits immediately, and procd respawns it every 5s
3. Test writes a valid config and auth file to the default paths the service reads
4. Next respawn finds both and starts successfully
5. In test `nordvpnlite start` fails with `DaemonIsRunning`

olekoliinyk: IMHO, so it more that we need to have a `NotConfigured` state in nordvpnlite rather than exit when there is no valid config

stalowyjez:
@tomasz-grz somehow I cannot reply directly to your comment:

>Was this less of an issue with the previous implementation?

The issue appears when the test is run first (or maybe early in the beginning of the suite), so it was always there, just with the current shuffle it become apparent

nat-lab/run_local.py
stalowyjez: removed

stalowyjez: Eh, actually we have a bit chicken and egg problem here - libtelio-build changes need this PR to run properly, so I guess I'll merge it first, then libtelio-build and in the and bump libtelio-build tag here and remove this leftovers in the same PR, so CI will be happy all the time

olekoliinyk:
@stalowyjez I think you would have to rebase from main to unblock the pipeline

nat-lab/tests/test_openwrt.py
stalowyjez: @gytsto removed that change from this PR (and rebased on the one added to main branch)

nat-lab/run_local.py
stalowyjez: So removed it here: https://github.com/NordSecurity/libtelio/pull/1986

tomasz-grz (approved):

gytsto (approved):
+1

PR #1938, opened 2026-08-07
Add performance test with enabled libfirewall

asledzinskiy:
### Problem
*--Currently all performance tests are executed with disabled libfirewall. We need to track possible performance degradations caused by libfirewall--*

### Solution
*--Add Mac OS NepTun test with enabled libfirewall as only Mac OS is currently using libfirewall--*

- [x] README.md is updated


.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v6.0.1
+TRIGGERED_REF=LLT-7543_perf_test_enabled_libfirewall
```
gytsto: update before merging

gytsto (commented):
+2

gytsto (approved):

PR #1942, opened 2026-08-11
Backport LLT-7581

tomaszklak:
Enable client side keepalives for ENS. This is a backport of #1934.

### :ballot_box_with_check: Definition of Done checklist
- [x] Commit history is clean ([requirements](../blob/main/docs/git_commit_messages_requirements.md))
- [x] README.md is updated
- [x] Functionality is covered by unit or integration tests

PR #1946, opened 2026-08-12
Backport LLT-7581: ens client side keepalives

tomaszklak:
Some of the apps are still on v6.2.x so we need to also backport the fix there. Those are the same changes as in #1942 

- [x] README.md is updated

### DIFF
--- a/crates/telio-model/src/features.rs
+++ b/crates/telio-model/src/features.rs
@@ -588,6 +588,22 @@ pub struct FeatureErrorNotificationService {
     /// DER encoded root certificate to be used for verification of all TLS connections
     /// to gRPC ENS endpoint in place of the hardcoded one
     pub root_certificate_override: Option<Vec<u8>>,
+
+    /// Interval between the keep alive messages sent over the ENS connection, in seconds.
+    /// When the underlying tcp connection stops working, the dead connection will be
+    /// detected after at most keepalive_interval_s + keepalive_timeout_s seconds and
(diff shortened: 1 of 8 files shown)


crates/telio-model/src/features.rs
```diff
@@ -588,6 +588,22 @@ pub struct FeatureErrorNotificationService {
     /// DER encoded root certificate to be used for verification of all TLS connections
     /// to gRPC ENS endpoint in place of the hardcoded one
     pub root_certificate_override: Option<Vec<u8>>,
+
+    /// Interval between the keep alive messages sent over the ENS connection, in seconds.
+    /// When the underlying tcp connection stops working, the dead connection will be
+    /// detected after at most keepalive_interval_s + keepalive_timeout_s seconds and
+    /// a new connection will be created after the ENS backoff elapses.
+    /// When set to null, the keep alives are disabled.
+    /// Setting it to 0 is an error.
```
lcruz99, line 597: as-is no error is thrown, but the default is used which can be even useful

lcruz99: https://github.com/NordSecurity/libtelio/pull/1946/changes#diff-45301ca5f5620646c57c108db449277cb2d9195510bd58929b3b5ed816234807R115

lcruz99 (approved):
+2

crates/telio-model/src/features.rs
tomaszklak: It's an error, just not a fatal one ;)

PR #1947, opened 2026-08-12
Remove second Windows VM from nat-lab

gytsto:
## Problem
`windows-client-02` (plus `windows-gw-03/04` and `windows-net-03/04`) is only used by the windows<->windows combination of `test_node_state_flickering_direct`, which is `@pytest.mark.long`. No CI job runs `long` — both `PYTEST_MARKS` and `NIGHTLY_PYTEST_MARKS` in libtelio-build exclude it. So it is a second 4G/2-core Windows VM that never runs a test.

## Solution
Remove it.

- [x] README.md is updated


asledzinskiy (approved):
+2

PR #1957, opened 2026-08-19
Temporary disable windows monitoring

olekoliinyk:
### Problem
Sometimes `VM_WINDOWS_1` vm kernel-panicks -> container exists and never restarts.
The bugcheck is `MEMORY_MANAGEMENT (0x1A)`, subcode `0x3453`, in `nt!MiDeleteFinalPageTables` - got from the `QMP GUEST_PANICKED` event and confirmed against the guest memory dump. 
Microsoft documents that subcode as "_not all the page table pages of an exited process could be deleted because of outstanding references._" 
Both times, the process being torn down was a `powershell.exe`.

Looks this is a kernel issue - no test-side change can cause it.
But the trigger appears to be the sheer volume of PowerShell process creation in the guest, and resource monitoring is the largest single contributor. 
Measured from the monitoring's own output logs: `~49 powershell.exe` creations per minute, `~600` over a VM's 15-minute lifetime.

### Solution
Disable resource monitoring while we trying to understand the root cause. 
That removes ~600 of the ~1000+ guest process create/destroy cycles per VM lifetime - the cheapest available experiment.

- [x] README.md is updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,7 +72,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


gytsto (approved):
+2

PR #1959, opened 2026-08-20
Fix WG-NT route setup racing concurrent route-table changes

gytsto:
## Problem

WG-NT adapter can come up unable to receive tunnel traffic, and the only log clue is `Unable to set IPv4 routes: 0`. Route setup races concurrent route-table writers (app routes, OS auto-routes), any hiccup aborts the whole interface config, the real error code gets lost, and the failure is swallowed silently.

## Solution

Route setup now converges instead of aborting: delete treats `ERROR_NOT_FOUND` as already gone, add overwrites an `ERROR_OBJECT_ALREADY_EXISTS` row with our parameters via `SetIpForwardEntry2` (it's always on our own adapter). Real error codes are kept and logged.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
+++ b/crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
@@ -360,6 +360,10 @@ impl InterfaceLuid {
     pub fn flush_ip_addresses(&self, address_family: ADDRESS_FAMILY) -> Result<(), NETIO_STATUS> {
         let mut p_table: PMIB_UNICASTIPADDRESS_TABLE = ptr::null_mut();
         let result = unsafe { GetUnicastIpAddressTable(address_family, &mut p_table) };
+        // ERROR_NOT_FOUND: no address entries for this family - nothing to flush
+        if ERROR_NOT_FOUND == result {
+            return Ok(());
+        }
         if NO_ERROR != result {
             return Err(result);
         }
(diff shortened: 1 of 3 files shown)


crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
```diff
@@ -360,23 +360,28 @@ impl InterfaceLuid {
     pub fn flush_ip_addresses(&self, address_family: ADDRESS_FAMILY) -> Result<(), NETIO_STATUS> {
         let mut p_table: PMIB_UNICASTIPADDRESS_TABLE = ptr::null_mut();
         let result = unsafe { GetUnicastIpAddressTable(address_family, &mut p_table) };
+        // ERROR_NOT_FOUND: no address entries for this family - nothing to flush
+        if ERROR_NOT_FOUND == result {
```
mathiaspeters: I can see this pattern in several places here, even places that you didn't edit now and I'm thinking that maybe we start changing them to be `if result == ERROR_NOT_FOUND` instead. It's functionally the same but at least personally I find it easier to read that way, and it would be more consistent with the rest of the code base

crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
```diff
@@ -697,3 +692,170 @@ impl InterfaceLuid {
...
+    let mut last_status = NO_ERROR;
+    for _ in 0..3 {
+        last_status = create();
+        if NO_ERROR == last_status {
+            return Ok(());
+        }
+        if ERROR_OBJECT_ALREADY_EXISTS != last_status {
+            return Err(last_status);
+        }
+        // ERROR_OBJECT_ALREADY_EXISTS: row raced into existence - converge its metric to ours
```
mathiaspeters: Looks to me like this comment should move up a few lines

crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
```diff
@@ -697,3 +692,170 @@ impl InterfaceLuid {
...
+        last_status = set();
+        if NO_ERROR == last_status {
+            return Ok(());
+        }
+        if ERROR_NOT_FOUND != last_status {
+            return Err(last_status);
+        }
+        // ERROR_NOT_FOUND: the row vanished between create and set - recreate
```
mathiaspeters: This comment should also move up a few lines

crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
gytsto: yeah, same for me, fixed

crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
gytsto: moved

crates/telio-wg/src/windows/tunnel/winipcfg/luid.rs
gytsto: moved

mathiaspeters:
+1 :rocket:

Paintree (approved):
+1

PR #1961, opened 2026-08-21
[LLT-6785] Report DaemonIsNotRunning instead of raw IO errors

olekoliinyk:
### Problem
`nordvpnlite is-alive` issued right after `nordvpnlite stop` fails with a raw OS error instead of `DaemonIsNotRunning`:
`Error: Io(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })
`
Before sending a command, the CLI only checks that the daemon's socket file exists, and it reports `DaemonIsNotRunning` only when that check fails. 
The daemon answers stop before it exits and removes the socket file, so a command landing in that window passes the check, connects, gets reset, and the IO error is propagated as-is.

### Solution
Report IO errors that mean nothing is listening (refused, reset, aborted, broken pipe, EOF, not found) as `DaemonIsNotRunning`, same as a missing socket file.

In nat-lab: drop the sleep, return False from is_alive() instead of re-raising. Without the sleep, quit() and kill() now poll for the process to exit and the socket to disappear.

- [x] README.md is updated

### DIFF
--- a/clis/nordvpnlite/src/main.rs
+++ b/clis/nordvpnlite/src/main.rs
@@ -3,6 +3,7 @@
 use clap::Parser;
 use daemonize::{Daemonize, Outcome};
 use std::fs::{self, OpenOptions};
+use std::io::{Error, ErrorKind};
 use tokio::time::{timeout, Duration};
 
 mod auth;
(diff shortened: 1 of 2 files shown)


clis/nordvpnlite/src/main.rs
```diff
@@ -121,7 +122,13 @@ async fn client_main(cmd: Cmd) -> Result<(), NordVpnLiteError> {
                 .await
                 .map_err(|_| NordVpnLiteError::ClientTimeoutError)?;
 
-                match CommandResponse::deserialize(&response?)? {
+                let response = match response {
+                    Ok(response) => response,
+                    Err(err) if is_daemon_gone(&err) => return handle_missing_daemon(cmd),
+                    Err(err) => return Err(err.into()),
+                };
```
djkarwowski: This block introduces three matching arms, out of which only one is significant and the other two only pass the values through, obfuscating the intent.

What we're really interested in here is checking if the daemon is gone, and we could emphasize it like this:
```
            if response.as_ref().is_err_and(|err| is_daemon_gone(&err)) {
                return handle_missing_daemon(cmd);
            }
```

The other two matching arms would be simply covered by restoring the `?` operator in the following call to `CommandResponse::deserialize`.

What do you think?

clis/nordvpnlite/src/main.rs
djkarwowski: With introduction of `handle_missing_daemon()` maybe it's worth to consider inverting the `if` like this:
```
            if !socket_path.exists() {
                return handle_missing_daemon(cmd);
            }
```

This would help reduce nesting and together with the [other suggestion](https://github.com/NordSecurity/libtelio/pull/1961/changes#r3879832526) would make the overall control flow in the function more straightforward (which I personally think is much easier to comprehend).

Feel free to skip it if you disagree :)

clis/nordvpnlite/src/main.rs
```diff
@@ -349,6 +375,60 @@ mod tests {
...
+        ] {
+            assert!(!is_daemon_gone(&Error::from(kind)), "{kind:?}");
+        }
+    }
+
+    #[test]
+    fn test_missing_daemon_reports_daemon_is_not_running() {
+        // Stopping an already stopped daemon is not an error, everything else is.
+        assert_matches!(handle_missing_daemon(ClientCmd::QuitDaemon), Ok(()));
+        for cmd in [ClientCmd::IsAlive, ClientCmd::GetStatus, ClientCmd::Reload] {
```
djkarwowski: If a new value is ever added to the `ClientCmd` enum in the future, this test will still compile and pass, putting it at risk of getting somehow stale.

Maybe it's worth making it more future-proof by ensuring the compiler complains about the non-exhaustive pattern in the test as well. What do you think?

djkarwowski (commented):
I've left few suggestions for consideration, but overall looks good to me.

+1.0

clis/nordvpnlite/src/main.rs
```diff
@@ -166,6 +165,30 @@ async fn client_main(cmd: Cmd) -> Result<(), NordVpnLiteError> {
...
+/// Whether an IO error means nothing is listening on the socket anymore, either because
+/// the socket file is stale or because the daemon exited while the command was in flight.
+fn is_daemon_gone(err: &Error) -> bool {
+    matches!(
+        err.kind(),
+        ErrorKind::NotFound
+            | ErrorKind::ConnectionRefused
+            | ErrorKind::BrokenPipe
+            | ErrorKind::ConnectionReset
+            | ErrorKind::UnexpectedEof
```
matislovas, line 178: Shouldn't it include `DaemonIsNotRunning` as well?

nat-lab/tests/nordvpnlite.py
matislovas, line 413:
```  
+        daemon_already_stopped = "Daemon is already stopped" in stdout
           assert (
               "Command executed successfully" in stdout
  -            or "Daemon is already stopped" in stdout
  +            or daemon_already_stopped
           ), f"Failed to execute stop command: {stderr}"

  +        if daemon_already_stopped:
  +            if await self.socket_exists():
  +                log.debug("Dangling socket found after stop no-op, removing it..")
  +                await self.remove_socket()
  +            return
```

Maybe this way to remove dangling socket?  `stop` can report success for a stale socket and there's already no daemon to remove it.

clis/nordvpnlite/src/main.rs
olekoliinyk: That list is `std::io::ErrorKind`, `DaemonIsNotRunning` is a `NordVpnLiteError`  which is different types, and I think it won't compile

nat-lab/tests/nordvpnlite.py
olekoliinyk: Done

clis/nordvpnlite/src/main.rs
matislovas: Oh shoot, yeah, I was referring to `ConnectionAborted`, shouldn't it be there?

olekoliinyk: Right, should be...Added

matislovas:
+1

matislovas (approved):

PR #1962, opened 2026-08-21
Add wireguard-nt adapter creation retry

matislovas:
### Problem
Windows had problems running telio-related tests in their CI, where VM is not shutdown after tests. This resulted in frequent 'Failed to create adapter ... File not found' errors.

### Solution
Introduce some retries on adapter creation with some GUIDs mixing. Tried on windows-app CI in long running tests with alwasy-on VM. Both - recovery with different GUID and cleanup works.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/adapter/windows_native_wg.rs
+++ b/crates/telio-wg/src/adapter/windows_native_wg.rs
@@ -39,6 +39,9 @@ use wireguard_nt::{
 use wireguard_uapi::xplatform;
 
 const REMOVAL_SLEEP_SECS: u64 = 2;
+const CREATE_ADAPTER_MAX_ATTEMPTS: usize = 10;
+const CREATE_ADAPTER_INITIAL_BACKOFF_SECS: u64 = 2;
+const CREATE_ADAPTER_MAX_BACKOFF_SECS: u64 = 14;
 
 /// Telio wrapper around wireguard-nt
 pub struct WindowsNativeWg {
@@ -338,13 +341,33 @@ impl WindowsNativeWg {
         Self::cleanup_orphaned_devices().await;
 
         let dll_path = "wireguard.dll";
-        let tmp_wg_dev = Self::create(name, dll_path, enable_dynamic_wg_nt_control.clone());
+        let mut backoff_secs = CREATE_ADAPTER_INITIAL_BACKOFF_SECS;
+        let mut attempt = 1;
+        let wg_dev = loop {
+            let tmp_wg_dev = Self::create(name, dll_path, enable_dynamic_wg_nt_control.clone());
 
-        telio_log_debug!("Print registry after adapter creation!");
-        Self::print_registry_key_contents(HKEY_LOCAL_MACHINE, service::GUID_DEVINTERFACE_NET_STR);
-        Self::print_registry_key_contents(HKEY_LOCAL_MACHINE, SWD_WIREGUARD);
+            telio_log_debug!("Print registry after adapter creation!");
+            Self::print_registry_key_contents(
+                HKEY_LOCAL_MACHINE,
+                service::GUID_DEVINTERFACE_NET_STR,
+            );
+            Self::print_registry_key_contents(HKEY_LOCAL_MACHINE, SWD_WIREGUARD);
 
-        let wg_dev = tmp_wg_dev?;
+            match tmp_wg_dev {
+                Ok(wg_dev) => break wg_dev,
+                Err(err) if attempt < CREATE_ADAPTER_MAX_ATTEMPTS => {
+                    telio_log_warn!(
+                        "Failed to create adapter '{}' with err: {err}. Attempt {attempt}/{}. Retrying in {backoff_secs}s",
+                        name,
+                        CREATE_ADAPTER_MAX_ATTEMPTS
+                    );
+                    sleep(Duration::from_secs(backoff_secs)).await;
+                    backoff_secs = (backoff_secs * 2).min(CREATE_ADAPTER_MAX_BACKOFF_SECS);
+                    attempt += 1;
+                }
+                Err(err) => return Err(err),
+            }
+        };
         telio_log_info!(
             "Adapter '{}' using created successfully. enable_dynamic_wg_nt_control: {}",
             name,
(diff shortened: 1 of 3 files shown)


crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -335,16 +430,46 @@ impl WindowsNativeWg {
 
         // Cleaning orphaned WireGuard-NT adapters inside the driver can be racey,
         // so we do it in a separate step, before creating a new adapter
-        Self::cleanup_orphaned_devices().await;
+        Self::cleanup_orphaned_devices(name).await;
```
gytsto, line 430: initially we decided to clean all devices just because wgnt would queue its own cleanup and that would race with creation of new adapter. So right now, when we clean only specific adapters, aren't  we getting back to that race we wanted to workaround with initial implementation? I don't really remember how wgnt did that cleanup, was it wiping like all orphaned devices or specific ones too.

Anyway, could you confirm or just check that race wouldn't happen?

matislovas: This `name` is used only for subsequent `cleanup_pool_guid_netcfg` where we need GUIDs of all possible adapters.  The `cleanup_orphaned_devices` is doing the same job that it was doing previously - taking over cleanup of orphaned adapters from `wireguard-nt`.

And the `cleanup_pool_guid_netcfg` is used to avoid duplicate adapters leftovers with the same name.

matislovas: Maybe I'll rename the function argument, for it not get mixed up with the one inside the function (even though, it is not affecting behavior).

gytsto: Nah, don't do anything :smile: it's just me being blind, somehow in my mind I tought you changed original function. All good here then :smile:

gytsto (commented):
+1

lcruz99:
Is there a plan to tackle the Windows VM not shutting down issue? Was the root-cause found? Is it related with libtelio?

matislovas:
> Is there a plan to tackle the Windows VM not shutting down issue? Was the root-cause found? Is it related with libtelio?

They're not shutting down it, because of lack of resources for CI. And the real root cause is not found. There are sometimes complaints from DCO driver failing to be installed the same way as Wireguard. IMHO it (the Windows) lacks some global mutex in NetworkSetup2 or in PnP. Wireguard-NT has its own global mutex for adapter operations, but it is in driver-scope, not in some more global scope.

So short story - no, libtelio is the victim here :smile:

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -102,19 +117,37 @@ impl WindowsNativeWg {
...
     fn create(
         name: &str,
+        adapter_guid: u128,
         path: &str,
         enable_dynamic_wg_nt_control: IsMeshnetEnabledCb,
     ) -> std::result::Result<Self, AdapterError> {
         // try to load dll
         match unsafe { wireguard_nt::load_from_path(path) } {
             Ok(wg_dll) => {
+                wireguard_nt::set_logger(&wg_dll, Some(wireguard_nt::default_logger));
```
lcruz99: nit: Remove `wireguard_nt` qualifier, it's being imported already

lcruz99: Isn't set_logger() being called on every loop iteration?

lcruz99: https://docs.rs/wireguard-nt/latest/src/wireguard_nt/log.rs.html#8-10 -> looks like there's no harm but still, isn't necessary

matislovas: Fixed

matislovas: > Isn't set_logger() being called on every loop iteration?

Nope, I tried it, and after failed attempt, we lose the logs, at least for some time :smile:

matislovas: They actually trying to set it, but singleton `SET_LOGGER` is not unset after failed attempt.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -102,19 +118,37 @@ impl WindowsNativeWg {
         guid
     }
 
+    /// Render a GUID the way Windows does in `SWD\WireGuard\{...}`.
+    ///
```
lcruz99: pls remove LLM generated comments

lcruz99: IMHO the code is self-explanatory and doesn't need this explanation

matislovas: Removed the following, but left some others, because they're actually useful IMHO.

lcruz99:
Please add nat-lab tests

matislovas:
> Please add nat-lab tests

That's a tricky part, we do not have reliable repro for this bug. So I cannot re-create the conditions, where it gets the error first time on adapter creation and succeeds the second time with different GUID.

lcruz99:
> That's a tricky part, we do not have reliable repro for this bug. So I cannot re-create the conditions, where it gets the error first time on adapter creation and succeeds the second time with different GUID.

Can't we use ` nat-lab/tests/test_wg_adapter.py::test_wg_adapter_cleanup` as a template? Even if we won't be able to reproduce the bug, this feature should be tested.

matislovas:
> > That's a tricky part, we do not have reliable repro for this bug. So I cannot re-create the conditions, where it gets the error first time on adapter creation and succeeds the second time with different GUID.
> 
> Can't we use ` nat-lab/tests/test_wg_adapter.py::test_wg_adapter_cleanup` as a template? Even if we won't be able to reproduce the bug, this feature should be tested.

The one in `test_wg_adapter_cleanup` actually tests the adapter cleanup on startup. For this PR to be tested, we should:
1. Receive an error from `wireguard-nt`, while creating adapter.
2. Than we would try different GUID.
3. Re-launch telio and see that the adapter from point 2 would be cleaned up.

But we do not know how to receive any error from windows stack, while creating wg adapter :smile:

lcruz99:
> Receive an error from wireguard-nt, while creating adapter.

Perhaps just making the wireguard.dll binary inaccessible?

tomaszklak:
> > Receive an error from wireguard-nt, while creating adapter.
> 
> Perhaps just making the wireguard.dll binary inaccessible?

~~Would that trigger the code in question? It would fail very early in `wireguard_nt::load_from_path(path)` without reaching `match wireguard_nt::Adapter::create(&wg_dll, name, name, Some(adapter_guid))`. So I'm not sure how useful that would be.~~

lcruz99 (approved):
+1

PR #1965, opened 2026-08-24
LLT-7669: Telio DNS cleanup

tomasz-grz:
### Problem
Tidying up in preparation for adding TCP Forwarder.

### Solution
- Rename UdpForwarder module file
- Define DNS_PORT constants
- Define DNS addresses constants
- Remove unused timeout
- Breakout upstream handling

- [x] README.md is updated


crates/telio-dns/src/dns.rs
```diff
@@ -38,10 +41,10 @@ pub trait DnsResolver {
     fn get_peer(&self, allowed_ips: Vec<IpNet>) -> Peer;
     /// Get default allowed IPs of this DNS server.
     fn get_default_dns_allowed_ips(&self) -> Vec<IpNet>;
-    /// Get DNS virtual peer addresses.
+    /// Get allowed IPs when connected to exit node.
```
lcruz99, line 47: 100.64.0.3?

lcruz99: isn't it the virtual peer address?

crates/telio-dns/src/dns.rs
```diff
@@ -156,25 +159,22 @@ impl DnsResolver for LocalDnsResolver {
 
     fn get_default_dns_allowed_ips(&self) -> Vec<IpNet> {
         vec![
-            IpAddr::V4(Ipv4Addr::new(100, 64, 0, 2)).into(),
-            IpAddr::V4(Ipv4Addr::new(100, 64, 0, 3)).into(),
-            IpAddr::V6(Ipv6Addr::new(0xfd74, 0x656c, 0x696f, 0, 0, 0, 0, 2)).into(),
-            IpAddr::V6(Ipv6Addr::new(0xfd74, 0x656c, 0x696f, 0, 0, 0, 0, 3)).into(),
+            IpAddr::V4(DNS_VIRTUAL_IPV4).into(),
+            IpAddr::V4(DNS_EXIT_IPV4).into(),
```
lcruz99: this is virtual DNS peer, not actual remote DNS IPV4 (I think the virtual keyword gives meaning and non-trivial)

crates/telio-dns/src/dns.rs
tomasz-grz: The local virtual peer address is `100.64.0.2`.

But when connected to a Meshnet exit node, we route `100.64.0.3` to the virtual peer of the exit node, not our local one.
So in that case the allowed IPs should only be `100.64.0.2`

I found the original comments and function name rather confusing, so I tried to "improve them" 😅 You be the judge of how effective that was.

crates/telio-dns/src/dns.rs
tomasz-grz: They are two virtual peers as documented by **_Reserved meshnet IP addresses_** on Confluence.
The local one `100.64.0.2`,`fd74:656c:696f::2` and the remote one when connected to exit node `100.64.0.3`, `fd74:656c:696f::3`.

Would you prefer to see something like `DNS_VIRTUAL_PEER_IPV4` and `DNS_VIRTUAL_PEER_ON_EXIT_IPV4`?

crates/telio-dns/src/dns.rs
tomasz-grz: So the original allowed ip functions were:

```rust
/// Get default allowed IPs of this DNS server.
get_default_dns_allowed_ips() -> [100.64.0.2, 100.64.0.3]
/// Get DNS virtual peer addresses.
get_exit_connected_dns_allowed_ips() -> [100.64.0.2]
```

I found this comment not reflecting properly when this is used and the function name.

tomasz-grz: And the original signature of
```rust
/// Get default DNS server IP addresses.
get_default_dns_servers() -> [100.64.0.3]
```

But it was only used when `is_meshnet_exit_node == true`:
```rust
if is_meshnet_exit_node {
}
```

Therefore I thought `get_exit_node_dns_servers()` would be more suitable name for it.

Paintree (dismissed):
+1

crates/telio-dns/src/dns.rs
lcruz99: or `DNS_VIRTUAL_WHEN_EXIT_IPV4` 😁 as you prefer.. My comment was just about omitting the "VIRTUAL" word.

tomasz-grz: Updated

crates/telio-dns/src/dns.rs
lcruz99: I see some redudancy here. Why not having split functions for virtual peer when meshnet and exit-node?

lcruz99: > get_exit_connected_dns_allowed_ips() -> [100.64.0.2]  

should be `get_exit_connected_dns_allowed_ips() -> [100.64.0.3]` no?

crates/telio-dns/src/dns.rs
```diff
@@ -38,10 +44,10 @@ pub trait DnsResolver {
     fn get_peer(&self, allowed_ips: Vec<IpNet>) -> Peer;
     /// Get default allowed IPs of this DNS server.
     fn get_default_dns_allowed_ips(&self) -> Vec<IpNet>;
-    /// Get DNS virtual peer addresses.
+    /// Get allowed IPs when connected to exit node.
     fn get_exit_connected_dns_allowed_ips(&self) -> Vec<IpNet>;
-    /// Get default DNS server IP addresses.
-    fn get_default_dns_servers(&self) -> Vec<IpAddr>;
+    /// Get DNS server IP addresses when connected to exit node.
+    fn get_exit_node_dns_servers(&self) -> Vec<IpAddr>;
```
lcruz99: maybe add `remote` keyword to the signature?

crates/telio-dns/src/lib.rs
```diff
@@ -35,3 +34,6 @@ pub mod fuzz {
     pub use super::packet_decoder::{find_nord_query, parse_dns_query_packet};
     pub use super::packet_encoder::fuzz_build_response;
 }
+
+/// DNS port number
+pub const DNS_PORT: u16 = 53;
```
lcruz99: shouldn't this belong to constants.rs?

tomasz-grz: Possibly, `telio-firewall` can use it, though I would rather avoid updating firewall in this PR

lcruz99: I didn't understand, how updating firewall would be necessary?

crates/telio-dns/src/dns.rs
tomasz-grz: I think we **don't** want to allow `100.64.0.3`, so that it goes through the tunnel to the exit node, instead of being processed by the local virtual peer.

crates/telio-dns/src/lib.rs
tomasz-grz: Well the point of putting it in `constants.rs` is for the constant to be reused by all the modules, instead of defining it locally, no?

crates/telio-dns/src/dns.rs
tomasz-grz: Added

crates/telio-dns/src/lib.rs
tomasz-grz: Moved to constants.rs

crates/telio-dns/src/dns.rs
lcruz99: I'm not refuting that, just that I'm not understanding why  `get_exit_connected_dns_allowed_ips()` is returning  `DNS_VIRTUAL_PEER_IPV4`  instead of `DNS_VIRTUAL_PEER_ON_EXIT_IPV4`.

In https://github.com/NordSecurity/libtelio/blob/9ddeafeaca50436a2abfa5c1b92c360251a4d3ca/crates/telio-core/src/device/wg_controller.rs#L772, shouldn't be:

```
        let dns_allowed_ips = if exit_node_exists {
            // resolver.get_exit_connected_dns_allowed_ips():
             vec![
                 IpAddr::V4(DNS_VIRTUAL_PEER_ON_EXIT_IPV4).into(),
                 IpAddr::V6(DNS_VIRTUAL_PEER_ON_EXIT_IPV6).into(),
             ]
        } else {
            // resolver.get_default_dns_allowed_ips() ->
             vec![
                 IpAddr::V4(DNS_VIRTUAL_PEER_IPV4).into(),
                 IpAddr::V6(DNS_VIRTUAL_PEER_IPV6).into(),
             ]
        };
```
?

tomasz-grz: The way I understand this, OS resolver is always set to `100.64.0.2` so we don't change that.

When we connect to Meshnet exit node: we set our Magic DNS forwarder upstream address to `100.64.0.3`, and remove `100.64.0.3` from our own allowed IPs.

When you make a DNS request, the OS is sending the request to `100.64.0.2`, Magic DNS is catching it, (if it's a local `.nord` requests it resolves it) and forwards it to `100.64.0.3`. Which get written back into the tunnel routed to the remote exit node.

If we keep `100.64.0.3` as allowed IPs, then then the requests forwarder to `100.64.0.3` would be caught again  our local magic DNS making a feedback loop.

On the remote peer, (the default meshnet settings) we want to keep both `100.64.0.2` and `100.64.0.3` in allowed IPs. To resolve both local requests addressed to `100.64.0.2`, and the requests arriving from the other peer addressed to `100.64.0.3`.

I didn't change any of the addresses, just gave them names to make it a little bit "less confusing"..

lcruz99 (approved):
+1

PR #1966, opened 2026-08-25
Retry wg-nt adapter state changes on device interface arrival

gytsto:
### Problem
Wg-nt adapter start intermittently fails on Windows with `Failed to set adapter's state to Up ... Os { code: 433 ... "A device which does not exist was specified." }`. Windows can transiently restart the device right after creation, and the LLT-5443 retry loop skips its sleep when the state read fails — so all 5 attempts burn out in milliseconds instead of riding the restart out.

### Solution
Sleep before every retry with exponential backoff (10 attempts, 200ms doubling, capped at 2s), and wake retries early via `CM_Register_Notification` device interface arrival events, falling back to plain sleeps when registration fails. The retry loop is extracted into a closure-driven function covered by unit tests.

- [x] README.md is updated

### DIFF
--- a/crates/telio-wg/src/adapter/windows_native_wg.rs
+++ b/crates/telio-wg/src/adapter/windows_native_wg.rs
@@ -63,9 +63,12 @@ const SET_STATE_MAX_BACKOFF: Duration = Duration::from_secs(2);
 struct DeviceInterfaceArrivalWatcher {
     handle: HCMNOTIFICATION,
     arrival: Arc<Notify>,
+    // The callback context's own reference into `arrival`, released in Drop
+    context: *const Notify,
 }
 
-// SAFETY: `handle` is only used by `CM_Unregister_Notification` in `Drop`, which is thread-safe
+// SAFETY: `handle` is only used by `CM_Unregister_Notification` in `Drop`, which is thread-safe,
+// and `context` only by the callback and Drop
 unsafe impl Send for DeviceInterfaceArrivalWatcher {}
 unsafe impl Sync for DeviceInterfaceArrivalWatcher {}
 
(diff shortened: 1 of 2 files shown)


crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -344,7 +460,17 @@ impl WindowsNativeWg {
...
+        wg_dev.iface_arrival = match DeviceInterfaceArrivalWatcher::register() {
+            Ok(watcher) => Some(watcher),
+            Err(error) => {
+                telio_log_warn!(
+                    "Failed to register device interface arrival notification: config manager error {}",
+                    error.0
+                );
+                None
+            }
+        };
```
mathiaspeters: I don't really have any strong feelings, just suggesting a different way of writing this. Go with whatever seems the most clear to you

```suggestion
        wg_dev.iface_arrival = DeviceInterfaceArrivalWatcher::register()
            .inspect_err(|error|telio_log_warn!(
                    "Failed to register device interface arrival notification: config manager error {}", error.0
            ))
            .ok();
```

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -39,6 +51,106 @@ use wireguard_nt::{
...
+impl Drop for DeviceInterfaceArrivalWatcher {
+    fn drop(&mut self) {
+        let result = unsafe { CM_Unregister_Notification(self.handle) };
+        if result != CR_SUCCESS {
+            telio_log_warn!(
+                "Failed to unregister device interface arrival notification: config manager error {}",
+                result.0
+            );
+            // The callback may still fire with a pointer into this Arc - leak it
+            std::mem::forget(self.arrival.clone());
```
mathiaspeters: I understand why it was solved this way, but I don't like this solution. Does the windows API give any sort of upper bound on when the unregister is actually done so that we can properly clean up?

Paintree (dismissed):
+1

crates/telio-wg/src/adapter/windows_native_wg.rs
gytsto: `CM_Unregister_Notification` is itself the completion barrier - it blocks until in-flight callbacks return. A failed `unregister` has no completion signal at all, so leaking the one context reference is the only option we have. The rework just makes that ownership explicit.

crates/telio-wg/src/adapter/windows_native_wg.rs
gytsto: Applied

crates/telio-wg/src/adapter/windows_native_wg.rs
mathiaspeters: What a nice API :sweat_smile:

mathiaspeters (approved):
+1

PR #1968, opened 2026-08-25
LLT-7587: natlab: wait for starting containers before releasing control

gytsto:
### Problem
A container crashing during `compose up --wait` aborts the whole wait early. `manage_containers()` recovery only restarted missing/unhealthy services and treated health `starting` as fine, so tests started while slow-boot VMs were still booting. Setup checks then hit a half-booted VM (SSH connection refused) and killed the session (libtelio-build job 38345856 on v6.1.4: openwrt-gw-03 crashed once, Windows VM got ~2.5 min instead of ~10).

### Solution
`manage_containers()` now also tracks `starting` services and polls until they resolve to healthy/unhealthy, with a 1200s backstop (every healthcheck resolves `starting` within ~300s by config). `starting` flipping to unhealthy still goes through the existing restart path.

- [x] README.md is updated

### DIFF
--- a/nat-lab/natlab.py
+++ b/nat-lab/natlab.py
@@ -7,6 +7,7 @@ import os
 import re
 import subprocess
 import sys
+import time
 from typing import List, Tuple
 
 # isort: off
@@ -15,6 +16,7 @@ sys.path += [f"{PROJECT_ROOT}/ci"]
 from env import LIBTELIO_ENV_NAT_LAB_DEPS_TAG  # type: ignore # pylint: disable=import-error, wrong-import-position
 
 NATLAB_CONTAINER_RESTART_ATTEMPTS = 5
+NATLAB_CONTAINER_HEALTHY_WAIT_SECONDS = 1200
 
 DIRECT_ROUTING_OVERRIDE_FILE = "docker-compose.override.yml"
 DIRECT_ROUTING_OVERRIDE = """\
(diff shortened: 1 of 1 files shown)


nat-lab/natlab.py
```diff
@@ -15,6 +16,8 @@
 from env import LIBTELIO_ENV_NAT_LAB_DEPS_TAG  # type: ignore # pylint: disable=import-error, wrong-import-position
 
 NATLAB_CONTAINER_RESTART_ATTEMPTS = 5
+# Backstop only: every healthcheck resolves "starting" within its own budget (<=300s).
+NATLAB_CONTAINER_HEALTHY_WAIT_SECONDS = 600
```
tomaszklak: I think that `android-client-01` can exceed that.

nat-lab/natlab.py
```diff
@@ -379,42 +382,66 @@ def _report_container_failures(
...
     while restart_attempts < NATLAB_CONTAINER_RESTART_ATTEMPTS:
-        missing, unhealthy = find_failing_containers(services_to_start)
+        missing, unhealthy, starting = find_failing_containers(services_to_start)
         failed = missing + [s for s in unhealthy if s not in missing]
-        if not failed:
+        if failed:
+            print("Missing services: ", missing)
+            print("Unhealthy services: ", unhealthy)
+            quick_restart_container(
+                failed, env={"COMPOSE_DOCKER_CLI_BUILD": "1", "DOCKER_BUILDKIT": "1"}
```
tomaszklak, line 428: nit: maybe it would be useful to `dump_docker_logs` for services from `failed`?

nat-lab/natlab.py
gytsto: good point, missed that

nat-lab/natlab.py
gytsto: added

tomaszklak (commented):
+1

nat-lab/natlab.py
tomaszklak: Please also update PR description.

Paintree (approved):
+1

nat-lab/natlab.py
gytsto: updated

PR #1971, opened 2026-08-26
Register nat-lab log collector before client start steps

gytsto:
### Problem
A client that fails during startup leaves no telio, system event, or setupapi logs in the test artifacts — `LogCollector` is registered only after the adapter is up, so exactly the failures that need device-side logs collect none.

### Solution
Register the collector right after the remote process is ready, before the start steps. Log fetching already tolerates a missing log file, and `LOG_COLLECTORS` is cleared per test, so nothing double-collects.

- [x] README.md is updated


matislovas (approved):
+2

PR #1976, opened 2026-09-04
Add perf test on android with libfirewall

asledzinskiy:
### Problem
Android platform is one of the main users of libfirewall. After enabling it android team started seeing some performance degradations but there was no proof it was caused by libfirewall. We want to make sure that performance isn't affected by enabled libfirewall

### Solution
Download libfirewall android artifact, upload to nat-lab android VM and measure performance with iperf

- [x] README.md is updated

### DIFF
--- a/nat-lab/bin/android/natlab-python
+++ b/nat-lab/bin/android/natlab-python
@@ -11,7 +11,7 @@ PREFIX=/data/data/com.termux/files/usr
 export PREFIX
 export HOME=/data/data/com.termux/files/home
 export PATH="$PREFIX/bin:/system/bin"
-export LD_LIBRARY_PATH="$PREFIX/lib"
+export LD_LIBRARY_PATH="$HOME/work:$PREFIX/lib"
 export LD_PRELOAD="$PREFIX/lib/libtermux-exec.so"
 export TMPDIR="$PREFIX/tmp"
 cd "$HOME/work"
--- a/nat-lab/performance_tests/conftest.py
+++ b/nat-lab/performance_tests/conftest.py
@@ -12,6 +12,7 @@ from tests.utils.connection import ConnectionTag, clear_ephemeral_setups_set
 from tests.utils.logger import log
 from tests.utils.router import IPStack
 from tests.utils.testing import get_current_test_log_path
+from tests.utils.vm import android_vm_util
 
 
 def _cancel_all_tasks(loop: asyncio.AbstractEventLoop):
@@ -154,6 +155,27 @@ async def collect_kernel_logs(items, suffix):
     os.makedirs(log_dir, exist_ok=True)
     save_dmesg_from_host(suffix)
     save_audit_log_from_host(suffix)
+
+
+def _skip_android_libfirewall_tests(items):
+    if os.path.exists(android_vm_util.LIBFIREWALL_SO):
+        return
+    for item in items:
+        if item.get_closest_marker("android") and item.get_closest_marker(
+            "libfirewall"
+        ):
+            item.add_marker(
+                pytest.mark.skip(
+                    reason=(
+                        "libfirewall not available at"
+                        f" {android_vm_util.LIBFIREWALL_SO}"
+                    )
+                )
+            )
+
+
+def pytest_collection_modifyitems(items):
+    _skip_android_libfirewall_tests(items)
 
 
 def pytest_runtestloop(session):
--- a/nat-lab/tests/utils/vm/android_vm_util.py
+++ b/nat-lab/tests/utils/vm/android_vm_util.py
@@ -1,6 +1,7 @@
 import os
-from tests.config import ANDROID_DEVICE_TMP, UNIFFI_PATH_VM_ANDROID
+from tests.config import ANDROID_DEVICE_TMP, get_root_path, UNIFFI_PATH_VM_ANDROID
 from tests.utils.connection.adb_connection import AdbConnection
+from tests.utils.logger import log
 
 # Where the libtelio runtime files land on the guest (inside Termux).
 WORK_DIR = UNIFFI_PATH_VM_ANDROID.rstrip("/")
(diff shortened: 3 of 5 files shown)


lcruz99:
nit: PR description `--` is not recognized and "We need to test performance on android with enabled libfirewall" is not a _problem_.

nat-lab/bin/android/natlab-python
```diff
@@ -11,7 +11,7 @@ PREFIX=/data/data/com.termux/files/usr
 export PREFIX
 export HOME=/data/data/com.termux/files/home
 export PATH="$PREFIX/bin:/system/bin"
-export LD_LIBRARY_PATH="$PREFIX/lib"
+export LD_LIBRARY_PATH="$HOME/work:$PREFIX/lib"
```
lcruz99, line 14: what's the reason for this change?

nat-lab/performance_tests/conftest.py
```diff
@@ -156,6 +157,23 @@ async def collect_kernel_logs(items, suffix):
...
+            "libfirewall"
+        ):
+            item.add_marker(
+                pytest.mark.skip(
+                    reason=(
+                        "libfirewall not available at"
+                        f" {android_vm_util.LIBFIREWALL_SO}"
+                    )
+                )
+            )
```
lcruz99, line 174: for future proofing in case anyone wants to extend  `pytest_collection_modifyitems`, move this code into its own function.

nat-lab/tests/utils/vm/android_vm_util.py
```diff
@@ -34,6 +38,14 @@ async def copy_binaries(connection: AdbConnection) -> None:
         (f"{_UNIFFI}/serialization.py", "serialization.py"),
     ]
 
+    if os.path.exists(LIBFIREWALL_SO):
+        runtime_files.append((f"{_DIST_ANDROID}/libfirewall.so", "libfirewall.so"))
```
lcruz99, line 39: Some level of redundancy with `LIBFIREWALL_SO` and `_DIST`. Use `LIBFIREWALL_SO` here or drop it and `_DIST` everywhere.

lcruz99: I would keep it concise by using `_DIST`, just like `libtelio.so`.

nat-lab/tests/utils/vm/android_vm_util.py
```diff
@@ -34,6 +38,14 @@ async def copy_binaries(connection: AdbConnection) -> None:
         (f"{_UNIFFI}/serialization.py", "serialization.py"),
     ]
 
+    if os.path.exists(LIBFIREWALL_SO):
```
lcruz99, line 38: Any reason for having this check for `LIBFIREWALL_SO` and not for other binaries? Perhaps this check isn't necessary.

nat-lab/bin/android/natlab-python
asledzinskiy: `libtelio` loads `libfirewall` by name only (not by full path) so Android is looking in `LD_LIBRARY_PATH` path. 
As we copy `libfirewall` to `UNIFFI_PATH_VM_ANDROID = /data/data/com.termux/files/home/work/` - we need to add `$HOME/work` to `LD_LIBRARY_PATH`

nat-lab/tests/utils/vm/android_vm_util.py
asledzinskiy: `libfirewall` is pulled from separate repo instead of being built in the pipeline. But we check the same in `performance_tests/conftest.py` so this check can be dropped

.env
```diff
@@ -1 +1 @@
-TRIGGERED_REF=v6.2.1
+TRIGGERED_REF=LLT-7674_android_libfirewall_perf_test
```
lcruz99: reminder to change this before merge

nat-lab/performance_tests/conftest.py
asledzinskiy: Done

nat-lab/tests/utils/vm/android_vm_util.py
asledzinskiy: I only refactored common path into one variable:
```
DIST_SUBPATH = f"dist/android/{os.getenv('TELIO_BIN_PROFILE', 'release')}/x86_64"
_DIST_ANDROID = f"{_MOUNT}/{_DIST_SUBPATH}"
LIBFIREWALL_SO = get_root_path(f"{_DIST_SUBPATH}/libfirewall.so")
```

Couldn't only use `_DIST_ANDROID` as it is container mounted path but we check `libfirewall.so` on the host with `os.path.exists`

.env
asledzinskiy: done

nat-lab/tests/utils/vm/android_vm_util.py
asledzinskiy: I had to keep this check because the same code path with copying binaries will be executed for `android_neptun` perf test which doesn't require `libfirewall`

lcruz99: shouldn't the `libfirewall` marker be used instead for that effect?

lcruz99: This comments also scopes to other places: `libfirewall` marker should be used to check which tests should be run. The existence of `libfirewall.so` should then be asserted and aligned with the pytest mark, i.e.: If the mark is present but the binary isn't, then pytest should exit with an error.

asledzinskiy: What you describe is already asserted in `_skip_android_libfirewall_tests` but we have to check `libfirewall.so` existence in `android_vm_util.py` as well because it is shared module used by all android tests.

The same approach is used by Mac tests (but `try/except` instead of checking file existence) - https://github.com/NordSecurity/libtelio/blob/main/nat-lab/tests/utils/vm/mac_vm_util.py#L74

Paintree (dismissed):
+1

lcruz99:
+1

lcruz99 (approved):

PR #1978, opened 2026-09-07
LLT-7068: Adapter MTU API

stalowyjez:
### Problem
Currently there is no way to set MTU for any adapter while it can be added and it would be useful for some telio users

### Solution
Two ways of configuring MTU are added: adding in during adapter startup and `set_mtu` usable on running device. Because there are already many adapter starting functions I think instead of adding to each of them another version with mtu argument I think it makes more sense to add a config with optional fields and MTU among them. The solution works currently only with WindowsNT adapter, returns an error when used with other one.

- [x] README.md is updated

### DIFF
--- a/crates/telio-traversal/src/endpoint_providers/upnp.rs
+++ b/crates/telio-traversal/src/endpoint_providers/upnp.rs
@@ -869,6 +869,7 @@ mod tests {
             async fn stop(self);
             async fn reset_existing_connections(&self, exit_pubkey: PublicKey) -> Result1<()>;
             async fn set_ip_stack(&self, ip_stack: Option<IpStack>) -> Result1<()>;
+            async fn set_adapter_mtu(&self, mtu: u32) -> Result1<()>;
         }
     }
 
--- a/crates/telio-traversal/src/endpoint_providers/upnp.rs
+++ b/crates/telio-traversal/src/endpoint_providers/upnp.rs
@@ -818,7 +818,6 @@ impl<Wg: WireGuard, I: UpnpEpCommands, E: Backoff> Runtime for State<Wg, I, E> {
 mod tests {
     use super::{
         EPHEMERAL_PORT_RANGE, EndpointCandidate, MockUpnpEpCommands, UpnpEndpointProvider,
-        async_trait,
     };
 
     use std::{
@@ -832,46 +831,21 @@ mod tests {
     use crate::endpoint_providers::Error;
     use crate::ping_pong_handler::PingPongHandler;
     use lazy_static::lazy_static;
-    use mockall::mock;
     use parking_lot::Mutex;
     use serial_test::serial;
     use telio_crypto::PublicKey;
     use telio_crypto::SecretKey;
-    use telio_model::mesh::LinkState;
     use telio_sockets::{NativeProtector, SocketPool};
     use telio_utils::exponential_backoff::{
         ExponentialBackoff, ExponentialBackoffBounds, MockBackoff,
     };
-    use telio_utils::ip_stack::IpStack;
     use telio_wg::{
-        Error as wgError, WireGuard,
+        MockWireGuard,
         uapi::{Interface, Peer},
     };
     use tokio::sync::Mutex as TMutex;
 
     type Result<T> = std::result::Result<T, Error>;
-    type Result1<T> = std::result::Result<T, wgError>;
-
-    mock! {
-        pub Wg {}
-        #[async_trait]
-        impl WireGuard for Wg {
-            async fn get_interface(&self) -> Result1<Interface,>;
-            async fn get_adapter_luid(&self) -> Result1<u64>;
-            async fn wait_for_listen_port(&self, d: Duration) -> Result1<u16>;
-            async fn get_link_state(&self, key: PublicKey) -> Result1<Option<LinkState>>;
-            async fn set_secret_key(&self, key: SecretKey) -> Result1<()>;
-            async fn set_fwmark(&self, fwmark: u32) -> Result1<()>;
-            async fn add_peer(&self, peer: Peer) -> Result1<()>;
-            async fn del_peer(&self, key: PublicKey) -> Result1<()>;
-            async fn drop_connected_sockets(&self) -> Result1<()>;
-            async fn time_since_last_rx(&self, public_key: PublicKey) -> Result1<Option<Duration>>;
-            async fn stop(self);
-            async fn reset_existing_connections(&self, exit_pubkey: PublicKey) -> Result1<()>;
-            async fn set_ip_stack(&self, ip_stack: Option<IpStack>) -> Result1<()>;
-            async fn set_adapter_mtu(&self, mtu: u32) -> Result1<()>;
-        }
-    }
 
     lazy_static! {
         static ref IGD_IS_AVAILABLE: Arc<Mutex<bool>> = Arc::new(Mutex::new(false));
@@ -922,7 +896,7 @@ mod tests {
 
     pub async fn prepare_test_setup(
         is_battery_optimization_on: bool,
-    ) -> UpnpEndpointProvider<MockWg, MockUpnpEpCommands, MockBackoff> {
+    ) -> UpnpEndpointProvider<MockWireGuard, MockUpnpEpCommands, MockBackoff> {
         let spool = SocketPool::new(
             NativeProtector::new(
                 #[cfg(target_os = "macos")]
@@ -954,7 +928,7 @@ mod tests {
         epc.udp.set_port(2000);
 
         // These are not properly used yet, just dummy variables
-        let mut wg = MockWg::default();
+        let mut wg = MockWireGuard::default();
         let wg_port = 55345;
         let wg_peers = Vec::<(PublicKey, Peer)>::new();
         let backoff_array = [100, 200, 400, 800, 1600, 3200];
@@ -1108,7 +1082,7 @@ mod tests {
         }
 
         // These are not properly used yet, just dummy variables
-        let mut wg = MockWg::default();
+        let mut wg = MockWireGuard::default();
         let wg_port = 55345;
 
         wg.expect_get_interface().returning(move || {
--- a/crates/telio-wg/src/adapter/windows_native_wg.rs
+++ b/crates/telio-wg/src/adapter/windows_native_wg.rs
@@ -1,7 +1,10 @@
 use super::{Adapter, Error as AdapterError, IsMeshnetEnabledCb, Tun as NativeTun};
 use crate::{
     uapi::{Cmd, Cmd::Get, Cmd::Set, Interface, Peer, Response},
-    windows::{service, tunnel::interfacewatcher::InterfaceWatcher},
+    windows::{
+        service,
+        tunnel::{interfacewatcher::InterfaceWatcher, mtumonitor::set_interface_mtu},
+    },
 };
 use async_trait::async_trait;
 use sha2::{Digest, Sha256};
@@ -30,6 +33,7 @@ use tokio::sync::Notify;
 use tokio::time::{sleep, timeout};
 use utf16_lit::utf16_null;
 use uuid::Uuid;
+use winapi::shared::ws2def::{ADDRESS_FAMILY, AF_INET, AF_INET6};
 use windows::core::GUID;
 use windows::core::PCWSTR;
 use windows::Win32::Devices::DeviceAndDriverInstallation::GUID_DEVCLASS_NET;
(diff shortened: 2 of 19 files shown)


crates/telio-traversal/src/endpoint_providers/upnp.rs
```diff
@@ -869,6 +869,7 @@ mod tests {
             async fn stop(self);
             async fn reset_existing_connections(&self, exit_pubkey: PublicKey) -> Result1<()>;
             async fn set_ip_stack(&self, ip_stack: Option<IpStack>) -> Result1<()>;
+            async fn set_adapter_mtu(&self, mtu: u32) -> Result1<()>;
         }
     }
```
tomaszklak: This seems to be copy-pasted from the stun.rs - please extract the mock into one common place.

nat-lab/tests/test_adapter.py
```diff
@@ -45,6 +49,58 @@ async def get_interface_state(client_conn, client):
...
+            if actual == expected:
+                return
+            await asyncio.sleep(0.5)
+
+    try:
+        await asyncio.wait_for(poll(), MTU_CHANGE_TIMEOUT_S)
+    except asyncio.TimeoutError:
+        raise AssertionError(
+            f"Expected adapter MTUs {expected}, last read {actual}"
+        ) from None
```
tomaszklak: I think this is the first time I see this construct (`from None`) - would be good idea to explain (comment? commit description) why it is used here. Also, why are we raising Assertion error - shouldn't it be only used by the actual assertions?

nat-lab/tests/test_adapter.py
```diff
@@ -487,3 +543,48 @@ async def test_adapter_state_for_meshnet(
     await client_alpha.set_mesh_off()
     state = await get_interface_state(client_conn, client_alpha)
     assert state == expected_idle_state
+
+
+@pytest.mark.parametrize(
```
tomaszklak: If there is only one version of the param, there is no need for `parametrize`.

nat-lab/tests/test_adapter.py
```diff
@@ -487,3 +543,48 @@ async def test_adapter_state_for_meshnet(
...
+    async def test_changes_mtu_at_runtime(
+        self,
+        alpha_setup_params: SetupParameters,  # pylint: disable=unused-argument
+        env: Environment,
+    ) -> None:
+        client_conn, *_ = [conn.connection for conn in env.connections]
+        client_alpha, *_ = env.clients
+        expected_mtu = 1320
+
+        await client_alpha.set_adapter_mtu(expected_mtu)
```
tomaszklak, line 585: Please add an assertion that verifies that the mtu is not as expected before the call to `set_adapter_mtu`.

src/ffi.rs
```diff
@@ -452,12 +491,37 @@ impl Telio {
...
+    ///
+    /// Adapter will attempt to open its own tunnel.
+    pub fn start_with_config(
+        &self,
+        private_key: SecretKey,
+        adapter: TelioAdapterType,
+        config: StartConfig,
+    ) -> FfiResult<()> {
+        telio_log_info!(
+            "Telio::start entry with instance id: {}. Public key: {:?}. Adapter: {:?}. Config: {:?}",
```
tomaszklak: nit: It's `Telio::start_with_config` not `Telio::start`.

src/ffi.rs
```diff
@@ -452,12 +491,37 @@ impl Telio {
...
     }
 
+    /// Start telio with the specified adapter and the options in `config`.
+    ///
+    /// Adapter will attempt to open its own tunnel.
+    pub fn start_with_config(
+        &self,
+        private_key: SecretKey,
+        adapter: TelioAdapterType,
+        config: StartConfig,
```
tomaszklak, line 477: I think we should return an error when adapter is not windows and config.mtu is not None.

crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
```diff
@@ -141,19 +144,28 @@ impl InterfaceWatcher {
...
 
         telio_log_trace!("--- InterfaceWatcher::stop");
     }
 
+    pub fn set_forced_mtu(&mut self, mtu: u32) {
+        telio_log_trace!("+++ InterfaceWatcher::set_forced_mtu");
+
+        if let Ok(mut watched_adapter) = self.watched_adapter.clone().lock() {
+            watched_adapter.forced_mtu = Some(mtu);
+            Self::stop_mtu_monitors(&mut watched_adapter);
```
tomaszklak, line 162: If I `set_adapter_mtu(Some(...))` and later on `set_adapter_mtu(None)` I would expect that the monitors will be restarted.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -51,6 +55,8 @@ use wireguard_nt::{
 use wireguard_uapi::xplatform;
 
 const REMOVAL_SLEEP_SECS: u64 = 2;
+// IPv6 minimum link MTU, the interface MTU applies to both address families
+const MIN_MTU: u32 = 1280;
```
tomaszklak: If the intention is to add `set_adapter_mtu` support to other adapter types in future, the lower bound validation should be moved to `device.rs` (or maybe `wg.rs`). Unless this validation is somehow windows specific?

nat-lab/tests/test_adapter.py
```diff
@@ -487,3 +543,48 @@ async def test_adapter_state_for_meshnet(
...
+                features=default_features(enable_dynamic_wg_nt_control=False),
+            ),
+            marks=[pytest.mark.windows],
+        ),
+    ],
+)
+class TestAdapterMtu:
+    """Setting the adapter MTU, supported on Windows only."""
+
+    async def test_starts_with_configured_mtu(
```
tomaszklak: There should be a test for how `start_with_config` behaves on other platforms.

nat-lab/tests/test_adapter.py
```diff
@@ -487,3 +543,48 @@ async def test_adapter_state_for_meshnet(
...
+        client_conn, *_ = [conn.connection for conn in env.connections]
+        client_alpha, *_ = env.clients
+        expected_mtu = 1360
+
+        await client_alpha.stop_device()
+        await client_alpha.start_with_config(StartConfig(mtu=expected_mtu))
+
+        await wait_for_interface_mtu(client_conn, client_alpha, expected_mtu)
+
+    async def test_changes_mtu_at_runtime(
```
tomaszklak: There should be a test for how `set_adapter_mtu` behaves on other platforms.

nat-lab/tests/test_adapter.py
```diff
@@ -487,3 +543,48 @@ async def test_adapter_state_for_meshnet(
...
+        self,
+        alpha_setup_params: SetupParameters,  # pylint: disable=unused-argument
+        env: Environment,
+    ) -> None:
+        client_conn, *_ = [conn.connection for conn in env.connections]
+        client_alpha, *_ = env.clients
+        expected_mtu = 1320
```
tomaszklak, line 579: Test for rejection of too low mtu is missing.

src/libtelio.udl
```diff
@@ -306,6 +306,12 @@ interface Telio {
     [Throws=TelioError]
     void start_named_ext_if_filter(SecretKey secret_key, TelioAdapterType adapter, string name, sequence<string> ext_if_filter);
 
+    /// Start telio with the specified adapter and the options in `config`.
+    ///
+    /// Adapter will attempt to open its own tunnel.
+    [Throws=TelioError]
+    void start_with_config(SecretKey secret_key, TelioAdapterType adapter, StartConfig config);
```
tomaszklak, line 313: Can you express all the other `start` methods from udl using this new one?

nat-lab/tests/test_adapter.py
```diff
@@ -487,3 +543,48 @@ async def test_adapter_state_for_meshnet(
...
+                connection_tag=ConnectionTag.VM_WINDOWS_1,
+                adapter_type_override=TelioAdapterType.WINDOWS_NATIVE_TUN,
+                is_meshnet=False,
+                features=default_features(enable_dynamic_wg_nt_control=False),
+            ),
+            marks=[pytest.mark.windows],
+        ),
+    ],
+)
+class TestAdapterMtu:
```
tomaszklak, line 554: Please add a case with `enable_dynamic_wg_nt_control=True` - set the mtu while the adapter is down, connect to VPN so that it's up, disconnect. The mtu should be set correctly during the whole test after the initial mtu setting.

stalowyjez: Added

src/libtelio.udl
stalowyjez: Custom adapter is kinda hard to express, but rewrote the rest - wdyt?

nat-lab/tests/test_adapter.py
stalowyjez: Added

nat-lab/tests/test_adapter.py
stalowyjez: Added

nat-lab/tests/test_adapter.py
stalowyjez: Added

crates/telio-wg/src/adapter/windows_native_wg.rs
stalowyjez: Moved

crates/telio-wg/src/windows/tunnel/interfacewatcher.rs
stalowyjez: Well, that implies adding option to set_mtu API, but makes sense, added

src/ffi.rs
stalowyjez: Fixed

src/ffi.rs
stalowyjez: Fixed

nat-lab/tests/test_adapter.py
stalowyjez: Removed

nat-lab/tests/test_adapter.py
stalowyjez: Simplified that a bit

crates/telio-traversal/src/endpoint_providers/upnp.rs
stalowyjez: Let's just use default one in both locations

nat-lab/tests/test_adapter.py
stalowyjez: i.e. it was on the other test, here there are some cases, as we check if it is unsupported for different setups

src/ffi.rs
```diff
@@ -145,6 +145,61 @@ pub extern "C" fn Java_com_nordsec_telio_TelioCert_initCertStore(
...
+        adapter: TelioAdapterType,
+    ) -> FfiResult<DeviceConfig> {
+        #[cfg(not(target_os = "windows"))]
+        let tun = self.tun.map(|fd| {
+            use std::os::fd::{FromRawFd, OwnedFd};
+            // SAFETY: the caller hands over an open descriptor it no longer uses
+            unsafe { OwnedFd::from_raw_fd(fd) }
+        });
+        #[cfg(target_os = "windows")]
+        let tun = None;
```
tomaszklak, line 182: I think it would be best to move this right before the return. Otherwise the failed mtu check will close the fd passed by the caller.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -699,6 +708,46 @@ impl WindowsNativeWg {
...
+            match set_interface_mtu(self.luid, family, mtu) {
+                Ok(()) => telio_log_info!("Set adapter MTU {} for family {}", mtu, family),
+                Err(err) => {
+                    telio_log_warn!(
+                        "Failed to set adapter MTU {} for family {}: {}",
+                        mtu,
+                        family,
+                        err
+                    );
+                    if failed {
```
tomaszklak, line 739: Should we undo the effects of `set_forced_mtu(Some(...))` here?

src/doc/integrating_telio.md
```diff
@@ -337,10 +337,11 @@ let adapter = get_default_adapter();
 
 let telio = Telio::new(Default::default(), Box::new(EventHandler)).unwrap();
 
-// There are three ways to start telio:
+// There are four ways to start telio:
 // * start - telio does everything
 // * start_with_tun - use existing tun (android, apple)
 // * start_with_name - create tun with name (windows, linux)
+// * start_with_config - like start, with optional settings (name, ext_if_filter, mtu)
```
tomaszklak, line 344: Here (and in other places) not all settings are listed.

tomaszklak (commented):
Please cleanup the commit history (move review fixes to appropriate commits, so that they already appear in the final correct form).

PR #1981, opened 2026-09-08
Probe SSH in openwrt healthcheck

olekoliinyk:
### Problem
`test_openwrt_simulate_network_down` failed  in `openwrt-gw-03`: the test could not open an SSH session to the router and timed out after 30s.
The container was reported as healthy, because the healthcheck only opened a TCP connection to port 22.

### Solution
- Make the healthcheck ask for the SSH greeting instead of just opening a socket (drop it as SSH is more complete check in this case)
- Raise openwrt healthcheck timeouts to 30s, since the check can now wait up to 20s
- Drop the QEMU guest agent block: no guest agent socket is ever created, and it queried the QMP socket instead, so it could never have worked

- [x] README.md is updated


asledzinskiy:
+1

matislovas (approved):
+1

PR #1982, opened 2026-09-08
LLT-7632: Use private key from requested state to configure DERP

tomaszpatejko:
### Problem

If a DERP peer sets a secret key on its WG interface incorrectly (due to, for example, error returned by WG interface), other DERP peers, will not be able to identify this peer with a public key derived from its secret key.

This happens because DERP relay receives the incorrect secret key in the DERP config and uses this key to establish direct connection between peers.

### Solution
A secret key from `RequestedState::device_config` is used to configure DERP. This key in turn is used for WG interface reconfiguration in `consolidate_wg_state` function.

Fixes LLT-7632.

- [x] README.md is updated

### DIFF
--- a/crates/telio-core/src/device.rs
+++ b/crates/telio-core/src/device.rs
@@ -2070,7 +2070,7 @@ impl Runtime {
             }
 
             let derp_config = DerpConfig {
-                secret_key,
+                secret_key: self.requested_state.device_config.private_key.clone(),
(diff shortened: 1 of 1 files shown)


matislovas:
+1

mathiaspeters (commented):
+1

mathiaspeters (approved):

PR #1985, opened 2026-09-09
Restart nordvpnlite instead of start in test_openwrt_router_restart

olekoliinyk:
### Problem
When `test_openwrt_router_restart` executed in the first order in pytest session we've got a following:
1. Installs nordvpnlite and starts the procd service
2. Daemon has no credentials, exits immediately, and procd respawns it every 5s
3. Test writes a valid config and auth file to the default paths the service reads
4. Next respawn finds both and starts successfully
5. In test nordvpnlite start fails with `DaemonIsRunning`

It's not happening when procd already died (crash loop is already over).

### Solution
Temporary  solution is to restart nordvpnlite instead of start in test_openwrt_router_restart.
The permanent fix is described in LLT-7676.

- [x] README.md is updated

### DIFF
--- a/nat-lab/tests/test_openwrt.py
+++ b/nat-lab/tests/test_openwrt.py
@@ -656,17 +656,24 @@ async def test_openwrt_router_restart(
                 ["mkdir", "-p", "/etc/nordvpnlite"]
             ).execute()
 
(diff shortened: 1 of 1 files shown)


stalowyjez (approved):
Looks good! +2

PR #1989, opened 2026-09-11
TCP forwarder proxy

tomasz-grz:
## Problem
libtelio only forwards DNS over UDP. A client that uses TCP, either because of truncated UDP or because it just prefers it, gets a connection RST.

DNS over TCP is a byte stream, messages have to be reassembled rather than arriving one per packet. And the connections can be long lived.

## Solution
Adds a `tcp_forwarder` module to with two parts:

### Frame
A frame assembler that buffers bytes and hands back one complete query at a time.
A check whether a frame contains a `.nord` query.

### Proxy
A per connection task that walks the configured upstreams until one connects and answers.
One query is outstanding at a time. If an upstream drops the connection, the outstanding query resent to the next upstream. An upstream that has already answered once is retried once before moving on.

- [x] README.md is updated


crates/telio-dns/src/packet_decoder.rs
```diff
@@ -65,12 +65,12 @@ pub fn parse_dns_query_packet(packet_bytes: &[u8]) -> Result<DnsPacket<'_>, DnsP
 pub fn find_nord_query(dns_packet: &DnsPacket) -> Option<DnsQuery> {
     let opcode = dns_packet.get_opcode();
     if opcode != Opcodes::StandardQuery {
-        telio_log_warn!("Unsupported Opcode for nord query: {opcode:?}");
+        telio_log_debug!("Unsupported Opcode for nord query: {opcode:?}");
```
tomaszpatejko: Why was this changed? I feel it could still be kept as warning.

Paintree (dismissed):
+1, seems 🆗

crates/telio-dns/src/packet_decoder.rs
tomasz-grz: I was worried that this could spam the production logs.. But good point, it's still far from getting there, could possibly stay and we can reduce the level in case we encounter any issues 😅

crates/telio-dns/src/tcp_forwarder/proxy.rs
```diff
@@ -0,0 +1,1010 @@
...
+
+/// Per connection proxy task.
+///
+/// Walks the configured upstream list until one
+/// connects and replies, serializing queries through [`run_proxy_pipe`].
+///
+/// `generation` identifies this connection
+/// so the engine can discard events from a past connections.
+pub(crate) async fn run_proxy(
+    upstreams: Arc<Mutex<UpstreamList>>,
```
tomaszpatejko, line 131: I noticed that whenever `UpstreamList` is used it's wrapped in `Arc<Mutex<>>` and uses generation counter internally so I'm wondering how likely is that a shared instance of `UpstreamList` is modified at runtime while connections are being established?

crates/telio-dns/src/tcp_forwarder/test_utils.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/// Mint an opaque `SocketHandle`.
+pub(crate) fn test_handle() -> SocketHandle {
+    let rx = tcp::SocketBuffer::new(vec![0u8; 1]);
+    let tx = tcp::SocketBuffer::new(vec![0u8; 1]);
+    SocketSet::new(Vec::new()).add(tcp::Socket::new(rx, tx))
+}
+
+pub(crate) enum StubBehavior {
```
tomaszpatejko, line 55: What is `Stub` here? Is that some kind of mock? A mock server?

crates/telio-dns/src/packet_decoder.rs
tomasz-grz: Removed

crates/telio-dns/src/tcp_forwarder/proxy.rs
tomasz-grz: To be honest, I think not so likely..
If I'm not mistaken, the Apps might even just restart whole tunnel once you change toggle TP-lite (the most probable source of changing upstreams).

crates/telio-dns/src/tcp_forwarder/test_utils.rs
tomasz-grz: Sort of, I would say it's more like a "fake" than a "mock".

We spawn the stub upstream server locally, with the given hard-coded behavior. So we have control over both sides of the connection and simulate some scenarios we might encounter in the wild. 

Since you can't really mock code dealing with sockets, it's still doing networking, but over loop-back interface.

crates/telio-dns/src/tcp_forwarder/proxy.rs
tomasz-grz: It might change though when switching from active Meshnet to VPN connection for example, Apple as far as I know sends a lot of DNS requests in that moment.

tomaszpatejko: Understood. Resolving.

tomaszpatejko:
+1

PR #1992, opened 2026-09-13
#1991: Fix DERP reconnect storm: unchecked RecvPacket panic + backoff bypass

### LINKED ISSUE #1991: DERP relay: unchecked frame-length panic + reconnect-backoff bypass enable unthrottled reconnect storm
# Summary

A DERP relay (trusted, TLS-verified) can force the client into a continuous, unthrottled reconnect loop by sending one malformed frame on every connection. Two defects combine:

1. proto.rs:200, start_read(), FrameType::RecvPacket arm:
   data.drain(0..KEY_SIZE) panics if the server-declared frame body is shorter than KEY_SIZE (32 bytes) — no length check before the drain, unlike the sibling read_server_key() (proto.rs:360), which guards with if bytes.len()< 40.
2. derp.rs, start_connecting() (~L211) + wait_with_update (~L786): theexponential backoff only engages when config.servers.get_next() == None (whole server list exhausted on connect failure). start_connecting takesconfig by value and every reconnect passes a fresh config.clone(), so current_server_num and sleep_time never persist across reconnects. A server that connects fine and then panics immediately after never reaches the None branch — no cooldown ever applies to this failure class.

# Impact

Impacket is DOS but with a lot pre conditions that why i am reporting it here, and it is worth fixing, as Sustained, attacker-controlled resource drain (CPU/socket/log churn) and indefinite denial of meshnet relay connectivity for as long as the malicious/compromised relay keeps repeating the pattern. 

# Fix

1. proto.rs: replace data.drain(0..KEY_SIZE) with a length-checked path (data.get(0..KEY_SIZE).ok_or(...)?), matching read_server_key().
2. derp.rs: persist current_server_num/sleep_time at the State level (not inside the per-call discarded Config clone), and/or enforce a minimum cooldown between "connection established" and "next reconnect attempt" regardless of which get_next() branch is hit.

michealkeines:
Fixes https://github.com/NordSecurity/libtelio/issues/1991

### Problem

A malicious/compromised DERP relay could send a RecvPacket frame with a body shorter than KEY_SIZE (32 bytes), causing an unchecked Vec::drain(0..KEY_SIZE) panic in start_read. Because the reconnect backoff only engaged when every server in the list failed to connect (not when a connection died immediately after connecting), this let a hostile relay hold the client in an unthrottled reconnect loop indefinitely.

### Solution

- `proto.rs`: `start_read`'s `RecvPacket` handling called `data.drain(0..KEY_SIZE)` with no length check, panicking on a frame shorter than 32 bytes. Now returns an error instead, matching the existing guard in `read_server_key`.
- `derp.rs`: the reconnect backoff only engaged when every configured server failed to *connect* — a server that connected and then died immediately (e.g. via the panic above) reconnected with no cooldown at all. Added `last_connected_at` tracking and a 1s `MIN_RECONNECT_COOLDOWN` before any new connection attempt.

- [x] README.md is updated


nord-llt:
[CLA assistant check](https://cla.nordsec.com/NordSecurity/libtelio?pullRequest=1992) <br/>All committers have signed the CLA.

crates/telio-relay/src/derp.rs
```diff
@@ -15,7 +15,7 @@ use futures::{future::select_all, Future};
 use std::collections::{HashMap, HashSet};
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
```
tomasz-grz: We generally prefer `use telio_utils::Instant;` instead, since it's suspend aware.

tomasz-grz (commented):
Hi @michealkeines thank you for your contribution!

Generally looks OK, just some steps to make our CI pass:
- Could you please create an empty change-log file in `.unreleased` like `touch .unreleased/derp-reconnect-storm`.
- You might need to re-base off `main` again due to some advisory fixes.

crates/telio-relay/src/derp.rs
michealkeines: updated

crates/telio-relay/src/derp.rs
```diff
@@ -795,6 +808,7 @@ impl Runtime for State {
                             Ok((server, conn)) => {
                                 self.server = Some(server.clone());
                                 self.conn = Some(conn);
+                                self.last_connected_at = Some(Instant::now());
```
tomasz-grz, line 811: Do we need to clear `last_connected_at` on `disconnect()` to prevent slowdowns on deliberate reconnects (like roaming or network change)?

crates/telio-relay/src/derp.rs
```diff
@@ -783,6 +791,11 @@ impl Runtime for State {
                 let connecting = if let Some(connecting) = &mut self.connecting {
                     connecting
                 } else {
+                    if let Some(elapsed) = self.last_connected_at.map(|t| t.elapsed()) {
+                        if elapsed < MIN_RECONNECT_COOLDOWN {
+                            sleep(MIN_RECONNECT_COOLDOWN - elapsed).await;
```
tomasz-grz, line 796: Does the `sleep()` in the runtime loop block other tasks from being processed?

tomasz-grz, line 796: Would be nice to have a `telio_log_debug!` to track when we enter cooldown.

crates/telio-relay/src/derp.rs
tomasz-grz: Nice thank you

crates/telio-relay/src/derp.rs
tomaszpatejko, line 796: I'm wondering if this sleep strategy makes sense for the case when `elapsed` is close but lower than `MIN_RECONNECT_COOLDOWN`. Then `sleep` will last ~1ms (granularity of `tokio::time::sleep`). Does it make sense to `sleep` for such short period of time? Is the case actually worth considering?
@tomasz-grz @Jauler

crates/telio-relay/src/derp.rs
michealkeines: yes, i have reviewed all the places and replaced them

crates/telio-relay/src/derp.rs
michealkeines: yes it would block, i have fixed it by moving the sleep into the spawned start_connecting task instead of the actor loop wait_with_update now reaches select! immediately

would this be enough? or overkill

michealkeines: added

michealkeines: catching the smaller elapsed (1-15ms) will be racing the scheduler, that is also not worth the extra check, right?

tomaszpatejko: Yes, but I was also wondering if handling short sleeping periods separately is something worth considering. 

I think what code covers is fine.

michealkeines:
@tomasz-grz please review, I have added the changes

PR #1995, opened 2026-09-14
Migrate to rust-toolchain.toml format

tomasz-grz:
### Problem
`rust-toolchain` file is outdated compared to `rust-toolchain.toml` format. And some newer tools are not picking it up.

### Solution
- Update project to use `rust-toolchain.toml`.
- Update linter job to use built in runner `cargo` instead of archived `actions-rs/cargo` action. It should automatically pick the component version from `rust-toolchain.toml`.

Screenshot 2026-09-14 at 13 15 33

- [x] README.md is updated


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.89.0"
+components = ["rustfmt", "clippy", "rust-analyzer"]
```
tomasz-grz, line 3: actually gonna double check how adding `rust-analyzer` is impacting CI speed.

tomasz-grz: `["rustfmt", "clippy", "rust-analyzer"]` Takes 10s
`["rustfmt"]` Takes 9s

So I guess it's OK to keep it for developer convenience

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.89.0"
```
sfraczek, line 2: Can we bump to 1.89.1?

sfraczek (commented):
+1

rust-toolchain.toml
tomasz-grz: There is [no](https://releases.rs/docs/1.89.1/) `1.89.1` :D

tomasz-grz: I would rather update straight to `1.9x` but part of different PR

mathiaspeters (approved):
+1

PR #2008, opened 2026-09-18
LLT-7662: update dependencies

breiker:
### Problem
Dependencies need to be updated.

### Solution

Compatible:
* `time` 0.3.47 to 0.3.55
  `time::format_description::parse` is deprecated, switch to `parse_borrowed::<3>`

Incompatible:
* `base64` 0.22.1 to 0.23.1
* `dirs` 6.0.0 to 7.0.0
* `mockall` 0.14.0 to 0.15.0
* `rstest` 0.26.1 to 0.27.0
* `surge-ping` 0.8 to 0.9
  Use NordSecurity's fork of pnet_packet for surge-ping dependency (both on 0.35.0)
  Pin to 0.9.0, 0.9.1 uses AtomicU64, not available on 32bit MIPS.
* `netlink-packet-core` 0.8.2 to 0.9.0,
  `netlink-packet-route` 0.31.0 to 0.33.0
* `system-configuration` 0.7.0 to 0.8.0,
  `core-foundation` 0.9 to 0.10
* `rustls-platform-verifier` 0.6.2 to 0.7.0, `jni` 0.21 to 0.22
  `rustls-platform-verifier` 0.7 depends on `jni` 0.22.
  `jni` 0.22 splits `JNIEnv` into `EnvUnowned` and `Env`.
  `rustls-platform-verifier` removes `init_hosted` for `init_with_env`.
  Exported function signature stays the same.

Not upgraded:
* `enum-map` `2.7.3` to `3.1.0` - needs rust `1.95.0`
* `histogram` 0.6.9 to 1.6.0 - changes in behavior, will be done in new PR
* `uniffi` 0.28.3 to 0.32.1 - newest version of Nord's `uniffi-generators`
* `rustyline` 17.0.2 to 18.0.1 - nightly rust in nagger is too old (by 5 days, File::[lock,lock_shared]), need >= 1.89.0

- [x] README.md is updated


Cargo.toml
```diff
@@ -158,7 +158,8 @@ smart-default = "0.7.1"
 sn_fake_clock = "0.4"
 socket2 = "0.6"
 strum = { version = "0.28.0", features = ["derive"] }
-surge-ping = { version = "0.8" }
+# 0.9.1 uses AtomicU64, which is not available on 32-bit MIPS (mipsel-unknown-linux-musl)
+surge-ping = { version = "=0.9.0" }
```
sfraczek: nit: could we reference a ticket or an upstream issue here, so it's clear when this exact pin can be dropped? The comment explains why 0.9.1 is a problem, but not what has to happen to unpin it (surge-ping moving to portable-atomic / AtomicUsize, or us dropping 32-bit MIPS). Worth keeping in mind that an `=` requirement will also hard-conflict if anything else in the graph ever requires ^0.9.1.

Cargo.toml
```diff
@@ -227,3 +228,8 @@ normal = ["telio-starcast"]
 
 [workspace.metadata.cargo-machete]
 ignored = ["tracing"]
+
+# surge-ping has dependency on pnet_packet, use our fork in both places (git tag "v0.36.0" is still pnet 0.35.0) 
+# surge-ping doesn't use removed gre or DNS parts which fork touches
+[patch.crates-io]
+pnet_packet = { git = "https://github.com/NordSecurity/libpnet.git", tag = "v0.36.0" }
```
sfraczek, line 235: Non-blocking: could you confirm that `Cargo.lock` ends up with a single `pnet_packet` entry (the fork) rather than two copies? That's the main thing this patch is supposed to buy us.

(Edit: dropping my earlier question about downstream repos - `telio` is `crate-type = ["cdylib"]` and is consumed through the generated bindings, not as a Rust git/path dependency, so this workspace is always the build root and the patch always applies.)

clis/nordvpnlite/Cargo.toml
```diff
@@ -35,7 +35,7 @@ reqwest = { version = "0.13.3", default-features = false, features = [
 ] }
 uuid = { workspace = true, features = ["serde"] }
 anyhow.workspace = true
-base64 = "0.22.1"
+base64 = "0.23.1"
```
sfraczek: nit: nordvpnlite is the only crate in the repo that declares base64 on its own - tcli, derpcli, telio-core, telio-crypto, telio-model, telio-proto and telio-pq all use base64.workspace = true. That's why this bump had to be made in two places, and the two now differ slightly (0.23 in the workspace vs 0.23.1 here). No features/default-features difference here, so the workspace dep should be a drop-in. Can you change it here?
```suggestion
base64.workspace = true
```

crates/telio-lana/src/event_log_file.rs
```diff
@@ -21,9 +21,10 @@ fn event_log(
 ) -> std::result::Result<usize, moose::Error> {
     match get_logfile() {
         Ok(mut file) => {
-            let format_string =
-                time::format_description::parse("[year]-[month]-[day] [hour]:[minute]:[second] ")
-                    .map_err(|_| moose::Error::EventLogError)?;
+            let format_string = time::format_description::parse_borrowed::<3>(
```
sfraczek: optional, not for this PR: the format string is a literal, so time::macros::format_description! would move the parsing to compile time and remove this error branch (and the per-call parse) entirely. Only worth it if enabling the time/macros feature for telio-lana is acceptable.

.unreleased/LLT-7662-update_dependencies
sfraczek: This entry is empty (GitHub reports it as whitespace-only). Is that intentional for a dependency-only change, or should it carry a one-line description so it shows up properly in changelog.md?

sfraczek (dismissed):
+1

clis/nordvpnlite/Cargo.toml
breiker: Updated

Cargo.toml
breiker: Yes, with this change there is only one copy of `pnet_packet`. Without it there were two: NordSecurity's and crates.io.

djkarwowski (approved):
+1.0

From what the diff says, the `netlink-packet-core` was bumped from `0.8.1`, but the description says from `0.8.2`. Can you please update the description accordingly?

breiker:
> From what the diff says, the `netlink-packet-core` was bumped from `0.8.1`, but the description says from `0.8.2`. Can you please update the description accordingly?

Updated PR description and description in commit. I mistakenly put `0.8.2` because `0.8.1` to `0.8.2` was compatible update done by `cargo update`.

PR #2009, opened 2026-09-21
Add wgnt device creation retry forwardport

matislovas:
Forwardport of https://github.com/NordSecurity/libtelio/pull/1962

- [x] README.md is updated


crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -216,10 +221,21 @@ impl WindowsNativeWg {
...
+    /// primary adapter keeps its identity - and with it the persistent firewall, DNS and
+    /// network profile state keyed on the interface GUID.
+    ///
+    /// Retries walk a small fixed pool instead of a random GUID, which keeps the amount of
+    /// per-GUID state Windows accumulates bounded at `ADAPTER_GUID_POOL_SIZE` per name.
+    fn get_adapter_guid_from_name_hash(name: &str, attempt: usize) -> u128 {
         let mut hasher = Sha256::new();
         hasher.update(name);
+        if attempt > 0 {
+            hasher.update([0u8]);
```
Jauler: Could you clarify the purpose of this `0` hash? is it the separator to avoid same-postfix match?

matislovas: :crying_cat_face:  a leftover, but updated the following line, for GUIDs to be consisting between 32 and 64 bit platforms.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -446,6 +479,63 @@ impl WindowsNativeWg {
...
+                .open_subkey(format!(r"{net_key}\Connection"))
+                .and_then(|key| key.get_value::<String, _>("PnpInstanceID"))
+                .map(|id| id.to_uppercase().starts_with(r"SWD\WIREGUARD\"))
+                .unwrap_or(false);
+            if !owned_by_us {
+                continue;
+            }
+
+            for key in [
+                net_key,
```
Jauler: aahm, are we deleting _everything_ under NET?

Jauler: oh, we are not. NVM ignore.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -446,6 +479,63 @@ impl WindowsNativeWg {
...
+        const NET_CLASS_KEY: &str =
+            r"SYSTEM\CurrentControlSet\Control\Network\{4d36e972-e325-11ce-bfc1-08002be10318}";
+
+        let hklm = RegKey::predef(HKEY_LOCAL_MACHINE);
+        let mut removed: bool = false;
+
+        for attempt in 1..ADAPTER_GUID_POOL_SIZE {
+            let guid = Self::get_adapter_guid_from_name_hash(name, attempt);
+            let guid_str = Self::guid_to_windows_string(guid);
+            let net_key = format!(r"{NET_CLASS_KEY}\{guid_str}");
```
Jauler: I know this is a nitpick, but could we rename `net_key` to `interface_key`. As I believe it may be misleading, and we have to really be careful when deleting stuff in registry.

matislovas: Fixed

Jauler (dismissed):
+1.0. But please merge only if NordLayer confirms that this will not break their integration 🙏.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -446,6 +479,63 @@ impl WindowsNativeWg {
...
+    ///
+    /// Deliberately bounded and targeted: only the `ADAPTER_GUID_POOL_SIZE` GUIDs we derive
+    /// ourselves, only slots `1..` (slot 0 is the primary adapter, whose settings we want to
+    /// keep), and only when the `Connection\PnpInstanceID` confirms the entry really was ours.
+    ///
+    /// `Control\NetworkSetup2\Interfaces\{GUID}` is intentionally *not* touched - it is
+    /// SYSTEM-owned on Windows 10 and later, and a stale entry there is inert.
+    fn cleanup_pool_guid_netcfg(name: &str) -> bool {
+        const NET_CLASS_KEY: &str =
+            r"SYSTEM\CurrentControlSet\Control\Network\{4d36e972-e325-11ce-bfc1-08002be10318}";
```
tomasz-grz, line 493: What is this magic hard coded key? How is it derived and does it ever change?

matislovas: Nope, this ones is permanent. LLM explanation:

```
The registry path SYSTEM\CurrentControlSet\Control\Network\{4d36e972-e325-11ce-bfc1-08002be10318} 
stores configuration and naming data for network connections 
and network interface cards (NICs) in Windows.
```

Jauler: Its microsoft's constant: https://learn.microsoft.com/en-us/windows-hardware/drivers/install/system-defined-device-setup-classes-available-to-vendors

tomasz-grz: Got it, thanks!

tomasz-grz (dismissed):
+1

crates/telio-model/src/features.rs
```diff
@@ -95,6 +99,24 @@ pub struct FeatureWireguard {
     pub max_inter_thread_batched_pkts: Option<u32>,
 }
 
+fn default_wg_nt_guid_rotation() -> bool {
+    true
```
Jauler, line 103: IMHO it would be better to make it disabled by default. 

This is mostly because, feature config servers two purposes, risk mitigation of undiscovered bugs (but all is good with this one). And backwards compatibility. In our specific case, this flag allows nordlayer to postpone adaptation of new GUIDs until they have time for it. If we make it enabled by default - nordlayer _will_ have to adapt anyways (either by setting the flag to false, or by actually adapting GUIDs), which kind of makes the flag no longer serve its purpose. 

NordVpn Windows team will be adapting anyways, meaning, that they are directly impacted by existance of this feature anyways.

crates/telio-wg/src/adapter/windows_native_wg.rs
```diff
@@ -463,16 +558,53 @@ impl WindowsNativeWg {
 
         // Cleaning orphaned WireGuard-NT adapters inside the driver can be racey,
         // so we do it in a separate step, before creating a new adapter
-        Self::cleanup_orphaned_devices().await;
+        Self::cleanup_orphaned_devices(name).await;
```
Jauler, line 561: I wonder 🤔 would it make sense to respect enable_guid_rotation here?

crates/telio-model/src/features.rs
matislovas: For NordVpn Windows team this would be quite weird, because `v7.0.9` was released without this flag (and is already in progress to being adopted). So to avoid ambiguity for them, I would leave it by default enabled.

And I do not see any problems for nordlayer, if they're warned in advance to disable this flag. And this flag literally fixes the bug or hacks-around it (although, we're not sure that the bug is in our cleanup process or in OS itself).

crates/telio-wg/src/adapter/windows_native_wg.rs
matislovas: There's a comment in `cleanup_pool_guid_netcfg`:

```
    /// Runs regardless of the GUID rotation feature flag, so turning rotation off still
    /// clears whatever earlier runs left in the pool.
```

So ... that's why it was left out :smile:

crates/telio-model/src/features.rs
Jauler: but is it useful then? If nordlayer will have to do changes anyways. I assume the converting single GUID to a list is comparable to setting the flag to false.

matislovas: So nordlayer will only have to set the flag to false, it is a major upgrade anyways (in terms of v8) so is expected not to be a drop-in replacement to a previous version.

Jauler: but v8 is already released. And v8.0.0 would be compatible with nordlayer's flow, while v8.0.1 (I suspect next release is v8.0.1) is not.

matislovas: Done

PR #2011, opened 2026-09-22
refactor(ffi): unify exit node connect API via config builder

tomasz-kumor:
### Problem

Adding a dedicated `connect_to_exit_node_*` function per variant does not scale. With AEGIS cipher support, new variants will be needed. 

### Solution

This commit adds a single `connect_to_exit_node_with_config` entry point that accepts a `ConnectionConfig` produced by dedicated builders for VPN and meshnet connections. Future variants can extend the config without introducing new connection functions.

- [x] README.md is updated


crates/telio-core/src/connection_config.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// Direct UDP endpoint of the exit node.
+    /// Always `Some` for VPN connections; always `None` for meshnet.
+    pub endpoint: Option<SocketAddr>,
+    /// When `true`, a post-quantum tunnel is established.
+    /// Only meaningful for VPN connections.
+    pub post_quantum: bool,
+}
+
+pub struct VpnConnectionConfigBuilder {
+    config: Mutex<ConnectionConfig>,
```
matislovas, line 31: Do we really need a lock here 🤔

src/ffi.rs
```diff
@@ -765,19 +815,15 @@ impl Telio {
...
-            allowed_ips,
-            endpoint: Some(endpoint),
-        };
-        catch_ffi_panic(|| {
-            self.device_op(true, |dev| {
-                dev.connect_vpn_post_quantum(&node)
-                    .log_result("Telio::connect_vpn_post_quantum")
-            })
-        })
+        let mut builder = Arc::new(VpnConnectionConfigBuilder::new(public_key, endpoint));
```
matislovas, line 795: Do we need `Arc` here?

src/ffi.rs
```diff
@@ -725,22 +771,26 @@ impl Telio {
             allowed_ips,
             endpoint,
         );
-        let identifier = identifier.unwrap_or_else(|| Uuid::new_v4().to_string());
-        let node = ExitNode {
-            identifier,
-            public_key,
-            allowed_ips,
-            endpoint,
+        let config = if let Some(ep) = endpoint {
```
matislovas, line 759: I would leave `connect_to_exit_node` and `connect_to_exit_node_postquantum` implementations untouched (to not introduce any unwanted into them, do not touch the parts that work), and would only keep the `connect_to_exit_node_with_config` separate and the only one, which can work with custom ciphers. But I'm not strict with this approach.

src/ffi.rs
```diff
@@ -683,7 +686,50 @@ impl Telio {
...
+    /// This is the canonical implementation. All other `connect_to_exit_node*` variants
+    /// are thin wrappers that build a `ConnectionConfig` and delegate here.
+    ///
+    /// Routing should be set by the user accordingly.
+    ///
+    /// # Parameters
+    /// - `config`: Connection configuration produced by [`connection_config::VpnConnectionConfigBuilder`]
+    ///   or [`connection_config::MeshnetConnectionConfigBuilder`].
+    pub fn connect_to_exit_node_with_config(&self, config: ConnectionConfig) -> FfiResult<()> {
+        telio_log_info!(
```
matislovas, line 700: This log will be duplicated with calling from `connect_to_exit_node_postquantum` and `connect_to_exit_node_with_id`.

matislovas:
Can you add some infor about it to `README` and `integrating_telio.md`?

crates/telio-core/src/connection_config.rs
tomasz-kumor: I was inspired by `FeaturesDefaultsBuilder`. My understanding is that both `Mutex` and `Arc` are needed because this object is crossing FFI boundary. If no FFI  neither would be needed.

src/ffi.rs
tomasz-kumor: see comment above

src/ffi.rs
tomasz-kumor: I did it according to Rytis suggestion in RFC 

> For backwards compatibility - we would have to leave old functions operating in the same manner like before, but internally it may be wrappers around for connect_to_exit_node_with_config()

src/ffi.rs
tomasz-kumor: I have spotted that but forgot to change it

tomasz-kumor:
> Can you add some infor about it to `README` and `integrating_telio.md`?

Updated `integrating_telio.md`

matislovas:
+1

src/ffi.rs
djkarwowski, line 1: What do you think about adding a small extension to the `connection_config` module:

```
pub trait ApplySome: Sized {
    fn apply_some<T>(self, opt: Option<T>, f: impl FnOnce(Self, T) -> Self) -> Self {
        match opt {
            Some(v) => f(self, v),
            None => self,
        }
    }
}

impl<S> ApplySome for S {}
```

And then the call sites could use more fluent form (plus some type aliasing to make it a bit more concise):

```
type Vpn = VpnConnectionConfigBuilder;
type Mesh = MeshnetConnectionConfigBuilder;
```

```
    pub fn connect_to_exit_node_with_id(
        &self,
        identifier: Option<String>,
        public_key: PublicKey,
        allowed_ips: Option<Vec<IpNet>>,
        endpoint: Option<SocketAddr>,
    ) -> FfiResult<()> {
        let config = if let Some(ep) = endpoint {
            Arc::new(Vpn::new(public_key, ep))
                .apply_some(allowed_ips, Vpn::with_allowed_ips)
                .apply_some(identifier,  Vpn::with_identifier)
                .build()
        } else {
            Arc::new(Mesh::new(public_key))
                .apply_some(allowed_ips, Mesh::with_allowed_ips)
                .build()
        };
        self.connect_to_exit_node_with_config(config)
    }
```

and 

```
    pub fn connect_to_exit_node_postquantum(
        &self,
        identifier: Option<String>,
        public_key: PublicKey,
        allowed_ips: Option<Vec<IpNet>>,
        endpoint: SocketAddr,
    ) -> FfiResult<()> {
        let config = Arc::new(Vpn::new(public_key, endpoint))
            .apply_some(allowed_ips, Vpn::with_allowed_ips)
            .apply_some(identifier,  Vpn::with_identifier)
            .force_pq()
            .build();
        self.connect_to_exit_node_with_config(config)
    }
```

djkarwowski (commented):
Nice work! I'd just propose a small enhancement to allow for more fluent (and thus more readable, I believe) usage of the builder in our codebase as well.

+1.0

djkarwowski (approved):

